PR #676, opened 2022-08-31
Update duchy to handle reach-only requests

matthewclegg:
If reach_and_frequency.maximum_frequency_per_user is set to 1 in the MeasurementSpec for a ReachAndFrequency request, then an alternative path is followed for handling the request.  This alternative path differs from the standard path in the following ways:

1. The Duchy sets the key for each register to "destroyed" and sets the count to a random value.
2. Frequency-related noise registers are not added to the sketch.
3. After the reach value has been computed, the protocol terminates early and returns just this value.

- - -
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/676)

### DIFF
--- a/src/main/cc/wfa/measurement/common/crypto/constants.h
+++ b/src/main/cc/wfa/measurement/common/crypto/constants.h
@@ -57,7 +57,7 @@ inline const std::pair<std::string, std::string> kGenerateNewCompositeCipher = {
 inline const std::pair<std::string, std::string>
     kGenerateNewParitialCompositeCipher = {"", ""};
 // Elliptic curve to be used when none is specified.
-inline const int kDefaultEllipticCurveId = 415;
+inline const int kDefaultEllipticCurveId = 415; // NID_X9_62_prime256v1
 
 }  // namespace wfa::measurement::common::crypto
 
--- a/src/main/cc/wfa/measurement/common/crypto/constants.h
+++ b/src/main/cc/wfa/measurement/common/crypto/constants.h
@@ -56,6 +56,8 @@ inline const std::pair<std::string, std::string> kGenerateNewCompositeCipher = {
     "", ""};
 inline const std::pair<std::string, std::string>
     kGenerateNewParitialCompositeCipher = {"", ""};
+// Elliptic curve to be used when none is specified.
+inline const int kDefaultEllipticCurveId = 415;
 
 }  // namespace wfa::measurement::common::crypto
 
--- a/src/main/cc/wfa/measurement/common/crypto/constants.h
+++ b/src/main/cc/wfa/measurement/common/crypto/constants.h
@@ -57,7 +57,7 @@ inline const std::pair<std::string, std::string> kGenerateNewCompositeCipher = {
 inline const std::pair<std::string, std::string>
     kGenerateNewParitialCompositeCipher = {"", ""};
 // Elliptic curve to be used when none is specified.
-inline const int kDefaultEllipticCurveId = 415; // NID_X9_62_prime256v1
+inline const int kDefaultEllipticCurveId = 415;  // NID_X9_62_prime256v1
 
 }  // namespace wfa::measurement::common::crypto
 
--- a/src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/liquidlegionsv2/LiquidLegionsV2Mill.kt
+++ b/src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/liquidlegionsv2/LiquidLegionsV2Mill.kt
@@ -858,7 +858,8 @@ class LiquidLegionsV2Mill(
     val measurementSpec =
       MeasurementSpec.parseFrom(token.computationDetails.kingdomComputation.measurementSpec)
 		val measurementSpecMaximumFrequency: Int = measurementSpec.reachAndFrequency.maximumFrequencyPerUser
-		return min(llv2MaximumFrequency, measurementSpecMaximumFrequency)
+//		return min(llv2MaximumFrequency, measurementSpecMaximumFrequency)
+		return llv2MaximumFrequency
 	}
 
   companion object {
--- a/src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/liquidlegionsv2/LiquidLegionsV2Mill.kt
+++ b/src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/liquidlegionsv2/LiquidLegionsV2Mill.kt
@@ -719,33 +719,6 @@ class LiquidLegionsV2Mill(
     }
   }
 
-  private suspend fun sendResultToKingdom(
-    token: ComputationToken,
-    reach: Long,
-    frequency: Map<Long, Double>
-  ) {
-    val reachAndFrequency = ReachAndFrequency(reach, frequency)
-    val kingdomComputation = token.computationDetails.kingdomComputation
-    val serializedPublicApiEncryptionPublicKey: ByteString
-    val encryptedResult =
-      when (Version.fromString(kingdomComputation.publicApiVersion)) {
-        Version.V2_ALPHA -> {
-          val signedResult = signResult(reachAndFrequency.toV2AlphaMeasurementResult(), signingKey)
-          val publicApiEncryptionPublicKey =
-            kingdomComputation.measurementPublicKey.toV2AlphaEncryptionPublicKey()
-          serializedPublicApiEncryptionPublicKey = publicApiEncryptionPublicKey.toByteString()
-          encryptResult(signedResult, publicApiEncryptionPublicKey)
-        }
-        Version.VERSION_UNSPECIFIED -> error("Public api version is invalid or unspecified.")
-      }
-    sendResultToKingdom(
-      globalId = token.globalComputationId,
-      certificate = consentSignalCert,
-      resultPublicKey = serializedPublicApiEncryptionPublicKey,
-      encryptedResult = encryptedResult
-    )
-  }
-
   private suspend fun completeExecutionPhaseThreeAtAggregator(
     token: ComputationToken
   ): ComputationToken {
@@ -817,6 +790,33 @@ class LiquidLegionsV2Mill(
 
     // This duchy's responsibility for the computation is done. Mark it COMPLETED locally.
     return completeComputation(nextToken, CompletedReason.SUCCEEDED)
+  }
+
+  private suspend fun sendResultToKingdom(
+    token: ComputationToken,
+    reach: Long,
+    frequency: Map<Long, Double>
+  ) {
+    val reachAndFrequency = ReachAndFrequency(reach, frequency)
+    val kingdomComputation = token.computationDetails.kingdomComputation
+    val serializedPublicApiEncryptionPublicKey: ByteString
+    val encryptedResult =
+      when (Version.fromString(kingdomComputation.publicApiVersion)) {
+        Version.V2_ALPHA -> {
+          val signedResult = signResult(reachAndFrequency.toV2AlphaMeasurementResult(), signingKey)
+          val publicApiEncryptionPublicKey =
+            kingdomComputation.measurementPublicKey.toV2AlphaEncryptionPublicKey()
+          serializedPublicApiEncryptionPublicKey = publicApiEncryptionPublicKey.toByteString()
+          encryptResult(signedResult, publicApiEncryptionPublicKey)
+        }
+        Version.VERSION_UNSPECIFIED -> error("Public api version is invalid or unspecified.")
+      }
+    sendResultToKingdom(
+      globalId = token.globalComputationId,
+      certificate = consentSignalCert,
+      resultPublicKey = serializedPublicApiEncryptionPublicKey,
+      encryptedResult = encryptedResult
+    )
   }
 
   private fun nextDuchyStub(
--- a/src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/liquidlegionsv2/LiquidLegionsV2Mill.kt
+++ b/src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/liquidlegionsv2/LiquidLegionsV2Mill.kt
@@ -496,7 +496,10 @@ class LiquidLegionsV2Mill(
             combinedRegisterVector = readAndCombineAllInputBlobs(token, 1)
             totalSketchesCount = token.requisitionsCount
           }
-        if (llv2Parameters.noise.hasFrequencyNoiseConfig()) {
+        if (
+          llv2Parameters.noise.hasFrequencyNoiseConfig() &&
+            (getMaximumRequestedFrequency(token) > 1)
+        ) {
           requestBuilder.noiseParameters = getFrequencyNoiseParams(token, llv2Parameters)
         }
         val cryptoResult: CompleteExecutionPhaseOneAtAggregatorResponse =
@@ -600,7 +603,7 @@ class LiquidLegionsV2Mill(
             globalReachDpNoise = noiseConfig.reachNoiseConfig.globalReachDpNoise
           }
         }
-        if (noiseConfig.hasFrequencyNoiseConfig()) {
+        if (noiseConfig.hasFrequencyNoiseConfig() && (getMaximumRequestedFrequency(token) > 1)) {
           requestBuilder.frequencyNoiseParametersBuilder.apply {
             contributorsCount = workerStubs.size + 1
             maximumFrequency = getMaximumRequestedFrequency(token)
@@ -636,6 +639,12 @@ class LiquidLegionsV2Mill(
           )
           .token
       }
+
+    // If this is a reach-only computation, then our job is done.
+    if (getMaximumRequestedFrequency(token) == 1) {
+      sendResultToKingdom(token, reach, emptyMap<Long, Double>())
+      return completeComputation(nextToken, CompletedReason.SUCCEEDED)
+    }
 
     // Passes the computation to the next duchy.
     sendAdvanceComputationRequest(
@@ -673,7 +682,9 @@ class LiquidLegionsV2Mill(
         if (llv2Parameters.noise.hasFrequencyNoiseConfig()) {
           requestBuilder.apply {
             partialCompositeElGamalPublicKey = llv2Details.partiallyCombinedPublicKey
-            noiseParameters = getFrequencyNoiseParams(token, llv2Parameters)
+            if (getMaximumRequestedFrequency(token) > 1) {
+              noiseParameters = getFrequencyNoiseParams(token, llv2Parameters)
+            }
           }
         }
 
@@ -694,10 +705,42 @@ class LiquidLegionsV2Mill(
       stub = nextDuchyStub(llv2Details.participantList)
     )
 
-    return dataClients.transitionComputationToStage(
-      nextToken,
-      inputsToNextStage = nextToken.outputPathList(),
-      stage = Stage.WAIT_EXECUTION_PHASE_THREE_INPUTS.toProtocolStage()
+    if (getMaximumRequestedFrequency(token) == 1) {
+      // If this is a reach-only computation, then our job is done.
+      return completeComputation(nextToken, CompletedReason.SUCCEEDED)
+    } else {
+      return dataClients.transitionComputationToStage(
+        nextToken,
+        inputsToNextStage = nextToken.outputPathList(),
+        stage = Stage.WAIT_EXECUTION_PHASE_THREE_INPUTS.toProtocolStage()
+      )
+    }
+  }
+
+  private suspend fun sendResultToKingdom(
+    token: ComputationToken,
+    reach: Long,
+    frequency: Map<Long, Double>
+  ) {
+    val reachAndFrequency = ReachAndFrequency(reach, frequency)
+    val kingdomComputation = token.computationDetails.kingdomComputation
+    val serializedPublicApiEncryptionPublicKey: ByteString
+    val encryptedResult =
+      when (Version.fromString(kingdomComputation.publicApiVersion)) {
+        Version.V2_ALPHA -> {
+          val signedResult = signResult(reachAndFrequency.toV2AlphaMeasurementResult(), signingKey)
+          val publicApiEncryptionPublicKey =
+            kingdomComputation.measurementPublicKey.toV2AlphaEncryptionPublicKey()
+          serializedPublicApiEncryptionPublicKey = publicApiEncryptionPublicKey.toByteString()
+          encryptResult(signedResult, publicApiEncryptionPublicKey)
+        }
+        Version.VERSION_UNSPECIFIED -> error("Public api version is invalid or unspecified.")
+      }
+    sendResultToKingdom(
+      globalId = token.globalComputationId,
+      certificate = consentSignalCert,
+      resultPublicKey = serializedPublicApiEncryptionPublicKey,
+      encryptedResult = encryptedResult
     )
   }
 
@@ -733,29 +776,7 @@ class LiquidLegionsV2Mill(
       CompleteExecutionPhaseThreeAtAggregatorResponse.parseFrom(bytes.flatten())
         .frequencyDistributionMap
 
-    val reachAndFrequency =
-      ReachAndFrequency(llv2Details.reachEstimate.reach, frequencyDistributionMap)
-
-    val kingdomComputation = token.computationDetails.kingdomComputation
-    val serializedPublicApiEncryptionPublicKey: ByteString
-    val encryptedResult =
-      when (Version.fromString(kingdomComputation.publicApiVersion)) {
-        Version.V2_ALPHA -> {
-          val signedResult = signResult(reachAndFrequency.toV2AlphaMeasurementResult(), signingKey)
-          val publicApiEncryptionPublicKey =
-            kingdomComputation.measurementPublicKey.toV2AlphaEncryptionPublicKey()
-          serializedPublicApiEncryptionPublicKey = publicApiEncryptionPublicKey.toByteString()
-          encryptResult(signedResult, publicApiEncryptionPublicKey)
-        }
-        Version.VERSION_UNSPECIFIED -> error("Public api version is invalid or unspecified.")
-      }
-    sendResultToKingdom(
-      globalId = token.globalComputationId,
-      certificate = consentSignalCert,
-      resultPublicKey = serializedPublicApiEncryptionPublicKey,
-      encryptedResult = encryptedResult
-    )
-
+    sendResultToKingdom(token, llv2Details.reachEstimate.reach, frequencyDistributionMap)
     return completeComputation(nextToken, CompletedReason.SUCCEEDED)
   }
 
@@ -817,13 +838,13 @@ class LiquidLegionsV2Mill(
   }
 
   private fun ByteString.toCompleteSetupPhaseRequest(
-		token: ComputationToken,
+    token: ComputationToken,
     llv2Details: LiquidLegionsSketchAggregationV2.ComputationDetails,
     totalRequisitionsCount: Int
   ): CompleteSetupPhaseRequest {
     val noiseConfig = llv2Details.parameters.noise
     val requestBuilder = CompleteSetupPhaseRequest.newBuilder().setCombinedRegisterVector(this)
-		requestBuilder.setMaximumFrequency(getMaximumRequestedFrequency(token))
+    requestBuilder.setMaximumFrequency(getMaximumRequestedFrequency(token))
     if (noiseConfig.hasReachNoiseConfig()) {
       requestBuilder.noiseParametersBuilder.apply {
         compositeElGamalPublicKey = llv2Details.combinedPublicKey
@@ -841,7 +862,7 @@ class LiquidLegionsV2Mill(
   }
 
   private fun getFrequencyNoiseParams(
-		token: ComputationToken,
+    token: ComputationToken,
     llv2Parameters: Parameters
   ): FlagCountTupleNoiseGenerationParameters {
     return FlagCountTupleNoiseGenerationParameters.newBuilder()
@@ -853,14 +874,17 @@ class LiquidLegionsV2Mill(
       .build()
   }
 
-	private fun getMaximumRequestedFrequency(token: ComputationToken): Int {
-		val llv2MaximumFrequency : Int = token.computationDetails.liquidLegionsV2.parameters.maximumFrequency
+  private fun getMaximumRequestedFrequency(token: ComputationToken): Int {
+    var maximumRequestedFrequency =
+      token.computationDetails.liquidLegionsV2.parameters.maximumFrequency
     val measurementSpec =
       MeasurementSpec.parseFrom(token.computationDetails.kingdomComputation.measurementSpec)
-		val measurementSpecMaximumFrequency: Int = measurementSpec.reachAndFrequency.maximumFrequencyPerUser
-//		return min(llv2MaximumFrequency, measurementSpecMaximumFrequency)
-		return llv2MaximumFrequency
-	}
+    val measurementSpecMaximumFrequency = measurementSpec.reachAndFrequency.maximumFrequencyPerUser
+    if (measurementSpecMaximumFrequency > 0) {
+      maximumRequestedFrequency = min(maximumRequestedFrequency, measurementSpecMaximumFrequency)
+    }
+    return maximumRequestedFrequency
+  }
 
   companion object {
     init {
--- a/src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/liquidlegionsv2/LiquidLegionsV2Mill.kt
+++ b/src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/liquidlegionsv2/LiquidLegionsV2Mill.kt
@@ -17,6 +17,7 @@ package org.wfanet.measurement.duchy.daemon.mill.liquidlegionsv2
 import com.google.protobuf.ByteString
 import java.nio.file.Paths
 import java.time.Clock
+import kotlin.math.min
 import org.wfanet.anysketch.crypto.CombineElGamalPublicKeysRequest
 import org.wfanet.measurement.api.Version
 import org.wfanet.measurement.api.v2alpha.MeasurementSpec
@@ -425,7 +426,7 @@ class LiquidLegionsV2Mill(
           dataClients
             .readAllRequisitionBlobs(token, duchyId)
             .concat(readAndCombineAllInputBlobs(token, workerStubs.size))
-            .toCompleteSetupPhaseRequest(llv2Details, token.requisitionsCount)
+            .toCompleteSetupPhaseRequest(token, llv2Details, token.requisitionsCount)
         val cryptoResult: CompleteSetupPhaseResponse = cryptoWorker.completeSetupPhase(request)
         logStageDurationMetric(token, CRYPTO_LIB_CPU_DURATION, cryptoResult.elapsedCpuTimeMillis)
         cryptoResult.combinedRegisterVector
@@ -456,7 +457,7 @@ class LiquidLegionsV2Mill(
         val request =
           dataClients
             .readAllRequisitionBlobs(token, duchyId)
-            .toCompleteSetupPhaseRequest(llv2Details, token.requisitionsCount)
+            .toCompleteSetupPhaseRequest(token, llv2Details, token.requisitionsCount)
         val cryptoResult: CompleteSetupPhaseResponse = cryptoWorker.completeSetupPhase(request)
         logStageDurationMetric(token, CRYPTO_LIB_CPU_DURATION, cryptoResult.elapsedCpuTimeMillis)
         cryptoResult.combinedRegisterVector
@@ -496,7 +497,7 @@ class LiquidLegionsV2Mill(
             totalSketchesCount = token.requisitionsCount
           }
         if (llv2Parameters.noise.hasFrequencyNoiseConfig()) {
-          requestBuilder.noiseParameters = getFrequencyNoiseParams(llv2Parameters)
+          requestBuilder.noiseParameters = getFrequencyNoiseParams(token, llv2Parameters)
         }
         val cryptoResult: CompleteExecutionPhaseOneAtAggregatorResponse =
           cryptoWorker.completeExecutionPhaseOneAtAggregator(requestBuilder.build())
@@ -585,7 +586,7 @@ class LiquidLegionsV2Mill(
             compositeElGamalPublicKey = llv2Details.combinedPublicKey
             curveId = llv2Parameters.ellipticCurveId.toLong()
             flagCountTuples = readAndCombineAllInputBlobs(token, 1)
-            maximumFrequency = llv2Parameters.maximumFrequency
+            maximumFrequency = getMaximumRequestedFrequency(token)
             liquidLegionsParametersBuilder.apply {
               decayRate = llv2Parameters.liquidLegionsSketch.decayRate
               size = llv2Parameters.liquidLegionsSketch.size
@@ -602,7 +603,7 @@ class LiquidLegionsV2Mill(
         if (noiseConfig.hasFrequencyNoiseConfig()) {
           requestBuilder.frequencyNoiseParametersBuilder.apply {
             contributorsCount = workerStubs.size + 1
-            maximumFrequency = llv2Parameters.maximumFrequency
+            maximumFrequency = getMaximumRequestedFrequency(token)
             dpParams = noiseConfig.frequencyNoiseConfig
           }
         }
@@ -672,7 +673,7 @@ class LiquidLegionsV2Mill(
         if (llv2Parameters.noise.hasFrequencyNoiseConfig()) {
           requestBuilder.apply {
             partialCompositeElGamalPublicKey = llv2Details.partiallyCombinedPublicKey
-            noiseParameters = getFrequencyNoiseParams(llv2Parameters)
+            noiseParameters = getFrequencyNoiseParams(token, llv2Parameters)
           }
         }
 
@@ -714,7 +715,7 @@ class LiquidLegionsV2Mill(
             localElGamalKeyPair = llv2Details.localElgamalKey
             curveId = llv2Parameters.ellipticCurveId.toLong()
             sameKeyAggregatorMatrix = readAndCombineAllInputBlobs(token, 1)
-            maximumFrequency = llv2Parameters.maximumFrequency
+            maximumFrequency = getMaximumRequestedFrequency(token)
           }
         if (llv2Parameters.noise.hasFrequencyNoiseConfig()) {
           requestBuilder.globalFrequencyDpNoisePerBucketBuilder.apply {
@@ -816,11 +817,13 @@ class LiquidLegionsV2Mill(
   }
 
   private fun ByteString.toCompleteSetupPhaseRequest(
+		token: ComputationToken,
     llv2Details: LiquidLegionsSketchAggregationV2.ComputationDetails,
     totalRequisitionsCount: Int
   ): CompleteSetupPhaseRequest {
     val noiseConfig = llv2Details.parameters.noise
     val requestBuilder = CompleteSetupPhaseRequest.newBuilder().setCombinedRegisterVector(this)
+		requestBuilder.setMaximumFrequency(getMaximumRequestedFrequency(token))
     if (noiseConfig.hasReachNoiseConfig()) {
       requestBuilder.noiseParametersBuilder.apply {
         compositeElGamalPublicKey = llv2Details.combinedPublicKey
@@ -838,16 +841,25 @@ class LiquidLegionsV2Mill(
   }
 
   private fun getFrequencyNoiseParams(
+		token: ComputationToken,
     llv2Parameters: Parameters
   ): FlagCountTupleNoiseGenerationParameters {
     return FlagCountTupleNoiseGenerationParameters.newBuilder()
       .apply {
-        maximumFrequency = llv2Parameters.maximumFrequency
+        maximumFrequency = getMaximumRequestedFrequency(token)
         contributorsCount = workerStubs.size + 1
         dpParams = llv2Parameters.noise.frequencyNoiseConfig
       }
       .build()
   }
+
+	private fun getMaximumRequestedFrequency(token: ComputationToken): Int {
+		val llv2MaximumFrequency : Int = token.computationDetails.liquidLegionsV2.parameters.maximumFrequency
+    val measurementSpec =
+      MeasurementSpec.parseFrom(token.computationDetails.kingdomComputation.measurementSpec)
+		val measurementSpecMaximumFrequency: Int = measurementSpec.reachAndFrequency.maximumFrequencyPerUser
+		return min(llv2MaximumFrequency, measurementSpecMaximumFrequency)
+	}
 
   companion object {
     init {
--- a/src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/liquidlegionsv2/LiquidLegionsV2Mill.kt
+++ b/src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/liquidlegionsv2/LiquidLegionsV2Mill.kt
@@ -575,6 +575,7 @@ class LiquidLegionsV2Mill(
       "invalid role for this function."
     }
     var reach = 0L
+    val maximumRequestedFrequency = getMaximumRequestedFrequency(token)
     val (bytes, tempToken) =
       existingOutputOr(token) {
         when (Version.fromString(token.computationDetails.kingdomComputation.publicApiVersion)) {
@@ -589,7 +590,7 @@ class LiquidLegionsV2Mill(
             compositeElGamalPublicKey = llv2Details.combinedPublicKey
             curveId = llv2Parameters.ellipticCurveId.toLong()
             flagCountTuples = readAndCombineAllInputBlobs(token, 1)
-            maximumFrequency = getMaximumRequestedFrequency(token)
+            maximumFrequency = maximumRequestedFrequency
             liquidLegionsParametersBuilder.apply {
               decayRate = llv2Parameters.liquidLegionsSketch.decayRate
               size = llv2Parameters.liquidLegionsSketch.size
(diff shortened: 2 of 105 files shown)


src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/liquidlegionsv2/LiquidLegionsV2Mill.kt
```diff
@@ -636,6 +640,12 @@ class LiquidLegionsV2Mill(
           .token
       }
 
+    // If this is a reach-only computation, then our job is done.
+    if (getMaximumRequestedFrequency(token) == 1) {
+      sendResultToKingdom(token, reach, emptyMap<Long, Double>())
```
pasin30055, line 645: Would it make more sense to output a map from 1 to 1 (i.e. 100% has frequency >= 1) here?

src/test/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/liquid_legions_v2_encryption_utility_test.cc
```diff
@@ -687,6 +687,66 @@ TEST(CompleteSetupPhase, WrongInputSketchSizeShouldThrow) {
               StatusIs(absl::StatusCode::kInvalidArgument, "not divisible"));
 }
```
pasin30055, line 689: Would it make sense to also add a unit test for `DestroyKeysAndCounts`?

pasin30055 (commented):
Thanks Matt. This looks good overall; I left a few (mostly minor) comments.

renjiezh (commented):
> Reviewed 3 of 8 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/676)* status: 3 of 8 files reviewed, 9 unresolved discussions (waiting on @matthewclegg, @pasin30055, and @renjiezh)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/liquidlegionsv2/LiquidLegionsV2Mill.kt` line 592 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/676#-NAzKboo31LE5Os2e3B4:-NAzKboo31LE5Os2e3B5:b-t4emii) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/505f7ca22a969fd1da049baded1a1d2ebab01c37/src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/liquidlegionsv2/LiquidLegionsV2Mill.kt#L592)):*
> ```Kotlin
>             curveId = llv2Parameters.ellipticCurveId.toLong()
>             flagCountTuples = readAndCombineAllInputBlobs(token, 1)
>             maximumFrequency = getMaximumRequestedFrequency(token)
> ```

nit: getMaximumRequestedFrequency() is called several times. It might be better to extract it to a variable `maximumRequestedFreqency = getMaximumRequestedFrequency(token)`
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/liquidlegionsv2/LiquidLegionsV2Mill.kt` line 645 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/676#-NAlkU8LX3KHjyLT5CwH-r1-645:-NAzOk-XCnLZVGsQVyR1:b32v57) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/505f7ca22a969fd1da049baded1a1d2ebab01c37/src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/liquidlegionsv2/LiquidLegionsV2Mill.kt#L645)):*
<details><summary><i>Previously, pasin30055 (Pasin Manurangsi) wrote…</i></summary><blockquote>

Would it make more sense to output a map from 1 to 1 (i.e. 100% has frequency >= 1) here?
</blockquote></details>

+1
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/liquidlegionsv2/LiquidLegionsV2Mill.kt` line 846 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/676#-NAzC1fQ-UWeGJQ4Pxrj:-NAzC1fQ-UWeGJQ4Pxrk:bspfv9m) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/505f7ca22a969fd1da049baded1a1d2ebab01c37/src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/liquidlegionsv2/LiquidLegionsV2Mill.kt#L846)):*
> ```Kotlin
>   ): CompleteSetupPhaseRequest {
>     val noiseConfig = llv2Details.parameters.noise
>     val requestBuilder = CompleteSetupPhaseRequest.newBuilder().setCombinedRegisterVector(this)
> ```

nit: Use Dsl
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/frontend/FrontendSimulator.kt` line 217 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/676#-NAz6zjx7MrRS3dkh6Ub:-NAz6zjx7MrRS3dkh6Uc:b-lf8ayh) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/505f7ca22a969fd1da049baded1a1d2ebab01c37/src/main/kotlin/org/wfanet/measurement/loadtest/frontend/FrontendSimulator.kt#L217)):*
> ```Kotlin
>     var reachOnlyResult = getReachAndFrequencyResult(createdReachOnlyMeasurement.name)
>     var nAttempts = 0
>     while (reachOnlyResult == null && (nAttempts < 4)) {
> ```

I saw other cases did not add the limit of attempts. Will this bring risk to break the test?
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/frontend/FrontendSimulator.kt` line 223 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/676#-NAz80TT8bzjjx3qcBVD:-NAz80TT8bzjjx3qcBVE:b-96v2lr) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/505f7ca22a969fd1da049baded1a1d2ebab01c37/src/main/kotlin/org/wfanet/measurement/loadtest/frontend/FrontendSimulator.kt#L223)):*
> ```Kotlin
>       reachOnlyResult = getReachAndFrequencyResult(createdReachOnlyMeasurement.name)
>     }
>     if (reachOnlyResult == null) {
> ```

nit: when assertion fails, we will have information of the failure. Guess logger does not necessarily need to print a message. Thus `assertThat(reachOnlyResult).isNotNull()` might suffice.

matthewclegg (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/676)* status: 2 of 8 files reviewed, 8 unresolved discussions (waiting on @pasin30055 and @renjiezh)

___
*[`src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/liquid_legions_v2_encryption_utility.cc` line 583 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/676#-NAlkU8LX3KHjyLT5CwG-r1-583:-NB3EVpV0-sUrKmruZTr:bebp470) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/505f7ca22a969fd1da049baded1a1d2ebab01c37/src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/liquid_legions_v2_encryption_utility.cc#L583)):*
<details><summary><i>Previously, pasin30055 (Pasin Manurangsi) wrote…</i></summary><blockquote>

I'm not sure what this second case means. I.e. when do we omit noise parameters?
</blockquote></details>

As far as I can tell, the noise parameters are only omitted in debugging.  In production, these parameters are always included.  I have added a comment to that effect.
___
*[`src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/liquid_legions_v2_encryption_utility.cc` line 628 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/676#-NAlkU8LX3KHjyLT5CwG-r1-628:-NB3Em905ouH42hljSLs:b-af5dko) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/505f7ca22a969fd1da049baded1a1d2ebab01c37/src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/liquid_legions_v2_encryption_utility.cc#L628)):*
<details><summary><i>Previously, pasin30055 (Pasin Manurangsi) wrote…</i></summary><blockquote>

Is this `random_count` string used anywhere?
</blockquote></details>

Ah, thanks for finding that.  I meant to use 'random_count' in place of 'register_id' in line 631.
___
*[`src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/liquid_legions_v2_encryption_utility.cc` line 630 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/676#-NAlkU8LX3KHjyLT5CwG-r1-630:-NB3F4ATBo75x3xpYvYO:b-ovahg3) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/505f7ca22a969fd1da049baded1a1d2ebab01c37/src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/liquid_legions_v2_encryption_utility.cc#L630)):*
<details><summary><i>Previously, pasin30055 (Pasin Manurangsi) wrote…</i></summary><blockquote>

(Also I guess we don't even fully run SKA, so there shouldn't be any problem with not randomizing the count, right?)
</blockquote></details>

I suppose that if we are bypassing SKA, then there is probably no danger of leaking the number of publishers that contributed to a given register.  On the other hand, if we randomize the register counts, it seems to give an extra level of assurance.  Do you think there could be a problem with this?
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/liquidlegionsv2/LiquidLegionsV2Mill.kt` line 645 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/676#-NAlkU8LX3KHjyLT5CwH-r1-645:-NB3INTu6V-uqL5yFKGc:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/505f7ca22a969fd1da049baded1a1d2ebab01c37/src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/liquidlegionsv2/LiquidLegionsV2Mill.kt#L645)):*

+1
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/liquidlegionsv2/LiquidLegionsV2Mill.kt` line 846 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/676#-NAzC1fQ-UWeGJQ4Pxrj:-NB3Qkqk7fZC9kFh1WtX:bhyryhu) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/505f7ca22a969fd1da049baded1a1d2ebab01c37/src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/liquidlegionsv2/LiquidLegionsV2Mill.kt#L846)):*

nit: Use Dsl
</blockquote></details>

This is part of the legacy code, so I hope you don't mind if I leave it as is.  In any event, and please correct me if I am wrong, but I don't think that DSL will work in this case.  In this context `this` refers to the receiver argument of the function.  I think if you were to try to switch to DSL, `this` would refer to the enclosing Builder object.
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/frontend/FrontendSimulator.kt` line 217 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/676#-NAz6zjx7MrRS3dkh6Ub:-NB3KTW565A6Lg4Amn6S:b-658c22) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/505f7ca22a969fd1da049baded1a1d2ebab01c37/src/main/kotlin/org/wfanet/measurement/loadtest/frontend/FrontendSimulator.kt#L217)):*

I saw other cases did not add the limit of attempts. Will this bring risk to break the test?
</blockquote></details>

In my view, it's the other tests that are broken. ;). If the number of attempts is not limited, then there is a potential for an infinite loop.  This could happen if there is a bug in the Duchy code.  In that case, the unit test would never complete.  A better behavior is that the unit test should eventually time out and generate a failure.
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/frontend/FrontendSimulator.kt` line 223 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/676#-NAz80TT8bzjjx3qcBVD:-NB3RxqHC6Tjs83uakxf:bcb8a5f) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/505f7ca22a969fd1da049baded1a1d2ebab01c37/src/main/kotlin/org/wfanet/measurement/loadtest/frontend/FrontendSimulator.kt#L223)):*

nit: when assertion fails, we will have information of the failure. Guess logger does not necessarily need to print a message. Thus `assertThat(reachOnlyResult).isNotNull()` might suffice.
</blockquote></details>

Thanks, I have changed the assertion as you have suggested.  I also kept the enclosing if { ... }.  The reason for this is to allow Kotlin to infer that reachOnlyResult cannot be null in the call to `assertDpResultsEqual()` later in this function.
___
*[`src/test/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/liquid_legions_v2_encryption_utility_test.cc` line 689 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/676#-NAlkU8LX3KHjyLT5CwL-r1-689:-NB3MMbw3bASCqX57Eru:b-nwch3l) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/505f7ca22a969fd1da049baded1a1d2ebab01c37/src/test/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/liquid_legions_v2_encryption_utility_test.cc#L689)):*
<details><summary><i>Previously, pasin30055 (Pasin Manurangsi) wrote…</i></summary><blockquote>

Would it make sense to also add a unit test for `DestroyKeysAndCounts`?
</blockquote></details>

That is essentially what is tested in `FrequencyOneWorksAsExpectedWithoutNoise` and `FrequencyOneWorksAsExpectedWithNoise`.  If you think there are other things that should be tested, please let me know.

renjiezh (approved):
:lgtm:

> Reviewed 2 of 8 files at r1, 1 of 3 files at r2.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/676#-:-NB4W9-s-Ubi7-60AXLb:bnfp4nl)* status: 4 of 8 files reviewed, 6 unresolved discussions (waiting on @matthewclegg, @pasin30055, and @renjiezh)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/frontend/FrontendSimulator.kt` line 223 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/676#-NAz80TT8bzjjx3qcBVD:-NB4VqE_DM2BDYYHwKEB:b2p8shw) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/505f7ca22a969fd1da049baded1a1d2ebab01c37/src/main/kotlin/org/wfanet/measurement/loadtest/frontend/FrontendSimulator.kt#L223)):*

Thanks, I have changed the assertion as you have suggested.  I also kept the enclosing if { ... }.  The reason for this is to allow Kotlin to infer that reachOnlyResult cannot be null in the call to `assertDpResultsEqual()` later in this function.
</blockquote></details>

One alternative is to use `reachOnlyResult!!`afterwards. But I think the choice is from you.

pasin30055 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/676)* status: 4 of 8 files reviewed, 2 unresolved discussions (waiting on @matthewclegg, @pasin30055, and @renjiezh)

___
*[`src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/liquid_legions_v2_encryption_utility.cc` line 583 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/676#-NAlkU8LX3KHjyLT5CwG-r1-583:-NB9--Kf15_Wt4G_HhR9:b753b8h) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/505f7ca22a969fd1da049baded1a1d2ebab01c37/src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/liquid_legions_v2_encryption_utility.cc#L583)):*

As far as I can tell, the noise parameters are only omitted in debugging.  In production, these parameters are always included.  I have added a comment to that effect.
</blockquote></details>

I see. Thanks for the clarification. This is fine then.
___
*[`src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/liquid_legions_v2_encryption_utility.cc` line 628 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/676#-NAlkU8LX3KHjyLT5CwG-r1-628:-NB9-g-B4ICquyHVsvfM:b-nl2bop) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/505f7ca22a969fd1da049baded1a1d2ebab01c37/src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/liquid_legions_v2_encryption_utility.cc#L628)):*

Ah, thanks for finding that.  I meant to use 'random_count' in place of 'register_id' in line 631.
</blockquote></details>

LGTM now. Thanks!
___
*[`src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/liquid_legions_v2_encryption_utility.cc` line 630 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/676#-NAlkU8LX3KHjyLT5CwG-r1-630:-NB9-N2x71Wva9_e3zui:b-dyakug) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/505f7ca22a969fd1da049baded1a1d2ebab01c37/src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/liquid_legions_v2_encryption_utility.cc#L630)):*

I suppose that if we are bypassing SKA, then there is probably no danger of leaking the number of publishers that contributed to a given register.  On the other hand, if we randomize the register counts, it seems to give an extra level of assurance.  Do you think there could be a problem with this?
</blockquote></details>

No, I don't think there is any problem with randomizing both, so the current code is fine.
___
*[`src/test/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/liquid_legions_v2_encryption_utility_test.cc` line 689 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/676#-NAlkU8LX3KHjyLT5CwL-r1-689:-NB90cFd1pu1zw3Lnfzy:b-hlqskb) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/505f7ca22a969fd1da049baded1a1d2ebab01c37/src/test/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/liquid_legions_v2_encryption_utility_test.cc#L689)):*

That is essentially what is tested in `FrequencyOneWorksAsExpectedWithoutNoise` and `FrequencyOneWorksAsExpectedWithNoise`.  If you think there are other things that should be tested, please let me know.
</blockquote></details>

The current tests LGTM. Thanks.

pasin30055 (approved):

wangyaopw (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/676)* status: 4 of 8 files reviewed, 7 unresolved discussions (waiting on @benjaminkreuter, @duliomatos1, @matthewclegg, and @renjiezh)

___
*[`src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/liquid_legions_v2_encryption_utility.cc` line 564 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/676#-NBY8Eai8cFDYzlwRjup:-NBY8Eai8cFDYzlwRjuq:b-lmfwtc) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6029678bb4bb187b65f7d17439076be80a3c682c/src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/liquid_legions_v2_encryption_utility.cc#L564)):*
> ```cpp
> 
> // Copies encrypted registers, replacing all keys with
> // the destroyed register flag and all counts with random values.
> ```

I feel this is not necessary.

The count doesn't need to be destroyed since it is not used anywhere at all and we are not leaking anything if we don't destroy it.

For the keys, we are still adding noise for the reach. If we are fine with the current LLV2 protocol, there is nothing extra we need to hide for the reach only protocol.



_Code quote:_
```cpp
DestroyKeysAndCounts(
```
___
*[`src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/liquid_legions_v2_encryption_utility.cc` line 573 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/676#-NBY0IoL83POKfU93LIg:-NBY0IoL83POKfU93LIh:bhc1qpv) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6029678bb4bb187b65f7d17439076be80a3c682c/src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/liquid_legions_v2_encryption_utility.cc#L573)):*
> ```cpp
> 
>   if (register_count == 0) {
>     return dest;
> ```

nit: just move this check after `  std::string source = request.combined_register_vector();` and check if source is empty.



_Code quote:_
```cpp
  if (register_count == 0) {
    return dest;
  }
```
___
*[`src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/liquid_legions_v2_encryption_utility.cc` line 831 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/676#-NBYBTby9vZUr6kebaoL:-NBYBTby9vZUr6kebaoM:b-6h561c) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6029678bb4bb187b65f7d17439076be80a3c682c/src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/liquid_legions_v2_encryption_utility.cc#L831)):*
> ```cpp
>   // Add noise (flag_a, flag_b, flag_c, count) tuples if configured to.
>     RETURN_IF_ERROR(AddAllFrequencyNoise(*protocol_cryptor, request.curve_id(),
> ```

You are still adding frequency noise when the maximum_frequency is set to 1.

But we can not just simply stop adding this noise here, since when we compensate for frequency noise when estimating reach at the aggregator. If we modify here, we need to adjust the code for reach estimate too.
___
*[`src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/liquid_legions_v2_encryption_utility.cc` line 884 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/676#-NBYB1_9DmS7ujt-bm0w:-NBYB1_9DmS7ujt-bm0x:b-6h561c) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6029678bb4bb187b65f7d17439076be80a3c682c/src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/liquid_legions_v2_encryption_utility.cc#L884)):*
> ```cpp
>   // Add noise (flag_a, flag_b, flag_c, count) tuples if configured to.
>     RETURN_IF_ERROR(AddAllFrequencyNoise(*protocol_cryptor, request.curve_id(),
> ```

You are still adding frequency noise when the maximum_frequency is set to 1.

But we can not just simply stop adding this noise here, since when we compensate for frequency noise when estimating reach at the aggregator. If we modify here, we need to adjust the code for reach estimate too.



_Code quote:_
```cpp
AddAllFrequencyNoise
```
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/liquidlegionsv2/LiquidLegionsV2Mill.kt` line 722 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/676#-NBY1pSm1vK_NvzE1_jZ:-NBY1pSm1vK_NvzE1_j_:b-uahua4) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6029678bb4bb187b65f7d17439076be80a3c682c/src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/liquidlegionsv2/LiquidLegionsV2Mill.kt#L722)):*
> ```Kotlin
>   }
> 
>   private suspend fun sendResultToKingdom(
> ```

nit: move this function to the bottom, so all `completeExecutionPhaseXAtY()` methods are together



_Code quote:_
```Kotlin
  private suspend fun sendResultToKingdom(
```

src/main/cc/wfa/measurement/common/crypto/constants.h
```diff
@@ -56,6 +56,8 @@ inline const std::pair<std::string, std::string> kGenerateNewCompositeCipher = {
     "", ""};
 inline const std::pair<std::string, std::string>
     kGenerateNewParitialCompositeCipher = {"", ""};
+// Elliptic curve to be used when none is specified.
```
benjaminkreuter, line 59: We should indicate what curve this is (the name of the curve), right now it is just a magic number.

kungfucraig (commented):
> Reviewed 3 of 8 files at r1, 1 of 3 files at r2.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/676)* status: 3 of 8 files reviewed, 10 unresolved discussions (waiting on @duliomatos1, @kungfucraig, @matthewclegg, @renjiezh, and @wangyaopw)

___
*[`src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/liquid_legions_v2_encryption_utility.cc` line 564 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/676#-NBY8Eai8cFDYzlwRjup:-NBYHIFR9YPkmeRw810h:bpk17z9) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6029678bb4bb187b65f7d17439076be80a3c682c/src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/liquid_legions_v2_encryption_utility.cc#L564)):*
<details><summary><i>Previously, wangyaopw (Yao Wang) wrote…</i></summary><blockquote>

I feel this is not necessary.

The count doesn't need to be destroyed since it is not used anywhere at all and we are not leaking anything if we don't destroy it.

For the keys, we are still adding noise for the reach. If we are fine with the current LLV2 protocol, there is nothing extra we need to hide for the reach only protocol.
</blockquote></details>

I think the reason to do this would be to be able to guarantee frequency is not revealed, even to a worker.
___
*[`src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/liquid_legions_v2_encryption_utility.cc` line 671 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/676#-NBZDV8a2wEM9ftZKQq5:-NBZDV8a2wEM9ftZKQq6:bucui6e) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6029678bb4bb187b65f7d17439076be80a3c682c/src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/liquid_legions_v2_encryption_utility.cc#L671)):*
> ```cpp
> 
>   if (request.maximum_frequency() == 1) {
>     *response_crv = *DestroyKeysAndCounts(request);
> ```

Why not just strip them when the protocol starts (i.e. setup phase) and you can stop carrying them around? I appreciate this may be a much more disruptive change and that it may not be desirable if you're trying to get this done as quick as possible. 
___
*[`src/main/proto/wfa/measurement/internal/duchy/protocol/liquid_legions_v2_encryption_methods.proto` line 115 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/676#-NBZCh9J44k7wv5Z0-q9:-NBZCh9J44k7wv5Z0-qA:b-zgxexd) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6029678bb4bb187b65f7d17439076be80a3c682c/src/main/proto/wfa/measurement/internal/duchy/protocol/liquid_legions_v2_encryption_methods.proto#L115)):*
> ```protobuf
>   // The maximum frequency that should be computed.
>   // If set to 1, then no frequency histogram is returned.
>   int32 maximum_frequency = 3;
> ```

Using in band signaling, in this case measurement frequency, to indicate fundamentally different behavior is generally undesirable. It's better to be explicit. A field called "compute_reach_only" would make more sense. That said, setting a maximum frequency has its own usefulness.

https://www.atlasobscura.com/articles/capn-crunch-whistle :)

kungfucraig (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/676)* status: 3 of 8 files reviewed, 10 unresolved discussions (waiting on @duliomatos1, @kungfucraig, @matthewclegg, @renjiezh, and @wangyaopw)

___
*[`src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/liquid_legions_v2_encryption_utility.cc` line 564 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/676#-NBY8Eai8cFDYzlwRjup:-NBZEHhm2S2xH34YtHqN:bku3207) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6029678bb4bb187b65f7d17439076be80a3c682c/src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/liquid_legions_v2_encryption_utility.cc#L564)):*

I think the reason to do this would be to be able to guarantee frequency is not revealed, even to a worker.
</blockquote></details>

Consider computing frequency at some later time from the residue of the reach calculation. Virtual "shredding" of the counts and keys precludes this.

wangyaopw (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/676)* status: 2 of 8 files reviewed, 10 unresolved discussions (waiting on @duliomatos1, @kungfucraig, @matthewclegg, and @renjiezh)

___
*[`src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/liquid_legions_v2_encryption_utility.cc` line 564 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/676#-NBY8Eai8cFDYzlwRjup:-NBZIEOl5q0G2lG_34L2:b-jdqyk9) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6029678bb4bb187b65f7d17439076be80a3c682c/src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/liquid_legions_v2_encryption_utility.cc#L564)):*

Consider computing frequency at some later time from the residue of the reach calculation. Virtual "shredding" of the counts and keys precludes this.
</blockquote></details>

For reach only measurement, all workers will mark the computation COMPLETED after phase two. Unless all workers collude (a case we don't consider), we don't need to worry about any worker can learn the frequency.

As a temporary solution, I feel there is no need to add such precaution.

matthewclegg (commented):
> Dismissed @pasin30055 from a discussion.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/676)* status: 2 of 8 files reviewed, 8 unresolved discussions (waiting on @benjaminkreuter, @duliomatos1, @kungfucraig, @renjiezh, and @wangyaopw)

___
*[`src/main/cc/wfa/measurement/common/crypto/constants.h` line 59 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/676#-NAlkU8LX3KHjyLT5CwF-r3-59:-NBZKSCU1038w4cI10Jq:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0ebc78c2d6691562a4109125d89555f336f1da0a/src/main/cc/wfa/measurement/common/crypto/constants.h#L59)):*
<details><summary><i>Previously, benjaminkreuter wrote…</i></summary><blockquote>

We should indicate what curve this is (the name of the curve), right now it is just a magic number.
</blockquote></details>

Done.
___
*[`src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/liquid_legions_v2_encryption_utility.cc` line 564 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/676#-NBY8Eai8cFDYzlwRjup:-NBYzL70--wJaHBwvf6M:b2u5zeo) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6029678bb4bb187b65f7d17439076be80a3c682c/src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/liquid_legions_v2_encryption_utility.cc#L564)):*
<details><summary><i>Previously, wangyaopw (Yao Wang) wrote…</i></summary><blockquote>

I feel this is not necessary.

The count doesn't need to be destroyed since it is not used anywhere at all and we are not leaking anything if we don't destroy it.

For the keys, we are still adding noise for the reach. If we are fine with the current LLV2 protocol, there is nothing extra we need to hide for the reach only protocol.
</blockquote></details>

To me it seemed like randomizing the counts added an extra level of safety to the code.  After this function returns, one can be mathematically certain that there is no danger of information about the counts being leaked by subsequent operations, even if there are bugs in the code or someone in the future accidentally changes the code.  Nonetheless, you are the second person to make this comment.  If you feel strongly about this, let me know and I will remove the code that randomizes the counts.
___
*[`src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/liquid_legions_v2_encryption_utility.cc` line 573 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/676#-NBY0IoL83POKfU93LIg:-NBZ-rTh9VRU2GHqybLR:b6j42ih) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6029678bb4bb187b65f7d17439076be80a3c682c/src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/liquid_legions_v2_encryption_utility.cc#L573)):*
<details><summary><i>Previously, wangyaopw (Yao Wang) wrote…</i></summary><blockquote>

nit: just move this check after `  std::string source = request.combined_register_vector();` and check if source is empty.
</blockquote></details>

I made this change, but I am just wondering, could you explain how this improves the code?
___
*[`src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/liquid_legions_v2_encryption_utility.cc` line 671 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/676#-NBZDV8a2wEM9ftZKQq5:-NBZJPeL9GrNn_dZagPT:bro1boo) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6029678bb4bb187b65f7d17439076be80a3c682c/src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/liquid_legions_v2_encryption_utility.cc#L671)):*

Why not just strip them when the protocol starts (i.e. setup phase) and you can stop carrying them around? I appreciate this may be a much more disruptive change and that it may not be desirable if you're trying to get this done as quick as possible. 
</blockquote></details>

Agreed, that would be a much better long-term solution.  But if you are going to do that, one might as well take the extra step of actually creating a separate reach-only protocol.
___
*[`src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/liquid_legions_v2_encryption_utility.cc` line 831 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/676#-NBYBTby9vZUr6kebaoL:-NBZ0R3T1FEIdGD1NOz2:b-2nwm77) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6029678bb4bb187b65f7d17439076be80a3c682c/src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/liquid_legions_v2_encryption_utility.cc#L831)):*
<details><summary><i>Previously, wangyaopw (Yao Wang) wrote…</i></summary><blockquote>

You are still adding frequency noise when the maximum_frequency is set to 1.

But we can not just simply stop adding this noise here, since when we compensate for frequency noise when estimating reach at the aggregator. If we modify here, we need to adjust the code for reach estimate too.
</blockquote></details>

If the frequency is 1, then `completeExecutePhaseOneAtAggregator` in `LiquidLegionsV2Mill.kt` will not set the `noise_parameters` field.  Hence, the test at line 830 will be false and line 831 will not be executed.  See line 501 of `LiquidLegionsV2Mill.kt`.

The reach estimate is adjusted at line 945 in this file.  Lines 944-945 now read:

```
    if (flags[0] && !flags[1] && !flags[2] &&
        (request.maximum_frequency() > 1)) {
```

Note that the code is now checking that the `maximum_frequency` is greater than 1.
___
*[`src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/liquid_legions_v2_encryption_utility.cc` line 884 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/676#-NBYB1_9DmS7ujt-bm0w:-NBZ1rVP5T443Ze9za3p:bx8jl71) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6029678bb4bb187b65f7d17439076be80a3c682c/src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/liquid_legions_v2_encryption_utility.cc#L884)):*
<details><summary><i>Previously, wangyaopw (Yao Wang) wrote…</i></summary><blockquote>

You are still adding frequency noise when the maximum_frequency is set to 1.

But we can not just simply stop adding this noise here, since when we compensate for frequency noise when estimating reach at the aggregator. If we modify here, we need to adjust the code for reach estimate too.
</blockquote></details>

If the `maximum_frequency` is 1, then `completeExecutionPhaseTwoAtNonAggregator` will not set the noise parameters in the `CompleteExecutionPhaseTwoRequest` proto message.  See line 687 of `LiquidLegionsV2Mill.kt`.  Consequently, the test at line 883 will fail and line 884 will not be executed.
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/frontend/FrontendSimulator.kt` line 223 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/676#-NAz80TT8bzjjx3qcBVD:-NBYyMbP9otVqXrsD-Tf:b-7h37jh) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/505f7ca22a969fd1da049baded1a1d2ebab01c37/src/main/kotlin/org/wfanet/measurement/loadtest/frontend/FrontendSimulator.kt#L223)):*

One alternative is to use `reachOnlyResult!!`afterwards. But I think the choice is from you.
</blockquote></details>

The !! operator is unsafe and I think its use is discouraged in Kotlin.

https://kotlinlang.org/docs/null-safety.html
___
*[`src/main/proto/wfa/measurement/internal/duchy/protocol/liquid_legions_v2_encryption_methods.proto` line 115 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/676#-NBZCh9J44k7wv5Z0-q9:-NBZJeZNBlw8CaEGfv5i:b-5dbg9a) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6029678bb4bb187b65f7d17439076be80a3c682c/src/main/proto/wfa/measurement/internal/duchy/protocol/liquid_legions_v2_encryption_methods.proto#L115)):*

Using in band signaling, in this case measurement frequency, to indicate fundamentally different behavior is generally undesirable. It's better to be explicit. A field called "compute_reach_only" would make more sense. That said, setting a maximum frequency has its own usefulness.

https://www.atlasobscura.com/articles/capn-crunch-whistle :)
</blockquote></details>

I love the Cap'n Crunch example (do they still make that?), but I think I would respectfully beg to disagree on this point.  In my view, the contract is defined by the MeasurementSpec proto.  I am not changing the contract or adding hidden functionality.  I am optimizing the implementation of one particular execution path.  
___
*[`src/test/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/liquid_legions_v2_encryption_utility_test.cc` line 689 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/676#-NAlkU8LX3KHjyLT5CwL-r1-689:-NBYxtW58bCPGQkw8kGH:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/505f7ca22a969fd1da049baded1a1d2ebab01c37/src/test/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/liquid_legions_v2_encryption_utility_test.cc#L689)):*
<details><summary><i>Previously, pasin30055 (Pasin Manurangsi) wrote…</i></summary><blockquote>

The current tests LGTM. Thanks.
</blockquote></details>

Done.

matthewclegg (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/676)* status: 2 of 8 files reviewed, 8 unresolved discussions (waiting on @benjaminkreuter, @duliomatos1, @kungfucraig, @renjiezh, and @wangyaopw)

___
*[`src/main/proto/wfa/measurement/internal/duchy/protocol/liquid_legions_v2_encryption_methods.proto` line 115 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/676#-NBZCh9J44k7wv5Z0-q9:-NBZMf7fBvb7sGfwT40k:bjf79wa) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6029678bb4bb187b65f7d17439076be80a3c682c/src/main/proto/wfa/measurement/internal/duchy/protocol/liquid_legions_v2_encryption_methods.proto#L115)):*

I love the Cap'n Crunch example (do they still make that?), but I think I would respectfully beg to disagree on this point.  In my view, the contract is defined by the MeasurementSpec proto.  I am not changing the contract or adding hidden functionality.  I am optimizing the implementation of one particular execution path.  
</blockquote></details>

Maybe there is a larger point here, though.  The reason for this PR is that I discovered that an MC could submit a measurement request that causes the Duchies to perform arbitrarily long computations.  This could be the basis of a denial of service attack, and perhaps there should be some hardening built into the system to safeguard against that.

In a separate PR, of course!

wangyaopw (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/676)* status: 2 of 8 files reviewed, 7 unresolved discussions (waiting on @benjaminkreuter, @duliomatos1, @kungfucraig, @matthewclegg, and @renjiezh)

___
*[`src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/liquid_legions_v2_encryption_utility.cc` line 564 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/676#-NBY8Eai8cFDYzlwRjup:-NBZSPCi7jzXVIYq0anM:br02ali) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6029678bb4bb187b65f7d17439076be80a3c682c/src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/liquid_legions_v2_encryption_utility.cc#L564)):*

To me it seemed like randomizing the counts added an extra level of safety to the code.  After this function returns, one can be mathematically certain that there is no danger of information about the counts being leaked by subsequent operations, even if there are bugs in the code or someone in the future accidentally changes the code.  Nonetheless, you are the second person to make this comment.  If you feel strongly about this, let me know and I will remove the code that randomizes the counts.
</blockquote></details>

It doesn't matter if we are going to replace this with a real reach-only protocol in the future.

up to you to keep this or not
___
*[`src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/liquid_legions_v2_encryption_utility.cc` line 573 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/676#-NBY0IoL83POKfU93LIg:-NBZRRTQBLoBzJj0KsLl:b-bi2ja5) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6029678bb4bb187b65f7d17439076be80a3c682c/src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/liquid_legions_v2_encryption_utility.cc#L573)):*

I made this change, but I am just wondering, could you explain how this improves the code?
</blockquote></details>

No big difference. 

If `source` is empty, there is no need to call `GetNumberOfBlocks()` to know the result

:smile: 



_Code quote:_
```cpp
GetNumberOfBlocks
```
___
*[`src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/liquid_legions_v2_encryption_utility.cc` line 831 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/676#-NBYBTby9vZUr6kebaoL:-NBZRsG-5D6ekJ4ydDmo:bsv1uu9) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6029678bb4bb187b65f7d17439076be80a3c682c/src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/liquid_legions_v2_encryption_utility.cc#L831)):*

If the frequency is 1, then `completeExecutePhaseOneAtAggregator` in `LiquidLegionsV2Mill.kt` will not set the `noise_parameters` field.  Hence, the test at line 830 will be false and line 831 will not be executed.  See line 501 of `LiquidLegionsV2Mill.kt`.

The reach estimate is adjusted at line 945 in this file.  Lines 944-945 now read:

```
    if (flags[0] && !flags[1] && !flags[2] &&
        (request.maximum_frequency() > 1)) {
```

Note that the code is now checking that the `maximum_frequency` is greater than 1.


</blockquote></details>

I see. That works. Nice

wangyaopw (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/676)* status: 2 of 8 files reviewed, 4 unresolved discussions (waiting on @benjaminkreuter, @duliomatos1, @kungfucraig, and @renjiezh)

___
*[`src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/liquid_legions_v2_encryption_utility.cc` line 564 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/676#-NBY8Eai8cFDYzlwRjup:-NBZXRQ92ijNz1IxY5HA:bumpu9o) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6029678bb4bb187b65f7d17439076be80a3c682c/src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/liquid_legions_v2_encryption_utility.cc#L564)):*
<details><summary><i>Previously, wangyaopw (Yao Wang) wrote…</i></summary><blockquote>

It doesn't matter if we are going to replace this with a real reach-only protocol in the future.

up to you to keep this or not
</blockquote></details>

An even better solution is to update the sketch encryptor used by the publishers to use random numbers for key and count if the measurement is REACH_ONLY.  then we can preclude the issue from the source.

duliomatos1 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/676)* status: 2 of 8 files reviewed, 4 unresolved discussions (waiting on @benjaminkreuter, @duliomatos1, @kungfucraig, @matthewclegg, @pasin30055, @renjiezh, and @wangyaopw)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/frontend/FrontendSimulator.kt` line 223 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/676#-NAz80TT8bzjjx3qcBVD:-NC5mM0F8pBL9vSpwIBS:bw8pula) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/505f7ca22a969fd1da049baded1a1d2ebab01c37/src/main/kotlin/org/wfanet/measurement/loadtest/frontend/FrontendSimulator.kt#L223)):*

The !! operator is unsafe and I think its use is discouraged in Kotlin.

https://kotlinlang.org/docs/null-safety.html
</blockquote></details>

we' ve been using `checkNotNull()` for these checks (It throws `IllegalStateException`), and you can pass a block with a message e.g.

RequisitionReader.kt:
```
163:          checkNotNull(DuchyIds.getExternalId(duchyId)) {
164-            "Duchy with internal ID $duchyId not found"
165-          }
```


`requireNotNull()` is similar, semantically would be used for confirming function inputs contracts (throws IllegalArgumentException)

benjaminkreuter (approved):
> Reviewed 4 of 8 files at r1, 1 of 3 files at r2, 2 of 2 files at r3, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/676)* status: all files reviewed, 4 unresolved discussions (waiting on @benjaminkreuter, @kungfucraig, @matthewclegg, and @renjiezh)

duliomatos1 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/676)* status: all files reviewed, 4 unresolved discussions (waiting on @benjaminkreuter, @kungfucraig, @matthewclegg, and @renjiezh)

matthewclegg (commented):
> Reviewed 1 of 8 files at r1, 1 of 3 files at r2, 1 of 2 files at r3, 1 of 1 files at r5, 2 of 2 files at r6, all commit messages.
> Dismissed @kungfucraig from 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/676)* status: all files reviewed, 2 unresolved discussions (waiting on @benjaminkreuter and @duliomatos1)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/frontend/FrontendSimulator.kt` line 223 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/676#-NAz80TT8bzjjx3qcBVD:-NC6qhmO1easAuPqS4rf:b-h5rv17) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/505f7ca22a969fd1da049baded1a1d2ebab01c37/src/main/kotlin/org/wfanet/measurement/loadtest/frontend/FrontendSimulator.kt#L223)):*
<details><summary><i>Previously, duliomatos1 (Dulio Matos) wrote…</i></summary><blockquote>

we' ve been using `checkNotNull()` for these checks (It throws `IllegalStateException`), and you can pass a block with a message e.g.

RequisitionReader.kt:
```
163:          checkNotNull(DuchyIds.getExternalId(duchyId)) {
164-            "Duchy with internal ID $duchyId not found"
165-          }
```


`requireNotNull()` is similar, semantically would be used for confirming function inputs contracts (throws IllegalArgumentException)
</blockquote></details>

Thanks for the suggestion!

matthewclegg (commented):
> Reviewed 2 of 8 files at r1, 1 of 2 files at r3.
> Dismissed @benjaminkreuter from a discussion.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/676)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @matthewclegg)

PR #678, opened 2022-08-31
Attempts to solve unordered task claimed by Mill

renjiezh:
Order unclaimedTasks by UpdatedTime then by LockExpirationTime

- - -
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/678)

### DIFF
--- a/src/main/kotlin/org/wfanet/measurement/duchy/deploy/gcloud/spanner/computation/UnclaimedTasksQuery.kt
+++ b/src/main/kotlin/org/wfanet/measurement/duchy/deploy/gcloud/spanner/computation/UnclaimedTasksQuery.kt
@@ -41,7 +41,7 @@ class UnclaimedTasksQuery<StageT>(
         AND c.LockExpirationTime IS NOT NULL
         AND c.UpdateTime IS NOT NULL
         AND c.LockExpirationTime <= @current_time
-      ORDER BY c.LockExpirationTime ASC, c.UpdateTime ASC
+      ORDER BY c.UpdateTime ASC, c.LockExpirationTime ASC
       LIMIT 50
       """
   }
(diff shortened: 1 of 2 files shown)


matthewclegg (changes requested):
Hi Renjie, This seems like a significant improvement.  The above graph shows the behavior of the Duchy before and after your patch.  For each curve, I submitted 50 identical reach-only requests to the Kingdom.  These requests were submitted concurrently, e.g., I did not wait for a request to complete before transmitting the next request.  I then measured the elapsed time from when the request was first transmitted until when the result was received.  The graph plots the elapsed time to receive a response as a function of the the request number.  The average amount of time needed to  process a request was 114 seconds for the first curve, 110 seconds for the second.  The elapsed time from when the first request was transmitted until the result was received was 1991 seconds for the first curve, 1367 seconds for the second.  

Having said that, I also tried sending a single request to the Kingdom (without other concurrent requests).  This was handled in 282 seconds.  Then, I tried another experiment where I sent 5 requests concurrently.  In this experiment, the first request was handled in 436 seconds and the last was handled after 591 seconds.  So, I don't think this PR has fully solved the starvation issue.

> Reviewed 1 of 1 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/678#-:-NBcJqKA6RK62psxfEo_:b-1mgs63)* status: all files reviewed, 1 unresolved discussion (waiting on @renjiezh)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/gcloud/spanner/computation/UnclaimedTasksQuery.kt` line 45 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/678#-NBcYPE-CTBdAjXU2ElD:-NBcYPE-CTBdAjXU2ElE:bqbotkd) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e9d59c892d9e58b0b0aa0c317a4fed00fe7f4013/src/main/kotlin/org/wfanet/measurement/duchy/deploy/gcloud/spanner/computation/UnclaimedTasksQuery.kt#L45)):*
> ```Kotlin
>         AND c.LockExpirationTime <= @current_time
>       ORDER BY c.UpdateTime ASC, c.LockExpirationTime ASC
>       LIMIT 50
> ```

Why the limit of 50?  To me, it seems like this should be 1.

PR #684, opened 2022-09-08
Catch StatusException instead of StatusRuntimeException

renjiezh:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/684)

### DIFF
--- a/src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/AccountsService.kt
+++ b/src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/AccountsService.kt
@@ -16,7 +16,7 @@ package org.wfanet.measurement.kingdom.service.api.v2alpha
 
 import com.google.gson.JsonParser
 import io.grpc.Status
-import io.grpc.StatusRuntimeException
+import io.grpc.StatusException
 import java.io.IOException
 import java.security.GeneralSecurityException
 import org.wfanet.measurement.api.AccountConstants
@@ -88,10 +88,10 @@ class AccountsService(
     val result =
       try {
         internalAccountsStub.createAccount(internalCreateAccountRequest)
-      } catch (ex: StatusRuntimeException) {
-        when (ex.status) {
-          Status.NOT_FOUND -> failGrpc(Status.NOT_FOUND, ex) { "Creator's Account not found." }
-          Status.PERMISSION_DENIED ->
+      } catch (ex: StatusException) {
+        when (ex.status.code) {
+          Status.Code.NOT_FOUND -> failGrpc(Status.NOT_FOUND, ex) { "Creator's Account not found." }
+          Status.Code.PERMISSION_DENIED ->
             failGrpc(Status.PERMISSION_DENIED, ex) {
               "Caller does not own the owned measurement consumer."
             }
@@ -135,15 +135,15 @@ class AccountsService(
     val result =
       try {
         internalAccountsStub.activateAccount(internalActivateAccountRequest)
-      } catch (ex: StatusRuntimeException) {
-        when (ex.status) {
-          Status.PERMISSION_DENIED ->
+      } catch (ex: StatusException) {
+        when (ex.status.code) {
+          Status.Code.PERMISSION_DENIED ->
             failGrpc(Status.PERMISSION_DENIED, ex) {
               "Activation token is not valid for this account."
             }
-          Status.FAILED_PRECONDITION ->
+          Status.Code.FAILED_PRECONDITION ->
             failGrpc(Status.FAILED_PRECONDITION, ex) { ex.message ?: "Failed precondition." }
-          Status.NOT_FOUND ->
+          Status.Code.NOT_FOUND ->
             failGrpc(Status.NOT_FOUND, ex) { "Account to activate has not been found." }
           else -> failGrpc(Status.UNKNOWN, ex) { "Unknown exception." }
         }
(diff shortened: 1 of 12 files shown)


SanjayVas (approved):
> Reviewed 12 of 12 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/684)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @renjiezh)

SanjayVas (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/684)* status: all files reviewed, 1 unresolved discussion (waiting on @renjiezh and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/AccountsService.kt` line 93 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/684#-NBZkKIvCEAEL0lkKpXF:-NBZkKIw0tMYUCgbO_ng:b-h6igwe) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/899fbdf52bc9fbf270ff545c94acc7b00f2fe69c/src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/AccountsService.kt#L93)):*
> ```Kotlin
>       } catch (ex: StatusException) {
>         when (ex.status.code) {
>           Status.Code.NOT_FOUND -> failGrpc(Status.NOT_FOUND, ex) { "Creator's Account not found." }
> ```

nit: I prefer the following pattern instead of using `failGrpc`, as it doesn't require specifying a message supplier. It also turns out that using a lambda for exception messages doesn't add anything. That pattern was (incorrectly) copied from logging, where it can potentially save on building the message if the log level isn't hit.

```kotlin
        throw when (ex.status.code) {
          Status.Code.NOT_FOUND -> Status.NOT_FOUND.withDescription("Creator's Account not found")
          Status.Code.PERMISSION_DENIED -> Status.PERMISSION_DENIED.withDescription(
            "Caller does not own the owned measurement consumer")
          else -> Status.UNKNOWN
        }.withCause(e).asRuntimeException()
```



_Code quote:_
```Kotlin
failGrpc
```

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/684)* status: all files reviewed, 1 unresolved discussion (waiting on @SanjayVas and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/AccountsService.kt` line 93 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/684#-NBZkKIvCEAEL0lkKpXF:-NBmytvA-28XQNaO-tLe:bd7kvi7) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/899fbdf52bc9fbf270ff545c94acc7b00f2fe69c/src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/AccountsService.kt#L93)):*

nit: I prefer the following pattern instead of using `failGrpc`, as it doesn't require specifying a message supplier. It also turns out that using a lambda for exception messages doesn't add anything. That pattern was (incorrectly) copied from logging, where it can potentially save on building the message if the log level isn't hit.

```kotlin
        throw when (ex.status.code) {
          Status.Code.NOT_FOUND -> Status.NOT_FOUND.withDescription("Creator's Account not found")
          Status.Code.PERMISSION_DENIED -> Status.PERMISSION_DENIED.withDescription(
            "Caller does not own the owned measurement consumer")
          else -> Status.UNKNOWN
        }.withCause(e).asRuntimeException()
```

</blockquote></details>

If we don't like the way of using lambda, what about change the interface as `failGrpc(Status.NOT_FOUND, ex, "Some message")`?
It is more code to write as `throw Status.Code.NOT_FOUND..withCause(e).withDescription("message").asRuntimeException()`

SanjayVas (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/684)* status: all files reviewed, 1 unresolved discussion (waiting on @renjiezh and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/AccountsService.kt` line 93 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/684#-NBZkKIvCEAEL0lkKpXF:-NBnGFIo0sbk4DPTi82T:bva1ry8) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/899fbdf52bc9fbf270ff545c94acc7b00f2fe69c/src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/AccountsService.kt#L93)):*

If we don't like the way of using lambda, what about change the interface as `failGrpc(Status.NOT_FOUND, ex, "Some message")`?
It is more code to write as `throw Status.Code.NOT_FOUND..withCause(e).withDescription("message").asRuntimeException()`
</blockquote></details>

I don't see the `throw` and `withCause(e).asRuntimeException()` as much of a big deal in terms of excess repeated code, as those are only repeated once per catch block.

Changing the failGrpc method is also fine. To avoid lots of churn, I'd recommend just marking the existing message supplier overloads as deprecated, and adding a new one for an optional string message. Note that this requires a common-jvm change.

I put this as a nit here since the PR you have written will still work, so feel free to do any cleanups in a separate PR.

stevenwarejones (approved):
> Reviewed 12 of 12 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/684)* status: all files reviewed, 1 unresolved discussion (waiting on @renjiezh)

Issue #695, opened 2022-09-21
Configure gRPC retry policies using service config


SanjayVas:
Rather than having custom retry logic, we should use what is built-in to gRPC itself. See https://github.com/grpc/grpc/blob/master/doc/service_config.md

PR #674, opened 2022-08-24
Herald catches errors and reports to Kingdom

renjiezh:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/674)

- Herald catches exceptions without rethrowing outside.
- Report to Kingdom for non-transient errors.
- Propagate herald error message to MC.
- Add unit test and integration test.

PR #664, opened 2022-08-12
Convert uncaught Exceptions to StatusException

renjiezh:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/664)

Issue #585, opened 2022-06-03
Herald retries computation start/update unconditionally.


SanjayVas:
Retries for start/update were added in f02986eb64fdd45b3427442637559a0ee657d02d to deal with a potential race condition. This was prior to the codebase being open-sourced, so the following is copied from the Google-internal issue b/168551392:

> The Herald has a race condition:
> 
> 1. the mill in the last unconfirmed duchy checks the local requisition and finds all are available
> 2. the mill confirms to the kingdom
> 3. the mill updates local computation to WAIT_TO_START
> 4. the kingdom updates the computation to RUNNING and sends out the status change to all connected heralds
> 5. the herald tries to update the local computation to TO_ADD_NOISE

Unfortunately, this retry is unconditional on any exception. `syncStatuses` also wraps the higher-level `processSystemComputationChange` in retries, but those check the gRPC status to determine if the operation is retriable. This would likely be insufficient for the start/update case as it's a state mismatch.

PR #697, opened 2022-09-22
Improve handling of long-lived streaming RPCs.

SanjayVas:
This prevents said RPCs from overloading the Kingdom.

- - -
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/697)

### DIFF
--- a/src/main/k8s/dev/kingdom_gke.cue
+++ b/src/main/k8s/dev/kingdom_gke.cue
@@ -19,7 +19,17 @@ _secret_name: string @tag("secret_name")
 // Name of K8s service account for the internal API server.
 #InternalServerServiceAccount: "internal-server"
 
-#DataServerResourceRequirements: #ResourceRequirements & {
+// Number of gRPC threads for the internal API server.
+#InternalServerGrpcThreads: 7
+
+// Number of gRPC threads for the system API server.
+//
+// This serves long-lived streaming RPCs from each Herald which will each occupy
+// a thread, so this should be greater than the number of Heralds.
+#SystemServerGrpcThreads: 5
+
+#InternalServerResourceRequirements: #ResourceRequirements & {
+	requests: cpu:  "500m"
 	limits: memory: "512Mi"
 }
 
@@ -61,10 +71,16 @@ kingdom: #Kingdom & {
 		"gcp-kingdom-data-server": {
 			_container: {
 				_javaOptions: maxRamPercentage: 40.0
-				resources: #DataServerResourceRequirements
+				_grpcThreadPoolSize: #InternalServerGrpcThreads
+				resources:           #InternalServerResourceRequirements
 			}
 			spec: template: spec: #ServiceAccountPodSpec & {
 				serviceAccountName: #InternalServerServiceAccount
+			}
+		}
+		"system-api-server": {
+			_container: {
+				_grpcThreadPoolSize: #SystemServerGrpcThreads
 			}
 		}
 	}
--- a/src/main/k8s/duchy.cue
+++ b/src/main/k8s/duchy.cue
@@ -40,6 +40,7 @@ import ("strings")
 
 	_images: [Name=_]: string
 	_duchy_image_pull_policy: string
+	_millPollingInterval?:    string
 
 	_akid_to_principal_map_file_flag:                   "--authority-key-identifier-to-principal-map-file=/etc/\(#AppName)/config-files/authority_key_identifier_to_principal_map.textproto"
 	_async_computations_control_service_target_flag:    "--async-computation-control-service-target=" + (#Target & {name: "\(_name)-async-computation-control-server"}).target
@@ -95,8 +96,7 @@ import ("strings")
 				_duchy_protocols_setup_config_flag,
 				_kingdom_system_api_target_flag,
 				_kingdom_system_api_cert_host_flag,
-				"--channel-shutdown-timeout=3s",
-				"--polling-interval=1m",
+				_debug_verbose_grpc_client_logging_flag,
 			]
 		}
 		"liquid-legions-v2-mill-daemon-deployment": {
@@ -113,8 +113,7 @@ import ("strings")
 						_duchy_cs_cert_rename_name_flag,
 						_kingdom_system_api_target_flag,
 						_kingdom_system_api_cert_host_flag,
-						"--channel-shutdown-timeout=3s",
-						"--polling-interval=1s",
+						if (_millPollingInterval != _|_) {"--polling-interval=\(_millPollingInterval)"},
 			] + _blob_storage_flags + _computation_control_target_flags
 			spec: template: spec: _dependencies: [
 				"\(_name)-spanner-computations-server", "\(_name)-computation-control-server",
--- a/src/main/k8s/kingdom.cue
+++ b/src/main/k8s/kingdom.cue
@@ -58,63 +58,74 @@ package k8s
 		_secretName: _kingdom_secret_name
 		_system:     "kingdom"
 		_container: {
+			_grpcThreadPoolSize?: int32 & >0
+			_commonServerFlags: [
+				"--port=8443",
+				"--health-port=8080",
+				if _grpcThreadPoolSize != _|_ {
+					"--grpc-thread-pool-size=\(_grpcThreadPoolSize)"
+				},
+			]
+
 			image:           _images[_name]
 			imagePullPolicy: _kingdom_image_pull_policy
 		}
 	}
 	deployments: {
 		"gcp-kingdom-data-server": {
-			_container: args: [
-						_duchy_info_config_flag,
-						_duchy_id_config_flag,
-						_kingdom_tls_cert_file_flag,
-						_kingdom_tls_key_file_flag,
-						_kingdom_cert_collection_file_flag,
-						_debug_verbose_grpc_server_logging_flag,
-						"--port=8443",
-						"--health-port=8080",
-			] + _spannerConfig.flags
+			_container: Container={
+				args: [
+					_duchy_info_config_flag,
+					_duchy_id_config_flag,
+					_kingdom_tls_cert_file_flag,
+					_kingdom_tls_key_file_flag,
+					_kingdom_cert_collection_file_flag,
+					_debug_verbose_grpc_server_logging_flag,
+				] + Container._commonServerFlags + _spannerConfig.flags
+			}
 			_updateSchemaContainer: Container=#Container & {
 				image:           _images[Container.name]
 				imagePullPolicy: _container.imagePullPolicy
 				args:            _spannerConfig.flags
 			}
-			spec: template: spec: _initContainers: {
-				"update-kingdom-schema": _updateSchemaContainer
+			spec: template: spec: {
+				_initContainers: {
+					"update-kingdom-schema": _updateSchemaContainer
+				}
 			}
 		}
 
 		"system-api-server": {
-			_container: args: [
-				_debug_verbose_grpc_client_logging_flag,
-				_debug_verbose_grpc_server_logging_flag,
-				_duchy_info_config_flag,
-				_kingdom_tls_cert_file_flag,
-				_kingdom_tls_key_file_flag,
-				_kingdom_cert_collection_file_flag,
-				_internal_api_target_flag,
-				_internal_api_cert_host_flag,
-				"--port=8443",
-				"--health-port=8080",
-			]
+			_container: Container={
+				args: [
+					_debug_verbose_grpc_client_logging_flag,
+					_debug_verbose_grpc_server_logging_flag,
+					_duchy_info_config_flag,
+					_kingdom_tls_cert_file_flag,
+					_kingdom_tls_key_file_flag,
+					_kingdom_cert_collection_file_flag,
+					_internal_api_target_flag,
+					_internal_api_cert_host_flag,
+				] + Container._commonServerFlags
+			}
 			spec: template: spec: _dependencies: ["gcp-kingdom-data-server"]
 		}
 
 		"v2alpha-public-api-server": {
-			_container: args: [
-				_debug_verbose_grpc_client_logging_flag,
-				_debug_verbose_grpc_server_logging_flag,
-				_llv2_protocol_config_config,
-				_kingdom_tls_cert_file_flag,
-				_kingdom_tls_key_file_flag,
-				_kingdom_cert_collection_file_flag,
-				_internal_api_target_flag,
-				_internal_api_cert_host_flag,
-				_akid_to_principal_map_file_flag,
-				_open_id_redirect_uri_flag,
-				"--port=8443",
-				"--health-port=8080",
-			]
+			_container: Container={
+				args: [
+					_debug_verbose_grpc_client_logging_flag,
+					_debug_verbose_grpc_server_logging_flag,
+					_llv2_protocol_config_config,
+					_kingdom_tls_cert_file_flag,
+					_kingdom_tls_key_file_flag,
+					_kingdom_cert_collection_file_flag,
+					_internal_api_target_flag,
+					_internal_api_cert_host_flag,
+					_akid_to_principal_map_file_flag,
+					_open_id_redirect_uri_flag,
+				] + Container._commonServerFlags
+			}
 			spec: template: spec: {
 				_mounts: "config-files": #ConfigMapMount
 				_dependencies: ["gcp-kingdom-data-server"]
--- a/src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/Herald.kt
+++ b/src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/Herald.kt
@@ -15,22 +15,23 @@
 package org.wfanet.measurement.duchy.daemon.herald
 
 import io.grpc.Status
+import io.grpc.StatusException
 import java.time.Clock
 import java.time.Duration
 import java.util.logging.Level
 import java.util.logging.Logger
+import kotlin.coroutines.coroutineContext
 import kotlin.math.pow
 import kotlin.random.Random
+import kotlinx.coroutines.CancellationException
 import kotlinx.coroutines.coroutineScope
 import kotlinx.coroutines.delay
-import kotlinx.coroutines.flow.collect
-import kotlinx.coroutines.flow.onEach
+import kotlinx.coroutines.flow.catch
+import kotlinx.coroutines.isActive
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.sync.Semaphore
 import org.wfanet.measurement.common.grpc.grpcStatusCode
-import org.wfanet.measurement.common.logAndSuppressExceptionSuspend
 import org.wfanet.measurement.common.protoTimestamp
-import org.wfanet.measurement.common.throttler.Throttler
 import org.wfanet.measurement.duchy.daemon.utils.MeasurementType
 import org.wfanet.measurement.duchy.daemon.utils.key
 import org.wfanet.measurement.duchy.daemon.utils.toMeasurementType
@@ -76,20 +77,17 @@ class Herald(
   private val clock: Clock,
   private val blobStorageBucket: String = "computation-blob-storage",
   private val maxAttempts: Int = 5,
-  private val maxConcurrency: Int = 5,
+  private val maxStreamingAttempts: Int = 5,
+  maxConcurrency: Int = 5,
   private val retryBackoff: ExponentialBackoff = ExponentialBackoff(),
 ) {
   private val semaphore = Semaphore(maxConcurrency)
 
   /**
    * Syncs the status of computations stored at the kingdom with those stored locally continually in
-   * a forever loop. The [pollingThrottler] is used to limit how often the kingdom and local
-   * computation storage service are polled.
-   *
-   * @param pollingThrottler throttles how often to get active computations from the Global
-   * Computation Service
+   * a forever loop.
    */
-  suspend fun continuallySyncStatuses(pollingThrottler: Throttler) {
+  suspend fun continuallySyncStatuses() {
     logger.info("Server starting...")
     // Token signifying the last computation in an active state at the kingdom that was processed by
     // this job. When empty, all active computations at the kingdom will be streamed in the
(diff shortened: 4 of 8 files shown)


renjiezh (changes requested):
> Reviewed 4 of 8 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/697)* status: 4 of 8 files reviewed, 3 unresolved discussions (waiting on @matthewclegg and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/Herald.kt` line 97 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/697#-NCafnmF3di8Jb10SxlE:-NCafnmF3di8Jb10SxlF:by0ioji) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/dade6930a309f0e29838361a86bf15cd0dd68ceb/src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/Herald.kt#L97)):*
> ```Kotlin
>     // TODO(world-federation-of-advertisers/cross-media-measurement#695): Use gRPC service config
>     // rather than custom retry logic w/backoff.
>     var attemptNumber = 1
> ```

nit: Do we need a attempt limit to crash the process so it is easier to know streaming goes wrong?
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/Herald.kt` line 101 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/697#-NCafVgI9BIOloROrWVE:-NCafVgI9BIOloROrWVF:byf7o0) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/dade6930a309f0e29838361a86bf15cd0dd68ceb/src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/Herald.kt#L101)):*
> ```Kotlin
>       try {
>         lastProcessedContinuationToken = syncStatuses(lastProcessedContinuationToken)
>       } catch (e: StreamingException) {
> ```

nit: Can we explicitly have logs that it is hit by a StreamingException.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/Herald.kt` line 104 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/697#-NCaidunANX6vUJde5Fk:-NCaidunANX6vUJde5Fl:b-rnmkm3) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/dade6930a309f0e29838361a86bf15cd0dd68ceb/src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/Herald.kt#L104)):*
> ```Kotlin
>         when (e.cause.status.code) {
>           Status.Code.UNAVAILABLE,
>           Status.Code.DEADLINE_EXCEEDED -> {
> ```

I wonder if backoff delay needed for DEADLINE_EXCEEDED.
If I understand correctly, when everything goes well, herald will receive DEADLINE_EXCEEDED every 10 min (duration depends on the server setting). This is not an error. It should not increase the attempts number which eventually leads herald sleep for too long.

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/697)* status: 4 of 8 files reviewed, 3 unresolved discussions (waiting on @matthewclegg, @renjiezh, and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/Herald.kt` line 97 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/697#-NCafnmF3di8Jb10SxlE:-NCb0nXcEcd-9pqpKnTX:b-1cxfn) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/dade6930a309f0e29838361a86bf15cd0dd68ceb/src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/Herald.kt#L97)):*

nit: Do we need a attempt limit to crash the process so it is easier to know streaming goes wrong?
</blockquote></details>

I thought about this, but crashing will just result in the Herald restarting with it's own CrashLoopBackoff. I'm not entirely sure if that's better. I suppose it allows one to configure a failure limit at the K8s level and have the job just stop running entirely.

If we throw at this limit, then we should also just throw when we hit unexpected error codes.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/Herald.kt` line 104 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/697#-NCaidunANX6vUJde5Fk:-NCb1-JrBy7w4bJ1nqGt:bsmpwvj) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/dade6930a309f0e29838361a86bf15cd0dd68ceb/src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/Herald.kt#L104)):*

I wonder if backoff delay needed for DEADLINE_EXCEEDED.
If I understand correctly, when everything goes well, herald will receive DEADLINE_EXCEEDED every 10 min (duration depends on the server setting). This is not an error. It should not increase the attempts number which eventually leads herald sleep for too long.
</blockquote></details>

It won't receive a `DEADLINE_EXCEEDED` under normal operation. Instead, the server just completes the flow at 10 minutes.

It would hit `DEADLINE_EXCEEDED` if it hits an RPC deadline anywhere in the chain.  I have an upcoming PR that adds a default deadline for calls to the Kingdom internal server from the API servers.

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/697)* status: 4 of 8 files reviewed, 2 unresolved discussions (waiting on @matthewclegg and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/Herald.kt` line 97 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/697#-NCafnmF3di8Jb10SxlE:-NCb4I_O-wpyQK822Nm2:b-wlcqeg) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/dade6930a309f0e29838361a86bf15cd0dd68ceb/src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/Herald.kt#L97)):*

I thought about this, but crashing will just result in the Herald restarting with it's own CrashLoopBackoff. I'm not entirely sure if that's better. I suppose it allows one to configure a failure limit at the K8s level and have the job just stop running entirely.

If we throw at this limit, then we should also just throw when we hit unexpected error codes.
</blockquote></details>

I was thinking the way to detect failure. If a permanent disconnection to Kingdom happens, a restart count or even CrashLoopBackoff can help us be aware, while keeping delaying could hide the failure.

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/697)* status: 3 of 8 files reviewed, all discussions resolved (waiting on @matthewclegg and @renjiezh)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/Herald.kt` line 97 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/697#-NCafnmF3di8Jb10SxlE:-NCb5wWk9K1gVPzPQbze:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/dade6930a309f0e29838361a86bf15cd0dd68ceb/src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/Herald.kt#L97)):*

I was thinking the way to detect failure. If a permanent disconnection to Kingdom happens, a restart count or even CrashLoopBackoff can help us be aware, while keeping delaying could hide the failure.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/Herald.kt` line 101 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/697#-NCafVgI9BIOloROrWVE:-NCb5xMf9iUJm5D5tu8W:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/dade6930a309f0e29838361a86bf15cd0dd68ceb/src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/Herald.kt#L101)):*

nit: Can we explicitly have logs that it is hit by a StreamingException.
</blockquote></details>

Done.

renjiezh (approved):
> Reviewed 2 of 8 files at r1, 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/697)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @matthewclegg)

matthewclegg (changes requested):
> Reviewed 1 of 8 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/697)* status: all files reviewed, 4 unresolved discussions (waiting on @SanjayVas)

___
*[`src/main/k8s/duchy.cue` line 99 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/697#-NCfQgee1G_PtL9DRxtF:-NCfQgee1G_PtL9DRxtG:bq33qhr) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/65335a00a47dcc756e24f152863275ca5e164505/src/main/k8s/duchy.cue#L99)):*
> ```CUE
> 				_kingdom_system_api_target_flag,
> 				_kingdom_system_api_cert_host_flag,
> 				_debug_verbose_grpc_client_logging_flag,
> ```

Please add Emad as a reviewer.  We will need to make corresponding changes to the yaml file of our internal Duchy deployment.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/Herald.kt` line 99 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/697#-NCfRIQi22llLpvYJ3hw:-NCfRIQjBvuh9AGCDakU:b-3mjfe3) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/65335a00a47dcc756e24f152863275ca5e164505/src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/Herald.kt#L99)):*
> ```Kotlin
>     // TODO(world-federation-of-advertisers/cross-media-measurement#695): Use gRPC service config
>     // rather than custom retry logic w/backoff.
>     var attemptNumber = 1
> ```

Could you please explain why you are removing the pollingThrottler? 
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/Herald.kt` line 100 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/697#-NCfSgjF2J_ZMkNXag00:-NCfSgjF2J_ZMkNXag01:b-3qnwhb) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/65335a00a47dcc756e24f152863275ca5e164505/src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/Herald.kt#L100)):*
> ```Kotlin
>     // rather than custom retry logic w/backoff.
>     var attemptNumber = 1
>     while (coroutineContext.isActive) {
> ```

Why would the context not be active?
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/Herald.kt` line 142 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/697#-NCfTmIy8rBkaM-VpuSw:-NCfTmIy8rBkaM-VpuSx:b-nxirqn) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/65335a00a47dcc756e24f152863275ca5e164505/src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/Herald.kt#L142)):*
> ```Kotlin
>     var lastProcessedContinuationToken = continuationToken
>     coroutineScope {
>       systemComputationsClient
> ```

I think a cleaner design would be to create a wrapper for systemComputationsClient.streamActiveComputations() that catches and handles any exceptions that might occur.  What exceptions are we expecting to have to handle, and why?

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/697)* status: all files reviewed, 4 unresolved discussions (waiting on @matthewclegg)

___
*[`src/main/k8s/duchy.cue` line 99 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/697#-NCfQgee1G_PtL9DRxtF:-NCv2idICO7Yv3TXyedf:b-4mdb4q) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/65335a00a47dcc756e24f152863275ca5e164505/src/main/k8s/duchy.cue#L99)):*

Please add Emad as a reviewer.  We will need to make corresponding changes to the yaml file of our internal Duchy deployment.
</blockquote></details>

Pinged him offline as he's not yet in the GitHub org.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/Herald.kt` line 99 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/697#-NCfRIQi22llLpvYJ3hw:-NCv2yVH7A_RE3-ZpC4S:b-cor5yr) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/65335a00a47dcc756e24f152863275ca5e164505/src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/Herald.kt#L99)):*

Could you please explain why you are removing the pollingThrottler? 
</blockquote></details>

We don't need to poll on an interval, but rather reconnect the streaming RPC whenever it completes. We generally want to do this immediately, unless there's an error in which case we delay using a backoff.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/Herald.kt` line 100 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/697#-NCfSgjF2J_ZMkNXag00:-NCv2zSV1k_ufDaj4Gcm:b642jk) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/65335a00a47dcc756e24f152863275ca5e164505/src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/Herald.kt#L100)):*

Why would the context not be active?
</blockquote></details>

Coroutine cancellation. This is the equivalent of `while (true)`, but what one should use instead when there's a coroutine context (e.g. in a suspendable function).

See https://kotlinlang.org/docs/cancellation-and-timeouts.html#making-computation-code-cancellable
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/Herald.kt` line 142 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/697#-NCfTmIy8rBkaM-VpuSw:-NCv5KZJ5-4IvAwzfhUW:b-2epmaf) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/65335a00a47dcc756e24f152863275ca5e164505/src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/Herald.kt#L142)):*

I think a cleaner design would be to create a wrapper for systemComputationsClient.streamActiveComputations() that catches and handles any exceptions that might occur.  What exceptions are we expecting to have to handle, and why?
</blockquote></details>

The only exception we should be getting under normal operation from any RPC should be `StatusException`, and it should be caught immediately rather than bubbled up in order to know which RPC resulted in that exception.

Basically every RPC call should have the equivalent of

```kotlin
try {
  client.methodCall()
catch (e: StatusException) {
  // Handle
}
```

This is the equivalent for streaming RPCs, as those return flows. #664 and #684 were intended to do this everywhere, but some cases were missed.

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/697)* status: 6 of 8 files reviewed, 4 unresolved discussions (waiting on @matthewclegg and @renjiezh)

*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/697#-ND5TZi24PYT6090cfAp:-ND5TZi24PYT6090cfAq:bdwqz5x) (no related file):*
@renjiezh you may want to take another look, as I tweaked some of the exception handling from #674 and noted you in a couple TODOs.

Basically:
1. Coroutine `CancellationException` should usually be bubbled up rather than caught to ensure coroutine cancellation works as expected. See https://kotlinlang.org/docs/cancellation-and-timeouts.html#cancellation-is-cooperative
1. In a similar vein, you generally want your "catch-all" blocks to use `Exception` rather than `Throwable`, so that you don't suppress serious JVM errors. These are usually unrecoverable and should crash the application. See https://docs.oracle.com/javase/8/docs/api/java/lang/Error.html
1. `StatusException` should be caught at the lowest level, i.e. the call site of the RPC. This is because how a given error should be handled depends on which method was being called. If it needs to be handled at a higher level, it should be wrapped in another exception type to make that explicit to calling code.

___

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/697)* status: 6 of 8 files reviewed, 4 unresolved discussions (waiting on @matthewclegg and @renjiezh)

*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/697#-ND5TZi24PYT6090cfAp:-ND5WPsNDrURQjQiHQP_:b-8hm32q) (no related file):*

@renjiezh you may want to take another look, as I tweaked some of the exception handling from #674 and noted you in a couple TODOs.

Basically:
1. Coroutine `CancellationException` should usually be bubbled up rather than caught to ensure coroutine cancellation works as expected. See https://kotlinlang.org/docs/cancellation-and-timeouts.html#cancellation-is-cooperative
1. In a similar vein, you generally want your "catch-all" blocks to use `Exception` rather than `Throwable`, so that you don't suppress serious JVM errors. These are usually unrecoverable and should crash the application. See https://docs.oracle.com/javase/8/docs/api/java/lang/Error.html
1. `StatusException` should be caught at the lowest level, i.e. the call site of the RPC. This is because how a given error should be handled depends on which method was being called. If it needs to be handled at a higher level, it should be wrapped in another exception type to make that explicit to calling code.
</blockquote></details>

4.  `UNKNOWN` is not a status code that are considered safe to retry. I didn't change this, as I don't know if it's part of the workaround for the race condition mentioned in #585.

    `RESOURCE_EXHAUSTED` should usually only be retried when you know that the resource is no longer exhausted (ideally there's a way to check this). Often it's better just to avoid retrying on this code and just returning an error.

___

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/697)* status: 6 of 8 files reviewed, 4 unresolved discussions (waiting on @matthewclegg and @renjiezh)

*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/697#-ND5TZi24PYT6090cfAp:-ND5Xvlx9vAuXELC682z:bym4lrc) (no related file):*

4.  `UNKNOWN` is not a status code that are considered safe to retry. I didn't change this, as I don't know if it's part of the workaround for the race condition mentioned in #585.

    `RESOURCE_EXHAUSTED` should usually only be retried when you know that the resource is no longer exhausted (ideally there's a way to check this). Often it's better just to avoid retrying on this code and just returning an error.
</blockquote></details>

Thank you. I saw these TODOs you noted. They will be fixed while resolving the [issue 585](https://github.com/world-federation-of-advertisers/cross-media-measurement/issues/585). It is a P1 task on the plan.

___

matthewclegg (approved):
> Reviewed 1 of 8 files at r1, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/697)* status: 7 of 8 files reviewed, 1 unresolved discussion (waiting on @renjiezh and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/Herald.kt` line 142 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/697#-NCfTmIy8rBkaM-VpuSw:-ND5c2Eu2J4xZXOypeXV:b-yymlow) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/65335a00a47dcc756e24f152863275ca5e164505/src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/Herald.kt#L142)):*

The only exception we should be getting under normal operation from any RPC should be `StatusException`, and it should be caught immediately rather than bubbled up in order to know which RPC resulted in that exception.

Basically every RPC call should have the equivalent of

```kotlin
try {
  client.methodCall()
catch (e: StatusException) {
  // Handle
}
```

This is the equivalent for streaming RPCs, as those return flows. #664 and #684 were intended to do this everywhere, but some cases were missed.
</blockquote></details>

Thank you for the really helpful explanations!

stevenwarejones (commented):
> Reviewed 4 of 8 files at r1, 1 of 3 files at r2, 1 of 1 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/697)* status: all files reviewed, 2 unresolved discussions (waiting on @SanjayVas)

___
*[`src/main/k8s/duchy.cue` line 116 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/697#-ND9kuP89z4H7n8aaDb-:-ND9kuP89z4H7n8aaDb0:b-7q2yvv) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7b73738a065b51269656c2d7d6f836e2f9c8149e/src/main/k8s/duchy.cue#L116)):*
> ```CUE
> 						_kingdom_system_api_target_flag,
> 						_kingdom_system_api_cert_host_flag,
> 						"--channel-shutdown-timeout=3s",
> ```

why remove this shutdown timeout?

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/697)* status: all files reviewed, 2 unresolved discussions (waiting on @SanjayVas and @stevenwarejones)

___
*[`src/main/k8s/duchy.cue` line 116 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/697#-ND9kuP89z4H7n8aaDb-:-ND9qCO22qcPJNCmg9N_:bjppm69) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7b73738a065b51269656c2d7d6f836e2f9c8149e/src/main/k8s/duchy.cue#L116)):*

why remove this shutdown timeout?
</blockquote></details>

3s is already the default, so no need to specify it here.

stevenwarejones (approved):
> Reviewed 1 of 8 files at r1.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/697)* status: all files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

SanjayVas (commented):
> Dismissed @matthewclegg from a discussion.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/697)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/Herald.kt` line 142 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/697#-NCfTmIy8rBkaM-VpuSw:-ND_DGvVDg7QadsCuieh:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/65335a00a47dcc756e24f152863275ca5e164505/src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/Herald.kt#L142)):*

Thank you for the really helpful explanations!
</blockquote></details>

Done.

PR #718, opened 2022-10-03
Add check if measurement exists in reporting queries.

tristanvuong2021:


SanjayVas (changes requested):
> Reviewed 2 of 3 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/718)* status: 2 of 3 files reviewed, 4 unresolved discussions (waiting on @SanjayVas and @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/postgres/writers/SetMeasurementFailure.kt` line 34 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/718#-NDV8HPU8FbT-iZgh8T4:-NDV8HPU8FbT-iZgh8T5:b-xhtqak) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a3f724784901a50d70e792fa514dfb81b9869ed8/src/main/kotlin/org/wfanet/measurement/reporting/deploy/postgres/writers/SetMeasurementFailure.kt#L34)):*
> ```Kotlin
>   PostgresWriter<Measurement>() {
>   override suspend fun TransactionScope.runTransaction(): Measurement {
>     MeasurementReader()
> ```

nit: It took me a bit to realize that you're relying on the fact that this method throws a `MeasurementNotFoundException`.

It would be clearer to have `readMeasurementByReferenceIds` return null in this case and have calling code determine whether that should be an exception.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/postgres/writers/SetMeasurementFailure.kt` line 34 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/718#-NDV920w59QkmRRfTeo9:-NDV920w59QkmRRfTeoA:b-lyspig) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a3f724784901a50d70e792fa514dfb81b9869ed8/src/main/kotlin/org/wfanet/measurement/reporting/deploy/postgres/writers/SetMeasurementFailure.kt#L34)):*
This should check the measurement state so it can throw an error if the transition is not allowed.

From https://google.aip.dev/216:
> If the state transition is not allowed, the service **must** error with `FAILED_PRECONDITION` (HTTP 400).
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/postgres/writers/SetMeasurementResult.kt` line 53 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/718#-NDVAuME8e7SKUc3HWUd:-NDVAuME8e7SKUc3HWUe:bq1s866) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a3f724784901a50d70e792fa514dfb81b9869ed8/src/main/kotlin/org/wfanet/measurement/reporting/deploy/postgres/writers/SetMeasurementResult.kt#L53)):*
> ```Kotlin
> 
>   override suspend fun TransactionScope.runTransaction(): Measurement {
>     MeasurementReader()
> ```

Same here re: error for invalid state transition.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/internal/testing/MeasurementsServiceTest.kt` line 515 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/718#-NDVBXOD-TnLitBIIU0L:-NDVBXOD-TnLitBIIU0M:bawax7m) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a3f724784901a50d70e792fa514dfb81b9869ed8/src/main/kotlin/org/wfanet/measurement/reporting/service/internal/testing/MeasurementsServiceTest.kt#L515)):*
> ```Kotlin
> 
>   @Test
>   fun `setMeasurementResult does nothing if the same request is called twice`() {
> ```

I don't think that this should be the default behavior. Non-GET methods are not inherently idempotent.

I think that this behavior should be controlled explicitly from the public API using request IDs (see https://google.aip.dev/155), rather than assuming that the caller always wants idempotent behavior.



_Code quote:_
```Kotlin
`setMeasurementResult does nothing if the same request is called twice`
```

SanjayVas (approved):
> Reviewed 7 of 7 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/718)* status: all files reviewed, 1 unresolved discussion (waiting on @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/postgres/readers/MeasurementReader.kt` line 44 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/718#-NDZ_SzP5iCX9xadDzT9:-NDZ_SzQ2sS7iJIqO1dq:bviv55s) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/99cb90fef04d08cad6e87fc6a32d7744faba25c1/src/main/kotlin/org/wfanet/measurement/reporting/deploy/postgres/readers/MeasurementReader.kt#L44)):*
> ```Kotlin
>   fun translate(row: ResultRow): Result = Result(buildMeasurement(row))
> 
>   /**
> ```

nit: maybe leave this in, but explain that it returns `null` when the measurement isn't found

duliomatos1 (approved):
> Reviewed 6 of 7 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/718)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @tristanvuong2021)

PR #719, opened 2022-10-04
Add additional error code case for reporting retries

tristanvuong2021:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/719)


SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/719)* status: all files reviewed, 1 unresolved discussion (waiting on @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/postgres/SerializableErrors.kt` line 39 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/719#-NDZOis4B5SK5V8wZgd9:-NDZOis4B5SK5V8wZgdA:brpjagw) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/37f37229d9968fdc8a05adcaf9f0d3d3c02091ba/src/main/kotlin/org/wfanet/measurement/reporting/deploy/postgres/SerializableErrors.kt#L39)):*
> If the serializable error doesn't propagate in time to other coroutines, the failed transaction will be used

I'm not quite sure I understand. What does propagation mean here? What does "the failed transaction will be used" mean? How is this specific to coroutines?

What I could find on this error code is that it generally has to do with a transaction not being closed (committed or rolled back). Maybe we're instead missing a case in our R2DBC code where `commitTransaction` isn't always run?

tristanvuong2021 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/719)* status: all files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/postgres/SerializableErrors.kt` line 39 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/719#-NDZOis4B5SK5V8wZgd9:-NDZSr6h18Psd4XiuND4:b-yq4mui) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/37f37229d9968fdc8a05adcaf9f0d3d3c02091ba/src/main/kotlin/org/wfanet/measurement/reporting/deploy/postgres/SerializableErrors.kt#L39)):*

> If the serializable error doesn't propagate in time to other coroutines, the failed transaction will be used

I'm not quite sure I understand. What does propagation mean here? What does "the failed transaction will be used" mean? How is this specific to coroutines?

What I could find on this error code is that it generally has to do with a transaction not being closed (committed or rolled back). Maybe we're instead missing a case in our R2DBC code where `commitTransaction` isn't always run?
</blockquote></details>

I can't find it in my action history right now, but my reports concurrent test  seems to occasionally fail with this error. In createReport, I use launch. If the serializable error occurs in a launch block, it propagates the exception https://kotlinlang.org/docs/exception-handling.html. I can't test it for sure, but I am thinking it is possible for the transaction to fail, but another launch block runs its queries before the block with the initial error is able to propagate the error to the rest. In this case, the new error wil be the one that surfaces.

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/719)* status: all files reviewed, 1 unresolved discussion (waiting on @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/postgres/SerializableErrors.kt` line 39 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/719#-NDZOis4B5SK5V8wZgd9:-NDZYRQE3AZCQgc3tDWv:bsmn9uq) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/37f37229d9968fdc8a05adcaf9f0d3d3c02091ba/src/main/kotlin/org/wfanet/measurement/reporting/deploy/postgres/SerializableErrors.kt#L39)):*
> If the serializable error occurs in a launch block, it propagates the exception

One of the key differences between `launch` and `async` is that the former *doesn't* propagate exceptions on `join`.

What I'm guessing is happening is something with coroutine cancellation, where the exception in the `launch` block tries to cancel the coroutine scope. This likely isn't handled properly somewhere, either by not expecting a coroutine CancellationException or not checking whether the coroutine context is still active. I'm also not 100% certain if calling `close()` on a `Connection` is sufficient or whether we need to make sure `rollbackTransaction()` is called on error.

tristanvuong2021:
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/postgres/SerializableErrors.kt` line 39 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/719#-NDZOis4B5SK5V8wZgd9:-NDZ_z4B4gbHAqgTBbMT:bo1g30n) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/37f37229d9968fdc8a05adcaf9f0d3d3c02091ba/src/main/kotlin/org/wfanet/measurement/reporting/deploy/postgres/SerializableErrors.kt#L39)):*

> If the serializable error occurs in a launch block, it propagates the exception

One of the key differences between `launch` and `async` is that the former *doesn't* propagate exceptions on `join`.

What I'm guessing is happening is something with coroutine cancellation, where the exception in the `launch` block tries to cancel the coroutine scope. This likely isn't handled properly somewhere, either by not expecting a coroutine CancellationException or not checking whether the coroutine context is still active. I'm also not 100% certain if calling `close()` on a `Connection` is sufficient or whether we need to make sure `rollbackTransaction()` is called on error.
</blockquote></details>

I don't actually use join in CreateReport.

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/719)* status: all files reviewed, 1 unresolved discussion (waiting on @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/postgres/SerializableErrors.kt` line 39 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/719#-NDZOis4B5SK5V8wZgd9:-NDZaKRU3aVNUU0ZCjNU:b45d140) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/37f37229d9968fdc8a05adcaf9f0d3d3c02091ba/src/main/kotlin/org/wfanet/measurement/reporting/deploy/postgres/SerializableErrors.kt#L39)):*

I don't actually use join in CreateReport.
</blockquote></details>

I assume you use `coroutineScope`, which calls `join`.

tristanvuong2021 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/719)* status: all files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/postgres/SerializableErrors.kt` line 39 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/719#-NDZOis4B5SK5V8wZgd9:-NDZbzXf0ZeSkdcv9_C-:b6hgxcr) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/37f37229d9968fdc8a05adcaf9f0d3d3c02091ba/src/main/kotlin/org/wfanet/measurement/reporting/deploy/postgres/SerializableErrors.kt#L39)):*

I assume you use `coroutineScope`, which calls `join`.
</blockquote></details>

It should still cancel then.  So queries are run using a failed transaction before the parent scope can cancel everything and the serializable error isn't surfaced.

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/719)* status: all files reviewed, 1 unresolved discussion (waiting on @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/postgres/SerializableErrors.kt` line 39 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/719#-NDZOis4B5SK5V8wZgd9:-NDZpgl6CZhAITbp8Wsr:b-maf6n6) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/37f37229d9968fdc8a05adcaf9f0d3d3c02091ba/src/main/kotlin/org/wfanet/measurement/reporting/deploy/postgres/SerializableErrors.kt#L39)):*

It should still cancel then.  So queries are run using a failed transaction before the parent scope can cancel everything and the serializable error isn't surfaced.
</blockquote></details>

Possibly. That's the "not checking whether the coroutine context is still active" case, where the scope cancellation isn't promptly handled by the sibling coroutines.

Either way, the fix isn't retrying on this error code but rather preventing it from happening. The first place I'd look is making sure that the transaction is closed even in the case of coroutine cancellation.

PR #728, opened 2022-10-11
Kingdom supports repeated protocols in public protocol_config

iverson52000:
- Kingdom supports repeated protocols in public protocol_config for requisition to specify if a measurement is direct reach and frequency measurement(single data provider). This way EDP can decide to do direct or MPC based reach and frequency computation for direct measurement. The internal protocol_config will remain unchanged
- Add Kingdom flag --allow-mpc-protocols-for-single-data-provider to control adding MPC protocol for single-edp reach and frequency measurement
- Update relevant unit tests and integrations tests

- - -
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/728)

### DIFF
--- a/build/repositories.bzl
+++ b/build/repositories.bzl
@@ -39,7 +39,7 @@ def wfa_measurement_system_repositories():
     wfa_repo_archive(
         name = "wfa_measurement_proto",
         repo = "cross-media-measurement-api",
-        commit = "9064af16be7521eac562139145e0456351f76361",
+        commit = "595462a63840d8a77f696b960e00906ecdf10028",
         sha256 = None,
         #        sha256 = "69ee69cbfa11ba90ca172d3141a9465a4408883e1aa559d56ef740bd01d474ff",
         #        version = "0.23.0",
--- a/build/repositories.bzl
+++ b/build/repositories.bzl
@@ -35,11 +35,14 @@ def wfa_measurement_system_repositories():
         version = "0.8.0",
     )
 
+    # DO_NOT_SUBMIT(world-federation-of-advertisers/cross-media-measurement-api PR): Replace with version once released from main.
     wfa_repo_archive(
         name = "wfa_measurement_proto",
         repo = "cross-media-measurement-api",
-        sha256 = "69ee69cbfa11ba90ca172d3141a9465a4408883e1aa559d56ef740bd01d474ff",
-        version = "0.23.0",
+        commit = "9064af16be7521eac562139145e0456351f76361",
+        sha256 = None,
+        #        sha256 = "69ee69cbfa11ba90ca172d3141a9465a4408883e1aa559d56ef740bd01d474ff",
+        #        version = "0.23.0",
     )
 
     wfa_repo_archive(
--- a/build/repositories.bzl
+++ b/build/repositories.bzl
@@ -35,14 +35,12 @@ def wfa_measurement_system_repositories():
         version = "0.8.0",
     )
 
-    # DO_NOT_SUBMIT(world-federation-of-advertisers/cross-media-measurement-api PR): Replace with version once released from main.
     wfa_repo_archive(
         name = "wfa_measurement_proto",
         repo = "cross-media-measurement-api",
         commit = "595462a63840d8a77f696b960e00906ecdf10028",
-        sha256 = None,
-        #        sha256 = "69ee69cbfa11ba90ca172d3141a9465a4408883e1aa559d56ef740bd01d474ff",
-        #        version = "0.23.0",
+        sha256 = "8412e478f15119b624e6696b578ca308b55f61a240e83ea2f72444692118d1ff",
+        version = "0.24.0",
     )
 
     wfa_repo_archive(
--- a/build/repositories.bzl
+++ b/build/repositories.bzl
@@ -38,7 +38,6 @@ def wfa_measurement_system_repositories():
     wfa_repo_archive(
         name = "wfa_measurement_proto",
         repo = "cross-media-measurement-api",
-        commit = "595462a63840d8a77f696b960e00906ecdf10028",
         sha256 = "8412e478f15119b624e6696b578ca308b55f61a240e83ea2f72444692118d1ff",
         version = "0.24.0",
     )
--- a/src/main/kotlin/org/wfanet/measurement/integration/common/InProcessKingdom.kt
+++ b/src/main/kotlin/org/wfanet/measurement/integration/common/InProcessKingdom.kt
@@ -14,6 +14,17 @@
 
 package org.wfanet.measurement.integration.common
 
+import io.grpc.Channel
+import java.util.concurrent.TimeUnit
+import java.util.logging.Logger
+import org.junit.rules.TestRule
+import org.junit.runner.Description
+import org.junit.runners.model.Statement
+import org.wfanet.measurement.api.v2alpha.testing.withMetadataPrincipalIdentities
+import org.wfanet.measurement.common.grpc.testing.GrpcTestServerRule
+import org.wfanet.measurement.common.grpc.withDefaultDeadline
+import org.wfanet.measurement.common.identity.testing.withMetadataDuchyIdentities
+import org.wfanet.measurement.common.testing.chainRulesSequentially
 import org.wfanet.measurement.internal.kingdom.AccountsGrpcKt.AccountsCoroutineStub as InternalAccountsCoroutineStub
 import org.wfanet.measurement.internal.kingdom.ApiKeysGrpcKt.ApiKeysCoroutineStub as InternalApiKeysCoroutineStub
 import org.wfanet.measurement.internal.kingdom.CertificatesGrpcKt.CertificatesCoroutineStub as InternalCertificatesCoroutineStub
@@ -28,21 +39,6 @@ import org.wfanet.measurement.internal.kingdom.MeasurementLogEntriesGrpcKt.Measu
 import org.wfanet.measurement.internal.kingdom.MeasurementsGrpcKt.MeasurementsCoroutineStub as InternalMeasurementsCoroutineStub
 import org.wfanet.measurement.internal.kingdom.PublicKeysGrpcKt.PublicKeysCoroutineStub as InternalPublicKeysCoroutineStub
 import org.wfanet.measurement.internal.kingdom.RequisitionsGrpcKt.RequisitionsCoroutineStub as InternalRequisitionsCoroutineStub
-import org.wfanet.measurement.kingdom.service.system.v1alpha.ComputationLogEntriesService as systemComputationLogEntriesService
-import org.wfanet.measurement.kingdom.service.system.v1alpha.ComputationParticipantsService as systemComputationParticipantsService
-import org.wfanet.measurement.kingdom.service.system.v1alpha.ComputationsService as systemComputationsService
-import org.wfanet.measurement.kingdom.service.system.v1alpha.RequisitionsService as systemRequisitionsService
-import io.grpc.Channel
-import java.util.concurrent.TimeUnit
-import java.util.logging.Logger
-import org.junit.rules.TestRule
-import org.junit.runner.Description
-import org.junit.runners.model.Statement
-import org.wfanet.measurement.api.v2alpha.testing.withMetadataPrincipalIdentities
-import org.wfanet.measurement.common.grpc.testing.GrpcTestServerRule
-import org.wfanet.measurement.common.grpc.withDefaultDeadline
-import org.wfanet.measurement.common.identity.testing.withMetadataDuchyIdentities
-import org.wfanet.measurement.common.testing.chainRulesSequentially
 import org.wfanet.measurement.kingdom.deploy.common.service.DataServices
 import org.wfanet.measurement.kingdom.deploy.common.service.toList
 import org.wfanet.measurement.kingdom.service.api.v2alpha.AccountsService
@@ -59,6 +55,10 @@ import org.wfanet.measurement.kingdom.service.api.v2alpha.PublicKeysService
 import org.wfanet.measurement.kingdom.service.api.v2alpha.RequisitionsService
 import org.wfanet.measurement.kingdom.service.api.v2alpha.withAccountAuthenticationServerInterceptor
 import org.wfanet.measurement.kingdom.service.api.v2alpha.withApiKeyAuthenticationServerInterceptor
+import org.wfanet.measurement.kingdom.service.system.v1alpha.ComputationLogEntriesService as systemComputationLogEntriesService
+import org.wfanet.measurement.kingdom.service.system.v1alpha.ComputationParticipantsService as systemComputationParticipantsService
+import org.wfanet.measurement.kingdom.service.system.v1alpha.ComputationsService as systemComputationsService
+import org.wfanet.measurement.kingdom.service.system.v1alpha.RequisitionsService as systemRequisitionsService
 
 /** TestRule that starts and stops all Kingdom gRPC services. */
 class InProcessKingdom(
@@ -127,29 +127,29 @@ class InProcessKingdom(
       logger.info("Building Kingdom's public API services")
 
       listOf(
-        ApiKeysService(internalApiKeysClient)
+          ApiKeysService(internalApiKeysClient)
             .withAccountAuthenticationServerInterceptor(internalAccountsClient, redirectUri),
-        CertificatesService(internalCertificatesClient)
+          CertificatesService(internalCertificatesClient)
             .withMetadataPrincipalIdentities()
             .withApiKeyAuthenticationServerInterceptor(internalApiKeysClient),
-        DataProvidersService(internalDataProvidersClient)
+          DataProvidersService(internalDataProvidersClient)
             .withMetadataPrincipalIdentities()
             .withApiKeyAuthenticationServerInterceptor(internalApiKeysClient),
-        EventGroupsService(internalEventGroupsClient)
+          EventGroupsService(internalEventGroupsClient)
             .withMetadataPrincipalIdentities()
             .withApiKeyAuthenticationServerInterceptor(internalApiKeysClient),
-        MeasurementsService(internalMeasurementsClient, allowMpcProtocolsForSingleDataProvider)
+          MeasurementsService(internalMeasurementsClient, allowMpcProtocolsForSingleDataProvider)
             .withMetadataPrincipalIdentities()
             .withApiKeyAuthenticationServerInterceptor(internalApiKeysClient),
-        PublicKeysService(internalPublicKeysClient)
+          PublicKeysService(internalPublicKeysClient)
             .withMetadataPrincipalIdentities()
             .withApiKeyAuthenticationServerInterceptor(internalApiKeysClient),
-        RequisitionsService(allowMpcProtocolsForSingleDataProvider, internalRequisitionsClient)
+          RequisitionsService(allowMpcProtocolsForSingleDataProvider, internalRequisitionsClient)
             .withMetadataPrincipalIdentities()
             .withApiKeyAuthenticationServerInterceptor(internalApiKeysClient),
-        AccountsService(internalAccountsClient, redirectUri)
+          AccountsService(internalAccountsClient, redirectUri)
             .withAccountAuthenticationServerInterceptor(internalAccountsClient, redirectUri),
-        MeasurementConsumersService(internalMeasurementConsumersClient)
+          MeasurementConsumersService(internalMeasurementConsumersClient)
             .withMetadataPrincipalIdentities()
             .withAccountAuthenticationServerInterceptor(internalAccountsClient, redirectUri)
             .withApiKeyAuthenticationServerInterceptor(internalApiKeysClient)
--- a/src/main/kotlin/org/wfanet/measurement/integration/common/InProcessKingdom.kt
+++ b/src/main/kotlin/org/wfanet/measurement/integration/common/InProcessKingdom.kt
@@ -14,17 +14,6 @@
 
 package org.wfanet.measurement.integration.common
 
-import io.grpc.Channel
-import java.util.concurrent.TimeUnit
-import java.util.logging.Logger
-import org.junit.rules.TestRule
-import org.junit.runner.Description
-import org.junit.runners.model.Statement
-import org.wfanet.measurement.api.v2alpha.testing.withMetadataPrincipalIdentities
-import org.wfanet.measurement.common.grpc.testing.GrpcTestServerRule
-import org.wfanet.measurement.common.grpc.withDefaultDeadline
-import org.wfanet.measurement.common.identity.testing.withMetadataDuchyIdentities
-import org.wfanet.measurement.common.testing.chainRulesSequentially
 import org.wfanet.measurement.internal.kingdom.AccountsGrpcKt.AccountsCoroutineStub as InternalAccountsCoroutineStub
 import org.wfanet.measurement.internal.kingdom.ApiKeysGrpcKt.ApiKeysCoroutineStub as InternalApiKeysCoroutineStub
 import org.wfanet.measurement.internal.kingdom.CertificatesGrpcKt.CertificatesCoroutineStub as InternalCertificatesCoroutineStub
@@ -39,6 +28,21 @@ import org.wfanet.measurement.internal.kingdom.MeasurementLogEntriesGrpcKt.Measu
 import org.wfanet.measurement.internal.kingdom.MeasurementsGrpcKt.MeasurementsCoroutineStub as InternalMeasurementsCoroutineStub
 import org.wfanet.measurement.internal.kingdom.PublicKeysGrpcKt.PublicKeysCoroutineStub as InternalPublicKeysCoroutineStub
 import org.wfanet.measurement.internal.kingdom.RequisitionsGrpcKt.RequisitionsCoroutineStub as InternalRequisitionsCoroutineStub
+import org.wfanet.measurement.kingdom.service.system.v1alpha.ComputationLogEntriesService as systemComputationLogEntriesService
+import org.wfanet.measurement.kingdom.service.system.v1alpha.ComputationParticipantsService as systemComputationParticipantsService
+import org.wfanet.measurement.kingdom.service.system.v1alpha.ComputationsService as systemComputationsService
+import org.wfanet.measurement.kingdom.service.system.v1alpha.RequisitionsService as systemRequisitionsService
+import io.grpc.Channel
+import java.util.concurrent.TimeUnit
+import java.util.logging.Logger
+import org.junit.rules.TestRule
+import org.junit.runner.Description
+import org.junit.runners.model.Statement
+import org.wfanet.measurement.api.v2alpha.testing.withMetadataPrincipalIdentities
+import org.wfanet.measurement.common.grpc.testing.GrpcTestServerRule
+import org.wfanet.measurement.common.grpc.withDefaultDeadline
+import org.wfanet.measurement.common.identity.testing.withMetadataDuchyIdentities
+import org.wfanet.measurement.common.testing.chainRulesSequentially
 import org.wfanet.measurement.kingdom.deploy.common.service.DataServices
 import org.wfanet.measurement.kingdom.deploy.common.service.toList
 import org.wfanet.measurement.kingdom.service.api.v2alpha.AccountsService
@@ -55,10 +59,6 @@ import org.wfanet.measurement.kingdom.service.api.v2alpha.PublicKeysService
 import org.wfanet.measurement.kingdom.service.api.v2alpha.RequisitionsService
 import org.wfanet.measurement.kingdom.service.api.v2alpha.withAccountAuthenticationServerInterceptor
 import org.wfanet.measurement.kingdom.service.api.v2alpha.withApiKeyAuthenticationServerInterceptor
-import org.wfanet.measurement.kingdom.service.system.v1alpha.ComputationLogEntriesService as systemComputationLogEntriesService
-import org.wfanet.measurement.kingdom.service.system.v1alpha.ComputationParticipantsService as systemComputationParticipantsService
-import org.wfanet.measurement.kingdom.service.system.v1alpha.ComputationsService as systemComputationsService
-import org.wfanet.measurement.kingdom.service.system.v1alpha.RequisitionsService as systemRequisitionsService
 
 /** TestRule that starts and stops all Kingdom gRPC services. */
 class InProcessKingdom(
@@ -66,6 +66,7 @@ class InProcessKingdom(
   val verboseGrpcLogging: Boolean = true,
   /** The open id client redirect uri when creating the authentication uri. */
   private val redirectUri: String,
+  private val allowMpcProtocolsForSingleDataProvider: Boolean,
 ) : TestRule {
   private val kingdomDataServices by lazy { dataServicesProvider() }
 
@@ -126,29 +127,29 @@ class InProcessKingdom(
       logger.info("Building Kingdom's public API services")
 
       listOf(
-          ApiKeysService(internalApiKeysClient)
+        ApiKeysService(internalApiKeysClient)
             .withAccountAuthenticationServerInterceptor(internalAccountsClient, redirectUri),
-          CertificatesService(internalCertificatesClient)
+        CertificatesService(internalCertificatesClient)
             .withMetadataPrincipalIdentities()
             .withApiKeyAuthenticationServerInterceptor(internalApiKeysClient),
-          DataProvidersService(internalDataProvidersClient)
+        DataProvidersService(internalDataProvidersClient)
             .withMetadataPrincipalIdentities()
             .withApiKeyAuthenticationServerInterceptor(internalApiKeysClient),
-          EventGroupsService(internalEventGroupsClient)
+        EventGroupsService(internalEventGroupsClient)
             .withMetadataPrincipalIdentities()
             .withApiKeyAuthenticationServerInterceptor(internalApiKeysClient),
-          MeasurementsService(internalMeasurementsClient)
+        MeasurementsService(internalMeasurementsClient, allowMpcProtocolsForSingleDataProvider)
             .withMetadataPrincipalIdentities()
             .withApiKeyAuthenticationServerInterceptor(internalApiKeysClient),
-          PublicKeysService(internalPublicKeysClient)
+        PublicKeysService(internalPublicKeysClient)
             .withMetadataPrincipalIdentities()
             .withApiKeyAuthenticationServerInterceptor(internalApiKeysClient),
-          RequisitionsService(internalRequisitionsClient)
+        RequisitionsService(allowMpcProtocolsForSingleDataProvider, internalRequisitionsClient)
             .withMetadataPrincipalIdentities()
             .withApiKeyAuthenticationServerInterceptor(internalApiKeysClient),
-          AccountsService(internalAccountsClient, redirectUri)
+        AccountsService(internalAccountsClient, redirectUri)
             .withAccountAuthenticationServerInterceptor(internalAccountsClient, redirectUri),
-          MeasurementConsumersService(internalMeasurementConsumersClient)
+        MeasurementConsumersService(internalMeasurementConsumersClient)
             .withMetadataPrincipalIdentities()
             .withAccountAuthenticationServerInterceptor(internalAccountsClient, redirectUri)
             .withApiKeyAuthenticationServerInterceptor(internalApiKeysClient)
--- a/src/main/kotlin/org/wfanet/measurement/integration/common/InProcessLifeOfAMeasurementIntegrationTest.kt
+++ b/src/main/kotlin/org/wfanet/measurement/integration/common/InProcessLifeOfAMeasurementIntegrationTest.kt
@@ -14,6 +14,14 @@
 
 package org.wfanet.measurement.integration.common
 
+import org.wfanet.measurement.api.v2alpha.AccountsGrpcKt.AccountsCoroutineStub as PublicAccountsCoroutineStub
+import org.wfanet.measurement.api.v2alpha.ApiKeysGrpcKt.ApiKeysCoroutineStub as PublicApiKeysCoroutineStub
+import org.wfanet.measurement.api.v2alpha.CertificatesGrpcKt.CertificatesCoroutineStub as PublicCertificatesCoroutineStub
+import org.wfanet.measurement.api.v2alpha.DataProvidersGrpcKt.DataProvidersCoroutineStub as PublicDataProvidersCoroutineStub
+import org.wfanet.measurement.api.v2alpha.EventGroupsGrpcKt.EventGroupsCoroutineStub as PublicEventGroupsCoroutineStub
+import org.wfanet.measurement.api.v2alpha.MeasurementConsumersGrpcKt.MeasurementConsumersCoroutineStub as PublicMeasurementConsumersCoroutineStub
+import org.wfanet.measurement.api.v2alpha.MeasurementsGrpcKt.MeasurementsCoroutineStub as PublicMeasurementsCoroutineStub
+import org.wfanet.measurement.api.v2alpha.RequisitionsGrpcKt.RequisitionsCoroutineStub as PublicRequisitionsCoroutineStub
 import java.time.Duration
 import kotlinx.coroutines.runBlocking
 import org.junit.After
@@ -22,15 +30,7 @@ import org.junit.BeforeClass
 import org.junit.Rule
 import org.junit.Test
 import org.junit.rules.TestRule
-import org.wfanet.measurement.api.v2alpha.AccountsGrpcKt.AccountsCoroutineStub as PublicAccountsCoroutineStub
-import org.wfanet.measurement.api.v2alpha.ApiKeysGrpcKt.ApiKeysCoroutineStub as PublicApiKeysCoroutineStub
-import org.wfanet.measurement.api.v2alpha.CertificatesGrpcKt.CertificatesCoroutineStub as PublicCertificatesCoroutineStub
-import org.wfanet.measurement.api.v2alpha.DataProvidersGrpcKt.DataProvidersCoroutineStub as PublicDataProvidersCoroutineStub
 import org.wfanet.measurement.api.v2alpha.EventGroup
-import org.wfanet.measurement.api.v2alpha.EventGroupsGrpcKt.EventGroupsCoroutineStub as PublicEventGroupsCoroutineStub
-import org.wfanet.measurement.api.v2alpha.MeasurementConsumersGrpcKt.MeasurementConsumersCoroutineStub as PublicMeasurementConsumersCoroutineStub
-import org.wfanet.measurement.api.v2alpha.MeasurementsGrpcKt.MeasurementsCoroutineStub as PublicMeasurementsCoroutineStub
-import org.wfanet.measurement.api.v2alpha.RequisitionsGrpcKt.RequisitionsCoroutineStub as PublicRequisitionsCoroutineStub
 import org.wfanet.measurement.api.v2alpha.differentialPrivacyParams
 import org.wfanet.measurement.common.identity.DuchyInfo
 import org.wfanet.measurement.common.testing.ProviderRule
@@ -52,6 +52,7 @@ private val OUTPUT_DP_PARAMS = differentialPrivacyParams {
 }
 private const val REDIRECT_URI = "https://localhost:2048"
 private val RESULT_POLLING_DELAY = Duration.ofSeconds(10)
+private const val ALLOW_MPC_PROTOCOLS_FOR_SINGLE_DATA_PROVIDER = true
 
 /**
  * Test that everything is wired up properly.
@@ -74,7 +75,8 @@ abstract class InProcessLifeOfAMeasurementIntegrationTest {
     InProcessKingdom(
       dataServicesProvider = { kingdomDataServices },
       verboseGrpcLogging = false,
-      REDIRECT_URI
+      REDIRECT_URI,
+      ALLOW_MPC_PROTOCOLS_FOR_SINGLE_DATA_PROVIDER
     )
 
   private val duchies: List<InProcessDuchy> by lazy {
--- a/src/main/kotlin/org/wfanet/measurement/integration/common/InProcessLifeOfAMeasurementIntegrationTest.kt
+++ b/src/main/kotlin/org/wfanet/measurement/integration/common/InProcessLifeOfAMeasurementIntegrationTest.kt
@@ -14,14 +14,6 @@
 
 package org.wfanet.measurement.integration.common
 
-import org.wfanet.measurement.api.v2alpha.AccountsGrpcKt.AccountsCoroutineStub as PublicAccountsCoroutineStub
-import org.wfanet.measurement.api.v2alpha.ApiKeysGrpcKt.ApiKeysCoroutineStub as PublicApiKeysCoroutineStub
-import org.wfanet.measurement.api.v2alpha.CertificatesGrpcKt.CertificatesCoroutineStub as PublicCertificatesCoroutineStub
-import org.wfanet.measurement.api.v2alpha.DataProvidersGrpcKt.DataProvidersCoroutineStub as PublicDataProvidersCoroutineStub
-import org.wfanet.measurement.api.v2alpha.EventGroupsGrpcKt.EventGroupsCoroutineStub as PublicEventGroupsCoroutineStub
-import org.wfanet.measurement.api.v2alpha.MeasurementConsumersGrpcKt.MeasurementConsumersCoroutineStub as PublicMeasurementConsumersCoroutineStub
-import org.wfanet.measurement.api.v2alpha.MeasurementsGrpcKt.MeasurementsCoroutineStub as PublicMeasurementsCoroutineStub
-import org.wfanet.measurement.api.v2alpha.RequisitionsGrpcKt.RequisitionsCoroutineStub as PublicRequisitionsCoroutineStub
 import java.time.Duration
 import kotlinx.coroutines.runBlocking
 import org.junit.After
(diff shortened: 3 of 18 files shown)


matthewclegg (approved):
> Reviewed 7 of 7 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/728)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @rmirisola, @robinsons, @SanjayVas, and @xnegron)

SanjayVas (changes requested):
> Reviewed 4 of 7 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/728)* status: all files reviewed, 3 unresolved discussions (waiting on @iverson52000, @rmirisola, @robinsons, @stevenwarejones, and @xnegron)

___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/ProtoConversions.kt` line 24 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/728#-NENH97tBK6z5Kmgp9hC:-NENH97tBK6z5Kmgp9hD:b5wts2s) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/339f40f33c17c63addffdc33012c5a7a99e59253/src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/ProtoConversions.kt#L24)):*
> ```Kotlin
> import org.wfanet.measurement.internal.kingdom.Measurement as InternalMeasurement
> import org.wfanet.measurement.internal.kingdom.ProtocolConfig as InternalProtocolConfig
> import org.wfanet.measurement.internal.kingdom.ProtocolConfigKt.direct as internalDirect
> ```

This could trigger https://youtrack.jetbrains.com/issue/KTIJ-16471/

Avoid import aliases for functions on objects/classes. Instead, alias the enclosing type. e.g.

```kotlin
import org.wfanet.measurement.internal.kingdom.ProtocolConfigKt as InternalProtocolConfigs

//...

InternalProtocolConfigs.direct {
}
```



_Code quote:_
```Kotlin
import org.wfanet.measurement.internal.kingdom.ProtocolConfigKt.direct as internalDirect
```
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/ProtoConversions.kt` line 205 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/728#-NENJd-rEDIlLHeX8zRT:-NENJd-rEDIlLHeX8zRU:b-7khm7a) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/339f40f33c17c63addffdc33012c5a7a99e59253/src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/ProtoConversions.kt#L205)):*
> ```Kotlin
>     dataProviders +=
>       source.dataProvidersMap.entries.map(Map.Entry<Long, DataProviderValue>::toDataProviderEntry)
>     if (source.details.hasProtocolConfig()) {
> ```

Note that this must be able to handle older persisted DB entries, but always produce the newest version of the v2alpha message.

This means that e.g. if there was an old impression measurement that was persisted before this PR, this code must produce the same message as if it was persisted after this PR.
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt` line 209 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/728#-NENIVRN5nyOvBQFC4QZ:-NENIVRN5nyOvBQFC4Q_:b-k7xv79) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/339f40f33c17c63addffdc33012c5a7a99e59253/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt#L209)):*
> ```Kotlin
>       }
> 
>       if (requisition.protocolConfig.protocolsList.last().hasDirect()) {
> ```

This is unsafe. The order of protocols in the API is not defined. Therefore this should look for whether *any* protocol in the list is direct.



_Code quote:_
```Kotlin
last().hasDirect()
```

iverson52000:
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/ProtoConversions.kt` line 205 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/728#-NENJd-rEDIlLHeX8zRT:-NEbOpDMAxVL78TXvTyt:b-svxx3z) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/339f40f33c17c63addffdc33012c5a7a99e59253/src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/ProtoConversions.kt#L205)):*

Note that this must be able to handle older persisted DB entries, but always produce the newest version of the v2alpha message.

This means that e.g. if there was an old impression measurement that was persisted before this PR, this code must produce the same message as if it was persisted after this PR.
</blockquote></details>

Do you mean here public `protocolConfig` has to be generated even if `source.details.protocolConfig` is missing?



_Code quote:_
```Kotlin
source.details.hasProtocolConfig()
```

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/728)* status: all files reviewed, 3 unresolved discussions (waiting on @iverson52000, @rmirisola, @robinsons, @stevenwarejones, and @xnegron)

___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/ProtoConversions.kt` line 205 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/728#-NENJd-rEDIlLHeX8zRT:-NEfwI3DEyhpuviZrelz:b4f968o) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/339f40f33c17c63addffdc33012c5a7a99e59253/src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/ProtoConversions.kt#L205)):*

Do you mean here public `protocolConfig` has to be generated even if `source.details.protocolConfig` is missing?
</blockquote></details>

Yes.

Basically we need to follow the new public API definition regardless of what may be in existing DB entries.

iverson52000:
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/ProtoConversions.kt` line 205 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/728#-NENJd-rEDIlLHeX8zRT:-NEgFX0oA6auD-kHW4h8:b-c5yt8c) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/339f40f33c17c63addffdc33012c5a7a99e59253/src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/ProtoConversions.kt#L205)):*

Yes.

Basically we need to follow the new public API definition regardless of what may be in existing DB entries.
</blockquote></details>

Thanks so it's something like this? But here where do I get `protocolConfig.name` and `protocolConfig.measurementType` from? The source object of this function is `InternalMeasurement` and seems those two fields are in `InternalMeasurement.protocolConfig` and it's missing in this case



_Code snippet:_
```Kotlin
if (source.details.hasProtocolConfig()) {
      protocolConfig = source.details.protocolConfig.toProtocolConfig()
    } else {
      protocolConfig = protocolConfig {
        name = ProtocolConfigKey(source.externalProtocolConfigId).toName()
        @Suppress("WHEN_ENUM_CAN_BE_NULL_IN_JAVA") // Proto enum fields are never null.
        measurementType =
          when (source.measurementType) {
            InternalProtocolConfig.MeasurementType.MEASUREMENT_TYPE_UNSPECIFIED ->
              ProtocolConfig.MeasurementType.MEASUREMENT_TYPE_UNSPECIFIED
            InternalProtocolConfig.MeasurementType.REACH_AND_FREQUENCY ->
              ProtocolConfig.MeasurementType.REACH_AND_FREQUENCY
            InternalProtocolConfig.MeasurementType.IMPRESSION ->
              ProtocolConfig.MeasurementType.IMPRESSION
            InternalProtocolConfig.MeasurementType.DURATION ->
              ProtocolConfig.MeasurementType.DURATION
            InternalProtocolConfig.MeasurementType.UNRECOGNIZED ->
              error("MeasurementType unrecognized.")
          }
        protocols += protocol { direct = direct {} }
      }
    }
```

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/728)* status: all files reviewed, 3 unresolved discussions (waiting on @iverson52000, @rmirisola, @robinsons, @stevenwarejones, and @xnegron)

___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/ProtoConversions.kt` line 205 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/728#-NENJd-rEDIlLHeX8zRT:-NEl-ztx674qZIo_VRfj:b-si6n6f) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/339f40f33c17c63addffdc33012c5a7a99e59253/src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/ProtoConversions.kt#L205)):*
> The source object of this function is InternalMeasurement and seems those two fields are in InternalMeasurement.protocolConfig and it's missing in this case

You may need to pass in additional parameters to have enough information. For example, items from the MeasurementSpec.

iverson52000 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/728)* status: 3 of 8 files reviewed, 3 unresolved discussions (waiting on @matthewclegg, @rmirisola, @robinsons, @SanjayVas, @stevenwarejones, and @xnegron)

___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/ProtoConversions.kt` line 24 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/728#-NENH97tBK6z5Kmgp9hC:-NEqGSkO42qUpZsfiEn4:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/339f40f33c17c63addffdc33012c5a7a99e59253/src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/ProtoConversions.kt#L24)):*

This could trigger https://youtrack.jetbrains.com/issue/KTIJ-16471/

Avoid import aliases for functions on objects/classes. Instead, alias the enclosing type. e.g.

```kotlin
import org.wfanet.measurement.internal.kingdom.ProtocolConfigKt as InternalProtocolConfigs

//...

InternalProtocolConfigs.direct {
}
```
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/ProtoConversions.kt` line 205 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/728#-NENJd-rEDIlLHeX8zRT:-NEqGWLA1GPw9jQV2v41:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/339f40f33c17c63addffdc33012c5a7a99e59253/src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/ProtoConversions.kt#L205)):*

> The source object of this function is InternalMeasurement and seems those two fields are in InternalMeasurement.protocolConfig and it's missing in this case

You may need to pass in additional parameters to have enough information. For example, items from the MeasurementSpec.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt` line 209 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/728#-NENIVRN5nyOvBQFC4QZ:-NEqGY8W5LgYU0FkWOhY:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/339f40f33c17c63addffdc33012c5a7a99e59253/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt#L209)):*

This is unsafe. The order of protocols in the API is not defined. Therefore this should look for whether *any* protocol in the list is direct.
</blockquote></details>

Done.

stevenwarejones (changes requested):
> Reviewed 1 of 7 files at r1, 2 of 5 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/728)* status: 5 of 8 files reviewed, 4 unresolved discussions (waiting on @iverson52000, @matthewclegg, @rmirisola, @robinsons, @SanjayVas, and @xnegron)

___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/ProtoConversions.kt` line 317 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/728#-NEv9B-2357A8UriUHUK:-NEv9B-2357A8UriUHUL:b-ochqk3) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d5ee7a410c5859f1b172ea8106c8bffd9b1a0757/src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/ProtoConversions.kt#L317)):*
> ```Kotlin
>             liquidLegionsV2 = Llv2ProtocolConfig.protocolConfig
>             protocols +=
>               InternalProtocolConfigs.protocol {
> ```

I'm re-thinking that there should be a kingdom-level flag whether single edp measurements should allow liquidLegionsV2. I realize this is kind of a hybrid of what we discussed yesterday. But I think it's fine to allow a list of protocols, but by default, the kingdom should not add liquid legions as a protocol choice for single edp measurements unless that setting is on.

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/728)* status: 5 of 8 files reviewed, 4 unresolved discussions (waiting on @iverson52000, @matthewclegg, @rmirisola, @robinsons, @SanjayVas, and @xnegron)

___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/ProtoConversions.kt` line 317 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/728#-NEv9B-2357A8UriUHUK:-NEv9zgHEPNJ29WUlbr7:b6enexp) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d5ee7a410c5859f1b172ea8106c8bffd9b1a0757/src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/ProtoConversions.kt#L317)):*

I'm re-thinking that there should be a kingdom-level flag whether single edp measurements should allow liquidLegionsV2. I realize this is kind of a hybrid of what we discussed yesterday. But I think it's fine to allow a list of protocols, but by default, the kingdom should not add liquid legions as a protocol choice for single edp measurements unless that setting is on.
</blockquote></details>

To clarify, I'm imagining the kingdom-level setting is for giving edps the additional option of mpc fulfillment of single edp measurements rather than forcing an edp to one or the other.

iverson52000:
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/ProtoConversions.kt` line 317 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/728#-NEv9B-2357A8UriUHUK:-NFA0YpNBZkrQROKOQn0:be43b22) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d5ee7a410c5859f1b172ea8106c8bffd9b1a0757/src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/ProtoConversions.kt#L317)):*

To clarify, I'm imagining the kingdom-level setting is for giving edps the additional option of mpc fulfillment of single edp measurements rather than forcing an edp to one or the other.
</blockquote></details>

Yes this PR was created before our meeting last week. @sanjayvas what's your suggestions here?

SanjayVas (changes requested):
> Reviewed 1 of 7 files at r1, 5 of 5 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/728)* status: all files reviewed, 6 unresolved discussions (waiting on @iverson52000, @rmirisola, @robinsons, @stevenwarejones, and @xnegron)

___
*[`build/repositories.bzl` line 41 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/728#-NFAPMcMCT2Vxf3mHuNA:-NFAPMcMCT2Vxf3mHuNB:bsxknah) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/958579b1b4577d1a984e9c9b1117115ac23f3513/build/repositories.bzl#L41)):*
> ```Starlark
>         name = "wfa_measurement_proto",
>         repo = "cross-media-measurement-api",
>         commit = "595462a63840d8a77f696b960e00906ecdf10028",
> ```

Delete this. You should only have one of (commit, version).



_Code quote:_
```Starlark
        commit = "595462a63840d8a77f696b960e00906ecdf10028",
```
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/ProtoConversions.kt` line 205 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/728#-NENJd-rEDIlLHeX8zRT:-NFAQbj2BQTLjQKM3ZDN:b-ybvzpw) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/339f40f33c17c63addffdc33012c5a7a99e59253/src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/ProtoConversions.kt#L205)):*

Done.
</blockquote></details>

I suppose this is okay, but I'm thinking it actually might be better for now to just avoid changing the internal protocol config at all, aside from maybe deprecating its MeasurementType enum and always reading it from the measurement spec.

We can change the internal protos if we need to support more complicated protocol config scenarios in the future. (Basically, the reason we use separate protos for persistence and for the API is so they can evolve independently like this).
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/ProtoConversions.kt` line 317 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/728#-NEv9B-2357A8UriUHUK:-NFARILjBaPSsWopvd6-:b-j7mzec) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d5ee7a410c5859f1b172ea8106c8bffd9b1a0757/src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/ProtoConversions.kt#L317)):*

Yes this PR was created before our meeting last week. @sanjayvas what's your suggestions here? 
</blockquote></details>

Yes, what we discussed was having a flag that controls whether single-EDP R/F measurements allow an MPC protocol (no strong opinion yet on how specific the flag should be, e.g. `--allow-mpc-protocols-for-single-data-provider` vs. being specific about LLv2 and R/F).
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/ProtoConversions.kt` line 172 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/728#-NFAQ53u4tjWSpj6UH8P:-NFAQ53u4tjWSpj6UH8Q:bsr0r6w) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/958579b1b4577d1a984e9c9b1117115ac23f3513/src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/ProtoConversions.kt#L172)):*
> ```Kotlin
> 
>     source.protocolsList.forEach { sourceProtocol ->
>       if (sourceProtocol.hasLiquidLegionsV2()) {
> ```

Use an exhaustive `when` here on `protocolCase` so that we're required to handle any new items added to the oneof or else get a compilation error.
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/ProtoConversions.kt` line 212 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/728#-NFAavfEBd2SOsj5OnWj:-NFAavfEBd2SOsj5OnWk:b-rxbvx9) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/958579b1b4577d1a984e9c9b1117115ac23f3513/src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/ProtoConversions.kt#L212)):*
> ```Kotlin
>       // don't have protocolConfig, generate here.
> 
>       val parsedMeasurementSpec = MeasurementSpec.parseFrom(this.measurementSpec.data)
> ```

This is not safe without checking that the `api_version` in the internal `Measurement.Details` is `v2alpha`.



_Code quote:_
```Kotlin
val parsedMeasurementSpec = MeasurementSpec.parseFrom(this.measurementSpec.data)
```
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/RequisitionsService.kt` line 309 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/728#-NFAnKQz0eXwmIRgpd54:-NFAnKQz0eXwmIRgpd55:buhtmr3) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/958579b1b4577d1a984e9c9b1117115ac23f3513/src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/RequisitionsService.kt#L309)):*
> ```Kotlin
>       signature = parentMeasurement.measurementSpecSignature
>     }
>     if (parentMeasurement.hasProtocolConfig()) {
> ```

Won't this always be true now?

stevenwarejones (changes requested):
> Reviewed 3 of 5 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/728)* status: all files reviewed, 8 unresolved discussions (waiting on @iverson52000, @rmirisola, @robinsons, @SanjayVas, and @xnegron)

___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/ProtoConversions.kt` line 317 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/728#-NEv9B-2357A8UriUHUK:-NFEtL0s-9RADkXQuWbq:bgqyemo) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d5ee7a410c5859f1b172ea8106c8bffd9b1a0757/src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/ProtoConversions.kt#L317)):*

Yes, what we discussed was having a flag that controls whether single-EDP R/F measurements allow an MPC protocol (no strong opinion yet on how specific the flag should be, e.g. `--allow-mpc-protocols-for-single-data-provider` vs. being specific about LLv2 and R/F).
</blockquote></details>

+1 on what Sanjay suggests - works for me. `--allow-mpc-protocols-for-single-data-provider-measurement`
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/ProtoConversions.kt` line 172 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/728#-NFAQ53u4tjWSpj6UH8P:-NFEt9BO-BjHyXVX-Aub:b32v57) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/958579b1b4577d1a984e9c9b1117115ac23f3513/src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/ProtoConversions.kt#L172)):*

Use an exhaustive `when` here on `protocolCase` so that we're required to handle any new items added to the oneof or else get a compilation error.
</blockquote></details>

+1
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt` line 209 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/728#-NFAkybh4KugZgJcqBBT:-NFAkybh4KugZgJcqBBU:buyseqv) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/958579b1b4577d1a984e9c9b1117115ac23f3513/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt#L209)):*
> ```Kotlin
>       }
> 
>       if (requisition.protocolConfig.protocolsList.any { protocol -> protocol.hasDirect() }) {
> ```

```
if (!requisition.hasProtocolConfig() || requisition.protocolConfig.protocolsList.any { protocol -> protocol.hasDirect() }) {
```
to support deprecated requisitions without a protocolConfig
___
*[`src/test/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulatorTest.kt` line 695 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/728#-NFEu62-EhEIsrMGqBGt:-NFEu62-EhEIsrMGqBGu:b-odznz) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/958579b1b4577d1a984e9c9b1117115ac23f3513/src/test/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulatorTest.kt#L695)):*
> ```Kotlin
>       encryptedRequisitionSpec = ENCRYPTED_REQUISITION_ONE_SPEC
>       protocolConfig = protocolConfig {
>         liquidLegionsV2 =
> ```

this field is deprecated. Can we just remove?

stevenwarejones (changes requested):
> Reviewed 16 of 18 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/728)* status: 16 of 18 files reviewed, 7 unresolved discussions (waiting on @iverson52000, @rmirisola, @robinsons, @SanjayVas, and @xnegron)

___
*[`src/main/proto/wfa/measurement/internal/kingdom/protocol_config.proto` line 33 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/728#-NFoe8lx0Ot_pGhetHlZ:-NFoe8lx0Ot_pGhetHl_:b-vuqirc) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d5ee7a410c5859f1b172ea8106c8bffd9b1a0757/src/main/proto/wfa/measurement/internal/kingdom/protocol_config.proto#L33)):*
> ```protobuf
>     MEASUREMENT_TYPE_UNSPECIFIED = 0;
>     REACH_AND_FREQUENCY = 1;
>     IMPRESSION = 2;
> ```

why is this removed?

iverson52000 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/728)* status: 16 of 18 files reviewed, 7 unresolved discussions (waiting on @rmirisola, @robinsons, @SanjayVas, @stevenwarejones, and @xnegron)

___
*[`build/repositories.bzl` line 41 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/728#-NFAPMcMCT2Vxf3mHuNA:-NFooukP0E9pQmQuWy80:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/958579b1b4577d1a984e9c9b1117115ac23f3513/build/repositories.bzl#L41)):*

Delete this. You should only have one of (commit, version).
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/ProtoConversions.kt` line 205 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/728#-NENJd-rEDIlLHeX8zRT:-NFooUoZDFWUUDgHOKL1:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/339f40f33c17c63addffdc33012c5a7a99e59253/src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/ProtoConversions.kt#L205)):*

I suppose this is okay, but I'm thinking it actually might be better for now to just avoid changing the internal protocol config at all, aside from maybe deprecating its MeasurementType enum and always reading it from the measurement spec.

We can change the internal protos if we need to support more complicated protocol config scenarios in the future. (Basically, the reason we use separate protos for persistence and for the API is so they can evolve independently like this).
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/ProtoConversions.kt` line 317 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/728#-NEv9B-2357A8UriUHUK:-NFopEK471TcL6WnV69a:bi12dxz) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d5ee7a410c5859f1b172ea8106c8bffd9b1a0757/src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/ProtoConversions.kt#L317)):*
Created the Kingdom flag:
> -allow-mpc-protocols-for-single-data-provider-measurement
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/ProtoConversions.kt` line 172 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/728#-NFAQ53u4tjWSpj6UH8P:-NFooxZn8yrtRC55X1VO:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/958579b1b4577d1a984e9c9b1117115ac23f3513/src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/ProtoConversions.kt#L172)):*

+1
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/ProtoConversions.kt` line 212 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/728#-NFAavfEBd2SOsj5OnWj:-NFop2NiAmxyFA2_jQ_6:b-w5pfyz) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/958579b1b4577d1a984e9c9b1117115ac23f3513/src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/ProtoConversions.kt#L212)):*

This is not safe without checking that the `api_version` in the internal `Measurement.Details` is `v2alpha`.
</blockquote></details>

API version is checked at the top of the function
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/RequisitionsService.kt` line 309 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/728#-NFAnKQz0eXwmIRgpd54:-NFooln38TaeewDklEIi:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/958579b1b4577d1a984e9c9b1117115ac23f3513/src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/RequisitionsService.kt#L309)):*

Won't this always be true now?
</blockquote></details>

Done.
___
*[`src/main/proto/wfa/measurement/internal/kingdom/protocol_config.proto` line 33 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/728#-NFoe8lx0Ot_pGhetHlZ:-NFopOCuESGv6bXuTiLH:b-h388lj) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d5ee7a410c5859f1b172ea8106c8bffd9b1a0757/src/main/proto/wfa/measurement/internal/kingdom/protocol_config.proto#L33)):*
Per Sanjay's suggestion, leaving internal `protocol_config` unchanged for now

> I suppose this is okay, but I'm thinking it actually might be better for now to just avoid changing the internal protocol config at all, aside from maybe deprecating its MeasurementType enum and always reading it from the measurement spec.
> We can change the internal protos if we need to support more complicated protocol config scenarios in the future. (Basically, the reason we use separate protos for persistence and for the API is so they can evolve independently like this).

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/728)* status: 16 of 18 files reviewed, 6 unresolved discussions (waiting on @iverson52000, @rmirisola, @robinsons, @SanjayVas, and @xnegron)

___
*[`src/main/proto/wfa/measurement/internal/kingdom/protocol_config.proto` line 33 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/728#-NFoe8lx0Ot_pGhetHlZ:-NFtjKQo7O4OCvq-BJ7E:b-5iylg0) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d5ee7a410c5859f1b172ea8106c8bffd9b1a0757/src/main/proto/wfa/measurement/internal/kingdom/protocol_config.proto#L33)):*

Per Sanjay's suggestion, leaving internal `protocol_config` unchanged for now

> I suppose this is okay, but I'm thinking it actually might be better for now to just avoid changing the internal protocol config at all, aside from maybe deprecating its MeasurementType enum and always reading it from the measurement spec.
> We can change the internal protos if we need to support more complicated protocol config scenarios in the future. (Basically, the reason we use separate protos for persistence and for the API is so they can evolve independently like this).

</blockquote></details>

what is the use case for keeping MeasurementType at all here?

SanjayVas (changes requested):
> Reviewed 4 of 18 files at r4, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/728)* status: 17 of 18 files reviewed, 4 unresolved discussions (waiting on @iverson52000, @rmirisola, @robinsons, @stevenwarejones, and @xnegron)

___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/readers/RequisitionReader.kt` line 58 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/728#-NFyXDTD09WWKMKUHm_9:-NFyXDTD09WWKMKUHm_A:bvkydqh) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5fca28083d94c8b3c8fe6347f6018cc1a0130e5d/src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/readers/RequisitionReader.kt#L58)):*
> ```Kotlin
>     Measurements.State AS MeasurementState,
>     MeasurementDetails,
>     (
> ```

I don't think you need this. There is exactly one Requisition for each DataProvider in a given Measurement, meaning DataProvidersCount is the number of Requisitions.
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/ProtoConversions.kt` line 136 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/728#-NFyYM-v20GC1F59zJCK:-NFyYM-v20GC1F59zJCL:bmpbigr) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5fca28083d94c8b3c8fe6347f6018cc1a0130e5d/src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/ProtoConversions.kt#L136)):*
> ```Kotlin
> /** @throws [IllegalStateException] if MeasurementType not specified */
> fun InternalProtocolConfig.toProtocolConfig(
>   parsedMeasurementSpec: MeasurementSpec,
> ```

nit: I think this only uses the measurement type (measurementTypeCase) from the MeasurementSpec, so just pass that



_Code quote:_
```Kotlin
parsedMeasurementSpec: MeasurementSpec,
```
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/ProtoConversions.kt` line 325 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/728#-NFyZAZ50hq6XaRtQKcU:-NFyZAZ50hq6XaRtQKcV:b-q1k41n) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5fca28083d94c8b3c8fe6347f6018cc1a0130e5d/src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/ProtoConversions.kt#L325)):*
> ```Kotlin
>           }
>         }
>         // No protocol for impression or duration type.
> ```

nit: not strictly true. It's just that we don't need to persist any info for the Direct protocol.



_Code quote:_
```Kotlin
No protocol
```
___
*[`src/main/proto/wfa/measurement/internal/kingdom/protocol_config.proto` line 33 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/728#-NFoe8lx0Ot_pGhetHlZ:-NFyTrOH8prYm04B8cBs:b-tve6lh) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d5ee7a410c5859f1b172ea8106c8bffd9b1a0757/src/main/proto/wfa/measurement/internal/kingdom/protocol_config.proto#L33)):*

what is the use case for keeping MeasurementType at all here?
</blockquote></details>

+1 to Steven's question. I had suggested dropping it in the quoted comment. This data is redundant as it's already in MeasurementSpec. It may have made sense to denormalize in the past so we could avoid deserializing the MeasurementSpec, but now I believe we always need to anyway.

Dropping a field is a backwards-compatible change. Just make sure you add a `reserved` line so the field number doesn't get reused. See https://developers.google.com/protocol-buffers/docs/proto3#reserved

iverson52000 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/728)* status: 17 of 18 files reviewed, 4 unresolved discussions (waiting on @rmirisola, @robinsons, @SanjayVas, @stevenwarejones, and @xnegron)

___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/readers/RequisitionReader.kt` line 58 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/728#-NFyXDTD09WWKMKUHm_9:-NFzDn-k5najDDWYXyT8:bpnsjh2) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5fca28083d94c8b3c8fe6347f6018cc1a0130e5d/src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/readers/RequisitionReader.kt#L58)):*

I don't think you need this. There is exactly one Requisition for each DataProvider in a given Measurement, meaning DataProvidersCount is the number of Requisitions.
</blockquote></details>

Thanks do you mean this subquery is not needed at all? How do I get the number of Requisitions then?

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/728)* status: 17 of 18 files reviewed, 4 unresolved discussions (waiting on @iverson52000, @rmirisola, @robinsons, @stevenwarejones, and @xnegron)

___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/readers/RequisitionReader.kt` line 58 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/728#-NFyXDTD09WWKMKUHm_9:-NFzEQEr6T_S4DmoSAZx:b-kthi6f) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5fca28083d94c8b3c8fe6347f6018cc1a0130e5d/src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/readers/RequisitionReader.kt#L58)):*

Thanks do you mean this subquery is not needed at all? How do I get the number of Requisitions then?
</blockquote></details>

I mean you can simplify this, as you just need to count the number of requisitions for a given Measurement. No need to join with DataProviders/DataProviderCertificates.

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/728)* status: 17 of 18 files reviewed, 4 unresolved discussions (waiting on @iverson52000, @rmirisola, @robinsons, @stevenwarejones, and @xnegron)

___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/readers/RequisitionReader.kt` line 58 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/728#-NFyXDTD09WWKMKUHm_9:-NFzFJZM6q7DET38uktK:b-e794ks) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5fca28083d94c8b3c8fe6347f6018cc1a0130e5d/src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/readers/RequisitionReader.kt#L58)):*

I mean you can simplify this, as you just need to count the number of requisitions for a given Measurement. No need to join with DataProviders/DataProviderCertificates.
</blockquote></details>

I'd call this something like MeasurementRequisitionCount rather than DataProvidersCount for this query, as number of data providers for a single requisition is always 1.

iverson52000 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/728)* status: 17 of 18 files reviewed, 4 unresolved discussions (waiting on @rmirisola, @robinsons, @SanjayVas, @stevenwarejones, and @xnegron)

___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/readers/RequisitionReader.kt` line 58 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/728#-NFyXDTD09WWKMKUHm_9:-NFzH-AB6z7d6EL_GlBR:b-cna4qw) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5fca28083d94c8b3c8fe6347f6018cc1a0130e5d/src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/readers/RequisitionReader.kt#L58)):*

I'd call this something like MeasurementRequisitionCount rather than DataProvidersCount for this query, as number of data providers for a single requisition is always 1.
</blockquote></details>

Got it. Thanks for clarifying

iverson52000:
*[`src/main/proto/wfa/measurement/internal/kingdom/protocol_config.proto` line 33 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/728#-NFoe8lx0Ot_pGhetHlZ:-NFzKiqM2Jsjxy2ZAgzB:b-b74lrw) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d5ee7a410c5859f1b172ea8106c8bffd9b1a0757/src/main/proto/wfa/measurement/internal/kingdom/protocol_config.proto#L33)):*

+1 to Steven's question. I had suggested dropping it in the quoted comment. This data is redundant as it's already in MeasurementSpec. It may have made sense to denormalize in the past so we could avoid deserializing the MeasurementSpec, but now I believe we always need to anyway.

Dropping a field is a backwards-compatible change. Just make sure you add a `reserved` line so the field number doesn't get reused. See https://developers.google.com/protocol-buffers/docs/proto3#reserved
</blockquote></details>

So here I will just remove `enum MeasurementType{}` and then reserve field number = 2? (`MeasurementType measurement_type = 2;`)

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/728)* status: 17 of 18 files reviewed, 4 unresolved discussions (waiting on @iverson52000, @rmirisola, @robinsons, @stevenwarejones, and @xnegron)

___
*[`src/main/proto/wfa/measurement/internal/kingdom/protocol_config.proto` line 33 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/728#-NFoe8lx0Ot_pGhetHlZ:-NFzMlOF3pvTimCw5ZnE:b-agqd1x) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d5ee7a410c5859f1b172ea8106c8bffd9b1a0757/src/main/proto/wfa/measurement/internal/kingdom/protocol_config.proto#L33)):*

So here I will just remove `enum MeasurementType{}` and then reserve field number = 2? (`MeasurementType measurement_type = 2;`)
</blockquote></details>

Yep, remove the MeasurementType enum as well as the corresponding field and mark that field number as reserved.

iverson52000:
*[`src/main/proto/wfa/measurement/internal/kingdom/protocol_config.proto` line 33 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/728#-NFoe8lx0Ot_pGhetHlZ:-NG2q6ad1gAU7wVhQzTc:b-n9u7ix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d5ee7a410c5859f1b172ea8106c8bffd9b1a0757/src/main/proto/wfa/measurement/internal/kingdom/protocol_config.proto#L33)):*

Yep, remove the MeasurementType enum as well as the corresponding field and mark that field number as reserved.
</blockquote></details>

Done. Thanks Steven and Sanjay for the advices

iverson52000:
*[`src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/readers/RequisitionReader.kt` line 58 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/728#-NFyXDTD09WWKMKUHm_9:-NG2qPHTDaABu2xNkfYd:bb6un1j) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5fca28083d94c8b3c8fe6347f6018cc1a0130e5d/src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/readers/RequisitionReader.kt#L58)):*

Got it. Thanks for clarifying
</blockquote></details>

Done

SanjayVas (changes requested):
> Reviewed 3 of 18 files at r4, 14 of 14 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/728)* status: all files reviewed, 4 unresolved discussions (waiting on @iverson52000, @rmirisola, @robinsons, @stevenwarejones, and @xnegron)

___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/ProtoConversions.kt` line 136 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/728#-NG32sWLD_UcVmLRKMHW:-NG32sWLD_UcVmLRKMHX:bfoyc0a) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/23a0babd414ff6982991379a82c9188257ecbf48/src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/ProtoConversions.kt#L136)):*
> ```Kotlin
> /** @throws [IllegalStateException] if MeasurementType not specified */
> fun InternalProtocolConfig.toProtocolConfig(
>   measurementTypeCase: MeasurementSpec.MeasurementTypeCase,
> ```

nit: just `measurementType` is fine. The "case" part is an artifact of protobuf codegen.



_Code quote:_
```Kotlin
measurementTypeCase
```
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt` line 210 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/728#-NG339BYDJTV4Etw8199:-NG339BZ5THSwFk7g0n1:bwnf8dk) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/23a0babd414ff6982991379a82c9188257ecbf48/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt#L210)):*
> ```Kotlin
> 
>       if (
>         !requisition.hasProtocolConfig() ||
> ```

Won't there always be a protocol config now?



_Code quote:_
```Kotlin
!requisition.hasProtocolConfig() ||
```
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt` line 357 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/728#-NG33KXF5DEM8GPDsbQj:-NG33KXF5DEM8GPDsbQk:bfsc57l) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/23a0babd414ff6982991379a82c9188257ecbf48/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt#L357)):*
> ```Kotlin
>     val liquidLegionsV2: ProtocolConfig.LiquidLegionsV2 =
>       requisition.protocolConfig.protocolsList
>         .find { protocol -> protocol.hasLiquidLegionsV2() }!!
> ```

Only use this coercion when you can guarantee it, but the compiler can't figure it out. This is not guaranteed from a code perspective, just a data one. Therefore you should use requireNotNull instead.



_Code quote:_
```Kotlin
!!
```

stevenwarejones (approved):
> Reviewed 14 of 14 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/728)* status: all files reviewed, 3 unresolved discussions (waiting on @iverson52000, @rmirisola, @robinsons, and @xnegron)

iverson52000 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/728)* status: 17 of 18 files reviewed, 3 unresolved discussions (waiting on @rmirisola, @robinsons, @SanjayVas, @stevenwarejones, and @xnegron)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt` line 210 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/728#-NG339BYDJTV4Etw8199:-NG3DVyiAclONMIVIjaO:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/23a0babd414ff6982991379a82c9188257ecbf48/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt#L210)):*

Won't there always be a protocol config now?
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt` line 357 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/728#-NG33KXF5DEM8GPDsbQj:-NG3DYek1PVvVmkVfuIq:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/23a0babd414ff6982991379a82c9188257ecbf48/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt#L357)):*

Only use this coercion when you can guarantee it, but the compiler can't figure it out. This is not guaranteed from a code perspective, just a data one. Therefore you should use requireNotNull instead.
</blockquote></details>

Done.

SanjayVas (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/728)* status: all files reviewed, 1 unresolved discussion (waiting on @iverson52000, @rmirisola, @robinsons, and @xnegron)

iverson52000 (commented):
> Reviewed 4 of 18 files at r4, 13 of 14 files at r6, 1 of 1 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/728)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @rmirisola, @robinsons, and @xnegron)

PR #560, opened 2022-05-10
Use kustomization rules for secret generation.

SanjayVas:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/560)

PR #733, opened 2022-10-19
Add Bazel test target for CMMS KinD correctness test.

SanjayVas:
The test is implemented using JUnit and assumes that KinD is installed and in the test PATH. It uses the Kubernetes API to interact with the cluster after creation. The test methods are run locally using port forwarding rather than deploying the test harness (FrontendSimular) to the cluster. Resource setup is also run locally.

This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/733)

https://rally1.rallydev.com/#/?detail=/task/668307013823&fdp=true


SanjayVas:
If you're using [hybrid development](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/main/docs/building.md#hybrid-development), you can build the test within a container and then run it locally using the following command:

```shell
tools/bazel-container-run //src/test/kotlin/org/wfanet/measurement/integration/k8s:CorrectnessTest
```

SanjayVas:
Depends on #560

SanjayVas:
Manual run passed on GitHub Actions: https://github.com/world-federation-of-advertisers/cross-media-measurement/actions/runs/3363026638

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/733)* status: 0 of 34 files reviewed, all discussions resolved (waiting on @matthewclegg)

___
*[`src/main/k8s/local/README.md` line 25 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/733#-NFyinKB2JQGinY4aFUs:-NFyinKB2JQGinY4aFUt:blcvm7m) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4f2a0ea6ae33c238e20c6616b1bc27c373d242c4/src/main/k8s/local/README.md#L25)):*
> ```Markdown
> ## Initial Setup
> 
> ### Set Default Resource Requirements
> ```

This was dropped due to resource limits not being respected in KinD. See https://github.com/kubernetes-sigs/kind/issues/2514#issuecomment-951265965

stevenwarejones (commented):
> Reviewed 2 of 5 files at r1, 26 of 30 files at r2, 4 of 4 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/733)* status: 32 of 34 files reviewed, 4 unresolved discussions (waiting on @matthewclegg and @SanjayVas)

___
*[`src/main/k8s/local/README.md` line 25 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/733#-NFyinKB2JQGinY4aFUs:-NFzOPFI7OyjSiFEcLlv:b-28agm2) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4f2a0ea6ae33c238e20c6616b1bc27c373d242c4/src/main/k8s/local/README.md#L25)):*
> kubernetes-sigs/kind#2514 (comment)
can you add a comment about this? That its not respected?
___
*[`src/main/kotlin/org/wfanet/measurement/common/k8s/JsonPatchOperation.kt` line 20 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/733#-NFzOcADA7Xit46yiSDE:-NFzOcADA7Xit46yiSDF:bb7jbea) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/11dc5c01bc8f56f4182d86219979efff1e6645ff/src/main/kotlin/org/wfanet/measurement/common/k8s/JsonPatchOperation.kt#L20)):*
> ```Kotlin
> 
> internal sealed class JsonPatchOperation
> private constructor(
> ```

kdoc
___
*[`src/main/kotlin/org/wfanet/measurement/common/k8s/testing/KindCluster.kt` line 37 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/733#-NFzPPAqEbTVOqyq6XCQ:-NFzPPAqEbTVOqyq6XCR:b-d80gty) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/11dc5c01bc8f56f4182d86219979efff1e6645ff/src/main/kotlin/org/wfanet/measurement/common/k8s/testing/KindCluster.kt#L37)):*
> ```Kotlin
>  */
> class KindCluster : TestRule {
>   lateinit var uuid: UUID
> ```

why late init and not just make it `by lazy`?



_Code quote:_
```Kotlin
uuid
```
___
*[`src/main/kotlin/org/wfanet/measurement/common/k8s/testing/KindCluster.kt` line 74 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/733#-NFzSOH84m7Gf4Dv6sfU:-NFzSOH98dgyiVXKxsDt:b9aizch) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/11dc5c01bc8f56f4182d86219979efff1e6645ff/src/main/kotlin/org/wfanet/measurement/common/k8s/testing/KindCluster.kt#L74)):*
> ```Kotlin
> 
> @Blocking
> private fun runCommand(vararg command: String) {
> ```

where are we logging these commands?
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/frontend/FrontendSimulator.kt` line 404 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/733#-NFzSeSNEhsS8-MlsCCR:-NFzSeSNEhsS8-MlsCCS:b49clxl) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/11dc5c01bc8f56f4182d86219979efff1e6645ff/src/main/kotlin/org/wfanet/measurement/loadtest/frontend/FrontendSimulator.kt#L404)):*
> ```Kotlin
>     val resultPair = measurement.resultsList[0]
>     val result = parseAndVerifyResult(resultPair)
>     assertThat(result.hasReach()).isTrue()
> ```

are there some sanity check numbers we can verify here?

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/733)* status: 31 of 34 files reviewed, 4 unresolved discussions (waiting on @matthewclegg and @stevenwarejones)

___
*[`src/main/k8s/local/README.md` line 25 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/733#-NFyinKB2JQGinY4aFUs:-NFzVFECErnYAH4lDRrH:b-gd35br) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4f2a0ea6ae33c238e20c6616b1bc27c373d242c4/src/main/k8s/local/README.md#L25)):*

> kubernetes-sigs/kind#2514 (comment)
can you add a comment about this? That its not respected?
</blockquote></details>

Do you have any recommendations on a good place to put such a comment? The main reason I mentioned it in code review instead is because I couldn't find such a place, as we're effectively left with no natural place where such limits would be specified.
___
*[`src/main/kotlin/org/wfanet/measurement/common/k8s/JsonPatchOperation.kt` line 20 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/733#-NFzOcADA7Xit46yiSDE:-NFz_YqzEdorVWh-Dbxa:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/11dc5c01bc8f56f4182d86219979efff1e6645ff/src/main/kotlin/org/wfanet/measurement/common/k8s/JsonPatchOperation.kt#L20)):*

kdoc
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/common/k8s/testing/KindCluster.kt` line 37 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/733#-NFzPPAqEbTVOqyq6XCQ:-NFzWX6CA0vHEAB6ftJP:b2bqs5r) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/11dc5c01bc8f56f4182d86219979efff1e6645ff/src/main/kotlin/org/wfanet/measurement/common/k8s/testing/KindCluster.kt#L37)):*

why late init and not just make it `by lazy`?
</blockquote></details>

So that it's guaranteed not to be set until `evaluate` is called. It should throw an error if someone tries to access it before then, which is exactly the lateinit behavior.
___
*[`src/main/kotlin/org/wfanet/measurement/common/k8s/testing/KindCluster.kt` line 74 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/733#-NFzSOH84m7Gf4Dv6sfU:-NFzWnWW3pQfT0DnalRM:bq9onr3) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/11dc5c01bc8f56f4182d86219979efff1e6645ff/src/main/kotlin/org/wfanet/measurement/common/k8s/testing/KindCluster.kt#L74)):*

where are we logging these commands?
</blockquote></details>

The commands themselves are never logged, but we log before/after them depending on whether the command itself has useful output.
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/frontend/FrontendSimulator.kt` line 404 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/733#-NFzSeSNEhsS8-MlsCCR:-NFzXcN3C2bDo-q-Fhtn:bss6jsg) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/11dc5c01bc8f56f4182d86219979efff1e6645ff/src/main/kotlin/org/wfanet/measurement/loadtest/frontend/FrontendSimulator.kt#L404)):*

are there some sanity check numbers we can verify here?
</blockquote></details>

The number verification already happens after this call. No need to do anything here.

stevenwarejones (commented):
> Reviewed 1 of 1 files at r4.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/733)* status: 32 of 34 files reviewed, 2 unresolved discussions (waiting on @matthewclegg and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/common/k8s/testing/KindCluster.kt` line 37 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/733#-NFzPPAqEbTVOqyq6XCQ:-NFzgTPL5jVcSc7GbhQV:b32v57) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/11dc5c01bc8f56f4182d86219979efff1e6645ff/src/main/kotlin/org/wfanet/measurement/common/k8s/testing/KindCluster.kt#L37)):*

So that it's guaranteed not to be set until `evaluate` is called. It should throw an error if someone tries to access it before then, which is exactly the lateinit behavior.
</blockquote></details>

+1
___
*[`src/main/kotlin/org/wfanet/measurement/common/k8s/testing/KindCluster.kt` line 74 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/733#-NFzSOH84m7Gf4Dv6sfU:-NFzgNvGDWST2YLa2bwd:b75rhg5) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/11dc5c01bc8f56f4182d86219979efff1e6645ff/src/main/kotlin/org/wfanet/measurement/common/k8s/testing/KindCluster.kt#L74)):*

The commands themselves are never logged, but we log before/after them depending on whether the command itself has useful output.
</blockquote></details>

it would probably be good to log them

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/733)* status: 32 of 34 files reviewed, 1 unresolved discussion (waiting on @matthewclegg and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/common/k8s/testing/KindCluster.kt` line 74 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/733#-NFzSOH84m7Gf4Dv6sfU:-NFzsLwiEGdCAjTzaM5y:b-df4alb) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/11dc5c01bc8f56f4182d86219979efff1e6645ff/src/main/kotlin/org/wfanet/measurement/common/k8s/testing/KindCluster.kt#L74)):*

it would probably be good to log them
</blockquote></details>

Sorry, should clarify that they do also get logged on error. No need to log on success.

stevenwarejones (changes requested):
> Reviewed 1 of 5 files at r1, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/733)* status: all files reviewed, 8 unresolved discussions (waiting on @matthewclegg and @SanjayVas)

___
*[`src/main/k8s/local/README.md` line 25 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/733#-NFyinKB2JQGinY4aFUs:-NG2ziZ8BsEZxxq0i1Rp:bll2chb) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4f2a0ea6ae33c238e20c6616b1bc27c373d242c4/src/main/k8s/local/README.md#L25)):*

Do you have any recommendations on a good place to put such a comment? The main reason I mentioned it in code review instead is because I couldn't find such a place, as we're effectively left with no natural place where such limits would be specified.
</blockquote></details>

i think it belongs in this doc as a warning, perhaps?
___
*[`src/main/kotlin/org/wfanet/measurement/common/k8s/KubernetesClient.kt` line 72 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/733#-NG2zp0F2mL58fEitdN7:-NG2zp0F2mL58fEitdN8:b25exlo) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3f607ed7063c091b6deea81db3d0a58a1b5617f8/src/main/kotlin/org/wfanet/measurement/common/k8s/KubernetesClient.kt#L72)):*
> ```Kotlin
>   suspend fun restartDeployment(
>     name: String,
>     namespace: String = Namespaces.NAMESPACE_DEFAULT
> ```

would prefer to not have a default
___
*[`src/main/kotlin/org/wfanet/measurement/common/k8s/KubernetesClient.kt` line 95 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/733#-NG2zr_o29dlONCoRv9_:-NG2zr_o29dlONCoRv9a:b-7ropln) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3f607ed7063c091b6deea81db3d0a58a1b5617f8/src/main/kotlin/org/wfanet/measurement/common/k8s/KubernetesClient.kt#L95)):*
> ```Kotlin
>     key: String,
>     value: String,
>     namespace: String = Namespaces.NAMESPACE_DEFAULT
> ```

ditto
___
*[`src/main/kotlin/org/wfanet/measurement/common/k8s/testing/KindCluster.kt` line 74 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/733#-NFzSOH84m7Gf4Dv6sfU:-NG301KK1yrksvHHtty_:bw09iso) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/11dc5c01bc8f56f4182d86219979efff1e6645ff/src/main/kotlin/org/wfanet/measurement/common/k8s/testing/KindCluster.kt#L74)):*

Sorry, should clarify that they do also get logged on error. No need to log on success.
</blockquote></details>

would be good to add fine logging for success to
___
*[`src/test/kotlin/org/wfanet/measurement/integration/k8s/CorrectnessTest.kt` line 112 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/733#-NG31gPk9qHfoNo6G0df:-NG31gPm8mBD2ihUJ57E:b-uvp8wk) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3f607ed7063c091b6deea81db3d0a58a1b5617f8/src/test/kotlin/org/wfanet/measurement/integration/k8s/CorrectnessTest.kt#L112)):*
> ```Kotlin
>     companion object {
>       fun from(resources: Iterable<Resources.Resource>): ResourceInfo {
>         var aggregatorCert: String? = null
> ```

could you make these `lateinit val` instead to ensure they are only updated 1x? A misconfiguration could accidentally add two aggregator certs, right, and we'd want that to throw an error?
___
*[`src/test/kotlin/org/wfanet/measurement/integration/k8s/CorrectnessTest.kt` line 127 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/733#-NG31rNc5oDb-HOZLU6g:-NG31rNd7aK1DcFP8Pri:b-7lfe9i) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3f607ed7063c091b6deea81db3d0a58a1b5617f8/src/test/kotlin/org/wfanet/measurement/integration/k8s/CorrectnessTest.kt#L127)):*
> ```Kotlin
>             }
>             Resources.Resource.ResourceCase.DATA_PROVIDER -> {
>               dataProviders[resource.dataProvider.displayName] = resource.name
> ```

require that the key doesn't already exist in the map
___
*[`src/test/kotlin/org/wfanet/measurement/integration/k8s/CorrectnessTest.kt` line 134 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/733#-NG30PCY3BSXC1fspZ9f:-NG30PCY3BSXC1fspZ9g:b-st5l2j) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3f607ed7063c091b6deea81db3d0a58a1b5617f8/src/test/kotlin/org/wfanet/measurement/integration/k8s/CorrectnessTest.kt#L134)):*
> ```Kotlin
>                 "worker1" -> worker1Cert = resource.name
>                 "worker2" -> worker2Cert = resource.name
>               }
> ```

shouldn't this list be exhaustive with an else that throws?
___
*[`src/test/kotlin/org/wfanet/measurement/integration/k8s/CorrectnessTest.kt` line 159 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/733#-NG321UW21Q_nnR9tRfA:-NG321UW21Q_nnR9tRfB:b-o52qwf) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3f607ed7063c091b6deea81db3d0a58a1b5617f8/src/test/kotlin/org/wfanet/measurement/integration/k8s/CorrectnessTest.kt#L159)):*
> ```Kotlin
>     }
> 
> ```

I would get rid of the explicit type, just
```
private val logger = Logger.getLogger(this::class.java.name)
```

___
*[`src/test/kotlin/org/wfanet/measurement/integration/k8s/CorrectnessTest.kt` line 539 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/733#-NG32YVFEEIFkEWV07aU:-NG32YVG4Is1jvwnvhPN:b-4gzbst) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3f607ed7063c091b6deea81db3d0a58a1b5617f8/src/test/kotlin/org/wfanet/measurement/integration/k8s/CorrectnessTest.kt#L539)):*
> ```Kotlin
>             .listEventGroups(
>               listEventGroupsRequest {
>                 parent = "dataProviders/-"
> ```

why is it just set to `"dataProviders/-"`

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/733)* status: all files reviewed, 8 unresolved discussions (waiting on @matthewclegg and @stevenwarejones)

___
*[`src/main/k8s/local/README.md` line 25 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/733#-NFyinKB2JQGinY4aFUs:-NG3y56W6vyfNzA4hj8P:b-t8znvr) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4f2a0ea6ae33c238e20c6616b1bc27c373d242c4/src/main/k8s/local/README.md#L25)):*

i think it belongs in this doc as a warning, perhaps?
</blockquote></details>

Rephrased the title of the doc and restructured it a bit to find a place to fit this. It didn't fit in the existing doc.
___
*[`src/main/kotlin/org/wfanet/measurement/common/k8s/KubernetesClient.kt` line 72 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/733#-NG2zp0F2mL58fEitdN7:-NG41WsOAfZMq3pxqcW0:bwyywn5) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3f607ed7063c091b6deea81db3d0a58a1b5617f8/src/main/kotlin/org/wfanet/measurement/common/k8s/KubernetesClient.kt#L72)):*

would prefer to not have a default
</blockquote></details>

This default is consistent with the `kubectl` CLI. It's in fact rare to operate on a different namespace. Namespaces are used to organize/hide things like K8s system pods by default.

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/733)* status: all files reviewed, 6 unresolved discussions (waiting on @matthewclegg and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/common/k8s/KubernetesClient.kt` line 72 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/733#-NG2zp0F2mL58fEitdN7:-NGHQi1TC2V5Cdc1DLXy:bay358r) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3f607ed7063c091b6deea81db3d0a58a1b5617f8/src/main/kotlin/org/wfanet/measurement/common/k8s/KubernetesClient.kt#L72)):*

This default is consistent with the `kubectl` CLI. It's in fact rare to operate on a different namespace. Namespaces are used to organize/hide things like K8s system pods by default.
</blockquote></details>

i understand but prefer this be more explicit but I'm okay with it the way it is too.

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/733)* status: 31 of 34 files reviewed, 6 unresolved discussions (waiting on @matthewclegg and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/common/k8s/testing/KindCluster.kt` line 74 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/733#-NFzSOH84m7Gf4Dv6sfU:-NGOu8Gn4TxLxX_ZkuDS:b-5flu7x) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/11dc5c01bc8f56f4182d86219979efff1e6645ff/src/main/kotlin/org/wfanet/measurement/common/k8s/testing/KindCluster.kt#L74)):*

would be good to add fine logging for success to
</blockquote></details>

Done. (Moved into companion object to get access to a logger).
___
*[`src/test/kotlin/org/wfanet/measurement/integration/k8s/CorrectnessTest.kt` line 112 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/733#-NG31gPk9qHfoNo6G0df:-NGOhL9X42pFS3XcxON0:byd6ooa) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3f607ed7063c091b6deea81db3d0a58a1b5617f8/src/test/kotlin/org/wfanet/measurement/integration/k8s/CorrectnessTest.kt#L112)):*

could you make these `lateinit val` instead to ensure they are only updated 1x? A misconfiguration could accidentally add two aggregator certs, right, and we'd want that to throw an error?
</blockquote></details>

`lateinit var` wouldn't error if it's set more than once (there's no such thing as `lateinit val`). It just gives an initialization error if it's accessed before it's ever set, which is already what the `requireNotNull` call does later.

We also don't care if these are set more than once. ResourceSetup can create more resources than we use. We just need there to be at least one of each for the purposes of this test.
___
*[`src/test/kotlin/org/wfanet/measurement/integration/k8s/CorrectnessTest.kt` line 127 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/733#-NG31rNc5oDb-HOZLU6g:-NGOsluP8GzMpH26H5jm:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3f607ed7063c091b6deea81db3d0a58a1b5617f8/src/test/kotlin/org/wfanet/measurement/integration/k8s/CorrectnessTest.kt#L127)):*

require that the key doesn't already exist in the map
</blockquote></details>

Done.
___
*[`src/test/kotlin/org/wfanet/measurement/integration/k8s/CorrectnessTest.kt` line 134 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/733#-NG30PCY3BSXC1fspZ9f:-NGOsmAX09y07vii69Ab:bon3umt) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3f607ed7063c091b6deea81db3d0a58a1b5617f8/src/test/kotlin/org/wfanet/measurement/integration/k8s/CorrectnessTest.kt#L134)):*

shouldn't this list be exhaustive with an else that throws?
</blockquote></details>

We don't really care if ResourceSetup makes more Duchy certificates. All we care about is that it created at least one certificate for each of these duchies.
___
*[`src/test/kotlin/org/wfanet/measurement/integration/k8s/CorrectnessTest.kt` line 159 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/733#-NG321UW21Q_nnR9tRfA:-NGOtU-F2FbDAm1MIc4e:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3f607ed7063c091b6deea81db3d0a58a1b5617f8/src/test/kotlin/org/wfanet/measurement/integration/k8s/CorrectnessTest.kt#L159)):*

I would get rid of the explicit type, just
```
private val logger = Logger.getLogger(this::class.java.name)
```

</blockquote></details>

Done.
___
*[`src/test/kotlin/org/wfanet/measurement/integration/k8s/CorrectnessTest.kt` line 539 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/733#-NG32YVFEEIFkEWV07aU:-NGOp_XoAT72WBt6LUvR:b-lze3t8) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3f607ed7063c091b6deea81db3d0a58a1b5617f8/src/test/kotlin/org/wfanet/measurement/integration/k8s/CorrectnessTest.kt#L539)):*

why is it just set to `"dataProviders/-"`
</blockquote></details>

That's the way to specify a wildcard for data providers. This is listing all event groups across all data providers for a single MC. (The parent of EventGroup is DataProvider, but it has a reference to the MC).

stevenwarejones (changes requested):
> Reviewed 3 of 3 files at r7, 4 of 4 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/733)* status: all files reviewed, 1 unresolved discussion (waiting on @matthewclegg and @SanjayVas)

___
*[`src/test/kotlin/org/wfanet/measurement/integration/k8s/CorrectnessTest.kt` line 134 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/733#-NG30PCY3BSXC1fspZ9f:-NGT6zV-EsZ7juNlT7gG:bsp7rav) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3f607ed7063c091b6deea81db3d0a58a1b5617f8/src/test/kotlin/org/wfanet/measurement/integration/k8s/CorrectnessTest.kt#L134)):*

We don't really care if ResourceSetup makes more Duchy certificates. All we care about is that it created at least one certificate for each of these duchies.
</blockquote></details>

what is the use case for additional duchy ids getting passed in where we don't want to throw an error?

matthewclegg (approved):
> Reviewed 2 of 5 files at r1, 20 of 30 files at r2, 4 of 4 files at r3, 1 of 1 files at r4, 1 of 1 files at r5, 3 of 3 files at r7, 4 of 4 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/733)* status: all files reviewed, 4 unresolved discussions (waiting on @SanjayVas and @stevenwarejones)

___
*[`src/main/k8s/base.cue` line 82 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/733#-NGTuetU6kDRYerSwt_w:-NGTuetU6kDRYerSwt_x:b-rbnqxi) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a964a243809bfda1c66e47bb54fc0dc4fc70cb5f/src/main/k8s/base.cue#L82)):*
> ```CUE
> #JavaOptions: {
> 	maxRamPercentage?:      float
> 	initialRamPercentage?:  float
> ```

Am I correct in understanding that these changes are independent of the code for supporting the CMMS correctness test?  If so, you might want to consider moving them to a separate PR.  This PR is already a little large.
___
*[`src/test/kotlin/org/wfanet/measurement/integration/k8s/CorrectnessTest.kt` line 134 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/733#-NG30PCY3BSXC1fspZ9f:-NGTm1AUD1pFtGZwVbdf:bdx4yb6) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3f607ed7063c091b6deea81db3d0a58a1b5617f8/src/test/kotlin/org/wfanet/measurement/integration/k8s/CorrectnessTest.kt#L134)):*

No current known use case, but that doesn't mean it needs to be an error. ResourceSetup can be somewhat decoupled from this particular test, so long as it creates at least the resources that this test needs. Other tests can also theoretically use ResourceSetup unchanged. It's an independent decision that doesn't need to be made now as to whether we'd want tests with slightly different needs to use a different version/configuration of ResourceSetup.
</blockquote></details>

To my knowledge, we have never tested the CMMS in a configuration other than 2 workers + 1 aggregator.  Would it be worth considering testing other configurations, such as 1 worker + 1 aggregator or 3 workers + 1 aggregator?
___
*[`src/test/kotlin/org/wfanet/measurement/integration/k8s/CorrectnessTest.kt` line 100 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/733#-NGTlKpYEH3WDSC5fExp:-NGTlKpYEH3WDSC5fExq:b-y950mo) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a964a243809bfda1c66e47bb54fc0dc4fc70cb5f/src/test/kotlin/org/wfanet/measurement/integration/k8s/CorrectnessTest.kt#L100)):*
> ```Kotlin
>   fun `reach and frequency measurement completes with expected result`() = runBlocking {
>     testHarness.executeReachAndFrequency("$runId-reach-and-freq")
>   }
> ```

Would it make sense to have separate tests for direct measurements vs MPC-based measurements?
What about a test for reach-only measurement?
___
*[`src/test/kotlin/org/wfanet/measurement/integration/k8s/CorrectnessTest.kt` line 170 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/733#-NGTmUohA1VaKaZwZH_w:-NGTmUohA1VaKaZwZH_x:b-tf5ers) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a964a243809bfda1c66e47bb54fc0dc4fc70cb5f/src/test/kotlin/org/wfanet/measurement/integration/k8s/CorrectnessTest.kt#L170)):*
> ```Kotlin
>     private const val FORWARDED_STORAGE_DEPLOYMENT_NAME = "fake-storage-server-deployment"
>     private val OUTPUT_DP_PARAMS = differentialPrivacyParams {
>       epsilon = 0.1
> ```

You might want to consider using DP params that are close to those that we intend to use in production.  We know that the behavior of the system is influenced to a certain extent by the choice of DP params.

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/733)* status: all files reviewed, 4 unresolved discussions (waiting on @matthewclegg and @stevenwarejones)

___
*[`src/main/k8s/base.cue` line 82 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/733#-NGTuetU6kDRYerSwt_w:-NGTyYQCAo-A0i4VqFhB:b-h1oa4p) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a964a243809bfda1c66e47bb54fc0dc4fc70cb5f/src/main/k8s/base.cue#L82)):*

Am I correct in understanding that these changes are independent of the code for supporting the CMMS correctness test?  If so, you might want to consider moving them to a separate PR.  This PR is already a little large.
</blockquote></details>

They are not independent, but I can extract them to a prefactoring PR that this PR would depend on if desired. These changes are necessary for the automated correctness test to pass on GitHub Actions.

More specifically, I needed to tighten the memory used in order to fit in the GitHub Actions runner. I also had to switch away from using RAM percentage options for KinD since it turns out KinD doesn't properly implement memory limits to the container. I have a follow-up PR that does the same for our dev GKE configuration, as it turns out the heap percentage limits with container support were a bit half-baked: They don't allow independent control of heap vs. direct and all of the other memory parameters use fixed sizes
___
*[`src/test/kotlin/org/wfanet/measurement/integration/k8s/CorrectnessTest.kt` line 134 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/733#-NG30PCY3BSXC1fspZ9f:-NGU-6dlChEsemI-DFz9:bblnzee) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3f607ed7063c091b6deea81db3d0a58a1b5617f8/src/test/kotlin/org/wfanet/measurement/integration/k8s/CorrectnessTest.kt#L134)):*

To my knowledge, we have never tested the CMMS in a configuration other than 2 workers + 1 aggregator.  Would it be worth considering testing other configurations, such as 1 worker + 1 aggregator or 3 workers + 1 aggregator?
</blockquote></details>

I have no opinion on testing other configurations. This PR is just trying to make an automated version of the existing test. We can certainly consider prioritizing adding more tests and test configurations in the future.

Note that this is orthogonal to the point of having this test not care if ResourceSetup creates extra resources that the test doesn't use.
___
*[`src/test/kotlin/org/wfanet/measurement/integration/k8s/CorrectnessTest.kt` line 100 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/733#-NGTlKpYEH3WDSC5fExp:-NGTzMVUDEpaQlDCAqOV:b-vzjynj) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a964a243809bfda1c66e47bb54fc0dc4fc70cb5f/src/test/kotlin/org/wfanet/measurement/integration/k8s/CorrectnessTest.kt#L100)):*

Would it make sense to have separate tests for direct measurements vs MPC-based measurements?
What about a test for reach-only measurement?

</blockquote></details>

This file just runs the exact same tests as are run when using the manual process. See [`FrontendSimulatorRunner`](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/main/src/main/kotlin/org/wfanet/measurement/loadtest/frontend/FrontendSimulatorRunner.kt).

We can certainly add more tests in the future, but I assume we'd want to do so in both places. Either that or a future refactor to share more of the same codebase for testing.
___
*[`src/test/kotlin/org/wfanet/measurement/integration/k8s/CorrectnessTest.kt` line 170 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/733#-NGTmUohA1VaKaZwZH_w:-NGU-XjKCqivBE_ZER12:b-5p2e4s) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a964a243809bfda1c66e47bb54fc0dc4fc70cb5f/src/test/kotlin/org/wfanet/measurement/integration/k8s/CorrectnessTest.kt#L170)):*

You might want to consider using DP params that are close to those that we intend to use in production.  We know that the behavior of the system is influenced to a certain extent by the choice of DP params.
</blockquote></details>

This is copying exactly what's in the manual version of the test. We can adjust the params in both tests in sync in a future PR.

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/733)* status: all files reviewed, 4 unresolved discussions (waiting on @matthewclegg)

___
*[`src/test/kotlin/org/wfanet/measurement/integration/k8s/CorrectnessTest.kt` line 134 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/733#-NG30PCY3BSXC1fspZ9f:-NGqvyKD4xu9JHVJrFVo:b-96x27n) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3f607ed7063c091b6deea81db3d0a58a1b5617f8/src/test/kotlin/org/wfanet/measurement/integration/k8s/CorrectnessTest.kt#L134)):*

I have no opinion on testing other configurations. This PR is just trying to make an automated version of the existing test. We can certainly consider prioritizing adding more tests and test configurations in the future.

Note that this is orthogonal to the point of having this test not care if ResourceSetup creates extra resources that the test doesn't use.
</blockquote></details>

i'd prefer to throw an error if an unknown duchy id is encountered.



_Code quote:_
```Kotlin
when (resource.duchyCertificate.duchyId) {
```

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/733)* status: 34 of 35 files reviewed, 4 unresolved discussions (waiting on @matthewclegg and @stevenwarejones)

___
*[`src/test/kotlin/org/wfanet/measurement/integration/k8s/CorrectnessTest.kt` line 134 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/733#-NG30PCY3BSXC1fspZ9f:-NGrAhsU3t3gFxnau-9z:b-wndkss) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3f607ed7063c091b6deea81db3d0a58a1b5617f8/src/test/kotlin/org/wfanet/measurement/integration/k8s/CorrectnessTest.kt#L134)):*

i'd prefer to throw an error if an unknown duchy id is encountered.
</blockquote></details>

Done. I realized in this case it would actually be an error as ResourceSetup runs with the configuration that this test sets up, and therefore should be unable to create resources for another Duchy.

stevenwarejones (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/733)* status: all files reviewed, 3 unresolved discussions (waiting on @matthewclegg)

SanjayVas (commented):
> Dismissed @matthewclegg from 3 discussions.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/733)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @matthewclegg)

SanjayVas (commented):
> Reviewed 1 of 2 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/733)* status: 35 of 36 files reviewed, all discussions resolved (waiting on @matthewclegg and @stevenwarejones)

SanjayVas (commented):
> Reviewed 1 of 2 files at r11.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/733)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @matthewclegg and @stevenwarejones)

PR #734, opened 2022-10-27
Update configuration of OpenTelemetry Java agent to use Prometheus exporter

tristanvuong2021:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/734)


tristanvuong2021:
The Prometheus exporter that the OpenTelemetry Collector uses is not specific enough to properly convert some metrics.

PR #742, opened 2022-11-03
add cancel measurement

uakyol:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/742)


TNATALI:
No longer valid

PR #745, opened 2022-11-10
Use fixed sizes for JVM memory in dev environment.

SanjayVas:
Percentage sizes do not allow for scaling heap memory independently of non-heap memory.

- - -
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/745)


SanjayVas:
CC @matthewclegg

PR #752, opened 2022-11-21
Prevent decryption of event group metadata when it is not set.

tristanvuong2021:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/752)

### DIFF
--- a/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v1alpha/BUILD.bazel
+++ b/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v1alpha/BUILD.bazel
@@ -34,6 +34,7 @@ kt_jvm_library(
     deps = [
         ":encryption_key_pair_store",
         ":resource_key",
+        "//imports/java/org/projectnessie/cel:core",
         "//src/main/kotlin/org/wfanet/measurement/api:api_key_constants",
         "//src/main/kotlin/org/wfanet/measurement/api/v2alpha:event_group_metadata_parser",
         "//src/main/kotlin/org/wfanet/measurement/eventdataprovider/eventfiltration:event_filters",
--- a/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v1alpha/EventGroupsService.kt
+++ b/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v1alpha/EventGroupsService.kt
@@ -28,6 +28,20 @@ import org.wfanet.measurement.api.v2alpha.ListEventGroupsRequestKt.filter
 import org.wfanet.measurement.api.v2alpha.MeasurementConsumerKey
 import org.wfanet.measurement.api.v2alpha.batchGetEventGroupMetadataDescriptorsRequest
 import org.wfanet.measurement.api.v2alpha.listEventGroupsRequest as cmmsListEventGroupsRequest
+import com.google.api.expr.v1alpha1.Decl
+import com.google.protobuf.DescriptorProtos.DescriptorProto
+import com.google.protobuf.DescriptorProtos.FileDescriptorProto
+import com.google.protobuf.Descriptors
+import com.google.protobuf.Descriptors.FieldDescriptor
+import com.google.protobuf.Descriptors.FileDescriptor
+import com.google.protobuf.DynamicMessage
+import com.google.protobuf.Message
+import com.google.protobuf.copy
+import com.google.protobuf.fileDescriptorProto
+import org.projectnessie.cel.Env
+import org.projectnessie.cel.EnvOption
+import org.projectnessie.cel.checker.Decls
+import org.projectnessie.cel.common.types.pb.ProtoTypeRegistry
 import org.wfanet.measurement.api.withAuthenticationKey
 import org.wfanet.measurement.common.grpc.failGrpc
 import org.wfanet.measurement.common.grpc.grpcRequireNotNull
@@ -41,6 +55,8 @@ import org.wfanet.measurement.reporting.v1alpha.ListEventGroupsRequest
 import org.wfanet.measurement.reporting.v1alpha.ListEventGroupsResponse
 import org.wfanet.measurement.reporting.v1alpha.eventGroup
 import org.wfanet.measurement.reporting.v1alpha.listEventGroupsResponse
+
+private const val EVENT_GROUP_METADATA_FIELD_DESCRIPTOR_NAME_PREFIX = "wfa.measurement.reporting.v1alpha.EventGroup.Metadata."
 
 class EventGroupsService(
   private val cmmsEventGroupsStub: EventGroupsCoroutineStub,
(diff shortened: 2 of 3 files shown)


SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/752)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v1alpha/EventGroupsService.kt` line 95 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/752#-NH_k6ld8KtDNoYIYjbK:-NH_k6ld8KtDNoYIYjbL:b-26nvv3) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/bca993b6a47ace489d5c753f609fa987477b4a2b/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v1alpha/EventGroupsService.kt#L95)):*
> ```Kotlin
>         }
> 
>     if (request.filter.isEmpty() || parsedEventGroupMetadataMap.isEmpty())
> ```

Why does it matter if there's no metadata? The filter can reference any field in the `EventGroup` message, not just the metadata. e.g. `event_group_reference_id`.

tristanvuong2021 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/752)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v1alpha/EventGroupsService.kt` line 95 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/752#-NH_k6ld8KtDNoYIYjbK:-NH_udNB1q_AF-6KbyBY:bjfseis) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/bca993b6a47ace489d5c753f609fa987477b4a2b/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v1alpha/EventGroupsService.kt#L95)):*

Why does it matter if there's no metadata? The filter can reference any field in the `EventGroup` message, not just the metadata. e.g. `event_group_reference_id`.
</blockquote></details>

The rest of the function requires the map to have something in it, although maybe event groups shouldn't be included if there is a filter and no metadata.

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/752)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v1alpha/EventGroupsService.kt` line 95 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/752#-NH_k6ld8KtDNoYIYjbK:-NHa2OZf-tYyFcbC-Zy1:bc1xiqk) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/bca993b6a47ace489d5c753f609fa987477b4a2b/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v1alpha/EventGroupsService.kt#L95)):*

The rest of the function requires the map to have something in it, although maybe event groups shouldn't be included if there is a filter and no metadata.
</blockquote></details>

No filter expression means a filter predicate that always returns `true`, i.e. that *all* EventGroups for the specified parent should be returned. Metadata is optional, and only has an effect on whether an EventGroup is returned *if* the filter happens to reference a metadata field.

tristanvuong2021 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/752)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v1alpha/EventGroupsService.kt` line 95 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/752#-NH_k6ld8KtDNoYIYjbK:-NHaD3235X7ssv5H74OG:b-ffb7e2) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/bca993b6a47ace489d5c753f609fa987477b4a2b/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v1alpha/EventGroupsService.kt#L95)):*

No filter expression means a filter predicate that always returns `true`, i.e. that *all* EventGroups for the specified parent should be returned. Metadata is optional, and only has an effect on whether an EventGroup is returned *if* the filter happens to reference a metadata field.
</blockquote></details>

Then it looks good here. This filter is only meant for the metadata. https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8487b70ec1c30daa91aa55a7db783be6a186b96f/src/main/proto/wfa/measurement/reporting/v1alpha/event_groups_service.proto#L62

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/752)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v1alpha/EventGroupsService.kt` line 95 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/752#-NH_k6ld8KtDNoYIYjbK:-NHaZ87WBmh1CCpCvdh5:bjybl3b) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/bca993b6a47ace489d5c753f609fa987477b4a2b/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v1alpha/EventGroupsService.kt#L95)):*

Then it looks good here. This filter is only meant for the metadata. https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8487b70ec1c30daa91aa55a7db783be6a186b96f/src/main/proto/wfa/measurement/reporting/v1alpha/event_groups_service.proto#L62
</blockquote></details>

That comment is misleading. The filter is supposed to be applied to `EventGroup` messages. See https://google.aip.dev/160

tristanvuong2021 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/752)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v1alpha/EventGroupsService.kt` line 95 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/752#-NH_k6ld8KtDNoYIYjbK:-NHzOPAM6GwSdb2cXWTa:box7gyk) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/bca993b6a47ace489d5c753f609fa987477b4a2b/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v1alpha/EventGroupsService.kt#L95)):*

That comment is misleading. The filter is supposed to be applied to `EventGroup` messages. See https://google.aip.dev/160
</blockquote></details>

This should be fixed in a separate PR. I don't understand how it works at the moment and it would be better to get this fix out first.

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/752)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v1alpha/EventGroupsService.kt` line 95 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/752#-NH_k6ld8KtDNoYIYjbK:-NHaZoC31GzjvKJte-nC:b-cu8shf) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/bca993b6a47ace489d5c753f609fa987477b4a2b/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v1alpha/EventGroupsService.kt#L95)):*

This should be fixed in a separate PR. I don't understand how it works at the moment and it would be better to get this fix out first.
</blockquote></details>

That bug (what the filter is applying to) is actually the more important one to fix, as it means that *every* query with a filter is being handled incorrectly. The bug for event group metadata only affects event groups that don't have any metadata specified.

That is to say if you want to split into separate PRs, the one for incorrectly assuming that event group metadata is always specified should come second.

PR #778, opened 2022-12-05
Duchy deletes Computations of certain states

renjiezh:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/778)


SanjayVas (changes requested):
> Reviewed 21 of 29 files at r1.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/778)* status: 21 of 29 files reviewed, 11 unresolved discussions (waiting on @matthewclegg and @renjiezh)

___
*[`src/main/k8s/duchy.cue` line 27 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/778#-NIiFsOZ1l_U7syl96tA:-NIiFsOZ1l_U7syl96tB:b-rwvc3t) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/130d7c039ea9e6b12dd1430e11f21db2d5ca7aa3/src/main/k8s/duchy.cue#L27)):*
> ```CUE
> 	_duchy_secret_name: string
> 	_computation_control_targets: [Name=_]: string
> 	_deletable_computation_states: [...string]
> ```

nit: fields should use `lowerCamelCase`

I know we have a mix with some using `snake_case`, but they should all be cleaned up



_Code quote:_
```CUE
_deletable_computation_states
```
___
*[`src/main/k8s/local/duchies.cue` line 82 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/778#-NIiEJn130hDa_69SFem:-NIiEJn2Ck5XkL0AYjaG:b5vhysm) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/130d7c039ea9e6b12dd1430e11f21db2d5ca7aa3/src/main/k8s/local/duchies.cue#L82)):*
> ```CUE
> 		_duchy_image_pull_policy: "Never"
> 		_verbose_grpc_logging:    "true"
> 		_deletable_computation_states: ["SUCCEEDED", "CANCELLED"]
> ```

Why do we want a different set of states in local vs. GKE? I figured for now we'd just want to give a default set of states in the base Duchy config and let it be overridden.

Something like:

Base Duchy:
```cue
#TerminalComputationState: "SUCCEEDED" | "FAILED" | "CANCELLED"

_deletableComputationStates: [...#TerminalComputationState] | *["SUCCEEDED"]
```
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/Herald.kt` line 86 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/778#-NIiG5IL8NAe8s_FInNP:-NIiG5IL8NAe8s_FInNQ:b3l6vyt) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/130d7c039ea9e6b12dd1430e11f21db2d5ca7aa3/src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/Herald.kt#L86)):*
> ```Kotlin
>   maxConcurrency: Int = 5,
>   private val retryBackoff: ExponentialBackoff = ExponentialBackoff(),
>   deletableStates: List<String> = emptyList(),
> ```

1. Convert to strong type as early as possible.
2. Use a set since we know there shouldn't be duplicates



_Suggestion:_
```Kotlin
Set<Computation.State>
```
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/Herald.kt` line 90 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/778#-NIiN4CiC7X0DNCIGwwv:-NIiN4CiC7X0DNCIGwww:bo9z5yd) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/130d7c039ea9e6b12dd1430e11f21db2d5ca7aa3/src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/Herald.kt#L90)):*
> ```Kotlin
>   private val semaphore = Semaphore(maxConcurrency)
>   private val continuationTokenManager = ContinuationTokenManager(continuationTokenClient)
> ```

Require that all deleteableStates are terminal.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/Herald.kt` line 319 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/778#-NIiIFgoB4URNvFbcedS:-NIiIFgoB4URNvFbcedT:b9q5unq) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/130d7c039ea9e6b12dd1430e11f21db2d5ca7aa3/src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/Herald.kt#L319)):*
> ```Kotlin
>   }
> 
>   private suspend fun clearComputation(computation: Computation) {
> ```

nit: why does this use a different verb? "delete" is reference everywhere else.

If following the pattern of other methods in this class, this would be `deleteComputationAtDuchy` to distinguish from a "global" Kingdom-level deletion.



_Code quote:_
```Kotlin
clear
```
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/db/computation/testing/FakeComputationsDatabase.kt` line 45 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/778#-NIiIsMWA9WS95yHwb4B:-NIiIsMWA9WS95yHwb4C:brm6kz) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/130d7c039ea9e6b12dd1430e11f21db2d5ca7aa3/src/main/kotlin/org/wfanet/measurement/duchy/db/computation/testing/FakeComputationsDatabase.kt#L45)):*
> ```Kotlin
> import org.wfanet.measurement.internal.duchy.requisitionMetadata
> 
> private const val COMPUTATION_BLOB_KEY_PREFIX = "computations"
> ```

Unused?

These prefixes are supposed to be private to the `Store` anyway



_Code quote:_
```Kotlin
private const val COMPUTATION_BLOB_KEY_PREFIX = "computations"
private const val REQUISITION_BLOB_KEY_PREFIX = "requisitions"
```
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/daemon/herald/HeraldDaemon.kt` line 78 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/778#-NIiHU_AB4NjBhQa76bj:-NIiHU_AB4NjBhQa76bk:b-brfplu) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/130d7c039ea9e6b12dd1430e11f21db2d5ca7aa3/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/daemon/herald/HeraldDaemon.kt#L78)):*
> ```Kotlin
>   @CommandLine.Option(
>     names = ["--deletable-computation-state"],
>     description = ["States of Computations to be deleted."],
> ```

nit: "Terminal State in which the Computation can be deleted. This can be specified multiple times."



_Code quote:_
```Kotlin
States of Computations to be deleted.
```
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/daemon/herald/HeraldDaemon.kt` line 82 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/778#-NIiGh0r06Ny_qdu5WRD:-NIiGh0r06Ny_qdu5WRE:bkej8z4) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/130d7c039ea9e6b12dd1430e11f21db2d5ca7aa3/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/daemon/herald/HeraldDaemon.kt#L82)):*
> ```Kotlin
>     defaultValue = "",
>   )
>   lateinit var deletableComputationStates: List<String>
> ```

Picocli has built-in conversion for enums, and also supports the `Set` collection type.



_Suggestion:_
```Kotlin
Set<Computation.State>
```
___
*[`src/main/proto/wfa/measurement/internal/duchy/computations_service.proto` line 39 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/778#-NIiKOelCNrefGt2-29W:-NIiKOelCNrefGt2-29X:b-gh9w6x) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/130d7c039ea9e6b12dd1430e11f21db2d5ca7aa3/src/main/proto/wfa/measurement/internal/duchy/computations_service.proto#L39)):*
> ```protobuf
>   // Deletes a computation.
>   rpc DeleteComputation(DeleteComputationRequest)
>       returns (DeleteComputationResponse);
> ```

nit: Could also use `google.protobuf.Empty` instead of having a specific response message. This is what the AIPs specify to use for hard deletions.



_Code quote:_
```protobuf
DeleteComputationResponse
```
___
*[`src/main/proto/wfa/measurement/internal/duchy/computations_service.proto` line 104 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/778#-NIiJpNH5k3SQh9wvoig:-NIiJpNH5k3SQh9wvoih:b45h) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/130d7c039ea9e6b12dd1430e11f21db2d5ca7aa3/src/main/proto/wfa/measurement/internal/duchy/computations_service.proto#L104)):*
> ```protobuf
> message DeleteComputationRequest {
>   // The local identifier of the computation.
>   int64 local_computation_id = 2;
> ```





_Suggestion:_
```protobuf
1
```
___
*[`src/test/kotlin/org/wfanet/measurement/duchy/daemon/herald/HeraldTest.kt` line 307 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/778#-NIiLafM4rddkVDGEHyG:-NIiLafM4rddkVDGEHyH:bw1kleh) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/130d7c039ea9e6b12dd1430e11f21db2d5ca7aa3/src/test/kotlin/org/wfanet/measurement/duchy/daemon/herald/HeraldTest.kt#L307)):*
> ```Kotlin
>         Clock.systemUTC(),
>       )
>     mockHerald =
> ```

This is called `mockHerald` but is instantiated with a concrete Herald.

If you need a Herald instance that is instantiated differently for a test method, then instantiate it as a local variable in that method. If you have a Herald instance that's used in all test methods, that's when to have it as a property and instantiate it in a `@Before` method as is done here.

I'm wondering if it would be better to extract functions that can be used in the relevant test methods for instantiating a Herald with the appropriate configuration. Unless you plan on testing with different `deletableStates`, then just adding these states to one or both of (aggregator, non-aggregator) should be fine.



_Code quote:_
```Kotlin
mockHerald
```

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/778)* status: 13 of 31 files reviewed, 7 unresolved discussions (waiting on @matthewclegg and @SanjayVas)

___
*[`src/main/k8s/local/duchies.cue` line 82 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/778#-NIiEJn130hDa_69SFem:-NIiXkYW5kvLwH7Z8629:b-6yosl1) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/130d7c039ea9e6b12dd1430e11f21db2d5ca7aa3/src/main/k8s/local/duchies.cue#L82)):*

Why do we want a different set of states in local vs. GKE? I figured for now we'd just want to give a default set of states in the base Duchy config and let it be overridden.

Something like:

Base Duchy:
```cue
#TerminalComputationState: "SUCCEEDED" | "FAILED" | "CANCELLED"

_deletableComputationStates: [...#TerminalComputationState] | *["SUCCEEDED"]
```
</blockquote></details>

Done.
Thanks for the code sample.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/Herald.kt` line 86 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/778#-NIiG5IL8NAe8s_FInNP:-NIijcQ85hgSHBm0Rypz:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/130d7c039ea9e6b12dd1430e11f21db2d5ca7aa3/src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/Herald.kt#L86)):*

1. Convert to strong type as early as possible.
2. Use a set since we know there shouldn't be duplicates
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/Herald.kt` line 90 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/778#-NIiN4CiC7X0DNCIGwwv:-NIiko5M8P71p8tuZzrP:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/130d7c039ea9e6b12dd1430e11f21db2d5ca7aa3/src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/Herald.kt#L90)):*

Require that all deleteableStates are terminal.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/db/computation/testing/FakeComputationsDatabase.kt` line 45 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/778#-NIiIsMWA9WS95yHwb4B:-NIilURT5TGGKeA59zLV:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/130d7c039ea9e6b12dd1430e11f21db2d5ca7aa3/src/main/kotlin/org/wfanet/measurement/duchy/db/computation/testing/FakeComputationsDatabase.kt#L45)):*

Unused?

These prefixes are supposed to be private to the `Store` anyway
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/daemon/herald/HeraldDaemon.kt` line 82 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/778#-NIiGh0r06Ny_qdu5WRD:-NIilpYL8CSzlkslF8TF:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/130d7c039ea9e6b12dd1430e11f21db2d5ca7aa3/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/daemon/herald/HeraldDaemon.kt#L82)):*

Picocli has built-in conversion for enums, and also supports the `Set` collection type.
</blockquote></details>

Done.
___
*[`src/main/proto/wfa/measurement/internal/duchy/computations_service.proto` line 104 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/778#-NIiJpNH5k3SQh9wvoig:-NIipIFL922gfXVFDpbw:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/130d7c039ea9e6b12dd1430e11f21db2d5ca7aa3/src/main/proto/wfa/measurement/internal/duchy/computations_service.proto#L104)):*
> ```protobuf
> message DeleteComputationRequest {
>   // The local identifier of the computation.
>   int64 local_computation_id = 2;
> ```

Done.
___
*[`src/test/kotlin/org/wfanet/measurement/duchy/daemon/herald/HeraldTest.kt` line 307 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/778#-NIiLafM4rddkVDGEHyG:-NIirQ4I8zEyES0RUzhX:b73sbkd) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/130d7c039ea9e6b12dd1430e11f21db2d5ca7aa3/src/test/kotlin/org/wfanet/measurement/duchy/daemon/herald/HeraldTest.kt#L307)):*

This is called `mockHerald` but is instantiated with a concrete Herald.

If you need a Herald instance that is instantiated differently for a test method, then instantiate it as a local variable in that method. If you have a Herald instance that's used in all test methods, that's when to have it as a property and instantiate it in a `@Before` method as is done here.

I'm wondering if it would be better to extract functions that can be used in the relevant test methods for instantiating a Herald with the appropriate configuration. Unless you plan on testing with different `deletableStates`, then just adding these states to one or both of (aggregator, non-aggregator) should be fine.
</blockquote></details>

Thanks for the comment. I moved `mockHerald` into local variables.
The reason to use `mockHerald` is that with `fakeComputationDatabase`, it is hard to control the API behavior. So I created the `mockHerald` with mockito. It is used for couple of the test case but not all.

SanjayVas (changes requested):
> Reviewed 9 of 11 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/778)* status: 22 of 31 files reviewed, 2 unresolved discussions (waiting on @matthewclegg and @renjiezh)

___
*[`src/test/kotlin/org/wfanet/measurement/duchy/daemon/herald/HeraldTest.kt` line 307 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/778#-NIiLafM4rddkVDGEHyG:-NInm7n1ADRScYFWC-Lj:b-adqwyg) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/130d7c039ea9e6b12dd1430e11f21db2d5ca7aa3/src/test/kotlin/org/wfanet/measurement/duchy/daemon/herald/HeraldTest.kt#L307)):*
>  I created the mockHerald with mockito.

This statement appears to be false. All instances of `mockHerald` I can see are concrete Herald instances, not Mockito mocks.
___
*[`src/test/kotlin/org/wfanet/measurement/duchy/daemon/herald/HeraldTest.kt` line 272 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/778#-NInmi2r2a3D5VfCv0_t:-NInmi2r2a3D5VfCv0_u:b-rjchkq) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3df49f9494e6b2442f78eaa8c75e7e8bca8d5365/src/test/kotlin/org/wfanet/measurement/duchy/daemon/herald/HeraldTest.kt#L272)):*
> ```Kotlin
> 
>   private val internalComputationsMock: InternalComputationsCoroutineImplBase = mockService {}
>   private val mockTestServerRule = GrpcTestServerRule { addService(internalComputationsMock) }
> ```

Neither of these are Mockito mocks and therefore should not have `mock` in the name. (Don't call something a mock just because it happens to depend on a mock, only if it actually *is* a mock).



_Code quote:_
```Kotlin
  private val mockTestServerRule = GrpcTestServerRule { addService(internalComputationsMock) }
  private val mockInternalComputationsStub =
```

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/778)* status: 22 of 31 files reviewed, 2 unresolved discussions (waiting on @matthewclegg and @SanjayVas)

___
*[`src/test/kotlin/org/wfanet/measurement/duchy/daemon/herald/HeraldTest.kt` line 307 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/778#-NIiLafM4rddkVDGEHyG:-NIsMfSx2qKWHZk8Nq3p:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/130d7c039ea9e6b12dd1430e11f21db2d5ca7aa3/src/test/kotlin/org/wfanet/measurement/duchy/daemon/herald/HeraldTest.kt#L307)):*

>  I created the mockHerald with mockito.

This statement appears to be false. All instances of `mockHerald` I can see are concrete Herald instances, not Mockito mocks.
</blockquote></details>

Done.
___
*[`src/test/kotlin/org/wfanet/measurement/duchy/daemon/herald/HeraldTest.kt` line 272 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/778#-NInmi2r2a3D5VfCv0_t:-NIsMg1FDO1aDO87uJcx:b9h2ceq) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3df49f9494e6b2442f78eaa8c75e7e8bca8d5365/src/test/kotlin/org/wfanet/measurement/duchy/daemon/herald/HeraldTest.kt#L272)):*

Neither of these are Mockito mocks and therefore should not have `mock` in the name. (Don't call something a mock just because it happens to depend on a mock, only if it actually *is* a mock).
</blockquote></details>

Thanks for the explanation.
I renamed them. Do you have suggestion of better names here?

SanjayVas (approved):
> Reviewed 7 of 29 files at r1, 1 of 11 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/778)* status: all files reviewed, 2 unresolved discussions (waiting on @matthewclegg and @renjiezh)

___
*[`src/test/kotlin/org/wfanet/measurement/duchy/daemon/herald/HeraldTest.kt` line 272 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/778#-NInmi2r2a3D5VfCv0_t:-NItNOpl2TCqAUUag_EQ:b-28zbdv) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3df49f9494e6b2442f78eaa8c75e7e8bca8d5365/src/test/kotlin/org/wfanet/measurement/duchy/daemon/herald/HeraldTest.kt#L272)):*

Thanks for the explanation.
I renamed them. Do you have suggestion of better names here?
</blockquote></details>

Not really.

The general way I'd avoid this is by not having things defined as class properties that are used mutually exclusively by some test methods. The alternative is instantiating what you need within the test method. Of course, that's tough to do if you want to use GrpcTestServerRule rather than manually making sure the server resource you instantiate in a specific test method are cleaned up.

The other alternative is to swap test doubles at the same level for all tests. It's strange that this test uses a real ComputationsService instance with a fake database for some tests, and a mock service for others. ComputationsService has its own test and doesn't really need to be exercised here.

That said, either of these would require a significant amount of reworking to the test.
___
*[`src/test/kotlin/org/wfanet/measurement/duchy/db/computation/ComputationDataClientsTest.kt` line 71 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/778#-NItSGyT-7NCmXA36KoR:-NItSGyT-7NCmXA36KoS:b-bej5xg) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6accde610e8992fe4fb993f0b83800696970db67/src/test/kotlin/org/wfanet/measurement/duchy/db/computation/ComputationDataClientsTest.kt#L71)):*
> ```Kotlin
> 
>   val grpcTestServerRule = GrpcTestServerRule {
>     val storageClient = FileSystemStorageClient(tempDirectory.root)
> ```

Note that if your test blobs aren't particularly large, you can use InMemoryStorageClient and avoid the need to use TemporaryFolder.
___
*[`src/test/kotlin/org/wfanet/measurement/duchy/service/internal/computations/ComputationsServiceTest.kt` line 101 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/778#-NItUQpD3EevBI9JOIXV:-NItUQpD3EevBI9JOIXW:b-vp7198) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6accde610e8992fe4fb993f0b83800696970db67/src/test/kotlin/org/wfanet/measurement/duchy/service/internal/computations/ComputationsServiceTest.kt#L101)):*
> ```Kotlin
>   @get:Rule val ruleChain = chainRulesSequentially(tempDirectory, grpcTestServerRule)
> 
>   private val fakeService: ComputationsService by lazy {
> ```

nit: mind renaming this to just `service`? This makes it sound like it's a fake rather than the actual subject under test (SUT).



_Code quote:_
```Kotlin
fakeService
```

duliomatos1 (approved):
lgtm, some nits

> Reviewed 16 of 29 files at r1, 10 of 11 files at r2, 5 of 5 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/778#-:-NJ6hMOn0Tlls2cRObgU:byf8m3d)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @matthewclegg)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/Herald.kt` line 95 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/778#-NJ6fkb92w9DMAZ1omPZ:-NJ6fkb92w9DMAZ1omP_:bbvz5a7) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/48a91d3213adb34ff1f10381797a0146f785c7eb/src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/Herald.kt#L95)):*
> ```Kotlin
>   init {
>     for (state in deletableComputationStates) {
>       require(state in TERMINAL_STATES)
> ```

nit: could put user readable message in the case someone pass an unexpected state in the flags e.g.



_Code snippet:_
```Kotlin
 require(state in TERMINAL_STATES) { "Unexpected deletable computation state ${state} while initializing Herald" }
```
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/daemon/herald/HeraldDaemon.kt` line 81 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/778#-NJ6gycLCQS17xv5Ukaa:-NJ6gycLCQS17xv5Ukab:bbz1zz9) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/48a91d3213adb34ff1f10381797a0146f785c7eb/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/daemon/herald/HeraldDaemon.kt#L81)):*
> ```Kotlin
>     description =
>       [
>         "Terminal State in which the Computation can be deleted. This can be specified multiple times."
> ```

nit. could specify it should be one or more of (SUCCEDED, FAILED, CANCELLED)

renjiezh (commented):
> Reviewed 13 of 29 files at r1, 7 of 11 files at r2, 5 of 5 files at r4, 2 of 2 files at r5, 3 of 3 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/778)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @matthewclegg)

PR #792, opened 2022-12-15
Respect collection_interval and filter in EventQuery implementations.

SanjayVas:
The result is that any EDP simulator instance using a known EventQuery impl other than RandomEventQuery should respect these fields.

This also fixes an issue where event filters with empty expressions were not properly being treated as filter predicates that always evaluate to true. Finally, this readjusts test templates to better match real-world usage (e.g. dropping age and gender fields from the Video template, having them on the common Person template instead.)

- - -
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/792)

https://rally1.rallydev.com/#/?detail=/task/683916759667&fdp=true


SanjayVas:
I added a `Person` event template to `TestEvent` to hold all fields that relate to the virtual person. I don't think any of the templates had the social grade field.

I expect that in practice these will be part of a common template for a given market and that templates for specific ad formats will be mutually exclusive (`oneof` in protobuf terms).

CC @iverson52000 and @jcorilla 

There are two changes in this PR that are relevant for the CSV event query:
1. The date format in our synthetic event CSV files is `dd/MM/yyyy`. I updated the test data to make this clearer.
2. An empty filter expression should be treated as an expression that always evaluates to true. There was a test that was instead verifying that an error is thrown in this case.

SanjayVas:
Swapping reviewers since @uakyol is OOO.

iverson52000 (commented):
For 2., when we were working on this, @uakyol explicitly mentioned that eventFilter cannot be empty so that's why we have this check. May need to verify with him if this constraint is still valid

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/792#-:-NJMVlZxDftAtK5aClie:b-kfv1o4)* status: 0 of 13 files reviewed, all discussions resolved (waiting on @jcorilla)

SanjayVas (commented):
event_filter is explicitly documented as optional. See https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/fe8763130763dc07e2fe80a838df9481ef3f928d/src/main/proto/wfa/measurement/api/v2alpha/requisition_spec.proto#L44

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/792#-:-NL3UIA09asepi9vyrP-:b-4kf2zs)* status: 0 of 14 files reviewed, all discussions resolved (waiting on @jcorilla)

robinsons (changes requested):
> Reviewed 2 of 13 files at r1, 1 of 5 files at r2, 49 of 49 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/792)* status: all files reviewed, 4 unresolved discussions (waiting on @SanjayVas and @uakyol)

___
*[`src/main/kotlin/org/wfanet/measurement/common/LocalDateProgression.kt` line 26 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/792#-NMA2iacDifCj7ZV5nBp:-NMA2iacDifCj7ZV5nBq:b-1u5443) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f632ede38945339ae1acd9f3781e0b35b7ed5047/src/main/kotlin/org/wfanet/measurement/common/LocalDateProgression.kt#L26)):*
> ```Kotlin
>   override val start: LocalDate,
>   override val endInclusive: LocalDate,
>   val step: TemporalAmount = IsoChronology.INSTANCE.period(0, 0, 1)
> ```

nit: What do you think of `Period.ofDays(1)`? I find it a bit more readable.



_Code quote:_
```Kotlin
IsoChronology.INSTANCE.period(0, 0, 1)
```
___
*[`src/main/kotlin/org/wfanet/measurement/common/OpenEndedTimeRange.kt` line 28 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/792#-NMA38CB5tM1SIrvYDVc:-NMA38CB5tM1SIrvYDVd:b7l3vbe) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f632ede38945339ae1acd9f3781e0b35b7ed5047/src/main/kotlin/org/wfanet/measurement/common/OpenEndedTimeRange.kt#L28)):*
> ```Kotlin
>  * TODO(@SanjayVas): Move to common-jvm.
>  */
> data class OpenEndedTimeRange(val start: Instant, val endExclusive: Instant) {
> ```

Do we have a dependency on Guava? We could use [Range](https://guava.dev/releases/19.0/api/docs/com/google/common/collect/Range.html) instead of defining our own version.

If we don't end up using Range, I think the name OpenEndedTimeRange is a bit ambiguous. As an interval, this could be construed as "open on both ends" even though the intent is "closed on the left, open on the right". Colloquially, "open ended" could also be taken as something like "unbounded" (although I hope that people wouldn't misunderstand it in this way). Do we have to expose the open-ness? In other words, could we call this something like TimeRange or TimeInterval and put factory methods in the companion object like `closedOpen(start, endExclusive)`?
___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/privacybudgetmanagement/PrivacyLandscape.kt` line 22 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/792#-NMAAXp_CAoj7ylFWOec:-NMAAXp_CAoj7ylFWOed:baswwy3) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f632ede38945339ae1acd9f3781e0b35b7ed5047/src/main/kotlin/org/wfanet/measurement/eventdataprovider/privacybudgetmanagement/PrivacyLandscape.kt#L22)):*
> ```Kotlin
>   const val PRIVACY_BUCKET_VID_SAMPLE_WIDTH = 1f / 300f
> 
>   val datePeriod: ChronoPeriod = IsoChronology.INSTANCE.period(1, 0, 0)
> ```

nit: Same thing here, `Period.ofYears(1)` would be more readable. (I'm assuming it's years; I don't know the order of arguments to `IsoChronology.INSTANCE.period(x, y, z)`.



_Code quote:_
```Kotlin
IsoChronology.INSTANCE.period(1, 0, 0)
```
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/BigQueryEventQuery.kt` line 71 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/792#-NMABebrBVkR44GINjGt:-NMABebrBVkR44GINjGu:b-6zzhdp) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f632ede38945339ae1acd9f3781e0b35b7ed5047/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/BigQueryEventQuery.kt#L71)):*
> ```Kotlin
>     val resultJob: Job? = queryJob.waitFor()
>     if (resultJob == null || resultJob.status.error != null) {
>       throw Exception("Error running query.")
> ```

nit: Do we often throw top-level Exception instead of a more specific type? Also, does `resultJob.status.error` have any message that would be worth showing in the thrown exception?

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/792)* status: all files reviewed, 4 unresolved discussions (waiting on @robinsons and @uakyol)

___
*[`src/main/kotlin/org/wfanet/measurement/common/OpenEndedTimeRange.kt` line 28 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/792#-NMA38CB5tM1SIrvYDVc:-NMADSfg2RiNafVwZBXy:buiv6ev) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f632ede38945339ae1acd9f3781e0b35b7ed5047/src/main/kotlin/org/wfanet/measurement/common/OpenEndedTimeRange.kt#L28)):*

Do we have a dependency on Guava? We could use [Range](https://guava.dev/releases/19.0/api/docs/com/google/common/collect/Range.html) instead of defining our own version.

If we don't end up using Range, I think the name OpenEndedTimeRange is a bit ambiguous. As an interval, this could be construed as "open on both ends" even though the intent is "closed on the left, open on the right". Colloquially, "open ended" could also be taken as something like "unbounded" (although I hope that people wouldn't misunderstand it in this way). Do we have to expose the open-ness? In other words, could we call this something like TimeRange or TimeInterval and put factory methods in the companion object like `closedOpen(start, endExclusive)`?
</blockquote></details>

Once we update to Kotlin 1.7+, this will implement [`OpenEndRange`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.ranges/-open-end-range/). The structure and naming is based on that.

robinsons (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/792)* status: all files reviewed, 3 unresolved discussions (waiting on @SanjayVas and @uakyol)

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/792)* status: 38 of 52 files reviewed, all discussions resolved (waiting on @robinsons and @uakyol)

___
*[`src/main/kotlin/org/wfanet/measurement/common/LocalDateProgression.kt` line 26 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/792#-NMA2iacDifCj7ZV5nBp:-NMB3oN80tnaIZFEEiYn:b-3z400p) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f632ede38945339ae1acd9f3781e0b35b7ed5047/src/main/kotlin/org/wfanet/measurement/common/LocalDateProgression.kt#L26)):*

nit: What do you think of `Period.ofDays(1)`? I find it a bit more readable.
</blockquote></details>

Done. I didn't realize that Period implements TemporalAmount and ChronoPeriod.
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/BigQueryEventQuery.kt` line 71 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/792#-NMABebrBVkR44GINjGt:-NMB4CmMAvt9zsMq9GOG:bo2do7o) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f632ede38945339ae1acd9f3781e0b35b7ed5047/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/BigQueryEventQuery.kt#L71)):*

nit: Do we often throw top-level Exception instead of a more specific type? Also, does `resultJob.status.error` have any message that would be worth showing in the thrown exception?
</blockquote></details>

We do sometimes when we
1. Don't expect the caller to need to handle it in a special way
2. Want to include a cause

Switched to `kotlin.error` here, which uses IllegalStateException underneath.
___
*[`src/main/kotlin/org/wfanet/measurement/common/OpenEndedTimeRange.kt` line 28 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/792#-NMA38CB5tM1SIrvYDVc:-NMB3Yp5CLlnGd7dAdgR:bevcblo) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f632ede38945339ae1acd9f3781e0b35b7ed5047/src/main/kotlin/org/wfanet/measurement/common/OpenEndedTimeRange.kt#L28)):*

Once we update to Kotlin 1.7+, this will implement [`OpenEndRange`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.ranges/-open-end-range/). The structure and naming is based on that.
</blockquote></details>

Added TODO to implement OpenEndRange and renamed to OpenEndTimeRange to more closely parallel its naming.

robinsons (approved):
> Reviewed 3 of 3 files at r4, 11 of 11 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/792)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @uakyol)

uakyol (approved):
Looks good. Related to https://github.com/world-federation-of-advertisers/cross-media-measurement/issues/799#issuecomment-1397556770 and https://github.com/world-federation-of-advertisers/cross-media-measurement/issues/819, I think we should include the top level (e.g. person.age) as the operative field for has support and derive the person.age.value for actual validation and conversion to normal form. Lets keep those open, I am putting together a PR to address this. Better not to do it in this one since this is huge already.

SanjayVas:
Swapped reviewer as Dulio is out.

stevenwarejones (changes requested):
> Reviewed 2 of 13 files at r1, 1 of 5 files at r2, 25 of 49 files at r3, 1 of 3 files at r4, 7 of 11 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/792)* status: all files reviewed, 8 unresolved discussions (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/common/LocalDateProgression.kt` line 23 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/792#-NN7pKM0E5tNe4064pbW:-NN7pKM0E5tNe4064pbX:b9avlzt) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/132d06d2522967d2e8590f0b5db35c0c4e964dcc/src/main/kotlin/org/wfanet/measurement/common/LocalDateProgression.kt#L23)):*
> ```Kotlin
> import java.time.temporal.TemporalAmount
> 
> data class LocalDateProgression(
> ```

todo to move to common please
___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/eventfiltration/EventFilters.kt` line 80 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/792#-NN7sQFX1IVjLwCT_kYa:-NN7sQFX1IVjLwCT_kYb:b-uyov4p) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/132d06d2522967d2e8590f0b5db35c0c4e964dcc/src/main/kotlin/org/wfanet/measurement/eventdataprovider/eventfiltration/EventFilters.kt#L80)):*
> ```Kotlin
>    * event template as fields. See `event_annotations.proto`.
>    */
>   fun createEnv(eventMessageDescriptor: Descriptors.Descriptor): Env {
> ```

if this is going to be public, it needs some tests. or change it back to private.
___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/privacybudgetmanagement/PrivacyBucketFilter.kt` line 67 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/792#-NN7t_d78B-WIaI-qYNb:-NN7t_d78B-WIaI-qYNc:bsw57pg) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/132d06d2522967d2e8590f0b5db35c0c4e964dcc/src/main/kotlin/org/wfanet/measurement/eventdataprovider/privacybudgetmanagement/PrivacyBucketFilter.kt#L67)):*
> ```Kotlin
>       (today.minus(PrivacyLandscape.datePeriod)..today).filter {
>         val startOfDay = it.atStartOfDay().toInstant(ZoneOffset.UTC)
>         val endOfDayExclusive = it.plusDays(1).atStartOfDay().toInstant(ZoneOffset.UTC)
> ```

why not just use `fromClosedDateRange`
___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/privacybudgetmanagement/PrivacyLandscape.kt` line 21 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/792#-NN7u-ed6cRDzykZuXVM:-NN7u-ed6cRDzykZuXVN:bf79iw6) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/b784afe905b0aa1dc12fd50e49d3477a7ce190fb/src/main/kotlin/org/wfanet/measurement/eventdataprovider/privacybudgetmanagement/PrivacyLandscape.kt#L21)):*
> ```Kotlin
>   const val PRIVACY_BUCKET_VID_SAMPLE_WIDTH = 1f / 300f
> 
>   val dates: List<LocalDate> = (0..400).map { LocalDate.now().minusDays(it.toLong()) }
> ```

why was this 400 before?
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/config/PrivacyBudgets.kt` line 36 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/792#-NN7uUS51XPJ1GPL8ClF:-NN7uUS66eH3Xkt_CZAR:bjwqs0x) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/132d06d2522967d2e8590f0b5db35c0c4e964dcc/src/main/kotlin/org/wfanet/measurement/loadtest/config/PrivacyBudgets.kt#L36)):*
> ```Kotlin
>     try {
>       compileProgram(
>         LoadTestEvent.getDescriptor(),
> ```

should this message type get passed into the constructor?
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EventQuery.kt` line 42 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/792#-NN7wfzgEMQu7UedMKw4:-NN7wfzgEMQu7UedMKw5:b56w631) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/132d06d2522967d2e8590f0b5db35c0c4e964dcc/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EventQuery.kt#L42)):*
> ```Kotlin
>     ): Program {
>       if (eventFilter.expression.isEmpty()) {
>         return Program { TRUE_EVAL_RESULT }
> ```

it looks like you're checking in two places if this expression is empty and replacing it with a tautology. Can you combine that logic?
___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_templates/testing/test_banner_template.proto` line 41 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/792#-NN8-BB21WKtru_t9UHO:-NN8-BB21WKtru_t9UHP:bxzp0xg) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/132d06d2522967d2e8590f0b5db35c0c4e964dcc/src/main/proto/wfa/measurement/api/v2alpha/event_templates/testing/test_banner_template.proto#L41)):*
> ```protobuf
>   // pixels on the screen. If not specified, viewability may not have been
>   // measured.
>   ViewableField viewable = 1;
> ```

is there a reason we are adding viewability to these tests?
___
*[`src/main/kotlin/org/wfanet/measurement/common/OpenEndTimeRange.kt` line 26 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/792#-NN2Sm2w-zRuueJB4-cD:-NN2Sm2w-zRuueJB4-cE:ber5e3i) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/132d06d2522967d2e8590f0b5db35c0c4e964dcc/src/main/kotlin/org/wfanet/measurement/common/OpenEndTimeRange.kt#L26)):*
> ```Kotlin
>  * Represents a range of [Instant]s where the upper bound is not included in the range.
>  *
>  * TODO(@SanjayVas): Move to common-jvm.
> ```

is there a reason you don't want to take care of this TODO now?

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/792)* status: all files reviewed, 7 unresolved discussions (waiting on @stevenwarejones and @uakyol)

___
*[`src/main/kotlin/org/wfanet/measurement/common/LocalDateProgression.kt` line 23 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/792#-NN7pKM0E5tNe4064pbW:-NNE02o0Dt1itl2Bjb81:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/132d06d2522967d2e8590f0b5db35c0c4e964dcc/src/main/kotlin/org/wfanet/measurement/common/LocalDateProgression.kt#L23)):*

todo to move to common please
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/eventfiltration/EventFilters.kt` line 80 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/792#-NN7sQFX1IVjLwCT_kYa:-NNDWAW2AZzKTczUYcl3:b7yzb8i) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/132d06d2522967d2e8590f0b5db35c0c4e964dcc/src/main/kotlin/org/wfanet/measurement/eventdataprovider/eventfiltration/EventFilters.kt#L80)):*

if this is going to be public, it needs some tests. or change it back to private.
</blockquote></details>

Done (made private)
___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/privacybudgetmanagement/PrivacyBucketFilter.kt` line 67 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/792#-NN7t_d78B-WIaI-qYNb:-NNDXPhE85Z8OeArO9wd:bvvspvs) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/132d06d2522967d2e8590f0b5db35c0c4e964dcc/src/main/kotlin/org/wfanet/measurement/eventdataprovider/privacybudgetmanagement/PrivacyBucketFilter.kt#L67)):*

why not just use `fromClosedDateRange`
</blockquote></details>

Done.

Because I had written this before writing `fromClosedDateRange` :P
___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/privacybudgetmanagement/PrivacyLandscape.kt` line 21 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/792#-NN7u-ed6cRDzykZuXVM:-NNDY08c87kEQP0yxacN:b-dk9k46) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/b784afe905b0aa1dc12fd50e49d3477a7ce190fb/src/main/kotlin/org/wfanet/measurement/eventdataprovider/privacybudgetmanagement/PrivacyLandscape.kt#L21)):*

why was this 400 before?
</blockquote></details>

My understanding was that we were supposed to cover one year. Perhaps @uakyol can shed some light on why this was specified as 400 days instead of an ISO 8601 standard year. AFAIK, there shouldn't be any need to pad this out.

The previous code attempted to fully enumerate the set of dates in the landscape, though this unfortunately is impossible with a static member as it forces "today" to be evaluated statically rather than at request time.
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/config/PrivacyBudgets.kt` line 36 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/792#-NN7uUS51XPJ1GPL8ClF:-NNDb1RI9lZxdEKVR_7E:bzw843) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/132d06d2522967d2e8590f0b5db35c0c4e964dcc/src/main/kotlin/org/wfanet/measurement/loadtest/config/PrivacyBudgets.kt#L36)):*

should this message type get passed into the constructor?
</blockquote></details>

No. `PrivacyBucketMapper` implementations are tied to specific message types. e.g. the expression `privacy.filterable == true` assumes a specific structure, and `toEventMessage` needs to return the right message type.

See `org.wfanet.measurement.eventdataprovider.privacybudgetmanagement.testing.TestPrivacyBucketMapper` for another implementation that uses a different event message type, and note how the `toEventMessage` logic depends on that type.
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EventQuery.kt` line 42 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/792#-NN7wfzgEMQu7UedMKw4:-NNDclzH5ArCpCI1lZZE:b-c2ej1r) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/132d06d2522967d2e8590f0b5db35c0c4e964dcc/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EventQuery.kt#L42)):*

it looks like you're checking in two places if this expression is empty and replacing it with a tautology. Can you combine that logic?
</blockquote></details>

Added a comment indicating that this is only an optimization.

`EventFilterValidator.compile` is a public method, so putting the logic there ensures that it applies regardless of the entrypoint. This check here is an optimization that can skip Env creation.
___
*[`src/main/kotlin/org/wfanet/measurement/common/OpenEndTimeRange.kt` line 26 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/792#-NN2Sm2w-zRuueJB4-cD:-NNE03ZeEzkxXdr6tS0G:bn6cr0r) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/132d06d2522967d2e8590f0b5db35c0c4e964dcc/src/main/kotlin/org/wfanet/measurement/common/OpenEndTimeRange.kt#L26)):*

is there a reason you don't want to take care of this TODO now?
</blockquote></details>

Now that all the code has been seen in context I suppose I could, but at this point I'd rather get it in as this PR has been open for awhile.

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/792)* status: all files reviewed, 4 unresolved discussions (waiting on @robinsons, @stevenwarejones, and @uakyol)

___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/privacybudgetmanagement/PrivacyBucketFilter.kt` line 67 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/792#-NN7t_d78B-WIaI-qYNb:-NNcOXxE2BhUYT9l6KPl:bgqpowg) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/132d06d2522967d2e8590f0b5db35c0c4e964dcc/src/main/kotlin/org/wfanet/measurement/eventdataprovider/privacybudgetmanagement/PrivacyBucketFilter.kt#L67)):*

Done.

Because I had written this before writing `fromClosedDateRange` :P
</blockquote></details>

Sorry, I was wrong. It
___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_templates/testing/person_template.proto` line 28 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/792#-NNMoi-xC8U2Lr3g8d6c:-NNcL3X--yYHZL_E2WUi:b9pu46p) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/619252397d3646d74ae832cbb6cc3f687db2d1e4/src/main/proto/wfa/measurement/api/v2alpha/event_templates/testing/person_template.proto#L28)):*

maybe call this message `TestPersonTemplate` like the other templates
</blockquote></details>

Renamed the other templates instead.
___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_templates/testing/test_banner_template.proto` line 41 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/792#-NN8-BB21WKtru_t9UHO:-NNOMYLc6znBsZKJYtA_:b-rgsmzl) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/132d06d2522967d2e8590f0b5db35c0c4e964dcc/src/main/proto/wfa/measurement/api/v2alpha/event_templates/testing/test_banner_template.proto#L41)):*

I think the comment is slightly confusing. Unspecified just defaults to false right? Maybe there's a better way of structuring viewability to give the advertiser the option of having viewable+unspecified or just explicitly viewable. Maybe an enum of:
UNDEFINED, NOT_VIEWABLE, VIEWABLE..... wdyt?
</blockquote></details>

Callers have a way to determine whether the value was unspecified. Since the field is of message type, you can determine whether the field itself was set using the `has` CEL function. Therefore unset is distinct from not viewable.

Right now all template fields be of message type and therefore have presence as a separate bit. This is because the `event_field` option can only apply to messages.
___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_templates/testing/test_event.proto` line 29 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/792#-NNMnPlA6RJDSSXbesP3:-NNcLEIJ3jr1055HEGPe:b29wbjd) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/619252397d3646d74ae832cbb6cc3f687db2d1e4/src/main/proto/wfa/measurement/api/v2alpha/event_templates/testing/test_event.proto#L29)):*

I would think its best practice to keep these in numerical order? Is that not correct?
</blockquote></details>

Not necessarily. These are generally listed in some logical order that is independent of their tag numbering. On initial definition, the logical order will match the tag number ordering since there's no reason not to. If you want to maintain backwards compatibility, you cannot change field tag numbers but you can add fields with new tag numbers.

That said, this is not a backwards-compatible change and does not need to be as the message is only used by tests, so I updated the tag numbers/ordering.

stevenwarejones (changes requested):
> Reviewed 23 of 23 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/792)* status: all files reviewed, 2 unresolved discussions (waiting on @robinsons, @SanjayVas, and @uakyol)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_templates/testing/test_banner_template.proto` line 41 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/792#-NN8-BB21WKtru_t9UHO:-NNcf5TEDyUPK2RL2jRl:b-fqqiga) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/132d06d2522967d2e8590f0b5db35c0c4e964dcc/src/main/proto/wfa/measurement/api/v2alpha/event_templates/testing/test_banner_template.proto#L41)):*

Callers have a way to determine whether the value was unspecified. Since the field is of message type, you can determine whether the field itself was set using the `has` CEL function. Therefore unset is distinct from not viewable.

Right now all template fields be of message type and therefore have presence as a separate bit. This is because the `event_field` option can only apply to messages.
</blockquote></details>

the `has` operator is a little confusing since it only works like you are saying on `oneof` and `message` types. Since this is a message, a sophisticated cel expression writer will know this is possible. However, supposing that the `viewable` field is set but `ViewableField.value` is not set, then what you are saying is not possible, according to my reading of the spec.

```
If f is some other singular field, has(e.f) indicates whether the field's value 
is its default value (zero for numeric fields, false for booleans, empty for 
strings and bytes)
```


This requires an edp to know to know to not populate the viewable field entirely rather than including it but not populating the value. 

I guess we could just include a more specific comment that `viewable` should not be populated at all if the value is unknown.

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/792)* status: all files reviewed, 2 unresolved discussions (waiting on @robinsons, @stevenwarejones, and @uakyol)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_templates/testing/test_banner_template.proto` line 41 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/792#-NN8-BB21WKtru_t9UHO:-NNcw2Uc2auHxz6DUx0f:b-yk7ogk) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/132d06d2522967d2e8590f0b5db35c0c4e964dcc/src/main/proto/wfa/measurement/api/v2alpha/event_templates/testing/test_banner_template.proto#L41)):*
> supposing that the `viewable` field is set but `ViewableField.value` is not set

> This requires an edp to know to know to not populate the viewable field entirely rather than including it but not populating the value.

Yes, in this case the `value` field being unset is indistinguishable from it being set to false. Therefore the EDP should set the field accordingly. Their knowledge of the protobuf spec is assumed/required.

> I guess we could just include a more specific comment that viewable should not be populated at all if the value is unknown.

This is part of how protobuf works and doesn't need a specific explanation.  The protobuf developer guide has a [section on field presence](https://protobuf.dev/programming-guides/field_presence/#presence-in-proto3-apis) in proto3 that covers this.

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/792)* status: all files reviewed, 2 unresolved discussions (waiting on @robinsons, @SanjayVas, and @uakyol)

___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/privacybudgetmanagement/PrivacyLandscape.kt` line 21 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/792#-NN7u-ed6cRDzykZuXVM:-NNd-UPaDPkuRf176RgE:b8pvcsx) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/b784afe905b0aa1dc12fd50e49d3477a7ce190fb/src/main/kotlin/org/wfanet/measurement/eventdataprovider/privacybudgetmanagement/PrivacyLandscape.kt#L21)):*

My understanding was that we were supposed to cover one year. Perhaps @uakyol can shed some light on why this was specified as 400 days instead of an ISO 8601 standard year. AFAIK, there shouldn't be any need to pad this out.

The previous code attempted to fully enumerate the set of dates in the landscape, though this unfortunately is impossible with a static member as it forces "today" to be evaluated statically rather than at request time.
</blockquote></details>

@uakyol ?
___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_templates/testing/test_banner_template.proto` line 41 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/792#-NN8-BB21WKtru_t9UHO:-NNd-_cHCyNiw0XYEyfk:b-mee1m0) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/132d06d2522967d2e8590f0b5db35c0c4e964dcc/src/main/proto/wfa/measurement/api/v2alpha/event_templates/testing/test_banner_template.proto#L41)):*

> supposing that the `viewable` field is set but `ViewableField.value` is not set

> This requires an edp to know to know to not populate the viewable field entirely rather than including it but not populating the value.

Yes, in this case the `value` field being unset is indistinguishable from it being set to false. Therefore the EDP should set the field accordingly. Their knowledge of the protobuf spec is assumed/required.

> I guess we could just include a more specific comment that viewable should not be populated at all if the value is unknown.

This is part of how protobuf works and doesn't need a specific explanation.  The protobuf developer guide has a [section on field presence](https://protobuf.dev/programming-guides/field_presence/#presence-in-proto3-apis) in proto3 that covers this.
</blockquote></details>

I do not assume as sophisticated an edp as you do. The CEL documentation does explicitly document this although as you mentioned, one could have guessed this behavior from the protobuf documentation on default values. 

Can you change `  // If not specified, viewability may not have been measured.
` to `  // If not specified, viewability was not determined.`

Additionally, given unsophisticated EDPs, we probably need helper libraries in the future to construct these messages based on input data.

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/792)* status: all files reviewed, 2 unresolved discussions (waiting on @robinsons, @stevenwarejones, and @uakyol)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_templates/testing/test_banner_template.proto` line 41 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/792#-NN8-BB21WKtru_t9UHO:-NNd8yRP0J5D_vmdjS3q:b-h2vas1) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/132d06d2522967d2e8590f0b5db35c0c4e964dcc/src/main/proto/wfa/measurement/api/v2alpha/event_templates/testing/test_banner_template.proto#L41)):*
> Can you change `// If not specified, viewability may not have been measured.` to `// If not specified, viewability was not determined.`

Done.

Helper libraries are a separate discussion as that topic can get much broader (e.g. which languages and dependency versions to support/maintain).

uakyol (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/792)* status: 57 of 58 files reviewed, 2 unresolved discussions (waiting on @robinsons and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/privacybudgetmanagement/PrivacyLandscape.kt` line 21 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/792#-NN7u-ed6cRDzykZuXVM:-NNgVW397Vzqg0WrnEHC:b-lj0c0) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/b784afe905b0aa1dc12fd50e49d3477a7ce190fb/src/main/kotlin/org/wfanet/measurement/eventdataprovider/privacybudgetmanagement/PrivacyLandscape.kt#L21)):*

@uakyol ?
</blockquote></details>

sorry for the late reply. 400 was just a number we had talked informally. We wanted it to be larger than a year. This should basically stretch as far back as the maximum allowed measurement date range into the past - which was thought of as a year. I had a plan to implement a Requistion Validator class to do some sanity checks on the received requisitions in the EDP Simulator. Here is my proposal : 

Lets add a comment here 
// This is the maximum number of days into the past for a measurement.   
// TODO(@uakyol): Implement RequisitionValidators to enforce this in EDPSimulator.

This way, any EDP using this codebase either uses the code unchanged and it works fine as long as measurements stretch less than a year in to the past, or they change this parameter to accommodate their needs.


wdyt?

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/792)* status: 57 of 58 files reviewed, 2 unresolved discussions (waiting on @robinsons, @SanjayVas, and @uakyol)

___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/privacybudgetmanagement/PrivacyLandscape.kt` line 21 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/792#-NN7u-ed6cRDzykZuXVM:-NNgq_udBG9dVQabY6ns:b-hys47f) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/b784afe905b0aa1dc12fd50e49d3477a7ce190fb/src/main/kotlin/org/wfanet/measurement/eventdataprovider/privacybudgetmanagement/PrivacyLandscape.kt#L21)):*

sorry for the late reply. 400 was just a number we had talked informally. We wanted it to be larger than a year. This should basically stretch as far back as the maximum allowed measurement date range into the past - which was thought of as a year. I had a plan to implement a Requistion Validator class to do some sanity checks on the received requisitions in the EDP Simulator. Here is my proposal : 

Lets add a comment here 
// This is the maximum number of days into the past for a measurement.   
// TODO(@uakyol): Implement RequisitionValidators to enforce this in EDPSimulator.

This way, any EDP using this codebase either uses the code unchanged and it works fine as long as measurements stretch less than a year in to the past, or they change this parameter to accommodate their needs.


wdyt?
</blockquote></details>

i think we need to allow this to be overwritten by the EDP if they want a shorter or longer timeframe then, right?
___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/privacybudgetmanagement/PrivacyLandscape.kt` line 20 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/792#-NNgqU8e0ySYJUKKQear:-NNgqU8e0ySYJUKKQeas:b-pdiwri) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1c5b045af906c8688f3f79bc5894819c949c7016/src/main/kotlin/org/wfanet/measurement/eventdataprovider/privacybudgetmanagement/PrivacyLandscape.kt#L20)):*
> ```Kotlin
> 
> object PrivacyLandscape {
>   const val PRIVACY_BUCKET_VID_SAMPLE_WIDTH = 1f / 300f
> ```

how was this width decided @uakyol ?

uakyol (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/792)* status: 57 of 58 files reviewed, 2 unresolved discussions (waiting on @robinsons and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/privacybudgetmanagement/PrivacyLandscape.kt` line 21 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/792#-NN7u-ed6cRDzykZuXVM:-NNh6R-P-W6tnhAdPx22:b-88yzzs) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/b784afe905b0aa1dc12fd50e49d3477a7ce190fb/src/main/kotlin/org/wfanet/measurement/eventdataprovider/privacybudgetmanagement/PrivacyLandscape.kt#L21)):*

i think we need to allow this to be overwritten by the EDP if they want a shorter or longer timeframe then, right?
</blockquote></details>

yes of course.
___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/privacybudgetmanagement/PrivacyLandscape.kt` line 20 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/792#-NNgqU8e0ySYJUKKQear:-NNh6YGD35zwxXLyKgd6:ba1y0x6) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1c5b045af906c8688f3f79bc5894819c949c7016/src/main/kotlin/org/wfanet/measurement/eventdataprovider/privacybudgetmanagement/PrivacyLandscape.kt#L20)):*

how was this width decided @uakyol ?
</blockquote></details>

it is an a priori decision/optimization made before all this implementation.  You can find more details here:  https://docs.google.com/presentation/d/1tM3n2KQzmn2YLvPvyR41K2krrHPQsyMtx0wOO3ZBoCg/edit?resourcekey=0-0wVT-O_6pUCHL2xFIl4gcg#slide=id.p
 
Although, this too can be overwritten by the EDPs. Keep in mind PrivacyBudgetManager is an EDP specific component that allows EDPs to maintain differential privacy for their data. Thus, they can even opt out of using a PBM and this is just a reference implementation.

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/792)* status: 55 of 59 files reviewed, 2 unresolved discussions (waiting on @robinsons and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/privacybudgetmanagement/PrivacyLandscape.kt` line 21 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/792#-NN7u-ed6cRDzykZuXVM:-NNhBq5T1QvPgBzxxTeG:b-rstpkc) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/b784afe905b0aa1dc12fd50e49d3477a7ce190fb/src/main/kotlin/org/wfanet/measurement/eventdataprovider/privacybudgetmanagement/PrivacyLandscape.kt#L21)):*
I added the following in `EdpSimulator.verifySpecifications` instead (where we verify the requisition spec):

```kotlin
// TODO(@uakyol): Validate that collection interval is not outside of privacy landscape.
```

>  We wanted it to be larger than a year. This should basically stretch as far back as the maximum allowed measurement date range into the past - which was thought of as a year.

@uakyol So should it be longer than one year or is it fine to be exactly one year? I imagine markets will tend to have a requirement of "at least one year", which "exactly one year" satisfies.

Regardless, the privacy budget manager reference implementation isn't something that EDPs can use directly unchanged even if they are using JVM. Markets will have different requirements, and the implementation is tied to a specific schema (3 specific age groups and 2 specific genders).

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/792)* status: 54 of 59 files reviewed, 2 unresolved discussions (waiting on @robinsons and @uakyol)

___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/privacybudgetmanagement/PrivacyLandscape.kt` line 20 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/792#-NNgqU8e0ySYJUKKQear:-NNhFiWS5RWjykfIORQj:b-i27a09) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1c5b045af906c8688f3f79bc5894819c949c7016/src/main/kotlin/org/wfanet/measurement/eventdataprovider/privacybudgetmanagement/PrivacyLandscape.kt#L20)):*

it is an a priori decision/optimization made before all this implementation.  You can find more details here:  https://docs.google.com/presentation/d/1tM3n2KQzmn2YLvPvyR41K2krrHPQsyMtx0wOO3ZBoCg/edit?resourcekey=0-0wVT-O_6pUCHL2xFIl4gcg#slide=id.p
 
Although, this too can be overwritten by the EDPs. Keep in mind PrivacyBudgetManager is an EDP specific component that allows EDPs to maintain differential privacy for their data. Thus, they can even opt out of using a PBM and this is just a reference implementation.
</blockquote></details>

also - does this need to be the same as the sample width for a requisition?

uakyol (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/792)* status: 54 of 59 files reviewed, 2 unresolved discussions (waiting on @robinsons and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/privacybudgetmanagement/PrivacyLandscape.kt` line 20 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/792#-NNgqU8e0ySYJUKKQear:-NNlx1GqB4LeLgQtg4yd:blu6dmj) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1c5b045af906c8688f3f79bc5894819c949c7016/src/main/kotlin/org/wfanet/measurement/eventdataprovider/privacybudgetmanagement/PrivacyLandscape.kt#L20)):*

also - does this need to be the same as the sample width for a requisition?
</blockquote></details>

No, this is the unit width in the VID interval. 
VID interval is the unit interval - (0..1). 
We divide this into 300 equal buckets. 
When a requisition is sent, it specifies a range in this interval - e.g. (0.2 - 0.3) 
PBM then finds all the buckets that are in that range (0.2, 0.2+1/300, 0.2+2/300 ..... 0.3) together with other dimensions and charges them.

stevenwarejones (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/792)* status: 54 of 59 files reviewed, all discussions resolved (waiting on @robinsons)

___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/privacybudgetmanagement/PrivacyLandscape.kt` line 20 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/792#-NNgqU8e0ySYJUKKQear:-NNm7LN-6hRxak4rBUM0:bsczom8) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1c5b045af906c8688f3f79bc5894819c949c7016/src/main/kotlin/org/wfanet/measurement/eventdataprovider/privacybudgetmanagement/PrivacyLandscape.kt#L20)):*

No, this is the unit width in the VID interval. 
VID interval is the unit interval - (0..1). 
We divide this into 300 equal buckets. 
When a requisition is sent, it specifies a range in this interval - e.g. (0.2 - 0.3) 
PBM then finds all the buckets that are in that range (0.2, 0.2+1/300, 0.2+2/300 ..... 0.3) together with other dimensions and charges them.

</blockquote></details>

thx - that was helpful

SanjayVas (commented):
> Reviewed 1 of 1 files at r8, 3 of 3 files at r9, 1 of 1 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/792)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @robinsons)

PR #793, opened 2022-12-16
Add padding to reporting frequency result values list.

tristanvuong2021:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/793)

### DIFF
--- a/src/main/kotlin/org/wfanet/measurement/reporting/deploy/postgres/writers/SetMeasurementResult.kt
+++ b/src/main/kotlin/org/wfanet/measurement/reporting/deploy/postgres/writers/SetMeasurementResult.kt
@@ -23,6 +23,7 @@ import java.util.concurrent.TimeUnit
 import org.wfanet.measurement.common.db.r2dbc.boundStatement
 import org.wfanet.measurement.internal.reporting.Measurement
 import org.wfanet.measurement.internal.reporting.Metric
+import org.wfanet.measurement.internal.reporting.Metric.MeasurementCalculation
 import org.wfanet.measurement.internal.reporting.PeriodicTimeInterval
 import org.wfanet.measurement.internal.reporting.Report
 import org.wfanet.measurement.internal.reporting.ReportKt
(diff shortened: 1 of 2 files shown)


riemanli (commented):
Hi Tristan, could you elaborate more about what issue/improvement this PR is aiming for?

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/793#-:-NJSnjJx9XRCpndgPQAY:bvoiycp)* status: 0 of 2 files reviewed, all discussions resolved

riemanli (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/793)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/internal/testing/MeasurementsServiceTest.kt` line 436 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/793#-NJSxzZv-6HUBom2RPDb:-NJSxzZv-6HUBom2RPDc:bsglol1) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/26cc0294d7ef72a594134bd5c9309777ec1620c2/src/main/kotlin/org/wfanet/measurement/reporting/service/internal/testing/MeasurementsServiceTest.kt#L436)):*
> ```Kotlin
>                   setOperations += 260.0
>                   setOperations += 439.0
>                   setOperations += 0.0
> ```

Sorry I am a bit confused here. Why are there only 3 values in the column when there are 6 rows? Also, could you explain how [260, 439] are computed? Thanks!



_Code quote:_
```Kotlin
                  setOperations += 260.0
                  setOperations += 439.0
                  setOperations += 0.0
```

tristanvuong2021 (commented):
It is possible for there to be less frequency values than rows for a given time interval. If there are multiple time intervals, just looking at the result you wouldn't be able to tell which values belong to which row. So I added sorting to make sure the frequency matches and I added padding so there will always be a 1:1 number of rows to values.

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/793#-:-NJfZIEa08Zp50XDYheD:b-myblb3)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @riemanli)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/internal/testing/MeasurementsServiceTest.kt` line 436 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/793#-NJSxzZv-6HUBom2RPDb:-NJfZ-BH-5Xce6sUuN-s:bipk71u) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/26cc0294d7ef72a594134bd5c9309777ec1620c2/src/main/kotlin/org/wfanet/measurement/reporting/service/internal/testing/MeasurementsServiceTest.kt#L436)):*

Sorry I am a bit confused here. Why are there only 3 values in the column when there are 6 rows? Also, could you explain how [260, 439] are computed? Thanks!
</blockquote></details>

I used a shared named set operation defined at the bottom and calculated the values using that and the measurements in the unit test itself. I also changed the named set operation so the number of values would match the number of rows.

riemanli (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/793)* status: 0 of 2 files reviewed, all discussions resolved

duliomatos1 (approved):
> Reviewed 1 of 1 files at r4, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/793)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @tristanvuong2021)

Issue #800, opened 2023-01-06
Potential timing issue in Herald syncStatuses


SanjayVas:
**Describe the bug**
HeraldTest passes locally but failed on GitHub Actions. It does not always fail, i.e. it's flaky. This implies a timing issue. It may be that it's more likely to fail in the resource-constrained environment of GHA.

**Steps to reproduce**
1. Run Bazel build and test action on GHA

**Component(s) affected**
Duchy

**Version**
750a2c993bc4dc4abb7ff34595f7bd6761a4aa9c, based on HEAD at 6184fb99fa02e25c15d66c12a86dabb341471822

**Environment**
Local testing

**Additional context**
Trace from [GitHub Actions run](https://github.com/world-federation-of-advertisers/cross-media-measurement/actions/runs/3857672262):
```
syncStatuses creates new computations(org.wfanet.measurement.duchy.daemon.herald.HeraldTest)
expected: token_for_321
but was : token_for_454647484950
	at org.wfanet.measurement.duchy.daemon.herald.HeraldTest$syncStatuses creates new computations$1.invokeSuspend(HeraldTest.kt:347)
	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
	at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:106)
	at kotlinx.coroutines.test.TestDispatcher.processEvent$kotlinx_coroutines_test(TestDispatcher.kt:28)
	at kotlinx.coroutines.test.TestCoroutineScheduler.tryRunNextTask(TestCoroutineScheduler.kt:95)
	at kotlinx.coroutines.test.TestCoroutineScheduler.advanceUntilIdle(TestCoroutineScheduler.kt:110)
	at kotlinx.coroutines.test.TestBuildersKt__TestBuildersKt.runTestCoroutine(TestBuilders.kt:212)
	at kotlinx.coroutines.test.TestBuildersKt.runTestCoroutine(Unknown Source)
	at kotlinx.coroutines.test.TestBuildersKt__TestBuildersKt$runTestCoroutine$1.invokeSuspend(TestBuilders.kt)
	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
	at kotlinx.coroutines.selects.SelectBuilderImpl.resumeWith(Select.kt:300)
[...]

PR #802, opened 2023-01-09
Fix flaky herald test

renjiezh:
Related issue https://github.com/world-federation-of-advertisers/cross-media-measurement/issues/800

Cause: ContinuationToken service updates the latest token without checking the timestamp. However the order of update requests are not strictly in time series. This is designed to keep the logic simple. 
Solution: update unit test to tolerate.

- - -
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/802)


SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/802)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @renjiezh)

*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/802#-NLRoBH5CFuZqccvj09F:-NLRoBH61QQD9NJ6G7JM:bey8woi) (no related file):*
Sorry, shouldn't the token updating be deterministic (especially in the controlled circumstances of a unit test)?

___

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/802)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/802#-NLRoBH5CFuZqccvj09F:-NLRoYcF1Im1efCGyLmI:b3titxc) (no related file):*

Sorry, shouldn't the token updating be deterministic (especially in the controlled circumstances of a unit test)?
</blockquote></details>

https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/main/src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/ContinuationTokenManager.kt#L100

Herald is now using `launch`to run multiple computations in parallel. Suppose Computation1 and Computation2 complete almost at the same time, task1 will update token1 and task2 will update token2. On the ComputationService side, it receives these two update requests and the order of update is not deterministic.
I think this is tolerable.
If we want to eliminate this case, just move the update call under synchronized block.

___

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/802)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @renjiezh)

*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/802#-NLRoBH5CFuZqccvj09F:-NLRyHlN6mP7xqxd2vzY:b-gc7smv) (no related file):*

https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/main/src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/ContinuationTokenManager.kt#L100

Herald is now using `launch`to run multiple computations in parallel. Suppose Computation1 and Computation2 complete almost at the same time, task1 will update token1 and task2 will update token2. On the ComputationService side, it receives these two update requests and the order of update is not deterministic.
I think this is tolerable.
If we want to eliminate this case, just move the update call under synchronized block.
</blockquote></details>

Aren't computation tokens strictly ordered, such that we don't update the last processed token unless all previous tokens have been processed? e.g. we get an update with token1 and token2 and record as pending in that order, processing them asynchronously. If token2 completes first, the last processed token is still token0. Only once token1 is also processed does token2 get marked as last processed.

syncStatuses suspends until all tokens are processed and we only have one syncStatuses in-flight at a time.

The strict ordering assumption is important for the StreamActiveComputations call, as it only returns tokens that are later than the one specified in the request.

___

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/802)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/802#-NLRoBH5CFuZqccvj09F:-NLS-3kO5wqKH6Yf66l9:b-sr88al) (no related file):*

Aren't computation tokens strictly ordered, such that we don't update the last processed token unless all previous tokens have been processed? e.g. we get an update with token1 and token2 and record as pending in that order, processing them asynchronously. If token2 completes first, the last processed token is still token0. Only once token1 is also processed does token2 get marked as last processed.

syncStatuses suspends until all tokens are processed and we only have one syncStatuses in-flight at a time.

The strict ordering assumption is important for the StreamActiveComputations call, as it only returns tokens that are later than the one specified in the request.
</blockquote></details>

Yeah, that order is guaranteed by the Herald. But the issue is, if token1 completes then token2 completes, there will be two different update requests sending to the ComputationService. When ComputationService handing the requests and the time between those two requests are too close, this is a race condition to update the value in spanner. 

___

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/802)* status: 0 of 7 files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/802#-NLRoBH5CFuZqccvj09F:-NLY9Y_RB9GOiXbLVccT:bfnyxa9) (no related file):*

Yeah, that order is guaranteed by the Herald. But the issue is, if token1 completes then token2 completes, there will be two different update requests sending to the ComputationService. When ComputationService handing the requests and the time between those two requests are too close, this is a race condition to update the value in spanner. 
</blockquote></details>

Rewrite the logic of setContinuationToken so it checks the timestamps. Please review.

___

SanjayVas (changes requested):
> Reviewed 1 of 1 files at r1, 4 of 6 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/802)* status: 5 of 7 files reviewed, 4 unresolved discussions (waiting on @renjiezh)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/gcloud/spanner/continuationtoken/SpannerContinuationTokensService.kt` line 45 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/802#-NLbiZZlCs1D7Ds2BlCQ:-NLbiZZlCs1D7Ds2BlCR:b-4evdra) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/b5dc7a839bc31d70526f1c8cd0eb982fa4d3233e/src/main/kotlin/org/wfanet/measurement/duchy/deploy/gcloud/spanner/continuationtoken/SpannerContinuationTokensService.kt#L45)):*
> ```Kotlin
>     try {
>       SetContinuationToken(request.token).execute(client)
> ```

We don't know for certain that all exceptions are caused by an invalid continuation token. You may want to throw a custom exception type so we can identify whether this was indeed due to the new continuation token being older than the existing one.



_Code quote:_
```Kotlin
} catch (e: Exception) {
```
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/gcloud/spanner/continuationtoken/SpannerContinuationTokensService.kt` line 46 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/802#-NLbikqe8k0anZDAgbIM:-NLbikqe8k0anZDAgbIN:b-ett7b9) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/b5dc7a839bc31d70526f1c8cd0eb982fa4d3233e/src/main/kotlin/org/wfanet/measurement/duchy/deploy/gcloud/spanner/continuationtoken/SpannerContinuationTokensService.kt#L46)):*
> ```Kotlin
>       SetContinuationToken(request.token).execute(client)
>       failGrpc(Status.INVALID_ARGUMENT) { e.message ?: "Invalid continuation token." }
> ```

`INVALID_ARGUMENT` in gRPC is for when the request is malformed regardless of system state. This should be `FAILED_PRECONDITION` instead.

Note that this is different than the definition of IllegalArgumentException in Java.



_Suggestion:_
```Kotlin
FAILED_PRECONDITION
```
___
*[`src/test/kotlin/org/wfanet/measurement/duchy/daemon/herald/HeraldTest.kt` line 346 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/802#-NLbk20MAtjXF2BGmCA1:-NLbk20MAtjXF2BGmCA2:bsls6nz) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/b5dc7a839bc31d70526f1c8cd0eb982fa4d3233e/src/test/kotlin/org/wfanet/measurement/duchy/daemon/herald/HeraldTest.kt#L346)):*
> ```Kotlin
>     aggregatorHerald.syncStatuses()
> 
>     assertThat(continuationTokensService.latestContinuationToken).isNotEmpty()
> ```

I think this can be changed back to checking for a specific token, as the behavior is now properly determinstic.



_Code quote:_
```Kotlin
isNotEmpty()
```

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/802)* status: 1 of 8 files reviewed, 4 unresolved discussions (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/gcloud/spanner/continuationtoken/SpannerContinuationTokensService.kt` line 45 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/802#-NLbiZZlCs1D7Ds2BlCQ:-NLc_4A85vkRmSG7bPK8:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/b5dc7a839bc31d70526f1c8cd0eb982fa4d3233e/src/main/kotlin/org/wfanet/measurement/duchy/deploy/gcloud/spanner/continuationtoken/SpannerContinuationTokensService.kt#L45)):*

We don't know for certain that all exceptions are caused by an invalid continuation token. You may want to throw a custom exception type so we can identify whether this was indeed due to the new continuation token being older than the existing one.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/gcloud/spanner/continuationtoken/SpannerContinuationTokensService.kt` line 46 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/802#-NLbikqe8k0anZDAgbIM:-NLc_4XC1hkR_cRU-hOv:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/b5dc7a839bc31d70526f1c8cd0eb982fa4d3233e/src/main/kotlin/org/wfanet/measurement/duchy/deploy/gcloud/spanner/continuationtoken/SpannerContinuationTokensService.kt#L46)):*

`INVALID_ARGUMENT` in gRPC is for when the request is malformed regardless of system state. This should be `FAILED_PRECONDITION` instead.

Note that this is different than the definition of IllegalArgumentException in Java.
</blockquote></details>

Done.
___
*[`src/test/kotlin/org/wfanet/measurement/duchy/daemon/herald/HeraldTest.kt` line 346 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/802#-NLbk20MAtjXF2BGmCA1:-NLc_6TP3Td5vySbAuwG:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/b5dc7a839bc31d70526f1c8cd0eb982fa4d3233e/src/test/kotlin/org/wfanet/measurement/duchy/daemon/herald/HeraldTest.kt#L346)):*

I think this can be changed back to checking for a specific token, as the behavior is now properly determinstic.
</blockquote></details>

Done.

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/802)* status: 1 of 8 files reviewed, 7 unresolved discussions (waiting on @renjiezh)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/Herald.kt` line 166 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/802#-NLgkO4v9EQo3q6bHBV2:-NLgkO4v9EQo3q6bHBV3:b-pikpal) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c7d5ba292356ef9daac4d8a34960670859e715ad/src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/Herald.kt#L166)):*
> ```Kotlin
>               processSystemComputation(response.computation, MAX_ATTEMPTS)
>               continuationTokenManager.markTokenProcessed(response.continuationToken)
>             } catch (e: StatusException) {
> ```

This should be at the lowest level, i.e. around the RPC itself.



_Code quote:_
```Kotlin
```
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/Herald.kt` line 168 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/802#-NLgklHo-T7YVR2BvVa7:-NLgklHo-T7YVR2BvVa8:b45h) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c7d5ba292356ef9daac4d8a34960670859e715ad/src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/Herald.kt#L168)):*
> ```Kotlin
>             } catch (e: StatusException) {
>               if (e.status.code == Status.FAILED_PRECONDITION.code) {
>                 logger.warning("Failure happened during process computation. $e")
> ```





_Suggestion:_
```Kotlin
logger.log(Level.WARNING, e) { "Failure happened during process computation" }
```
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/Herald.kt` line 170 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/802#-NLgkdAt1OPoqUoUqBYG:-NLgkdAt1OPoqUoUqBYH:bsdd75) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c7d5ba292356ef9daac4d8a34960670859e715ad/src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/Herald.kt#L170)):*
> ```Kotlin
>                 logger.warning("Failure happened during process computation. $e")
>               } else {
>                 throw e
> ```

Don't re-throw StatusException. We want to limit StatusException to just the RPC. Wrap this in another exception type.



_Code quote:_
```Kotlin
throw e
```

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/802)* status: 1 of 8 files reviewed, 5 unresolved discussions (waiting on @renjiezh)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/gcloud/spanner/continuationtoken/SpannerContinuationTokensService.kt` line 50 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/802#-NLgl60M1BRwjswpYth1:-NLgl60M1BRwjswpYth2:bbhx5ez) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c7d5ba292356ef9daac4d8a34960670859e715ad/src/main/kotlin/org/wfanet/measurement/duchy/deploy/gcloud/spanner/continuationtoken/SpannerContinuationTokensService.kt#L50)):*
> ```Kotlin
>         is InvalidContinuationToken ->
>           failGrpc(Status.FAILED_PRECONDITION) { e.message ?: "Invalid continuation token." }
>         is InvalidProtocolBufferException ->
> ```

Don't catch the base exception when there's no shared code between exception types. Instead, use separate catch block  for each exception type.



_Code quote:_
```Kotlin
    } catch (e: Exception) {
      when (e) {
        is InvalidContinuationToken ->
          failGrpc(Status.FAILED_PRECONDITION) { e.message ?: "Invalid continuation token." }
        is InvalidProtocolBufferException ->
          failGrpc(Status.INVALID_ARGUMENT) { e.message ?: "Malformed continuation token." }
```

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/802)* status: 1 of 8 files reviewed, 5 unresolved discussions (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/Herald.kt` line 166 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/802#-NLgkO4v9EQo3q6bHBV2:-NM-xoEhEP9YafNSfYVt:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c7d5ba292356ef9daac4d8a34960670859e715ad/src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/Herald.kt#L166)):*

This should be at the lowest level, i.e. around the RPC itself.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/Herald.kt` line 168 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/802#-NLgklHo-T7YVR2BvVa7:-NM-wuGB5hE9GRdWe3T0:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c7d5ba292356ef9daac4d8a34960670859e715ad/src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/Herald.kt#L168)):*
> ```Kotlin
>             } catch (e: StatusException) {
>               if (e.status.code == Status.FAILED_PRECONDITION.code) {
>                 logger.warning("Failure happened during process computation. $e")
> ```

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/Herald.kt` line 170 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/802#-NLgkdAt1OPoqUoUqBYG:-NM-wtuNB5lqgsGJ_ZGd:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c7d5ba292356ef9daac4d8a34960670859e715ad/src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/Herald.kt#L170)):*

Don't re-throw StatusException. We want to limit StatusException to just the RPC. Wrap this in another exception type.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/gcloud/spanner/continuationtoken/SpannerContinuationTokensService.kt` line 50 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/802#-NLgl60M1BRwjswpYth1:-NM-yIW_34NYcnJwNh4D:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c7d5ba292356ef9daac4d8a34960670859e715ad/src/main/kotlin/org/wfanet/measurement/duchy/deploy/gcloud/spanner/continuationtoken/SpannerContinuationTokensService.kt#L50)):*

Don't catch the base exception when there's no shared code between exception types. Instead, use separate catch block  for each exception type.
</blockquote></details>

Done.

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/802)* status: 1 of 9 files reviewed, 5 unresolved discussions (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/Herald.kt` line 166 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/802#-NLgkO4v9EQo3q6bHBV2:-NM03BKA0XcV1x5lZ5jO:bs7wumx) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c7d5ba292356ef9daac4d8a34960670859e715ad/src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/Herald.kt#L166)):*

Done.
</blockquote></details>

Done.

SanjayVas (changes requested):
> Reviewed 4 of 7 files at r3, 1 of 2 files at r4, 2 of 2 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/802)* status: 8 of 9 files reviewed, 7 unresolved discussions (waiting on @renjiezh)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/ContinuationTokenManager.kt` line 106 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/802#-NM5PpNt8qtPo5kh8JzM:-NM5PpNt8qtPo5kh8JzN:b-htcf9v) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/bc0cb736b034b176c02ee0cc60f56b5540aeb0d8/src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/ContinuationTokenManager.kt#L106)):*
> ```Kotlin
>       continuationTokenClient.setContinuationToken(setRequest)
>     } catch (e: StatusException) {
>       logger.log(Level.WARNING, e) { "Failure happened during setContinuationToken" }
> ```

Don't blindly log all StatusExceptions as WARNING. Check the exception info (e.g. status code) to determine what we want to do with it. E.g. is it one that we want to log as warning vs. log as severe vs. wrap and re-throw.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/gcloud/spanner/continuationtoken/SetContinuationToken.kt` line 36 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/802#-NM5RBmXDOkeIkr9dfTU:-NM5RBmY8csgz_D1gSMO:b-1zi0wo) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/bc0cb736b034b176c02ee0cc60f56b5540aeb0d8/src/main/kotlin/org/wfanet/measurement/duchy/deploy/gcloud/spanner/continuationtoken/SetContinuationToken.kt#L36)):*
> ```Kotlin
>           ?: StreamActiveComputationsContinuationToken.getDefaultInstance()
> 
>       if (
> ```

nit: rewrite to avoid a layer of nesting

```
if (condition that throws exception) {
  throw exception
}
do normal thing
```
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/gcloud/spanner/continuationtoken/SetContinuationToken.kt` line 55 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/802#-NM5SKVG2OqaWDNGRrjF:-NM5SKVG2OqaWDNGRrjG:b-461g8k) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/bc0cb736b034b176c02ee0cc60f56b5540aeb0d8/src/main/kotlin/org/wfanet/measurement/duchy/deploy/gcloud/spanner/continuationtoken/SetContinuationToken.kt#L55)):*
> ```Kotlin
>   }
> 
>   private fun String.decode(): StreamActiveComputationsContinuationToken =
> ```

nit: prefer not to add extensions to String that don't apply to all strings, or at least be specific enough in the extension method name what kind of string is expected.

e.g. `fun decodeContinuationToken(encoded: String)` or `fun String.decodeContinuationToken()`

Going a step further would be to use a wrapper data class that we use ASAP so the type system can differentiate from a String (might be overkill in this case).

```kotlin
data class EncodedContinuationToken(val value: String) {
  val decoded: StreamActiveComputationsContinuationToken = // eagerly decode
}
```
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/gcloud/spanner/continuationtoken/SetContinuationToken.kt` line 59 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/802#-NM5Tyam1JGhSfVhRPCU:-NM5Tyam1JGhSfVhRPCV:b-4i27u6) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/bc0cb736b034b176c02ee0cc60f56b5540aeb0d8/src/main/kotlin/org/wfanet/measurement/duchy/deploy/gcloud/spanner/continuationtoken/SetContinuationToken.kt#L59)):*
> ```Kotlin
> }
> 
> class InvalidContinuationToken(message: String) : Exception(message) {}
> ```

By convention, all Exception class names end in `Exception`.

Additionally, I think you're using this exception specifically for the case where the new continuation token is older than the existing one rather then cases where the token itself is invalid (e.g. malformed) so you probably want a different name.



_Code quote:_
```Kotlin
InvalidContinuationToken
```
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/gcloud/spanner/continuationtoken/SpannerContinuationTokensService.kt` line 50 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/802#-NM5UTB39eL2nkouZgEW:-NM5UTB39eL2nkouZgEX:b-kg9oa6) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/bc0cb736b034b176c02ee0cc60f56b5540aeb0d8/src/main/kotlin/org/wfanet/measurement/duchy/deploy/gcloud/spanner/continuationtoken/SpannerContinuationTokensService.kt#L50)):*
> ```Kotlin
>     } catch (e: InvalidProtocolBufferException) {
>       failGrpc(Status.INVALID_ARGUMENT) { e.message ?: "Malformed continuation token." }
> ```

1. This isn't necessary as the gRPC framework will wrap anything that isn't a StatusException/StatusRuntimeException in an UNKNOWN Status for you.
2. It's not safe to expose the message for all exceptions, as it may contain sensitive data or internals that we don't want to leak. We only pass on the exception message for cases where we know it's suitable for the user of this API.



_Code quote:_
```Kotlin
    } catch (e: Exception) {
      failGrpc(Status.UNKNOWN) { e.message ?: "Error during setContinuationToken." }
    }
```
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/service/internal/testing/InMemoryContinuationTokensService.kt` line 29 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/802#-NM5VAZdBNulnpIVxcTG:-NM5VAZdBNulnpIVxcTH:btbo9co) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/bc0cb736b034b176c02ee0cc60f56b5540aeb0d8/src/main/kotlin/org/wfanet/measurement/duchy/service/internal/testing/InMemoryContinuationTokensService.kt#L29)):*
> ```Kotlin
>  * In memory continuation token service for testing purpose.
>  *
>  * Tokens are integers in string form standing for the time order. Larger value means newer token.
> ```

I don't think the testing service implementation should change anything with the API contract. This should still speak the same token format.



_Code quote:_
```Kotlin
Tokens are integers in string form standing for the time order.
```

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/802)* status: 8 of 9 files reviewed, 7 unresolved discussions (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/service/internal/testing/InMemoryContinuationTokensService.kt` line 29 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/802#-NM5VAZdBNulnpIVxcTG:-NMV-e80COexW0BIElrD:b8yf1cx) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/bc0cb736b034b176c02ee0cc60f56b5540aeb0d8/src/main/kotlin/org/wfanet/measurement/duchy/service/internal/testing/InMemoryContinuationTokensService.kt#L29)):*

I don't think the testing service implementation should change anything with the API contract. This should still speak the same token format.
</blockquote></details>

Hi sanjay, given the current structure of HeraldTest, I found it difficult to use real continuationToken (instead of a random simple string) to write assertions. Because the continuationToken is generated during streaming mock, the timestamp of the continuation token is not available in assertion block. Either we only verify that the last continuation token is not empty, or bypassing the timestamp using such different but simpler format of token. Do you have any idea to solve this?

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/802)* status: 8 of 9 files reviewed, 7 unresolved discussions (waiting on @renjiezh)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/service/internal/testing/InMemoryContinuationTokensService.kt` line 29 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/802#-NM5VAZdBNulnpIVxcTG:-NMV8xhG5tUOD6qRZaSn:b-ghnr0p) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/bc0cb736b034b176c02ee0cc60f56b5540aeb0d8/src/main/kotlin/org/wfanet/measurement/duchy/service/internal/testing/InMemoryContinuationTokensService.kt#L29)):*

Hi sanjay, given the current structure of HeraldTest, I found it difficult to use real continuationToken (instead of a random simple string) to write assertions. Because the continuationToken is generated during streaming mock, the timestamp of the continuation token is not available in assertion block. Either we only verify that the last continuation token is not empty, or bypassing the timestamp using such different but simpler format of token. Do you have any idea to solve this? 
</blockquote></details>

Without going into too much detail on this case specifically, let's explore the test strategy w.r.t. ContinuationTokenManager and HeraldTest:

Is ContinuationTokenManager an implementation detail of HeraldTest, or something that warrants testing its behavior directly? It looks like it's the latter, given that it's a separate class with its own tests. This means that you can theoretically use a test double in its place for the HeraldTest, providing that your test double behaves sufficiently closely to the real thing as far as the Herald is concerned.

Would using a mock ContinuationTokenManager for HeraldTest help you here?

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/802)* status: 8 of 9 files reviewed, 7 unresolved discussions (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/service/internal/testing/InMemoryContinuationTokensService.kt` line 29 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/802#-NM5VAZdBNulnpIVxcTG:-NMVnXOp7bTF--pBEqkX:bdywgcw) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/bc0cb736b034b176c02ee0cc60f56b5540aeb0d8/src/main/kotlin/org/wfanet/measurement/duchy/service/internal/testing/InMemoryContinuationTokensService.kt#L29)):*

Without going into too much detail on this case specifically, let's explore the test strategy w.r.t. ContinuationTokenManager and HeraldTest:

Is ContinuationTokenManager an implementation detail of HeraldTest, or something that warrants testing its behavior directly? It looks like it's the latter, given that it's a separate class with its own tests. This means that you can theoretically use a test double in its place for the HeraldTest, providing that your test double behaves sufficiently closely to the real thing as far as the Herald is concerned.

Would using a mock ContinuationTokenManager for HeraldTest help you here?
</blockquote></details>

The Herald construction takes the `continuationTokenClient` as the parameter so a mock ContinuationTokenManager cannot be used directly.

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/802)* status: 8 of 9 files reviewed, 7 unresolved discussions (waiting on @renjiezh)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/service/internal/testing/InMemoryContinuationTokensService.kt` line 29 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/802#-NM5VAZdBNulnpIVxcTG:-NMVoUbJ3ft1RQMsTgxT:bbcc6c1) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/bc0cb736b034b176c02ee0cc60f56b5540aeb0d8/src/main/kotlin/org/wfanet/measurement/duchy/service/internal/testing/InMemoryContinuationTokensService.kt#L29)):*

The Herald construction takes the `continuationTokenClient` as the parameter so a mock ContinuationTokenManager cannot be used directly.
</blockquote></details>

You can modify the constructor to support dependency injection at the correct level (i.e. ContinuationTokenManager instead of continuationTokenClient).

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/802)* status: 8 of 9 files reviewed, 7 unresolved discussions (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/service/internal/testing/InMemoryContinuationTokensService.kt` line 29 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/802#-NM5VAZdBNulnpIVxcTG:-NMVorJQBbyFDJWHOsgS:b-wv2qoq) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/bc0cb736b034b176c02ee0cc60f56b5540aeb0d8/src/main/kotlin/org/wfanet/measurement/duchy/service/internal/testing/InMemoryContinuationTokensService.kt#L29)):*

You can modify the constructor to support dependency injection at the correct level (i.e. ContinuationTokenManager instead of continuationTokenClient).
</blockquote></details>

Sounds good. Thanks for the suggestion!

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/802)* status: 3 of 11 files reviewed, 5 unresolved discussions (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/ContinuationTokenManager.kt` line 106 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/802#-NM5PpNt8qtPo5kh8JzM:-NM_USXm6ovbtc2D1pSW:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/bc0cb736b034b176c02ee0cc60f56b5540aeb0d8/src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/ContinuationTokenManager.kt#L106)):*

Don't blindly log all StatusExceptions as WARNING. Check the exception info (e.g. status code) to determine what we want to do with it. E.g. is it one that we want to log as warning vs. log as severe vs. wrap and re-throw.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/gcloud/spanner/continuationtoken/SetContinuationToken.kt` line 59 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/802#-NM5Tyam1JGhSfVhRPCU:-NM_UcjL1dICtwZwea7m:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/bc0cb736b034b176c02ee0cc60f56b5540aeb0d8/src/main/kotlin/org/wfanet/measurement/duchy/deploy/gcloud/spanner/continuationtoken/SetContinuationToken.kt#L59)):*

By convention, all Exception class names end in `Exception`.

Additionally, I think you're using this exception specifically for the case where the new continuation token is older than the existing one rather then cases where the token itself is invalid (e.g. malformed) so you probably want a different name.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/gcloud/spanner/continuationtoken/SpannerContinuationTokensService.kt` line 50 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/802#-NM5UTB39eL2nkouZgEW:-NM_Ue5e889jSVIKrGiO:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/bc0cb736b034b176c02ee0cc60f56b5540aeb0d8/src/main/kotlin/org/wfanet/measurement/duchy/deploy/gcloud/spanner/continuationtoken/SpannerContinuationTokensService.kt#L50)):*

1. This isn't necessary as the gRPC framework will wrap anything that isn't a StatusException/StatusRuntimeException in an UNKNOWN Status for you.
2. It's not safe to expose the message for all exceptions, as it may contain sensitive data or internals that we don't want to leak. We only pass on the exception message for cases where we know it's suitable for the user of this API.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/service/internal/testing/InMemoryContinuationTokensService.kt` line 29 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/802#-NM5VAZdBNulnpIVxcTG:-NM_Uj3OAZcpIM9WxzjV:b-wu07q9) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/bc0cb736b034b176c02ee0cc60f56b5540aeb0d8/src/main/kotlin/org/wfanet/measurement/duchy/service/internal/testing/InMemoryContinuationTokensService.kt#L29)):*

Sounds good. Thanks for the suggestion!
</blockquote></details>

Uses ContinutationTokenService mock for HeraldTest.

SanjayVas (changes requested):
> Reviewed 8 of 8 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/802)* status: all files reviewed, 3 unresolved discussions (waiting on @renjiezh)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/ContinuationTokenManager.kt` line 110 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/802#-NMeN9DB5g0i_XlEZ5Kg:-NMeN9DB5g0i_XlEZ5Kh:bi8vjzu) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e14d90db927f2c3bbcf9427050e602704d65b4a9/src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/ContinuationTokenManager.kt#L110)):*
> ```Kotlin
>         logger.log(Level.WARNING, e) { "Failure happened during setContinuationToken" }
>       } else {
>         throw e
> ```

Don't re-throw StatusException. Wrap it in another exception type.



_Code quote:_
```Kotlin
throw e
```
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/gcloud/spanner/continuationtoken/SetContinuationToken.kt` line 29 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/802#-NMeNfut2I164s3ZyHIx:-NMeNfut2I164s3ZyHIy:b-2k0thu) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e14d90db927f2c3bbcf9427050e602704d65b4a9/src/main/kotlin/org/wfanet/measurement/duchy/deploy/gcloud/spanner/continuationtoken/SetContinuationToken.kt#L29)):*
> ```Kotlin
>     databaseClient.readWriteTransaction().execute { ctx ->
>       val newContinuationToken = decodeContinuationToken(continuationToken)
>       val oldContinuationTokenString =
> ```

Won't this throw an error if there's no old continuation token? It will attempt to parse empty bytes as a `StreamActiveComputationsContinuationToken`.

Maybe something like this would be clearer:

```kotlin
val oldContinuationToken: StreamActiveComputationsContinuationToken? = ctx
  .readRow("HeraldContinuationTokens", Key.of(true), listOf("ContinuationToken"))
  ?.getString("ContinuationToken")
  ?.let { decodeContinuationToken(it) }
if (oldContinuationToken != null) {
  // Comparison check
}
```



_Code quote:_
```Kotlin
      val oldContinuationTokenString =
        ctx
          .readRow("HeraldContinuationTokens", Key.of(true), listOf("ContinuationToken"))
          ?.getString("ContinuationToken")
          ?: ""
      val oldContinuationToken = decodeContinuationToken(oldContinuationTokenString)
```
___
*[`src/test/kotlin/org/wfanet/measurement/duchy/daemon/herald/HeraldTest.kt` line 862 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/802#-NMeR7DS-Ioll6dFFAAW:-NMeR7DS-Ioll6dFFAAX:biay8i2) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e14d90db927f2c3bbcf9427050e602704d65b4a9/src/test/kotlin/org/wfanet/measurement/duchy/daemon/herald/HeraldTest.kt#L862)):*
> ```Kotlin
>   /** Verify that SetContinuationToken requests contains the expected last token */
>   private fun verifyLastContinuationToken(lastToken: String) {
>     val setContinuationTokenCaptor: KArgumentCaptor<SetContinuationTokenRequest> = argumentCaptor()
> ```

The use of the phrase "last token" makes it sound like this is verifying the last (most recent) call to setContinuationToken. Instead this is verifying that any call was made using this token. Maybe call this `verifySetContinuationTokenCalledWith` or something that makes this more clear.

Taking a step back however, I think this can be done without argument captors and therefore be simple enough that you don't need a separate function.

To verify that the method was never called:
```kotlin
verifyBlocking(continuationTokensService, never()) {setContinuationToken(any())}
```

To verify it was called with a specific value at least once:
```kotlin
verifyBlocking(continuationTokensService, atLeastOnce()) {
  setContinuationToken(eq(setContinuationTokenRequest { this.token = token }))
}
```



_Code quote:_
```Kotlin
  /** Verify that SetContinuationToken requests contains the expected last token */
  private fun verifyLastContinuationToken(lastToken: String) {
```

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/802)* status: 8 of 11 files reviewed, 3 unresolved discussions (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/ContinuationTokenManager.kt` line 110 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/802#-NMeN9DB5g0i_XlEZ5Kg:-NMfmRDc44Hk7JLdJ9O9:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e14d90db927f2c3bbcf9427050e602704d65b4a9/src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/ContinuationTokenManager.kt#L110)):*

Don't re-throw StatusException. Wrap it in another exception type.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/gcloud/spanner/continuationtoken/SetContinuationToken.kt` line 29 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/802#-NMeNfut2I164s3ZyHIx:-NMfmS4QCcj2IhceuHsD:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e14d90db927f2c3bbcf9427050e602704d65b4a9/src/main/kotlin/org/wfanet/measurement/duchy/deploy/gcloud/spanner/continuationtoken/SetContinuationToken.kt#L29)):*

Won't this throw an error if there's no old continuation token? It will attempt to parse empty bytes as a `StreamActiveComputationsContinuationToken`.

Maybe something like this would be clearer:

```kotlin
val oldContinuationToken: StreamActiveComputationsContinuationToken? = ctx
  .readRow("HeraldContinuationTokens", Key.of(true), listOf("ContinuationToken"))
  ?.getString("ContinuationToken")
  ?.let { decodeContinuationToken(it) }
if (oldContinuationToken != null) {
  // Comparison check
}
```
</blockquote></details>

Done.
___
*[`src/test/kotlin/org/wfanet/measurement/duchy/daemon/herald/HeraldTest.kt` line 862 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/802#-NMeR7DS-Ioll6dFFAAW:-NMfju4L3i3QRQ1UPdng:br5qcgr) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e14d90db927f2c3bbcf9427050e602704d65b4a9/src/test/kotlin/org/wfanet/measurement/duchy/daemon/herald/HeraldTest.kt#L862)):*

The use of the phrase "last token" makes it sound like this is verifying the last (most recent) call to setContinuationToken. Instead this is verifying that any call was made using this token. Maybe call this `verifySetContinuationTokenCalledWith` or something that makes this more clear.

Taking a step back however, I think this can be done without argument captors and therefore be simple enough that you don't need a separate function.

To verify that the method was never called:
```kotlin
verifyBlocking(continuationTokensService, never()) {setContinuationToken(any())}
```

To verify it was called with a specific value at least once:
```kotlin
verifyBlocking(continuationTokensService, atLeastOnce()) {
  setContinuationToken(eq(setContinuationTokenRequest { this.token = token }))
}
```
</blockquote></details>

Great to know that! Thanks for the suggestion.

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/802)* status: 8 of 11 files reviewed, 3 unresolved discussions (waiting on @SanjayVas)

___
*[`src/test/kotlin/org/wfanet/measurement/duchy/daemon/herald/HeraldTest.kt` line 862 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/802#-NMeR7DS-Ioll6dFFAAW:-NMfmYh_-RwcvTGcf5Nf:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e14d90db927f2c3bbcf9427050e602704d65b4a9/src/test/kotlin/org/wfanet/measurement/duchy/daemon/herald/HeraldTest.kt#L862)):*

Great to know that! Thanks for the suggestion.
</blockquote></details>

Done.

SanjayVas (approved):
> Reviewed 3 of 3 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/802)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @renjiezh)

Issue #816, opened 2023-01-18
Frequency Histogram report results return more results than expected


mariolamassaavedra:
**Describe the bug**
Different report definitions result in the results coming back with extra columns. Some of the test cases we have executed so far are:

Frequency histogram, LHS union RHS Increment = 24h, intervalCount = 1, maxFrequency = 1
Expected result single number for the single interval 
Actual result 7 columns 
Attachment: Histogram_freq1.txt

Frequency histogram, LHS union RHS  start time= 21-07-2022, end time = 28-07-2022, maxFrequency = 5
Expected result single column, 5 results 
Actual result single column 6 results
Attachment: Histogram_freq5.txt

Frequency histogram, LHS union RHS (two different reporting sets) Increment = 7days, intervalCount = 9, maxFrequency = 5
Expected result single column, 45 results 
Actual result single column 62 results
Attachment: Histogram_9_5.txt

**Steps to reproduce**
1. Create a report definition using the scenarios described above
2. Get report result
3. Validate results

**Component(s) affected**
Reporting Server 

**Version**
37bf941150cead85842b46d9726186ee52f316f3

**Environment**
Origin Dev Environment

**Additional context**
[histogram_9_5.txt](https://github.com/world-federation-of-advertisers/cross-media-measurement/files/10449449/histogram_9_5.txt)
[histogram_freq5.txt](https://github.com/world-federation-of-advertisers/cross-media-measurement/files/10449450/histogram_freq5.txt)
[histogram_freq1.txt](https://github.com/world-federation-of-advertisers/cross-media-measurement/files/10449451/histogram_freq1.txt)

duliomatos1:
@mariolamassaavedra, thanks for clarifying. @tristanvuong2021 can you confirm if this is a bug in the reporting server? They are using a fairly recent version from 5 days ago.

SanjayVas:
@kungfucraig Could this be related to the comment you made on the new Reporting API about documenting that max_frequency_per_user is at a per-EDP level rather than a cutoff of for the entire result?

tristanvuong2021:
I opened the above PR, but I am not sure if this solves the issue because I had issues replicating it with the report request parameters. @mariolamassaavedra could you test it with the same requests to see if it fixes it.

mariolamassaavedra:
Apologies still not getting GitHub emails for some reason ... yes we can test it. Will ask the team to take a look tomorrow

PR #820, opened 2023-01-20
Aggregate reporting frequency results when measurement results have more than the max frequency per user

tristanvuong2021:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/820)


riemanli (commented):
Hi Tristan, could you explain why the result can have frequencies more than the cut-off frequency?

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/820#-:-NMFV7n71Trd8VW_N1WH:b-gkxlzk)* status: 0 of 2 files reviewed, all discussions resolved

tristanvuong2021:
This issue came up https://github.com/world-federation-of-advertisers/cross-media-measurement/issues/816. The only thing I could think of is that it may be possible the kingdom returns more than asked for. I will have to see if this solves the issue.

kungfucraig (commented):
On the MPC side perhaps the max freq is not making it's way into the protocol config? Or on the direct measurement side, the EDP is not honoring it?

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/820#-:-NMVqrfTAeDLfY22dBJ5:binf0wa)* status: 0 of 2 files reviewed, all discussions resolved

PR #826, opened 2023-01-25
Fix Bazel build output not being saved to Actions cache.

SanjayVas:
There was a previous misconfiguration in an attempt to get the cache to save only if the correctness test passes, but the action is configured to only save when the all targets wildcard is used (`//...`).

- - -
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/826)

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -34,7 +34,6 @@ jobs:
         uses: world-federation-of-advertisers/actions/bazel-build-test@v1
         with:
           cache-version: 1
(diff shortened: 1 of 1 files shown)


stevenwarejones (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/826)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @SanjayVas)

PR #831, opened 2023-01-26
Add ComputationsCleaner to periodically delete outdated Computations

renjiezh:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/831)


SanjayVas (changes requested):
> Reviewed 1 of 9 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/831)* status: 1 of 9 files reviewed, 1 unresolved discussion (waiting on @renjiezh)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/server/ComputationsServer.kt` line 96 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/831#-NNCvVOp4yQsTNbIQlhQ:-NNCvVOp4yQsTNbIQlhR:buzds9k) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5b1c6a0b241b3f9a110e0b0ec4ca513d1055280e/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/server/ComputationsServer.kt#L96)):*
> ```Kotlin
> 
>     val computationsCleaner =
>       ComputationsCleaner(
> ```

The intent was to have the cleaner job run as a [Kubernetes CronJob](https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/), not as part of the Computations server. It should be a separate binary.

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/831)* status: 1 of 9 files reviewed, 1 unresolved discussion (waiting on @renjiezh and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/server/ComputationsServer.kt` line 96 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/831#-NNCvVOp4yQsTNbIQlhQ:-NNDMjBJ8iLsWjYBkd6X:b-ds0djt) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5b1c6a0b241b3f9a110e0b0ec4ca513d1055280e/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/server/ComputationsServer.kt#L96)):*

The intent was to have the cleaner job run as a [Kubernetes CronJob](https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/), not as part of the Computations server. It should be a separate binary.
</blockquote></details>

Got it. Working on it.

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/831)* status: 0 of 18 files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/server/ComputationsServer.kt` line 96 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/831#-NNCvVOp4yQsTNbIQlhQ:-NNNaCifDagTgHcteBkJ:bs02h1q) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5b1c6a0b241b3f9a110e0b0ec4ca513d1055280e/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/server/ComputationsServer.kt#L96)):*

Got it. Working on it.
</blockquote></details>

Done.

Tested on KiND. I will try it on GCP once the Kingdom is back alive.

SanjayVas (changes requested):
> Reviewed 5 of 9 files at r1, 13 of 13 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/831)* status: all files reviewed, 7 unresolved discussions (waiting on @renjiezh)

___
*[`src/main/k8s/base.cue` line 559 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/831#-NNOIctw95jvH26Y345N:-NNOIctw95jvH26Y345O:b-nqzh9j) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0bb1ce9aece29eccb6d4652173d2b1d9174f93cb/src/main/k8s/base.cue#L559)):*
> ```CUE
> 						}
> 						_containers: "\(_name)-container": _container
> 						restartPolicy: restartPolicy | *"OnFailure"
> ```

I think we may want cron jobs not to restart, and just wait for the next scheduled run to try again.



_Code quote:_
```CUE
*"OnFailure"
```
___
*[`src/main/k8s/local/duchies.cue` line 79 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/831#-NNOIvOr8JweHEbsCWea:-NNOIvOr8JweHEbsCWeb:barpyzs) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0bb1ce9aece29eccb6d4652173d2b1d9174f93cb/src/main/k8s/local/duchies.cue#L79)):*
> ```CUE
> 			"spanner-computations-server":      "bazel/src/main/kotlin/org/wfanet/measurement/duchy/deploy/gcloud/server:forwarded_storage_spanner_computations_server_image"
> 			"update-duchy-schema":              "bazel/src/main/kotlin/org/wfanet/measurement/duchy/deploy/gcloud/spanner/tools:update_schema_image"
> 			"computations-cleaner":             "bazel/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/job:computations_cleaner_cron_job_image"
> ```

nit: I don't think you need `cron_job` in the image name



_Code quote:_
```CUE
cron_job_
```
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/db/computation/ComputationsDatabase.kt` line 63 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/831#-NNOJ1st2lqQIhoLt_Rm:-NNOJ1st2lqQIhoLt_Rn:b5hu289) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0bb1ce9aece29eccb6d4652173d2b1d9174f93cb/src/main/kotlin/org/wfanet/measurement/duchy/db/computation/ComputationsDatabase.kt#L63)):*
> ```Kotlin
> 
>   /** Gets list of globalIds of all outdated Computations */
>   suspend fun readOutdatedComputationGlobalIds(ttlSecond: Long): List<String>
> ```

nit: maybe pass this as a "before" timestamp instead? The TTL is more of a concept for the job, but the DB layer doesn't need to know that.



_Code quote:_
```Kotlin
ttlSecond
```
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/job/ComputationsCleanerCronJob.kt` line 44 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/831#-NNOJxE80Q70zJdm5NMS:-NNOJxE80Q70zJdm5NMT:b84lds3) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0bb1ce9aece29eccb6d4652173d2b1d9174f93cb/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/job/ComputationsCleanerCronJob.kt#L44)):*
> ```Kotlin
>     names = ["--computations-ttl-days"],
>     defaultValue = "90",
>     description = ["TTL in days of Computations in database and storage."],
> ```

nit: expand this abbreviation, as readers might not be familiar with it. Also add more detail.

e.g. "Time to live (TTL) for a Computation. Computations that have been in a terminal state longer than this value will be cleaned."



_Code quote:_
```Kotlin
TTL
```
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/job/ComputationsCleanerCronJob.kt` line 47 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/831#-NNOJcwv0rp4MF7an03t:-NNOJcwv0rp4MF7an03u:b-vp1vk5) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0bb1ce9aece29eccb6d4652173d2b1d9174f93cb/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/job/ComputationsCleanerCronJob.kt#L47)):*
> ```Kotlin
>     required = false
>   )
>   var computationsTtlDays by Delegates.notNull<Long>()
> ```

Use a Duration instead of seconds. This way we can parse from something human-readable like "90d".



_Code quote:_
```Kotlin
Delegates.notNull<Long>()
```
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/service/internal/computations/ComputationsService.kt` line 155 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/831#-NNOKy7v9Ii8SDbgn3RB:-NNOKy7w4bGsMLoLq98w:bjtp1tn) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0bb1ce9aece29eccb6d4652173d2b1d9174f93cb/src/main/kotlin/org/wfanet/measurement/duchy/service/internal/computations/ComputationsService.kt#L155)):*
> ```Kotlin
> 
>   override suspend fun deleteOutdatedComputations(
>     request: DeleteOutdatedComputationsRequest
> ```

nit: instead of referring to "outdated" maybe "terminal" or "completed", as that's the operative part that's not explicitly in the request. i.e. the request indicates the update time, but the method always operates on computations that are in a terminal state.



_Code quote:_
```Kotlin
Outdated
```
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/service/internal/computations/ComputationsService.kt` line 159 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/831#-NNOLNQZ-nJoXEj_SMJ9:-NNOLNQZ-nJoXEj_SMJA:baf5uf3) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0bb1ce9aece29eccb6d4652173d2b1d9174f93cb/src/main/kotlin/org/wfanet/measurement/duchy/service/internal/computations/ComputationsService.kt#L159)):*
> ```Kotlin
>     var deleted = 0
>     try {
>       val globalIds = computationsDatabase.readOutdatedComputationGlobalIds(request.ttlSecond)
> ```

I think you can do this as a single DELETE statement rather than a separate read and write. The `executeUpdate` method returns the number of rows affected.

Something like `DELETE FROM Computations WHERE State IN (terminal states) AND UpdateTime < @threshold`

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/831)* status: all files reviewed, 7 unresolved discussions (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/service/internal/computations/ComputationsService.kt` line 159 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/831#-NNOLNQZ-nJoXEj_SMJ9:-NOzejCF1TNs2nKUMv4S:bljycth) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0bb1ce9aece29eccb6d4652173d2b1d9174f93cb/src/main/kotlin/org/wfanet/measurement/duchy/service/internal/computations/ComputationsService.kt#L159)):*

I think you can do this as a single DELETE statement rather than a separate read and write. The `executeUpdate` method returns the number of rows affected.

Something like `DELETE FROM Computations WHERE State IN (terminal states) AND UpdateTime < @threshold`

</blockquote></details>

I think one of the goals is to clean pending computations that are too old. It requires duchy to fail them at kingdom. Besides, blobs in storage also need to clean. Does batch db deletion work?

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/831)* status: 10 of 18 files reviewed, 4 unresolved discussions (waiting on @SanjayVas)

___
*[`src/main/k8s/base.cue` line 559 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/831#-NNOIctw95jvH26Y345N:-NP-St7e6hE3JGKhS0WP:bulrga3) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0bb1ce9aece29eccb6d4652173d2b1d9174f93cb/src/main/k8s/base.cue#L559)):*

I think we may want cron jobs not to restart, and just wait for the next scheduled run to try again.
</blockquote></details>

Done.
Replaced with "Never"
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/service/internal/computations/ComputationsService.kt` line 155 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/831#-NNOKy7v9Ii8SDbgn3RB:-NP-T1y86qMhU5pr_UDO:b-5i4eve) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0bb1ce9aece29eccb6d4652173d2b1d9174f93cb/src/main/kotlin/org/wfanet/measurement/duchy/service/internal/computations/ComputationsService.kt#L155)):*

nit: instead of referring to "outdated" maybe "terminal" or "completed", as that's the operative part that's not explicitly in the request. i.e. the request indicates the update time, but the method always operates on computations that are in a terminal state.
</blockquote></details>

Do we need to delete pending computations as well?
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/job/ComputationsCleanerCronJob.kt` line 47 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/831#-NNOJcwv0rp4MF7an03t:-NP-SzBg684W-TX3Zuu3:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0bb1ce9aece29eccb6d4652173d2b1d9174f93cb/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/job/ComputationsCleanerCronJob.kt#L47)):*

Use a Duration instead of seconds. This way we can parse from something human-readable like "90d".
</blockquote></details>

Done.

SanjayVas (changes requested):
> Reviewed 8 of 15 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/831)* status: 12 of 19 files reviewed, 3 unresolved discussions (waiting on @renjiezh)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/job/ComputationsCleanerJob.kt` line 45 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/831#-NP3rnKeAeWlM-9bDWjf:-NP3rnKeAeWlM-9bDWjg:b-rrwsnl) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/952b6e89d8412a2215c8d3b880efb99a8585f8ad/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/job/ComputationsCleanerJob.kt#L45)):*
> ```Kotlin
>     defaultValue = "90d",
>     description =
>       ["Time to live (TTL) for a Computation. Computations live longer than this will be cleaned."],
> ```

There are two TTLs that we care about:
1. How long a computation can be in a pending (non-terminal) state before being cancelled (I believe Joe's work should handle this at the Kingdom, but theoretically we could handle it from the Duchy side too by failing at the Kingdom)
2. How long a computation can be in a completed (terminal) state before being deleted.

Make it clear which one this is.



_Code quote:_
```Kotlin
Computations live longer than this will be cleaned
```

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/831)* status: 12 of 19 files reviewed, 3 unresolved discussions (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/job/ComputationsCleanerJob.kt` line 45 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/831#-NP3rnKeAeWlM-9bDWjf:-NP4R01C0BCd7C6Uvw6k:b-8ncz1r) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/952b6e89d8412a2215c8d3b880efb99a8585f8ad/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/job/ComputationsCleanerJob.kt#L45)):*

There are two TTLs that we care about:
1. How long a computation can be in a pending (non-terminal) state before being cancelled (I believe Joe's work should handle this at the Kingdom, but theoretically we could handle it from the Duchy side too by failing at the Kingdom)
2. How long a computation can be in a completed (terminal) state before being deleted.

Make it clear which one this is.
</blockquote></details>

Thanks for explain that. It sounds like eventually we need two TTLs to delete both pending and terminal computations. 
Do you think we need to do that for now? Or only delete terminal ones and leave pending computations to Kingdom in order to debug?

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/831)* status: 12 of 19 files reviewed, 3 unresolved discussions (waiting on @kungfucraig)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/job/ComputationsCleanerJob.kt` line 45 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/831#-NP3rnKeAeWlM-9bDWjf:-NP4x8KgC2uIjyaG7g23:b8et5ii) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/952b6e89d8412a2215c8d3b880efb99a8585f8ad/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/job/ComputationsCleanerJob.kt#L45)):*

Thanks for explain that. It sounds like eventually we need two TTLs to delete both pending and terminal computations. 
Do you think we need to do that for now? Or only delete terminal ones and leave pending computations to Kingdom in order to debug?
</blockquote></details>

@kungfucraig WDYT? Should the Duchy just let the Kingdom take care of cancelling pending computations?

I believe the Duchy already does cleanup when it finds out from the Kingdom that the computation has completed (see deleteComputationAtDuchy in the Herald).

kungfucraig (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/831)* status: 12 of 19 files reviewed, 3 unresolved discussions (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/job/ComputationsCleanerJob.kt` line 45 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/831#-NP3rnKeAeWlM-9bDWjf:-NPIqF6I-t3KjpGoXzQ8:b-pobypt) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/952b6e89d8412a2215c8d3b880efb99a8585f8ad/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/job/ComputationsCleanerJob.kt#L45)):*

@kungfucraig WDYT? Should the Duchy just let the Kingdom take care of cancelling pending computations?

I believe the Duchy already does cleanup when it finds out from the Kingdom that the computation has completed (see deleteComputationAtDuchy in the Herald).
</blockquote></details>

Yes, let's have the Kingdom be in charge of cancelling. We can see how that goes and always add additional cancel logic in the Duchy if we are deficient.

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/831)* status: 7 of 21 files reviewed, 2 unresolved discussions (waiting on @kungfucraig and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/service/internal/computations/ComputationsService.kt` line 159 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/831#-NNOLNQZ-nJoXEj_SMJ9:-NPUf0xg1j8ZOA6pWItm:b-edj3tl) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0bb1ce9aece29eccb6d4652173d2b1d9174f93cb/src/main/kotlin/org/wfanet/measurement/duchy/service/internal/computations/ComputationsService.kt#L159)):*

I think one of the goals is to clean pending computations that are too old. It requires duchy to fail them at kingdom. Besides, blobs in storage also need to clean. Does batch db deletion work?
</blockquote></details>

Now using GlobalIdQuery to query specific stages.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/job/ComputationsCleanerJob.kt` line 45 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/831#-NP3rnKeAeWlM-9bDWjf:-NPUeyis4A-m-G62BqWD:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/952b6e89d8412a2215c8d3b880efb99a8585f8ad/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/job/ComputationsCleanerJob.kt#L45)):*

Yes, let's have the Kingdom be in charge of cancelling. We can see how that goes and always add additional cancel logic in the Duchy if we are deficient.
</blockquote></details>

Done.

SanjayVas (changes requested):
> Reviewed 1 of 15 files at r3, 13 of 13 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/831)* status: all files reviewed, 8 unresolved discussions (waiting on @kungfucraig and @renjiezh)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/server/ComputationsServer.kt` line 145 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/831#-NPUyu1s2770a8VH6H75:-NPUyu1s2770a8VH6H76:bcxe32s) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/49a146249bf63a7c32670481a71d9fd1b7cefe64/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/server/ComputationsServer.kt#L145)):*
> ```Kotlin
> 
>     @set:CommandLine.Option(
>       names = ["--dry-run-batch-deletion"],
> ```

This should be a flag on the job binary, not on the server. It would translate to a field in the RPC request.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/gcloud/spanner/computation/GcpSpannerComputationsDatabaseReader.kt` line 61 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/831#-NPUzIls0h4lQ_5hUSiV:-NPUzIls0h4lQ_5hUSiW:b-awi4sn) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/49a146249bf63a7c32670481a71d9fd1b7cefe64/src/main/kotlin/org/wfanet/measurement/duchy/deploy/gcloud/spanner/computation/GcpSpannerComputationsDatabaseReader.kt#L61)):*
> ```Kotlin
>   override suspend fun readGlobalComputationIds(
>     stages: Set<ComputationStage>,
>     before: Instant?
> ```

nit: be clear what this is before



_Suggestion:_
```Kotlin
updatedBefore
```
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/gcloud/spanner/computation/GlobalIdsQuery.kt` line 32 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/831#-NPUzRrOEUf4PTEy_8tV:-NPUzRrOEUf4PTEy_8tW:b2ni8l0) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/49a146249bf63a7c32670481a71d9fd1b7cefe64/src/main/kotlin/org/wfanet/measurement/duchy/deploy/gcloud/spanner/computation/GlobalIdsQuery.kt#L32)):*
> ```Kotlin
>   filterToStages: Set<StageT>,
>   computationType: ComputationType,
>   before: Timestamp? = null
> ```

nit



_Suggestion:_
```Kotlin
updatedBefore
```
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/service/internal/computations/ComputationsService.kt` line 155 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/831#-NNOKy7v9Ii8SDbgn3RB:-NPUzY7K4xbeyUUDO04W:boi8cd9) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0bb1ce9aece29eccb6d4652173d2b1d9174f93cb/src/main/kotlin/org/wfanet/measurement/duchy/service/internal/computations/ComputationsService.kt#L155)):*

Do we need to delete pending computations as well?
</blockquote></details>

No. We want to ensure a minimum amount of time that a computation is in a terminal state prior to deletion.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/service/internal/computations/ComputationsService.kt` line 159 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/831#-NNOLNQZ-nJoXEj_SMJ9:-NPV-xmu7TU7aJXeNLXs:b4y0n2m) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0bb1ce9aece29eccb6d4652173d2b1d9174f93cb/src/main/kotlin/org/wfanet/measurement/duchy/service/internal/computations/ComputationsService.kt#L159)):*

Now using GlobalIdQuery to query specific stages.
</blockquote></details>

That doesn't really help the issue I was mentioning re: having everything be in a single transaction.

If it's a dry run, you do a SELECT. If it's not, you do a DELETE. This means either having two separate methods or passing the dry run option into the DB method.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/service/internal/computations/ComputationsService.kt` line 176 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/831#-NPV0Tsy1iW_iRkYOb-Y:-NPV0TszEm80uFflVzJA:b-msgp95) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/49a146249bf63a7c32670481a71d9fd1b7cefe64/src/main/kotlin/org/wfanet/measurement/duchy/service/internal/computations/ComputationsService.kt#L176)):*
> ```Kotlin
>         logger.log(
>           Level.INFO,
>           "Dry run batch deletion. # of Computations to delete is ${globalIds.size}"
> ```

This information needs to be returned to the caller job so it can print it. You can just return the count that would be deleted in the response. The caller should know it passed `dry_run: true` in the request and therefore this number doesn't mean anything was actually deleted.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/service/internal/computations/ComputationsService.kt` line 394 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/831#-NPV-dxH6yFQeNSh0ETL:-NPV-dxH6yFQeNSh0ETM:b-irhqrp) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/49a146249bf63a7c32670481a71d9fd1b7cefe64/src/main/kotlin/org/wfanet/measurement/duchy/service/internal/computations/ComputationsService.kt#L394)):*
> ```Kotlin
> 
>     private val MIN_COMPUTATIONS_TIME_TO_LIVE = Duration.ofDays(90)
> ```

Why do we have a 90-day minimum hard-coded? This should be purely based on configurable policy.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/job/ComputationsCleanerJob.kt` line 45 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/831#-NP3rnKeAeWlM-9bDWjf:-NPUyFtP9Ae4deXfm8FG:bosq1go) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/952b6e89d8412a2215c8d3b880efb99a8585f8ad/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/job/ComputationsCleanerJob.kt#L45)):*

Done.
</blockquote></details>

The description still isn't clear that this refers to how long the Computations are in these states as opposed to a total TTL since creation for Computations in these states.

The difference is whether we're comparing to CreateTime or UpdateTime. We want to use UpdateTime.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/job/ComputationsCleanerJob.kt` line 32 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/831#-NPUypcFBHDo5do-cd1-:-NPUypcFBHDo5do-cd10:b4a0ujy) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/49a146249bf63a7c32670481a71d9fd1b7cefe64/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/job/ComputationsCleanerJob.kt#L32)):*
> ```Kotlin
> import picocli.CommandLine
> 
> private val RPC_DEADLINE_LIMIT = Duration.ofMinutes(5)
> ```

Make this a flag.



_Code quote:_
```Kotlin
private val RPC_DEADLINE_LIMIT = Duration.ofMinutes(5)
```

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/831)* status: all files reviewed, 8 unresolved discussions (waiting on @kungfucraig and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/service/internal/computations/ComputationsService.kt` line 159 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/831#-NNOLNQZ-nJoXEj_SMJ9:-NPVR6db4ts-M4hIe8kI:b-gu24ld) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0bb1ce9aece29eccb6d4652173d2b1d9174f93cb/src/main/kotlin/org/wfanet/measurement/duchy/service/internal/computations/ComputationsService.kt#L159)):*

That doesn't really help the issue I was mentioning re: having everything be in a single transaction.

If it's a dry run, you do a SELECT. If it's not, you do a DELETE. This means either having two separate methods or passing the dry run option into the DB method.
</blockquote></details>

Regarding the implementation, as we also need to deal with blobs, it deletes Computations one by one. 
If I understand your suggestion correctly, we will have two steps: 1. delete all blobs for all outdated computations. 2. delete all entries in db. 
Which way is better in your opinion?
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/service/internal/computations/ComputationsService.kt` line 394 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/831#-NPV-dxH6yFQeNSh0ETL:-NPVSeN9DrSNse9za0Pz:b-mn0f8i) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/49a146249bf63a7c32670481a71d9fd1b7cefe64/src/main/kotlin/org/wfanet/measurement/duchy/service/internal/computations/ComputationsService.kt#L394)):*

Why do we have a 90-day minimum hard-coded? This should be purely based on configurable policy.
</blockquote></details>

I add this on SpannerComputationServer as a protection in case it receives request (accidentally from nowhere) that delete recent computations. This duration could be 30 days which makes more sense. 
Do you want me to drop it?

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/831)* status: all files reviewed, 7 unresolved discussions (waiting on @kungfucraig and @renjiezh)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/service/internal/computations/ComputationsService.kt` line 159 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/831#-NNOLNQZ-nJoXEj_SMJ9:-NPYKxtx3Qpk2AiHmHAe:bd8kce4) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0bb1ce9aece29eccb6d4652173d2b1d9174f93cb/src/main/kotlin/org/wfanet/measurement/duchy/service/internal/computations/ComputationsService.kt#L159)):*

Regarding the implementation, as we also need to deal with blobs, it deletes Computations one by one. 
If I understand your suggestion correctly, we will have two steps: 1. delete all blobs for all outdated computations. 2. delete all entries in db. 
Which way is better in your opinion?


</blockquote></details>

Ah sorry, I forgot about the blobs. This is fine then.

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/831)* status: 10 of 21 files reviewed, 5 unresolved discussions (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/server/ComputationsServer.kt` line 145 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/831#-NPUyu1s2770a8VH6H75:-NPdOsrn4gprVHsNUR4D:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/49a146249bf63a7c32670481a71d9fd1b7cefe64/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/server/ComputationsServer.kt#L145)):*

This should be a flag on the job binary, not on the server. It would translate to a field in the RPC request.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/service/internal/computations/ComputationsService.kt` line 176 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/831#-NPV0Tsy1iW_iRkYOb-Y:-NPdOyV5Cj8XlzyoFGe2:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/49a146249bf63a7c32670481a71d9fd1b7cefe64/src/main/kotlin/org/wfanet/measurement/duchy/service/internal/computations/ComputationsService.kt#L176)):*

This information needs to be returned to the caller job so it can print it. You can just return the count that would be deleted in the response. The caller should know it passed `dry_run: true` in the request and therefore this number doesn't mean anything was actually deleted.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/job/ComputationsCleanerJob.kt` line 45 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/831#-NP3rnKeAeWlM-9bDWjf:-NPdOqiZC1whaC7pmHM_:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/952b6e89d8412a2215c8d3b880efb99a8585f8ad/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/job/ComputationsCleanerJob.kt#L45)):*

The description still isn't clear that this refers to how long the Computations are in these states as opposed to a total TTL since creation for Computations in these states.

The difference is whether we're comparing to CreateTime or UpdateTime. We want to use UpdateTime.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/job/ComputationsCleanerJob.kt` line 32 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/831#-NPUypcFBHDo5do-cd1-:-NPdOmMG91pVa1F2TX4d:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/49a146249bf63a7c32670481a71d9fd1b7cefe64/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/job/ComputationsCleanerJob.kt#L32)):*

Make this a flag.
</blockquote></details>

Done.

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/831)* status: 16 of 23 files reviewed, 8 unresolved discussions (waiting on @kungfucraig and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/service/internal/computations/ComputationsService.kt` line 394 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/831#-NPV-dxH6yFQeNSh0ETL:-NPtQKAt7ovLuKRbPzsn:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/49a146249bf63a7c32670481a71d9fd1b7cefe64/src/main/kotlin/org/wfanet/measurement/duchy/service/internal/computations/ComputationsService.kt#L394)):*

Yes. The caller should be able to delete arbitrarily. The place where the policy is effectively configured is the flag to the cleaner job.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/service/internal/computations/ComputationsService.kt` line 138 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/831#-NPsnq792mwOWYjBEyJg:-NPtPO8N4myKwzaUB40H:b-g1nvea) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c6dd4b8922b25ce5652b519fa11d4d448c3c21e0/src/main/kotlin/org/wfanet/measurement/duchy/service/internal/computations/ComputationsService.kt#L138)):*

The enum type is `Status.Code`.
</blockquote></details>

Done.
I just realized either way works. But still change it as you suggested to keep our code in the same format.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/service/internal/computations/ComputationsService.kt` line 173 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/831#-NPsmd4h-fFiBXcOjBVl:-NPtetO7BtEncghC4YKN:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c6dd4b8922b25ce5652b519fa11d4d448c3c21e0/src/main/kotlin/org/wfanet/measurement/duchy/service/internal/computations/ComputationsService.kt#L173)):*

You can drop the log message. Verbose gRPC logging will include the request and response.
</blockquote></details>

Done.
___
*[`src/main/proto/wfa/measurement/internal/duchy/computations_service.proto` line 43 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/831#-NPsnF7l8qn2CHUdbfof:-NPteuJW605EzTZg48_k:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c6dd4b8922b25ce5652b519fa11d4d448c3c21e0/src/main/proto/wfa/measurement/internal/duchy/computations_service.proto#L43)):*

Document that this is non-transactional such that some Computations may have been deleted even if the response status is not `OK`.
</blockquote></details>

Done.
___
*[`src/main/proto/wfa/measurement/internal/duchy/computations_service.proto` line 44 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/831#-NPsp8JAEm2-15mflNuo:-NPteywbEIrrjtscYoiM:bx2yo8l) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c6dd4b8922b25ce5652b519fa11d4d448c3c21e0/src/main/proto/wfa/measurement/internal/duchy/computations_service.proto#L44)):*

nit: it turns out this pattern has an associated AIP: https://google.aip.dev/165

You don't technically need to follow that since this is an internal API method that does not have a corresponding public API method, but it may be nice to follow some of the language and semantics. e.g. call the method `Purge`, put the `updated_before` field inside of a `Filter` sub-message
</blockquote></details>

The following are done:
1. rename to purgeComputations.
2. In the request, use --dry-run to set force option. Instead of having parent and filter, keep using states and updated_before.
3. In the response, rename count to purge-count. Add purge-sample of globalIds for dry-run mode.
___
*[`src/main/proto/wfa/measurement/internal/duchy/computations_service.proto` line 116 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/831#-NPsqKTp5pFjlSVYeesM:-NPtic4w3Qf7CrSzhFn2:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c6dd4b8922b25ce5652b519fa11d4d448c3c21e0/src/main/proto/wfa/measurement/internal/duchy/computations_service.proto#L116)):*

Rather than the TTL, the method should take in the `updated_before` timestamp. The cleaner job can compute this timestamp using the TTL flag.
</blockquote></details>

Done.
___
*[`src/main/proto/wfa/measurement/internal/duchy/computations_service.proto` line 117 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/831#-NPsoAwMB3_hcPPPwjfB:-NPtfE099P5JIYXZLqa1:b-a073sp) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c6dd4b8922b25ce5652b519fa11d4d448c3c21e0/src/main/proto/wfa/measurement/internal/duchy/computations_service.proto#L117)):*

I'm not sure this is actually used, as we have logic to determine if the Computation is in a terminal state.

This field can just be dropped.
</blockquote></details>

I kept it to make it flexible if we want to delete more than terminal computations.

SanjayVas (changes requested):
> Reviewed 7 of 7 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/831)* status: all files reviewed, 2 unresolved discussions (waiting on @renjiezh)

*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/831#-NPxRq992lzk8imnillc:-NPxRq992lzk8imnilld:b-eemgvx) (no related file):*
FWIW, I prefer the pattern of using separate List and (Batch)Delete methods. Since we're only concerned with terminal states, we arguably don't even need an etag mechanism. See the similar comment in #872

___
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/service/internal/computations/ComputationsCleaner.kt` line 46 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/831#-NPxRSKUEI2ZLqYmSEYp:-NPxRSKUEI2ZLqYmSEYq:b-3u0ei8) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/2fcf6c777fd4f867fbc3cfcce19933786082b264/src/main/kotlin/org/wfanet/measurement/duchy/service/internal/computations/ComputationsCleaner.kt#L46)):*
> ```Kotlin
>           stages += Stage.COMPLETE.toProtocolStage()
>           force = dryRun
>         }
> ```

`force` and `dry_run` are opposites



_Code quote:_
```Kotlin
force = dryRun
```

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/831)* status: 21 of 23 files reviewed, 2 unresolved discussions (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/service/internal/computations/ComputationsCleaner.kt` line 46 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/831#-NPxRSKUEI2ZLqYmSEYp:-NPx_sha7QOhkUOl_Vxw:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/2fcf6c777fd4f867fbc3cfcce19933786082b264/src/main/kotlin/org/wfanet/measurement/duchy/service/internal/computations/ComputationsCleaner.kt#L46)):*

`force` and `dry_run` are opposites
</blockquote></details>

Done.

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/831)* status: 21 of 23 files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/831#-NPxRq992lzk8imnillc:-NPx_vkh36O5JnN-28Ir:b-54snbz) (no related file):*

FWIW, I prefer the pattern of using separate List and (Batch)Delete methods. Since we're only concerned with terminal states, we arguably don't even need an etag mechanism. See the similar comment in #872
</blockquote></details>

Acknowledged.

___

SanjayVas (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/831)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @renjiezh)

PR #842, opened 2023-02-06
infra as code for cross media measurement

wfa-mdharmalingam:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842)


wfa-siyengar:
cpplint is more to lint through the CPP code. Please use a terraform linter.

terraform/vm.tf
```diff
@@ -0,0 +1,22 @@
+/*
+
+What is this VM meant for.
+
+*/
+resource "google_compute_instance" "vm_instance" {
```
wfa-siyengar, line 6: Request few relevant comments for each file above.

terraform/variables.tf
```diff
@@ -0,0 +1,39 @@
...
+locals {
+  project = "wfa"
+  tags = {
+    Environment     = var.env
+  }
+  prefix      = "${var.env}-${local.project}"
+  prefix_path = "${var.env}/${local.project}"
+}
+
+data "template_file" "user_data" {
```
wfa-siyengar, line 13: Request comments about the file.

wfa-siyengar (approved):
Sparing the missing comments, code looks fine.
We also need to update to use terraform lint in place of the cpplint

kungfucraig:
Is this ready for review by other team members?

bdomen-ggl (commented):
You're also missing the KMS keys. See step 3 in the doc.

https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/main/docs/gke/kingdom-deployment.md

Also, you'll probably get lint errors that you're missing the apache license comment blocks on the files.

As another note, when I was working on some Terraform stuff, Sanjay noted that we probably want to nest this in a GCS folder since we might also have an AWS version (possibly others?) at some point. Probably OK to move when we have the requirement for another provider.

> Reviewed 8 of 9 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842#-:-NNwlKhp4jOJrCHwg35a:bumqb4q)* status: all files reviewed, 15 unresolved discussions (waiting on @SanjayVas, @wfa-mdharmalingam, and @wfa-siyengar)

___
*[`.github/workflows/terraform.yml` line 19 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842#-NNwvslt3cnehS4mi-Qr:-NNwvslt3cnehS4mi-Qs:b1wjkdg) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d1dc6bbf1d5c3ac8652dc72f73cc49ec87a42cba/.github/workflows/terraform.yml#L19)):*
> ```YAML
> #
> #         # The name of the Terraform Cloud workspace to store Terraform state files in.
> #         workspaces {
> ```

Looks like a discrepancy between the comments and the code. Are we using workspaces or no? The actual files just have like
```
terraform {
  backend "gcs" {
    bucket  = "terraform-state-mohanraj"
    prefix  = "wfa"
  }
}
```
___
*[`.github/workflows/terraform.yml` line 33 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842#-NNwiTDE48EVx4tLW4ZT:-NNwiTDE48EVx4tLW4ZU:bcozu5r) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d1dc6bbf1d5c3ac8652dc72f73cc49ec87a42cba/.github/workflows/terraform.yml#L33)):*
> ```YAML
> #
> #
> # 2. Generate a Terraform Cloud user API token and store it as a GitHub secret (e.g. TF_API_TOKEN) on this repository.
> ```

General question to reviewers, are we OK with storing secrets/tokens in GitHub? I've had mixed feelings from other companies before, but there's been a lot of security breaches recently and not sure how we feel about this. The risk is that in order for Terraform to work, it will need admin permissions (what most people do) or a curated list of permissions (which is a lot of maintenance overhead). The first being a risk to security and the latter being a lot of extra dev work, which in many cases ends up being an awful lot of access anyways.

I know we decided to use GitHub Actions, but a common alternative is to use GitHub Webhooks to notify your server of an action, then the server (that you control) can do whatever it needs. This is how you can configure for example GCS Cloud Build or AWS Code Pipeline.
___
*[`.github/workflows/terraform.yml` line 47 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842#-NNwxAiw0K1Hm0uT0mNy:-NNwxAiw0K1Hm0uT0mNz:b4g9ndu) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d1dc6bbf1d5c3ac8652dc72f73cc49ec87a42cba/.github/workflows/terraform.yml#L47)):*
> ```YAML
> name: 'Terraform_kingdom'
> 
> on:
> ```

I think you said you were going to make multiple actions? Doing so would align with the comments above.



_Code quote:_
```YAML
on:
  push:
    branches: [ "main" ]
  pull_request:
```
___
*[`.github/workflows/terraform.yml` line 104 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842#-NNwkSNy14UFVeel6fO_:-NNwkSNy14UFVeel6fOa:b-7mfjig) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d1dc6bbf1d5c3ac8652dc72f73cc49ec87a42cba/.github/workflows/terraform.yml#L104)):*
> ```YAML
>         working-directory: ${{env.ROOT_PATH}}
> 
>       # Applies the above created plan
> ```

As an FYI for others, even if you run fmt, validate, and plan, there can still be errors when you run apply that are GCS specific - like a duplicate DB name exists (some can be caught by plan) or permission denied or an improper name format.
___
*[`.github/workflows/terraform.yml` line 109 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842#-NNw_IsOBEIoayDlK-Lm:-NNw_IsP0S7Ol5CMMn6T:bm339r3) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d1dc6bbf1d5c3ac8652dc72f73cc49ec87a42cba/.github/workflows/terraform.yml#L109)):*
> ```YAML
>         working-directory: ${{env.ROOT_PATH}}
> 
>       # Destroy the infra created.
> ```

I think you said you were going to move this to a separate action.
___
*[`terraform/db.tf` line 11 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842#-NNw_R4MDVzWgkZxbfOe:-NNw_R4MDVzWgkZxbfOf:b-w3o8ij) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d1dc6bbf1d5c3ac8652dc72f73cc49ec87a42cba/terraform/db.tf#L11)):*
> ```HCL
>   name     = "kingdom_database"
>   version_retention_period = "3d"
>   ddl = [
> ```

We use liquibase to manage the schema, so we don't need anything here.
___
*[`terraform/db.tf` line 15 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842#-NNwcHmyBx_je86VB6s5:-NNwcHmyBx_je86VB6s6:bxordqc) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d1dc6bbf1d5c3ac8652dc72f73cc49ec87a42cba/terraform/db.tf#L15)):*
> ```HCL
>     "CREATE TABLE t2 (t2 INT64 NOT NULL,) PRIMARY KEY(t2)",
>   ]
>   deletion_protection = false
> ```

`deletion_protection = false`

This is generally dangerous and not recommended in the Terraform docs. It's used to provide a simple example, but this will destroy the database and data without warning. If we want to leave this as false to be managed completely by Terraform, we'll need to make sure we have strong data backup and recovery processes in place.

https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/spanner_database

Even if we modify the GitHub actions, we need to be extremely careful when deleting the database. Also I don't know how Terraform with GCS works, but even on certain attribute changes AWS will destroy the original resource and create a new one - like changing a DDB table name. So even if we don't run terraform destroy, it's possible that changing an attribute in the terraform definition can destroy the database as well (again, I'm not positive how Terraform works with GCS, so this may be OK).
___
*[`terraform/gke.tf` line 5 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842#-NNweKFP1KC3TslE-cFT:-NNweKFP1KC3TslE-cFU:b-auecxm) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d1dc6bbf1d5c3ac8652dc72f73cc49ec87a42cba/terraform/gke.tf#L5)):*
> ```HCL
> module "gke" {
>   source                     = "terraform-google-modules/kubernetes-engine/google"
>   project_id                 = ""
> ```

I'm curious, does an empty project_id work? I see it's marked as required and I don't see anywhere the module would replace empty with the provider's project id. It's a big repo, so I didn't check all the usages.
___
*[`terraform/gke.tf` line 12 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842#-NNwfsQT3412IvMD7jwE:-NNwfsQU24BoKcmdryQl:bd49tzr) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d1dc6bbf1d5c3ac8652dc72f73cc49ec87a42cba/terraform/gke.tf#L12)):*
> ```HCL
>   zones                      = ["us-central1-a", "us-central1-b", "us-central1-f"]
>   network                    = "terraform-network"
>   subnetwork                 = ""
> ```

I think we should have a subnetwork. Probably using the default example is OK. I'm not terribly familiar with cloud networking, so I could be wrong. Also I think it's related to the `ip_range_pods` and `ip_range_services`

https://github.com/terraform-google-modules/terraform-google-kubernetes-engine

EDIT: After double checking our GCP, it looks like we use the default subnet. Does leaving this blank do that?
___
*[`terraform/gke.tf` line 18 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842#-NNwgasQ69KSRNBmN29x:-NNwgasQ69KSRNBmN29y:b-e3oh1h) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d1dc6bbf1d5c3ac8652dc72f73cc49ec87a42cba/terraform/gke.tf#L18)):*
> ```HCL
>   filestore_csi_driver       = false
> 
>   node_pools = [
> ```

Double check the cluster configuration, some of the values are off like machine_type, min_count, and max_count.

https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/main/docs/gke/kingdom-deployment.md
___
*[`terraform/network.tf` line 2 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842#-NNwaAkrDMe87WoMsD1w:-NNwaAkrDMe87WoMsD1x:b-3pbty5) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d1dc6bbf1d5c3ac8652dc72f73cc49ec87a42cba/terraform/network.tf#L2)):*
> ```HCL
> resource "google_compute_network" "vpc_network" {
>   name = "terraform-network"
> ```

Do we need this? It looks like the google kubernetes module creates a vpc network and subnetwork.
___
*[`terraform/service-account.tf` line 7 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842#-NNwhAmi408bHyjNotP7:-NNwhAmi408bHyjNotP8:bp3d0es) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d1dc6bbf1d5c3ac8652dc72f73cc49ec87a42cba/terraform/service-account.tf#L7)):*
> ```HCL
>   prefix        = "test-sa"
>   names         = ["first", "second"]
> }
> ```

I recall you said you were going to fix the names to be more meaningful (and that we needed four?). Also I'd update the prefix to be more meaningful as well.



_Code quote:_
```HCL
  prefix        = "test-sa"
  names         = ["first", "second"]
```
___
*[`terraform/variables.tf` line 21 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842#-NNx-FMyBF6FrFfzqjk5:-NNx-FMz4kuACxsgwP39:bsgudrx) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d1dc6bbf1d5c3ac8652dc72f73cc49ec87a42cba/terraform/variables.tf#L21)):*
> ```HCL
> }
> 
> variable projectName {
> ```

Guess it's not required with default values, but it's nice to define `type` as well.

Also, we don't have a module so these are really more `locals` than `variable` unless you want to specify them from the command line in the future.
___
*[`terraform/vm.tf` line 6 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842#-NNaw6kbD_1ZsYGXKbsy-r1-6:-NNx0hAVBpZydPZwqVSg:b-7blvgo) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0d47cfa51534f7c093b934de79d25af62df67bc6/terraform/vm.tf#L6)):*
<details><summary><i>Previously, wfa-siyengar wrote…</i></summary><blockquote>

Request few relevant comments for each file above.
</blockquote></details>

I think except for the pre_install script, everything makes sense in terms of what each file is doing.

kungfucraig (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842)* status: all files reviewed, 15 unresolved discussions (waiting on @bdomen-ggl, @SanjayVas, @wfa-mdharmalingam, and @wfa-siyengar)

___
*[`.github/workflows/terraform.yml` line 33 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842#-NNwiTDE48EVx4tLW4ZT:-NNxWU1i7cBmDCcYKbVS:be30wtb) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d1dc6bbf1d5c3ac8652dc72f73cc49ec87a42cba/.github/workflows/terraform.yml#L33)):*

General question to reviewers, are we OK with storing secrets/tokens in GitHub? I've had mixed feelings from other companies before, but there's been a lot of security breaches recently and not sure how we feel about this. The risk is that in order for Terraform to work, it will need admin permissions (what most people do) or a curated list of permissions (which is a lot of maintenance overhead). The first being a risk to security and the latter being a lot of extra dev work, which in many cases ends up being an awful lot of access anyways.

I know we decided to use GitHub Actions, but a common alternative is to use GitHub Webhooks to notify your server of an action, then the server (that you control) can do whatever it needs. This is how you can configure for example GCS Cloud Build or AWS Code Pipeline.
</blockquote></details>

This YAML file should not be submitted. Let's just shoot for TF that can be run manually at the CLI and work out how to pass any secrets in on the CLI.

We should explore webhooks when we get to release automation.

I only see big downsides to putting secrets in the repo.

bdomen-ggl:
*[`.github/workflows/terraform.yml` line 33 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842#-NNwiTDE48EVx4tLW4ZT:-NOLPGL-297-iH8DJgTl:b57wn8b) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d1dc6bbf1d5c3ac8652dc72f73cc49ec87a42cba/.github/workflows/terraform.yml#L33)):*

This YAML file should not be submitted. Let's just shoot for TF that can be run manually at the CLI and work out how to pass any secrets in on the CLI.

We should explore webhooks when we get to release automation.

I only see big downsides to putting secrets in the repo. 
</blockquote></details>

Ah, the secrets wouldn't be in the repo itself, but GitHub has like a secrets manager to securely store sensitive things like secrets/tokens/credentials. I know lots of people are OK with it, but I'm personally not a fan of trusting 3P services to hold my secrets (like how LastPass got hacked).

Also running TF ourselves is rather easy. We just need appropriate GCS access and run: `gcloud auth login`

wfa-siyengar (approved):
Looks good for an initial version of IAC

bdomen-ggl (commented):
Looking closer. I also just realized you have `terraform` as a top level folder. We are actually putting it under `src/main/terraform` like here:

https://github.com/world-federation-of-advertisers/panel-exchange-client/tree/main/src/main/terraform

Also, I think we would generally benefit a lot if we converted this into a module rather than have everything flat. You can also see the above link regarding how that looks. This way we can define everything in Terraform rather than have some things in Terraform and some things in the command line invocation. All the command line invocation will do is pick a particular configuration.

> Reviewed 13 of 13 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842#-:-NOuN24mAWrMqx8f72Vr:b-gjcgpr)* status: all files reviewed, 22 unresolved discussions (waiting on @kungfucraig, @SanjayVas, @wfa-mdharmalingam, and @wfa-siyengar)

___
*[`terraform/build_image.sh` line 15 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842#-NOuCRlw6lrBxjNoA8l1:-NOuCRlw6lrBxjNoA8l2:br2unk3) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/554a1e4cda3c2aaacab806a2143a8e15072f3f5a/terraform/build_image.sh#L15)):*
> ```Shell
> 
> 7. Build the images and push to registry ( this takes around 1.5 hours for the first time )
> ```

Sanjay is working on a GitHub action for this, so we won't need to do this as part of the TF deployment.

All we'll do with TF is deploy the infrastructure. Other steps will handle things like building the image, attaching the image to the K8S cluster, applying manifests, running tests, etc.
___
*[`terraform/db.tf` line 11 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842#-NNw_R4MDVzWgkZxbfOe:-NOuCd9iBzBO8Yk45apw:b76cx9z) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d1dc6bbf1d5c3ac8652dc72f73cc49ec87a42cba/terraform/db.tf#L11)):*

We use liquibase to manage the schema, so we don't need anything here.
</blockquote></details>

Can just remove this code instead of comment it out.
___
*[`terraform/gcr.tf` line 16 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842#-NOuCl0G6M6Hb6h8ZFTN:-NOuCl0G6M6Hb6h8ZFTO:bcqyqng) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/554a1e4cda3c2aaacab806a2143a8e15072f3f5a/terraform/gcr.tf#L16)):*
> ```HCL
> 
> /*
> resource "google_container_registry" "registry" {
> ```

@SanjayVas I forgot, how are you planning to manage the container registry? Should we create it as part of Terraform or is that a manually managed step? If you want the builds to be separate from Terraform, I suppose it would be better to decouple the steps and have container registry manually managed, or at least managed outside of these TF files.
___
*[`terraform/gke.tf` line 18 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842#-NOuMNgnE7OUPsnVvrWc:-NOuMNgnE7OUPsnVvrWd:b-kwc9xc) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/554a1e4cda3c2aaacab806a2143a8e15072f3f5a/terraform/gke.tf#L18)):*
> ```HCL
> # Create a GKE cluster with one node
> resource "google_container_cluster" "cluster" {
>   name     = "cluster"
> ```

We should have a more easily identifiable cluster name.
___
*[`terraform/gke.tf` line 35 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842#-NOuLrAq0ofXNmbK-5QZ:-NOuLrAq0ofXNmbK-5Q_:bdppb87) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/554a1e4cda3c2aaacab806a2143a8e15072f3f5a/terraform/gke.tf#L35)):*
> ```HCL
> 
> 
>  
> ```

There's some strange extra whitespace at the end
___
*[`terraform/kms.tf` line 15 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842#-NOuJWlY1HtbKCd90t5v:-NOuJWlY1HtbKCd90t5w:b-cqbhwv) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/554a1e4cda3c2aaacab806a2143a8e15072f3f5a/terraform/kms.tf#L15)):*
> ```HCL
> 
> /*
> ```

As a note, it's probably easier to track if we use `count` + `condition` to create the resources or not.

Count: https://developer.hashicorp.com/terraform/language/meta-arguments/count
Condition: https://developer.hashicorp.com/terraform/language/expressions/conditionals

Rather than having to know what files to comment in/out code, we can just change variables.
___
*[`terraform/kms.tf` line 29 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842#-NOuDScWCM18Od00VS9o:-NOuDScWCM18Od00VS9p:b-jlv27t) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/554a1e4cda3c2aaacab806a2143a8e15072f3f5a/terraform/kms.tf#L29)):*
> ```HCL
> 
> resource "google_kms_crypto_key" "default" {
>   name = "gce_east1_symm_key"
> ```

Don't think we need the region in the name. Also, it's central1 not east1, right? Otherwise, why are we using a mix of regions?
___
*[`terraform/kms.tf` line 32 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842#-NOuEa7s1h6yw_KGIlgr:-NOuEa7s1h6yw_KGIlgs:blbadec) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/554a1e4cda3c2aaacab806a2143a8e15072f3f5a/terraform/kms.tf#L32)):*
> ```HCL
>   key_ring = data.google_kms_key_ring.my_key_ring.id
> }
> data "google_iam_policy" "default" {
> ```

I don't see this used anywhere. `data` isn't a resource that gets created and needs to be used by a resource.

https://registry.terraform.io/providers/hashicorp/google/latest/docs/data-sources/iam_policy

If it's not used, we can delete it. Although I think we need this to associate the KMS key with the K8S service account.
___
*[`terraform/network.tf` line 33 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842#-NOuM3jk8RKF4kss-Tdm:-NOuM3jk8RKF4kss-Tdn:be85i90) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/554a1e4cda3c2aaacab806a2143a8e15072f3f5a/terraform/network.tf#L33)):*
> ```HCL
>   ip_cidr_range = "10.0.0.0/24"
>   region        = "us-central1"
>   
> ```

Some weird space here



_Code quote:_
```HCL
··
```
___
*[`terraform/pre_install.sh.tpl` line 1 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842#-NOuDpuw4CKPv8xfALHh:-NOuDpuw4CKPv8xfALHi:b-qp9k32) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/554a1e4cda3c2aaacab806a2143a8e15072f3f5a/terraform/pre_install.sh.tpl#L1)):*
> ```Smarty
> # Copyright 2020 The Cross-Media Measurement Authors
> ```

This file looks like it's a copy of the above. If so we don't need both. If not, documentation somewhere (in the files or a README) would be nice.
___
*[`terraform/variables.tf` line 34 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842#-NOuKZYD62h-9YVqA-6T:-NOuKZYD62h-9YVqA-6U:b-j9nj1b) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/554a1e4cda3c2aaacab806a2143a8e15072f3f5a/terraform/variables.tf#L34)):*
> ```HCL
> }
> 
> variable projectName {
> ```

I see several variables not referenced anywhere. Let's either use them or remove the ones we don't need.
___
*[`terraform/variables.tf` line 63 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842#-NOuKf8S4pgcLdkCFjG9:-NOuKf8S4pgcLdkCFjGA:bdnwzy7) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/554a1e4cda3c2aaacab806a2143a8e15072f3f5a/terraform/variables.tf#L63)):*
> ```HCL
> }
> 
> variable db_password {
> ```

FYI: I don't see this used anywhere, but we shouldn't store credentials in plaintext in TF. We can use GitHub secrets manager as you noted in the GitHub actions file (now removed).

Or my preference is to use GCP's secret manager. I've done this in AWS but not GCS and didn't read the whole article, but the section on reading secrets sounds about right.

https://gmusumeci.medium.com/how-to-use-gcp-secret-manager-to-manage-your-secrets-using-terraform-488a75a65b28



_Code quote:_
```HCL
db_user
```
___
*[`terraform/variables.tf` line 85 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842#-NOuIJI9C5VDw3fNFMpk:-NOuIJI9C5VDw3fNFMpl:bmagrhq) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/554a1e4cda3c2aaacab806a2143a8e15072f3f5a/terraform/variables.tf#L85)):*
> ```HCL
> }
> 
> variable "project_id" {
> ```

We can set project id like this:

https://registry.terraform.io/providers/hashicorp/google/latest/docs/data-sources/project

In order for Terraform to work, it uses the `provider` 

https://cloud.google.com/sdk/gcloud/reference/config/set

Rather than having the provider be configurable, we usually want to specify the provider configuration. If we want multiple providers (say to deploy to different GCPs) we can create multiple providers and have Terraform use a specific one.
___
*[`terraform/variables.tf` line 101 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842#-NOuKDPn3nYIgZ3bDqpf:-NOuKDPn3nYIgZ3bDqpg:bflqlh7) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/554a1e4cda3c2aaacab806a2143a8e15072f3f5a/terraform/variables.tf#L101)):*
> ```HCL
> variable "ring_location" {
>   type = string
>   default ="us-east1"
> ```

Let's be consistent with our regions. I see a mix of us-central1 and us-east1.
___
*[`terraform/vm.tf` line 6 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842#-NNaw6kbD_1ZsYGXKbsy-r1-6:-NOuLV2H-qRkK3NwziPD:bfyixah) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0d47cfa51534f7c093b934de79d25af62df67bc6/terraform/vm.tf#L6)):*

I think except for the pre_install script, everything makes sense in terms of what each file is doing.
</blockquote></details>

I should have mentioned though, while it's clear what this file is doing, it's not clear why we need it. I would guess it's to build and deploy the code, but we'll be doing that as a separate step from Terraform.
___
*[`terraform/vm.tf` line 41 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842#-NOuHPq4565fpJQ337Wk:-NOuHPq4565fpJQ337Wl:bv8h357) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/554a1e4cda3c2aaacab806a2143a8e15072f3f5a/terraform/vm.tf#L41)):*
> ```HCL
>     
>   
>   
> ```

There's a bunch of weird whitespace here

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842)* status: all files reviewed, 22 unresolved discussions (waiting on @bdomen-ggl, @kungfucraig, @wfa-mdharmalingam, and @wfa-siyengar)

___
*[`terraform/gcr.tf` line 16 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842#-NOuCl0G6M6Hb6h8ZFTN:-NOzG2V5DK1kR3V_fTUg:b-y7gy4) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/554a1e4cda3c2aaacab806a2143a8e15072f3f5a/terraform/gcr.tf#L16)):*

@SanjayVas I forgot, how are you planning to manage the container registry? Should we create it as part of Terraform or is that a manually managed step? If you want the builds to be separate from Terraform, I suppose it would be better to decouple the steps and have container registry manually managed, or at least managed outside of these TF files.
</blockquote></details>

Building and publishing container images would be a separate GitHub Action and not handled by Terraform. We basically want to port the workflow that I wrote in the containers repo to this one: https://github.com/world-federation-of-advertisers/containers/blob/main/.github/workflows/push-images.yml.

The container images would be in GitHub packages.

wfa-mdharmalingam:
> Looking closer. I also just realized you have `terraform` as a top level folder. We are actually putting it under `src/main/terraform` like here:
> 
> https://github.com/world-federation-of-advertisers/panel-exchange-client/tree/main/src/main/terraform
> 
> Also, I think we would generally benefit a lot if we converted this into a module rather than have everything flat. You can also see the above link regarding how that looks. This way we can define everything in Terraform rather than have some things in Terraform and some things in the command line invocation. All the command line invocation will do is pick a particular configuration.
> 
> > Reviewed 13 of 13 files at r4, all commit messages.
> > _[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842#-:-NOuN24mAWrMqx8f72Vr:b-gjcgpr)_ status: all files reviewed, 22 unresolved discussions (waiting on @kungfucraig, @SanjayVas, @wfa-mdharmalingam, and @wfa-siyengar)
> 
> _[`terraform/build_image.sh` line 15 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842#-NOuCRlw6lrBxjNoA8l1:-NOuCRlw6lrBxjNoA8l2:br2unk3) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/554a1e4cda3c2aaacab806a2143a8e15072f3f5a/terraform/build_image.sh#L15)):_
> 
> > ```shell
> > # limitations under the License.
> > 
> > 7. Build the images and push to registry ( this takes around 1.5 hours for the first time )
> > ```
> 
> Sanjay is working on a GitHub action for this, so we won't need to do this as part of the TF deployment.
> 
> All we'll do with TF is deploy the infrastructure. Other steps will handle things like building the image, attaching the image to the K8S cluster, applying manifests, running tests, etc.
> 
> _[`terraform/db.tf` line 11 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842#-NNw_R4MDVzWgkZxbfOe:-NOuCd9iBzBO8Yk45apw:b76cx9z) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d1dc6bbf1d5c3ac8652dc72f73cc49ec87a42cba/terraform/db.tf#L11)):_
> 
> Previously, bdomen-ggl wrote…
> Can just remove this code instead of comment it out.
> 
> _[`terraform/gcr.tf` line 16 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842#-NOuCl0G6M6Hb6h8ZFTN:-NOuCl0G6M6Hb6h8ZFTO:bcqyqng) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/554a1e4cda3c2aaacab806a2143a8e15072f3f5a/terraform/gcr.tf#L16)):_
> 
> > ```terraform
> > /*
> > resource "google_container_registry" "registry" {
> > ```
> 
> @SanjayVas I forgot, how are you planning to manage the container registry? Should we create it as part of Terraform or is that a manually managed step? If you want the builds to be separate from Terraform, I suppose it would be better to decouple the steps and have container registry manually managed, or at least managed outside of these TF files.
> 
> _[`terraform/gke.tf` line 18 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842#-NOuMNgnE7OUPsnVvrWc:-NOuMNgnE7OUPsnVvrWd:b-kwc9xc) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/554a1e4cda3c2aaacab806a2143a8e15072f3f5a/terraform/gke.tf#L18)):_
> 
> > ```terraform
> > # Create a GKE cluster with one node
> > resource "google_container_cluster" "cluster" {
> >   name     = "cluster"
> > ```
> 
> We should have a more easily identifiable cluster name.
> 
> _[`terraform/gke.tf` line 35 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842#-NOuLrAq0ofXNmbK-5QZ:-NOuLrAq0ofXNmbK-5Q_:bdppb87) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/554a1e4cda3c2aaacab806a2143a8e15072f3f5a/terraform/gke.tf#L35)):_
> 
> > ```terraform
> >  
> > ```
> 
> There's some strange extra whitespace at the end
> 
> _[`terraform/kms.tf` line 15 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842#-NOuJWlY1HtbKCd90t5v:-NOuJWlY1HtbKCd90t5w:b-cqbhwv) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/554a1e4cda3c2aaacab806a2143a8e15072f3f5a/terraform/kms.tf#L15)):_
> 
> > ```terraform
> > # limitations under the License.
> > 
> > /*
> > ```
> 
> As a note, it's probably easier to track if we use `count` + `condition` to create the resources or not.
> 
> Count: https://developer.hashicorp.com/terraform/language/meta-arguments/count Condition: https://developer.hashicorp.com/terraform/language/expressions/conditionals
> 
> Rather than having to know what files to comment in/out code, we can just change variables.
> 
> _[`terraform/kms.tf` line 29 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842#-NOuDScWCM18Od00VS9o:-NOuDScWCM18Od00VS9p:b-jlv27t) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/554a1e4cda3c2aaacab806a2143a8e15072f3f5a/terraform/kms.tf#L29)):_
> 
> > ```terraform
> > resource "google_kms_crypto_key" "default" {
> >   name = "gce_east1_symm_key"
> > ```
> 
> Don't think we need the region in the name. Also, it's central1 not east1, right? Otherwise, why are we using a mix of regions?
> 
> _[`terraform/kms.tf` line 32 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842#-NOuEa7s1h6yw_KGIlgr:-NOuEa7s1h6yw_KGIlgs:blbadec) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/554a1e4cda3c2aaacab806a2143a8e15072f3f5a/terraform/kms.tf#L32)):_
> 
> > ```terraform
> >   key_ring = data.google_kms_key_ring.my_key_ring.id
> > }
> > data "google_iam_policy" "default" {
> > ```
> 
> I don't see this used anywhere. `data` isn't a resource that gets created and needs to be used by a resource.
> 
> https://registry.terraform.io/providers/hashicorp/google/latest/docs/data-sources/iam_policy
> 
> If it's not used, we can delete it. Although I think we need this to associate the KMS key with the K8S service account.
> 
> _[`terraform/network.tf` line 33 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842#-NOuM3jk8RKF4kss-Tdm:-NOuM3jk8RKF4kss-Tdn:be85i90) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/554a1e4cda3c2aaacab806a2143a8e15072f3f5a/terraform/network.tf#L33)):_
> 
> > ```terraform
> >   ip_cidr_range = "10.0.0.0/24"
> >   region        = "us-central1"
> >   
> > ```
> 
> Some weird space here
> 
> _Code quote:_
> 
> ```terraform
> ··
> ```
> 
> _[`terraform/pre_install.sh.tpl` line 1 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842#-NOuDpuw4CKPv8xfALHh:-NOuDpuw4CKPv8xfALHi:b-qp9k32) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/554a1e4cda3c2aaacab806a2143a8e15072f3f5a/terraform/pre_install.sh.tpl#L1)):_
> 
> > ```
> > # Copyright 2020 The Cross-Media Measurement Authors
> > ```
> 
> This file looks like it's a copy of the above. If so we don't need both. If not, documentation somewhere (in the files or a README) would be nice.
> 
> _[`terraform/variables.tf` line 34 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842#-NOuKZYD62h-9YVqA-6T:-NOuKZYD62h-9YVqA-6U:b-j9nj1b) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/554a1e4cda3c2aaacab806a2143a8e15072f3f5a/terraform/variables.tf#L34)):_
> 
> > ```terraform
> > }
> > 
> > variable projectName {
> > ```
> 
> I see several variables not referenced anywhere. Let's either use them or remove the ones we don't need.
> 
> _[`terraform/variables.tf` line 63 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842#-NOuKf8S4pgcLdkCFjG9:-NOuKf8S4pgcLdkCFjGA:bdnwzy7) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/554a1e4cda3c2aaacab806a2143a8e15072f3f5a/terraform/variables.tf#L63)):_
> 
> > ```terraform
> > }
> > 
> > variable db_password {
> > ```
> 
> FYI: I don't see this used anywhere, but we shouldn't store credentials in plaintext in TF. We can use GitHub secrets manager as you noted in the GitHub actions file (now removed).
> 
> Or my preference is to use GCP's secret manager. I've done this in AWS but not GCS and didn't read the whole article, but the section on reading secrets sounds about right.
> 
> https://gmusumeci.medium.com/how-to-use-gcp-secret-manager-to-manage-your-secrets-using-terraform-488a75a65b28
> 
> _Code quote:_
> 
> ```terraform
> db_user
> ```
> 
> _[`terraform/variables.tf` line 85 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842#-NOuIJI9C5VDw3fNFMpk:-NOuIJI9C5VDw3fNFMpl:bmagrhq) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/554a1e4cda3c2aaacab806a2143a8e15072f3f5a/terraform/variables.tf#L85)):_
> 
> > ```terraform
> > }
> > 
> > variable "project_id" {
> > ```
> 
> We can set project id like this:
> 
> https://registry.terraform.io/providers/hashicorp/google/latest/docs/data-sources/project
> 
> In order for Terraform to work, it uses the `provider`
> 
> https://cloud.google.com/sdk/gcloud/reference/config/set
> 
> Rather than having the provider be configurable, we usually want to specify the provider configuration. If we want multiple providers (say to deploy to different GCPs) we can create multiple providers and have Terraform use a specific one.
> 
> _[`terraform/variables.tf` line 101 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842#-NOuKDPn3nYIgZ3bDqpf:-NOuKDPn3nYIgZ3bDqpg:bflqlh7) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/554a1e4cda3c2aaacab806a2143a8e15072f3f5a/terraform/variables.tf#L101)):_
> 
> > ```terraform
> > variable "ring_location" {
> >   type = string
> >   default ="us-east1"
> > ```
> 
> Let's be consistent with our regions. I see a mix of us-central1 and us-east1.
> 
> _[`terraform/vm.tf` line 6 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842#-NNaw6kbD_1ZsYGXKbsy-r1-6:-NOuLV2H-qRkK3NwziPD:bfyixah) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0d47cfa51534f7c093b934de79d25af62df67bc6/terraform/vm.tf#L6)):_
> 
> Previously, bdomen-ggl wrote…
> I should have mentioned though, while it's clear what this file is doing, it's not clear why we need it. I would guess it's to build and deploy the code, but we'll be doing that as a separate step from Terraform.
> 
> _[`terraform/vm.tf` line 41 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842#-NOuHPq4565fpJQ337Wk:-NOuHPq4565fpJQ337Wl:bv8h357) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/554a1e4cda3c2aaacab806a2143a8e15072f3f5a/terraform/vm.tf#L41)):_
> 
> > ```terraform
> >     
> >   
> >   
> > ```
> 
> There's a bunch of weird whitespace here

As

wfa-mdharmalingam:
> src/main/terraform

we have moved the terraform code under the path src/main/terraform

bdomen-ggl (commented):
Looks pretty good for now, but it looks like the PR is moved to "closed"?

Just change a couple names and I think we can merge this.
- DB: kingdom-qa
- K8S cluster: halo-cmm-kingdom-qa

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842#-:-NP3FVnW6yjMbjT_t08e:busbjy0)* status: 1 of 15 files reviewed, 22 unresolved discussions

kungfucraig (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842)* status: 1 of 15 files reviewed, 24 unresolved discussions (waiting on @bdomen-ggl, @wfa-mdharmalingam, and @wfa-siyengar)

___
*[`src/main/terraform/variables.tf` line 35 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842#-NP3SWxN3fHOqhmVhBI_:-NP3SWxN3fHOqhmVhBIa:b-rhdyd1) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/10e04938da0d1e64fa0d462620919b278b568570/src/main/terraform/variables.tf#L35)):*
> ```HCL
> 
> variable projectName {
>   default = "wfa"
> ```

default = "halo"
___
*[`src/main/terraform/variables.tf` line 36 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842#-NP3SZ9hCwY-GIJNUHo2:-NP3SZ9hCwY-GIJNUHo3:b-jle3bx) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/10e04938da0d1e64fa0d462620919b278b568570/src/main/terraform/variables.tf#L36)):*
> ```HCL
> variable projectName {
>   default = "wfa"
>   description = "The IAC project of WFA "
> ```

description = "Infrastructure as Code for Halo"

bdomen-ggl (commented):
Still looking for the name changes ^
Also we don't need the shell scripts. Sanjay has some bazel commands that we'll hook into GitHub actions to perform those tasks. So the actions will eventually chain something like: Deploy Infrastructure (TF) > Build Images > Push Images > etc.

> Reviewed 6 of 14 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842#-:-NPJLzI59iDyFFjCFsof:b-rh2cjm)* status: 7 of 17 files reviewed, 24 unresolved discussions (waiting on @kungfucraig, @wfa-mdharmalingam, and @wfa-siyengar)

hiq-mdharmalingam:
> Is this ready for review by other team members?
Yes @kungfucraig, Kingdon code is ready for review. in the recent change we have added a the code to build image and push it to  the GCR registry

wfa-siyengar (changes requested):
> Reviewed 1 of 13 files at r4, 3 of 14 files at r5, 3 of 6 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842)* status: 10 of 17 files reviewed, 23 unresolved discussions (waiting on @bdomen-ggl, @kungfucraig, and @wfa-mdharmalingam)

___
*[`.github/workflows/terraform.yml` line 33 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842#-NNwiTDE48EVx4tLW4ZT:-NPNLP2BB-e-t3-QLyyl:b-4t0v47) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d1dc6bbf1d5c3ac8652dc72f73cc49ec87a42cba/.github/workflows/terraform.yml#L33)):*
> min permissions (what most people do) or a curated list of permissions (which is a lot of maintenance overhead). The first being a risk to security and the latter being a lot of extra dev work, which in many cases ends up being an awful lot of access anyways.
We intend to try and use the workload Identity.
___
*[`.github/workflows/terraform.yml` line 104 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842#-NNwkSNy14UFVeel6fO_:-NPNL_FaEx0WGrpeyCIl:bjjdr2) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d1dc6bbf1d5c3ac8652dc72f73cc49ec87a42cba/.github/workflows/terraform.yml#L104)):*

As an FYI for others, even if you run fmt, validate, and plan, there can still be errors when you run apply that are GCS specific - like a duplicate DB name exists (some can be caught by plan) or permission denied or an improper name format.
</blockquote></details>

We have a sandbox now. We test it there.
___
*[`src/main/terraform/variables.tf` line 35 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842#-NP3SWxN3fHOqhmVhBI_:-NPNKK738tNBpZEE8USc:b-u080ec) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/10e04938da0d1e64fa0d462620919b278b568570/src/main/terraform/variables.tf#L35)):*

default = "halo"
</blockquote></details>

Agreed. changed!
___
*[`src/main/terraform/variables.tf` line 36 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842#-NP3SZ9hCwY-GIJNUHo2:-NPNKVBZ4sO0fo35zqR5:b-a2ujfl) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/10e04938da0d1e64fa0d462620919b278b568570/src/main/terraform/variables.tf#L36)):*

description = "Infrastructure as Code for Halo"
</blockquote></details>

Changed.
___
*[`terraform/build_image.sh` line 15 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842#-NOuCRlw6lrBxjNoA8l1:-NPNFEWh4qko9bxrbhYp:b-176nph) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/554a1e4cda3c2aaacab806a2143a8e15072f3f5a/terraform/build_image.sh#L15)):*

Sanjay is working on a GitHub action for this, so we won't need to do this as part of the TF deployment.

All we'll do with TF is deploy the infrastructure. Other steps will handle things like building the image, attaching the image to the K8S cluster, applying manifests, running tests, etc.
</blockquote></details>

Agreed. This Script might need some changes with respect to Paths.
___
*[`terraform/db.tf` line 11 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842#-NNw_R4MDVzWgkZxbfOe:-NPNFm_Q1nvxzTNl3JGv:b-9tygx0) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d1dc6bbf1d5c3ac8652dc72f73cc49ec87a42cba/terraform/db.tf#L11)):*

Can just remove this code instead of comment it out.
</blockquote></details>

Sure, will remove the code.
Just to be sure, we then will not include DB creation as part of the IAC.
___
*[`terraform/db.tf` line 15 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842#-NNwcHmyBx_je86VB6s5:-NPNGCX76drwlOJvZtzw:b8que14) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d1dc6bbf1d5c3ac8652dc72f73cc49ec87a42cba/terraform/db.tf#L15)):*

`deletion_protection = false`

This is generally dangerous and not recommended in the Terraform docs. It's used to provide a simple example, but this will destroy the database and data without warning. If we want to leave this as false to be managed completely by Terraform, we'll need to make sure we have strong data backup and recovery processes in place.

https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/spanner_database

Even if we modify the GitHub actions, we need to be extremely careful when deleting the database. Also I don't know how Terraform with GCS works, but even on certain attribute changes AWS will destroy the original resource and create a new one - like changing a DDB table name. So even if we don't run terraform destroy, it's possible that changing an attribute in the terraform definition can destroy the database as well (again, I'm not positive how Terraform works with GCS, so this may be OK).
</blockquote></details>

Agreed, we will set it to true before we finally deploy. This was set to false for us to establish the straight path. You can be sure that we have a tab on it.
___
*[`terraform/gcr.tf` line 16 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842#-NOuCl0G6M6Hb6h8ZFTN:-NPNGd3L598hNAaYAZfO:bi5akbh) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/554a1e4cda3c2aaacab806a2143a8e15072f3f5a/terraform/gcr.tf#L16)):*

Building and publishing container images would be a separate GitHub Action and not handled by Terraform. We basically want to port the workflow that I wrote in the containers repo to this one: https://github.com/world-federation-of-advertisers/containers/blob/main/.github/workflows/push-images.yml.

The container images would be in GitHub packages.
</blockquote></details>

Sanjay, Having Image creation separated means disjoint set of automations, shouldn't we think of keeping them connected. 
 I was visualizing a scenario where we just click and the whole Infra with its artifact is created, deployed and Presented as ready for use.
@SanjayVas 
___
*[`terraform/gke.tf` line 12 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842#-NNwfsQT3412IvMD7jwE:-NPNHWbJ7Yka2bBb8Olf:b-uffv77) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d1dc6bbf1d5c3ac8652dc72f73cc49ec87a42cba/terraform/gke.tf#L12)):*

I think we should have a subnetwork. Probably using the default example is OK. I'm not terribly familiar with cloud networking, so I could be wrong. Also I think it's related to the `ip_range_pods` and `ip_range_services`

https://github.com/terraform-google-modules/terraform-google-kubernetes-engine

EDIT: After double checking our GCP, it looks like we use the default subnet. Does leaving this blank do that?
</blockquote></details>

Using a Default VPC and default subnet is generally not a practice we have followed with any of our clients. 
But open to try.
___
*[`terraform/gke.tf` line 18 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842#-NNwgasQ69KSRNBmN29x:-NPNHrp85NYP7TR_HTCY:bgft6pk) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d1dc6bbf1d5c3ac8652dc72f73cc49ec87a42cba/terraform/gke.tf#L18)):*

Double check the cluster configuration, some of the values are off like machine_type, min_count, and max_count.

https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/main/docs/gke/kingdom-deployment.md
</blockquote></details>

@wfa-mdharmalingam request your attention on this and confirm.
___
*[`terraform/gke.tf` line 18 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842#-NOuMNgnE7OUPsnVvrWc:-NPNHHVy9OqP5Sk2DIky:b-m2n93i) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/554a1e4cda3c2aaacab806a2143a8e15072f3f5a/terraform/gke.tf#L18)):*

We should have a more easily identifiable cluster name.
</blockquote></details>

Request suggestion on the name as we are still getting used to the CMM lingo.
___
*[`terraform/gke.tf` line 35 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842#-NOuLrAq0ofXNmbK-5QZ:-NPNIzk0E--v8nwJoeN8:bvk9aed) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/554a1e4cda3c2aaacab806a2143a8e15072f3f5a/terraform/gke.tf#L35)):*

There's some strange extra whitespace at the end
</blockquote></details>

Stands cleared. @wfa-mdharmalingam request confirmation.
___
*[`terraform/kms.tf` line 29 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842#-NOuDScWCM18Od00VS9o:-NPNJKfX9WbqcYNF0Bf9:b-iitacw) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/554a1e4cda3c2aaacab806a2143a8e15072f3f5a/terraform/kms.tf#L29)):*
> https://developer.hashicorp.com/terraform/language/expressions/conditionals
Thanks We will put this to use.
___
*[`terraform/kms.tf` line 32 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842#-NOuEa7s1h6yw_KGIlgr:-NPNJPK96qnR-mfJQ00u:bdf61v3) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/554a1e4cda3c2aaacab806a2143a8e15072f3f5a/terraform/kms.tf#L32)):*

I don't see this used anywhere. `data` isn't a resource that gets created and needs to be used by a resource.

https://registry.terraform.io/providers/hashicorp/google/latest/docs/data-sources/iam_policy

If it's not used, we can delete it. Although I think we need this to associate the KMS key with the K8S service account.
</blockquote></details>

Will check.
___
*[`terraform/network.tf` line 33 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842#-NOuM3jk8RKF4kss-Tdm:-NPNJfOJ8pnZmHeFl-Li:brpmqc3) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/554a1e4cda3c2aaacab806a2143a8e15072f3f5a/terraform/network.tf#L33)):*

Some weird space here
</blockquote></details>

removed.
___
*[`terraform/variables.tf` line 13 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842#-NNaw6kbD_1ZsYGXKbsx-r1-13:-NPNK41B5Cp6XZypw7cX:bb6un1j) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0d47cfa51534f7c093b934de79d25af62df67bc6/terraform/variables.tf#L13)):*
<details><summary><i>Previously, wfa-siyengar wrote…</i></summary><blockquote>

Request comments about the file.
</blockquote></details>

Done
___
*[`terraform/variables.tf` line 21 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842#-NNx-FMyBF6FrFfzqjk5:-NPNK7SP6ZZSnykEqdAP:b6u0cpk) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d1dc6bbf1d5c3ac8652dc72f73cc49ec87a42cba/terraform/variables.tf#L21)):*

Guess it's not required with default values, but it's nice to define `type` as well.

Also, we don't have a module so these are really more `locals` than `variable` unless you want to specify them from the command line in the future.
</blockquote></details>

Will be done.
___
*[`terraform/variables.tf` line 34 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842#-NOuKZYD62h-9YVqA-6T:-NPNKArU1w0H0ZUeBkVE:baka19w) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/554a1e4cda3c2aaacab806a2143a8e15072f3f5a/terraform/variables.tf#L34)):*

I see several variables not referenced anywhere. Let's either use them or remove the ones we don't need.
</blockquote></details>

Will look and remove variables not used. It will anyway be gone when puthem as modules.
___
*[`terraform/variables.tf` line 85 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842#-NOuIJI9C5VDw3fNFMpk:-NPNKhJCCaD7SCW8Z5xp:b-dhzm24) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/554a1e4cda3c2aaacab806a2143a8e15072f3f5a/terraform/variables.tf#L85)):*
> https://registry.terraform.io/providers/hashicorp/google/latest/docs/data-sources/project
Sure, Data resources is what you are pointing at.
___
*[`terraform/vm.tf` line 6 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842#-NNaw6kbD_1ZsYGXKbsy-r1-6:-NPNKyBO4g-U2xBrNxH7:b-pbdqzg) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0d47cfa51534f7c093b934de79d25af62df67bc6/terraform/vm.tf#L6)):*

I should have mentioned though, while it's clear what this file is doing, it's not clear why we need it. I would guess it's to build and deploy the code, but we'll be doing that as a separate step from Terraform.
</blockquote></details>

Agreed!
___
*[`terraform/vm.tf` line 41 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842#-NOuHPq4565fpJQ337Wk:-NPNL5fN1j9cHZ2A0UwV:b67fbhj) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/554a1e4cda3c2aaacab806a2143a8e15072f3f5a/terraform/vm.tf#L41)):*
<details><summary><i>Previously, wfa-mdharmalingam wrote…</i></summary><blockquote>

As 
</blockquote></details>

Cleared

wfa-siyengar (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842#-:-NPNLoNK5EQyVHAoReMe:bnfp4nl)* status: 10 of 17 files reviewed, 23 unresolved discussions (waiting on @bdomen-ggl, @kungfucraig, and @wfa-mdharmalingam)

wfa-siyengar:
We will need a working meet to decide on the approach.

TNATALI:
@kungfucraig @bdomen-ggl please review this PR

wfa-mdharmalingam:
> > Reviewed all commit messages.
> > _[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842)_ status: 1 of 15 files reviewed, 24 unresolved discussions (waiting on @bdomen-ggl, @wfa-mdharmalingam, and @wfa-siyengar)
> 
> _[`src/main/terraform/variables.tf` line 35 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842#-NP3SWxN3fHOqhmVhBI_:-NP3SWxN3fHOqhmVhBIa:b-rhdyd1) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/10e04938da0d1e64fa0d462620919b278b568570/src/main/terraform/variables.tf#L35)):_
> 
> > ```terraform
> > variable projectName {
> >   default = "wfa"
> > ```
> 
> default = "halo"
> 
> _[`src/main/terraform/variables.tf` line 36 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842#-NP3SZ9hCwY-GIJNUHo2:-NP3SZ9hCwY-GIJNUHo3:b-jle3bx) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/10e04938da0d1e64fa0d462620919b278b568570/src/main/terraform/variables.tf#L36)):_
> 
> > ```terraform
> > variable projectName {
> >   default = "wfa"
> >   description = "The IAC project of WFA "
> > ```
> 
> description = "Infrastructure as Code for Halo"

@kungfucraig project name and description has been changed

duliomatos1 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842)* status: 1 of 17 files reviewed, 23 unresolved discussions (waiting on @bdomen-ggl, @kungfucraig, @SanjayVas, @wfa-mdharmalingam, and @wfa-siyengar)

___
*[`src/main/terraform/variables.tf` line 36 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842#-NP3SZ9hCwY-GIJNUHo2:-NPO04-6014ejelyQw6u:b-kk3ut2) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/10e04938da0d1e64fa0d462620919b278b568570/src/main/terraform/variables.tf#L36)):*
<details><summary><i>Previously, wfa-mdharmalingam wrote…</i></summary><blockquote>

@kungfucraig project name and description has been changed 
</blockquote></details>

solved
___
*[`terraform/db.tf` line 15 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842#-NNwcHmyBx_je86VB6s5:-NPOFAULAT3S6bgnLq7B:b48gkmf) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d1dc6bbf1d5c3ac8652dc72f73cc49ec87a42cba/terraform/db.tf#L15)):*
<details><summary><i>Previously, wfa-siyengar wrote…</i></summary><blockquote>

Agreed, we will set it to true before we finally deploy. This was set to false for us to establish the straight path. You can be sure that we have a tab on it.
</blockquote></details>

If you need to disable deletion protection when you're testing the code, can you enable deletion outside the script, only when needed? I think the concern here is to have something unsafe in code that we could forget about later. e.g. passing through a command
`terraform apply -var="deletion_protection=false"`

I see there's a deletion protection you have to setup in Cloud console or gcloud command https://cloud.google.com/sql/docs/postgres/deletion-protection, you may have to enable/disable the protection outside terraform as well as in a variable

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842)* status: 1 of 17 files reviewed, 23 unresolved discussions (waiting on @bdomen-ggl, @kungfucraig, @wfa-mdharmalingam, and @wfa-siyengar)

___
*[`terraform/gcr.tf` line 16 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842#-NOuCl0G6M6Hb6h8ZFTN:-NPOUN0FCHnrtxVorgc_:b-2hlqr6) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/554a1e4cda3c2aaacab806a2143a8e15072f3f5a/terraform/gcr.tf#L16)):*
<details><summary><i>Previously, wfa-siyengar wrote…</i></summary><blockquote>

Sanjay, Having Image creation separated means disjoint set of automations, shouldn't we think of keeping them connected. 
 I was visualizing a scenario where we just click and the whole Infra with its artifact is created, deployed and Presented as ready for use.
@SanjayVas 
</blockquote></details>

Publishing of artifacts is independent from updating the QA environment. That is to say that the artifacts for a release need to be published in a way that is accessible to everyone. Therefore the artifacts should be published to GitHub Packages, as the artifacts are tied to a release in the GitHub repo.

It just happens to be that both events (publishing the artifacts and updating the QA environment) are both triggered by the creation of a release in GitHub, but that only matters insofar as the QA environment update process can rely on the fact that there are versioned artifacts already published and available in a known stable location.

duliomatos1 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842)* status: 1 of 17 files reviewed, 23 unresolved discussions (waiting on @bdomen-ggl, @kungfucraig, @wfa-mdharmalingam, and @wfa-siyengar)

___
*[`terraform/gke.tf` line 12 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842#-NNwfsQT3412IvMD7jwE:-NPOXo4z61s_ioliOhcG:b9eriip) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d1dc6bbf1d5c3ac8652dc72f73cc49ec87a42cba/terraform/gke.tf#L12)):*
<details><summary><i>Previously, wfa-siyengar wrote…</i></summary><blockquote>

Using a Default VPC and default subnet is generally not a practice we have followed with any of our clients. 
But open to try.
</blockquote></details>

Feel free to suggest a better practice whenever required, I agree it's better to create a separate VPC and not go with the default one

bdomen-ggl (commented):
> Reviewed 1 of 14 files at r7.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842)* status: 2 of 17 files reviewed, 24 unresolved discussions (waiting on @kungfucraig, @wfa-mdharmalingam, and @wfa-siyengar)

___
*[`src/main/terraform/kingdom/gcloud/db.tf` line 24 at r9](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842#-NPOcQhz4skFqkk0r1bv:-NPOcQhz4skFqkk0r1bw:bcdzi1w) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/937b0ba15642c1377a4539000abb202809f05aa9/src/main/terraform/kingdom/gcloud/db.tf#L24)):*
> ```HCL
> resource "google_spanner_database" "database" {
>   instance = google_spanner_instance.main.name
>   name     = "spanner_database"
> ```

Oops to make it clear, this should be `kingdom-qa` and I think that's good for now.
___
*[`terraform/db.tf` line 11 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842#-NNw_R4MDVzWgkZxbfOe:-NPNsBLc3CoHB3MSknex:b-4g8mgs) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d1dc6bbf1d5c3ac8652dc72f73cc49ec87a42cba/terraform/db.tf#L11)):*
<details><summary><i>Previously, wfa-siyengar wrote…</i></summary><blockquote>

Sure, will remove the code.
Just to be sure, we then will not include DB creation as part of the IAC.
</blockquote></details>

Oh no. DB creation is needed, but the table creations aren't. So we don't need just the ddl section.
___
*[`terraform/gcr.tf` line 16 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842#-NOuCl0G6M6Hb6h8ZFTN:-NPNsSf18TTYFx-6ibNO:b-n2byx3) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/554a1e4cda3c2aaacab806a2143a8e15072f3f5a/terraform/gcr.tf#L16)):*
<details><summary><i>Previously, wfa-siyengar wrote…</i></summary><blockquote>

Sanjay, Having Image creation separated means disjoint set of automations, shouldn't we think of keeping them connected. 
 I was visualizing a scenario where we just click and the whole Infra with its artifact is created, deployed and Presented as ready for use.
@SanjayVas 
</blockquote></details>

There's two high level approaches. One is your suggestion to have one thing do all the work. The other is to use a combination. I guess Sanjay is going for the later. They both have their pros and cons.

In this specific case, the argument to split up the pieces is that there's a lot of things we need that Terraform isn't really designed to handle - specifically creating and executing all these scripts. I'm not that familiar with running scripts with Terraform, but I don't think it is well equipped especially in our case of running scripts in a VM to handle and report error conditions. The nice part about GitHub actions is since we're running custom scripts, we can return errors to GitHub actions and it will be clear if we ran into an error or not even on the custom scripts.

In a more general scenario I would rather have one tool too (eg. Terraform) rather than a combination (eg. Terraform + GitHub Actions w/ custom scripts).
___
*[`terraform/gke.tf` line 12 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842#-NNwfsQT3412IvMD7jwE:-NPNv9KoALHtBv3NjhKN:bqqmiu8) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d1dc6bbf1d5c3ac8652dc72f73cc49ec87a42cba/terraform/gke.tf#L12)):*
<details><summary><i>Previously, wfa-siyengar wrote…</i></summary><blockquote>

Using a Default VPC and default subnet is generally not a practice we have followed with any of our clients. 
But open to try.
</blockquote></details>

Right, it doesn't hurt to add a non-default subnet I suppose and if you have it working then great. Just didn't want to add extra work since we're currently using the default one.
___
*[`terraform/gke.tf` line 18 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842#-NOuMNgnE7OUPsnVvrWc:-NPNufH04U5ExD-Pi-k3:b-vomfnh) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/554a1e4cda3c2aaacab806a2143a8e15072f3f5a/terraform/gke.tf#L18)):*
<details><summary><i>Previously, wfa-siyengar wrote…</i></summary><blockquote>

Request suggestion on the name as we are still getting used to the CMM lingo.
</blockquote></details>

You can look at the existing ones in halo-cmm-dev, but in case it's not clear which one's which or you don't have access, here you go.

`halo-cmm-kingdom-demo-cluster`

It doesn't have to be EXACTLY the same, but something like that. Maybe since this is the QA version we could call it `halo-cmm-kingdom-qa-cluster`

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842)* status: 2 of 17 files reviewed, 24 unresolved discussions (waiting on @bdomen-ggl, @kungfucraig, @wfa-mdharmalingam, and @wfa-siyengar)

___
*[`terraform/gcr.tf` line 16 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842#-NOuCl0G6M6Hb6h8ZFTN:-NPOh_0r-DufRmfU1uhA:b532i9e) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/554a1e4cda3c2aaacab806a2143a8e15072f3f5a/terraform/gcr.tf#L16)):*

There's two high level approaches. One is your suggestion to have one thing do all the work. The other is to use a combination. I guess Sanjay is going for the later. They both have their pros and cons.

In this specific case, the argument to split up the pieces is that there's a lot of things we need that Terraform isn't really designed to handle - specifically creating and executing all these scripts. I'm not that familiar with running scripts with Terraform, but I don't think it is well equipped especially in our case of running scripts in a VM to handle and report error conditions. The nice part about GitHub actions is since we're running custom scripts, we can return errors to GitHub actions and it will be clear if we ran into an error or not even on the custom scripts.

In a more general scenario I would rather have one tool too (eg. Terraform) rather than a combination (eg. Terraform + GitHub Actions w/ custom scripts).
</blockquote></details>

Right now we have one tool we're already using: GitHub Actions. That tool will continue to be in use no matter what.

We want the scope of what we use Terraform to be used for to be minimal: just setting up the QA environment within the cloud project. This means the databases, clusters, etc. It does *not* include the Kubernetes objects such as deployments, pods, etc. Therefore the container images would never be referenced by Terraform since this Terraform would not be responsible for bringing up any Kubernetes pods.

Basically the release will kick off the following actions:
1. Build and publish artifacts
2. Provision the cloud project infra, i.e. run Terraform to apply any necessary changes
3. Update the Kubernetes objects. This includes applying updated object configurations pointing to the new images from step 3.

Note how Terraform is only involved in step 2.

bdomen-ggl (commented):
> Reviewed 1 of 14 files at r7.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/842)* status: 3 of 17 files reviewed, 24 unresolved discussions (waiting on @kungfucraig, @wfa-mdharmalingam, and @wfa-siyengar)

wfa-hkawalkar:
We have added Duchy code . Please let me know if any changes needs to be done.

wfa-hkawalkar:
The code will deploy Duchy VM with preinstall packages also build Duchy Image.

wfa-siyengar:
closing this PR

Issue #840, opened 2023-02-01
Kingdom internal server eventually deadlines on every RPC, runs out of memory


SanjayVas:
Observed on QA. First failure starts ~2023-01-27 10:50 AM PST. It eventually ran out of memory at 2023-01-28 10:30:16.921 PST. The server starts responding again for awhile after restarts before starting to fail again. It restarted at 2023-01-28 10:30:28.785 PST, 2023-01-28 10:31:05.808 PST, and 2023-01-28 10:33:07.228 PST. There was another `java.lang.OutOfMemoryError: Java heap space` at 2023-01-31 01:43:49.256 PST.

Heap dump analysis appears to show too many Spanner objects retained in heap:

```
Class Histogram (by retained heap)
----------------------------------
Percent      Size   Count  Name
  33.2%  29.71MiB    4369  com.google.cloud.spanner.AsyncResultSetImpl$ProduceRowsCallable
  33.1%  29.61MiB   98086  java.lang.Object[]
  31.6%  28.27MiB    4141  com.google.cloud.spanner.SessionPool$AutoClosingReadContext$AutoClosingReadContextAsyncResultSetImpl
  27.5%  24.64MiB  148673  byte[]
  22.4%  20.09MiB  132548  java.lang.String
  19.6%  17.53MiB    2541  com.google.cloud.spanner.SessionPool$SessionPoolAsyncRunner$$Lambda$402
  13.8%  12.39MiB    6909  java.util.concurrent.CancellationException
  13.8%  12.37MiB    4829  com.google.common.base.Suppliers$SupplierOfInstance
  13.6%  12.14MiB    4141  com.google.cloud.spanner.SessionPool$AutoClosingReadContext$1
  13.4%  12.02MiB    4141  com.google.cloud.spanner.SessionPool$AutoClosingReadContext$7
```

SanjayVas:
The theory of why we haven't experienced this before is that the QA environment is now under much heavier load, i.e. there are more Measurements flowing through it.

SanjayVas:
It [looks like](https://pantheon.corp.google.com/errors/detail/CP-Th6Xfqo2QbA;time=P30D?project=halo-cmm-dev) we didn't start seeing this behavior until Jan 25, and in earnest starting Jan 28.

This is around the time we updated dependencies related to Spanner, so that has a chance of being a potential cause.

SanjayVas:
[...]

PR #848, opened 2023-02-08
Make Duchy's endComputation idempotent

renjiezh:
https://github.com/world-federation-of-advertisers/cross-media-measurement/issues/840

To prevent mill from restlessly claiming abnormal COMPLETE Computations which did not reset the lock, make the function endCompution() idempotent by only resetting the lock.

- - -
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/848)


renjiezh:
The cause of Computations in wrong state (finished but not reset lock) is unclear, as the log before January 8th was lost. But this fix can help eliminate these abnormal Computations.

SanjayVas (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/848)* status: all files reviewed, 1 unresolved discussion (waiting on @kungfucraig and @renjiezh)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/gcloud/spanner/computation/GcpSpannerComputationsDatabaseTransactor.kt` line 300 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/848#-NNmUxZoCjo1cJ7gTLVF:-NNmUxZoCjo1cJ7gTLVG:b-1o56mq) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/9a19751821692273aff2ba42b269c5834ac7af3f/src/main/kotlin/org/wfanet/measurement/duchy/deploy/gcloud/spanner/computation/GcpSpannerComputationsDatabaseTransactor.kt#L300)):*
> ```Kotlin
>     runIfTokenFromLastUpdate(token) { ctx ->
>       val writeTime = clock.gcloudTimestamp()
>       // If the Computation is already in the endingStage, reset lockOwner and lockExpirationTime.
> ```

nit: "clear", or use the higher-level phrase "release the lock"



_Code quote:_
```Kotlin
reset
```

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/848)* status: all files reviewed, 2 unresolved discussions (waiting on @duliomatos1, @kungfucraig, and @renjiezh)

*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/848#-NNnCf5iAz963DNpFqiE:-NNnCf5j-MlRfk2SD3l2:bwbejnn) (no related file):*
Just to clarify, I would have assumed the issue wouldn't be related to the lock but rather the COMPLETE state. I would assume that nothing should try to work on a computation that's already marked as COMPLETE.

Just double-checking to make sure that this isn't the kind of band-aid that may bite us in the future.

___

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/848)* status: all files reviewed, 1 unresolved discussion (waiting on @duliomatos1, @kungfucraig, and @SanjayVas)

*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/848#-NNnCf5iAz963DNpFqiE:-NNnoCZkDv1nvABADRfw:b-kc30r3) (no related file):*

Just to clarify, I would have assumed the issue wouldn't be related to the lock but rather the COMPLETE state. I would assume that nothing should try to work on a computation that's already marked as COMPLETE.

Just double-checking to make sure that this isn't the kind of band-aid that may bite us in the future.
</blockquote></details>

The only situation of working on a COMPLETE computation that I can find is when Kingdom notify a state change of FAILED/CANCELED. At this moment, duchy will try to end a computation that is already COMPLETE.  This PR is to let the duchy to skip operations. But I still add operation of releasing lock to solve current abnormal Computations.

___

kungfucraig (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/848)* status: all files reviewed, 2 unresolved discussions (waiting on @duliomatos1, @renjiezh, and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/gcloud/spanner/computation/GcpSpannerComputationsDatabaseTransactor.kt` line 299 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/848#-NNrIV9Y1fd_Ac1kgmPy:-NNrIV9Y1fd_Ac1kgmPz:b-kg9mpk) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4538f0cc0168f982c3cbcfc796c89e9320e4f762/src/main/kotlin/org/wfanet/measurement/duchy/deploy/gcloud/spanner/computation/GcpSpannerComputationsDatabaseTransactor.kt#L299)):*
> ```Kotlin
>       val writeTime = clock.gcloudTimestamp()
>       // If the Computation is already in the endingStage, release the lock.
>       if (token.stage == endingStage) {
> ```

Can you find out where the actual state transition happens and have it also clear the lockOwner at that time. This change doesn't seem like it's addressing the root cause.

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/848)* status: all files reviewed, 2 unresolved discussions (waiting on @duliomatos1 and @renjiezh)

*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/848#-NNnCf5iAz963DNpFqiE:-NNrRI3Y85m8aRglth9K:b-aldyl4) (no related file):*

The only situation of working on a COMPLETE computation that I can find is when Kingdom notify a state change of FAILED/CANCELED. At this moment, duchy will try to end a computation that is already COMPLETE.  This PR is to let the duchy to skip operations. But I still add operation of releasing lock to solve current abnormal Computations.
</blockquote></details>

A one-time fix of abnormal computations can be done via a DB update I think. We don't want to have code permanently around to fix conditions that won't exist in the future.

___

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/848)* status: all files reviewed, 2 unresolved discussions (waiting on @duliomatos1 and @kungfucraig)

*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/848#-NNnCf5iAz963DNpFqiE:-NNrRqsM9KB5V0qZkzv7:b-t6lhsb) (no related file):*

A one-time fix of abnormal computations can be done via a DB update I think. We don't want to have code permanently around to fix conditions that won't exist in the future.
</blockquote></details>

That sounds good to me. We can delete those Computations manually and keep eyes on the case. If it happens again, I can trace it via logs. 
@kungfucraig How do you feel about it? Waiting for the root cause to expose itself again.

___

kungfucraig (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/848)* status: all files reviewed, 2 unresolved discussions (waiting on @duliomatos1, @renjiezh, and @SanjayVas)

*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/848#-NNnCf5iAz963DNpFqiE:-NNrvH-q7WIiHADgI9Um:b1oy9uc) (no related file):*

That sounds good to me. We can delete those Computations manually and keep eyes on the case. If it happens again, I can trace it via logs. 
@kungfucraig How do you feel about it? Waiting for the root cause to expose itself again.
</blockquote></details>

sgtm. Thanks!

___

PR #851, opened 2023-02-09
Include the correctness test in the "all targets" step.

SanjayVas:
This allows correctness test results to be cached so that the test is not re-run if no dependencies have changed. The downside is that we do not get streaming test output, but the test timeouts should make that unnecessary.

- - -
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/851)

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -37,23 +37,7 @@ jobs:
           workspace-path: .
           target-patterns: |
             //...
-          build-options: |
-            --compilation_mode=opt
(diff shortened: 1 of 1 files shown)


stevenwarejones (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/851)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @SanjayVas)

PR #847, opened 2023-02-07
Rearrange joins on ComputationParticipantReader.

SanjayVas:
This is to reduce the number of table rows scanned. Spanner Query Insights on QA showed an average latency of ~300ms and average rows scanned of ~4162.

- - -
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/847)

PR #854, opened 2023-02-10
Revert "Rearrange joins on ComputationParticipantReader."

SanjayVas:
Reverts world-federation-of-advertisers/cross-media-measurement#847

This actually made things significantly worse. Average rows scanned went up to 14k and average latency up to 5,261.36ms.

It looks like rearranging the MeasurementLogEntries join was a bad change, but rearranging the Measurements join was good. Despite that, doing a clean revert to keep things simple.


SanjayVas:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/854)

SanjayVas:
Nevermind. It appears that the issue was due to still having bad computations around with too many log entries (see #840). When I did the manual deletion, I deleted DuchyMeasurementLogEntries rows but not their corresponding MeasurementLogEntries rows.

Reverting the log entry join would still be an improvement, but this is still better than the old one.

From QA with `ExternalComputationId = 5239238807293094679 AND ComputationParticipants.DuchyId = 2345`:
* Old query: 2,778 rows scanned
* New query: 86 rows scanned
* Modified query reverting just the log entry join: 82 rows scanned.

PR #856, opened 2023-02-13
Update DB schema / add logic for reading-writing duchy-include-list /…

Marco-Premier:
[TA567](https://rally1.rallydev.com/#/604612930531d/iterationstatus?detail=%2Ftask%2F685508089433)

Scope of this PR:

- update DB in order to include (1:many) duchy include list
- update CreateResource picocli command to create a data provider
- update spanners readers / writers for data provider resource


SanjayVas:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/856)

stevenwarejones (changes requested):
> Reviewed 14 of 14 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/856)* status: all files reviewed, 3 unresolved discussions (waiting on @Marco-Premier and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/SpannerDataProvidersService.kt` line 41 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/856#-NOAqWVf6BPvv81Fzw8J:-NOAqWVf6BPvv81Fzw8K:b-7agtxq) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/edd625b331d9dd81952750bc6e5ab57c19a90916/src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/SpannerDataProvidersService.kt#L41)):*
> ```Kotlin
>       "Details field of DataProvider is missing fields."
>     }
>     grpcRequire(request.externalDuchyIdCount > 0) { "ExternalDuchyId list is empty" }
> ```

we don't need this requirement. the list can be empty and then no rows inserted into the table
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/readers/DataProviderReader.kt` line 89 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/856#-NOArKi_3IV2rcmHqzFi:-NOArKi_3IV2rcmHqzFj:b-m9awe3) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/edd625b331d9dd81952750bc6e5ab57c19a90916/src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/readers/DataProviderReader.kt#L89)):*
> ```Kotlin
>   private fun buildExternalDuchyIdList(struct: Struct): MutableList<Long> {
>     val duchiesExternalIdList = mutableListOf<Long>()
>     for (duchyIncludeListValue in struct.getStructList("DataProviderDuchyIncludeList")) {
> ```

```
return struct.getStructList("DataProviderDuchyIncludeList").map{ 
   it..getLong("ExternalDuchyId")
}
```

___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/writers/CreateDataProvider.kt` line 24 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/856#-NOAs3qd3Zm_m4GcGN0V:-NOAs3qd3Zm_m4GcGN0W:b-lp4qu0) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/edd625b331d9dd81952750bc6e5ab57c19a90916/src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/writers/CreateDataProvider.kt#L24)):*
> ```Kotlin
> import org.wfanet.measurement.internal.kingdom.copy
> 
> class CreateDataProvider(private val dataProvider: DataProvider) :
> ```

is there a separate write for editing a data provider?

SanjayVas (changes requested):
> Reviewed 5 of 14 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/856)* status: all files reviewed, 6 unresolved discussions (waiting on @Marco-Premier and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/writers/CreateDataProvider.kt` line 24 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/856#-NOAs3qd3Zm_m4GcGN0V:-NOCF2Eq4psnZamX-ju9:b-yv8mgm) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/edd625b331d9dd81952750bc6e5ab57c19a90916/src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/writers/CreateDataProvider.kt#L24)):*

is there a separate write for editing a data provider?
</blockquote></details>

There's currently no Update method for DataProviders in the public API. Probably something that needs to be added so EDPs can update their lists of required Duchies.
___
*[`src/main/proto/wfa/measurement/internal/kingdom/data_provider.proto` line 43 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/856#-NOCDu3C1zolYkDWb9dD:-NOCDu3DB_oupNiS4zMJ:b-7rvtve) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/edd625b331d9dd81952750bc6e5ab57c19a90916/src/main/proto/wfa/measurement/internal/kingdom/data_provider.proto#L43)):*
> ```protobuf
> 
>   // The list of duchies that must be included in all computations involving the DataProvider
>   repeated int64 external_duchy_id = 4;
> ```

 Maybe name this `required_external_duchy_ids`? I assume we'd call the public API field something like `required_duchies` so the purpose is clear.

Regardless, the names of repeated fields should be plural.



_Code quote:_
```protobuf
external_duchy_id
```
___
*[`src/main/resources/kingdom/spanner/create-duchy-include-list-schema.sql` line 18 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/856#-NOCED089f1cj42OQVFr:-NOCED089f1cj42OQVFs:be5gby5) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/edd625b331d9dd81952750bc6e5ab57c19a90916/src/main/resources/kingdom/spanner/create-duchy-include-list-schema.sql#L18)):*
> ```TSQL
> 
> -- changeset marcopremier:6 dbms:cloudspanner
> CREATE TABLE DataProviderDuchyIncludeList (
> ```

Maybe `DataProviderRequiredDuchies`? Table names should be plural.

Also in general not a fan of "include list":
1. It's not a list, it's a set (i.e. order doesn't matter and there are expected to be no duplicates)
2. It's not just what to include, but what is required (**must** be included)



_Code quote:_
```TSQL
DataProviderDuchyIncludeList
```
___
*[`src/main/resources/kingdom/spanner/create-duchy-include-list-schema.sql` line 20 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/856#-NOCEVZnAtOf6YIc8NqU:-NOCEVZnAtOf6YIc8NqV:b-xstt2f) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/edd625b331d9dd81952750bc6e5ab57c19a90916/src/main/resources/kingdom/spanner/create-duchy-include-list-schema.sql#L20)):*
> ```TSQL
> CREATE TABLE DataProviderDuchyIncludeList (
>     DataProviderId INT64 NOT NULL,
>     -- Duchy external id that must be included in every computation that involves this DP id.
> ```

All references should use internal IDs. We can join with the corresponding table (or in the case of Duchy IDs, use the config) for external ID lookups.



_Code quote:_
```TSQL
    -- Duchy external id that must be included in every computation that involves this DP id.
    ExternalDuchyId INT64 NOT NULL,
```

stevenwarejones (changes requested):
> Reviewed 21 of 21 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/856)* status: all files reviewed, 5 unresolved discussions (waiting on @Marco-Premier and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/readers/DataProviderReader.kt` line 92 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/856#-NOGzfBq2N4Dpr40tGrb:-NOGzfBq2N4Dpr40tGrc:b-3f301l) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/69606747d7d9bcaebe08723936c36fb0cc519b65/src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/readers/DataProviderReader.kt#L92)):*
> ```Kotlin
>       .getStructList("DataProviderRequiredDuchies")
>       .map {
>         checkNotNull(DuchyIds.getExternalId(it.getLong("DuchyId"))) {
> ```

I don't think you want to throw an error, right? you should filter out null 
```
.filterNotNull
```

___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/writers/CreateDataProvider.kt` line 24 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/856#-NOAs3qd3Zm_m4GcGN0V:-NOGzvJS2sOcDQY_r-Wa:b-2p61jw) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/edd625b331d9dd81952750bc6e5ab57c19a90916/src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/writers/CreateDataProvider.kt#L24)):*

There's currently no Update method for DataProviders in the public API. Probably something that needs to be added so EDPs can update their lists of required Duchies.
</blockquote></details>

can you add a TODO for that @Marco-Premier ?

SanjayVas (changes requested):
> Reviewed 1 of 14 files at r1, 21 of 21 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/856)* status: all files reviewed, 17 unresolved discussions (waiting on @Marco-Premier and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/readers/DataProviderReader.kt` line 88 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/856#-NOHDcna3B9j04S9vV9w:-NOHDcna3B9j04S9vV9x:undefined) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/69606747d7d9bcaebe08723936c36fb0cc519b65/src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/readers/DataProviderReader.kt#L88)):*
> ```Kotlin
>       .build()
> 
>   private fun buildExternalDuchyIdList(struct: Struct): Iterable<String?> {
> ```





_Suggestion:_
```Kotlin
String
```
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/readers/DataProviderReader.kt` line 88 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/856#-NOHDjM92nQgKKoMDb6y:-NOHDjM92nQgKKoMDb6z:b-qoqyyk) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/69606747d7d9bcaebe08723936c36fb0cc519b65/src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/readers/DataProviderReader.kt#L88)):*
> ```Kotlin
>       .build()
> 
>   private fun buildExternalDuchyIdList(struct: Struct): Iterable<String?> {
> ```

nit: You have method called buildXXXList which returns an `Iterable` not a `List`



_Code quote:_
```Kotlin
List
```
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/readers/DataProviderReader.kt` line 92 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/856#-NOGzfBq2N4Dpr40tGrb:-NOHBd7uAqCoxW1JK3bq:b-9ka7ve) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/69606747d7d9bcaebe08723936c36fb0cc519b65/src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/readers/DataProviderReader.kt#L92)):*

I don't think you want to throw an error, right? you should filter out null 
```
.filterNotNull
```

</blockquote></details>

Throwing an error seems correct. We shouldn't just silently drop entries that have an invalid ID. If we get a `null` from the lookup, it means something is wrong as an invariant was violated. From the design doc/discussion, Duchies shouldn't get deleted but may be marked inactive.

If we did want to support hard deletion of Duchies we would want to have cascading deletion behavior. Since Duchies is a config file rather than a real table, this behavior would have to be implemented manually.
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/readers/DataProviderReader.kt` line 96 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/856#-NOHDNlSATDbEe4H5q5S:-NOHDNlSATDbEe4H5q5T:b-cxvdng) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/69606747d7d9bcaebe08723936c36fb0cc519b65/src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/readers/DataProviderReader.kt#L96)):*
> ```Kotlin
>         }
>       }
>       .asIterable()
> ```

I believe `map` returns a `List` in this case, which already implements `Iterable`.



_Code quote:_
```Kotlin
      .asIterable()
```
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/writers/CreateDataProvider.kt` line 45 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/856#-NOHG-QUDj3Oc-siToCh:-NOHG-QVDP9rkmBHBh4B:b-gkutlf) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/69606747d7d9bcaebe08723936c36fb0cc519b65/src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/writers/CreateDataProvider.kt#L45)):*
> ```Kotlin
>     }
> 
>     // This takes O(n*m) where n is the list of duchies and m the list of required ones. Expected
> ```

This comment doesn't belong here as it is referring to an implementation detail of `DuchyIds.getInternalId`.

(`DuchyIds.getInternalId` also shouldn't have the note about being implemented as a linear scan in its public documentation, or should at least make it clear that it is not part of the contract and may change in the future.)



_Code quote:_
```Kotlin
    // This takes O(n*m) where n is the list of duchies and m the list of required ones. Expected
    // number of duchies should quite small (~5) and the required <= 2
```
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/deploy/tools/CreateResource.kt` line 152 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/856#-NOH5udVEEOnmVLqdLGC:-NOH5udVEEOnmVLqdLGD:b45h) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/69606747d7d9bcaebe08723936c36fb0cc519b65/src/main/kotlin/org/wfanet/measurement/kingdom/deploy/tools/CreateResource.kt#L152)):*
> ```Kotlin
>     required = false,
>   )
>   private lateinit var requiredDuchies: Array<String>
> ```





_Suggestion:_
```Kotlin
List
```
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/deploy/tools/CreateResource.kt` line 164 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/856#-NOH5hmY5-S80pxPzI-W:-NOH5hmZ5n-_mZmbRw50:undefined) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/69606747d7d9bcaebe08723936c36fb0cc519b65/src/main/kotlin/org/wfanet/measurement/kingdom/deploy/tools/CreateResource.kt#L164)):*
> ```Kotlin
>         }
> 
>       for (duchyExternalId in requiredDuchies) {
> ```





_Suggestion:_
```Kotlin
      requiredExternalDuchyIds += requiredDuchies
```
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/internal/testing/ComputationParticipantsServiceTest.kt` line 64 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/856#-NOHHwE60zpVUbacqtUG:-NOH6IJ35651gfFJYMDY:b-ho2fwi) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/69606747d7d9bcaebe08723936c36fb0cc519b65/src/main/kotlin/org/wfanet/measurement/kingdom/service/internal/testing/ComputationParticipantsServiceTest.kt#L64)):*
> ```Kotlin
> private const val RANDOM_SEED = 1
> private const val PROVIDED_MEASUREMENT_ID = "measurement"
> private val EXTERNAL_DUCHY_IDS = listOf("worker1", "worker2")
> ```

Why change this? This file appears to be otherwise untouched by this PR.
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/internal/testing/DataProvidersServiceTest.kt` line 123 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/856#-NOH6Q6Y80e1VSl4hZ88:-NOH6Q6Y80e1VSl4hZ89:undefined) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/69606747d7d9bcaebe08723936c36fb0cc519b65/src/main/kotlin/org/wfanet/measurement/kingdom/service/internal/testing/DataProvidersServiceTest.kt#L123)):*
> ```Kotlin
>             publicKeySignature = PUBLIC_KEY_SIGNATURE
>           }
>           addAllRequiredExternalDuchyIds(mutableListOf("worker1", "worker2"))
> ```





_Suggestion:_
```Kotlin
listOf
```
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/internal/testing/DataProvidersServiceTest.kt` line 192 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/856#-NOHAXnQCkGcuPju1kQo:-NOHAXnQCkGcuPju1kQp:undefined) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/69606747d7d9bcaebe08723936c36fb0cc519b65/src/main/kotlin/org/wfanet/measurement/kingdom/service/internal/testing/DataProvidersServiceTest.kt#L192)):*
> ```Kotlin
>             publicKeySignature = PUBLIC_KEY_SIGNATURE
>           }
>           addAllRequiredExternalDuchyIds(mutableListOf("worker1", "worker2"))
> ```





_Suggestion:_
```Kotlin
listOf
```
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/internal/testing/MeasurementLogEntriesServiceTest.kt` line 46 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/856#-NOHNyTD8q0frHEY5jXR:-NOHNyTD8q0frHEY5jXS:b-ncz8vr) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/69606747d7d9bcaebe08723936c36fb0cc519b65/src/main/kotlin/org/wfanet/measurement/kingdom/service/internal/testing/MeasurementLogEntriesServiceTest.kt#L46)):*
> ```Kotlin
> 
> private const val RANDOM_SEED = 1
> private val EXTERNAL_DUCHY_IDS = listOf("Buck", "Rippon", "Shoaks", "worker1", "worker2")
> ```

Same here re: interaction with Population.



_Code quote:_
```Kotlin
, "worker1", "worker2"
```
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/internal/testing/MeasurementsServiceTest.kt` line 72 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/856#-NOHP6r3DJY4VBMYwFKH:-NOHP6r3DJY4VBMYwFKI:b-ncz8vr) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/69606747d7d9bcaebe08723936c36fb0cc519b65/src/main/kotlin/org/wfanet/measurement/kingdom/service/internal/testing/MeasurementsServiceTest.kt#L72)):*
> ```Kotlin
> private const val API_VERSION = "v2alpha"
> private const val PROVIDED_MEASUREMENT_ID = "ProvidedMeasurementId"
> private val EXTERNAL_DUCHY_IDS = listOf("Buck", "Rippon", "Shoaks", "worker1", "worker2")
> ```

Same here re: interaction with Population.
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/internal/testing/Population.kt` line 142 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/856#-NOHItEt5lFdC0BKXcVk:-NOHItEt5lFdC0BKXcVl:b-fibymj) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/69606747d7d9bcaebe08723936c36fb0cc519b65/src/main/kotlin/org/wfanet/measurement/kingdom/service/internal/testing/Population.kt#L142)):*
> ```Kotlin
>             publicKeySignature = "EDP public key signature".toByteStringUtf8()
>           }
>         requiredExternalDuchyIds.addAll(mutableListOf("worker1", "worker2"))
> ```

Either this should come from a parameter to `createDataProvider` or the EXTERNAL_DUCHY_IDS constant should be moved to the companion object of this class so the same ones aren't being redefined in multiple tests in such a way that they need to be kept in sync.

If doing the latter, I'd recommend having the constant be ("worker1", "worker2", "aggregator") if you want to match a more realistic configuration. The required Duchies for the DataProvider can still leave out "aggregator". You can also make up this constant from three separate ones so they can be referenced individually without using list indices, e.g. `WORKER1_DUCHY_EXTERNAL_ID`.

Also Kotlin protobuf DSL lists implement the `+=` operator, so the preferred syntax would be either

```kotlin
requiredExternalDuchyIds += listOf("worker1", "worker2")
```

or to save on an extra list creation

```kotlin
requiredExternalDuchyIds += "worker1"
requiredExternalDuchyIds += "worker2"
```
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/internal/testing/RequisitionsServiceTest.kt` line 68 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/856#-NOHL6xv20eCyVv7TuMZ:-NOHL6xv20eCyVv7TuM_:b-u1zbz0) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/69606747d7d9bcaebe08723936c36fb0cc519b65/src/main/kotlin/org/wfanet/measurement/kingdom/service/internal/testing/RequisitionsServiceTest.kt#L68)):*
> ```Kotlin
> private const val NONCE_1 = 3127743798281582205L
> private const val NONCE_2 = -7004399847946251733L
> private val EXTERNAL_DUCHY_IDS = listOf("Buck", "Rippon", "Shoaks", "worker1", "worker2")
> ```

Either have `Population.createDataProvider` take the list of Duchies as a param so this can remain unchanged or make use of shared constants.

Note that in the latter case, you can still just use the longer list of Duchies here by e.g. concatenating with a shared list constant.



_Code quote:_
```Kotlin
listOf("Buck", "Rippon", "Shoaks", "worker1", "worker2")
```
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/resourcesetup/ResourceSetup.kt` line 245 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/856#-NOHM5aa8qOUV8uiMfjD:-NOHM5aa8qOUV8uiMfjE:b-gxn2l3) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/69606747d7d9bcaebe08723936c36fb0cc519b65/src/main/kotlin/org/wfanet/measurement/loadtest/resourcesetup/ResourceSetup.kt#L245)):*
> ```Kotlin
>                 publicKeySignature = signedPublicKey.signature
>               }
>             requiredExternalDuchyIds.addAll(mutableListOf("worker1", "worker2"))
> ```

Populate this from a command-line option defined in `ResourceSetupFlags`. We don't want the IDs to be hard-coded here.



_Code quote:_
```Kotlin
requiredExternalDuchyIds.addAll(mutableListOf("worker1", "worker2"))
```
___
*[`src/main/resources/kingdom/spanner/changelog.yaml` line 34 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/856#-NOHNPAP9llK2AwiEe-z:-NOHNPAQA0r1QBqMGtey:undefined) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/69606747d7d9bcaebe08723936c36fb0cc519b65/src/main/resources/kingdom/spanner/changelog.yaml#L34)):*
> ```YAML
>     relativeToChangeLogFile: true
> - include:
>     file: create-duchy-include-list-schema.sql
> ```





_Suggestion:_
```YAML
create-data-provider-required-duchies-table
```

Marco-Premier (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/856)* status: all files reviewed, 16 unresolved discussions (waiting on @SanjayVas and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/readers/DataProviderReader.kt` line 88 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/856#-NOHDcna3B9j04S9vV9w:-NOKPoAg51cMGrsDd6MG:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/69606747d7d9bcaebe08723936c36fb0cc519b65/src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/readers/DataProviderReader.kt#L88)):*
> ```Kotlin
>       .build()
> 
>   private fun buildExternalDuchyIdList(struct: Struct): Iterable<String?> {
> ```

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/readers/DataProviderReader.kt` line 92 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/856#-NOGzfBq2N4Dpr40tGrb:-NOKQFU1E3cfpzvzN279:bo96yka) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/69606747d7d9bcaebe08723936c36fb0cc519b65/src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/readers/DataProviderReader.kt#L92)):*

Throwing an error seems correct. We shouldn't just silently drop entries that have an invalid ID. If we get a `null` from the lookup, it means something is wrong as an invariant was violated. From the design doc/discussion, Duchies shouldn't get deleted but may be marked inactive.

If we did want to support hard deletion of Duchies we would want to have cascading deletion behavior. Since Duchies is a config file rather than a real table, this behavior would have to be implemented manually.
</blockquote></details>

I thought that if for some reason the lookup fails it would be better to throw an exception since the DP has an invalid "required" duchy. Furthermore I think that the exception would make eventual debugging easier
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/readers/DataProviderReader.kt` line 96 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/856#-NOHDNlSATDbEe4H5q5S:-NOKQ7AhAK1LTj_NxW12:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/69606747d7d9bcaebe08723936c36fb0cc519b65/src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/readers/DataProviderReader.kt#L96)):*

I believe `map` returns a `List` in this case, which already implements `Iterable`.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/writers/CreateDataProvider.kt` line 24 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/856#-NOAs3qd3Zm_m4GcGN0V:-NOKS6Qu7ewd6l7guGI5:b-n62xyi) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/edd625b331d9dd81952750bc6e5ab57c19a90916/src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/writers/CreateDataProvider.kt#L24)):*

can you add a TODO for that @Marco-Premier ?
</blockquote></details>

sure, added it to the SpannerDataProviderService
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/writers/CreateDataProvider.kt` line 45 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/856#-NOHG-QUDj3Oc-siToCh:-NOK5fBn2qnhnTuYEuJe:b-mnt1pt) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/69606747d7d9bcaebe08723936c36fb0cc519b65/src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/writers/CreateDataProvider.kt#L45)):*

This comment doesn't belong here as it is referring to an implementation detail of `DuchyIds.getInternalId`.

(`DuchyIds.getInternalId` also shouldn't have the note about being implemented as a linear scan in its public documentation, or should at least make it clear that it is not part of the contract and may change in the future.)
</blockquote></details>

I see. So no this kind of mentions. I'm going to remove it
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/deploy/tools/CreateResource.kt` line 152 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/856#-NOH5udVEEOnmVLqdLGC:-NOKSRro0OWz4HtcwtsS:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/69606747d7d9bcaebe08723936c36fb0cc519b65/src/main/kotlin/org/wfanet/measurement/kingdom/deploy/tools/CreateResource.kt#L152)):*
> ```Kotlin
>     required = false,
>   )
>   private lateinit var requiredDuchies: Array<String>
> ```

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/deploy/tools/CreateResource.kt` line 164 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/856#-NOH5hmY5-S80pxPzI-W:-NOKS_Ld5sFsO-f7UMmQ:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/69606747d7d9bcaebe08723936c36fb0cc519b65/src/main/kotlin/org/wfanet/measurement/kingdom/deploy/tools/CreateResource.kt#L164)):*
> ```Kotlin
>         }
> 
>       for (duchyExternalId in requiredDuchies) {
> ```

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/internal/testing/ComputationParticipantsServiceTest.kt` line 64 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/856#-NOHHwE60zpVUbacqtUG:-NOKSp0H2xGUwbSkPu3Q:b-d4zhr1) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/69606747d7d9bcaebe08723936c36fb0cc519b65/src/main/kotlin/org/wfanet/measurement/kingdom/service/internal/testing/ComputationParticipantsServiceTest.kt#L64)):*

Why change this? This file appears to be otherwise untouched by this PR.
</blockquote></details>

please see comment below on refactor using shared variables
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/internal/testing/DataProvidersServiceTest.kt` line 123 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/856#-NOH6Q6Y80e1VSl4hZ88:-NOKTx7e0ZdxhH4lfb_r:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/69606747d7d9bcaebe08723936c36fb0cc519b65/src/main/kotlin/org/wfanet/measurement/kingdom/service/internal/testing/DataProvidersServiceTest.kt#L123)):*
> ```Kotlin
>             publicKeySignature = PUBLIC_KEY_SIGNATURE
>           }
>           addAllRequiredExternalDuchyIds(mutableListOf("worker1", "worker2"))
> ```

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/internal/testing/DataProvidersServiceTest.kt` line 192 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/856#-NOHAXnQCkGcuPju1kQo:-NOKU1G_60wWZwCCM6yU:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/69606747d7d9bcaebe08723936c36fb0cc519b65/src/main/kotlin/org/wfanet/measurement/kingdom/service/internal/testing/DataProvidersServiceTest.kt#L192)):*
> ```Kotlin
>             publicKeySignature = PUBLIC_KEY_SIGNATURE
>           }
>           addAllRequiredExternalDuchyIds(mutableListOf("worker1", "worker2"))
> ```

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/internal/testing/MeasurementLogEntriesServiceTest.kt` line 46 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/856#-NOHNyTD8q0frHEY5jXR:-NOKU7LnCrsrG7DvISLI:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/69606747d7d9bcaebe08723936c36fb0cc519b65/src/main/kotlin/org/wfanet/measurement/kingdom/service/internal/testing/MeasurementLogEntriesServiceTest.kt#L46)):*

Same here re: interaction with Population.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/internal/testing/MeasurementsServiceTest.kt` line 72 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/856#-NOHP6r3DJY4VBMYwFKH:-NOKUAl6A7YlFPU8cdiS:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/69606747d7d9bcaebe08723936c36fb0cc519b65/src/main/kotlin/org/wfanet/measurement/kingdom/service/internal/testing/MeasurementsServiceTest.kt#L72)):*

Same here re: interaction with Population.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/internal/testing/Population.kt` line 142 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/856#-NOHItEt5lFdC0BKXcVk:-NOKE9Ck5pRnFmFOKB_J:bq6g59v) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/69606747d7d9bcaebe08723936c36fb0cc519b65/src/main/kotlin/org/wfanet/measurement/kingdom/service/internal/testing/Population.kt#L142)):*

Either this should come from a parameter to `createDataProvider` or the EXTERNAL_DUCHY_IDS constant should be moved to the companion object of this class so the same ones aren't being redefined in multiple tests in such a way that they need to be kept in sync.

If doing the latter, I'd recommend having the constant be ("worker1", "worker2", "aggregator") if you want to match a more realistic configuration. The required Duchies for the DataProvider can still leave out "aggregator". You can also make up this constant from three separate ones so they can be referenced individually without using list indices, e.g. `WORKER1_DUCHY_EXTERNAL_ID`.

Also Kotlin protobuf DSL lists implement the `+=` operator, so the preferred syntax would be either

```kotlin
requiredExternalDuchyIds += listOf("worker1", "worker2")
```

or to save on an extra list creation

```kotlin
requiredExternalDuchyIds += "worker1"
requiredExternalDuchyIds += "worker2"
```
</blockquote></details>

Great, thanks. 
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/internal/testing/RequisitionsServiceTest.kt` line 68 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/856#-NOHL6xv20eCyVv7TuMZ:-NOKJW6A8izZqjUJBqHu:bdt71cr) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/69606747d7d9bcaebe08723936c36fb0cc519b65/src/main/kotlin/org/wfanet/measurement/kingdom/service/internal/testing/RequisitionsServiceTest.kt#L68)):*

Either have `Population.createDataProvider` take the list of Duchies as a param so this can remain unchanged or make use of shared constants.

Note that in the latter case, you can still just use the longer list of Duchies here by e.g. concatenating with a shared list constant.
</blockquote></details>

To prevent the spreading of multiple external_duchy_ids names in different tests, I replaced Buck / Rippon / Shoaks with shared constants too
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/resourcesetup/ResourceSetup.kt` line 245 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/856#-NOHM5aa8qOUV8uiMfjD:-NOKYuiDC6j2ShMHajYT:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/69606747d7d9bcaebe08723936c36fb0cc519b65/src/main/kotlin/org/wfanet/measurement/loadtest/resourcesetup/ResourceSetup.kt#L245)):*

Populate this from a command-line option defined in `ResourceSetupFlags`. We don't want the IDs to be hard-coded here.
</blockquote></details>

Done.
___
*[`src/main/resources/kingdom/spanner/changelog.yaml` line 34 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/856#-NOHNPAP9llK2AwiEe-z:-NOKURwX3u8GidRBhZIz:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/69606747d7d9bcaebe08723936c36fb0cc519b65/src/main/resources/kingdom/spanner/changelog.yaml#L34)):*
> ```YAML
>     relativeToChangeLogFile: true
> - include:
>     file: create-duchy-include-list-schema.sql
> ```

Done.

stevenwarejones (approved):
> Reviewed 25 of 25 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/856)* status: all files reviewed, 15 unresolved discussions (waiting on @SanjayVas)

SanjayVas (changes requested):
> Reviewed 25 of 25 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/856)* status: all files reviewed, 4 unresolved discussions (waiting on @Marco-Premier)

___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/internal/testing/ComputationParticipantsServiceTest.kt` line 64 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/856#-NOHHwE60zpVUbacqtUG:-NOLaM_Z-ykJgucxDYkw:bg002ha) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/69606747d7d9bcaebe08723936c36fb0cc519b65/src/main/kotlin/org/wfanet/measurement/kingdom/service/internal/testing/ComputationParticipantsServiceTest.kt#L64)):*

please see comment below on refactor using shared variables
</blockquote></details>

Sorry, meant to drop this comment when I realized the change to Population.
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/internal/testing/Population.kt` line 74 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/856#-NOLa2yNExX9syxUioZX:-NOLa2yNExX9syxUioZY:bouqbrl) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4ff5d3df04a33d80984b9d34f37f5b741ce85c08/src/main/kotlin/org/wfanet/measurement/kingdom/service/internal/testing/Population.kt#L74)):*
> ```Kotlin
>     const val WORKER2_DUCHY_EXTERNAL_ID = "worker2"
>     const val AGGREGATOR_DUCHY_EXTERNAL_ID = "aggregator"
>     val EXTERNAL_DUCHY_IDS: List<String> by lazy {
> ```

nit: can just make this a list. This is only for testing and the memory overhead of always having this is low. We may not even actually be saving anything by doing this as lazy initialization has some of its own overhead.



_Code quote:_
```Kotlin
by lazy
```
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/internal/testing/RequisitionsServiceTest.kt` line 68 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/856#-NOHL6xv20eCyVv7TuMZ:-NOLalQE1SSCNfruonzG:bj4rc0k) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/69606747d7d9bcaebe08723936c36fb0cc519b65/src/main/kotlin/org/wfanet/measurement/kingdom/service/internal/testing/RequisitionsServiceTest.kt#L68)):*
> I replaced Buck / Rippon / Shoaks with shared constants too

I figured that might end up happening. RIP my easter egg reference to [The Six Duchies](https://robinhobbelderlings.fandom.com/wiki/The_Six_Duchies) (I was reading this book series when I wrote the tests :P )
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/internal/testing/RequisitionsServiceTest.kt` line 703 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/856#-NOLbVDRDQKNvmubfJxD:-NOLbVDRDQKNvmubfJxE:blezd23) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4ff5d3df04a33d80984b9d34f37f5b741ce85c08/src/main/kotlin/org/wfanet/measurement/kingdom/service/internal/testing/RequisitionsServiceTest.kt#L703)):*
> ```Kotlin
>           computedParams = computedRequisitionParams {
>             externalComputationId = measurement.externalComputationId
>             externalFulfillingDuchyId = Population.AGGREGATOR_DUCHY_EXTERNAL_ID
> ```

nit: requisitions would never be fulfilled at the aggregator Duchy, so this might be misleading to someone trying to understand the system via tests. Use one of the workers for `externalFulfillingDuchyId` instead.
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/resourcesetup/ResourceSetupFlags.kt` line 127 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/856#-NOLcuGbC_7cz9jA6Vg9:-NOLcuGbC_7cz9jA6VgA:b-pbnijo) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4ff5d3df04a33d80984b9d34f37f5b741ce85c08/src/main/kotlin/org/wfanet/measurement/loadtest/resourcesetup/ResourceSetupFlags.kt#L127)):*
> ```Kotlin
> 
>   @CommandLine.Option(
>     names = ["--required-duchies"],
> ```

We use repeated options when there multiple values, so it's clearer for the option name to be singular even though the resulting property is a collection and is therefore plural.



_Suggestion:_
```Kotlin
--required-duchy
```
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/resourcesetup/ResourceSetupFlags.kt` line 130 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/856#-NOLcz-d62yUQJ4ZHoq_:-NOLcz-d62yUQJ4ZHoqa:bqkptqa) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4ff5d3df04a33d80984b9d34f37f5b741ce85c08/src/main/kotlin/org/wfanet/measurement/loadtest/resourcesetup/ResourceSetupFlags.kt#L130)):*
> ```Kotlin
>     description =
>       [
>         "The set of duchies externals IDS that must be included in all computations involving this DataProvider."
> ```

nit: ResourceSetup creates 6 DataProviders, so the original comment is a bit inaccurate. Also, we only talk about internal vs. external IDs when referencing the internal API. "Computation" is the Duchy view of a Measurement, so we should reference "Measurement" here instead. Finally, we want to be clear that this is a repeated option.



_Suggestion:_
```Kotlin
      [
        "Duchy ID that must be included in all Measurements for the created DataProvider resources.",
        "This option may be specified multiple times."
      ],
```

stevenwarejones (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/856)* status: all files reviewed, 4 unresolved discussions (waiting on @Marco-Premier)

Marco-Premier (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/856)* status: all files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/internal/testing/RequisitionsServiceTest.kt` line 68 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/856#-NOHL6xv20eCyVv7TuMZ:-NOPOxA27LoWDs0_wlhx:bxp1zgn) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/69606747d7d9bcaebe08723936c36fb0cc519b65/src/main/kotlin/org/wfanet/measurement/kingdom/service/internal/testing/RequisitionsServiceTest.kt#L68)):*

> I replaced Buck / Rippon / Shoaks with shared constants too

I figured that might end up happening. RIP my easter egg reference to [The Six Duchies](https://robinhobbelderlings.fandom.com/wiki/The_Six_Duchies) (I was reading this book series when I wrote the tests :P )
</blockquote></details>

:) I might end up reading it too
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/resourcesetup/ResourceSetupFlags.kt` line 127 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/856#-NOLcuGbC_7cz9jA6Vg9:-NOPQ8AH5LI0JuMq0cjC:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4ff5d3df04a33d80984b9d34f37f5b741ce85c08/src/main/kotlin/org/wfanet/measurement/loadtest/resourcesetup/ResourceSetupFlags.kt#L127)):*

We use repeated options when there multiple values, so it's clearer for the option name to be singular even though the resulting property is a collection and is therefore plural.
</blockquote></details>

Done.

SanjayVas (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/856)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Marco-Premier)

SanjayVas (approved):
> Reviewed 3 of 3 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/856)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Marco-Premier)

Marco-Premier (commented):
Git was unable to merge the branch as it became out dated. I've automatically merged main into this branch using git

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/856#-:-NOsFNokC1nYplasre7j:b-r7to1q)* status: 25 of 26 files reviewed, all discussions resolved (waiting on @SanjayVas and @stevenwarejones)

Marco-Premier (commented):
branch became outdated. I rebased it against main

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/856#-:-NP2S-6pEDa6dk3Emr47:b-ialxiq)* status: 25 of 26 files reviewed, all discussions resolved (waiting on @SanjayVas and @stevenwarejones)

SanjayVas (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/856)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Marco-Premier)

src/main/kotlin/org/wfanet/measurement/kingdom/deploy/tools/CreateResource.kt
```diff
@@ -140,6 +140,16 @@ private class CreateMcCreationTokenCommand : Runnable {
...
+
+  @Option(
+    names = ["--required-duchies"],
+    description =
+      [
+        "The set of duchies externals IDS that must be included in all computations involving this DataProvider"
+      ],
+    required = false,
+  )
+  private lateinit var requiredDuchies: List<String>
```
renjiezh, line 152: This option was noted as optional, thus a default value is required. It raises runtime error "lateinit var not initialized". Could you please fix it? @Marco-Premier

PR #863, opened 2023-02-17
Duchy's failComputationAtDuchy() skips for terminal Computations

renjiezh:


SanjayVas:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/863)

SanjayVas (approved):
> Reviewed 1 of 2 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/863)* status: all files reviewed, 2 unresolved discussions (waiting on @renjiezh)

*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/863#-NOWXltZ8-fuexPMihwE:-NOWXltZ8-fuexPMihwF:b-w9wnc3) (no related file):*
Note that this seems like a band-aid. If the issue is getting an error from the Kingdom because the Computation is in the wrong state for the transition, a better approach would be having the system API Computation.State returned in the error details so that the Duchy can determine whether it should stop (e.g. Duchy can check whether the state is a terminal one so it knows that Fail is unnecessary).

Even that is a band-aid for the particular error we're dealing with, but it's at least a more generally applicable one.

___
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/LiquidLegionsV2Starter.kt` line 49 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/863#-NOWUf9c9pi3QG51lmUH:-NOWUf9c9pi3QG51lmUI:b-tkgmh4) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c17f726e8fad27af744c0540ff2c1b4c83e1a5d5/src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/LiquidLegionsV2Starter.kt#L49)):*
> ```Kotlin
>   private val logger: Logger = Logger.getLogger(this::class.java.name)
> 
>   val terminalStage = Stage.COMPLETE.toProtocolStage()
> ```

nit: use CONSTANT_CASE



_Suggestion:_
```Kotlin
TERMINAL_STAGE
```

duliomatos1 (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/863)* status: all files reviewed, 2 unresolved discussions (waiting on @renjiezh and @stevenwarejones)

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/863)* status: all files reviewed, 2 unresolved discussions (waiting on @stevenwarejones)

PR #865, opened 2023-02-22
Allow Mill to specify duration of work locks.

SanjayVas:
This enables Mill implementations to specify a lock duration depending on factors such the protocol. For the LiquidLegionsV2 Mill, this can be adjusted using the `--work-lock-duration` option.


emadlivani (approved):

wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/865)

PR #861, opened 2023-02-16
Drop allowMpcProtocolsForSingleDataProvider flag and set default to false.

iverson52000:

PR #868, opened 2023-02-24
Stop inserting ComputationParticipants rows for single-EDP R/F Measurements.

### LINKED ISSUE #869: Single-EDP R/F Measurements can result in invalid Duchy computation states
In the case where the Kingdom allows the EDP a choice to fulfill an R/F Measurement either at a Duchy or directly at a Kingdom, computations may be left in an invalid state in Duchies.

In order to support this case, the Kingdom must start the internal Measurement in the `PENDING_REQUISITION_PARAMS` state and insert `ComputationParticipants` rows. This is because Duchies need to set Requisition params for the LLv2 protocol before Requisitions are made available to EDPs. Otherwise an EDP cannot fulfill the Requisition at a Duchy using the LLv2 protocol.

If an EDP then fulfills such a Requisition directly at the Kingdom, the internal Measurement is transitioned to the `SUCCEEDED` state. Participant Duchies do not currently handle this transition, as they expect it to only occur if they complete the MPC computation.

Rather than updating Duchies to handle this case, we can simply only allow single-EDP R/F Measurements to be fulfilled directly. Indeed, allowing EDPs flexibility in how they fulfill these was only intended to be temporary to give EDPs time to implement direct fulfillment.

https://rally1.rallydev.com/#/?detail=/defect/687779117107&fdp=true

SanjayVas:
This also stops populating the liquid_legions_v2 field in Measurement.Details.protocol_config for these Measurements. As a result, the deprecated liquid_legions_v2 field in the ProtocolConfig message of the public API will no longer be populated.

This was missed in #861.
Fixes #869.
https://rally1.rallydev.com/#/?detail=/defect/687779117107&fdp=true

### DIFF
--- a/src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/writers/CreateMeasurement.kt
+++ b/src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/writers/CreateMeasurement.kt
@@ -69,14 +69,12 @@ class CreateMeasurement(private val measurement: Measurement) :
       }
     }
 
-    // protocol has to be set for the measurement to require computation
-    return if (
-      measurement.details.protocolConfig.protocolCase !=
-        ProtocolConfig.ProtocolCase.PROTOCOL_NOT_SET
-    ) {
-      createComputedMeasurement(measurement, measurementConsumerId)
-    } else {
-      createDirectMeasurement(measurement, measurementConsumerId)
+    @Suppress("WHEN_ENUM_CAN_BE_NULL_IN_JAVA") // Protobuf enum fields are never null.
+    return when (measurement.details.protocolConfig.protocolCase) {
+      ProtocolConfig.ProtocolCase.LIQUID_LEGIONS_V2 ->
+        createComputedMeasurement(measurement, measurementConsumerId)
+      ProtocolConfig.ProtocolCase.PROTOCOL_NOT_SET ->
+        createDirectMeasurement(measurement, measurementConsumerId)
     }
   }
 
--- a/src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/ProtoConversions.kt
+++ b/src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/ProtoConversions.kt
@@ -133,62 +133,66 @@ fun InternalDifferentialPrivacyParams.toDifferentialPrivacyParams(): Differentia
   }
 }
 
-/** @throws [IllegalStateException] if MeasurementType not specified */
 fun InternalProtocolConfig.toProtocolConfig(
   measurementTypeCase: MeasurementSpec.MeasurementTypeCase,
   dataProviderCount: Int,
 ): ProtocolConfig {
-  val source = this
+  require(dataProviderCount >= 1) { "Expected at least one DataProvider" }
 
+  val source = this
   return protocolConfig {
     name =
       if (source.externalProtocolConfigId.isNotEmpty())
         ProtocolConfigKey(source.externalProtocolConfigId).toName()
       else ProtocolConfigKey(Direct.getDescriptor().name).toName()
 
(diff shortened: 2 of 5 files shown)


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/868)

SanjayVas (commented):
I mean that it can no longer be populated for new single-EDP R/F measurements. Since it's a deprecated field that was intended to be removed, it's just never populated at all anymore. For multi-EDP R/F measurements, the LLv2 protocol config can be found in the `protocols` field.

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/868#-:-NPJGDQf1tHHil3vvwB8:bjnk31k)* status: 1 of 5 files reviewed, all discussions resolved (waiting on @iverson52000)

stevenwarejones (changes requested):
> Reviewed 2 of 2 files at r1, 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/868)* status: all files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/ProtoConversions.kt` line 183 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/868#-NPJPo8QBwO20s1XB4xK:-NPJPo8QBwO20s1XB4xL:bg6rica) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/73231d3e77f5c9d15bd32065208858dbce40fe22/src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/ProtoConversions.kt#L183)):*
> ```Kotlin
>           }
>         }
>         if (dataProviderCount == 1) {
> ```

i think the code reads slightly easier to move this up and make it an `if else`.
```
if (dataProviderCount == 1) {
   protocols += protocol { direct = direct {} }
} else {
    when (source.protocolCase) {     
       LIQUID_LEGIONS_V2 -> {
             ....
        } 
        else -> error("Invalid Protocol")
   }
}
```

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/868)* status: all files reviewed, 1 unresolved discussion (waiting on @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/ProtoConversions.kt` line 183 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/868#-NPJPo8QBwO20s1XB4xK:-NPJd8tI2oFvc9ZI3Zx0:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/73231d3e77f5c9d15bd32065208858dbce40fe22/src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/ProtoConversions.kt#L183)):*

i think the code reads slightly easier to move this up and make it an `if else`.
```
if (dataProviderCount == 1) {
   protocols += protocol { direct = direct {} }
} else {
    when (source.protocolCase) {     
       LIQUID_LEGIONS_V2 -> {
             ....
        } 
        else -> error("Invalid Protocol")
   }
}
```

</blockquote></details>

Done.

stevenwarejones (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/868)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @SanjayVas)

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/868)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @SanjayVas)

PR #872, opened 2023-03-01
Define internal Measurements Service api retention methods

jcorilla:
deletion of completed measurements and cancellation of pending measurements

### DIFF
--- a/src/main/proto/wfa/measurement/internal/kingdom/measurements_service.proto
+++ b/src/main/proto/wfa/measurement/internal/kingdom/measurements_service.proto
@@ -49,7 +49,7 @@ service Measurements {
       returns (CancelPendingMeasurementsResponse);
 
   // Permanently deletes Measurements based on update_time for states:
-  // SUCCEEDED, FAILED, CANCELLED, and STATE_UNSPECIFIED.
+  // SUCCEEDED, FAILED, and CANCELLED.
   rpc DeleteCompletedMeasurements(DeleteCompletedMeasurementsRequest)
       returns (DeleteCompletedMeasurementsResponse);
 }
@@ -121,9 +121,9 @@ message DeleteCompletedMeasurementsRequest {
 }
 
 message CancelPendingMeasurementsResponse {
-  repeated Measurement measurements = 1;
+  repeated fixed64 external_measurement_ids = 1;
 }
 
 message DeleteCompletedMeasurementsResponse {
-  repeated Measurement measurements = 1;
+  repeated fixed64 external_measurement_ids = 1;
 }
--- a/src/main/proto/wfa/measurement/internal/kingdom/measurements_service.proto
+++ b/src/main/proto/wfa/measurement/internal/kingdom/measurements_service.proto
@@ -50,8 +50,8 @@ service Measurements {
 
   // Permanently deletes a limited amount of Measurements based on update_time
   // for states: SUCCEEDED, FAILED, and CANCELLED.
-  rpc DeleteCompletedMeasurements(DeleteCompletedMeasurementsRequest)
-      returns (DeleteCompletedMeasurementsResponse);
+  rpc PurgeCompletedMeasurements(PurgeCompletedMeasurementsRequest)
+      returns (PurgeCompletedMeasurementsResponse);
 }
 
 message GetMeasurementRequest {
@@ -112,12 +112,20 @@ message CancelMeasurementRequest {
 
 message CancelPendingMeasurementsRequest {
   google.protobuf.Timestamp created_before = 1;
-  bool dry_run = 2;
+
+  // Execute the state transitions to CANCELLED. If `force` is set to false,
+  // the method will return a sample of measurement keys that would be cancelled
+  // without transitioning their states.
+  bool force = 2;
 }
 
-message DeleteCompletedMeasurementsRequest {
+message PurgeCompletedMeasurementsRequest {
   google.protobuf.Timestamp updated_before = 1;
-  bool dry_run = 2;
+
+  // Execute the deletion. If `force` is set to false, the method will return
+  // a sample of measurement keys that would be deleted without executing
+  // deletions.
+  bool force = 2;
 }
 
 message CancelPendingMeasurementsResponse {
@@ -128,10 +136,14 @@ message CancelPendingMeasurementsResponse {
   repeated MeasurementKey measurement_keys = 1;
 }
 
-message DeleteCompletedMeasurementsResponse {
+message PurgeCompletedMeasurementsResponse {
+  int32 purge_count = 1;
+
   message MeasurementKey {
     fixed64 external_measurement_consumer_id = 1;
     fixed64 external_measurement_id = 2;
   }
-  repeated MeasurementKey measurement_keys = 1;
+  // A sample of measurement keys that would be deleted. The sample size is at
+  // maximum 100 measurement keys. Only populated if `force` is false.
+  repeated MeasurementKey purge_sample = 2;
 }
--- a/src/main/proto/wfa/measurement/internal/kingdom/measurements_service.proto
+++ b/src/main/proto/wfa/measurement/internal/kingdom/measurements_service.proto
@@ -128,21 +128,18 @@ message PurgeCompletedMeasurementsRequest {
   bool force = 2;
 }
 
+message MeasurementKey {
+  fixed64 external_measurement_consumer_id = 1;
+  fixed64 external_measurement_id = 2;
+}
+
 message CancelPendingMeasurementsResponse {
-  message MeasurementKey {
-    fixed64 external_measurement_consumer_id = 1;
-    fixed64 external_measurement_id = 2;
-  }
   repeated MeasurementKey measurement_keys = 1;
 }
 
 message PurgeCompletedMeasurementsResponse {
   int32 purge_count = 1;
 
-  message MeasurementKey {
-    fixed64 external_measurement_consumer_id = 1;
-    fixed64 external_measurement_id = 2;
-  }
   // A sample of measurement keys that would be deleted. The sample size is at
   // maximum 100 measurement keys. Only populated if `force` is false.
   repeated MeasurementKey purge_sample = 2;
--- a/src/main/proto/wfa/measurement/internal/kingdom/measurements_service.proto
+++ b/src/main/proto/wfa/measurement/internal/kingdom/measurements_service.proto
@@ -43,13 +43,13 @@ service Measurements {
   // Marks a Measurement as CANCELLED by transitioning its state.
   rpc CancelMeasurement(CancelMeasurementRequest) returns (Measurement);
 
-  // Marks Measurements as CANCELLED by transitioning their states based on
-  // create_time for states beginning with 'PENDING'.
+  // Marks a limited amount of Measurements as CANCELLED by transitioning their
+  // states based on create_time for states beginning with 'PENDING'.
   rpc CancelPendingMeasurements(CancelPendingMeasurementsRequest)
       returns (CancelPendingMeasurementsResponse);
 
-  // Permanently deletes Measurements based on update_time for states:
-  // SUCCEEDED, FAILED, and CANCELLED.
+  // Permanently deletes a limited amount of Measurements based on update_time
+  // for states: SUCCEEDED, FAILED, and CANCELLED.
   rpc DeleteCompletedMeasurements(DeleteCompletedMeasurementsRequest)
       returns (DeleteCompletedMeasurementsResponse);
 }
@@ -121,9 +121,17 @@ message DeleteCompletedMeasurementsRequest {
 }
 
 message CancelPendingMeasurementsResponse {
-  repeated fixed64 external_measurement_ids = 1;
+  message MeasurementKey {
+    fixed64 external_measurement_consumer_id = 1;
+    fixed64 external_measurement_id = 2;
+  }
+  repeated MeasurementKey measurement_keys = 1;
 }
 
 message DeleteCompletedMeasurementsResponse {
-  repeated fixed64 external_measurement_ids = 1;
+  message MeasurementKey {
+    fixed64 external_measurement_consumer_id = 1;
+    fixed64 external_measurement_id = 2;
+  }
+  repeated MeasurementKey measurement_keys = 1;
 }
--- a/src/main/proto/wfa/measurement/internal/kingdom/measurements_service.proto
+++ b/src/main/proto/wfa/measurement/internal/kingdom/measurements_service.proto
@@ -42,6 +42,16 @@ service Measurements {
 
   // Marks a Measurement as CANCELLED by transitioning its state.
   rpc CancelMeasurement(CancelMeasurementRequest) returns (Measurement);
+
+  // Marks Measurements as CANCELLED by transitioning their states based on
+  // create_time for states beginning with 'PENDING'.
+  rpc CancelPendingMeasurements(CancelPendingMeasurementsRequest)
+      returns (CancelPendingMeasurementsResponse);
+
+  // Permanently deletes Measurements based on update_time for states:
+  // SUCCEEDED, FAILED, CANCELLED, and STATE_UNSPECIFIED.
+  rpc DeleteCompletedMeasurements(DeleteCompletedMeasurementsRequest)
+      returns (DeleteCompletedMeasurementsResponse);
 }
 
 message GetMeasurementRequest {
@@ -98,4 +108,22 @@ message SetMeasurementResultRequest {
 message CancelMeasurementRequest {
   fixed64 external_measurement_consumer_id = 1;
   fixed64 external_measurement_id = 2;
+}
+
+message CancelPendingMeasurementsRequest {
+  google.protobuf.Timestamp created_before = 1;
+  bool dry_run = 2;
+}
+
+message DeleteCompletedMeasurementsRequest {
+  google.protobuf.Timestamp updated_before = 1;
+  bool dry_run = 2;
+}
+
+message CancelPendingMeasurementsResponse {
+  repeated Measurement measurements = 1;
+}
+
+message DeleteCompletedMeasurementsResponse {
+  repeated Measurement measurements = 1;
 }


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/872)

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/872)* status: all files reviewed, 2 unresolved discussions (waiting on @jcorilla, @kungfucraig, and @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/internal/kingdom/measurements_service.proto` line 52 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/872#-NPUC9pG7Hz-_iwU0FZT:-NPUC9pG7Hz-_iwU0FZU:bo4ob3v) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/21a0327537b2a2ab7d1a4a5e27c9b04000225984/src/main/proto/wfa/measurement/internal/kingdom/measurements_service.proto#L52)):*
> ```protobuf
> 
>   // Permanently deletes Measurements based on update_time for states:
>   // SUCCEEDED, FAILED, CANCELLED, and STATE_UNSPECIFIED.
> ```

A Measurement should never be in this state to begin with. If you find one, it means some invariant was violated elsewhere. We shouldn't mention that state here.



_Code quote:_
```protobuf
, and STATE_UNSPECIFIED.
```
___
*[`src/main/proto/wfa/measurement/internal/kingdom/measurements_service.proto` line 124 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/872#-NPUCWzZ4o73MRmV0jCE:-NPUCWzZ4o73MRmV0jCF:b-hvn23y) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/21a0327537b2a2ab7d1a4a5e27c9b04000225984/src/main/proto/wfa/measurement/internal/kingdom/measurements_service.proto#L124)):*
> ```protobuf
> 
> message CancelPendingMeasurementsResponse {
>   repeated Measurement measurements = 1;
> ```

@kungfucraig Do we want to return the entire Measurement as opposed to just the IDs? It's a heftier query due to the joins.

I'm also thinking about the number of rows this might affect and whether the implementation might want to use [Partitioned DML](https://cloud.google.com/spanner/docs/dml-partitioned). In either case, it's more efficient if we can do a single DML statement and utilize the [THEN RETURN clause](https://cloud.google.com/spanner/docs/reference/standard-sql/dml-syntax#then_return_2) of the DELETE statement.

kungfucraig (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/872)* status: all files reviewed, 2 unresolved discussions (waiting on @jcorilla, @SanjayVas, and @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/internal/kingdom/measurements_service.proto` line 124 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/872#-NPUCWzZ4o73MRmV0jCE:-NPUIVfX1DODOaNi-E8_:bofhaa8) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/21a0327537b2a2ab7d1a4a5e27c9b04000225984/src/main/proto/wfa/measurement/internal/kingdom/measurements_service.proto#L124)):*

@kungfucraig Do we want to return the entire Measurement as opposed to just the IDs? It's a heftier query due to the joins.

I'm also thinking about the number of rows this might affect and whether the implementation might want to use [Partitioned DML](https://cloud.google.com/spanner/docs/dml-partitioned). In either case, it's more efficient if we can do a single DML statement and utilize the [THEN RETURN clause](https://cloud.google.com/spanner/docs/reference/standard-sql/dml-syntax#then_return_2) of the DELETE statement.
</blockquote></details>

Any issue just returning IDs? In dry-run mode we might dereference each and log them, but in wet run, we'd probably just echo the count (as they are deleted anyway).

jcorilla (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/872)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @SanjayVas and @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/internal/kingdom/measurements_service.proto` line 52 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/872#-NPUC9pG7Hz-_iwU0FZT:-NPUt1S-4SUG0AIx22jD:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/21a0327537b2a2ab7d1a4a5e27c9b04000225984/src/main/proto/wfa/measurement/internal/kingdom/measurements_service.proto#L52)):*

A Measurement should never be in this state to begin with. If you find one, it means some invariant was violated elsewhere. We shouldn't mention that state here.
</blockquote></details>

Done.
___
*[`src/main/proto/wfa/measurement/internal/kingdom/measurements_service.proto` line 124 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/872#-NPUCWzZ4o73MRmV0jCE:-NPUt22x6zzqLf1F9yBp:bht1xr7) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/21a0327537b2a2ab7d1a4a5e27c9b04000225984/src/main/proto/wfa/measurement/internal/kingdom/measurements_service.proto#L124)):*

Any issue just returning IDs? In dry-run mode we might dereference each and log them, but in wet run, we'd probably just echo the count (as they are deleted anyway).
</blockquote></details>

Changed response type to ids.

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/872)* status: all files reviewed, 1 unresolved discussion (waiting on @jcorilla, @kungfucraig, and @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/internal/kingdom/measurements_service.proto` line 124 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/872#-NPUCWzZ4o73MRmV0jCE:-NPUsZlD1AkOyZ0YD9RA:b-1ayzlj) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/21a0327537b2a2ab7d1a4a5e27c9b04000225984/src/main/proto/wfa/measurement/internal/kingdom/measurements_service.proto#L124)):*
> Changed response type to ids.

`external_measurement_id` is not sufficient to uniquely identify a Measurement. That ID is only guaranteed to be unique in the scope of a given MC. Therefore you need (`external_measurement_consumer_id`, `external_measurement_id`) tuples.

> Any issue just returning IDs? In dry-run mode we might dereference each and log them, but in wet run, we'd probably just echo the count (as they are deleted anyway).

Ah, I was thinking we'd want consistent behavior from this method regardless of whether `--dry-run` is used. If we want, the caller (cron job) can dereference after (we'd maybe want to add a batch get method to make that more efficient).

Returning IDs should be fine even without '--dry-run` as long as we don't need to use partitioned DML (i.e. we're mutating fewer than 40k rows, including indexes). That can just use the delete w/THEN RETURN as mentioned above. Note that even if we need to do more rows, we can batch by using LIMIT in our DML statement. In that case though we may want to stream the deleted IDs in batches so it's clear it's not a single transaction (which would also be the case with partitioned DML).

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/872)* status: all files reviewed, 1 unresolved discussion (waiting on @jcorilla, @kungfucraig, and @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/internal/kingdom/measurements_service.proto` line 124 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/872#-NPUCWzZ4o73MRmV0jCE:-NPUviE34XfSAicY90Il:bgevoex) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/21a0327537b2a2ab7d1a4a5e27c9b04000225984/src/main/proto/wfa/measurement/internal/kingdom/measurements_service.proto#L124)):*

> Changed response type to ids.

`external_measurement_id` is not sufficient to uniquely identify a Measurement. That ID is only guaranteed to be unique in the scope of a given MC. Therefore you need (`external_measurement_consumer_id`, `external_measurement_id`) tuples.

> Any issue just returning IDs? In dry-run mode we might dereference each and log them, but in wet run, we'd probably just echo the count (as they are deleted anyway).

Ah, I was thinking we'd want consistent behavior from this method regardless of whether `--dry-run` is used. If we want, the caller (cron job) can dereference after (we'd maybe want to add a batch get method to make that more efficient).

Returning IDs should be fine even without '--dry-run` as long as we don't need to use partitioned DML (i.e. we're mutating fewer than 40k rows, including indexes). That can just use the delete w/THEN RETURN as mentioned above. Note that even if we need to do more rows, we can batch by using LIMIT in our DML statement. In that case though we may want to stream the deleted IDs in batches so it's clear it's not a single transaction (which would also be the case with partitioned DML).

</blockquote></details>

Alternatively instead of streaming batches, we can just limit the number of mutations per method call. Basically a single call can still be a single transaction, but you may need to make multiple calls to get them all (e.g. call repeatedly until you affect 0 rows).

kungfucraig (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/872)* status: all files reviewed, 1 unresolved discussion (waiting on @SanjayVas and @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/internal/kingdom/measurements_service.proto` line 124 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/872#-NPUCWzZ4o73MRmV0jCE:-NPYu-iiASgAqRAsBogM:br4vs30) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/21a0327537b2a2ab7d1a4a5e27c9b04000225984/src/main/proto/wfa/measurement/internal/kingdom/measurements_service.proto#L124)):*

Alternatively instead of streaming batches, we can just limit the number of mutations per method call. Basically a single call can still be a single transaction, but you may need to make multiple calls to get them all (e.g. call repeatedly until you affect 0 rows).
</blockquote></details>

I don't think BatchGet is necessary. Dry run mode should not be used much, and we can afford to just "for each ID GetMeasurement(ID)". For the wet run we can print the IDs. We could decide to just print the IDs for dry run too for that matter.

Let's try to keep this simple. :)

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/872)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @jcorilla and @SanjayVas)

___
*[`src/main/proto/wfa/measurement/internal/kingdom/measurements_service.proto` line 115 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/872#-NPrt7bqC0-6gQpYx8in:-NPrt7bqC0-6gQpYx8io:brzuk7t) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/55a9ca61e1148e5aa34397ecbd252cbc4b05e8d9/src/main/proto/wfa/measurement/internal/kingdom/measurements_service.proto#L115)):*
> ```protobuf
> message CancelPendingMeasurementsRequest {
>   google.protobuf.Timestamp created_before = 1;
>   bool dry_run = 2;
> ```

why are we putting dry_run in the api itself? What are the use cases?

jcorilla (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/872)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @SanjayVas and @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/internal/kingdom/measurements_service.proto` line 124 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/872#-NPUCWzZ4o73MRmV0jCE:-NPsFEqN1CcsoDT2ia-1:b-uww83a) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/21a0327537b2a2ab7d1a4a5e27c9b04000225984/src/main/proto/wfa/measurement/internal/kingdom/measurements_service.proto#L124)):*

I don't think BatchGet is necessary. Dry run mode should not be used much, and we can afford to just "for each ID GetMeasurement(ID)". For the wet run we can print the IDs. We could decide to just print the IDs for dry run too for that matter.

Let's try to keep this simple. :)
</blockquote></details>

Updated the comment above the rpc definition to specify the new apis will apply a limited number of mutations (to avoid streaming batches). So it would be on the caller to make multiple calls if needed.
___
*[`src/main/proto/wfa/measurement/internal/kingdom/measurements_service.proto` line 115 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/872#-NPrt7bqC0-6gQpYx8in:-NPsHk092-qDfdbj36Ep:b-9t881o) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/55a9ca61e1148e5aa34397ecbd252cbc4b05e8d9/src/main/proto/wfa/measurement/internal/kingdom/measurements_service.proto#L115)):*

why are we putting dry_run in the api itself? What are the use cases?
</blockquote></details>

The use case was to safeguard when deploying to prod or even the qa environment. Another deployment which cleaned up Duchy Computations deleted more entries than expected - so a dry-run flag was suggested as a way to prevent that.

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/872)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/internal/kingdom/measurements_service.proto` line 115 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/872#-NPrt7bqC0-6gQpYx8in:-NPsNwYbAJ9LRfs39fgO:b-cwrnla) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/55a9ca61e1148e5aa34397ecbd252cbc4b05e8d9/src/main/proto/wfa/measurement/internal/kingdom/measurements_service.proto#L115)):*
<details><summary><i>Previously, jcorilla wrote…</i></summary><blockquote>

The use case was to safeguard when deploying to prod or even the qa environment. Another deployment which cleaned up Duchy Computations deleted more entries than expected - so a dry-run flag was suggested as a way to prevent that. 
</blockquote></details>

To clarify why it's in the internal API, it's due to the fact that we cannot have transactional behavior across multiple calls. Therefore this is passed in to the method that performs the single transaction.

SanjayVas (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/872)* status: all files reviewed, 3 unresolved discussions (waiting on @jcorilla and @stevenwarejones)

*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/872#-NPxJFAQ7TjyE9eCQJFV:-NPxJFAR8398tXGoUgYR:b-a0nq25) (no related file):*
FWIW, I think I still prefer the pattern of using separate List, Cancel, and Delete methods (adding Batch versions for efficiency as desired). If we add an `etag` mechanism (which for anything with an `update_time` can just be based on that), then we can guarantee that the resource wasn't changed between the List and Delete/Cancel operations.

___
___
*[`src/main/proto/wfa/measurement/internal/kingdom/measurements_service.proto` line 132 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/872#-NPxH_P70MaiuEbPk-Eb:-NPxH_P8-6LM_gvokkgi:b-tsi5q3) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/bf3ef1ff3e5df003ea22259c838c5decd2a67414/src/main/proto/wfa/measurement/internal/kingdom/measurements_service.proto#L132)):*
> ```protobuf
> 
> message CancelPendingMeasurementsResponse {
>   message MeasurementKey {
> ```

nit: define a single shared MeasurementKey message for both responses. The message can just be defined in this file since it's only used in responses.

SanjayVas (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/872)* status: all files reviewed, 3 unresolved discussions (waiting on @jcorilla and @stevenwarejones)

*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/872#-NPxJFAQ7TjyE9eCQJFV:-NPxJvWfCWJK-P5jcKcX:b-gtbvb5) (no related file):*

FWIW, I think I still prefer the pattern of using separate List, Cancel, and Delete methods (adding Batch versions for efficiency as desired). If we add an `etag` mechanism (which for anything with an `update_time` can just be based on that), then we can guarantee that the resource wasn't changed between the List and Delete/Cancel operations.
</blockquote></details>

@kungfucraig for opinion on the above, though I believe the stated preference was to keep the criteria-based methods.

___

stevenwarejones (approved):
> Reviewed 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/872)* status: all files reviewed, 1 unresolved discussion (waiting on @jcorilla)

jcorilla (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/872)* status: all files reviewed, 1 unresolved discussion (waiting on @kungfucraig)

*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/872#-NPxJFAQ7TjyE9eCQJFV:-NQaHMFL3fENYOorm4Di:bwuu951) (no related file):*

@kungfucraig for opinion on the above, though I believe the stated preference was to keep the criteria-based methods.
</blockquote></details>

Will be submitting new PRs to move from criteria based delete to using List - Cancel/Delete methods. This avoids adding the new purge method (https://google.aip.dev/165)

___

PR #875, opened 2023-03-02
Add single log entry for computation failure

Marco-Premier:
1. Unified StateTransitionMeasurementLogEntry and DuchyMeasurementLogEntry in the same MeasurementLogEntry when computation participant fails
2. deprecated Measurement.Failure proto fields since now failures are set in log messages


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/875)

SanjayVas (changes requested):
> Reviewed 5 of 13 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/875)* status: 5 of 13 files reviewed, 3 unresolved discussions (waiting on @Marco-Premier and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/MillBase.kt` line 225 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/875#-NPZZSNcBJe83uow_YcU:-NPZZSNcBJe83uow_YcV:b-6dcp6c) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ccfb7467cdf1b477c39f4ae8835640678e257746/src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/MillBase.kt#L225)):*
> ```Kotlin
>       else -> {
>         // Treat all other errors as transient.
>         logger.log(Level.WARNING, "$globalId@$millId: TRANSIENT error", e)
> ```

nit: move these lines into the `else` block as well



_Code quote:_
```Kotlin
        // Treat all other errors as transient.
        logger.log(Level.WARNING, "$globalId@$millId: TRANSIENT error", e)
```
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/writers/UpdateMeasurementState.kt` line 34 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/875#-NPZa5SM9fbROzZmwMYN:-NPZa5SM9fbROzZmwMYO:biacxpu) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ccfb7467cdf1b477c39f4ae8835640678e257746/src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/writers/UpdateMeasurementState.kt#L34)):*
> ```Kotlin
>   previousState: Measurement.State,
>   measurementLogEntryDetails: MeasurementLogEntry.Details,
>   duchyMeasurementLogEntry: DuchyMeasurementLogEntry? = null
> ```

We prefer not to use partially-filled protobuf messages. The only case we do so is in request messages, where the field behavior (e.g. whether it's output-only) tends to be more clearly documented.

Perhaps just call `insertDuchyMeasurementLogEntry` from the call site of `updateMeasurementState`.
___
*[`src/main/proto/wfa/measurement/internal/kingdom/measurement.proto` line 131 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/875#-NPZcJ646OVLM1Um2RY5:-NPZcJ646OVLM1Um2RY6:b-v6xpau) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ccfb7467cdf1b477c39f4ae8835640678e257746/src/main/proto/wfa/measurement/internal/kingdom/measurement.proto#L131)):*
> ```protobuf
>     // DEPRECATED: failure information are available in the corresponding
>     // MeasurementLogEntries row
>     Failure failure = 8 [deprecated = true];
> ```

I believe this field is still read in some places. You'll need to update those as well. For example, `org.wfanet.measurement.kingdom.service.api.v2alpha.toMeasurement`.

Note that this means you still need the failure information, so while this field is deprecated you'll need to add a new output-only version of the field in `Measurement` itself. Basically the idea is that this is no longer in `Measurement.Details` and therefore no longer needs to be persisted, but the data still needs to be read and returned from MeasurementReader.

Marco-Premier (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/875)* status: 3 of 13 files reviewed, 2 unresolved discussions (waiting on @SanjayVas and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/writers/UpdateMeasurementState.kt` line 34 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/875#-NPZa5SM9fbROzZmwMYN:-NPagUkF935EcKbpdAYX:bbenc9h) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ccfb7467cdf1b477c39f4ae8835640678e257746/src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/writers/UpdateMeasurementState.kt#L34)):*

We prefer not to use partially-filled protobuf messages. The only case we do so is in request messages, where the field behavior (e.g. whether it's output-only) tends to be more clearly documented.

Perhaps just call `insertDuchyMeasurementLogEntry` from the call site of `updateMeasurementState`.
</blockquote></details>

I moved it into ```FailComputationParticipant```, which is the only class that needs to ```insertDuchyMeasurementLogEntry``` along with ```CreateDuchyMeasurementLogEntry```
___
*[`src/main/proto/wfa/measurement/internal/kingdom/measurement.proto` line 131 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/875#-NPZcJ646OVLM1Um2RY5:-NPauFmKDx6qg1p0fPUF:b-duwdyk) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ccfb7467cdf1b477c39f4ae8835640678e257746/src/main/proto/wfa/measurement/internal/kingdom/measurement.proto#L131)):*

I believe this field is still read in some places. You'll need to update those as well. For example, `org.wfanet.measurement.kingdom.service.api.v2alpha.toMeasurement`.

Note that this means you still need the failure information, so while this field is deprecated you'll need to add a new output-only version of the field in `Measurement` itself. Basically the idea is that this is no longer in `Measurement.Details` and therefore no longer needs to be persisted, but the data still needs to be read and returned from MeasurementReader.
</blockquote></details>

I think I need to clarify this more:
Right now we store all the logging information (including failures) into the ```MeasurementLogEntries``` table and childs.

I thought we aimed to remove any failure information from the ```measurement``` message and leave them only into log messages.

I understand your suggestions, however ```measurement_log_entry.proto``` contains ```ErrorDetails``` which differs from the "deprecated" ```Failure``` message in ```measurement.proto```. 
This means that, if the goal is to keep failure info into ```measurement```, I may be reading failures info from logging tables and slightly change the "output-only" Failure field I should add into ```measurement```.

Is that what you meant @SanjayVas ?
thanks

SanjayVas (changes requested):
> Reviewed 3 of 13 files at r1, 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/875)* status: 9 of 13 files reviewed, 1 unresolved discussion (waiting on @Marco-Premier and @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/internal/kingdom/measurement.proto` line 131 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/875#-NPZcJ646OVLM1Um2RY5:-NPs8FrR-_QqcDv12BT4:b-u6p5v9) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ccfb7467cdf1b477c39f4ae8835640678e257746/src/main/proto/wfa/measurement/internal/kingdom/measurement.proto#L131)):*
> This means that, if the goal is to keep failure info into measurement, I may be reading failures info from logging tables and slightly change the "output-only" Failure field I should add into measurement.

Yes, you got it. That's precisely it.

stevenwarejones (approved):
> Reviewed 10 of 13 files at r1, 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/875)* status: all files reviewed, 1 unresolved discussion (waiting on @Marco-Premier)

Marco-Premier (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/875)* status: all files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

___
*[`src/main/proto/wfa/measurement/internal/kingdom/measurement.proto` line 131 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/875#-NPZcJ646OVLM1Um2RY5:-NRXHW5IASw2_5TERbdn:b-7g4r4m) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ccfb7467cdf1b477c39f4ae8835640678e257746/src/main/proto/wfa/measurement/internal/kingdom/measurement.proto#L131)):*

> This means that, if the goal is to keep failure info into measurement, I may be reading failures info from logging tables and slightly change the "output-only" Failure field I should add into measurement.

Yes, you got it. That's precisely it.
</blockquote></details>

A bit late but want to close this one @SanjayVas .
We don't store the reason of a failure in Measurement Log Entry tables. So if we deprecate and remove the Failure filed from measurement details, there's no way to return the ```Reason```  through public API. In other words, we don't know if a measurement failed because a certificate was revoked  or a requisition was refused etc.

Either we store this info in the MeasurementLogEntry table and read the info from there or we leave the Failure object inside Measurement .Details as it is and keep storing it into the DB.

I think we should proceed with the later and not deprecate the Failure message, but open to different suggestion.
Thanks

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/875)* status: all files reviewed, 1 unresolved discussion (waiting on @Marco-Premier)

___
*[`src/main/proto/wfa/measurement/internal/kingdom/measurement.proto` line 131 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/875#-NPZcJ646OVLM1Um2RY5:-NRZCKYE731bktOMnuQH:bhoppr6) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ccfb7467cdf1b477c39f4ae8835640678e257746/src/main/proto/wfa/measurement/internal/kingdom/measurement.proto#L131)):*

A bit late but want to close this one @SanjayVas .
We don't store the reason of a failure in Measurement Log Entry tables. So if we deprecate and remove the Failure filed from measurement details, there's no way to return the ```Reason```  through public API. In other words, we don't know if a measurement failed because a certificate was revoked  or a requisition was refused etc.

Either we store this info in the MeasurementLogEntry table and read the info from there or we leave the Failure object inside Measurement .Details as it is and keep storing it into the DB.

I think we should proceed with the later and not deprecate the Failure message, but open to different suggestion.
Thanks
</blockquote></details>

I'd prefer persisting this information in the MeasurementLogEntries table, but we can just leave it as-is for now as that would take more work to do and it's not a significant priority. Regardless, the public API should remain unchanged.

Marco-Premier (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/875)* status: 5 of 13 files reviewed, 1 unresolved discussion (waiting on @SanjayVas and @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/internal/kingdom/measurement.proto` line 131 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/875#-NPZcJ646OVLM1Um2RY5:-NRcC6vXE1zozx-vVlTS:b7jcgbu) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ccfb7467cdf1b477c39f4ae8835640678e257746/src/main/proto/wfa/measurement/internal/kingdom/measurement.proto#L131)):*

I'd prefer persisting this information in the MeasurementLogEntries table, but we can just leave it as-is for now as that would take more work to do and it's not a significant priority. Regardless, the public API should remain unchanged.
</blockquote></details>

Fair enough, I enable back MeasurementDetails to store the cause of failures.
Also I closed the PR for public API which will not be changed.

stevenwarejones (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/875)* status: all files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

Marco-Premier (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/875)* status: all files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

___
*[`src/main/proto/wfa/measurement/internal/kingdom/measurement.proto` line 131 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/875#-NPZcJ646OVLM1Um2RY5:-NSFl58HE0tM8p64vnAz:b-wmsb59) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ccfb7467cdf1b477c39f4ae8835640678e257746/src/main/proto/wfa/measurement/internal/kingdom/measurement.proto#L131)):*

Fair enough, I enable back MeasurementDetails to store the cause of failures.
Also I closed the PR for public API which will not be changed.
</blockquote></details>

@SanjayVas can you please review this? I reverted changes and left public API as they were. thank you

SanjayVas (approved):
> Reviewed 1 of 13 files at r1, 8 of 8 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/875)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Marco-Premier)

Marco-Premier (commented):
> Reviewed 3 of 13 files at r1, 5 of 8 files at r3, 4 of 5 files at r4, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/875)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Marco-Premier)

PR #882, opened 2023-03-06
v2alpha metrics service -- createMetric & batchCreateMetrics

riemanli:
Add a metrics service in the v2alpha. `createMetric` and `batchCreateMetrics` are added in this PR.


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/882)

SanjayVas (changes requested):
> Reviewed 7 of 27 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/882)* status: 7 of 28 files reviewed, 6 unresolved discussions (waiting on @riemanli)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/EncryptionKeyPairStore.kt` line 1 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/882#-NPy7LmPEd_EYOPay-YS:-NPy7LmPEd_EYOPay-YT:b25uxvk) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c615736b4de6d6158c744f9f8b98d32a7a89efec/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/EncryptionKeyPairStore.kt#L1)):*
> ```Kotlin
> ```

I'm not spotting anything specific to a given Reporting API version in this file. Perhaps it can be moved up a level and shared across versions?
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 142 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/882#-NPy7sJ3B788li73l_2q:-NPy7sJ3B788li73l_2r:b3irmj3) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c615736b4de6d6158c744f9f8b98d32a7a89efec/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L142)):*
> ```Kotlin
>       it * REACH_FREQUENCY_VID_SAMPLING_WIDTH
>   }
> private const val REACH_FREQUENCY_REACH_EPSILON = 0.0033
> ```

nit: A lot of these look like they should be coming from flags or a config file so they can be configured without recompilation.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 209 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/882#-NPy89wK6Rz90szvSqfN:-NPy89wK6Rz90szvSqfO:buc3kym) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c615736b4de6d6158c744f9f8b98d32a7a89efec/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L209)):*
> ```Kotlin
>      * Creates CMM public [Measurement]s and [InternalMeasurement]s from a list of [InternalMetric].
>      */
>     suspend fun createCmmsMeasurements(
> ```

private
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 544 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/882#-NPy8XzCB_jiEdST5EJC:-NPy8XzCB_jiEdST5EJD:b-lr5lqd) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c615736b4de6d6158c744f9f8b98d32a7a89efec/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L544)):*
> ```Kotlin
>     }
> 
>     /** Combines two event group filters. */
> ```

If it combines exactly two, it should take two params and not a `List`.



_Code quote:_
```Kotlin
two
```
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 624 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/882#-NPy9BqwCSx2IhWv8LE_:-NPy9BqwCSx2IhWv8LEa:b-6xnqdt) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c615736b4de6d6158c744f9f8b98d32a7a89efec/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L624)):*
> ```Kotlin
>   }
> 
>   private val measurementSupplier =
> ```

nit: move this to where the other top-level properties are declared
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 655 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/882#-NPyCBMkDX2ZkD56UzsJ:-NPyCBMkDX2ZkD56UzsK:b-l29307) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c615736b4de6d6158c744f9f8b98d32a7a89efec/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L655)):*
> ```Kotlin
>     }
> 
>     val uncreatedInternalMetric: InternalMetric =
> ```

The fact that you have an `uncreatedInternalMetric` and an `initialInternalMetric` suggests that this flow is a bit convoluted. I get that `createInitialInternalMetric` is intended to extract common functionality between Create and BatchCreate, but I'd expect this logic to be a little clearer to follow:

1. Build an internal CreateMetric request from the public Metric request
   * The internal request is just an internal Metric message, but you refer to it as `internalCreateMetricRequest` to differentiate from the value returned from the internal API.
   * The logic that determines whether an internal Metric already exists for a given request ID can be handled within the internal CreateMetric implementation so that the check can be done transactionally.
2. Iterate over the resulting internal Measurements, creating a CMMS Measurement for any internal Measurement that does not have a `cmms_measurement_id`.
    * On each CMMS CreateMeasurement response, update the corresponding internal Measurement with the ID
3. Convert the internal Metric to a public API one to return.

Am I missing anything?



_Code quote:_
```Kotlin
uncreatedInternalMetric
```

SanjayVas (changes requested):
> Reviewed 5 of 27 files at r1, 10 of 19 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/882)* status: 20 of 41 files reviewed, 8 unresolved discussions (waiting on @riemanli)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 624 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/882#-NPy9BqwCSx2IhWv8LE_:-NQ1wFwO1fVrkeiz7TyQ:b-r5ijog) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c615736b4de6d6158c744f9f8b98d32a7a89efec/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L624)):*

Do you mean outside the MetricsService class? Since `MeasurementSupplier` needs those service stubs from `MetricsService`, I am not sure how to initialize the object somewhere else.
</blockquote></details>

I just mean before the other method declarations in the class. While we don't strictly follow this ordering, I find it a good default: https://kotlinlang.org/docs/coding-conventions.html#class-layout
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 546 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/882#-NQ1ynffA0SdoJXmzb7F:-NQ1ynffA0SdoJXmzb7G:b-a9c6t3) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/fde2ee8627298b7ca400567b63edfa7bf05e2066/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L546)):*
> ```Kotlin
> 
>     /** Combines event group filters. */
>     private fun combineEventGroupFilters(filters: List<String?>): String? {
> ```

nit: be explicit about how this is combining things



_Suggestion:_
```Kotlin
buildConjunction
```
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 546 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/882#-NQ1yxbj0dJ1pMb52XWC:-NQ1yxbj0dJ1pMb52XWD:blmvxsc) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/fde2ee8627298b7ca400567b63edfa7bf05e2066/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L546)):*
> ```Kotlin
> 
>     /** Combines event group filters. */
>     private fun combineEventGroupFilters(filters: List<String?>): String? {
> ```

nit: Since this is only called in one place and therefore is more for organization than DRY principles, maybe make the caller handle the nulls rather than doing a pass-through. The first step would be having the parameter be `Collection<String>`, meaning the caller has to filter out the nulls first. Another consideration would be having the return type be `String`, where the function has a precondition that the input collection may not be empty.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/SetExpressionCompiler.kt` line 423 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/882#-NQ21YCOE7a9G01QNlj4:-NQ21YCOE7a9G01QNlj5:b-sxdisp) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/fde2ee8627298b7ca400567b63edfa7bf05e2066/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/SetExpressionCompiler.kt#L423)):*
> ```Kotlin
> }
> 
> /** Gets all resource names of the reporting sets used in this [SetExpression]. */
> ```

This doesn't return anything.



_Code quote:_
```Kotlin
Gets
```
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/SetExpressionCompiler.kt` line 424 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/882#-NQ21JNq9Moi05qiki8k:-NQ21JNq9Moi05qiki8l:bf0wnr) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/fde2ee8627298b7ca400567b63edfa7bf05e2066/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/SetExpressionCompiler.kt#L424)):*
> ```Kotlin
> 
> /** Gets all resource names of the reporting sets used in this [SetExpression]. */
> private fun SetExpression.storeReportingSetNames(reportingSetNames: MutableSet<String>) {
> ```

"Store" implies writing things to blob storage. Maybe "Add"?



_Code quote:_
```Kotlin
storeReportingSetNames
```

riemanli (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/882)* status: 20 of 41 files reviewed, 6 unresolved discussions (waiting on @SanjayVas and @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 142 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/882#-NPy7sJ3B788li73l_2q:-NQ2Eixf1JHHQFavxjbC:b8enypl) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c615736b4de6d6158c744f9f8b98d32a7a89efec/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L142)):*

nit: A lot of these look like they should be coming from flags or a config file so they can be configured without recompilation.
</blockquote></details>

Offline discussion: Put the constants into a .proto. See [measurement_consumer_config.proto](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/02db9b01de4034a2383601ac42b30c82c1882acb/src/main/proto/wfa/measurement/config/reporting/measurement_consumer_config.proto). Then read the config and make it be part of the constructor arguments of `MetricsService`.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 209 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/882#-NPy89wK6Rz90szvSqfN:-NQ2Ow7m4kSSLN6XLB1M:b40vp3b) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c615736b4de6d6158c744f9f8b98d32a7a89efec/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L209)):*

private
</blockquote></details>

Offline discussion: the function is inside `MeasurementSupplier` and should be public. No change is needed.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 624 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/882#-NPy9BqwCSx2IhWv8LE_:-NQ2R0cP4hn9sQUdvLgg:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c615736b4de6d6158c744f9f8b98d32a7a89efec/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L624)):*

I just mean before the other method declarations in the class. While we don't strictly follow this ordering, I find it a good default: https://kotlinlang.org/docs/coding-conventions.html#class-layout
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 655 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/882#-NPyCBMkDX2ZkD56UzsJ:-NQ2Syj32nuekdLfyOnB:b-jtjs7f) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c615736b4de6d6158c744f9f8b98d32a7a89efec/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L655)):*

The fact that you have an `uncreatedInternalMetric` and an `initialInternalMetric` suggests that this flow is a bit convoluted. I get that `createInitialInternalMetric` is intended to extract common functionality between Create and BatchCreate, but I'd expect this logic to be a little clearer to follow:

1. Build an internal CreateMetric request from the public Metric request
   * The internal request is just an internal Metric message, but you refer to it as `internalCreateMetricRequest` to differentiate from the value returned from the internal API.
   * The logic that determines whether an internal Metric already exists for a given request ID can be handled within the internal CreateMetric implementation so that the check can be done transactionally.
2. Iterate over the resulting internal Measurements, creating a CMMS Measurement for any internal Measurement that does not have a `cmms_measurement_id`.
    * On each CMMS CreateMeasurement response, update the corresponding internal Measurement with the ID
3. Convert the internal Metric to a public API one to return.

Am I missing anything?
</blockquote></details>

Offline discussion: The internal `createMetric` method should cover the case when the metric is already created given a request ID and the case when the metric state is not valid. As a result of the former one, there is no need for the internal method `getInternalMetricByIdempotencyKey`. cc: @tristanvuong2021 
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 546 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/882#-NQ1ynffA0SdoJXmzb7F:-NQ2PTH03HmnIPXszASr:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/fde2ee8627298b7ca400567b63edfa7bf05e2066/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L546)):*

nit: be explicit about how this is combining things
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 546 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/882#-NQ1yxbj0dJ1pMb52XWC:-NQ2QmbT284ycBAgp3Qr:b-bjudbj) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/fde2ee8627298b7ca400567b63edfa7bf05e2066/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L546)):*

nit: Since this is only called in one place and therefore is more for organization than DRY principles, maybe make the caller handle the nulls rather than doing a pass-through. The first step would be having the parameter be `Collection<String>`, meaning the caller has to filter out the nulls first. Another consideration would be having the return type be `String`, where the function has a precondition that the input collection may not be empty.
</blockquote></details>

Thanks for the suggestion! Done.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/SetExpressionCompiler.kt` line 423 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/882#-NQ21YCOE7a9G01QNlj4:-NQ2SBu1Evp0gQxg2TSu:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/fde2ee8627298b7ca400567b63edfa7bf05e2066/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/SetExpressionCompiler.kt#L423)):*

This doesn't return anything.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/SetExpressionCompiler.kt` line 424 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/882#-NQ21JNq9Moi05qiki8k:-NQ2RszU6cAgirAg2CFL:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/fde2ee8627298b7ca400567b63edfa7bf05e2066/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/SetExpressionCompiler.kt#L424)):*

"Store" implies writing things to blob storage. Maybe "Add"?
</blockquote></details>

Done.



_Code quote:_
```Kotlin
SetExpression.storeReportingSetNames
```

riemanli (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/882)* status: 20 of 41 files reviewed, 6 unresolved discussions (waiting on @SanjayVas and @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 255 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/882#-NQ2l8mq2QKpZNmqVGhy:-NQ2l8mq2QKpZNmqVGhz:b-o972oi) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/fde2ee8627298b7ca400567b63edfa7bf05e2066/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L255)):*
> ```Kotlin
>               )
>             }
>           )
> ```

@SanjayVas, a question about `async`: I was doing

```kotlin
val deferred = mutableListOf<Deferred<MeasurementInfo>>()

...

          deferred.add(
            async {
              measurementIds {
                externalMeasurementId = weightedMeasurement.measurement.externalMeasurementId
                val measurement = createCmmsMeasurement(
                  weightedMeasurement.measurement,
                  internalMetric.metricSpec,
                  internalPrimitiveReportingSetMap,
                  measurementConsumer,
                  principal,
                )
                cmmsMeasurementId = MeasurementKey.fromName(measurement.name)!!.measurementId
              }
            }
          )
```

But the unit test shows the externalMeasurementId didn't pair with cmmsMeasurementId correctly. My thought is that it's race condition due to `async` and so I used a map to store the correct mapping. But do you think that is the real root cause?



_Code quote:_
```Kotlin
          measurementRequestIdToExternalId[
            weightedMeasurement.measurement.cmmsCreateMeasurementRequestId] =
            weightedMeasurement.measurement.externalMeasurementId

          deferred.add(
            async {
              createCmmsMeasurement(
                weightedMeasurement.measurement,
                internalMetric.metricSpec,
                internalPrimitiveReportingSetMap,
                measurementConsumer,
                principal,
              )
            }
          )
```

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/882)* status: 20 of 41 files reviewed, 6 unresolved discussions (waiting on @riemanli and @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 255 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/882#-NQ2l8mq2QKpZNmqVGhy:-NQ3-WtI8-lZnmnnylow:bny14ka) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/fde2ee8627298b7ca400567b63edfa7bf05e2066/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L255)):*

@SanjayVas, a question about `async`: I was doing

```kotlin
val deferred = mutableListOf<Deferred<MeasurementInfo>>()

...

          deferred.add(
            async {
              measurementIds {
                externalMeasurementId = weightedMeasurement.measurement.externalMeasurementId
                val measurement = createCmmsMeasurement(
                  weightedMeasurement.measurement,
                  internalMetric.metricSpec,
                  internalPrimitiveReportingSetMap,
                  measurementConsumer,
                  principal,
                )
                cmmsMeasurementId = MeasurementKey.fromName(measurement.name)!!.measurementId
              }
            }
          )
```

But the unit test shows the externalMeasurementId didn't pair with cmmsMeasurementId correctly. My thought is that it's race condition due to `async` and so I used a map to store the correct mapping. But do you think that is the real root cause?
</blockquote></details>

I suspect it's an issue with the unit test. Since the order of the CMMS create RPCs are executed here is not deterministic, you need to account for that in your test w.r.t. any stubbing of mocks. For example, making sure that your stubbed response depends on the request ID.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 254 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/882#-NQ3-GDt8FFZ8u2GPeaZ:-NQ3-GDt8FFZ8u2GPea_:b-hoffq6) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7b26943a48eabed3ef4d09ffb403f1604a139b08/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L254)):*
> ```Kotlin
>         for (weightedMeasurement in internalMetric.weightedMeasurementsList) {
>           if (weightedMeasurement.measurement.cmmsMeasurementId.isNotBlank()) continue
> ```

Always use curly braces for `if` statements



_Code quote:_
```Kotlin
 continue
```

riemanli (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/882)* status: 20 of 41 files reviewed, 6 unresolved discussions (waiting on @SanjayVas and @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 255 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/882#-NQ2l8mq2QKpZNmqVGhy:-NQ3Os_qCLlbLQD6N9DR:b-klthsr) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/fde2ee8627298b7ca400567b63edfa7bf05e2066/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L255)):*

I suspect it's an issue with the unit test. Since the order of the CMMS create RPCs are executed here is not deterministic, you need to account for that in your test w.r.t. any stubbing of mocks. For example, making sure that your stubbed response depends on the request ID.
</blockquote></details>

Thanks for the feedback. The unit tests pass after I change let `measurementsMock` output based on the request ID.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 254 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/882#-NQ3-GDt8FFZ8u2GPeaZ:-NQ3OVpp-xP9q8krziBr:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7b26943a48eabed3ef4d09ffb403f1604a139b08/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L254)):*

Always use curly braces for `if` statements
</blockquote></details>

Done.

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/882)* status: 20 of 41 files reviewed, 5 unresolved discussions (waiting on @riemanli and @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 275 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/882#-NQaHtYACatqsN9q0Z06:-NQaHtYACatqsN9q0Z07:b-mltcuv) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a71eae33750266d7384e90defd97b306f5fe82d9/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L275)):*
> ```Kotlin
>       }
> 
>       // Set CMMs measurement IDs.
> ```

nit: CMMS for Cross-Media Measurement System



_Suggestion:_
```Kotlin
CMMS
```

tristanvuong2021 (commented):
> Reviewed 10 of 21 files at r3, 16 of 17 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/882)* status: 35 of 43 files reviewed, 5 unresolved discussions (waiting on @riemanli and @SanjayVas)

tristanvuong2021 (commented):
> Reviewed 9 of 29 files at r1, 3 of 21 files at r3, 1 of 17 files at r6.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/882)* status: 39 of 43 files reviewed, 5 unresolved discussions (waiting on @riemanli and @SanjayVas)

riemanli (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/882)* status: 14 of 64 files reviewed, 4 unresolved discussions (waiting on @SanjayVas and @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 142 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/882#-NPy7sJ3B788li73l_2q:-NSlqdgv3g1aoc5ERSgR:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c615736b4de6d6158c744f9f8b98d32a7a89efec/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L142)):*

Offline discussion: Put the constants into a .proto. See [measurement_consumer_config.proto](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/02db9b01de4034a2383601ac42b30c82c1882acb/src/main/proto/wfa/measurement/config/reporting/measurement_consumer_config.proto). Then read the config and make it be part of the constructor arguments of `MetricsService`.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 874 at r16](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/882#-NSlW5TrD1TYK1lHnnjy:-NSlW5TrD1TYK1lHnnjz:bs33sv8) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8c52414dc5a6460ea580ffdaf284040b2ff4143f/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L874)):*
> ```Kotlin
>             weightedSubsetUnion.primitiveReportingSetBasesList.map { primitiveReportingSetBasis ->
>               primitiveReportingSetBasis.copy { filters += metric.filtersList }
>             }
> ```

@tristanvuong2021 this is one of places where I need the primitive reporting set has itself in its WeightedSubsetUnion.primitiveReportingSetBases. It makes it easier to convert a `weightedSubsetUnion` to an internal measurement without any reporting set type check.



_Code quote:_
```Kotlin
          this.primitiveReportingSetBases +=
            weightedSubsetUnion.primitiveReportingSetBasesList.map { primitiveReportingSetBasis ->
              primitiveReportingSetBasis.copy { filters += metric.filtersList }
            }
```

tristanvuong2021 (changes requested):
> Reviewed 1 of 29 files at r1, 2 of 3 files at r9, 14 of 48 files at r15, 19 of 19 files at r16, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/882)* status: 49 of 64 files reviewed, 7 unresolved discussions (waiting on @riemanli and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 561 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/882#-NRtj9E31Knw4mhWGuTB:-NRtj9E31Knw4mhWGuTC:byltel6) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/71ab2f442e954c8a09a20d60a7d52214f023f924/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L561)):*
> ```Kotlin
>     }
> 
>     /** Get a [MeasurementConsumer] based on a CMMs ID. */
> ```

nit: lowercase s here too
___
*[`src/test/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsServiceTest.kt` line 132 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/882#-NRtb9sh7-_u3vYZFNqT:-NRtb6Jm2UbIzMx2FH9V:b-60rkgz) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/71ab2f442e954c8a09a20d60a7d52214f023f924/src/test/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsServiceTest.kt#L132)):*
> ```Kotlin
> import org.wfanet.measurement.internal.reporting.v2alpha.ReportingSet.SetExpression as InternalSetExpression
> import org.wfanet.measurement.internal.reporting.v2alpha.ReportingSetKt as InternalReportingSetKt
> import org.wfanet.measurement.internal.reporting.v2alpha.ReportingSetKt.SetExpressionKt.operand as internalOperand
> ```

Do you still have autocomplete? I believe the non top level aliases, the ones for the builders under ReportingSetKt, break it.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/EncryptionKeyPairStore.kt` line 1 at r17](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/882#-NSm14KU2wHK5lSguHlV:-NSm14KU2wHK5lSguHlW:b-ao6wp2) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f7510a5ac45eb76e0f2ad05031a01472ee91ed09/src/main/kotlin/org/wfanet/measurement/reporting/service/api/EncryptionKeyPairStore.kt#L1)):*
> ```Kotlin
> ```

Since this file was renamed, the year should stay as 2022.

tristanvuong2021 (changes requested):
> Reviewed 13 of 48 files at r15.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/882)* status: 62 of 64 files reviewed, 8 unresolved discussions (waiting on @riemanli and @SanjayVas)

___
*[`src/test/kotlin/org/wfanet/measurement/reporting/service/api/EncryptionKeyPairStoreTest.kt` line 1 at r17](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/882#-NSm3rze2P1aobczGLnk:-NSm3rze2P1aobczGLnl:bydi8ip) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f7510a5ac45eb76e0f2ad05031a01472ee91ed09/src/test/kotlin/org/wfanet/measurement/reporting/service/api/EncryptionKeyPairStoreTest.kt#L1)):*
> ```Kotlin
> ```

This file was also renamed so the year shouldn't change.

riemanli (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/882)* status: 62 of 64 files reviewed, 8 unresolved discussions (waiting on @SanjayVas and @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 561 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/882#-NRtj9E31Knw4mhWGuTB:-NSm41SM8qBW3paPBRc1:bsl5c9l) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/71ab2f442e954c8a09a20d60a7d52214f023f924/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L561)):*

nit: lowercase s here too
</blockquote></details>

I think you were looking at an old commit. It's fixed in the latest.
___
*[`src/test/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsServiceTest.kt` line 132 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/882#-NRtb9sh7-_u3vYZFNqT:-NSm4X9UD-gCNE2B1vSy:buvstzh) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/71ab2f442e954c8a09a20d60a7d52214f023f924/src/test/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsServiceTest.kt#L132)):*

Do you still have autocomplete? I believe the non top level aliases, the ones for the builders under ReportingSetKt, break it.
</blockquote></details>

I am not sure what you meant by autocomplete, but let me remove the aliases for the builder methods.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/EncryptionKeyPairStore.kt` line 1 at r17](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/882#-NSm14KU2wHK5lSguHlV:-NSm3RPV9yHwfh2gn-YW:b-4k0685) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f7510a5ac45eb76e0f2ad05031a01472ee91ed09/src/main/kotlin/org/wfanet/measurement/reporting/service/api/EncryptionKeyPairStore.kt#L1)):*

Since this file was renamed, the year should stay as 2022.
</blockquote></details>

Good point! Done.
___
*[`src/test/kotlin/org/wfanet/measurement/reporting/service/api/EncryptionKeyPairStoreTest.kt` line 1 at r17](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/882#-NSm3rze2P1aobczGLnk:-NSm4I60C-yshZubvmep:b-mws2j5) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f7510a5ac45eb76e0f2ad05031a01472ee91ed09/src/test/kotlin/org/wfanet/measurement/reporting/service/api/EncryptionKeyPairStoreTest.kt#L1)):*

This file was also renamed so the year shouldn't change.
</blockquote></details>

Thanks for pointing it out! Done.

riemanli (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/882)* status: 48 of 64 files reviewed, 7 unresolved discussions (waiting on @SanjayVas and @tristanvuong2021)

___
*[`src/test/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsServiceTest.kt` line 132 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/882#-NRtb9sh7-_u3vYZFNqT:-NSm6KBH7wL4x_myLkjT:b-896fja) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/71ab2f442e954c8a09a20d60a7d52214f023f924/src/test/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsServiceTest.kt#L132)):*

I am not sure what you meant by autocomplete, but let me remove the aliases for the builder methods.
</blockquote></details>

Done!

tristanvuong2021:
*[`src/test/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsServiceTest.kt` line 132 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/882#-NRtb9sh7-_u3vYZFNqT:-NSm6m7u6yyvyh7xH4G-:b-busxe3) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/71ab2f442e954c8a09a20d60a7d52214f023f924/src/test/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsServiceTest.kt#L132)):*

There is a bug in Intellij where certain aliases break autocomplete. MetricKt as InternalMetricKt is okay, but MetricKt.xxx as internalXXX breaks it.
</blockquote></details>

So you can have aliases, but just things like InternalReportingKt and InternalMetricKt, and not internalSetExpression

riemanli (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/882)* status: 48 of 64 files reviewed, 7 unresolved discussions (waiting on @SanjayVas and @tristanvuong2021)

___
*[`src/test/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsServiceTest.kt` line 132 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/882#-NRtb9sh7-_u3vYZFNqT:-NSm7KfyAMES-tBzUUai:b-xfyrp0) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/71ab2f442e954c8a09a20d60a7d52214f023f924/src/test/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsServiceTest.kt#L132)):*

So you can have aliases, but just things like InternalReportingKt and InternalMetricKt, and not internalSetExpression
</blockquote></details>

Oh that's interesting. I still had autocomplete working when I had those aliases. But anyway, I removed those for others' sake. Thanks for pointing it out!

SanjayVas (changes requested):
> Reviewed 13 of 48 files at r15, 2 of 15 files at r18.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/882)* status: 48 of 65 files reviewed, 13 unresolved discussions (waiting on @riemanli and @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 229 at r19](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/882#-NSqW7Y62uaxOBRV2FUo:-NSqW7Y62uaxOBRV2FUp:bhs84cs) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1e44c44d6f2a81f4955302ce9b15b73de4d35726/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L229)):*
> ```Kotlin
>       }
> 
>       // Set CMMS measurement IDs.
> ```

I think this comment was meant to apply to the following block instead.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 322 at r19](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/882#-NSqWl0I7DPsVq3-DuFg:-NSqWl0JDDlos1UnyYxc:bhuwwos) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1e44c44d6f2a81f4955302ce9b15b73de4d35726/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L322)):*
> ```Kotlin
>     ): SigningKeyHandle {
>       // TODO: Factor this out to a separate class similar to EncryptionKeyPairStore.
>       val signingPrivateKeyDer: ByteString =
> ```

Wrap the file system I/O in a `withContext` block that uses a context that has a coroutine dispatcher suitable for blocking calls.

The general pattern is to take in said context in the constructor and annotate it with `@BlockingExecutor`, defaulting to  Dispatchers.IO.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 523 at r19](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/882#-NSq_-xv1j7lxs8Pzd0e:-NSq_-xv1j7lxs8Pzd0f:b-7s19t9) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1e44c44d6f2a81f4955302ce9b15b73de4d35726/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L523)):*
> ```Kotlin
>     /**
>      * Builds a map of external reporting set IDs to [InternalReportingSet]s to minimize grpc calls.
>      */
> ```

nit: the summary fragment should indicate what the method does, with additional details in following paragraphs.

Also the C in RPC is "call", so "RPC call" is redundant :P



_Suggestion:_
```Kotlin
    /**
     * Builds a map of external reporting set IDs to [InternalReportingSet]s.
     *
     * This helps reduce the number of RPCs.
     */
```
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 598 at r19](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/882#-NSqaLdn6YVfMxD1tyzJ:-NSqaLdn6YVfMxD1tyzK:bhbt42r) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1e44c44d6f2a81f4955302ce9b15b73de4d35726/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L598)):*
> ```Kotlin
>     }
> 
>     val initialInternalMetric =
> ```

What's "initial" about this?



_Suggestion:_
```Kotlin
internalMetric
```
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 667 at r19](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/882#-NSqadpDA8dirF2LBGq6:-NSqadpDA8dirF2LBGq7:b-kyw6zu) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1e44c44d6f2a81f4955302ce9b15b73de4d35726/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L667)):*
> ```Kotlin
> 
>   /** Builds an [InternalMetric]. */
>   private suspend fun buildInternalMetric(
> ```

This isn't *really* building an [InternalMetric]. Strictly speaking it's building an instance of that class, but it's incomplete as it's part of the internal CreateMetricRequest. It's generally clearer to only refer to complete [InternalMetric] instances (ones that come from the internal API) as such.

Maybe make a `buildInternalCreateMetricRequest` instead?

Note that the "buildX" methods which build complete message are fine as-is, the issue is only when building an incomplete message where the server returns the complete one.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 886 at r19](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/882#-NSqe1tv8ON9CMqqSAEO:-NSqe1tv8ON9CMqqSAEP:bq5ssrp) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1e44c44d6f2a81f4955302ce9b15b73de4d35726/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L886)):*
> ```Kotlin
>       }
> 
>     grpcRequire(reportingSetKey.cmmsMeasurementConsumerId == cmmsMeasurementConsumerId) {
> ```

I assume this should be a PERMISSION_DENIED, assuming that `cmmsMeasurementConsumerId` comes from the authenticated principal. The default status code for grpcRequire is INVALID_ARGUMENT, which means the request message is invalid regardless of context.

tristanvuong2021 (commented):
> Reviewed 13 of 15 files at r18, 1 of 3 files at r20, 26 of 28 files at r21, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/882)* status: 63 of 65 files reviewed, 10 unresolved discussions (waiting on @riemanli and @SanjayVas)

tristanvuong2021 (commented):
> Reviewed 1 of 1 files at r22, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/882)* status: 63 of 65 files reviewed, 10 unresolved discussions (waiting on @riemanli and @SanjayVas)

tristanvuong2021 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/882)* status: 63 of 65 files reviewed, 11 unresolved discussions (waiting on @riemanli and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 915 at r16](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/882#-NSrIYm24QKGe2lc8xgZ:-NSrIYm24QKGe2lc8xg_:bxo361e) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8c52414dc5a6460ea580ffdaf284040b2ff4143f/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L915)):*
> ```Kotlin
> 
> /**
>  * Converts an [MetricSpecConfig.VidSamplingInterval] to an
> ```

A lot of these conversion functions look like the same ones in https://github.com/world-federation-of-advertisers/cross-media-measurement/pull/943. 
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 667 at r19](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/882#-NSqadpDA8dirF2LBGq6:-NSrGO9m6zBOYdjz-Aj1:bevshop) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1e44c44d6f2a81f4955302ce9b15b73de4d35726/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L667)):*

This isn't *really* building an [InternalMetric]. Strictly speaking it's building an instance of that class, but it's incomplete as it's part of the internal CreateMetricRequest. It's generally clearer to only refer to complete [InternalMetric] instances (ones that come from the internal API) as such.

Maybe make a `buildInternalCreateMetricRequest` instead?

Note that the "buildX" methods which build complete message are fine as-is, the issue is only when building an incomplete message where the server returns the complete one.
</blockquote></details>

This is a good idea. It can return the entire InternalCreateMetricRequest.

riemanli (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/882)* status: 63 of 65 files reviewed, 11 unresolved discussions (waiting on @SanjayVas and @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 915 at r16](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/882#-NSrIYm24QKGe2lc8xgZ:-NSrKEyOC1okZWqcy_nY:b4uzd13) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8c52414dc5a6460ea580ffdaf284040b2ff4143f/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L915)):*

A lot of these conversion functions look like the same ones in https://github.com/world-federation-of-advertisers/cross-media-measurement/pull/943. 
</blockquote></details>

They should be removed in the latest commit. Could you try to refresh the page?

riemanli (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/882)* status: 63 of 65 files reviewed, 10 unresolved discussions (waiting on @SanjayVas and @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 229 at r19](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/882#-NSqW7Y62uaxOBRV2FUo:-NSrL_aR0l6tYZhyQcIY:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1e44c44d6f2a81f4955302ce9b15b73de4d35726/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L229)):*

I think this comment was meant to apply to the following block instead.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 322 at r19](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/882#-NSqWl0I7DPsVq3-DuFg:-NSrMVcnBElRZLNri67-:b-kih9l6) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1e44c44d6f2a81f4955302ce9b15b73de4d35726/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L322)):*

Wrap the file system I/O in a `withContext` block that uses a context that has a coroutine dispatcher suitable for blocking calls.

The general pattern is to take in said context in the constructor and annotate it with `@BlockingExecutor`, defaulting to  Dispatchers.IO.
</blockquote></details>

Got it. Done.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 523 at r19](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/882#-NSq_-xv1j7lxs8Pzd0e:-NSrN2_N7pE_NYXMmbdd:bxx9oq) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1e44c44d6f2a81f4955302ce9b15b73de4d35726/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L523)):*

nit: the summary fragment should indicate what the method does, with additional details in following paragraphs.

Also the C in RPC is "call", so "RPC call" is redundant :P
</blockquote></details>

Thanks for pointing it out! Fixed!
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 598 at r19](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/882#-NSqaLdn6YVfMxD1tyzJ:-NSrO4-cEYOVP11Syicg:bss4a2b) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1e44c44d6f2a81f4955302ce9b15b73de4d35726/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L598)):*

What's "initial" about this?
</blockquote></details>

Removed.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 667 at r19](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/882#-NSqadpDA8dirF2LBGq6:-NSrPTVMEflpXDsTNtq1:bi36ttq) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1e44c44d6f2a81f4955302ce9b15b73de4d35726/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L667)):*

This is a good idea. It can return the entire InternalCreateMetricRequest.
</blockquote></details>

Thanks for the great suggestion. Done!
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 886 at r19](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/882#-NSqe1tv8ON9CMqqSAEO:-NSrQcaSDbF3mbDkq5R0:b-zf5xbg) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1e44c44d6f2a81f4955302ce9b15b73de4d35726/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L886)):*

I assume this should be a PERMISSION_DENIED, assuming that `cmmsMeasurementConsumerId` comes from the authenticated principal. The default status code for grpcRequire is INVALID_ARGUMENT, which means the request message is invalid regardless of context.
</blockquote></details>

Thanks for catching it! Fixed.

SanjayVas (changes requested):
> Reviewed 1 of 21 files at r3, 2 of 17 files at r6, 2 of 48 files at r15, 19 of 19 files at r16, 12 of 15 files at r18, 25 of 28 files at r21, 2 of 2 files at r24, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/882)* status: 64 of 65 files reviewed, 8 unresolved discussions (waiting on @riemanli and @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 322 at r19](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/882#-NSqWl0I7DPsVq3-DuFg:-NSvY64oEthL3BBpUOyV:b6bhifl) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1e44c44d6f2a81f4955302ce9b15b73de4d35726/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L322)):*

Got it. Done.
</blockquote></details>

This is using `Dispatchers.IO` directly, where it should be using a coroutine context passed in the constructor as I described. It's just that the constructor parameter can default to `Dispatchers.IO`.
___
*[`src/test/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsServiceTest.kt` line 464 at r24](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/882#-NSv_yVA5FIfqg1rG65l:-NSv_yVA5FIfqg1rG65m:bdrbw90) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7eb3c9ce6e616f954ed1b2b03fc763927bfb828f/src/test/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsServiceTest.kt#L464)):*
> ```Kotlin
>   }
> }
> private val INTERNAL_UNION_ALL_BUT_LAST_PUBLISHER_REPORTING_SET = internalReportingSet {
> ```

nit: there are way too many specific single-use constants here. It makes it difficult to figure out what each test method is actually verifying.

Just define the common ones that you'll need to reuse, and do the one-time modifications in the specific test method. You also don't need to build up each constant out of many other constants, unless those also are reused.

This is distinctly different than the pattern one would use for non-test code, where most literals should be extracted to constants.
___
*[`src/test/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsServiceTest.kt` line 1177 at r24](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/882#-NSvZrQgBx42K_xg52Bz:-NSvZrQgBx42K_xg52C-:bk6u269) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7eb3c9ce6e616f954ed1b2b03fc763927bfb828f/src/test/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsServiceTest.kt#L1177)):*
> ```Kotlin
>       )
>     // onBlocking { streamMetrics(any()) }
>     //   .thenReturn(
> ```

Don't leave commented code behind.



_Code quote:_
```Kotlin
    // onBlocking { streamMetrics(any()) }
    //   .thenReturn(
    //     flowOf(
    //       INTERNAL_PENDING_REACH_METRIC,
    //       INTERNAL_PENDING_IMPRESSION_METRIC,
    //       INTERNAL_PENDING_WATCH_DURATION_METRIC,
    //       INTERNAL_PENDING_FREQUENCY_HISTOGRAM_METRIC,
    //     )
    //   )
    // onBlocking { getMetric(any()) }
    //   .thenReturn(
    //     INTERNAL_SUCCEEDED_REACH_METRIC,
    //     INTERNAL_SUCCEEDED_IMPRESSION_METRIC,
    //     INTERNAL_SUCCEEDED_WATCH_DURATION_METRIC,
    //     INTERNAL_SUCCEEDED_FREQUENCY_HISTOGRAM_METRIC,
    //   )
```
___
*[`src/test/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsServiceTest.kt` line 1311 at r24](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/882#-NSvc3-R-JkktQg0-ZkX:-NSvc3-R-JkktQg0-ZkY:bha5bvp) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7eb3c9ce6e616f954ed1b2b03fc763927bfb828f/src/test/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsServiceTest.kt#L1311)):*
> ```Kotlin
> 
>   @Test
>   fun `createMetric without request ID returns incremental reach metric with RUNNING state`() {
> ```

nit: isn't the state determined by the internal CreateMetric response? If that's the case, then this test method name is misleading as it's not actually what's being verified.

Maybe `createMetric creates CMMS Measurements` for this one. Add the other part of the conditions to the "when" part when relevant, e.g. `createMetric reuses X when request ID specified`



_Suggestion:_
```Kotlin
createMetric creates CMMS Measurements
```
___
*[`src/test/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsServiceTest.kt` line 1368 at r24](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/882#-NSvdRUG3B3PptFY5TZW:-NSvdRUG3B3PptFY5TZX:b-gab0gb) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7eb3c9ce6e616f954ed1b2b03fc763927bfb828f/src/test/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsServiceTest.kt#L1368)):*
> ```Kotlin
>       .containsExactly(
>         getDataProviderRequest { name = DATA_PROVIDERS_LIST[0].name },
>         getDataProviderRequest { name = DATA_PROVIDERS_LIST[1].name },
> ```

Your stubbing for `dataProvidersMock` already matches on the EDP name, so you don't need to capture/verify the call arguments here.



_Code quote:_
```Kotlin
    // Verify proto argument of DataProvidersCoroutineImplBase::getDataProvider
    val dataProvidersCaptor: KArgumentCaptor<GetDataProviderRequest> = argumentCaptor()
    verifyBlocking(dataProvidersMock, times(5)) { getDataProvider(dataProvidersCaptor.capture()) }

    val capturedDataProviderRequests = dataProvidersCaptor.allValues
    assertThat(capturedDataProviderRequests)
      .containsExactly(
        getDataProviderRequest { name = DATA_PROVIDERS_LIST[0].name },
        getDataProviderRequest { name = DATA_PROVIDERS_LIST[1].name },
        getDataProviderRequest { name = DATA_PROVIDERS_LIST[0].name },
        getDataProviderRequest { name = DATA_PROVIDERS_LIST[1].name },
        getDataProviderRequest { name = DATA_PROVIDERS_LIST[2].name },
      )
```

riemanli (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/882)* status: 63 of 65 files reviewed, 8 unresolved discussions (waiting on @SanjayVas and @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 322 at r19](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/882#-NSqWl0I7DPsVq3-DuFg:-NSwpTTu8gLZfi8HRXc_:bib909w) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1e44c44d6f2a81f4955302ce9b15b73de4d35726/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L322)):*

This is using `Dispatchers.IO` directly, where it should be using a coroutine context passed in the constructor as I described. It's just that the constructor parameter can default to `Dispatchers.IO`.
</blockquote></details>

I added `CoroutineContext` to the constructor of `MetricsService` and the constructor of `MeasurementSupplier`, and let I/O use that `CoroutineContext` in `withContext`. Let me know if it's still wrong.
___
*[`src/test/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsServiceTest.kt` line 464 at r24](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/882#-NSv_yVA5FIfqg1rG65l:-NSx73qdDACvcLxHAC6t:b-o0f6vn) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7eb3c9ce6e616f954ed1b2b03fc763927bfb828f/src/test/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsServiceTest.kt#L464)):*

nit: there are way too many specific single-use constants here. It makes it difficult to figure out what each test method is actually verifying.

Just define the common ones that you'll need to reuse, and do the one-time modifications in the specific test method. You also don't need to build up each constant out of many other constants, unless those also are reused.

This is distinctly different than the pattern one would use for non-test code, where most literals should be extracted to constants.
</blockquote></details>

Maybe I tried too hard to cover all use cases :P

In addition, I thought doing this will be easier for the reader as the reader can also know every stage of the process. Anyway, I deleted the ones that are used only one or two times. 

Note that `INTERNAL_UNION_ALL_BUT_LAST_PUBLISHER_REPORTING_SET` is used many times in other places, so I still keep it.
___
*[`src/test/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsServiceTest.kt` line 1177 at r24](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/882#-NSvZrQgBx42K_xg52Bz:-NSwbirj2ZlEO4M4tTCo:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7eb3c9ce6e616f954ed1b2b03fc763927bfb828f/src/test/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsServiceTest.kt#L1177)):*

Don't leave commented code behind.
</blockquote></details>

Done.
___
*[`src/test/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsServiceTest.kt` line 1311 at r24](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/882#-NSvc3-R-JkktQg0-ZkX:-NSweb28D-oZfIx8jQ8e:blf7oq6) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7eb3c9ce6e616f954ed1b2b03fc763927bfb828f/src/test/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsServiceTest.kt#L1311)):*

nit: isn't the state determined by the internal CreateMetric response? If that's the case, then this test method name is misleading as it's not actually what's being verified.

Maybe `createMetric creates CMMS Measurements` for this one. Add the other part of the conditions to the "when" part when relevant, e.g. `createMetric reuses X when request ID specified`
</blockquote></details>

Good point. I renamed this one and the other testing functions. 
___
*[`src/test/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsServiceTest.kt` line 1368 at r24](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/882#-NSvdRUG3B3PptFY5TZW:-NSwrIhj-dou48x_dv80:b-2d29vb) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7eb3c9ce6e616f954ed1b2b03fc763927bfb828f/src/test/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsServiceTest.kt#L1368)):*

Your stubbing for `dataProvidersMock` already matches on the EDP name, so you don't need to capture/verify the call arguments here.
</blockquote></details>

Thanks for pointing it out! Removed the codes that verify call arguments of `GetDataProviderRequest` when  `getDataProvider` should be called.

SanjayVas (approved):
> Reviewed 2 of 2 files at r25, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/882)* status: all files reviewed, 1 unresolved discussion (waiting on @tristanvuong2021)

stevenwarejones (commented):
> Reviewed 2 of 21 files at r3, 2 of 17 files at r6, 2 of 48 files at r15, 19 of 19 files at r16, 12 of 15 files at r18, 25 of 28 files at r21, 1 of 2 files at r24, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/882)* status: all files reviewed, 1 unresolved discussion (waiting on @Marco-Premier and @tristanvuong2021)

Marco-Premier (approved):
> Reviewed 2 of 21 files at r3, 2 of 17 files at r6, 2 of 48 files at r15, 19 of 19 files at r16, 12 of 15 files at r18, 25 of 28 files at r21, 1 of 2 files at r24, 2 of 2 files at r25, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/882)* status: all files reviewed, 1 unresolved discussion (waiting on @tristanvuong2021)

riemanli (commented):
> Reviewed 2 of 21 files at r3, 2 of 17 files at r6, 2 of 48 files at r15, 19 of 19 files at r16, 12 of 15 files at r18, 25 of 28 files at r21, 1 of 2 files at r24, 2 of 2 files at r25, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/882)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @tristanvuong2021)

PR #895, opened 2023-03-17
V2alpha DP params API

riemanli:
* Make all fields in metric spec not required. Document the usage.
* Incorporate the new kingdom measurement design.
* Add metric spec config for storing default values.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -17,10 +17,6 @@ test --test_env=TESTCONTAINERS_RYUK_DISABLED
 build:debian-bullseye --platforms=//build/platforms:debian_bullseye
 build:ubuntu-bionic --platforms=//build/platforms:ubuntu_bionic
 
-# Enable explicit presence in proto3.
-# See https://github.com/protocolbuffers/protobuf/blob/main/docs/field_presence.md#how-to-enable-explicit-presence-in-proto3
-build --protocopt=--experimental_allow_proto3_optional
-
 import %workspace%/container.bazelrc
 import %workspace%/remote.bazelrc
 import %workspace%/results.bazelrc
--- a/.bazelrc
+++ b/.bazelrc
@@ -17,6 +17,10 @@ test --test_env=TESTCONTAINERS_RYUK_DISABLED
 build:debian-bullseye --platforms=//build/platforms:debian_bullseye
 build:ubuntu-bionic --platforms=//build/platforms:ubuntu_bionic
 
+# Enable explicit presence in proto3.
+# See https://github.com/protocolbuffers/protobuf/blob/main/docs/field_presence.md#how-to-enable-explicit-presence-in-proto3
+build --protocopt=--experimental_allow_proto3_optional
+
 import %workspace%/container.bazelrc
 import %workspace%/remote.bazelrc
 import %workspace%/results.bazelrc
--- a/src/main/proto/wfa/measurement/config/reporting/metric_spec_config.proto
+++ b/src/main/proto/wfa/measurement/config/reporting/metric_spec_config.proto
@@ -1,4 +1,4 @@
-// Copyright 2022 The Cross-Media Measurement Authors
+// Copyright 2023 The Cross-Media Measurement Authors
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/src/main/proto/wfa/measurement/config/reporting/metric_spec_config.proto
+++ b/src/main/proto/wfa/measurement/config/reporting/metric_spec_config.proto
@@ -31,12 +31,24 @@ message MetricSpecConfig {
     double delta = 2;
   }
 
-  // Parameters that are used to generate `Reach` metric or
-  // `Frequency Histogram` metric.
-  message ReachFrequencyParams {
+  // Parameters that are used to generate `Reach` metric.
+  message ReachParams {
     // Differential privacy parameters for reach.
     DifferentialPrivacyParams reach_privacy_params = 1;
-
+    // Differential privacy parameters for frequency.
+    //
+    // Small default values are selected for the reach computation execution.
+    DifferentialPrivacyParams frequency_privacy_params = 2;
+    // Maximum frequency cut-off value in the frequency histogram.
+    //
+    // Counts with frequency higher than `max_frequency` will be aggregated
+    // together. For reach, this is equal to 1.
+    int32 maximum_frequency_per_user = 3;
+  }
+  // Parameters that are used to generate `Frequency Histogram` metric.
+  message FrequencyHistogramParams {
+    // Differential privacy parameters for reach.
+    DifferentialPrivacyParams reach_privacy_params = 1;
     // Differential privacy parameters for frequency.
     DifferentialPrivacyParams frequency_privacy_params = 2;
     // Maximum frequency cut-off value in the frequency histogram.
@@ -70,10 +82,10 @@ message MetricSpecConfig {
 
   // Parameters for generating the count of unique audiences reached given a set
   // of event groups.
-  ReachFrequencyParams reach_params = 1;
+  ReachParams reach_params = 1;
   // Parameters for generating the reach frequency histogram given a set of
   // event groups.
-  ReachFrequencyParams frequency_histogram_params = 2;
+  FrequencyHistogramParams frequency_histogram_params = 2;
   // Parameters for generating the impression count given a set of event groups.
   ImpressionCountParams impression_count_params = 3;
   // Parameters for generating the watch duration given a set of event groups.
--- a/src/main/proto/wfa/measurement/config/reporting/metric_spec_config.proto
+++ b/src/main/proto/wfa/measurement/config/reporting/metric_spec_config.proto
@@ -31,20 +31,19 @@ message MetricSpecConfig {
     double delta = 2;
   }
 
-  // Parameters that are used to generate `Reach` metric.
-  message ReachParams {
-    // Differential privacy parameters.
-    DifferentialPrivacyParams privacy_params = 1;
-  }
-  // Parameters that are used to generate `Frequency Histogram` metric.
-  message FrequencyHistogramParams {
-    // Differential privacy parameters.
-    DifferentialPrivacyParams privacy_params = 1;
+  // Parameters that are used to generate `Reach` metric or
+  // `Frequency Histogram` metric.
+  message ReachFrequencyParams {
+    // Differential privacy parameters for reach.
+    DifferentialPrivacyParams reach_privacy_params = 1;
+
+    // Differential privacy parameters for frequency.
+    DifferentialPrivacyParams frequency_privacy_params = 2;
     // Maximum frequency cut-off value in the frequency histogram.
     //
     // Counts with frequency higher than `max_frequency` will be aggregated
     // together.
-    int32 maximum_frequency_per_user = 2;
+    int32 maximum_frequency_per_user = 3;
   }
   // Parameters that are used to generate `Impression Count` metric.
   message ImpressionCountParams {
--- a/src/main/proto/wfa/measurement/config/reporting/metric_spec_config.proto
+++ b/src/main/proto/wfa/measurement/config/reporting/metric_spec_config.proto
@@ -1,0 +1,81 @@
+// Copyright 2022 The Cross-Media Measurement Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+syntax = "proto3";
+
+package wfa.measurement.config;
+
+option java_package = "org.wfanet.measurement.config.reporting";
+option java_multiple_files = true;
+
+// The configuration for metric spec in Metric in the reporting server
+message MetricSpecConfig {
+  // Parameters for differential privacy (DP).
+  //
+  // For detail, refer to "Dwork, C. and Roth, A., 2014. The algorithmic
+  // foundations of differential privacy. Foundations and Trends in Theoretical
+  // Computer Science, 9(3-4), pp.211-407."
+  message DifferentialPrivacyParams {
+    double epsilon = 1;
+    double delta = 2;
+  }
+
+  // Parameters that are used to generate `Reach` metric.
+  message ReachParams {
+    // Differential privacy parameters.
+    DifferentialPrivacyParams privacy_params = 1;
+  }
+  // Parameters that are used to generate `Frequency Histogram` metric.
+  message FrequencyHistogramParams {
+    // Differential privacy parameters.
+    DifferentialPrivacyParams privacy_params = 1;
+    // Maximum frequency cut-off value in the frequency histogram.
+    //
+    // Counts with frequency higher than `max_frequency` will be aggregated
+    // together.
+    int32 maximum_frequency_per_user = 2;
+  }
+  // Parameters that are used to generate `Impression Count` metric.
+  message ImpressionCountParams {
+    // Differential privacy parameters.
+    DifferentialPrivacyParams privacy_params = 1;
+    // Maximum frequency per user that will be included in this metric. Enforced
+    // on a per EDP basis.
+    //
+    // Setting the maximum frequency for each user is for noising the impression
+    // estimation with the noise proportional to maximum_frequency_per_user to
+    // guarantee epsilon-DP, i.e. the higher maximum_frequency_per_user, the
+    // larger the variance. On the other hand, if maximum_frequency_per_user is
+    // too small, there's a truncation bias.
+    int32 maximum_frequency_per_user = 2;
+  }
+  // Parameters that are used to generate `Watch Duration` metric.
+  message WatchDurationParams {
+    // Differential privacy parameters.
+    DifferentialPrivacyParams privacy_params = 1;
+    // Maximum watch duration per user in second that will be included in this
+    // metric. Enforced on a per EDP basis.
+    int32 maximum_watch_duration_per_user = 2;
+  }
+
+  // Specifies a range of VIDs to be sampled.
+  message VidSamplingInterval {
+    // The start of the sampling interval in [0, 1)
+    float start = 1;
+    // The width of the sampling interval.
+    float width = 2;
+  }
+  // Range of VIDs that will be included in this measurement
+  VidSamplingInterval vid_sampling_interval = 5;
+}
--- a/src/main/proto/wfa/measurement/config/reporting/metric_spec_config.proto
+++ b/src/main/proto/wfa/measurement/config/reporting/metric_spec_config.proto
@@ -34,16 +34,7 @@ message MetricSpecConfig {
   // Parameters that are used to generate `Reach` metric.
   message ReachParams {
     // Differential privacy parameters for reach.
-    DifferentialPrivacyParams reach_privacy_params = 1;
-    // Differential privacy parameters for frequency.
-    //
-    // Small default values are selected for the reach computation execution.
-    DifferentialPrivacyParams frequency_privacy_params = 2;
-    // Maximum frequency cut-off value in the frequency histogram.
-    //
-    // Counts with frequency higher than `max_frequency` will be aggregated
-    // together. For reach, this is equal to 1.
-    int32 maximum_frequency_per_user = 3;
+    DifferentialPrivacyParams privacy_params = 1;
   }
   // Parameters that are used to generate `Frequency Histogram` metric.
   message FrequencyHistogramParams {
--- a/src/main/proto/wfa/measurement/config/reporting/metric_spec_config.proto
+++ b/src/main/proto/wfa/measurement/config/reporting/metric_spec_config.proto
@@ -68,6 +68,17 @@ message MetricSpecConfig {
     int32 maximum_watch_duration_per_user = 2;
   }
 
+  // Parameters for generating the count of unique audiences reached given a set
+  // of event groups.
+  ReachFrequencyParams reach_params = 1;
+  // Parameters for generating the reach frequency histogram given a set of
+  // event groups.
+  ReachFrequencyParams frequency_histogram_params = 2;
+  // Parameters for generating the impression count given a set of event groups.
+  ImpressionCountParams impression_count_params = 3;
+  // Parameters for generating the watch duration given a set of event groups.
+  WatchDurationParams watch_duration_params = 4;
+
   // Specifies a range of VIDs to be sampled.
   message VidSamplingInterval {
     // The start of the sampling interval in [0, 1)
--- a/src/main/proto/wfa/measurement/internal/reporting/v2/metric.proto
+++ b/src/main/proto/wfa/measurement/internal/reporting/v2/metric.proto
@@ -33,8 +33,9 @@ message MetricSpec {
     DifferentialPrivacyParams privacy_params = 1;
   }
   message FrequencyHistogramParams {
-    DifferentialPrivacyParams privacy_params = 1;
-    int32 maximum_frequency_per_user = 2;
+    DifferentialPrivacyParams reach_privacy_params = 1;
+    DifferentialPrivacyParams frequency_privacy_params = 2;
+    int32 maximum_frequency_per_user = 3;
   }
   message ImpressionCountParams {
     DifferentialPrivacyParams privacy_params = 1;
--- a/src/main/proto/wfa/measurement/reporting/v2alpha/metric.proto
+++ b/src/main/proto/wfa/measurement/reporting/v2alpha/metric.proto
@@ -42,7 +42,7 @@ message MetricSpec {
     // Differential privacy parameters for reach.
     //
     // If this field is unspecified in a request message, the service
-    // implementation will choose a value.
+    // implementation will use the value specified in the system defaults.
     DifferentialPrivacyParams privacy_params = 1;
   }
   // Parameters that are used to generate `Frequency Histogram` metric.
@@ -50,12 +50,12 @@ message MetricSpec {
     // Differential privacy parameters for reach.
     //
     // If this field is unspecified in a request message, the service
-    // implementation will choose a value.
+    // implementation will use the value specified in the system defaults.
     DifferentialPrivacyParams reach_privacy_params = 1;
     // Differential privacy parameters for frequency distribution.
     //
     // If this field is unspecified in a request message, the service
-    // implementation will choose a value.
+    // implementation will use the value specified in the system defaults.
     DifferentialPrivacyParams frequency_privacy_params = 2;
     // Maximum frequency cut-off value in the frequency histogram.
     //
@@ -68,7 +68,7 @@ message MetricSpec {
     // Differential privacy parameters.
     //
     // If this field is unspecified in a request message, the service
-    // implementation will choose a value.
+    // implementation will use the value specified in the system defaults.
     DifferentialPrivacyParams privacy_params = 1;
     // Maximum frequency per user that will be included in this metric. Enforced
     // on a per EDP basis.
@@ -88,7 +88,7 @@ message MetricSpec {
     // Differential privacy parameters.
     //
     // If this field is unspecified in a request message, the service
-    // implementation will choose a value.
+    // implementation will use the value specified in the system defaults.
     DifferentialPrivacyParams privacy_params = 1;
     // Maximum watch duration per user in second that will be included in this
     // metric.
@@ -131,7 +131,7 @@ message MetricSpec {
   // Range of VIDs that will be included in this measurement
   //
   // If this field is unspecified in a request message, the service
-  // implementation will choose a value.
+  // implementation will use the value specified in the system defaults.
   VidSamplingInterval vid_sampling_interval = 5;
 }
 
--- a/src/main/proto/wfa/measurement/reporting/v2alpha/metric.proto
+++ b/src/main/proto/wfa/measurement/reporting/v2alpha/metric.proto
@@ -33,25 +33,29 @@ message MetricSpec {
   // foundations of differential privacy. Foundations and Trends in Theoretical
   // Computer Science, 9(3-4), pp.211-407."
   message DifferentialPrivacyParams {
-    double epsilon = 1;
-    double delta = 2;
+    double epsilon = 1 [(google.api.field_behavior) = REQUIRED];
+    double delta = 2 [(google.api.field_behavior) = REQUIRED];
   }
 
-  // Parameters that are used to generate `Reach` metric or
-  // `Frequency Histogram` metric.
-  message ReachFrequencyParams {
+  // Parameters that are used to generate `Reach` metric.
+  message ReachParams {
     // Differential privacy parameters for reach.
     //
-    // This field is either unset or fully set. The parameters in this field
-    // cannot be set partially. If this field is not set, the default values
-    // will be used and outputted here.
-    DifferentialPrivacyParams reach_privacy_params = 1;
-
-    // Differential privacy parameters for frequency.
+    // If this field is unspecified in a request message, the service
+    // implementation will choose a value.
+    DifferentialPrivacyParams privacy_params = 1;
+  }
+  // Parameters that are used to generate `Frequency Histogram` metric.
+  message FrequencyHistogramParams {
+    // Differential privacy parameters for reach.
     //
-    // This field is either unset or fully set. The parameters in this field
-    // cannot be set partially. If this field is not set, the default values
-    // will be used and outputted here.
+    // If this field is unspecified in a request message, the service
+    // implementation will choose a value.
+    DifferentialPrivacyParams reach_privacy_params = 1;
+    // Differential privacy parameters for frequency distribution.
+    //
+    // If this field is unspecified in a request message, the service
+    // implementation will choose a value.
     DifferentialPrivacyParams frequency_privacy_params = 2;
     // Maximum frequency cut-off value in the frequency histogram.
     //
@@ -63,9 +67,8 @@ message MetricSpec {
   message ImpressionCountParams {
     // Differential privacy parameters.
     //
-    // This field is either unset or fully set. The parameters in this field
-    // cannot be set partially. If this field is not set, the default values
-    // will be used and outputted here.
+    // If this field is unspecified in a request message, the service
+    // implementation will choose a value.
     DifferentialPrivacyParams privacy_params = 1;
     // Maximum frequency per user that will be included in this metric. Enforced
     // on a per EDP basis.
@@ -84,9 +87,8 @@ message MetricSpec {
   message WatchDurationParams {
     // Differential privacy parameters.
     //
-    // This field is either unset or fully set. The parameters in this field
-    // cannot be set partially. If this field is not set, the default values
-    // will be used and outputted here.
+    // If this field is unspecified in a request message, the service
+    // implementation will choose a value.
     DifferentialPrivacyParams privacy_params = 1;
     // Maximum watch duration per user in second that will be included in this
     // metric.
@@ -101,13 +103,13 @@ message MetricSpec {
   // Types of metric with parameters.
   oneof type {
     // The count of unique audiences reached given a set of event groups.
-    ReachFrequencyParams reach = 1 [(google.api.field_behavior) = IMMUTABLE];
+    ReachParams reach = 1 [(google.api.field_behavior) = IMMUTABLE];
     // The reach frequency histogram given a set of event groups.
     //
     // Currently, we only support union operations for frequency histograms. Any
     // other operations on frequency histograms won't guarantee the result is a
     // frequency histogram.
-    ReachFrequencyParams frequency_histogram = 2
+    FrequencyHistogramParams frequency_histogram = 2
         [(google.api.field_behavior) = IMMUTABLE];
     // The impression count given a set of event groups.
     ImpressionCountParams impression_count = 3
@@ -120,16 +122,16 @@ message MetricSpec {
   // Specifies a range of VIDs to be sampled.
   message VidSamplingInterval {
     // The start of the sampling interval in [0, 1)
-    float start = 1;
+    float start = 1 [(google.api.field_behavior) = REQUIRED];
     // The width of the sampling interval.
     //
     // start + width cannot be larger than 1
-    float width = 2;
+    float width = 2 [(google.api.field_behavior) = REQUIRED];
   }
   // Range of VIDs that will be included in this measurement
   //
-  // The parameters cannot be set partially. If no parameters are set, the
-  // default values will be used and outputted here.
+  // If this field is unspecified in a request message, the service
+  // implementation will choose a value.
   VidSamplingInterval vid_sampling_interval = 5;
 }
 
--- a/src/main/proto/wfa/measurement/reporting/v2alpha/metric.proto
+++ b/src/main/proto/wfa/measurement/reporting/v2alpha/metric.proto
@@ -40,20 +40,30 @@ message MetricSpec {
   // Parameters that are used to generate `Reach` metric.
   message ReachParams {
     // Differential privacy parameters.
+    //
+    // The parameters cannot be set partially. If no parameters are set, the
+    // default values will be used and outputted here.
     DifferentialPrivacyParams privacy_params = 1;
   }
   // Parameters that are used to generate `Frequency Histogram` metric.
   message FrequencyHistogramParams {
     // Differential privacy parameters.
+    //
+    // The parameters cannot be set partially. If no parameters are set, the
+    // default values will be used and outputted here.
     DifferentialPrivacyParams privacy_params = 1;
-    // Maximum frequency cut-off value in the frequency histogram. Counts with
-    // frequency higher than `max_frequency` will be aggregated together.
-    int32 maximum_frequency_per_user = 2
-        [(google.api.field_behavior) = REQUIRED];
+    // Maximum frequency cut-off value in the frequency histogram.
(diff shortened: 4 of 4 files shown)


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/895)

SanjayVas (changes requested):
> Reviewed 1 of 2 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/895)* status: 1 of 2 files reviewed, 3 unresolved discussions (waiting on @kungfucraig, @riemanli, and @tristanvuong2021)

*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/895#-NQlvIvV8xU05eAILyJ6:-NQlvIvV8xU05eAILyJ7:b-97ulld) (no related file):*
> Due to the incorporation of DP params, reach metric and reach-frequency metric need to share the same params message.

Why? This seems quite odd, as you have frequency DP params even for a reach-only Metric.

___
___
*[`src/main/proto/wfa/measurement/reporting/v2alpha/metric.proto` line 36 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/895#-NQlvrp30mjGdbEgfye9:-NQlvrp30mjGdbEgfyeA:b-i78v10) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d7207e068b62029c3ce2090ff0c7cc72489bab05/src/main/proto/wfa/measurement/reporting/v2alpha/metric.proto#L36)):*
> ```protobuf
>   // Computer Science, 9(3-4), pp.211-407."
>   message DifferentialPrivacyParams {
>     double epsilon = 1;
> ```

Mark these as REQUIRED.



_Code quote:_
```protobuf
    double epsilon = 1;
    double delta = 2;
```
___
*[`src/main/proto/wfa/measurement/reporting/v2alpha/metric.proto` line 52 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/895#-NQlvZuiC_7VxE0BY37_:-NQlvZuiC_7VxE0BY37a:b4r5z9g) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d7207e068b62029c3ce2090ff0c7cc72489bab05/src/main/proto/wfa/measurement/reporting/v2alpha/metric.proto#L52)):*
> ```protobuf
>     // Differential privacy parameters for frequency.
>     //
>     // This field is either unset or fully set. The parameters in this field
> ```

This isn't necessary. Just phrasing the latter part as "If this field is unspecified in a request message, the service implementation will choose a value." should be sufficient. The fields inside the message are required, which already covers the fact that this field can only be specified when all of its fields are populated.

The only reason we have to document this at all is that we're not using proto3 defaults but rather choosing some other defaults at the application layer.



_Code quote:_
```protobuf
This field is either unset or fully set.
```

riemanli (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/895)* status: 1 of 2 files reviewed, 3 unresolved discussions (waiting on @jiayu-google, @kungfucraig, and @tristanvuong2021)

*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/895#-NQlvIvV8xU05eAILyJ6:-NQlzB288p9-s2PBUVp9:blh4dbx) (no related file):*

> Due to the incorporation of DP params, reach metric and reach-frequency metric need to share the same params message.

Why? This seems quite odd, as you have frequency DP params even for a reach-only Metric.
</blockquote></details>

As [this](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/main/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v1alpha/ReportsService.kt#L225-L236) shows, reach-only still needs to fill the privacy params for frequency because in the kingdom there is only one type "reach-frequency".  This follows the measurement spec in the cmms api .

Maybe @jiayu-google can confirm if `epsilon` for frequency would impact reach-only performance (in terms of PB consumption, speed and accuracy) or not. If not, we can fall back to two different params messages and use an internal default value for the frequency epsilon in reach-only.

___

jiayu-google (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/895)* status: 1 of 2 files reviewed, 3 unresolved discussions (waiting on @kungfucraig, @riemanli, @SanjayVas, and @tristanvuong2021)

*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/895#-NQlvIvV8xU05eAILyJ6:-NQxB34N4WOsJzNpMVQR:bbtuu0r) (no related file):*

As [this](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/main/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v1alpha/ReportsService.kt#L225-L236) shows, reach-only still needs to fill the privacy params for frequency because in the kingdom there is only one type "reach-frequency".  This follows the measurement spec in the cmms api .

Maybe @jiayu-google can confirm if `epsilon` for frequency would impact reach-only performance (in terms of PB consumption, speed and accuracy) or not. If not, we can fall back to two different params messages and use an internal default value for the frequency epsilon in reach-only.
</blockquote></details>

It impacts. Each reach-only measurement will consume `reach.epsilon + frequency.epsilon`.  Ideally we want to set `frequency.epsilon = 0` in reach-only msmts, but LLV2 requires `frequency.epsilon > 0`. So, we set a tiny frequency.epsilon = 1e-4.

___

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/895)* status: 1 of 2 files reviewed, 3 unresolved discussions (waiting on @jiayu-google, @kungfucraig, @riemanli, and @tristanvuong2021)

*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/895#-NQlvIvV8xU05eAILyJ6:-NQzd85Y9qu5mjvudeU5:b-jgqhiq) (no related file):*
<details><summary><i>Previously, jiayu-google wrote…</i></summary><blockquote>

It impacts. Each reach-only measurement will consume `reach.epsilon + frequency.epsilon`.  Ideally we want to set `frequency.epsilon = 0` in reach-only msmts, but LLV2 requires `frequency.epsilon > 0`. So, we set a tiny frequency.epsilon = 1e-4.
</blockquote></details>

Let's avoid leaking the fact that we don't have a true reach-only measurement implemented in the CMMS API and build this API assuming that we do support reach-only.

___

riemanli (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/895)* status: 0 of 2 files reviewed, 3 unresolved discussions (waiting on @jiayu-google, @kungfucraig, @SanjayVas, and @tristanvuong2021)

*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/895#-NQlvIvV8xU05eAILyJ6:-NR-2io00PAkb9bfkGrN:bmn5wj3) (no related file):*

Let's avoid leaking the fact that we don't have a true reach-only measurement implemented in the CMMS API and build this API assuming that we do support reach-only.
</blockquote></details>

@jiayu-google Thanks for confirming! Is a smaller frequency.epsilon possible? Ex: 1e--7

@SanjayVas I made the change to separate reach params and frequency params to two types in `Metric`. For `MetricSpecConfig`, I also separated them, but since we still need to store the default values of the frequency DP params and `maximum_frequency_per_user` for reach-only, the `ReachParams` in the spec config has two additional fields. Let me know if that looks fine to you.

___
___
*[`src/main/proto/wfa/measurement/reporting/v2alpha/metric.proto` line 36 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/895#-NQlvrp30mjGdbEgfye9:-NR--AZkE3QKkXSK_3A3:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d7207e068b62029c3ce2090ff0c7cc72489bab05/src/main/proto/wfa/measurement/reporting/v2alpha/metric.proto#L36)):*

Mark these as REQUIRED.
</blockquote></details>

Done.
___
*[`src/main/proto/wfa/measurement/reporting/v2alpha/metric.proto` line 52 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/895#-NQlvZuiC_7VxE0BY37_:-NR-4RG4AbRlI0yrmMqc:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d7207e068b62029c3ce2090ff0c7cc72489bab05/src/main/proto/wfa/measurement/reporting/v2alpha/metric.proto#L52)):*

This isn't necessary. Just phrasing the latter part as "If this field is unspecified in a request message, the service implementation will choose a value." should be sufficient. The fields inside the message are required, which already covers the fact that this field can only be specified when all of its fields are populated.

The only reason we have to document this at all is that we're not using proto3 defaults but rather choosing some other defaults at the application layer.
</blockquote></details>

Done.

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/895)* status: 0 of 2 files reviewed, 3 unresolved discussions (waiting on @jiayu-google, @kungfucraig, @riemanli, and @tristanvuong2021)

*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/895#-NQlvIvV8xU05eAILyJ6:-NR-5fcl3pGsXwAJGpkb:bra176l) (no related file):*
> , but since we still need to store the default values of the frequency DP params and maximum_frequency_per_user for reach-only, the ReachParams in the spec config has two additional fields. 

Do we need these fields visible in the public API?

___

riemanli (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/895)* status: 0 of 2 files reviewed, 3 unresolved discussions (waiting on @jiayu-google, @kungfucraig, @SanjayVas, and @tristanvuong2021)

*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/895#-NQlvIvV8xU05eAILyJ6:-NR-9p5uEVYgfHPnr5wX:bfpo22l) (no related file):*

> , but since we still need to store the default values of the frequency DP params and maximum_frequency_per_user for reach-only, the ReachParams in the spec config has two additional fields. 

Do we need these fields visible in the public API?
</blockquote></details>

Offline discussion: to make the API change in this PR more straightforward, we decided to add a reach-only measurement type to the CMMS API, and this PR will depend on it.

___

kungfucraig (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/895)* status: 0 of 2 files reviewed, 6 unresolved discussions (waiting on @jiayu-google, @riemanli, @SanjayVas, and @tristanvuong2021)

___
*[`src/main/proto/wfa/measurement/config/reporting/metric_spec_config.proto` line 23 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/895#-NR-azpl75qVR8q_JV_m:-NR-azpl75qVR8q_JV_n:b-d1vvnt) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1272650bb0af8b8a6db98d680d63701b96ccfc4b/src/main/proto/wfa/measurement/config/reporting/metric_spec_config.proto#L23)):*
> ```protobuf
> 
> // The configuration for metric spec in Metric in the reporting server
> message MetricSpecConfig {
> ```

Where is this being used?
___
*[`src/main/proto/wfa/measurement/config/reporting/metric_spec_config.proto` line 41 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/895#-NR-aLCM0wJz2osaBErg:-NR-aLCM0wJz2osaBErh:b-qfuc5y) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1272650bb0af8b8a6db98d680d63701b96ccfc4b/src/main/proto/wfa/measurement/config/reporting/metric_spec_config.proto#L41)):*
> ```protobuf
>     //
>     // Small default values are selected for the reach computation execution.
>     DifferentialPrivacyParams frequency_privacy_params = 2;
> ```

Do you mean to have this here?
___
*[`src/main/proto/wfa/measurement/reporting/v2alpha/metric.proto` line 45 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/895#-NR-arXS0-KV9m7vSX0Q:-NR-arXS0-KV9m7vSX0R:b-jldr5a) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1272650bb0af8b8a6db98d680d63701b96ccfc4b/src/main/proto/wfa/measurement/reporting/v2alpha/metric.proto#L45)):*
> ```protobuf
>     //
>     // If this field is unspecified in a request message, the service
>     // implementation will choose a value.
> ```

/choose a value/use the value specified in the system defaults.

riemanli (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/895)* status: 0 of 2 files reviewed, 5 unresolved discussions (waiting on @jiayu-google, @kungfucraig, @SanjayVas, and @tristanvuong2021)

___
*[`src/main/proto/wfa/measurement/config/reporting/metric_spec_config.proto` line 23 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/895#-NR-azpl75qVR8q_JV_m:-NR4DjHp27I5WZ34Vm_n:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1272650bb0af8b8a6db98d680d63701b96ccfc4b/src/main/proto/wfa/measurement/config/reporting/metric_spec_config.proto#L23)):*

Thanks! I missed that.
</blockquote></details>

Done.
___
*[`src/main/proto/wfa/measurement/config/reporting/metric_spec_config.proto` line 41 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/895#-NR-aLCM0wJz2osaBErg:-NR4DoRq9_pwevOpTGe4:bb77uf3) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1272650bb0af8b8a6db98d680d63701b96ccfc4b/src/main/proto/wfa/measurement/config/reporting/metric_spec_config.proto#L41)):*

The conclusion in the discussion above seems to indicate we'll add REACH as a measurement type in the Kingdom APIs, which would be dropping this. Is that right?
</blockquote></details>

Yes!

SanjayVas (changes requested):
> Reviewed 1 of 2 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/895)* status: all files reviewed, 4 unresolved discussions (waiting on @jiayu-google, @kungfucraig, @riemanli, and @tristanvuong2021)

___
*[`src/main/proto/wfa/measurement/reporting/v2alpha/metric.proto` line 58 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/895#-NR4T61Q-jIPptWsE-Au:-NR4T61Q-jIPptWsE-Av:b754prp) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/24fcc6619ad4f028a227599840e193c388df2dc1/src/main/proto/wfa/measurement/reporting/v2alpha/metric.proto#L58)):*
> ```protobuf
>     //
>     // If this field is unspecified in a request message, the service
>     // implementation will use the value specified in the system defaults.
> ```

nit: I avoided referencing defaults in my suggestion so that we would be free to change the implementation without changing the API. e.g. we could have this be determined dynamically.



_Suggestion:_
```protobuf
will choose a value
```

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/895)* status: all files reviewed, 4 unresolved discussions (waiting on @jiayu-google, @kungfucraig, and @tristanvuong2021)

___
*[`src/main/proto/wfa/measurement/reporting/v2alpha/metric.proto` line 58 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/895#-NR4T61Q-jIPptWsE-Au:-NR5K2FOD0sGMA5W7ohQ:bhl0he8) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/24fcc6619ad4f028a227599840e193c388df2dc1/src/main/proto/wfa/measurement/reporting/v2alpha/metric.proto#L58)):*

Undocumented behavior is annoying for someone who is developing against an API because if you need to understand how things work you are forced to go read the code. Plus an undocumented behavior will come to be relied upon to the point that you cannot change it anyway. 

If you feel strongly about this I'll back down though. :)
</blockquote></details>

It's just about having things documented in the right place. This is documenting the API for the caller, so we should leave out implementation details that the caller need not know.

e.g. the caller (intentionally) has no way to know what said defaults actually are, so mentioning that there are defaults isn't useful. They just have to observe what is returned in the response on a per-call basis. Stating that there are defaults may lead to the incorrect assumption that the response value is static and there is therefore no need to actually inspect each response.

jiayu-google (approved):
I confirm that all the needed parameters are included and are well explained in the documentation.

jiayu-google (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/895)* status: 1 of 3 files reviewed, 4 unresolved discussions (waiting on @kungfucraig, @SanjayVas, @stevenwarejones, and @tristanvuong2021)

___
*[`src/main/proto/wfa/measurement/reporting/v2alpha/metric.proto` line 36 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/895#-NQlvrp30mjGdbEgfye9:-NRsqtrS-1QOoLzhPOjL:b-12ppmy) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d7207e068b62029c3ce2090ff0c7cc72489bab05/src/main/proto/wfa/measurement/reporting/v2alpha/metric.proto#L36)):*

In general delta is not always required, but for our use cases (MPC + direct) I believe it is.

@jiayu-google can you confirm?
</blockquote></details>

sorry for replying late... Yes I confirm we require delta for Laplacian noise. (With CDP though one can specify replace epsilon-delta by a single rho.)

stevenwarejones (changes requested):
> Reviewed 1 of 1 files at r3, 2 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/895)* status: all files reviewed, 4 unresolved discussions (waiting on @kungfucraig, @SanjayVas, and @tristanvuong2021)

___
*[`src/main/proto/wfa/measurement/reporting/v2alpha/metric.proto` line 36 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/895#-NQlvrp30mjGdbEgfye9:-NRtW7im-qJzNN8uyIUT:blr21xs) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d7207e068b62029c3ce2090ff0c7cc72489bab05/src/main/proto/wfa/measurement/reporting/v2alpha/metric.proto#L36)):*
<details><summary><i>Previously, jiayu-google wrote…</i></summary><blockquote>

sorry for replying late... Yes I confirm we require delta for Laplacian noise. (With CDP though one can specify replace epsilon-delta by a single rho.)
</blockquote></details>

so, is it required based on the noiser? I don't think these should be required

riemanli (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/895)* status: all files reviewed, 4 unresolved discussions (waiting on @kungfucraig, @SanjayVas, and @tristanvuong2021)

___
*[`src/main/proto/wfa/measurement/reporting/v2alpha/metric.proto` line 36 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/895#-NQlvrp30mjGdbEgfye9:-NRtoFDA54yQSL3FNno7:b-60d28c) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d7207e068b62029c3ce2090ff0c7cc72489bab05/src/main/proto/wfa/measurement/reporting/v2alpha/metric.proto#L36)):*

so, is it required based on the noiser? I don't think these should be required
</blockquote></details>

Based on the offline discussion with @jiayu-google, delta is optional for most cases but it should have a lower bound to avoid MPC crashing. That being said, if we make delta not required, the check process will be like:
1. If differentialPrivacyParams is set and delta is unset, delta is equal to 0 based on the default value for Double in protobuf. In this case, we will substitute 0 with the default value for delta.
2. If differentialPrivacyParams is set and delta is set (>0), we need to check if delta is great than the required min value. If not, either raise an error or substitute it with the default value.
3. If differentialPrivacyParams is unset, use the default value for delta.

Does that sound good to everyone?

kungfucraig (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/895)* status: all files reviewed, 4 unresolved discussions (waiting on @SanjayVas and @tristanvuong2021)

___
*[`src/main/proto/wfa/measurement/reporting/v2alpha/metric.proto` line 36 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/895#-NQlvrp30mjGdbEgfye9:-NS6h5H8AFZmwF3K_sgq:b-505je7) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d7207e068b62029c3ce2090ff0c7cc72489bab05/src/main/proto/wfa/measurement/reporting/v2alpha/metric.proto#L36)):*

Based on the offline discussion with @jiayu-google, delta is optional for most cases but it should have a lower bound to avoid MPC crashing. That being said, if we make delta not required, the check process will be like:
1. If differentialPrivacyParams is set and delta is unset, delta is equal to 0 based on the default value for Double in protobuf. In this case, we will substitute 0 with the default value for delta.
2. If differentialPrivacyParams is set and delta is set (>0), we need to check if delta is great than the required min value. If not, either raise an error or substitute it with the default value.
3. If differentialPrivacyParams is unset, use the default value for delta.

Does that sound good to everyone?
</blockquote></details>

Should we update this structure so that we have (conceptually speaking) oneof { struct { epsilon, delta}, rho}?

With respect to the default delta. #1 sounds good, #2 should be an error, #3 sounds good. I think you can say the following:

If delta or epsilon are unset a configuration specific default is used. An error is thrown if delta is set and is less than a configuration specific minimum.

I imagine there should be a minimum epsilon as well.

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/895)* status: all files reviewed, 4 unresolved discussions (waiting on @kungfucraig, @SanjayVas, and @tristanvuong2021)

___
*[`src/main/proto/wfa/measurement/reporting/v2alpha/metric.proto` line 36 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/895#-NQlvrp30mjGdbEgfye9:-NS7MUOnAXTE3G1T0ll-:bxgpc3o) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d7207e068b62029c3ce2090ff0c7cc72489bab05/src/main/proto/wfa/measurement/reporting/v2alpha/metric.proto#L36)):*

Should we update this structure so that we have (conceptually speaking) oneof { struct { epsilon, delta}, rho}?

With respect to the default delta. #1 sounds good, #2 should be an error, #3 sounds good. I think you can say the following:

If delta or epsilon are unset a configuration specific default is used. An error is thrown if delta is set and is less than a configuration specific minimum.

I imagine there should be a minimum epsilon as well.

</blockquote></details>

do we want the API to distinguish an unset of delta from a delta of zero? If that's the case, then delta needs to be in its own message.

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/895)* status: all files reviewed, 4 unresolved discussions (waiting on @jiayu-google, @kungfucraig, @stevenwarejones, and @tristanvuong2021)

___
*[`src/main/proto/wfa/measurement/reporting/v2alpha/metric.proto` line 36 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/895#-NQlvrp30mjGdbEgfye9:-NS8MlVOEDz9am6eqRR0:b-l0cpsm) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d7207e068b62029c3ce2090ff0c7cc72489bab05/src/main/proto/wfa/measurement/reporting/v2alpha/metric.proto#L36)):*
I think I'd prefer the all-or-nothing behavior I initially described. If you specify a value for `epsilon`, it means you must also specify a non-zero value for `delta`. The alternative is that you instead make it so that `epsilon` and `delta` are both optional and will have defaults specified if unset. You basically never check whether fields of type `DifferentialPrivacyParams` are set in this case, and instead only check the presence of `epsilon` and `delta`.

> do we want the API to distinguish an unset of delta from a delta of zero? If that's the case, then delta needs to be in its own message.

proto3 has an `optional` keyword to add explicit field presence to scalar fields. See https://protobuf.dev/programming-guides/field_presence/#presence-in-proto3-apis

riemanli (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/895)* status: all files reviewed, 4 unresolved discussions (waiting on @jiayu-google, @kungfucraig, @SanjayVas, @stevenwarejones, and @tristanvuong2021)

___
*[`src/main/proto/wfa/measurement/reporting/v2alpha/metric.proto` line 36 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/895#-NQlvrp30mjGdbEgfye9:-NS8Y8vm3I_lRJT9uNYJ:bhe3da7) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d7207e068b62029c3ce2090ff0c7cc72489bab05/src/main/proto/wfa/measurement/reporting/v2alpha/metric.proto#L36)):*

I think I'd prefer the all-or-nothing behavior I initially described. If you specify a value for `epsilon`, it means you must also specify a non-zero value for `delta`. The alternative is that you instead make it so that `epsilon` and `delta` are both optional and will have defaults specified if unset. You basically never check whether fields of type `DifferentialPrivacyParams` are set in this case, and instead only check the presence of `epsilon` and `delta`.

> do we want the API to distinguish an unset of delta from a delta of zero? If that's the case, then delta needs to be in its own message.

proto3 has an `optional` keyword to add explicit field presence to scalar fields. See https://protobuf.dev/programming-guides/field_presence/#presence-in-proto3-apis
</blockquote></details>

For the latter option, could we do required `epsilon` and optional `delta` instead?

kungfucraig (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/895)* status: all files reviewed, 4 unresolved discussions (waiting on @jiayu-google, @SanjayVas, @stevenwarejones, and @tristanvuong2021)

___
*[`src/main/proto/wfa/measurement/reporting/v2alpha/metric.proto` line 36 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/895#-NQlvrp30mjGdbEgfye9:-NSCj3jN4n-JDfQ5CJKa:b5epznf) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d7207e068b62029c3ce2090ff0c7cc72489bab05/src/main/proto/wfa/measurement/reporting/v2alpha/metric.proto#L36)):*

> For the latter option, could we do required `epsilon` and optional `delta` instead?

I suppose. It basically means you have the following three valid scenarios:

1. A request field of type `DifferentialPrivacyParams` is not set.
    * Both `epsilon` and `delta` will have server-chosen defaults.
2. A request field of type `DifferentialPrivacyParams` is set, with its child `delta` field unspecified.
    * `delta` will have a server-chosen default.
3. A request field of type `DifferentialPrivacyParams` is set with all child fields specified.

You'd need to document this behavior succinctly for each request field of type `DifferentialPrivacyParams`, as the existing comment is no longer accurate. If instead you just have both `epsilon` and `delta` be optional, it's easier to describe the behavior of the `DifferentialPrivacyParams` message itself rather than on each field of that type: "If any of the fields are not specified in a request message, the service implementation will choose values."
</blockquote></details>

Making them both optional seems like the right thing to do. This means the user only has to provide a value if they really understand what they are doing.

riemanli (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/895)* status: 2 of 3 files reviewed, 4 unresolved discussions (waiting on @jiayu-google, @kungfucraig, @SanjayVas, @stevenwarejones, and @tristanvuong2021)

___
*[`src/main/proto/wfa/measurement/reporting/v2alpha/metric.proto` line 36 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/895#-NQlvrp30mjGdbEgfye9:-NSDCJyG5zUjkjwXNoZQ:b-nwq4vr) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d7207e068b62029c3ce2090ff0c7cc72489bab05/src/main/proto/wfa/measurement/reporting/v2alpha/metric.proto#L36)):*

Making them both optional seems like the right thing to do. This means the user only has to provide a value if they really understand what they are doing.
</blockquote></details>

Made `DifferentialPrivacyParams` a required field in each measurement type, but its fields are optional. 
___
*[`src/main/proto/wfa/measurement/reporting/v2alpha/metric.proto` line 58 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/895#-NR4T61Q-jIPptWsE-Au:-NSDCgLe2rrt8e6Sf90-:b-1ybeuq) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/24fcc6619ad4f028a227599840e193c388df2dc1/src/main/proto/wfa/measurement/reporting/v2alpha/metric.proto#L58)):*

It's just about having things documented in the right place. This is documenting the API for the caller, so we should leave out implementation details that the caller need not know.

e.g. the caller (intentionally) has no way to know what said defaults actually are, so mentioning that there are defaults isn't useful. They just have to observe what is returned in the response on a per-call basis. Stating that there are defaults may lead to the incorrect assumption that the response value is static and there is therefore no need to actually inspect each response.
</blockquote></details>

Documented as Sanjay suggested. Let me know if there is objection.

stevenwarejones (changes requested):
> Reviewed 2 of 2 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/895)* status: all files reviewed, 5 unresolved discussions (waiting on @jiayu-google, @kungfucraig, @riemanli, @SanjayVas, and @tristanvuong2021)

___
*[`.bazelrc` line 22 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/895#-NSGk14sDjuv7ME2nsxm:-NSGk14sDjuv7ME2nsxn:b-96xxut) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/637ddd3e2c41bad4edb1b929b8c0d49268ae26d9/.bazelrc#L22)):*
> ```
> # Enable explicit presence in proto3.
> # See https://github.com/protocolbuffers/protobuf/blob/main/docs/field_presence.md#how-to-enable-explicit-presence-in-proto3
> build --protocopt=--experimental_allow_proto3_optional
> ```

I don't think we should be enabling experimental flags unless we have a strong reason.
___
*[`src/main/proto/wfa/measurement/reporting/v2alpha/metric.proto` line 36 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/895#-NQlvrp30mjGdbEgfye9:-NSGkRG66yKwIoiHzZpn:b62k2to) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d7207e068b62029c3ce2090ff0c7cc72489bab05/src/main/proto/wfa/measurement/reporting/v2alpha/metric.proto#L36)):*

Made `DifferentialPrivacyParams` a required field in each measurement type, but its fields are optional. 
</blockquote></details>

see my other comment. I'd prefer to not need experimental flags. 

Another option is to put delta and epsilon in their own messages.

SanjayVas (approved):
> Reviewed 2 of 2 files at r4, 2 of 2 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/895)* status: all files reviewed, 3 unresolved discussions (waiting on @jiayu-google, @kungfucraig, @stevenwarejones, and @tristanvuong2021)

___
*[`.bazelrc` line 22 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/895#-NSGk14sDjuv7ME2nsxm:-NSHZeTZ0IBvQI4A14lK:bd5yc30) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/637ddd3e2c41bad4edb1b929b8c0d49268ae26d9/.bazelrc#L22)):*

I don't think we should be enabling experimental flags unless we have a strong reason.
</blockquote></details>

This feature has been generally available for awhile and is used broadly in Google internal code. I'm not actually sure why it's still marked as experimental in the open source protoc and also not just on by default. If I had to hazard a guess, it's just being conservative on backwards compatibility.
___
*[`src/main/proto/wfa/measurement/reporting/v2alpha/metric.proto` line 36 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/895#-NQlvrp30mjGdbEgfye9:-NSH_Fj18frUhFA-9W0Z:bxckthj) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d7207e068b62029c3ce2090ff0c7cc72489bab05/src/main/proto/wfa/measurement/reporting/v2alpha/metric.proto#L36)):*
> see my other comment. I'd prefer to not need experimental flags.

See my comment on the "experimental" status of the `optional` keyword in proto3.

riemanli (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/895)* status: all files reviewed, 3 unresolved discussions (waiting on @jiayu-google, @kungfucraig, @stevenwarejones, and @tristanvuong2021)

___
*[`.bazelrc` line 22 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/895#-NSGk14sDjuv7ME2nsxm:-NSHicSB0FA4MRGlNdyy:bo1ewz1) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/637ddd3e2c41bad4edb1b929b8c0d49268ae26d9/.bazelrc#L22)):*

This feature has been generally available for awhile and is used broadly in Google internal code. I'm not actually sure why it's still marked as experimental in the open source protoc and also not just on by default. If I had to hazard a guess, it's just being conservative on backwards compatibility.
</blockquote></details>

Based on what the readme described, it seems like if we force `protoc` version at least v3.15, it's on by default. Will that be a good workaround?

SanjayVas (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/895)* status: all files reviewed, 3 unresolved discussions (waiting on @jiayu-google, @kungfucraig, @stevenwarejones, and @tristanvuong2021)

___
*[`.bazelrc` line 22 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/895#-NSGk14sDjuv7ME2nsxm:-NSHndsZBqC260NBEXcP:b1842xd) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/637ddd3e2c41bad4edb1b929b8c0d49268ae26d9/.bazelrc#L22)):*

Based on what the readme described, it seems like if we force `protoc` version at least v3.15, it's on by default. Will that be a good workaround?
</blockquote></details>

I don't believe that's accurate. It's just that you need to be on at least that version to even have the option to turn it on. We're already on 3.21, I believe.

riemanli (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/895)* status: 3 of 4 files reviewed, 3 unresolved discussions (waiting on @jiayu-google, @kungfucraig, @SanjayVas, @stevenwarejones, and @tristanvuong2021)

___
*[`src/main/proto/wfa/measurement/reporting/v2alpha/metric.proto` line 36 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/895#-NQlvrp30mjGdbEgfye9:-NSI5ask9OMft_vbcTBF:b-bjroof) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d7207e068b62029c3ce2090ff0c7cc72489bab05/src/main/proto/wfa/measurement/reporting/v2alpha/metric.proto#L36)):*

> see my other comment. I'd prefer to not need experimental flags.

See my comment on the "experimental" status of the `optional` keyword in proto3.


</blockquote></details>

The `optional` tag can work without the experimental flags. Please see my comment in the other thread.
___
*[`.bazelrc` line 22 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/895#-NSGk14sDjuv7ME2nsxm:-NSI4RXeDmD8Mbo1Rktj:b5s2r60) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/637ddd3e2c41bad4edb1b929b8c0d49268ae26d9/.bazelrc#L22)):*

See https://github.com/protocolbuffers/protobuf/issues/9684 for some discussion on the protoc option.
</blockquote></details>

Sorry I probably did something wrong when I tested the `optional` tag without this `.bazelrc` arg. After testing it again with a fresh IntelliJ project (i.e. no leftover building cache), the `optional` tag works alone. So we should be fine using it. I will remove this `.bazelrc` arg.

riemanli (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/895)* status: 3 of 4 files reviewed, 4 unresolved discussions (waiting on @jiayu-google, @kungfucraig, @SanjayVas, @stevenwarejones, and @tristanvuong2021)

*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/895#-NSIbDxt4FK6b86tvKgS:-NSIbDxt4FK6b86tvKgT:bu6kdj5) (no related file):*
@stevenwarejones, @kungfucraig I think the design about the DP params is clear and determined after finding out that `optional` tag can be used without experimental tag. Let me know if there is any other concerns. Thanks!

___

kungfucraig (approved):
> Reviewed 2 of 2 files at r4, 1 of 2 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/895)* status: 3 of 4 files reviewed, 2 unresolved discussions (waiting on @jiayu-google, @SanjayVas, @stevenwarejones, and @tristanvuong2021)

jiayu-google (approved):

tristanvuong2021 (changes requested):
> Reviewed 2 of 2 files at r4, 1 of 2 files at r5, 1 of 1 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/895)* status: all files reviewed, 3 unresolved discussions (waiting on @riemanli and @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/config/reporting/metric_spec_config.proto` line 1 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/895#-NSNZvdS6NkKqjpooV6g:-NSNZvdS6NkKqjpooV6h:bb6ht38) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a91810538414e136479ff8e3e48c981d0f3e3d34/src/main/proto/wfa/measurement/config/reporting/metric_spec_config.proto#L1)):*
> ```protobuf
> // Copyright 2022 The Cross-Media Measurement Authors
> ```

2023

riemanli (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/895)* status: 3 of 4 files reviewed, 3 unresolved discussions (waiting on @kungfucraig, @SanjayVas, @stevenwarejones, and @tristanvuong2021)

___
*[`src/main/proto/wfa/measurement/config/reporting/metric_spec_config.proto` line 1 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/895#-NSNZvdS6NkKqjpooV6g:-NSNaR7SA0tE3H_B3CQj:b-x8yi4b) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a91810538414e136479ff8e3e48c981d0f3e3d34/src/main/proto/wfa/measurement/config/reporting/metric_spec_config.proto#L1)):*

2023
</blockquote></details>

Thanks for catching it! Done!

tristanvuong2021 (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/895)* status: all files reviewed, 2 unresolved discussions (waiting on @stevenwarejones)

stevenwarejones (approved):
> Reviewed 1 of 1 files at r6, 1 of 1 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/895)* status: all files reviewed, 1 unresolved discussion (waiting on @jiayu-google, @kungfucraig, @SanjayVas, and @tristanvuong2021)

riemanli (commented):
> Reviewed 1 of 2 files at r4, 1 of 2 files at r5, 1 of 1 files at r6, 1 of 1 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/895)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @riemanli)

PR #898, opened 2023-03-17
Its to be able to create Kingdom separately

wfa-siyengar:
With this PR we will be able to create a Kingdom separately.


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/898)

PR #902, opened 2023-03-20
Duchy claims tasks based on CreationTime

renjiezh:


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/902)

SanjayVas (changes requested):
> Reviewed 8 of 10 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/902)* status: all files reviewed, 4 unresolved discussions (waiting on @renjiezh)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/MillBase.kt` line 174 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/902#-NR4Kmdd4KqbNXHoQbdH:-NR4Kmdd4KqbNXHoQbdI:b1w4aky) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e3e24ea3d3f82e058e49fbb1b4bfb8e5bf5d423d/src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/MillBase.kt#L174)):*
> ```Kotlin
>       val token = claimWorkResponse.token
>       if (token.attempt > maximumAttempts) {
>         val errorMessage = "Failing computation due to too many failed ComputationStageAttempts."
> ```

This code block appears in multiple places. Extract it to a method?



_Code quote:_
```Kotlin
        val errorMessage = "Failing computation due to too many failed ComputationStageAttempts."
        logger.log(Level.SEVERE, "${token.globalComputationId}@$millId: $errorMessage")
        failComputationAtKingdom(token, errorMessage)
        completeComputation(token, CompletedReason.FAILED)
```
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/gcloud/spanner/computation/UnclaimedTasksQuery.kt` line 45 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/902#-NR4Ii0HCRzVF5GlsLOV:-NR4Ii0HCRzVF5GlsLOW:b-pyr0nb) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e3e24ea3d3f82e058e49fbb1b4bfb8e5bf5d423d/src/main/kotlin/org/wfanet/measurement/duchy/deploy/gcloud/spanner/computation/UnclaimedTasksQuery.kt#L45)):*
> ```Kotlin
>       WHERE c.Protocol = @protocol
>         AND c.LockExpirationTime IS NOT NULL
>         AND c.CreationTime IS NOT NULL
> ```

This clause means that previously created computations never get claimed. Instead, we just want them to be ordered such that they are claimed before ones with a creation time.
___
*[`src/main/resources/duchy/spanner/add-computation-creation-time.sql` line 18 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/902#-NR4JnbWCODT8xXhJPFP:-NR4JnbWCODT8xXhJPFQ:b3zuau4) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e3e24ea3d3f82e058e49fbb1b4bfb8e5bf5d423d/src/main/resources/duchy/spanner/add-computation-creation-time.sql#L18)):*
> ```TSQL
> 
> -- changeset renjiez:3 dbms:cloudspanner
> START BATCH DDL;
> ```

nit: we don't need to use BATCH mode when there's only one statement



_Code quote:_
```TSQL
START BATCH DDL;
```
___
*[`src/main/resources/duchy/spanner/add-computation-creation-time.sql` line 24 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/902#-NR4JSyK8v9wsNIA8mVl:-NR4JSyK8v9wsNIA8mVm:b-yhwdkl) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e3e24ea3d3f82e058e49fbb1b4bfb8e5bf5d423d/src/main/resources/duchy/spanner/add-computation-creation-time.sql#L24)):*
> ```TSQL
>     -- and option `allow_commit_timestamp`, this column has to be nullable. The
>     -- application code is supposed to check whether the value is null.
>     ADD COLUMN CreationTime TIMESTAMP OPTIONS(allow_commit_timestamp = true);
> ```

nit: `CreateTime` would match how to named this in other tables. That said, I goofed with `LockExpirationTime` so `CreationTime` is more similar grammatically. Your choice on which way to take this.



_Code quote:_
```TSQL
CreationTime
```

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/902)* status: 6 of 10 files reviewed, 3 unresolved discussions (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/MillBase.kt` line 174 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/902#-NR4Kmdd4KqbNXHoQbdH:-NR5Hj7J80_VRNufAqP2:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e3e24ea3d3f82e058e49fbb1b4bfb8e5bf5d423d/src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/MillBase.kt#L174)):*

This code block appears in multiple places. Extract it to a method?
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/gcloud/spanner/computation/UnclaimedTasksQuery.kt` line 45 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/902#-NR4Ii0HCRzVF5GlsLOV:-NR5I1ToCsZQ47GKBH94:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e3e24ea3d3f82e058e49fbb1b4bfb8e5bf5d423d/src/main/kotlin/org/wfanet/measurement/duchy/deploy/gcloud/spanner/computation/UnclaimedTasksQuery.kt#L45)):*

This clause means that previously created computations never get claimed. Instead, we just want them to be ordered such that they are claimed before ones with a creation time.
</blockquote></details>

Done.
___
*[`src/main/resources/duchy/spanner/add-computation-creation-time.sql` line 24 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/902#-NR4JSyK8v9wsNIA8mVl:-NR569Ss75EQ47LrJJ8R:b-bn2e8q) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e3e24ea3d3f82e058e49fbb1b4bfb8e5bf5d423d/src/main/resources/duchy/spanner/add-computation-creation-time.sql#L24)):*

nit: `CreateTime` would match how to named this in other tables. That said, I goofed with `LockExpirationTime` so `CreationTime` is more similar grammatically. Your choice on which way to take this.
</blockquote></details>

`CreationTime` has been used in duchy's `create-computations-schema.sql`. So I keep it the same way.

SanjayVas (approved):
> Reviewed 4 of 4 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/902)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @renjiezh)

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/902)* status: all files reviewed, 1 unresolved discussion (waiting on @duliomatos1, @Marco-Premier, @renjiezh, and @stevenwarejones)

*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/902#-NRAi2yCAAYazmvg95BR:-NRAi2yCAAYazmvg95BS:b9xsz50) (no related file):*
I noticed another potential issue here: it looks like the `EnqueueComputation` implementation makes some assumptions about LockExpirationTime being used to order the work queue. See https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f36d76ee351e1fb59b64d4003f21730b9b06d38b/src/main/kotlin/org/wfanet/measurement/duchy/deploy/gcloud/spanner/computation/GcpSpannerComputationsDatabaseTransactor.kt#L112

___

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/902)* status: all files reviewed, 1 unresolved discussion (waiting on @duliomatos1, @Marco-Premier, @SanjayVas, and @stevenwarejones)

*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/902#-NRAi2yCAAYazmvg95BR:-NRC6Mb--KWUG5qSxToD:bkwd4qw) (no related file):*

I noticed another potential issue here: it looks like the `EnqueueComputation` implementation makes some assumptions about LockExpirationTime being used to order the work queue. See https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f36d76ee351e1fb59b64d4003f21730b9b06d38b/src/main/kotlin/org/wfanet/measurement/duchy/deploy/gcloud/spanner/computation/GcpSpannerComputationsDatabaseTransactor.kt#L112
</blockquote></details>

Thanks for pointing that out. `enqueueComputation()` is used by mill to handle computations with transient failure. Given we claim tasks by `CreationTime`, it means this certain failed computation will be picked up multiple times in a row. This is not what we like.  
However,  the transient failure are most likely caused by either a race condition or unavailable server. For the latter reason, claiming other task won't solve the abnormal state. In the worst case, MaxAttempts protects the mill from stuck on the same Computation forever. 
So I think `enqueue` won't cause real trouble. Please let me know you thought. Thanks.

___

Marco-Premier (approved):
> Reviewed 4 of 10 files at r1, 2 of 2 files at r2, 4 of 4 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/902)* status: all files reviewed, 1 unresolved discussion (waiting on @duliomatos1, @SanjayVas, and @stevenwarejones)

SanjayVas (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/902)* status: all files reviewed, 1 unresolved discussion (waiting on @duliomatos1, @renjiezh, and @stevenwarejones)

*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/902#-NRAi2yCAAYazmvg95BR:-NREUIb70Gozd88fCV_J:b-4mwbhv) (no related file):*

Thanks for pointing that out. `enqueueComputation()` is used by mill to handle computations with transient failure. Given we claim tasks by `CreationTime`, it means this certain failed computation will be picked up multiple times in a row. This is not what we like.  
However,  the transient failure are most likely caused by either a race condition or unavailable server. For the latter reason, claiming other task won't solve the abnormal state. In the worst case, MaxAttempts protects the mill from stuck on the same Computation forever. 
So I think `enqueue` won't cause real trouble. Please let me know you thought. Thanks.
</blockquote></details>

My concern is whether the `delay_second` field in `EnqueueComputationRequest` would be respected. Is changing LockExpirationTime with no owner sufficient for that? I guess perhaps it is, since `ClaimWork` won't let you claim if there's a lock with an owner that isn't the caller.

___

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/902)* status: all files reviewed, 1 unresolved discussion (waiting on @duliomatos1, @SanjayVas, and @stevenwarejones)

*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/902#-NRAi2yCAAYazmvg95BR:-NREVDFDDyihk__HtyRx:bnjbe3l) (no related file):*

My concern is whether the `delay_second` field in `EnqueueComputationRequest` would be respected. Is changing LockExpirationTime with no owner sufficient for that? I guess perhaps it is, since `ClaimWork` won't let you claim if there's a lock with an owner that isn't the caller.
</blockquote></details>

Yeah, the unclaimedTask query requires expired lock to be claimed. 

___

stevenwarejones (approved):
> Reviewed 4 of 10 files at r1, 2 of 2 files at r2, 2 of 4 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/902)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @duliomatos1)

PR #904, opened 2023-03-21
Create Gaussian Noiser in EDP simulator for direct measurements

iverson52000:
And created a Noiser interface for Laplace and Gaussian Noisers

### DIFF
--- a/src/test/kotlin/org/wfanet/measurement/loadtest/dataprovider/GaussianNoiserTest.kt
+++ b/src/test/kotlin/org/wfanet/measurement/loadtest/dataprovider/GaussianNoiserTest.kt
@@ -16,6 +16,8 @@ package org.wfanet.measurement.loadtest.dataprovider
 
 import com.google.common.truth.Truth.assertThat
 import java.util.Random
+import kotlin.math.pow
+import kotlin.math.sqrt
 import org.junit.Test
 import org.junit.runner.RunWith
 import org.junit.runners.JUnit4
@@ -30,7 +32,7 @@ private const val RANDOM_SEED: Long = 1
 @RunWith(JUnit4::class)
 class GaussianNoiserTest {
   @Test
-  fun `calculate Gaussian noised direct reach and frequency correctly with epsilon equals to 1`() {
+  fun `calculate Gaussian noised direct reach and frequency correctly with seeded random number generator and epsilon equals to 1`() {
     val random = Random(RANDOM_SEED)
     val gaussianNoiser = GaussianNoiser(MEASUREMENT_SPEC.reachAndFrequency, random)
 
@@ -50,7 +52,7 @@ class GaussianNoiserTest {
     }
   }
   @Test
-  fun `calculate Gaussian noised direct reach and frequency correctly with epsilon equals to 1E-4`() {
+  fun `calculate Gaussian noised direct reach and frequency correctly with seeded random number generator and epsilon equals to 1E-4`() {
     val measurementSpec =
       MEASUREMENT_SPEC.copy {
         reachAndFrequency =
@@ -70,6 +72,7 @@ class GaussianNoiserTest {
       )
 
     val expectedNoisedReachValue = 52897
+    // Frequency values can be negative due to large variance
     val expectedNoisedFrequencyMap =
       mapOf(1L to -40.21444386138276, 2L to -73.03106359061003, 3L to -41.71217907726343)
 
@@ -77,6 +80,27 @@ class GaussianNoiserTest {
     noisedFrequencyMap.forEach { (frequency, percentage) ->
       assertThat(percentage).isEqualTo(expectedNoisedFrequencyMap[frequency])
     }
+  }
+  @Test
+  fun `standard deviation from noisedReachValues is close to the theoretical sigma`() {
+    val random = Random(RANDOM_SEED)
+    val gaussianNoiser = GaussianNoiser(MEASUREMENT_SPEC.reachAndFrequency, random)
+    val noisedReachValues =
+      List(1000) {
+        gaussianNoiser
+          .addPublisherNoise(
+            reachValue,
+            frequencyMap,
+          )
+          .reach
+      }
+
+    val sigma = calculateStandardDeviation(noisedReachValues)
+    // Sigma value with pre-set epsilon and delta
+    val expectedSigma = 6.557822067460045
+    val diffRatio = (expectedSigma - sigma) / expectedSigma
+
+    assertThat(diffRatio).isLessThan(0.2)
   }
   companion object {
     private val MEASUREMENT_SPEC = measurementSpec {
@@ -97,7 +121,13 @@ class GaussianNoiserTest {
           width = PrivacyLandscape.PRIVACY_BUCKET_VID_SAMPLE_WIDTH
         }
     }
-    private val reachValue = 500L
+    private const val reachValue = 500L
     private val frequencyMap = mapOf(1L to 0.6, 2L to 0.2, 3L to 0.2)
+    private fun calculateStandardDeviation(nums: List<Long>): Double {
+      val mean = nums.average()
+      val standardDeviation = nums.fold(0.0) { acc, num -> acc + (num - mean).pow(2.0) }
+
+      return sqrt(standardDeviation / nums.size)
+    }
   }
 }
--- a/src/test/kotlin/org/wfanet/measurement/loadtest/dataprovider/GaussianNoiserTest.kt
+++ b/src/test/kotlin/org/wfanet/measurement/loadtest/dataprovider/GaussianNoiserTest.kt
@@ -23,80 +23,38 @@ import org.junit.Test
 import org.junit.runner.RunWith
 import org.junit.runners.JUnit4
 import org.wfanet.measurement.api.v2alpha.MeasurementSpecKt
-import org.wfanet.measurement.api.v2alpha.copy
 import org.wfanet.measurement.api.v2alpha.differentialPrivacyParams
 import org.wfanet.measurement.api.v2alpha.measurementSpec
-import org.wfanet.measurement.eventdataprovider.privacybudgetmanagement.PrivacyLandscape
 
 private const val RANDOM_SEED: Long = 1
 
 @RunWith(JUnit4::class)
 class GaussianNoiserTest {
   @Test
-  fun `calculate Gaussian noised direct reach and frequency correctly with seeded random number generator and epsilon equals to 1`() {
+  fun `Gaussian noiser with random seed returns expected samples`() {
     val random = Random(RANDOM_SEED)
-    val gaussianNoiser = GaussianNoiser(MEASUREMENT_SPEC.reachAndFrequency, random)
-
-    val (noisedReachValue, noisedFrequencyMap) =
-      gaussianNoiser.addReachAndFrequencyPublisherNoise(
-        reachValue,
-        frequencyMap,
+    val gaussianNoiser =
+      GaussianNoiser(MEASUREMENT_SPEC.reachAndFrequency.reachPrivacyParams, random)
+    val samples = List(5) { gaussianNoiser.sample() }
+    val expectedSamples =
+      listOf(
+        10.240550327678935,
+        -3.9883454237630307,
+        -7.156064121340545,
+        -4.0956149786203895,
+        -7.333487792497553
       )
 
-    val expectedNoisedReachValue = 510
-    val expectedNoisedFrequencyMap =
-      mapOf(1L to 0.592023309152474, 2L to 0.1856878717573189, 3L to 0.19180877004275923)
-
-    assertThat(noisedReachValue).isEqualTo(expectedNoisedReachValue)
-    noisedFrequencyMap.forEach { (frequency, percentage) ->
-      assertThat(percentage).isEqualTo(expectedNoisedFrequencyMap[frequency])
-    }
+    assertThat(expectedSamples).isEqualTo(samples)
   }
   @Test
-  fun `calculate Gaussian noised direct reach and frequency correctly with seeded random number generator and epsilon equals to 1E-4`() {
-    val measurementSpec =
-      MEASUREMENT_SPEC.copy {
-        reachAndFrequency =
-          reachAndFrequency.copy {
-            reachPrivacyParams = reachPrivacyParams.copy { epsilon = 1E-4 }
-            frequencyPrivacyParams = frequencyPrivacyParams.copy { epsilon = 1E-4 }
-          }
-      }
-
+  fun `standard deviation from samples is close to the theoretical sigma`() {
     val random = Random(RANDOM_SEED)
-    val gaussianNoiser = GaussianNoiser(measurementSpec.reachAndFrequency, random)
+    val gaussianNoiser =
+      GaussianNoiser(MEASUREMENT_SPEC.reachAndFrequency.reachPrivacyParams, random)
+    val samples = List(1000) { gaussianNoiser.sample() }
 
-    val (noisedReachValue, noisedFrequencyMap) =
-      gaussianNoiser.addReachAndFrequencyPublisherNoise(
-        reachValue,
-        frequencyMap,
-      )
-
-    val expectedNoisedReachValue = 78766
-    // Frequency values can be negative due to large variance
-    val expectedNoisedFrequencyMap =
-      mapOf(1L to -60.364636044457335, 2L to -109.18542135016607, 3L to -62.40432585957863)
-
-    assertThat(noisedReachValue).isEqualTo(expectedNoisedReachValue)
-    noisedFrequencyMap.forEach { (frequency, percentage) ->
-      assertThat(percentage).isEqualTo(expectedNoisedFrequencyMap[frequency])
-    }
-  }
-  @Test
-  fun `standard deviation from noisedReachValues is close to the theoretical sigma`() {
-    val random = Random(RANDOM_SEED)
-    val gaussianNoiser = GaussianNoiser(MEASUREMENT_SPEC.reachAndFrequency, random)
-    val noisedReachValues =
-      List(1000) {
-        gaussianNoiser
-          .addReachAndFrequencyPublisherNoise(
-            reachValue,
-            frequencyMap,
-          )
-          .reach
-      }
-
-    val sigma = calculateStandardDeviation(noisedReachValues)
+    val sigma = calculateStandardDeviation(samples)
     // Sigma value with pre-set epsilon and delta
     val expectedSigma = 6.557822067460045
     val diffRatio = abs((expectedSigma - sigma) / expectedSigma)
@@ -111,20 +69,9 @@ class GaussianNoiserTest {
             epsilon = 1.0
             delta = 1E-12
           }
-          frequencyPrivacyParams = differentialPrivacyParams {
-            epsilon = 1.0
-            delta = 1E-12
-          }
-        }
-      vidSamplingInterval =
-        MeasurementSpecKt.vidSamplingInterval {
-          start = 0.0f
-          width = PrivacyLandscape.PRIVACY_BUCKET_VID_SAMPLE_WIDTH
         }
     }
-    private const val reachValue = 500L
-    private val frequencyMap = mapOf(1L to 0.6, 2L to 0.2, 3L to 0.2)
-    private fun calculateStandardDeviation(nums: List<Long>): Double {
+    private fun calculateStandardDeviation(nums: List<Double>): Double {
       val mean = nums.average()
       val standardDeviation = nums.fold(0.0) { acc, num -> acc + (num - mean).pow(2.0) }
 
--- a/src/test/kotlin/org/wfanet/measurement/loadtest/dataprovider/GaussianNoiserTest.kt
+++ b/src/test/kotlin/org/wfanet/measurement/loadtest/dataprovider/GaussianNoiserTest.kt
@@ -15,6 +15,7 @@
 package org.wfanet.measurement.loadtest.dataprovider
 
 import com.google.common.truth.Truth.assertThat
+import java.util.Random
 import org.junit.Test
 import org.junit.runner.RunWith
 import org.junit.runners.JUnit4
@@ -32,6 +33,9 @@ class GaussianNoiserTest {
     val reachValue = 500L
     val frequencyMap = mapOf(1L to 0.6, 2L to 0.2, 3L to 0.2)
 
+    val random = Random(RANDOM_SEED)
+    val gaussianNoiser = GaussianNoiser(MEASUREMENT_SPEC.reachAndFrequency, random)
+
     val (noisedReachValue, noisedFrequencyMap) =
       gaussianNoiser.addPublisherNoise(
         reachValue,
@@ -40,7 +44,7 @@ class GaussianNoiserTest {
 
     val expectedNoisedReachValue = 500
     val expectedNoisedFrequencyMap =
-      mapOf(1L to 0.6000013282959348, 2L to 0.20000065359491367, 3L to 0.20000174523507977)
+      mapOf(1L to 0.5999993912234646, 2L to 0.1999989077064648, 3L to 0.19999937484996114)
 
     assertThat(noisedReachValue).isEqualTo(expectedNoisedReachValue)
     noisedFrequencyMap.forEach { (frequency, percentage) ->
@@ -67,7 +71,5 @@ class GaussianNoiserTest {
           width = PrivacyLandscape.PRIVACY_BUCKET_VID_SAMPLE_WIDTH
         }
     }
-
-    val gaussianNoiser = GaussianNoiser(MEASUREMENT_SPEC.reachAndFrequency, RANDOM_SEED)
   }
 }
--- a/src/test/kotlin/org/wfanet/measurement/loadtest/dataprovider/GaussianNoiserTest.kt
+++ b/src/test/kotlin/org/wfanet/measurement/loadtest/dataprovider/GaussianNoiserTest.kt
@@ -40,7 +40,7 @@ class GaussianNoiserTest {
 
     val expectedNoisedReachValue = 500
     val expectedNoisedFrequencyMap =
-      mapOf(1L to 0.6017729973990973, 2L to 0.20087241257893376, 3L to 0.20232952399863266)
+      mapOf(1L to 0.6000013282959348, 2L to 0.20000065359491367, 3L to 0.20000174523507977)
 
     assertThat(noisedReachValue).isEqualTo(expectedNoisedReachValue)
     noisedFrequencyMap.forEach { (frequency, percentage) ->
--- a/src/test/kotlin/org/wfanet/measurement/loadtest/dataprovider/GaussianNoiserTest.kt
+++ b/src/test/kotlin/org/wfanet/measurement/loadtest/dataprovider/GaussianNoiserTest.kt
@@ -20,6 +20,7 @@ import org.junit.Test
 import org.junit.runner.RunWith
 import org.junit.runners.JUnit4
 import org.wfanet.measurement.api.v2alpha.MeasurementSpecKt
+import org.wfanet.measurement.api.v2alpha.copy
 import org.wfanet.measurement.api.v2alpha.differentialPrivacyParams
 import org.wfanet.measurement.api.v2alpha.measurementSpec
 import org.wfanet.measurement.eventdataprovider.privacybudgetmanagement.PrivacyLandscape
@@ -29,10 +30,7 @@ private const val RANDOM_SEED: Long = 1
 @RunWith(JUnit4::class)
 class GaussianNoiserTest {
   @Test
-  fun `calculate Gaussian noised direct reach and frequency correctly`() {
-    val reachValue = 500L
-    val frequencyMap = mapOf(1L to 0.6, 2L to 0.2, 3L to 0.2)
-
+  fun `calculate Gaussian noised direct reach and frequency correctly with epsilon equals to 1`() {
     val random = Random(RANDOM_SEED)
     val gaussianNoiser = GaussianNoiser(MEASUREMENT_SPEC.reachAndFrequency, random)
 
@@ -42,27 +40,55 @@ class GaussianNoiserTest {
         frequencyMap,
       )
 
-    val expectedNoisedReachValue = 500
+    val expectedNoisedReachValue = 512
     val expectedNoisedFrequencyMap =
-      mapOf(1L to 0.5999993912234646, 2L to 0.1999989077064648, 3L to 0.19999937484996114)
+      mapOf(1L to 0.590035536760728, 2L to 0.18212132343148613, 3L to 0.18976753501501636)
 
     assertThat(noisedReachValue).isEqualTo(expectedNoisedReachValue)
     noisedFrequencyMap.forEach { (frequency, percentage) ->
       assertThat(percentage).isEqualTo(expectedNoisedFrequencyMap[frequency])
     }
   }
+  @Test
+  fun `calculate Gaussian noised direct reach and frequency correctly with epsilon equals to 1E-4`() {
+    val measurementSpec =
+      MEASUREMENT_SPEC.copy {
+        reachAndFrequency =
+          reachAndFrequency.copy {
+            reachPrivacyParams = reachPrivacyParams.copy { epsilon = 1E-4 }
+            frequencyPrivacyParams = frequencyPrivacyParams.copy { epsilon = 1E-4 }
+          }
+      }
 
+    val random = Random(RANDOM_SEED)
+    val gaussianNoiser = GaussianNoiser(measurementSpec.reachAndFrequency, random)
+
+    val (noisedReachValue, noisedFrequencyMap) =
+      gaussianNoiser.addPublisherNoise(
+        reachValue,
+        frequencyMap,
+      )
+
+    val expectedNoisedReachValue = 52897
+    val expectedNoisedFrequencyMap =
+      mapOf(1L to -40.21444386138276, 2L to -73.03106359061003, 3L to -41.71217907726343)
+
+    assertThat(noisedReachValue).isEqualTo(expectedNoisedReachValue)
+    noisedFrequencyMap.forEach { (frequency, percentage) ->
+      assertThat(percentage).isEqualTo(expectedNoisedFrequencyMap[frequency])
+    }
+  }
   companion object {
     private val MEASUREMENT_SPEC = measurementSpec {
       reachAndFrequency =
         MeasurementSpecKt.reachAndFrequency {
           reachPrivacyParams = differentialPrivacyParams {
             epsilon = 1.0
-            delta = 1.0
+            delta = 1E-12
           }
           frequencyPrivacyParams = differentialPrivacyParams {
             epsilon = 1.0
-            delta = 1.0
+            delta = 1E-12
           }
         }
       vidSamplingInterval =
@@ -71,5 +97,7 @@ class GaussianNoiserTest {
           width = PrivacyLandscape.PRIVACY_BUCKET_VID_SAMPLE_WIDTH
         }
     }
+    private val reachValue = 500L
+    private val frequencyMap = mapOf(1L to 0.6, 2L to 0.2, 3L to 0.2)
   }
 }
--- a/src/test/kotlin/org/wfanet/measurement/loadtest/dataprovider/GaussianNoiserTest.kt
+++ b/src/test/kotlin/org/wfanet/measurement/loadtest/dataprovider/GaussianNoiserTest.kt
@@ -1,0 +1,73 @@
+// Copyright 2023 The Cross-Media Measurement Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package org.wfanet.measurement.loadtest.dataprovider
+
+import com.google.common.truth.Truth.assertThat
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.junit.runners.JUnit4
+import org.wfanet.measurement.api.v2alpha.MeasurementSpecKt
+import org.wfanet.measurement.api.v2alpha.differentialPrivacyParams
+import org.wfanet.measurement.api.v2alpha.measurementSpec
+import org.wfanet.measurement.eventdataprovider.privacybudgetmanagement.PrivacyLandscape
+
+private const val RANDOM_SEED: Long = 1
+
+@RunWith(JUnit4::class)
+class GaussianNoiserTest {
+  @Test
+  fun `calculate Gaussian noised direct reach and frequency correctly`() {
+    val reachValue = 500L
+    val frequencyMap = mapOf(1L to 0.6, 2L to 0.2, 3L to 0.2)
+
+    val (noisedReachValue, noisedFrequencyMap) =
+      gaussianNoiser.addPublisherNoise(
+        reachValue,
+        frequencyMap,
+      )
+
+    val expectedNoisedReachValue = 500
+    val expectedNoisedFrequencyMap =
+      mapOf(1L to 0.6017729973990973, 2L to 0.20087241257893376, 3L to 0.20232952399863266)
+
+    assertThat(noisedReachValue).isEqualTo(expectedNoisedReachValue)
+    noisedFrequencyMap.forEach { (frequency, percentage) ->
+      assertThat(percentage).isEqualTo(expectedNoisedFrequencyMap[frequency])
+    }
+  }
+
+  companion object {
+    private val MEASUREMENT_SPEC = measurementSpec {
+      reachAndFrequency =
+        MeasurementSpecKt.reachAndFrequency {
+          reachPrivacyParams = differentialPrivacyParams {
+            epsilon = 1.0
+            delta = 1.0
+          }
+          frequencyPrivacyParams = differentialPrivacyParams {
+            epsilon = 1.0
+            delta = 1.0
+          }
+        }
+      vidSamplingInterval =
+        MeasurementSpecKt.vidSamplingInterval {
+          start = 0.0f
+          width = PrivacyLandscape.PRIVACY_BUCKET_VID_SAMPLE_WIDTH
+        }
+    }
+
+    val gaussianNoiser = GaussianNoiser(MEASUREMENT_SPEC.reachAndFrequency, RANDOM_SEED)
+  }
+}
--- a/src/test/kotlin/org/wfanet/measurement/loadtest/dataprovider/GaussianNoiserTest.kt
+++ b/src/test/kotlin/org/wfanet/measurement/loadtest/dataprovider/GaussianNoiserTest.kt
@@ -16,6 +16,7 @@ package org.wfanet.measurement.loadtest.dataprovider
 
 import com.google.common.truth.Truth.assertThat
 import java.util.Random
+import kotlin.math.abs
 import kotlin.math.pow
 import kotlin.math.sqrt
 import org.junit.Test
@@ -37,14 +38,14 @@ class GaussianNoiserTest {
     val gaussianNoiser = GaussianNoiser(MEASUREMENT_SPEC.reachAndFrequency, random)
 
     val (noisedReachValue, noisedFrequencyMap) =
-      gaussianNoiser.addPublisherNoise(
+      gaussianNoiser.addReachAndFrequencyPublisherNoise(
         reachValue,
         frequencyMap,
       )
 
-    val expectedNoisedReachValue = 512
+    val expectedNoisedReachValue = 510
     val expectedNoisedFrequencyMap =
-      mapOf(1L to 0.590035536760728, 2L to 0.18212132343148613, 3L to 0.18976753501501636)
+      mapOf(1L to 0.592023309152474, 2L to 0.1856878717573189, 3L to 0.19180877004275923)
 
     assertThat(noisedReachValue).isEqualTo(expectedNoisedReachValue)
     noisedFrequencyMap.forEach { (frequency, percentage) ->
@@ -66,15 +67,15 @@ class GaussianNoiserTest {
     val gaussianNoiser = GaussianNoiser(measurementSpec.reachAndFrequency, random)
 
     val (noisedReachValue, noisedFrequencyMap) =
-      gaussianNoiser.addPublisherNoise(
+      gaussianNoiser.addReachAndFrequencyPublisherNoise(
         reachValue,
         frequencyMap,
       )
 
-    val expectedNoisedReachValue = 52897
+    val expectedNoisedReachValue = 78766
     // Frequency values can be negative due to large variance
     val expectedNoisedFrequencyMap =
-      mapOf(1L to -40.21444386138276, 2L to -73.03106359061003, 3L to -41.71217907726343)
+      mapOf(1L to -60.364636044457335, 2L to -109.18542135016607, 3L to -62.40432585957863)
 
     assertThat(noisedReachValue).isEqualTo(expectedNoisedReachValue)
     noisedFrequencyMap.forEach { (frequency, percentage) ->
@@ -88,7 +89,7 @@ class GaussianNoiserTest {
     val noisedReachValues =
       List(1000) {
         gaussianNoiser
-          .addPublisherNoise(
+          .addReachAndFrequencyPublisherNoise(
             reachValue,
             frequencyMap,
           )
@@ -98,7 +99,7 @@ class GaussianNoiserTest {
     val sigma = calculateStandardDeviation(noisedReachValues)
     // Sigma value with pre-set epsilon and delta
     val expectedSigma = 6.557822067460045
-    val diffRatio = (expectedSigma - sigma) / expectedSigma
+    val diffRatio = abs((expectedSigma - sigma) / expectedSigma)
 
     assertThat(diffRatio).isLessThan(0.2)
   }
--- a/src/main/kotlin/org/wfanet/measurement/integration/common/InProcessEdpSimulator.kt
+++ b/src/main/kotlin/org/wfanet/measurement/integration/common/InProcessEdpSimulator.kt
@@ -100,7 +100,8 @@ class InProcessEdpSimulator(
           100.0f
         ),
       trustedCertificates = trustedCertificates,
-      random
+      random,
+      NOISE_MECHANISM
     )
 
   private lateinit var edpJob: Job
@@ -128,5 +129,6 @@ class InProcessEdpSimulator(
     private val logger: Logger = Logger.getLogger(this::class.java.name)
     private const val RANDOM_SEED: Long = 1
     private val random = Random(RANDOM_SEED)
+    private const val NOISE_MECHANISM = "LAPLACE"
   }
 }
--- a/src/main/kotlin/org/wfanet/measurement/integration/common/InProcessEdpSimulator.kt
+++ b/src/main/kotlin/org/wfanet/measurement/integration/common/InProcessEdpSimulator.kt
@@ -19,6 +19,7 @@ import io.grpc.Channel
 import java.security.cert.X509Certificate
 import java.time.Clock
 import java.time.Duration
+import java.util.Random
 import java.util.logging.Level
 import java.util.logging.Logger
 import kotlin.coroutines.CoroutineContext
@@ -98,7 +99,8 @@ class InProcessEdpSimulator(
           100.0f,
           100.0f
         ),
-      trustedCertificates = trustedCertificates
+      trustedCertificates = trustedCertificates,
+      random
     )
 
   private lateinit var edpJob: Job
@@ -124,5 +126,7 @@ class InProcessEdpSimulator(
 
   companion object {
     private val logger: Logger = Logger.getLogger(this::class.java.name)
+    private const val randomSeed: Long = 1
+    private val random = Random(randomSeed)
   }
 }
--- a/src/main/kotlin/org/wfanet/measurement/integration/common/InProcessEdpSimulator.kt
+++ b/src/main/kotlin/org/wfanet/measurement/integration/common/InProcessEdpSimulator.kt
@@ -126,7 +126,7 @@ class InProcessEdpSimulator(
 
   companion object {
     private val logger: Logger = Logger.getLogger(this::class.java.name)
-    private const val randomSeed: Long = 1
-    private val random = Random(randomSeed)
+    private const val RANDOM_SEED: Long = 1
+    private val random = Random(RANDOM_SEED)
   }
 }
--- a/src/main/kotlin/org/wfanet/measurement/integration/common/InProcessEdpSimulator.kt
+++ b/src/main/kotlin/org/wfanet/measurement/integration/common/InProcessEdpSimulator.kt
@@ -44,6 +44,7 @@ import org.wfanet.measurement.eventdataprovider.privacybudgetmanagement.PrivacyB
 import org.wfanet.measurement.eventdataprovider.privacybudgetmanagement.testing.TestPrivacyBucketMapper
 import org.wfanet.measurement.loadtest.dataprovider.EdpData
 import org.wfanet.measurement.loadtest.dataprovider.EdpSimulator
+import org.wfanet.measurement.loadtest.dataprovider.NoiseMechanism
 import org.wfanet.measurement.loadtest.dataprovider.RandomEventQuery
 import org.wfanet.measurement.loadtest.dataprovider.SketchGenerationParams
 import org.wfanet.measurement.loadtest.storage.SketchStore
@@ -129,6 +130,6 @@ class InProcessEdpSimulator(
     private val logger: Logger = Logger.getLogger(this::class.java.name)
     private const val RANDOM_SEED: Long = 1
     private val random = Random(RANDOM_SEED)
-    private const val NOISE_MECHANISM = "LAPLACE"
+    private val NOISE_MECHANISM = NoiseMechanism.LAPLACE
   }
 }
--- a/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/AbstractNoiser.kt
+++ b/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/AbstractNoiser.kt
@@ -14,11 +14,13 @@
 
 package org.wfanet.measurement.loadtest.dataprovider
 
-/** The reach and frequency estimation of a computation. */
-data class ReachAndFrequencyPair(val reach: Long, val frequency: Map<Long, Double>)
+import org.apache.commons.math3.distribution.RealDistribution
 
 /** A base noiser class for direct measurement */
-interface Noiser {
+abstract class AbstractNoiser : Noiser {
+  protected abstract val distributionForReach: RealDistribution
+  protected abstract val distributionForFrequency: RealDistribution
+
   /**
    * Add publisher noise to calculated direct reach and frequency.
    *
@@ -26,8 +28,17 @@ interface Noiser {
    * @param frequencyMap Direct frequency.
    * @return Pair of noised reach value and frequency map.
    */
-  fun addPublisherNoise(
+  override fun addPublisherNoise(
     reachValue: Long,
     frequencyMap: Map<Long, Double>,
-  ): ReachAndFrequencyPair
+  ): ReachAndFrequencyPair {
+    val noisedReachValue = reachValue + distributionForReach.sample().toInt()
+    val noisedFrequencyMap =
+      frequencyMap.mapValues {
+        (it.value * reachValue.toDouble() + distributionForFrequency.sample()) /
+          reachValue.toDouble()
+      }
+
+    return ReachAndFrequencyPair(noisedReachValue, noisedFrequencyMap)
+  }
 }
--- a/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/AbstractNoiser.kt
+++ b/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/AbstractNoiser.kt
@@ -16,22 +16,11 @@ package org.wfanet.measurement.loadtest.dataprovider
 
 import org.apache.commons.math3.distribution.RealDistribution
 
-/** A base noiser class for direct measurement */
+/** A base noiser class for direct measurements */
 abstract class AbstractNoiser : Noiser {
-  protected abstract val distributionForReach: RealDistribution
-  protected abstract val distributionForFrequency: RealDistribution
+  protected abstract val distribution: RealDistribution
 
-  override fun addReachAndFrequencyPublisherNoise(
-    reachValue: Long,
-    frequencyMap: Map<Long, Double>,
-  ): ReachAndFrequencyPair {
-    val noisedReachValue = reachValue + distributionForReach.sample().toInt()
-    val noisedFrequencyMap =
-      frequencyMap.mapValues {
-        (it.value * reachValue.toDouble() + distributionForFrequency.sample()) /
-          reachValue.toDouble()
-      }
-
-    return ReachAndFrequencyPair(noisedReachValue, noisedFrequencyMap)
+  override fun sample(): Double {
+    return distribution.sample()
   }
 }
--- a/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/AbstractNoiser.kt
+++ b/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/AbstractNoiser.kt
@@ -21,7 +21,7 @@ abstract class AbstractNoiser : Noiser {
   protected abstract val distributionForReach: RealDistribution
   protected abstract val distributionForFrequency: RealDistribution
 
-  override fun addPublisherNoise(
+  override fun addReachAndFrequencyPublisherNoise(
     reachValue: Long,
     frequencyMap: Map<Long, Double>,
   ): ReachAndFrequencyPair {
--- a/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/AbstractNoiser.kt
+++ b/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/AbstractNoiser.kt
@@ -21,13 +21,6 @@ abstract class AbstractNoiser : Noiser {
   protected abstract val distributionForReach: RealDistribution
   protected abstract val distributionForFrequency: RealDistribution
 
-  /**
-   * Add publisher noise to calculated direct reach and frequency.
-   *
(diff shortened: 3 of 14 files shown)


SanjayVas (changes requested):
> Reviewed 3 of 8 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/904)* status: 3 of 8 files reviewed, 4 unresolved discussions (waiting on @iverson52000, @jiayu-google, @kungfucraig, and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt` line 730 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/904#-NR5F2pe6kUOrx6LL7tC:-NR5F2pe6kUOrx6LL7tD:b-vtg7hc) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4be505c2e8c0aa784f23fd48da6a9fd0bde0f4c9/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt#L730)):*
> ```Kotlin
> 
>     logger.info("Adding publisher noise to direct reach and frequency...")
>     // Reseed random number generator so the results can be verified in unit test and integration
> ```

Don't do this. Instead take in a java.util.Random in the EdpSimulator constructor. This way the caller has full flexibility on the implementation, including passing in one with a known seed.



_Code quote:_
```Kotlin
    // Reseed random number generator so the results can be verified in unit test and integration
    // tests.
```
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/LaplaceNoiser.kt` line 20 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/904#-NR5EdCX8X_FuLQmWFIL:-NR5EdCYC6M7RwngGaE-:b-pao1uh) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4be505c2e8c0aa784f23fd48da6a9fd0bde0f4c9/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/LaplaceNoiser.kt#L20)):*
> ```Kotlin
> import org.wfanet.measurement.api.v2alpha.MeasurementSpec
> 
> class LaplaceNoiser(reachAndFrequency: MeasurementSpec.ReachAndFrequency, randomSeed: Long?) :
> ```

Take in a`java.util.Random` instead of passing the seed. You can use the appropriate `LaplaceDistribution` constructor overload that takes in a `org.apache.commons.math3.random.RandomGenerator`. You can use `org.apache.commons.math3.random.RandomAdaptor` to wrap the `Random`.



_Code quote:_
```Kotlin
randomSeed: Long?
```
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/Noiser.kt` line 23 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/904#-NR5GEb-8kSJniJsmL1w:-NR5GEb-8kSJniJsmL1x:b-jzf90g) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4be505c2e8c0aa784f23fd48da6a9fd0bde0f4c9/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/Noiser.kt#L23)):*
> ```Kotlin
> 
> /** A base noiser class for direct measurement */
> interface Noiser {
> ```

Add an abstract base class so that you can make the `RealDistribution` properties `protected` and avoid leaking that the implementation uses Apache Commons Math. (I'm assuming that callers need not read the distributions).

e.g. have `interface Noiser` that just has `addPublisherNoise` declared, and then an `internal abstract class AbstractNoiser` with the properties and the implementation of that method.



_Code quote:_
```Kotlin
interface Noiser
```
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/Noiser.kt` line 24 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/904#-NR5H4IO2SNteCyL9aLO:-NR5H4IO2SNteCyL9aLP:blare0) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4be505c2e8c0aa784f23fd48da6a9fd0bde0f4c9/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/Noiser.kt#L24)):*
> ```Kotlin
> /** A base noiser class for direct measurement */
> interface Noiser {
>   val distributionForReach: AbstractRealDistribution
> ```

nit: Reference the interface rather than the abstract base implementation.



_Suggestion:_
```Kotlin
RealDistribution
```

stevenwarejones (changes requested):
> Reviewed 4 of 8 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/904)* status: 5 of 8 files reviewed, 7 unresolved discussions (waiting on @iverson52000, @jiayu-google, and @uakyol)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/LaplaceNoiser.kt` line 28 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/904#-NRKAZ7P9LZx68b4M-JP:-NRKAZ7P9LZx68b4M-JQ:bpsqjgb) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4be505c2e8c0aa784f23fd48da6a9fd0bde0f4c9/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/LaplaceNoiser.kt#L28)):*
> ```Kotlin
> 
>   init {
>     if (randomSeed != null) {
> ```

What is the default behavior if we don't reseed?
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/Noiser.kt` line 39 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/904#-NRKCqLX84p2XHT01Xdq:-NRKCqLX84p2XHT01Xdr:b-af5fuf) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4be505c2e8c0aa784f23fd48da6a9fd0bde0f4c9/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/Noiser.kt#L39)):*
> ```Kotlin
>   ): ReachAndFrequencyPair {
>     val noisedReachValue = reachValue + distributionForReach.sample().toInt()
>     val noisedFrequencyMap = mutableMapOf<Long, Double>()
> ```

how about avoiding a mutableMap and just using .mapValues?
```
val noisedFrequencyMap = frequencyMap.mapValues{
           (it.value * reachValue.toDouble() + distributionForFrequency.sample()) /
          reachValue.toDouble()

}
```

___
*[`src/test/kotlin/org/wfanet/measurement/loadtest/dataprovider/GaussianNoiserTest.kt` line 55 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/904#-NRKDfdP3DjKi7H0olp5:-NRKDfdP3DjKi7H0olp6:bvkjvob) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4be505c2e8c0aa784f23fd48da6a9fd0bde0f4c9/src/test/kotlin/org/wfanet/measurement/loadtest/dataprovider/GaussianNoiserTest.kt#L55)):*
> ```Kotlin
>       reachAndFrequency =
>         MeasurementSpecKt.reachAndFrequency {
>           reachPrivacyParams = differentialPrivacyParams {
> ```

don't we need to add noiserType as a field to `differentialPrivacyParams`?

src/test/kotlin/org/wfanet/measurement/loadtest/dataprovider/GaussianNoiserTest.kt
```diff
@@ -0,0 +1,75 @@
...
+    }
+  }
+
+  companion object {
+    private val MEASUREMENT_SPEC = measurementSpec {
+      reachAndFrequency =
+        MeasurementSpecKt.reachAndFrequency {
+          reachPrivacyParams = differentialPrivacyParams {
+            epsilon = 1.0
+            delta = 1.0
```
jiayu-google, line 61: The noise is too small when setting delta = 1.  delta = 1e-12 is realistic.

src/test/kotlin/org/wfanet/measurement/loadtest/dataprovider/GaussianNoiserTest.kt
```diff
@@ -0,0 +1,75 @@
...
+          }
+          frequencyPrivacyParams = differentialPrivacyParams {
+            epsilon = 1.0
+            delta = 1.0
```
jiayu-google, line 65: ditto

iverson52000 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/904)* status: 1 of 13 files reviewed, 6 unresolved discussions (waiting on @jiayu-google, @kungfucraig, @SanjayVas, @stevenwarejones, and @uakyol)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt` line 730 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/904#-NR5F2pe6kUOrx6LL7tC:-NRZsBsvAIM0C-2YqOy4:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4be505c2e8c0aa784f23fd48da6a9fd0bde0f4c9/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt#L730)):*

Don't do this. Instead take in a java.util.Random in the EdpSimulator constructor. This way the caller has full flexibility on the implementation, including passing in one with a known seed.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/LaplaceNoiser.kt` line 20 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/904#-NR5EdCX8X_FuLQmWFIL:-NRZsEDKEshz3R8Bn4U3:bdbgn2a) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4be505c2e8c0aa784f23fd48da6a9fd0bde0f4c9/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/LaplaceNoiser.kt#L20)):*

Take in a`java.util.Random` instead of passing the seed. You can use the appropriate `LaplaceDistribution` constructor overload that takes in a `org.apache.commons.math3.random.RandomGenerator`. You can use `org.apache.commons.math3.random.RandomAdaptor` to wrap the `Random`.
</blockquote></details>

Done. Used `RandomGeneratorFactory.createRandomGenerator(random)` to wrap `Random`
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/LaplaceNoiser.kt` line 28 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/904#-NRKAZ7P9LZx68b4M-JP:-NRZsR3VDsEeRh7khWQf:badlb6u) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4be505c2e8c0aa784f23fd48da6a9fd0bde0f4c9/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/LaplaceNoiser.kt#L28)):*

What is the default behavior if we don't reseed?
</blockquote></details>

Updated the classes based on Sanjay's comments. Now the default behavior is `java.util.Random` is passed in without pre-set seed
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/Noiser.kt` line 23 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/904#-NR5GEb-8kSJniJsmL1w:-NRZsptU8LcLBV55_Brn:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4be505c2e8c0aa784f23fd48da6a9fd0bde0f4c9/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/Noiser.kt#L23)):*

Add an abstract base class so that you can make the `RealDistribution` properties `protected` and avoid leaking that the implementation uses Apache Commons Math. (I'm assuming that callers need not read the distributions).

e.g. have `interface Noiser` that just has `addPublisherNoise` declared, and then an `internal abstract class AbstractNoiser` with the properties and the implementation of that method.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/Noiser.kt` line 39 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/904#-NRKCqLX84p2XHT01Xdq:-NRZssZ48kL6FOe1h3Fo:b-5b7t16) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4be505c2e8c0aa784f23fd48da6a9fd0bde0f4c9/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/Noiser.kt#L39)):*

how about avoiding a mutableMap and just using .mapValues?
```
val noisedFrequencyMap = frequencyMap.mapValues{
           (it.value * reachValue.toDouble() + distributionForFrequency.sample()) /
          reachValue.toDouble()

}
```

</blockquote></details>

Done. Good suggestion thanks!
___
*[`src/test/kotlin/org/wfanet/measurement/loadtest/dataprovider/GaussianNoiserTest.kt` line 55 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/904#-NRKDfdP3DjKi7H0olp5:-NRZt5XQ8CKIt4ouBelC:b1pdc7j) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4be505c2e8c0aa784f23fd48da6a9fd0bde0f4c9/src/test/kotlin/org/wfanet/measurement/loadtest/dataprovider/GaussianNoiserTest.kt#L55)):*

don't we need to add noiserType as a field to `differentialPrivacyParams`?
</blockquote></details>

Yeah maybe we should have a noiserType field in `Direct` protocol in `ProtocolConfig`.

src/test/kotlin/org/wfanet/measurement/loadtest/dataprovider/GaussianNoiserTest.kt
```diff
@@ -0,0 +1,75 @@
...
+
+    val (noisedReachValue, noisedFrequencyMap) =
+      gaussianNoiser.addPublisherNoise(
+        reachValue,
+        frequencyMap,
+      )
+
+    val expectedNoisedReachValue = 500
+    val expectedNoisedFrequencyMap =
+      mapOf(1L to 0.5999993912234646, 2L to 0.1999989077064648, 3L to 0.19999937484996114)
```
jiayu-google, line 47: This is a single sample of the noise. Ideally we want to have, say, 1000 samples and statistically test if the empirical variance is close to the theoretical variance. If you'd like to do it, we can discuss how to do the statistical test.

src/test/kotlin/org/wfanet/measurement/loadtest/dataprovider/GaussianNoiserTest.kt
```diff
@@ -0,0 +1,75 @@
...
+      assertThat(percentage).isEqualTo(expectedNoisedFrequencyMap[frequency])
+    }
+  }
+
+  companion object {
+    private val MEASUREMENT_SPEC = measurementSpec {
+      reachAndFrequency =
+        MeasurementSpecKt.reachAndFrequency {
+          reachPrivacyParams = differentialPrivacyParams {
+            epsilon = 1.0
```
jiayu-google, line 60: To exclude the possibility of overflow, it's great to test the maximum and minimum epsilon to use in practice. max epsilon = 1 as it is here. min epsilon = 0.0001.

jiayu-google (commented):
Thanks Albert!

`GaussianNoiser.kt` lgtm. I suggesting making its test more realistic.

stevenwarejones (changes requested):
> Reviewed 11 of 12 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/904)* status: 12 of 13 files reviewed, 9 unresolved discussions (waiting on @iverson52000, @SanjayVas, and @uakyol)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt` line 733 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/904#-NRcoqRX4wao5vntr2eL:-NRcoqRX4wao5vntr2eM:b-rjacep) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/95f822802c42ab00acd373eec9395831b551ec94/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt#L733)):*
> ```Kotlin
>     logger.info("Adding publisher noise to direct reach and frequency...")
> 
>     val laplaceNoiser = LaplaceNoiser(measurementSpec.reachAndFrequency, random)
> ```

add a TODO to use the noiser specified in the measurementSpec
___
*[`src/test/kotlin/org/wfanet/measurement/loadtest/dataprovider/GaussianNoiserTest.kt` line 55 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/904#-NRKDfdP3DjKi7H0olp5:-NRco5DZ7yQZvrdYy7Z4:bq6r9qi) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4be505c2e8c0aa784f23fd48da6a9fd0bde0f4c9/src/test/kotlin/org/wfanet/measurement/loadtest/dataprovider/GaussianNoiserTest.kt#L55)):*

Yeah maybe we should have a noiserType field in `Direct` protocol in `ProtocolConfig`.
</blockquote></details>

We also need the noiser for the aggregator too, right? Can you add a TODO for that in the mpc code?

iverson52000 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/904)* status: 8 of 13 files reviewed, 10 unresolved discussions (waiting on @jiayu-google, @renjiezh, @SanjayVas, @stevenwarejones, and @uakyol)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt` line 733 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/904#-NRcoqRX4wao5vntr2eL:-NRdqY0J-sMy8FGiDF5c:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/95f822802c42ab00acd373eec9395831b551ec94/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt#L733)):*

add a TODO to use the noiser specified in the measurementSpec
</blockquote></details>

Done.
___
*[`src/test/kotlin/org/wfanet/measurement/loadtest/dataprovider/GaussianNoiserTest.kt` line 55 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/904#-NRKDfdP3DjKi7H0olp5:-NRdqa5l-9Umyeywpz0W:b-kfnhx6) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4be505c2e8c0aa784f23fd48da6a9fd0bde0f4c9/src/test/kotlin/org/wfanet/measurement/loadtest/dataprovider/GaussianNoiserTest.kt#L55)):*

We also need the noiser for the aggregator too, right? Can you add a TODO for that in the mpc code?
</blockquote></details>

The noiser implemented here is for direct measurements only. The noiser for multi-EDP MPC measurement is already implemented in any-sketch. 
https://github.com/world-federation-of-advertisers/any-sketch/commit/c7eb1c403f5519ff84c5f1f2a59d283bea7b8992

@renjiezh is working on adding noising mechanism config in MPC code.
___
*[`src/test/kotlin/org/wfanet/measurement/loadtest/dataprovider/GaussianNoiserTest.kt` line 60 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/904#-NR42-EStKTx9KoX2Dsv-r2-60:-NRdqwdH4CXV79Mm9XOW:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/95f822802c42ab00acd373eec9395831b551ec94/src/test/kotlin/org/wfanet/measurement/loadtest/dataprovider/GaussianNoiserTest.kt#L60)):*
<details><summary><i>Previously, jiayu-google wrote…</i></summary><blockquote>

To exclude the possibility of overflow, it's great to test the maximum and minimum epsilon to use in practice. max epsilon = 1 as it is here. min epsilon = 0.0001.
</blockquote></details>

Done.
___
*[`src/test/kotlin/org/wfanet/measurement/loadtest/dataprovider/GaussianNoiserTest.kt` line 61 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/904#-NR42-EStKTx9KoX2Dsv-r2-61:-NRdqx8Q8wJaQ9R7gZpp:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/95f822802c42ab00acd373eec9395831b551ec94/src/test/kotlin/org/wfanet/measurement/loadtest/dataprovider/GaussianNoiserTest.kt#L61)):*
<details><summary><i>Previously, jiayu-google wrote…</i></summary><blockquote>

The noise is too small when setting delta = 1.  delta = 1e-12 is realistic.
</blockquote></details>

Done.
___
*[`src/test/kotlin/org/wfanet/measurement/loadtest/dataprovider/GaussianNoiserTest.kt` line 65 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/904#-NR42-EStKTx9KoX2Dsv-r2-65:-NRdqxhI4m9iDfjKclcz:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/95f822802c42ab00acd373eec9395831b551ec94/src/test/kotlin/org/wfanet/measurement/loadtest/dataprovider/GaussianNoiserTest.kt#L65)):*
<details><summary><i>Previously, jiayu-google wrote…</i></summary><blockquote>

ditto
</blockquote></details>

Done.
___
*[`src/test/kotlin/org/wfanet/measurement/loadtest/dataprovider/GaussianNoiserTest.kt` line 73 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/904#-NRdr0b85zXPQ7hUqlqr:-NRdr0b85zXPQ7hUqlqs:b-g10l5j) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/439c6d257039cdc9e75a178369b4f34deb4f4bbd/src/test/kotlin/org/wfanet/measurement/loadtest/dataprovider/GaussianNoiserTest.kt#L73)):*
> ```Kotlin
> 
>     val expectedNoisedReachValue = 52897
>     val expectedNoisedFrequencyMap =
> ```

Need to verify the correctness of the values with @jiayu-google

SanjayVas (changes requested):
> Reviewed 8 of 12 files at r2, 5 of 5 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/904)* status: all files reviewed, 11 unresolved discussions (waiting on @iverson52000, @jiayu-google, @renjiezh, and @uakyol)

___
*[`src/main/kotlin/org/wfanet/measurement/integration/common/InProcessEdpSimulator.kt` line 129 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/904#-NRiK68YBY7h1hHpZdtr:-NRiK68YBY7h1hHpZdts:b-ih3gcv) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/439c6d257039cdc9e75a178369b4f34deb4f4bbd/src/main/kotlin/org/wfanet/measurement/integration/common/InProcessEdpSimulator.kt#L129)):*
> ```Kotlin
>   companion object {
>     private const val randomSeed: Long = 1
> ```

CONST_CASE



_Suggestion:_
```Kotlin
RANDOM_SEED
```
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/AbstractNoiser.kt` line 26 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/904#-NRiKWdh23z-gZWD_VYA:-NRiKWdh23z-gZWD_VYB:b-83f7b0) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/439c6d257039cdc9e75a178369b4f34deb4f4bbd/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/AbstractNoiser.kt#L26)):*
> ```Kotlin
>   /**
>    * Add publisher noise to calculated direct reach and frequency.
>    *
> ```

This is overriding the KDoc of the base method with the same content. Just leave out the KDoc for overridden methods unless you actually need it to be different.



_Code quote:_
```Kotlin
  /**
   * Add publisher noise to calculated direct reach and frequency.
   *
   * @param reachValue Direct reach value.
   * @param frequencyMap Direct frequency.
   * @return Pair of noised reach value and frequency map.
   */
```
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulatorRunner.kt` line 82 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/904#-NRiLHJrEYEf_KI56Xin:-NRiLHJrEYEf_KI56Xio:bkqqssg) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/439c6d257039cdc9e75a178369b4f34deb4f4bbd/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulatorRunner.kt#L82)):*
> ```Kotlin
> 
>     val randomSeed: Long = 1
>     val random = java.util.Random(randomSeed)
> ```

This should be driven by an optional flag. By default, it should use a "real" Random instance without seeding.



_Code quote:_
```Kotlin
    val randomSeed: Long = 1
    val random = java.util.Random(randomSeed)
```
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/GaussianNoiser.kt` line 40 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/904#-NRiMln2ENCuGoEXVMOp:-NRiMln2ENCuGoEXVMOq:b-qizbjk) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/439c6d257039cdc9e75a178369b4f34deb4f4bbd/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/GaussianNoiser.kt#L40)):*
> ```Kotlin
> 
>   /**
>    * This implementation is adapted from jiayu-google. Assuming sensitivity = 1, solve delta given
> ```

Google is already listed in the AUTHORS file so we need not specify this if it's just written by a Googler for this project. If you want to credit Jiayu on this PR, you can [add him as a co-author](https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors).



_Code quote:_
```Kotlin
jiayu-google
```
___
*[`src/test/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulatorTest.kt` line 393 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/904#-NRiLnC_EwyTfatA1ocW:-NRiLnC_EwyTfatA1ocX:b-etcxbw) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/439c6d257039cdc9e75a178369b4f34deb4f4bbd/src/test/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulatorTest.kt#L393)):*
> ```Kotlin
> 
>       val allEvents = matchingEvents + nonMatchingEvents
>       val random = java.util.Random()
> ```

nit: import java.util.Random



_Suggestion:_
```Kotlin
Random
```

uakyol (approved):
> Reviewed 8 of 12 files at r2, 5 of 5 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/904)* status: all files reviewed, 11 unresolved discussions (waiting on @iverson52000, @jiayu-google, and @renjiezh)

renjiezh (approved):
> Reviewed 8 of 12 files at r2, 5 of 5 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/904)* status: all files reviewed, 11 unresolved discussions (waiting on @iverson52000, @jiayu-google, and @stevenwarejones)

___
*[`src/test/kotlin/org/wfanet/measurement/loadtest/dataprovider/GaussianNoiserTest.kt` line 55 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/904#-NRKDfdP3DjKi7H0olp5:-NRj2EoA3liiyRB7woUc:bvcbk41) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4be505c2e8c0aa784f23fd48da6a9fd0bde0f4c9/src/test/kotlin/org/wfanet/measurement/loadtest/dataprovider/GaussianNoiserTest.kt#L55)):*

we'l still need to choose a noiser based on the measurementspec, though, right?
</blockquote></details>

To align with multiparty measurements, the `noise_type` (called `noise_mechasm` actually) could be added into `ProtocolConfig.Direct`.
The noise type is not chosen by MeasurementConsumer. For multiparty measurements, it is chosen by the Kingdom. For direct measurements, it is by the EDPs.
Note that, MeasurementSpec is specified by the MC and cannot be modified later because it requires encryptions & signing, thus is somehow read-only.

iverson52000 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/904)* status: all files reviewed, 11 unresolved discussions (waiting on @jiayu-google, @renjiezh, and @stevenwarejones)

___
*[`src/test/kotlin/org/wfanet/measurement/loadtest/dataprovider/GaussianNoiserTest.kt` line 55 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/904#-NRKDfdP3DjKi7H0olp5:-NRj5QqV4N1HjannyHWN:bodfzkz) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4be505c2e8c0aa784f23fd48da6a9fd0bde0f4c9/src/test/kotlin/org/wfanet/measurement/loadtest/dataprovider/GaussianNoiserTest.kt#L55)):*

To align with multiparty measurements, the `noise_type` (called `noise_mechasm` actually) could be added into `ProtocolConfig.Direct`.
The noise type is not chosen by MeasurementConsumer. For multiparty measurements, it is chosen by the Kingdom. For direct measurements, it is by the EDPs.
Note that, MeasurementSpec is specified by the MC and cannot be modified later because it requires encryptions & signing, thus is somehow read-only.

</blockquote></details>

Yes we can add a field `noise_mechanism` into `ProtocolConfig.Direct` and have a flag in EDP to specify which noiser to use. For direct measurements, EDP has the flexibility to choose which dp noise to add but eventually, Gaussian noise is preferred due to better composition property.

I can work on adding `noise_mechanism` into `ProtocolConfig.Direct` in public API and create a flag in EDP in a separate PR

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/904)* status: all files reviewed, 10 unresolved discussions (waiting on @iverson52000, @jiayu-google, @renjiezh, and @stevenwarejones)

___
*[`src/test/kotlin/org/wfanet/measurement/loadtest/dataprovider/GaussianNoiserTest.kt` line 55 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/904#-NRKDfdP3DjKi7H0olp5:-NRnSZMH1LiSY7Shd4R4:bt8q2pl) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4be505c2e8c0aa784f23fd48da6a9fd0bde0f4c9/src/test/kotlin/org/wfanet/measurement/loadtest/dataprovider/GaussianNoiserTest.kt#L55)):*
> For multiparty measurements, it is chosen by the Kingdom. For direct measurements, it is by the EDPs.

> I can work on adding noise_mechanism into `ProtocolConfig.Direct` in public API

These two statements seem to be at odds. If it's chosen by each EDP, then the specific mechanism cannot be in the `ProtocolConfig` as that is set by the Kingdom.

jiayu-google (approved):
Thanks Albert! LGTM

stevenwarejones (changes requested):
> Reviewed 7 of 7 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/904)* status: all files reviewed, 10 unresolved discussions (waiting on @iverson52000, @jiayu-google, @renjiezh, and @SanjayVas)

___
*[`src/test/kotlin/org/wfanet/measurement/loadtest/dataprovider/GaussianNoiserTest.kt` line 55 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/904#-NRKDfdP3DjKi7H0olp5:-NS6f8Ez4sjU9w-ZIZ_m:b4oew7g) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4be505c2e8c0aa784f23fd48da6a9fd0bde0f4c9/src/test/kotlin/org/wfanet/measurement/loadtest/dataprovider/GaussianNoiserTest.kt#L55)):*

Here I mean for **direct measurements**, add `NoiseMechanism` enum into Direct protocol like we did for mpc measurements:
```
  // Protocol for direct measurement.
  message Direct {
    enum NoiseMechanism {
      NOISE_MECHANISM_UNSPECIFIED = 0;

      LAPLACE = 1;
      GAUSSIAN = 2;
    }
  }
```
In EDP, create a flag `directNoiseMechanism` which will be set by EDP(Laplace or Gaussian). And in the code, check
```
If (flag.directNoiseMechanism == ProtocolConfig.Direct.NoiseMechanismCase.GAUSSIAN) {
  useGaussianNoiser()
} else {
  useLaplaceNoiser()
}
```
Does this make sense?
</blockquote></details>

yes - this makes sense. I'm also saying that the noise mechanism chosen should affect the kind of noise added during the MPC process before the final result is revealed to the MC.

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/904)* status: all files reviewed, 10 unresolved discussions (waiting on @iverson52000, @jiayu-google, @renjiezh, and @SanjayVas)

___
*[`src/test/kotlin/org/wfanet/measurement/loadtest/dataprovider/GaussianNoiserTest.kt` line 55 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/904#-NRKDfdP3DjKi7H0olp5:-NS6fnLLAXTOqvsICDhc:b-6xr6wn) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4be505c2e8c0aa784f23fd48da6a9fd0bde0f4c9/src/test/kotlin/org/wfanet/measurement/loadtest/dataprovider/GaussianNoiserTest.kt#L55)):*
> world-federation-of-advertisers/any-sketch@c7eb1c4
I guess what I'm asking is where is the kingdom code that chooses with noiser to use based on the kingdom config? I must have missed that PR if it landed already.

iverson52000 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/904)* status: all files reviewed, 10 unresolved discussions (waiting on @jiayu-google, @renjiezh, @SanjayVas, and @stevenwarejones)

___
*[`src/test/kotlin/org/wfanet/measurement/loadtest/dataprovider/GaussianNoiserTest.kt` line 55 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/904#-NRKDfdP3DjKi7H0olp5:-NS7gxI99fyIqaizAU5N:b-eup1n) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4be505c2e8c0aa784f23fd48da6a9fd0bde0f4c9/src/test/kotlin/org/wfanet/measurement/loadtest/dataprovider/GaussianNoiserTest.kt#L55)):*

> world-federation-of-advertisers/any-sketch@c7eb1c4
I guess what I'm asking is where is the kingdom code that chooses with noiser to use based on the kingdom config? I must have missed that PR if it landed already.
</blockquote></details>

Got it. @renjiezh just opened up the PR last week for the MPC and Kingdom changes for multi-EDP noiser https://github.com/world-federation-of-advertisers/cross-media-measurement/pull/917

iverson52000 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/904)* status: all files reviewed, 10 unresolved discussions (waiting on @jiayu-google, @renjiezh, @SanjayVas, and @stevenwarejones)

___
*[`src/test/kotlin/org/wfanet/measurement/loadtest/dataprovider/GaussianNoiserTest.kt` line 55 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/904#-NRKDfdP3DjKi7H0olp5:-NS7vd9b7pBC0b_eONvN:b-pvjcbk) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4be505c2e8c0aa784f23fd48da6a9fd0bde0f4c9/src/test/kotlin/org/wfanet/measurement/loadtest/dataprovider/GaussianNoiserTest.kt#L55)):*

are you planning on adding the direct measurement noiser selector in this PR or a follow-up?
</blockquote></details>

I would prefer a follow-up PR for direct measurement noiser selector since the main focus of this PR is Noiser class and Gaussian Noiser creation

SanjayVas (changes requested):
> Reviewed 7 of 7 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/904)* status: all files reviewed, 7 unresolved discussions (waiting on @iverson52000, @jiayu-google, @renjiezh, and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulatorFlags.kt` line 100 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/904#-NS86dPs5kBu30a5kV_s:-NS86dPs5kBu30a5kV_t:bjdjqva) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d64122dc719621067354e30315e9aa9b458a6b69/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulatorFlags.kt#L100)):*
> ```Kotlin
>     required = false,
>   )
>   var randomSeed: Long? = 1
> ```

Default should be unset, i.e. don't specify a seed.



_Suggestion:_
```Kotlin
null
```
___
*[`src/test/kotlin/org/wfanet/measurement/loadtest/dataprovider/GaussianNoiserTest.kt` line 55 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/904#-NRKDfdP3DjKi7H0olp5:-NS87AyB7VDbK8h60-ar:bnd4woc) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4be505c2e8c0aa784f23fd48da6a9fd0bde0f4c9/src/test/kotlin/org/wfanet/measurement/loadtest/dataprovider/GaussianNoiserTest.kt#L55)):*

I would prefer a follow-up PR for direct measurement noiser selector since the main focus of this PR is Noiser class and Gaussian Noiser creation
</blockquote></details>

I'm still confused here. If there is a `noise_mechanism` field in the `Direct` `ProtocolConfig`, it means that it is selected by the Kingdom and that the EDP must respect that. Alternatively, the EDP simulator can have a flag that controls it (which would be akin to the EDP deciding on the noise mechanism as opposed to the Kingdom). The enum itself used for the flag can just be one locally defined in `EdpSimulatorFlags.kt` if there's not a suitable one already defined. We should **not** add an enum to the API that's only used by the EDP simulator runner.

iverson52000 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/904)* status: all files reviewed, 7 unresolved discussions (waiting on @jiayu-google, @renjiezh, @SanjayVas, and @stevenwarejones)

___
*[`src/test/kotlin/org/wfanet/measurement/loadtest/dataprovider/GaussianNoiserTest.kt` line 55 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/904#-NRKDfdP3DjKi7H0olp5:-NSCAabC52l4l07cjaTS:b9zwhfz) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4be505c2e8c0aa784f23fd48da6a9fd0bde0f4c9/src/test/kotlin/org/wfanet/measurement/loadtest/dataprovider/GaussianNoiserTest.kt#L55)):*

I'm still confused here. If there is a `noise_mechanism` field in the `Direct` `ProtocolConfig`, it means that it is selected by the Kingdom and that the EDP must respect that. Alternatively, the EDP simulator can have a flag that controls it (which would be akin to the EDP deciding on the noise mechanism as opposed to the Kingdom). The enum itself used for the flag can just be one locally defined in `EdpSimulatorFlags.kt` if there's not a suitable one already defined. We should **not** add an enum to the API that's only used by the EDP simulator runner.
</blockquote></details>

@SanjayVas Thanks for the clarification! I thought we should put the `NoiseMechanism` enum for direct measurements in public API as well. I am totally fine with creating a `NoiseMechanism` enum in `EdpSimulatorFlags.kt` and then specify the direct measurement noiser based on it. It’s straightforward and can be done in this PR. @stevenwarejones Does this approach sound good to you?

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/904)* status: all files reviewed, 7 unresolved discussions (waiting on @iverson52000, @jiayu-google, @renjiezh, and @SanjayVas)

___
*[`src/test/kotlin/org/wfanet/measurement/loadtest/dataprovider/GaussianNoiserTest.kt` line 55 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/904#-NRKDfdP3DjKi7H0olp5:-NSGle7GCyIjE_3rkYkv:b-lwss64) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4be505c2e8c0aa784f23fd48da6a9fd0bde0f4c9/src/test/kotlin/org/wfanet/measurement/loadtest/dataprovider/GaussianNoiserTest.kt#L55)):*

@SanjayVas Thanks for the clarification! I thought we should put the `NoiseMechanism` enum for direct measurements in public API as well. I am totally fine with creating a `NoiseMechanism` enum in `EdpSimulatorFlags.kt` and then specify the direct measurement noiser based on it. It’s straightforward and can be done in this PR. @stevenwarejones Does this approach sound good to you?
</blockquote></details>

I think the enum does need to be part of the public API. There should be a list of acceptable Noisers (including no noise) for direct measurements. The EDP should choose one of them *and report* back which one it used for PBM tracking.

Can you add a TODO for that somewhere?

iverson52000 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/904)* status: 3 of 14 files reviewed, 7 unresolved discussions (waiting on @jiayu-google, @renjiezh, @SanjayVas, @stevenwarejones, and @uakyol)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulatorFlags.kt` line 100 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/904#-NS86dPs5kBu30a5kV_s:-NSHLx0E7du21SQdfUJw:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d64122dc719621067354e30315e9aa9b458a6b69/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulatorFlags.kt#L100)):*

Default should be unset, i.e. don't specify a seed.
</blockquote></details>

Done.
___
*[`src/test/kotlin/org/wfanet/measurement/loadtest/dataprovider/GaussianNoiserTest.kt` line 55 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/904#-NRKDfdP3DjKi7H0olp5:-NSHLybH4krBxkJ4u2CF:b243wcs) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4be505c2e8c0aa784f23fd48da6a9fd0bde0f4c9/src/test/kotlin/org/wfanet/measurement/loadtest/dataprovider/GaussianNoiserTest.kt#L55)):*

I think the enum does need to be part of the public API. There should be a list of acceptable Noisers (including no noise) for direct measurements. The EDP should choose one of them *and report* back which one it used for PBM tracking.

Can you add a TODO for that somewhere?
</blockquote></details>

Yesterday I added the `NoiseMechanism` enum and flag in `EdpSimulatorFlags.kt` and sure we can move the enum into public API when there's the need to report back which one it used for PBM tracking. I also added the TODO in `EdpSimulatorFlags.kt` for it. Please review thanks!

SanjayVas (changes requested):
> Reviewed 7 of 11 files at r5, 2 of 2 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/904)* status: 12 of 14 files reviewed, 11 unresolved discussions (waiting on @iverson52000, @jiayu-google, @renjiezh, and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt` line 157 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/904#-NSHfarQ7rDi_kpzVw7I:-NSHfarQ7rDi_kpzVw7J:bvm02sf) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c2ae9e6b91c73f3484c47bb79bd1cc087da03c47/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt#L157)):*
> ```Kotlin
>   private val trustedCertificates: Map<ByteString, X509Certificate>,
>   private val random: Random,
>   private val noiseMechanism: String
> ```

Use enum



_Suggestion:_
```Kotlin
NoiseMechanism
```
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulatorFlags.kt` line 114 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/904#-NSHf7P90afskbsSG9YS:-NSHf7P90afskbsSG9YT:b-fmx09h) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c2ae9e6b91c73f3484c47bb79bd1cc087da03c47/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulatorFlags.kt#L114)):*
> ```Kotlin
>     names = ["--noise-mechanism"],
>     description = ["Differential privacy noise mechanism for direct measurements"],
>     required = true,
> ```

Required with a default value doesn't make sense. Note that required for a command-line option means the caller must specify it.



_Code quote:_
```Kotlin
required = true,
```
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulatorFlags.kt` line 117 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/904#-NSHfVdW6RK3mPflTojF:-NSHfVdW6RK3mPflTojG:bx550z2) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c2ae9e6b91c73f3484c47bb79bd1cc087da03c47/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulatorFlags.kt#L117)):*
> ```Kotlin
>     defaultValue = "LAPLACE",
>   )
>   var noiseMechanism by Delegates.notNull<String>()
> ```

Use the enum type, not String. Picocli knows how to convert it.



_Code quote:_
```Kotlin
by Delegates.notNull<String>()
```
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/GaussianNoiser.kt` line 26 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/904#-NSHg-gN8EKsMubspQVs:-NSHg-gN8EKsMubspQVt:b-6k3wg0) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c2ae9e6b91c73f3484c47bb79bd1cc087da03c47/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/GaussianNoiser.kt#L26)):*
> ```Kotlin
> class GaussianNoiser(reachAndFrequency: MeasurementSpec.ReachAndFrequency, random: Random) :
>   AbstractNoiser() {
>   override val distributionForReach: NormalDistribution by lazy {
> ```

Curious why you switched this to being lazy rather than eager. Note that `lazy` has some overhead in terms of memory and locking for thread safety, so we only want to use it when there's a clear reason to do so.

* It looks like the value *can* be computed during construction, i.e. all inputs are known at this time.
* If a property will always be used, then there's no point in delaying initialization. Read on if one or the other is not always used.
* Is NormalDistribution large or is computing it expensive? If not, then the lazy initialization overhead may be a waste.



_Code quote:_
```Kotlin
 by lazy
```
___
*[`src/test/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulatorTest.kt` line 393 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/904#-NRiLnC_EwyTfatA1ocW:-NSHj8TU6JIqcgT5hEJP:b-6ds5d3) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/439c6d257039cdc9e75a178369b4f34deb4f4bbd/src/test/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulatorTest.kt#L393)):*

`kotlin.random.Random` is also used in this test so that's why I didn't import `java.util.Random`
</blockquote></details>

Ah, apologies.

It turns out that you can trivially convert a `kotlin.random.Random` to a `java.util.Random` (on JVM, the Kotlin one actually just wraps a Java one anyway). So e.g.

```kotlin
val random = Random.Default
```

and then when you need a `java.util.random`:

```kotlin
random.asJavaRandom()
```

In fact, it may be better to have the EdpSimulator constructor take in a `kotlin.random.Random` instead.
___
*[`src/test/kotlin/org/wfanet/measurement/loadtest/dataprovider/GaussianNoiserTest.kt` line 55 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/904#-NRKDfdP3DjKi7H0olp5:-NSHkb68BjVv_OCisbAu:b-bf7se1) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4be505c2e8c0aa784f23fd48da6a9fd0bde0f4c9/src/test/kotlin/org/wfanet/measurement/loadtest/dataprovider/GaussianNoiserTest.kt#L55)):*

Yesterday I added the `NoiseMechanism` enum and flag in `EdpSimulatorFlags.kt` and sure we can move the enum into public API when there's the need to report back which one it used for PBM tracking. I also added the TODO in `EdpSimulatorFlags.kt` for it. Please review thanks!
</blockquote></details>

Ah, there's already a NoiseMechanism enum in the public API as of world-federation-of-advertisers/cross-media-measurement-api#127. Just use that.

The implementation for reporting the mechanism back would involve changing `FulfillDirectRequisitionRequest`, not `ProtocolConfig.Direct`.

iverson52000 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/904)* status: 9 of 14 files reviewed, 11 unresolved discussions (waiting on @jiayu-google, @renjiezh, @SanjayVas, and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt` line 157 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/904#-NSHfarQ7rDi_kpzVw7I:-NSI-GZq9rKYGWOiFaqF:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c2ae9e6b91c73f3484c47bb79bd1cc087da03c47/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt#L157)):*

Use enum
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulatorFlags.kt` line 114 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/904#-NSHf7P90afskbsSG9YS:-NSI-HZN1mg1BjIGFMmQ:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c2ae9e6b91c73f3484c47bb79bd1cc087da03c47/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulatorFlags.kt#L114)):*

Required with a default value doesn't make sense. Note that required for a command-line option means the caller must specify it.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulatorFlags.kt` line 117 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/904#-NSHfVdW6RK3mPflTojF:-NSI-Ivh4IrZKVybMP0n:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c2ae9e6b91c73f3484c47bb79bd1cc087da03c47/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulatorFlags.kt#L117)):*

Use the enum type, not String. Picocli knows how to convert it.
</blockquote></details>

Done.

iverson52000 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/904)* status: 9 of 14 files reviewed, 11 unresolved discussions (waiting on @jiayu-google, @renjiezh, @riemanli, @SanjayVas, and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/GaussianNoiser.kt` line 26 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/904#-NSHg-gN8EKsMubspQVs:-NSIUvXl99UB-cQMksIc:bdm5ey) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c2ae9e6b91c73f3484c47bb79bd1cc087da03c47/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/GaussianNoiser.kt#L26)):*

I spoked with @riemanli and he mentioned there's also needs to add publisherNoise to other direct measurement types(Impression, Duration). My idea here is to use the same Noiser for all direct measurement types and lazily initialized the distribution needed. For example, if we want to use it for Impression too:
```
class GaussianNoiser(
  reachAndFrequency: MeasurementSpec.ReachAndFrequency,
  impression: MeasurementSpec.Impression,
  random: Random
) : AbstractNoiser() {
  override val distributionForReach: NormalDistribution by lazy {
    getNormalDistribution(
      reachAndFrequency.reachPrivacyParams.epsilon,
      reachAndFrequency.reachPrivacyParams.delta,
      random
    )
  }

  override val distributionForFrequency: NormalDistribution by lazy {
    getNormalDistribution(
      reachAndFrequency.frequencyPrivacyParams.epsilon,
      reachAndFrequency.frequencyPrivacyParams.delta,
      random
    )
  }

  override val distributionForImpression: NormalDistribution by lazy {
    getNormalDistribution(impression.privacyParams.epsilon, impression.privacyParams.delta, random)
  }

  fun addImpressionPublisherNoise(
    impression: Long,
  ): NoisedImpression
```
`fun addImpressionPublisherNoise` will be in AbstractNoiser.kt. Is `lazy` suitable for this use case?

</blockquote></details>

Actually scratch that. I just realized this `Noiser` design may overcomplicate things. Maybe should just have a `Distribution` class which return random samples and handle the  noising outside

iverson52000 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/904)* status: 4 of 14 files reviewed, 11 unresolved discussions (waiting on @jiayu-google, @renjiezh, @riemanli, @SanjayVas, and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/GaussianNoiser.kt` line 26 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/904#-NSHg-gN8EKsMubspQVs:-NSMX1Gs5Zu_TUXpP2cI:b-e8xept) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c2ae9e6b91c73f3484c47bb79bd1cc087da03c47/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/GaussianNoiser.kt#L26)):*

Actually scratch that. I just realized this `Noiser` design may overcomplicate things. Maybe should just have a `Distribution` class which return random samples and handle the  noising outside
</blockquote></details>

Done. Refactor `Noiser` class to be initialized with `DifferentialPrivacyParams` and `Random`, and has a `sample()` method just returns sampled random value from the distribution. This design reconciles with the incoming changes in
https://github.com/world-federation-of-advertisers/cross-media-measurement/pull/911

cc @riemanli

SanjayVas (changes requested):
> Reviewed 2 of 4 files at r7, 6 of 7 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/904)* status: 12 of 14 files reviewed, 8 unresolved discussions (waiting on @iverson52000, @jiayu-google, @renjiezh, and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulatorFlags.kt` line 25 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/904#-NSMnB8wBvcPXjdt2iDk:-NSMnB8wBvcPXjdt2iDl:b-o915kw) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/b4e5940b1d859af1c731b9c63f4c61f320b94590/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulatorFlags.kt#L25)):*
> ```Kotlin
> import picocli.CommandLine
> 
> // TODO(@iverson52000): Move this to public API if EDP needs to report back which noiser is used for
> ```

KDoc syntax for type definitions



_Suggestion:_
```Kotlin
/**
 * Mechanism for generating noise.
 *
 * TODO(@iverson52000): Move this to public API if EDP needs to report back which noiser is used for PBM tracking.
 */
```
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulatorFlags.kt` line 28 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/904#-NSMn1aO5DSoxFTlYlU1:-NSMn1aO5DSoxFTlYlU2:b-ke1gig) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/b4e5940b1d859af1c731b9c63f4c61f320b94590/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulatorFlags.kt#L28)):*
> ```Kotlin
> // PBM tracking.
> enum class NoiseMechanism {
>   LAPLACE,
> ```

nit: define this in EdpSimulator.kt or in a separate file. It's weird for EdpSimulator.kt to depend on a value defined in EdpSimulatorFlags.kt



_Code quote:_
```Kotlin
enum class NoiseMechanism {
  LAPLACE,
  GAUSSIAN,
}
```
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulatorFlags.kt` line 116 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/904#-NSMm1sV2VS_EZGEb22H:-NSMm1sV2VS_EZGEb22I:b-dwoavs) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/b4e5940b1d859af1c731b9c63f4c61f320b94590/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulatorFlags.kt#L116)):*
> ```Kotlin
>     defaultValue = "LAPLACE",
>   )
>   var noiseMechanism by Delegates.notNull<NoiseMechanism>()
> ```

I think with enum values, you can use `lateinit var` instead of `Delegates.notNull`. We only use the latter in cases where `lateinit var` wouldn't work.



_Suggestion:_
```Kotlin
lateinit var noiseMechanism: NoiseMechanism
```
___
*[`src/test/kotlin/org/wfanet/measurement/loadtest/dataprovider/GaussianNoiserTest.kt` line 55 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/904#-NRKDfdP3DjKi7H0olp5:-NSMmTOaD9JF_PNroRYO:bfck4zo) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4be505c2e8c0aa784f23fd48da6a9fd0bde0f4c9/src/test/kotlin/org/wfanet/measurement/loadtest/dataprovider/GaussianNoiserTest.kt#L55)):*

The `ProtocolConfig.NoiseMechanism` enum we have right now in public API is **discrete** noise distribution for MPC(The sum of GEOMETRIC distribution approximates discrete Laplace distribution). For direct measurements, we are using **continuous** noise distribution which is technically from discrete ones so there's a need to differentiate discrete and continuous noising mechanisms for clarity. Maybe something like `DirectNoiseMechanism`?

I agree reporting the mechanism back would involve changing `FulfillDirectRequisitionRequest`
</blockquote></details>

Ah, I didn't realize the values are different. Fine to keep the enum here for now.

iverson52000 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/904)* status: 10 of 14 files reviewed, 7 unresolved discussions (waiting on @jiayu-google, @renjiezh, @SanjayVas, and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulatorFlags.kt` line 25 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/904#-NSMnB8wBvcPXjdt2iDk:-NSN8wj-4UqxOZxB5eeb:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/b4e5940b1d859af1c731b9c63f4c61f320b94590/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulatorFlags.kt#L25)):*

KDoc syntax for type definitions
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulatorFlags.kt` line 116 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/904#-NSMm1sV2VS_EZGEb22H:-NSN8yEy8bba4LOEzmNv:bwppva0) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/b4e5940b1d859af1c731b9c63f4c61f320b94590/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulatorFlags.kt#L116)):*

I think with enum values, you can use `lateinit var` instead of `Delegates.notNull`. We only use the latter in cases where `lateinit var` wouldn't work.
</blockquote></details>

Done. Thanks for the reminder.

jiayu-google (approved):

SanjayVas (changes requested):
> Reviewed 1 of 4 files at r7, 1 of 7 files at r8, 2 of 2 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/904)* status: all files reviewed, 6 unresolved discussions (waiting on @iverson52000, @jiayu-google, @renjiezh, and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulatorFlags.kt` line 116 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/904#-NSMm1sV2VS_EZGEb22H:-NSO1zDX2PomICjMO2XW:b-z48wg2) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/b4e5940b1d859af1c731b9c63f4c61f320b94590/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulatorFlags.kt#L116)):*

Done. Thanks for the reminder.
</blockquote></details>

I think you accidentally dropped the `private set`

stevenwarejones (approved):
> Reviewed 1 of 11 files at r5, 2 of 4 files at r7, 4 of 7 files at r8, 2 of 2 files at r9.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/904)* status: all files reviewed, 6 unresolved discussions (waiting on @iverson52000, @jiayu-google, and @renjiezh)

iverson52000 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/904)* status: 13 of 14 files reviewed, 6 unresolved discussions (waiting on @jiayu-google, @renjiezh, @SanjayVas, @set, and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulatorFlags.kt` line 116 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/904#-NSMm1sV2VS_EZGEb22H:-NSRmHcxCvjih92LIxWR:b-sv0bdl) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/b4e5940b1d859af1c731b9c63f4c61f320b94590/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulatorFlags.kt#L116)):*

I think you accidentally dropped the `private set`
</blockquote></details>

I did sorry. And removed @set: since we are not using the `notNull` property delegate here

SanjayVas (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/904)* status: all files reviewed, 5 unresolved discussions (waiting on @iverson52000, @jiayu-google, and @renjiezh)

jiayu-google (approved):

iverson52000 (commented):
> Reviewed 2 of 12 files at r2, 1 of 5 files at r3, 1 of 11 files at r5, 2 of 4 files at r7, 6 of 7 files at r8, 1 of 2 files at r9, 1 of 1 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/904)* status: all files reviewed, 4 unresolved discussions (waiting on @jiayu-google and @renjiezh)

iverson52000 (commented):
> Dismissed @jiayu-google from 4 discussions.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/904)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @renjiezh)

PR #909, opened 2023-03-22
Extract Kind-specific portions of correctness test.

SanjayVas:
This is to prepare the correctness test to be runnable against any K8s cluster.

Future work:
* Use an existing container image registry that already has the images.
* Use a specified kubeconfig to connect to the cluster.


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/909)

PR #911, opened 2023-03-23
Implement reach measurement type

riemanli:
Implement reach-only measurements without breaking the current usage in upstream services.

### DIFF
--- a/build/repositories.bzl
+++ b/build/repositories.bzl
@@ -37,9 +37,9 @@ def wfa_measurement_system_repositories():
 
     wfa_repo_archive(
         name = "wfa_measurement_proto",
+        commit = "265e1ac06d4d793f6debfdfec84201b59f11642b",
         repo = "cross-media-measurement-api",
-        sha256 = "8aae477270a606ecbd84a67274734ddfcb7164e0d0d4b9ccd0fd72ed44776ece",
-        version = "0.27.0",
+        sha256 = "4ddf9a98088d8e470f0b18fbe5fcbcdbc184c616492cc65ec42901e688f0f4fb",
     )
 
     wfa_repo_archive(
--- a/build/repositories.bzl
+++ b/build/repositories.bzl
@@ -35,7 +35,7 @@ def wfa_measurement_system_repositories():
         version = "0.8.0",
     )
 
-    # TODO: DO_NOT_SUBMIT
+    # DO_NOT_SUBMIT(world-federation-of-advertisers/cross-media-measurement-api#126): Use version once released.
     wfa_repo_archive(
         name = "wfa_measurement_proto",
         commit = "265e1ac06d4d793f6debfdfec84201b59f11642b",
--- a/build/repositories.bzl
+++ b/build/repositories.bzl
@@ -38,8 +38,8 @@ def wfa_measurement_system_repositories():
     wfa_repo_archive(
         name = "wfa_measurement_proto",
         repo = "cross-media-measurement-api",
-        sha256 = "cfde44921d11ccc04e22d4dd7a99b45751a3c3eb53e0cb3ba454100ffa616d27",
-        version = "0.28.0",
+        sha256 = "333ec3153cfe20d9f0ceeb9c73b0d11daa9f0b61382596de76d7090511bc591a",
+        version = "0.28.1",
     )
 
     wfa_repo_archive(
--- a/build/repositories.bzl
+++ b/build/repositories.bzl
@@ -35,6 +35,7 @@ def wfa_measurement_system_repositories():
         version = "0.8.0",
     )
 
+    # TODO: DO_NOT_SUBMIT
     wfa_repo_archive(
         name = "wfa_measurement_proto",
         commit = "265e1ac06d4d793f6debfdfec84201b59f11642b",
--- a/build/repositories.bzl
+++ b/build/repositories.bzl
@@ -35,12 +35,11 @@ def wfa_measurement_system_repositories():
         version = "0.8.0",
     )
 
-    # DO_NOT_SUBMIT(world-federation-of-advertisers/cross-media-measurement-api#126): Use version once released.
     wfa_repo_archive(
         name = "wfa_measurement_proto",
-        commit = "265e1ac06d4d793f6debfdfec84201b59f11642b",
         repo = "cross-media-measurement-api",
-        sha256 = "4ddf9a98088d8e470f0b18fbe5fcbcdbc184c616492cc65ec42901e688f0f4fb",
+        sha256 = "cfde44921d11ccc04e22d4dd7a99b45751a3c3eb53e0cb3ba454100ffa616d27",
+        version = "0.28.0",
     )
 
     wfa_repo_archive(
--- a/src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/Herald.kt
+++ b/src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/Herald.kt
@@ -230,6 +230,7 @@ class Herald(
     logger.info("[id=$globalId] Creating Computation...")
     try {
       when (systemComputation.toMeasurementType()) {
+        MeasurementType.REACH,
         MeasurementType.REACH_AND_FREQUENCY -> {
           LiquidLegionsV2Starter.createComputation(
             internalComputationsClient,
--- a/src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/LiquidLegionsV2Starter.kt
+++ b/src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/LiquidLegionsV2Starter.kt
@@ -298,27 +298,50 @@ object LiquidLegionsV2Starter {
           when (Version.fromString(publicApiVersion)) {
             Version.V2_ALPHA -> {
               val measurementSpec = MeasurementSpec.parseFrom(measurementSpec)
-              require(measurementSpec.hasReachAndFrequency()) {
-                "Missing ReachAndFrequency in the measurementSpec."
+
+              @Suppress("WHEN_ENUM_CAN_BE_NULL_IN_JAVA") // Proto enum fields are never null.
+              when (measurementSpec.measurementTypeCase) {
+                MeasurementSpec.MeasurementTypeCase.REACH -> {
+                  val reach = measurementSpec.reach
+                  require(reach.privacyParams.delta > 0) {
+                    "LLv2 requires that privacy_params.delta be greater than 0"
+                  }
+                  require(reach.privacyParams.epsilon > MIN_REACH_EPSILON) {
+                    "LLv2 requires that privacy_params.epsilon be greater than $MIN_REACH_EPSILON"
+                  }
+                  reachNoiseConfigBuilder.globalReachDpNoise =
+                    reach.privacyParams.toDuchyDifferentialPrivacyParams()
+                }
+                MeasurementSpec.MeasurementTypeCase.REACH_AND_FREQUENCY -> {
+                  val reachAndFrequency = measurementSpec.reachAndFrequency
+                  require(reachAndFrequency.reachPrivacyParams.delta > 0) {
+                    "LLv2 requires that reach_privacy_params.delta be greater than 0"
+                  }
+                  require(reachAndFrequency.reachPrivacyParams.epsilon > MIN_REACH_EPSILON) {
+                    "LLv2 requires that reach_privacy_params.epsilon be greater than $MIN_REACH_EPSILON"
+                  }
+                  require(reachAndFrequency.frequencyPrivacyParams.delta > 0) {
+                    "LLv2 requires that frequency_privacy_params.delta be greater than 0"
+                  }
+                  require(
+                    reachAndFrequency.frequencyPrivacyParams.epsilon > MIN_FREQUENCY_EPSILON
+                  ) {
+                    "LLv2 requires that frequency_privacy_params.epsilon be greater than " +
+                      "$MIN_FREQUENCY_EPSILON"
+                  }
+                  reachNoiseConfigBuilder.globalReachDpNoise =
+                    reachAndFrequency.reachPrivacyParams.toDuchyDifferentialPrivacyParams()
+                  frequencyNoiseConfig =
+                    reachAndFrequency.frequencyPrivacyParams.toDuchyDifferentialPrivacyParams()
+                }
+                MeasurementSpec.MeasurementTypeCase.IMPRESSION,
+                MeasurementSpec.MeasurementTypeCase.DURATION,
+                MeasurementSpec.MeasurementTypeCase.MEASUREMENTTYPE_NOT_SET -> {
+                  throw IllegalArgumentException(
+                    "Missing Reach and ReachAndFrequency in the measurementSpec."
+                  )
+                }
               }
-              val reachAndFrequency = measurementSpec.reachAndFrequency
-              require(reachAndFrequency.reachPrivacyParams.delta > 0) {
-                "LLv2 requires that reach_privacy_params.delta be greater than 0"
-              }
-              require(reachAndFrequency.reachPrivacyParams.epsilon > MIN_REACH_EPSILON) {
-                "LLv2 requires that reach_privacy_params.epsilon be greater than $MIN_REACH_EPSILON"
-              }
-              require(reachAndFrequency.frequencyPrivacyParams.delta > 0) {
-                "LLv2 requires that frequency_privacy_params.delta be greater than 0"
-              }
-              require(reachAndFrequency.frequencyPrivacyParams.epsilon > MIN_FREQUENCY_EPSILON) {
-                "LLv2 requires that frequency_privacy_params.epsilon be greater than " +
-                  "$MIN_FREQUENCY_EPSILON"
-              }
-              reachNoiseConfigBuilder.globalReachDpNoise =
-                reachAndFrequency.reachPrivacyParams.toDuchyDifferentialPrivacyParams()
-              frequencyNoiseConfig =
-                reachAndFrequency.frequencyPrivacyParams.toDuchyDifferentialPrivacyParams()
             }
             Version.VERSION_UNSPECIFIED -> error("Public api version is invalid or unspecified.")
           }
--- a/src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/LiquidLegionsV2Starter.kt
+++ b/src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/LiquidLegionsV2Starter.kt
@@ -35,6 +35,10 @@ import org.wfanet.measurement.internal.duchy.createComputationRequest
 import org.wfanet.measurement.internal.duchy.protocol.LiquidLegionsSketchAggregationV2.ComputationDetails.ComputationParticipant
 import org.wfanet.measurement.internal.duchy.protocol.LiquidLegionsSketchAggregationV2.ComputationDetails.Parameters
 import org.wfanet.measurement.internal.duchy.protocol.LiquidLegionsSketchAggregationV2.Stage
+import org.wfanet.measurement.internal.duchy.protocol.LiquidLegionsSketchAggregationV2Kt.ComputationDetailsKt.parameters
+import org.wfanet.measurement.internal.duchy.protocol.LiquidLegionsV2NoiseConfigKt.reachNoiseConfig
+import org.wfanet.measurement.internal.duchy.protocol.liquidLegionsSketchParameters
+import org.wfanet.measurement.internal.duchy.protocol.liquidLegionsV2NoiseConfig
 import org.wfanet.measurement.internal.duchy.updateComputationDetailsRequest
 import org.wfanet.measurement.system.v1alpha.Computation
 import org.wfanet.measurement.system.v1alpha.ComputationParticipant as SystemComputationParticipant
@@ -281,20 +285,20 @@ object LiquidLegionsV2Starter {
     }
     val llv2Config = mpcProtocolConfig.liquidLegionsV2
 
-    return Parameters.newBuilder()
-      .also {
-        it.maximumFrequency = llv2Config.maximumFrequency
-        it.liquidLegionsSketchBuilder.apply {
-          decayRate = llv2Config.sketchParams.decayRate
-          size = llv2Config.sketchParams.maxSize
-        }
-        it.noiseBuilder.apply {
-          reachNoiseConfigBuilder.apply {
-            val mpcNoise = llv2Config.mpcNoise
-            blindHistogramNoise = mpcNoise.blindedHistogramNoise.toDuchyDifferentialPrivacyParams()
-            noiseForPublisherNoise =
-              mpcNoise.noiseForPublisherNoise.toDuchyDifferentialPrivacyParams()
-          }
+    return parameters {
+      maximumFrequency = llv2Config.maximumFrequency
+      liquidLegionsSketch = liquidLegionsSketchParameters {
+        decayRate = llv2Config.sketchParams.decayRate
+        size = llv2Config.sketchParams.maxSize
+      }
+      ellipticCurveId = llv2Config.ellipticCurveId
+      noise = liquidLegionsV2NoiseConfig {
+        reachNoiseConfig = reachNoiseConfig {
+          val mpcNoise = llv2Config.mpcNoise
+          blindHistogramNoise = mpcNoise.blindedHistogramNoise.toDuchyDifferentialPrivacyParams()
+          noiseForPublisherNoise =
+            mpcNoise.noiseForPublisherNoise.toDuchyDifferentialPrivacyParams()
+
           when (Version.fromString(publicApiVersion)) {
             Version.V2_ALPHA -> {
               val measurementSpec = MeasurementSpec.parseFrom(measurementSpec)
@@ -309,8 +313,7 @@ object LiquidLegionsV2Starter {
                   require(reach.privacyParams.epsilon > MIN_REACH_EPSILON) {
                     "LLv2 requires that privacy_params.epsilon be greater than $MIN_REACH_EPSILON"
                   }
-                  reachNoiseConfigBuilder.globalReachDpNoise =
-                    reach.privacyParams.toDuchyDifferentialPrivacyParams()
+                  globalReachDpNoise = reach.privacyParams.toDuchyDifferentialPrivacyParams()
                 }
                 MeasurementSpec.MeasurementTypeCase.REACH_AND_FREQUENCY -> {
                   val reachAndFrequency = measurementSpec.reachAndFrequency
@@ -329,9 +332,9 @@ object LiquidLegionsV2Starter {
                     "LLv2 requires that frequency_privacy_params.epsilon be greater than " +
                       "$MIN_FREQUENCY_EPSILON"
                   }
-                  reachNoiseConfigBuilder.globalReachDpNoise =
+                  globalReachDpNoise =
                     reachAndFrequency.reachPrivacyParams.toDuchyDifferentialPrivacyParams()
-                  frequencyNoiseConfig =
+                  this@liquidLegionsV2NoiseConfig.frequencyNoiseConfig =
                     reachAndFrequency.frequencyPrivacyParams.toDuchyDifferentialPrivacyParams()
                 }
                 MeasurementSpec.MeasurementTypeCase.IMPRESSION,
@@ -346,8 +349,7 @@ object LiquidLegionsV2Starter {
             Version.VERSION_UNSPECIFIED -> error("Public api version is invalid or unspecified.")
           }
         }
-        it.ellipticCurveId = llv2Config.ellipticCurveId
       }
-      .build()
+    }
   }
 }
--- a/src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/MillBase.kt
+++ b/src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/MillBase.kt
@@ -410,10 +410,13 @@ abstract class MillBase(
   protected suspend fun existingOutputOr(
     token: ComputationToken,
     block: suspend () -> ByteString
-  ): ComputationResult {
+  ): EncryptedComputationResult {
     if (token.singleOutputBlobMetadata().path.isNotEmpty()) {
       // Reuse cached result if it exists
-      return ComputationResult(checkNotNull(dataClients.readSingleOutputBlob(token)), token)
+      return EncryptedComputationResult(
+        checkNotNull(dataClients.readSingleOutputBlob(token)),
+        token
+      )
     }
     val newResult: ByteString =
       try {
@@ -429,7 +432,10 @@ abstract class MillBase(
         // All errors from block() are permanent and would cause the computation to FAIL
         throw PermanentComputationError(error)
       }
-    return ComputationResult(flowOf(newResult), dataClients.writeSingleOutputBlob(token, newResult))
+    return EncryptedComputationResult(
+      flowOf(newResult),
+      dataClients.writeSingleOutputBlob(token, newResult)
+    )
   }
 
   /** Reads all input blobs and combines all the bytes together. */
@@ -531,7 +537,7 @@ const val CURRENT_RUNTIME_MEMORY_MAXIMUM = "current_runtime_memory_maximum"
 const val CURRENT_RUNTIME_MEMORY_TOTAL = "current_runtime_memory_total"
 const val CURRENT_RUNTIME_MEMORY_FREE = "current_runtime_memory_free"
 
-data class ComputationResult(val bytes: Flow<ByteString>, val token: ComputationToken)
+data class EncryptedComputationResult(val bytes: Flow<ByteString>, val token: ComputationToken)
 
 data class Certificate(
   // The public API name of this certificate.
--- a/src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/liquidlegionsv2/LiquidLegionsV2Mill.kt
+++ b/src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/liquidlegionsv2/LiquidLegionsV2Mill.kt
@@ -46,8 +46,8 @@ import org.wfanet.measurement.duchy.daemon.mill.MillBase
 import org.wfanet.measurement.duchy.daemon.mill.PermanentComputationError
 import org.wfanet.measurement.duchy.daemon.mill.liquidlegionsv2.crypto.LiquidLegionsV2Encryption
 import org.wfanet.measurement.duchy.daemon.utils.ComputationResult
-import org.wfanet.measurement.duchy.daemon.utils.Reach
-import org.wfanet.measurement.duchy.daemon.utils.ReachAndFrequency
+import org.wfanet.measurement.duchy.daemon.utils.ReachAndFrequencyResult
+import org.wfanet.measurement.duchy.daemon.utils.ReachResult
 import org.wfanet.measurement.duchy.daemon.utils.toAnySketchElGamalPublicKey
 import org.wfanet.measurement.duchy.daemon.utils.toCmmsElGamalPublicKey
 import org.wfanet.measurement.duchy.daemon.utils.toV2AlphaElGamalPublicKey
@@ -719,7 +719,7 @@ class LiquidLegionsV2Mill(
 
     // If this is a reach-only computation, then our job is done.
     if (measurementSpec.hasReach() || maximumRequestedFrequency == 1) {
-      sendResultToKingdom(token, Reach(reach))
+      sendResultToKingdom(token, ReachResult(reach))
       return completeComputation(nextToken, CompletedReason.SUCCEEDED)
     }
 
@@ -842,7 +842,7 @@ class LiquidLegionsV2Mill(
 
     sendResultToKingdom(
       token,
-      ReachAndFrequency(llv2Details.reachEstimate.reach, frequencyDistributionMap)
+      ReachAndFrequencyResult(llv2Details.reachEstimate.reach, frequencyDistributionMap)
     )
     return completeComputation(nextToken, CompletedReason.SUCCEEDED)
   }
--- a/src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/liquidlegionsv2/LiquidLegionsV2Mill.kt
+++ b/src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/liquidlegionsv2/LiquidLegionsV2Mill.kt
@@ -970,7 +970,9 @@ class LiquidLegionsV2Mill(
       token.computationDetails.liquidLegionsV2.parameters.maximumFrequency
     val measurementSpec =
       MeasurementSpec.parseFrom(token.computationDetails.kingdomComputation.measurementSpec)
-    val measurementSpecMaximumFrequency = measurementSpec.reachAndFrequency.maximumFrequencyPerUser
+    val measurementSpecMaximumFrequency =
+      if (measurementSpec.hasReach()) 1
+      else measurementSpec.reachAndFrequency.maximumFrequencyPerUser
     if (measurementSpecMaximumFrequency > 0) {
       maximumRequestedFrequency = min(maximumRequestedFrequency, measurementSpecMaximumFrequency)
     }
--- a/src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/liquidlegionsv2/LiquidLegionsV2Mill.kt
+++ b/src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/liquidlegionsv2/LiquidLegionsV2Mill.kt
@@ -45,12 +45,13 @@ import org.wfanet.measurement.duchy.daemon.mill.Certificate
 import org.wfanet.measurement.duchy.daemon.mill.MillBase
 import org.wfanet.measurement.duchy.daemon.mill.PermanentComputationError
 import org.wfanet.measurement.duchy.daemon.mill.liquidlegionsv2.crypto.LiquidLegionsV2Encryption
+import org.wfanet.measurement.duchy.daemon.utils.ComputationResult
+import org.wfanet.measurement.duchy.daemon.utils.Reach
 import org.wfanet.measurement.duchy.daemon.utils.ReachAndFrequency
 import org.wfanet.measurement.duchy.daemon.utils.toAnySketchElGamalPublicKey
 import org.wfanet.measurement.duchy.daemon.utils.toCmmsElGamalPublicKey
 import org.wfanet.measurement.duchy.daemon.utils.toV2AlphaElGamalPublicKey
 import org.wfanet.measurement.duchy.daemon.utils.toV2AlphaEncryptionPublicKey
-import org.wfanet.measurement.duchy.daemon.utils.toV2AlphaMeasurementResult
 import org.wfanet.measurement.duchy.db.computation.ComputationDataClients
 import org.wfanet.measurement.duchy.service.internal.computations.outputPathList
 import org.wfanet.measurement.duchy.service.system.v1alpha.advanceComputationHeader
@@ -645,14 +646,15 @@ class LiquidLegionsV2Mill(
     }
     var reach = 0L
     val maximumRequestedFrequency = getMaximumRequestedFrequency(token)
+    val measurementSpec =
+      MeasurementSpec.parseFrom(token.computationDetails.kingdomComputation.measurementSpec)
+
     val (bytes, tempToken) =
       existingOutputOr(token) {
         when (Version.fromString(token.computationDetails.kingdomComputation.publicApiVersion)) {
           Version.V2_ALPHA -> {}
           Version.VERSION_UNSPECIFIED -> error("Public api version is invalid or unspecified.")
         }
-        val measurementSpec =
-          MeasurementSpec.parseFrom(token.computationDetails.kingdomComputation.measurementSpec)
         val requestBuilder =
           CompleteExecutionPhaseTwoAtAggregatorRequest.newBuilder().apply {
             localElGamalKeyPair = llv2Details.localElgamalKey
@@ -716,8 +718,8 @@ class LiquidLegionsV2Mill(
       }
 
     // If this is a reach-only computation, then our job is done.
-    if (maximumRequestedFrequency == 1) {
-      sendResultToKingdom(token, reach, mapOf(1L to 1.0))
+    if (measurementSpec.hasReach() || maximumRequestedFrequency == 1) {
+      sendResultToKingdom(token, Reach(reach))
       return completeComputation(nextToken, CompletedReason.SUCCEEDED)
     }
 
@@ -786,7 +788,10 @@ class LiquidLegionsV2Mill(
       stub = nextDuchyStub(llv2Details.participantList)
     )
 
-    return if (maximumRequestedFrequency == 1) {
+    val measurementSpec =
+      MeasurementSpec.parseFrom(token.computationDetails.kingdomComputation.measurementSpec)
+
+    return if (measurementSpec.hasReach() || maximumRequestedFrequency == 1) {
       // If this is a reach-only computation, then our job is done.
       completeComputation(nextToken, CompletedReason.SUCCEEDED)
     } else {
@@ -835,7 +840,10 @@ class LiquidLegionsV2Mill(
       CompleteExecutionPhaseThreeAtAggregatorResponse.parseFrom(bytes.flatten())
         .frequencyDistributionMap
 
-    sendResultToKingdom(token, llv2Details.reachEstimate.reach, frequencyDistributionMap)
+    sendResultToKingdom(
+      token,
+      ReachAndFrequency(llv2Details.reachEstimate.reach, frequencyDistributionMap)
+    )
     return completeComputation(nextToken, CompletedReason.SUCCEEDED)
   }
 
@@ -883,16 +891,14 @@ class LiquidLegionsV2Mill(
 
   private suspend fun sendResultToKingdom(
     token: ComputationToken,
-    reach: Long,
-    frequency: Map<Long, Double>
+    computationResult: ComputationResult
   ) {
-    val reachAndFrequency = ReachAndFrequency(reach, frequency)
     val kingdomComputation = token.computationDetails.kingdomComputation
     val serializedPublicApiEncryptionPublicKey: ByteString
     val encryptedResult =
       when (Version.fromString(kingdomComputation.publicApiVersion)) {
         Version.V2_ALPHA -> {
-          val signedResult = signResult(reachAndFrequency.toV2AlphaMeasurementResult(), signingKey)
+          val signedResult = signResult(computationResult.toV2AlphaMeasurementResult(), signingKey)
           val publicApiEncryptionPublicKey =
             kingdomComputation.measurementPublicKey.toV2AlphaEncryptionPublicKey()
           serializedPublicApiEncryptionPublicKey = publicApiEncryptionPublicKey.toByteString()
--- a/src/main/kotlin/org/wfanet/measurement/duchy/daemon/utils/ComputationConversions.kt
+++ b/src/main/kotlin/org/wfanet/measurement/duchy/daemon/utils/ComputationConversions.kt
@@ -260,9 +260,9 @@ interface ComputationResult {
 }
 
 /** The result and frequency estimation of a computation. */
-data class ReachAndFrequency(val reach: Long, val frequency: Map<Long, Double>) :
+data class ReachAndFrequencyResult(val reach: Long, val frequency: Map<Long, Double>) :
   ComputationResult {
-  /** Converts a ReachAndFrequency object to the v2Alpha measurement result. */
+  /** Converts a ReachAndFrequencyResult object to the v2Alpha measurement result. */
   override fun toV2AlphaMeasurementResult(): Measurement.Result {
     val source = this
     return MeasurementKt.result {
@@ -272,9 +272,9 @@ data class ReachAndFrequency(val reach: Long, val frequency: Map<Long, Double>) 
   }
 }
 
-data class Reach(val reach: Long) : ComputationResult {
+data class ReachResult(val reach: Long) : ComputationResult {
 
-  /** Converts a Reach object to the v2Alpha measurement result. */
+  /** Converts a ReachResult object to the v2Alpha measurement result. */
   override fun toV2AlphaMeasurementResult(): Measurement.Result {
     val source = this
     return MeasurementKt.result { reach = reach { value = source.reach } }
--- a/src/main/kotlin/org/wfanet/measurement/duchy/daemon/utils/ComputationConversions.kt
+++ b/src/main/kotlin/org/wfanet/measurement/duchy/daemon/utils/ComputationConversions.kt
@@ -255,15 +255,29 @@ fun V2AlphaElGamalPublicKey.toDuchyElGamalPublicKey(): ElGamalPublicKey {
   }
 }
 
-/** The result and frequency estimation of a computation. */
-data class ReachAndFrequency(val reach: Long, val frequency: Map<Long, Double>)
+interface ComputationResult {
+  fun toV2AlphaMeasurementResult(): Measurement.Result
+}
 
-/** Converts a ReachAndFrequency object to the v2Alpha measurement result. */
-fun ReachAndFrequency.toV2AlphaMeasurementResult(): Measurement.Result {
-  val source = this
-  return MeasurementKt.result {
-    reach = reach { value = source.reach }
-    frequency = frequency { relativeFrequencyDistribution.putAll(source.frequency) }
+/** The result and frequency estimation of a computation. */
+data class ReachAndFrequency(val reach: Long, val frequency: Map<Long, Double>) :
+  ComputationResult {
+  /** Converts a ReachAndFrequency object to the v2Alpha measurement result. */
+  override fun toV2AlphaMeasurementResult(): Measurement.Result {
+    val source = this
+    return MeasurementKt.result {
+      reach = reach { value = source.reach }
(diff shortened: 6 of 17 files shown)


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/911)

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/911)* status: 6 of 7 files reviewed, 2 unresolved discussions (waiting on @riemanli)

*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/911#-NREWnqH1TMWU6JVV007:-NREWnqI6Mwrg-c77uWv:b-71h13o) (no related file):*
Where are we ensuring that `Measurement.Result` has no `frequency` value set when the measurement type is `REACH`?

___
___
*[`build/repositories.bzl` line 40 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/911#-NREWKNQ0JRxEviZkM-I:-NREWKNQ0JRxEviZkM-J:b-wmyxk2) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0ad502bac2217458cedc914abb7a419f8c334c37/build/repositories.bzl#L40)):*
> ```Starlark
>     wfa_repo_archive(
>         name = "wfa_measurement_proto",
>         commit = "265e1ac06d4d793f6debfdfec84201b59f11642b",
> ```

Add a DO_NOT_SUBMIT comment in TODO format to switch this to version.

riemanli (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/911)* status: 6 of 7 files reviewed, 2 unresolved discussions (waiting on @SanjayVas)

*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/911#-NREWnqH1TMWU6JVV007:-NREjVew8f9cwMnB4a4V:b-xqzfho) (no related file):*

Where are we ensuring that `Measurement.Result` has no `frequency` value set when the measurement type is `REACH`?
</blockquote></details>

That is a great question. Currently, even it's reach-only, the [LLV2 Mill](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/main/src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/liquidlegionsv2/LiquidLegionsV2Mill.kt#L720) still sends the result with a frequency map of size "1". We could either follow this or put additional logic to `sendResultToKingdom` function to enforce no `frequency`.

___
___
*[`build/repositories.bzl` line 40 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/911#-NREWKNQ0JRxEviZkM-I:-NREhmcW9WF2c5pUjkCH:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0ad502bac2217458cedc914abb7a419f8c334c37/build/repositories.bzl#L40)):*

Add a DO_NOT_SUBMIT comment in TODO format to switch this to version.
</blockquote></details>

Done.

SanjayVas (changes requested):
> Reviewed 1 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/911)* status: 6 of 9 files reviewed, 2 unresolved discussions (waiting on @riemanli)

*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/911#-NREWnqH1TMWU6JVV007:-NREmUfSE_SB5CoU8dMF:befkxv8) (no related file):*

That is a great question. Currently, even it's reach-only, the [LLV2 Mill](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/main/src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/liquidlegionsv2/LiquidLegionsV2Mill.kt#L720) still sends the result with a frequency map of size "1". We could either follow this or put additional logic to `sendResultToKingdom` function to enforce no `frequency`.
</blockquote></details>

The idea is that this is intended to eventually replace the hack of using R/F with max frequency of 1 and just ignoring the result frequency. We still need to support that behavior, but the point of the `REACH` type is being explicit that the result will not include a frequency. The fact that we're doing R/F LLv2 with max frequency of 1 under the hood is an implementation detail.

___
___
*[`build/repositories.bzl` line 40 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/911#-NREWKNQ0JRxEviZkM-I:-NREm7P963wViuiMQpBp:b-5npwuh) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0ad502bac2217458cedc914abb7a419f8c334c37/build/repositories.bzl#L40)):*

Done.
</blockquote></details>

What I meant by using TODO format is

```
# DO_NOT_SUBMIT(world-federation-of-advertisers/cross-media-measurement-api#126): Use version once released.
```

riemanli (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/911)* status: 3 of 15 files reviewed, 2 unresolved discussions (waiting on @SanjayVas)

___
*[`build/repositories.bzl` line 40 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/911#-NREWKNQ0JRxEviZkM-I:-NRigQCc-2klF4yfphUC:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0ad502bac2217458cedc914abb7a419f8c334c37/build/repositories.bzl#L40)):*

What I meant by using TODO format is

```
# DO_NOT_SUBMIT(world-federation-of-advertisers/cross-media-measurement-api#126): Use version once released.
```
</blockquote></details>

Done.

tristanvuong2021 (changes requested):
> Reviewed 3 of 7 files at r1, 1 of 3 files at r2, 9 of 10 files at r3, 3 of 3 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/911)* status: 16 of 17 files reviewed, 4 unresolved discussions (waiting on @iverson52000, @renjiezh, @riemanli, @SanjayVas, and @uakyol)

___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/MeasurementsService.kt` line 272 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/911#-NRtO30r0FAfD8uKG0wP:-NRtO30r0FAfD8uKG0wQ:b-ju7zt7) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/989c3fce69c6e3632d546775a61c8a7cb504a2c6/src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/MeasurementsService.kt#L272)):*
> ```Kotlin
>       }
> 
>       val vidSamplingInterval = vidSamplingInterval
> ```

This is unnecessary. Should fix it below too.
___
*[`src/test/kotlin/org/wfanet/measurement/duchy/daemon/herald/HeraldTest.kt` line 155 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/911#-NRtMUGKC_BxXfDt9QUU:-NRtM14zEI5LdsHhztN-:b91hp2t) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/989c3fce69c6e3632d546775a61c8a7cb504a2c6/src/test/kotlin/org/wfanet/measurement/duchy/daemon/herald/HeraldTest.kt#L155)):*
> ```Kotlin
> private val PUBLIC_API_REACH_ONLY_MEASUREMENT_SPEC =
>   MeasurementSpec.newBuilder()
>     .apply {
> ```

Use the other builder instead of builder/apply in new code.

renjiezh (approved):
> Reviewed 3 of 7 files at r1, 2 of 3 files at r2, 9 of 10 files at r3, 3 of 3 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/911)* status: all files reviewed, 6 unresolved discussions (waiting on @iverson52000, @riemanli, @SanjayVas, @tristanvuong2021, and @uakyol)

___
*[`build/repositories.bzl` line 42 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/911#-NRsZL6k4IWxfn7YmsHo:-NRsZL6k4IWxfn7YmsHp:b-4j6axn) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/989c3fce69c6e3632d546775a61c8a7cb504a2c6/build/repositories.bzl#L42)):*
> ```Starlark
>         repo = "cross-media-measurement-api",
>         sha256 = "cfde44921d11ccc04e22d4dd7a99b45751a3c3eb53e0cb3ba454100ffa616d27",
>         version = "0.28.0",
> ```

nip: 0.28.1 has been released if you want to keep this the latest. It should require no change to your code.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/daemon/utils/ComputationConversions.kt` line 275 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/911#-NRtVATSAJrpk7NuNAuN:-NRtVATSAJrpk7NuNAuO:bkrapj2) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/989c3fce69c6e3632d546775a61c8a7cb504a2c6/src/main/kotlin/org/wfanet/measurement/duchy/daemon/utils/ComputationConversions.kt#L275)):*
> ```Kotlin
> }
> 
> data class Reach(val reach: Long) : ComputationResult {
> ```

nit: What about renaming it to `ReachResult`?  As well as `ReachAndFrequencyResult`.
___
*[`src/test/kotlin/org/wfanet/measurement/duchy/daemon/herald/HeraldTest.kt` line 155 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/911#-NRtMUGKC_BxXfDt9QUU:-NRt_jva4Wo8R8nKCIP0:bbfceuj) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/989c3fce69c6e3632d546775a61c8a7cb504a2c6/src/test/kotlin/org/wfanet/measurement/duchy/daemon/herald/HeraldTest.kt#L155)):*

Use the other builder instead of builder/apply in new code.
</blockquote></details>

Yeah. Please replace old code by DSL as you can. Appreciate it!

riemanli (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/911)* status: 13 of 17 files reviewed, 4 unresolved discussions (waiting on @iverson52000, @renjiezh, @SanjayVas, @tristanvuong2021, and @uakyol)

___
*[`build/repositories.bzl` line 42 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/911#-NRsZL6k4IWxfn7YmsHo:-NRu7K-j7sgq_HOolk2N:belsklq) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/989c3fce69c6e3632d546775a61c8a7cb504a2c6/build/repositories.bzl#L42)):*

nip: 0.28.1 has been released if you want to keep this the latest. It should require no change to your code.
</blockquote></details>

Thanks for letting me know!
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/daemon/utils/ComputationConversions.kt` line 275 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/911#-NRtVATSAJrpk7NuNAuN:-NRu9jx49XzlhfZVckSN:bad59sg) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/989c3fce69c6e3632d546775a61c8a7cb504a2c6/src/main/kotlin/org/wfanet/measurement/duchy/daemon/utils/ComputationConversions.kt#L275)):*

nit: What about renaming it to `ReachResult`?  As well as `ReachAndFrequencyResult`.
</blockquote></details>

Thanks for the suggestion. Done
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/MeasurementsService.kt` line 272 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/911#-NRtO30r0FAfD8uKG0wP:-NRuApq17347kKJkboV5:bfhz2u7) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/989c3fce69c6e3632d546775a61c8a7cb504a2c6/src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/MeasurementsService.kt#L272)):*

This is unnecessary. Should fix it below too.
</blockquote></details>

Thanks for catching it. Done.
___
*[`src/test/kotlin/org/wfanet/measurement/duchy/daemon/herald/HeraldTest.kt` line 155 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/911#-NRtMUGKC_BxXfDt9QUU:-NRuAx654j10W1RL_fXn:b-ugd4pq) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/989c3fce69c6e3632d546775a61c8a7cb504a2c6/src/test/kotlin/org/wfanet/measurement/duchy/daemon/herald/HeraldTest.kt#L155)):*

Yeah. Please replace old code by DSL as you can. Appreciate it!
</blockquote></details>

Thanks for bringing this up. I will modify the one added by me and try my best to update the existing ones.

riemanli (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/911)* status: 10 of 17 files reviewed, 4 unresolved discussions (waiting on @iverson52000, @renjiezh, @SanjayVas, @tristanvuong2021, and @uakyol)

*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/911#-NREWnqH1TMWU6JVV007:-NRuZd8G5mmCdUmxpqB0:b-896fix) (no related file):*

The idea is that this is intended to eventually replace the hack of using R/F with max frequency of 1 and just ignoring the result frequency. We still need to support that behavior, but the point of the `REACH` type is being explicit that the result will not include a frequency. The fact that we're doing R/F LLv2 with max frequency of 1 under the hood is an implementation detail.
</blockquote></details>

Done.

___

SanjayVas (changes requested):
> Reviewed 5 of 10 files at r3, 2 of 3 files at r4, 4 of 4 files at r5, 2 of 3 files at r6, 1 of 1 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/911)* status: all files reviewed, 5 unresolved discussions (waiting on @iverson52000, @riemanli, and @uakyol)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt` line 739 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/911#-NS7LYl50SYxVvbthy7y:-NS7LYl50SYxVvbthy7z:b-4oiagi) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5e6ba9c003ef739bb2e3068ae662efb5753830ed/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt#L739)):*
> ```Kotlin
> 
>     logger.info("Generating publisher noise for direct reach and frequency...")
>     val publisherNoise = generatePublisherNoise(measurementSpec, frequencyMap.keys.toList())
> ```

Does the function need to take in a `List`, or would e.g. an `Iterable` be sufficient such that we could avoid this conversion?



_Code quote:_
```Kotlin
frequencyMap.keys.toList()
```
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt` line 909 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/911#-NS7Mzpp4QBHxtliLGWJ:-NS7MzpqCVsUefEHXvZi:b-fr50l) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5e6ba9c003ef739bb2e3068ae662efb5753830ed/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt#L909)):*
> ```Kotlin
>             // Reseed random number generator so the results can be verified in tests.
>             // TODO(@iverson52000): make randomSeed an input once create noiser class.
>             laplaceForReach.reseedRandomGenerator(1)
> ```

You'll likely need to reconcile this with #904, as that changes how we handle seeding the RNG.
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/frontend/FrontendSimulator.kt` line 63 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/911#-NS7NT_Y9eSr87v-l890:-NS7NT_Y9eSr87v-l891:b-xztp1h) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5e6ba9c003ef739bb2e3068ae662efb5753830ed/src/main/kotlin/org/wfanet/measurement/loadtest/frontend/FrontendSimulator.kt#L63)):*
> ```Kotlin
> import org.wfanet.measurement.api.v2alpha.MeasurementSpecKt.duration
> import org.wfanet.measurement.api.v2alpha.MeasurementSpecKt.impression
> import org.wfanet.measurement.api.v2alpha.MeasurementSpecKt.reach as reachSpec
> ```

This could trigger https://youtrack.jetbrains.com/issue/KTIJ-16471/

Avoid import aliases for functions on objects/classes. If you want to add an import alias, do so for the enclosing type.



_Suggestion:_
```Kotlin
import org.wfanet.measurement.api.v2alpha.MeasurementSpecKt as MeasurementSpecs
```
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/frontend/FrontendSimulator.kt` line 426 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/911#-NS7O7arDRhxO_DvSfHw:-NS7O7arDRhxO_DvSfHx:b-wuxyew) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5e6ba9c003ef739bb2e3068ae662efb5753830ed/src/main/kotlin/org/wfanet/measurement/loadtest/frontend/FrontendSimulator.kt#L426)):*
> ```Kotlin
>   private suspend fun getReachResult(measurementName: String): Result? {
>     val measurement = getMeasurement(measurementName)
>     logger.info("Current Measurement state is: " + measurement.state)
> ```

Extract this block, as it's repeated multiple times in this file. Maybe move the "current state" logging into `getMeasurement` and add a `checkMeasurementNotFailed` function?



_Code quote:_
```Kotlin
    val measurement = getMeasurement(measurementName)
    logger.info("Current Measurement state is: " + measurement.state)
    if (measurement.state == Measurement.State.FAILED) {
      val failure: Failure = measurement.failure
      throw Exception("Measurement failed with reason ${failure.reason}: ${failure.message}")
    }
```
___
*[`src/test/kotlin/org/wfanet/measurement/duchy/daemon/herald/HeraldTest.kt` line 450 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/911#-NS7Pfa13QxpugxIpkL-:-NS7Pfa13QxpugxIpkL0:b-k2cw16) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5e6ba9c003ef739bb2e3068ae662efb5753830ed/src/test/kotlin/org/wfanet/measurement/duchy/daemon/herald/HeraldTest.kt#L450)):*
> ```Kotlin
>         }
>       )
>   }
> ```

nit: blank line between function definitions

uakyol (changes requested):
> Reviewed 3 of 7 files at r1, 3 of 10 files at r3, 1 of 3 files at r4, 4 of 4 files at r5, 2 of 3 files at r6, 1 of 1 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/911)* status: all files reviewed, 7 unresolved discussions (waiting on @iverson52000 and @riemanli)

___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/MeasurementsService.kt` line 276 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/911#-NS7ogFt52iRzkprknlb:-NS7ogFt52iRzkprknlc:b-p0ltxp) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/de0c8df34d21704b98fb371f209fd9aac4ac1083/src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/MeasurementsService.kt#L276)):*
> ```Kotlin
>       }
> 
>       val vidSamplingInterval = vidSamplingInterval
> ```

what happened to this val? Was it not being used?
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt` line 140 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/911#-NS7xUFS9kv5fh2noS5B:-NS7xUFS9kv5fh2noS5C:b-e9gypw) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5e6ba9c003ef739bb2e3068ae662efb5753830ed/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt#L140)):*
> ```Kotlin
> )
> 
> data class PublisherNoise(
> ```

IIUC, you are using this only during populating the Measurement Result. While this data class might be useful if the objects were passed around and thus results  of calculations were passed around I see couple reasons to just extract the c

1) Member variables are not always all populated and when not used (like duration) are defaulted. This is confusing and not a good practice to operate on objects with fields that have not meaningful fields
2) As stated above, noising is not redone and thus not saving anything
3) In fact, after noising for frequency you iterate one more time over the map which can be avoided.

I would prefer to just extract functions to calculate the measurement result and put the noising logic in them. Wdyt? I might be missing something.

uakyol (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/911)* status: all files reviewed, 6 unresolved discussions (waiting on @iverson52000 and @riemanli)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt` line 140 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/911#-NS7xUFS9kv5fh2noS5B:-NS8533iATbs1nqUxJiV:bvx3thd) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5e6ba9c003ef739bb2e3068ae662efb5753830ed/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt#L140)):*

IIUC, you are using this only during populating the Measurement Result. While this data class might be useful if the objects were passed around and thus results  of calculations were passed around I see couple reasons to just extract the c

1) Member variables are not always all populated and when not used (like duration) are defaulted. This is confusing and not a good practice to operate on objects with fields that have not meaningful fields
2) As stated above, noising is not redone and thus not saving anything
3) In fact, after noising for frequency you iterate one more time over the map which can be avoided.

I would prefer to just extract functions to calculate the measurement result and put the noising logic in them. Wdyt? I might be missing something. 
</blockquote></details>

Sorry, something went wrong editing the above comment it seems, here is the correct version :  

IIUC, you are using this only during populating the Measurement Result. While this data class might be useful if the objects were passed around and thus results of calculations were cached,  I see couple reasons to just extract it in to couple of functions: 

1. Member variables are not always all populated and when not used (like duration) are defaulted. This is confusing and not a good practice to operate on objects with fields that have not meaningful fields
2. As stated above, noising is not redone and thus not saving anything
3. In fact, after noising for frequency you iterate one more time over the map which can be avoided.

I would prefer to just extract functions to calculate the measurement result and put the noising logic in them. Wdyt? I might be missing something.

riemanli (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/911)* status: all files reviewed, 5 unresolved discussions (waiting on @iverson52000, @SanjayVas, and @uakyol)

___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/MeasurementsService.kt` line 276 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/911#-NS7ogFt52iRzkprknlb:-NS80Q634jEjS-rAcbmq:b-nqw123) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/de0c8df34d21704b98fb371f209fd9aac4ac1083/src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/MeasurementsService.kt#L276)):*

what happened to this val? Was it not being used?
</blockquote></details>

Yes. The only place it was used is

``` kotlin
grpcRequire(vidSamplingInterval.width > 0) { "Vid sampling interval is unspecified" }
```

which doesn't need to create a variable for doing it.
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt` line 739 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/911#-NS7LYl50SYxVvbthy7y:-NS7mXZq6HcwDLBoPVtb:brydz6g) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5e6ba9c003ef739bb2e3068ae662efb5753830ed/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt#L739)):*

Does the function need to take in a `List`, or would e.g. an `Iterable` be sufficient such that we could avoid this conversion?
</blockquote></details>

Good point! Changed to `Iterable`.
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt` line 909 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/911#-NS7Mzpp4QBHxtliLGWJ:-NS7o-k96rn4Flky4H6W:b-p6uno3) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5e6ba9c003ef739bb2e3068ae662efb5753830ed/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt#L909)):*

You'll likely need to reconcile this with #904, as that changes how we handle seeding the RNG.
</blockquote></details>

Thanks for bringing this up. I synced up with @iverson52000 for this issue. We decided to let #904 merged, and I will resolve the conflicts.
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/frontend/FrontendSimulator.kt` line 63 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/911#-NS7NT_Y9eSr87v-l890:-NS8-cu62zYtJHgjGA4Q:bmqwo8h) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5e6ba9c003ef739bb2e3068ae662efb5753830ed/src/main/kotlin/org/wfanet/measurement/loadtest/frontend/FrontendSimulator.kt#L63)):*

This could trigger https://youtrack.jetbrains.com/issue/KTIJ-16471/

Avoid import aliases for functions on objects/classes. If you want to add an import alias, do so for the enclosing type.
</blockquote></details>

That's good to know! Modified.
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/frontend/FrontendSimulator.kt` line 426 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/911#-NS7O7arDRhxO_DvSfHw:-NS848ewDglZ_VTBo9oW:b-obdf53) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5e6ba9c003ef739bb2e3068ae662efb5753830ed/src/main/kotlin/org/wfanet/measurement/loadtest/frontend/FrontendSimulator.kt#L426)):*

Extract this block, as it's repeated multiple times in this file. Maybe move the "current state" logging into `getMeasurement` and add a `checkMeasurementNotFailed` function?
</blockquote></details>

I kept `getMeasurement` as what it is but create another function `getNotFailedMeasurement` to include the code block as `getFailure` needs to get a failed `Measurement`.
___
*[`src/test/kotlin/org/wfanet/measurement/duchy/daemon/herald/HeraldTest.kt` line 450 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/911#-NS7Pfa13QxpugxIpkL-:-NS84PVl5Oj95FLcZNrp:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5e6ba9c003ef739bb2e3068ae662efb5753830ed/src/test/kotlin/org/wfanet/measurement/duchy/daemon/herald/HeraldTest.kt#L450)):*

nit: blank line between function definitions
</blockquote></details>

Done.

SanjayVas (changes requested):
> Reviewed 3 of 3 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/911)* status: all files reviewed, 2 unresolved discussions (waiting on @iverson52000, @riemanli, and @uakyol)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/frontend/FrontendSimulator.kt` line 426 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/911#-NS7O7arDRhxO_DvSfHw:-NS8D6jO2XmI8-KzfS19:b-44xsbo) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5e6ba9c003ef739bb2e3068ae662efb5753830ed/src/main/kotlin/org/wfanet/measurement/loadtest/frontend/FrontendSimulator.kt#L426)):*

I kept `getMeasurement` as what it is but create another function `getNotFailedMeasurement` to include the code block as `getFailure` needs to get a failed `Measurement`.
</blockquote></details>

The reason I suggested moving the logging for current state is that I believe every call to `getMeasurement` is followed by logging the current state. I had meant the check to be a separate function that's called on the result of `getMeasurement`.

e.g.

```kotlin
private fun checkNotFailed(measurement: Measurement): Measurement {
  check(state != Measurement.State.FAILED) {
    val failure: Failure = measurement.failure
    "Measurement failed with reason ${failure.reason}: ${failure.message}"
  }
  return measurement
}
```

```kotlin
val measurement = checkNotFailed(getMeasurement(measurementName))
```

riemanli (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/911)* status: all files reviewed, 2 unresolved discussions (waiting on @iverson52000 and @uakyol)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt` line 140 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/911#-NS7xUFS9kv5fh2noS5B:-NS89uh8DmyHSxu8br7o:b-vivxga) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5e6ba9c003ef739bb2e3068ae662efb5753830ed/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt#L140)):*
Thanks for the feedback! The main reason I had this modification is because the function `addPublisherNoise` is specifically for `reachAndFrequency` only, which is not compatible if we have impression or duration in the future. In addition, we already had `fulfill*Measurement` those measurement-type specific functions for each measurement type. Using measurement-type specific functions also for generating noise seems a bit too much for me.

> Member variables are not always all populated and when not used (like duration) are defaulted. This is confusing and not a good practice to operate on objects with fields that have not meaningful fields

I can create an interface `PublisherNoise` and create child noise classes for each measurement type. This way there is no default values for unused noises.

If you insist, I can also fall back to the original design. Let me know what you think. Thanks!

riemanli (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/911)* status: 16 of 17 files reviewed, 2 unresolved discussions (waiting on @iverson52000, @SanjayVas, and @uakyol)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/frontend/FrontendSimulator.kt` line 426 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/911#-NS7O7arDRhxO_DvSfHw:-NS8IJFB2R5GZj9wZWNx:b73p2f7) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5e6ba9c003ef739bb2e3068ae662efb5753830ed/src/main/kotlin/org/wfanet/measurement/loadtest/frontend/FrontendSimulator.kt#L426)):*

The reason I suggested moving the logging for current state is that I believe every call to `getMeasurement` is followed by logging the current state. I had meant the check to be a separate function that's called on the result of `getMeasurement`.

e.g.

```kotlin
private fun checkNotFailed(measurement: Measurement): Measurement {
  check(state != Measurement.State.FAILED) {
    val failure: Failure = measurement.failure
    "Measurement failed with reason ${failure.reason}: ${failure.message}"
  }
  return measurement
}
```

```kotlin
val measurement = checkNotFailed(getMeasurement(measurementName))
```
</blockquote></details>

Got it. Thanks for correcting. Modified.

SanjayVas (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/911)* status: all files reviewed, 1 unresolved discussion (waiting on @iverson52000 and @uakyol)

uakyol (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/911)* status: all files reviewed, 1 unresolved discussion (waiting on @iverson52000)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt` line 140 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/911#-NS7xUFS9kv5fh2noS5B:-NS9OY0W4o9cHroLRid_:b-7eo5gi) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5e6ba9c003ef739bb2e3068ae662efb5753830ed/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt#L140)):*

Thanks for the feedback! The main reason I had this modification is because the function `addPublisherNoise` is specifically for `reachAndFrequency` only, which is not compatible if we have impression or duration in the future. In addition, we already had `fulfill*Measurement` those measurement-type specific functions for each measurement type. Using measurement-type specific functions also for generating noise seems a bit too much for me.

> Member variables are not always all populated and when not used (like duration) are defaulted. This is confusing and not a good practice to operate on objects with fields that have not meaningful fields

I can create an interface `PublisherNoise` and create child noise classes for each measurement type. This way there is no default values for unused noises.

If you insist, I can also fall back to the original design. Let me know what you think. Thanks!
</blockquote></details>

 Thanks Rieman.

If you can make it work with the interface + child noise classes it would be great but I think you will have to type cast after you return based on measurement type, which again is not good. 
I think the clean way to do this is actually creating treating each type differently  with helper functions and they can share code. Here is what I am thinking :  



_Code snippet:_
```Kotlin
val requisitionData = 
when (measurementSpec.measurementTypeCase) {
          MeasurementSpec.MeasurementTypeCase.REACH_AND_FREQUENCY -> {
          val (sampledReachValue, frequencyMap) = calculateDirectReachAndFrequency(vidList)
          // addReachAndFreqPublisherNoise is the same as addPublisherNoise from before, just passing in 2 epsilons rather than the 
          // reachAndFrequency message for code sharing
          val (sampledNoisedReachValue, noisedFrequencyMap) = addReachAndFreqPublisherNoise(sampledReachValue, frequencyMap, measurementSpec.reachAndFrequency.reach.epsilon, measurementSpec.reachAndFrequency frequency.epsiloon )
          val scaledNoisedReachValue = (sampledNoisedReachValue / vidSamplingIntervalWidth).toLong()
           MeasurementKt.result {
              reach = reach { value = scaledNoisedReachValue }
              frequency = frequency { relativeFrequencyDistribution.putAll(noisedFrequencyMap) }
            }
          }
          MeasurementSpec.MeasurementTypeCase.IMPRESSION,
          MeasurementSpec.MeasurementTypeCase.DURATION -> {
             error("Measurement type not supported.")
          }
          MeasurementSpec.MeasurementTypeCase.REACH -> {
            MeasurementKt.result {
              reach = reach {
              // addReachPublisherNoise is a new function since we pass in epsilon, we can code share with addReachAndFreqPublisherNoise function.
                val sampledNoisedReachValue = addReachPublisherNoise(sampledReachValue, frequencyMap, measurementSpec.reach.epsilon)
                val scaledNoisedReachValue = (sampledNoisedReachValue / vidSamplingIntervalWidth).toLong()
                value = scaledNoisedReachValue
              }
            }
          }
          MeasurementSpec.MeasurementTypeCase.MEASUREMENTTYPE_NOT_SET -> {
            error("Measurement type not set.")
          }
        }
```

riemanli (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/911)* status: 16 of 17 files reviewed, 1 unresolved discussion (waiting on @iverson52000, @SanjayVas, and @uakyol)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt` line 140 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/911#-NS7xUFS9kv5fh2noS5B:-NSDtPKx5FEX_BVPduq6:b-1znw5r) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5e6ba9c003ef739bb2e3068ae662efb5753830ed/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt#L140)):*

 Thanks Rieman.

If you can make it work with the interface + child noise classes it would be great but I think you will have to type cast after you return based on measurement type, which again is not good. 
I think the clean way to do this is actually creating treating each type differently  with helper functions and they can share code. Here is what I am thinking :  
</blockquote></details>

Thanks for the suggestion! I incorporated your idea and made the changes. Please take a look and let me know what you think. Thanks!

cc: @iverson52000 for #904.

uakyol (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/911)* status: 16 of 17 files reviewed, 1 unresolved discussion (waiting on @iverson52000, @riemanli, and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt` line 140 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/911#-NS7xUFS9kv5fh2noS5B:-NSE7GesApI8ZYqFAzBY:bcwjqvl) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5e6ba9c003ef739bb2e3068ae662efb5753830ed/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt#L140)):*

Thanks for the suggestion! I incorporated your idea and made the changes. Please take a look and let me know what you think. Thanks!

cc: @iverson52000 for #904.
</blockquote></details>

looks great Rieman!
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt` line 749 at r11](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/911#-NSE7ROH4Sjv1kvFgAYt:-NSE7ROH4Sjv1kvFgAYu:bmolyve) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0191a64498e1614a600d5dc1f293072ab188721e/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt#L749)):*
> ```Kotlin
>     return when (measurementSpec.measurementTypeCase) {
>       MeasurementSpec.MeasurementTypeCase.REACH_AND_FREQUENCY -> {
>         val (sampledReachValue, frequencyMap) = calculateDirectReachAndFrequency(vidList)
> ```

would you consider splitting calculateDirectReachAndFrequency function to `calculateDirectReach` and  `calculateFrequency` functions? It seems they can be completely separated without any computational cost. 

this way you don't have to unnecessarily calculate and then  ignore the frequencyMap value in the reach only case below.

uakyol:
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt` line 749 at r11](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/911#-NSE7ROH4Sjv1kvFgAYt:-NSE83EJ1bdVI6SYHvju:b-bxeybd) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0191a64498e1614a600d5dc1f293072ab188721e/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt#L749)):*

would you consider splitting calculateDirectReachAndFrequency function to `calculateDirectReach` and  `calculateFrequency` functions? It seems they can be completely separated without any computational cost. 

this way you don't have to unnecessarily calculate and then  ignore the frequencyMap value in the reach only case below.

</blockquote></details>

sorry, I meant `calculateDirectFrequency`

uakyol:
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt` line 740 at r11](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/911#-NSE8IMIC-hTTs4L4jge:-NSE8IMIC-hTTs4L4jgf:b-n9lr0a) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0191a64498e1614a600d5dc1f293072ab188721e/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt#L740)):*
> ```Kotlin
>    * @param measurementSpec Measurement spec.
>    * @param vidList List of VIDs.
>    * @return Noised reach value.
> ```

can you change this to. @return MeasurementResult.

uakyol (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/911)* status: 16 of 17 files reviewed, 4 unresolved discussions (waiting on @iverson52000, @riemanli, and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt` line 730 at r11](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/911#-NSE8W_p5bHPG8yFHjr0:-NSE8W_p5bHPG8yFHjr1:b-scc6tl) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0191a64498e1614a600d5dc1f293072ab188721e/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt#L730)):*
> ```Kotlin
>         }
> 
>     val requisitionData = buildMeasurementResult(measurementSpec, vidList)
> ```

lets rename this val to measurementResult here and in other places?
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt` line 831 at r11](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/911#-NSE8hkW9wQAQpBcPaVV:-NSE8hkW9wQAQpBcPaVW:b91fo7j) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0191a64498e1614a600d5dc1f293072ab188721e/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt#L831)):*
> ```Kotlin
>     requisitionSpec: RequisitionSpec,
>     measurementSpec: MeasurementSpec,
>     requisitionData: Measurement.Result
> ```

while we are here, would you mind changing this input name to measurementResult?

riemanli (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/911)* status: 16 of 17 files reviewed, 4 unresolved discussions (waiting on @iverson52000, @return, @SanjayVas, and @uakyol)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt` line 140 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/911#-NS7xUFS9kv5fh2noS5B:-NSE9bDP0hheo1IEhlwO:by84cxh) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5e6ba9c003ef739bb2e3068ae662efb5753830ed/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt#L140)):*

looks great Rieman!
</blockquote></details>

Great. Thanks Ugur!
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt` line 730 at r11](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/911#-NSE8W_p5bHPG8yFHjr0:-NSE9iyY0l0gg6hLpc6t:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0191a64498e1614a600d5dc1f293072ab188721e/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt#L730)):*

lets rename this val to measurementResult here and in other places?
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt` line 740 at r11](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/911#-NSE8IMIC-hTTs4L4jge:-NSE95xsAKOl40mEw6Pr:b-x8yi3y) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0191a64498e1614a600d5dc1f293072ab188721e/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt#L740)):*

can you change this to. @return MeasurementResult. 
</blockquote></details>

Thanks for catching it! Done.
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt` line 749 at r11](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/911#-NSE7ROH4Sjv1kvFgAYt:-NSFN-c854HgpawE1rgT:bkx3jeh) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0191a64498e1614a600d5dc1f293072ab188721e/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt#L749)):*

sorry, I meant `calculateDirectFrequency`
</blockquote></details>

Sure. Modified.
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt` line 831 at r11](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/911#-NSE8hkW9wQAQpBcPaVV:-NSE9ZoF4emD4NYhAaF3:brgq1k0) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0191a64498e1614a600d5dc1f293072ab188721e/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt#L831)):*

while we are here, would you mind changing this input name to measurementResult?
</blockquote></details>

Sure. Done.

iverson52000 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/911)* status: 16 of 17 files reviewed, 1 unresolved discussion (waiting on @riemanli and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt` line 742 at r12](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/911#-NSHU9Nj5DNnS-NZDf8J:-NSHU9Nj5DNnS-NZDf8K:b3zka27) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f8ed91d000222d0ec437053ac19faed420bcc11b/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt#L742)):*
> ```Kotlin
>    * @return [Measurement.Result].
>    */
>   private fun buildMeasurementResult(
> ```

Name this `buildDirectMeasurementResult`?

riemanli (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/911)* status: 16 of 17 files reviewed, 1 unresolved discussion (waiting on @iverson52000 and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt` line 742 at r12](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/911#-NSHU9Nj5DNnS-NZDf8J:-NSHhe7t92xSmEKXTgfE:b-xu0uci) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f8ed91d000222d0ec437053ac19faed420bcc11b/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt#L742)):*

Name this `buildDirectMeasurementResult`?
</blockquote></details>

Thanks for the suggestion. Done.

SanjayVas (approved):
> Reviewed 1 of 1 files at r13, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/911)* status: all files reviewed, 1 unresolved discussion (waiting on @iverson52000 and @stevenwarejones)

iverson52000 (approved):
> Reviewed 3 of 7 files at r1, 4 of 10 files at r3, 1 of 3 files at r4, 4 of 4 files at r5, 1 of 3 files at r6, 1 of 1 files at r8, 1 of 3 files at r9, 1 of 1 files at r10, 1 of 1 files at r13, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/911)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @stevenwarejones)

Marco-Premier (approved):
> Reviewed 3 of 7 files at r1, 4 of 10 files at r3, 1 of 3 files at r4, 4 of 4 files at r5, 1 of 3 files at r6, 1 of 1 files at r8, 1 of 3 files at r9, 1 of 1 files at r10, 1 of 1 files at r13, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/911)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @stevenwarejones)

riemanli (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/911)* status: 13 of 17 files reviewed, 1 unresolved discussion (waiting on @iverson52000, @Marco-Premier, @renjiezh, @SanjayVas, @stevenwarejones, @tristanvuong2021, and @uakyol)

*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/911#-NSgtQWGEEBfDbQcwiJt:-NSgtQWGEEBfDbQcwiJu:bwa26cu) (no related file):*
@uakyol @iverson52000 @Marco-Premier would you please review the new changes in EdpSimulator.kt? I made some small changes to adopt the latest main. Should be fairly easy to review. Thanks!

___

iverson52000 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/911)* status: 13 of 17 files reviewed, 3 unresolved discussions (waiting on @Marco-Premier, @renjiezh, @riemanli, @SanjayVas, @stevenwarejones, @tristanvuong2021, and @uakyol)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt` line 750 at r15](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/911#-NSh3I5V-GilG1NR5Hg0:-NSh3I5V-GilG1NR5Hg1:b-35h7dz) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/b4fc28ad9640930b1116e6a97916b510bd6e15db/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt#L750)):*
> ```Kotlin
> 
>   /**
>    * Add Laplace publisher noise to calculated direct reach.
> ```

Update comment: Add publisher noise to calculated direct reach.
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt` line 770 at r15](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/911#-NSh3Wt26O56uIYsGPYw:-NSh3Wt26O56uIYsGPYx:bhpixzf) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/b4fc28ad9640930b1116e6a97916b510bd6e15db/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt#L770)):*
> ```Kotlin
> 
>   /**
>    * Add Laplace publisher noise to calculated direct reach and frequency.
> ```

Update comment: Add publisher noise to calculated frequency.

iverson52000 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/911)* status: 13 of 17 files reviewed, 3 unresolved discussions (waiting on @Marco-Premier, @renjiezh, @riemanli, @SanjayVas, @stevenwarejones, @tristanvuong2021, and @uakyol)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt` line 770 at r15](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/911#-NSh3Wt26O56uIYsGPYw:-NSh5cyh2CiVismoTnFH:b-yvuzzb) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/b4fc28ad9640930b1116e6a97916b510bd6e15db/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt#L770)):*
> calculated frequency.
Sorry. calculated direct frequency.

riemanli (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/911)* status: 13 of 17 files reviewed, 3 unresolved discussions (waiting on @iverson52000, @Marco-Premier, @renjiezh, @SanjayVas, @stevenwarejones, @tristanvuong2021, and @uakyol)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt` line 750 at r15](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/911#-NSh3I5V-GilG1NR5Hg0:-NSh5f5wALCFHVzq976z:b-x8yi3y) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/b4fc28ad9640930b1116e6a97916b510bd6e15db/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt#L750)):*

Update comment: Add publisher noise to calculated direct reach.
</blockquote></details>

Thanks for catching it! Done.
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt` line 770 at r15](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/911#-NSh3Wt26O56uIYsGPYw:-NSh62Ez5FkIyUlkdGOH:b-mws2j5) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/b4fc28ad9640930b1116e6a97916b510bd6e15db/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt#L770)):*

> calculated frequency.
Sorry. calculated direct frequency.
</blockquote></details>

Thanks for pointing it out! Done.

iverson52000 (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/911)* status: 13 of 17 files reviewed, 1 unresolved discussion (waiting on @Marco-Premier, @renjiezh, @SanjayVas, @stevenwarejones, @tristanvuong2021, and @uakyol)

SanjayVas (approved):
> Reviewed 1 of 1 files at r14, 2 of 3 files at r15, 1 of 1 files at r16, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/911)* status: all files reviewed, 1 unresolved discussion (waiting on @Marco-Premier, @stevenwarejones, and @uakyol)

Marco-Premier (approved):
> Reviewed 1 of 1 files at r14, 2 of 3 files at r15, 1 of 1 files at r16, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/911)* status: all files reviewed, 1 unresolved discussion (waiting on @stevenwarejones and @uakyol)

riemanli (commented):
> Reviewed 3 of 7 files at r1, 2 of 10 files at r3, 1 of 3 files at r4, 4 of 4 files at r5, 1 of 3 files at r6, 1 of 1 files at r8, 1 of 3 files at r9, 1 of 1 files at r14, 2 of 3 files at r15, 1 of 1 files at r16, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/911)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @stevenwarejones and @uakyol)

riemanli (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/911)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @stevenwarejones and @uakyol)

PR #930, opened 2023-04-05
setting up bazel for GRPC Gateway

bdomen-ggl:


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/930)

bdomen-ggl (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/930)* status: 0 of 3 files reviewed, 5 unresolved discussions

___
*[`repositories.bzl` line 3 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/930#-NSHM5UH8Aj1j-iwiobz:-NSHM5UH8Aj1j-iwioc-:b-3qvcq) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/002bbc33c0db2909abc35fbd7ce962dcddc70fbb/repositories.bzl#L3)):*
> ```Starlark
> load("@bazel_gazelle//:deps.bzl", "go_repository")
> 
> def go_repositories():
> ```

This is required to exist and have when pulling in go dependencies. This was the file in the example, but I'm not sure which ones are specifically needed for a simple GRPC Gateway vs which ones are fully comprehensive. Also, the package that loads these dependencies apparently expects the specific file `./repositories.bzl` with the method `def go_repositories():` to exist. So I can't move this to another location as far as I can tell.
___
*[`WORKSPACE` line 19 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/930#-NSHMU7uCuCETTpXNkUK:-NSHMU7uCuCETTpXNkUL:b7vksti) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/002bbc33c0db2909abc35fbd7ce962dcddc70fbb/WORKSPACE#L19)):*
> ```Starlark
> 
> 
> ####### My STUFF #######
> ```

I'm trying to pull this out of WORKSPACE and into a file (or set of files) for like "load reporting UI resources" and "load GRPC Gateway resource". I'm running into the problem that most of these have to be loaded and run in a very specific order. Another problem despite pulling these out into many files to preserve order is that I still ran into a loading problem - possibly due to lazy loading or something. Comment below.
___
*[`WORKSPACE` line 114 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/930#-NSHN4F_9gaxuHTUGgAV:-NSHN4F_9gaxuHTUGgAW:b13dlqp) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/002bbc33c0db2909abc35fbd7ce962dcddc70fbb/WORKSPACE#L114)):*
> ```Starlark
> go_rules_dependencies()
> 
> go_register_toolchains(version = "1.19.4")
> ```

These lines were causing me problems even when I was trying to extract all this into separate files. For some reason, they didn't seem to be loading correctly (or at the correct time).

I also need the specific version (or would have to go dig through and fix a lot of dependency version) which causes problems below.



_Code quote:_
```Starlark
load("@io_bazel_rules_go//go:deps.bzl", "go_register_toolchains", "go_rules_dependencies")

go_rules_dependencies()

go_register_toolchains(version = "1.19.4")
```
___
*[`WORKSPACE` line 228 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/930#-NSHNQdW7ipPGT5txnUV:-NSHNQdXCc8wNeQtMdBF:bpwi9wx) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/002bbc33c0db2909abc35fbd7ce962dcddc70fbb/WORKSPACE#L228)):*
> ```Starlark
> )
> 
> # load("@wfa_common_jvm//build:common_jvm_extra_deps.bzl", "common_jvm_extra_deps")
> ```

One of the external dependencies called by this also calls `go_register_toolchains` with a different hardcoded version (1.18.1). The method throws an error stating that it has already been called. Since I'm not the one loading the packages, but this method is, I'm not sure how to address this nicely. The below works but it seems really bad.
___
*[`build/grpc_extra_deps.bzl` line 57 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/930#-NSHO0ii7oabWREopFV5:-NSHO0ii7oabWREopFV6:b3k44n4) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/002bbc33c0db2909abc35fbd7ce962dcddc70fbb/build/grpc_extra_deps.bzl#L57)):*
> ```Starlark
>     # go_rules_dependencies()
>     # go_register_toolchains(version = "1.18.1")
>     # gazelle_dependencies()
> ```

This file is from the external package and is the block in the external library that is calling `go_register_toolchains` with version 1.18.1. This is incompatible with the latest GRPC Gateway and I'm not sure the best way to fix this.



_Code quote:_
```Starlark
    # go_rules_dependencies()
    # go_register_toolchains(version = "1.18.1")
    # gazelle_dependencies()
```

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/930)* status: 0 of 3 files reviewed, 6 unresolved discussions (waiting on @bdomen-ggl)

___
*[`WORKSPACE` line 48 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/930#-NSHQMYJ1XqC8S9_p4Tq:-NSHQMYJ1XqC8S9_p4Tr:b-8cqiqw) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/002bbc33c0db2909abc35fbd7ce962dcddc70fbb/WORKSPACE#L48)):*
> ```Starlark
> # =============== GRPC  GATEWAY =============== #
> # This is very finicky with how the imports need to be loaded
> ```

You should be able to put all `http_archive`s first in any order, as it's just indicating where something can be downloaded. Nothing actually happens until you start `load`ing defs from those archives.



_Code quote:_
```Starlark
# This is very finicky with how the imports need to be loaded
```
___
*[`build/grpc_extra_deps.bzl` line 57 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/930#-NSHO0ii7oabWREopFV5:-NSHRqkK4SR2hR4rFjZm:b5pqp5r) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/002bbc33c0db2909abc35fbd7ce962dcddc70fbb/build/grpc_extra_deps.bzl#L57)):*

This file is from the external package and is the block in the external library that is calling `go_register_toolchains` with version 1.18.1. This is incompatible with the latest GRPC Gateway and I'm not sure the best way to fix this.
</blockquote></details>

It sounds like what we need to do is decompose grpc_extra_deps in common-jvm and have it use the newer version. We already pull in various dependencies at newer versions first so that grpc_deps/grpc_extra_deps doesn't pull in older ones. This works because they wrap `http_archive` targets in checks to see whether they're already defined, but I guess that doesn't work with go_register_toolchains.

As an aside, the Bazel external dependency story is a giant mess. Bzlmod is intended to fix this, but it needs the whole ecosystem to switch.

PR #946, opened 2023-04-13
Add flag to config mill retry backoff delay base.

renjiezh:
Two Questions to solve:
1. For duchy's api `enqueue()`, do we need extra check to see if the token's lock owner is the same as current token?
No. The computationService checks whether the token is the latest version which guarantees that the lockowner in the request must be the same as the current lockowner. 
Currently this api is only used by mills to request retries for failed computations.

2. Do we need mill's retry backoff mechanism given the FIFO behavior for computations.
Yes. It is good-to-have feature. When we set backoff base to 0, it means mills should immediately retry failed computations (because they are the oldest, of course). On the other side, we might want the mills to calm down and work on other computations. Thus this mechanism add a cooldown time for retriable computations.


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/946)

SanjayVas (changes requested):
> The computationService checks whether the token is the latest version which guarantees that the lockowner in the request must be the same as the current lockowner.

This just checks that the caller obtained a sufficiently recent token. AFAIK there's no check that ensures that the caller is the current lock owner.

> we might want the mills to calm down and work on other computations.

Don't we always want FIFO behavior though? Meaning shouldn't a Mill keep retrying the oldest computation until it hits the retry limit and then move onto the next oldest? @kungfucraig 

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/946#-:-NTFV2HJChQQAnB3_6yC:b-fnck95)* status: 6 of 7 files reviewed, 2 unresolved discussions (waiting on @kungfucraig and @renjiezh)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/MillBase.kt` line 113 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/946#-NTFWE8a7ZkuA-h7liRa:-NTFWE8a7ZkuA-h7liRb:b-g3hx43) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/108c12b939fe10a502356b6b9a2a0dd2d06b4033/src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/MillBase.kt#L113)):*
> ```Kotlin
>   private val requestChunkSizeBytes: Int,
>   private val maximumAttempts: Int,
>   private val retryBackoffDelayBaseSeconds: Int,
> ```

Use a Duration.



_Suggestion:_
```Kotlin
retryInitialDelay: Duration
```
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/MillBase.kt` line 485 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/946#-NTFW3bJAHYM-DThtpzx:-NTFW3bJAHYM-DThtpzy:b9qh6kf) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/108c12b939fe10a502356b6b9a2a0dd2d06b4033/src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/MillBase.kt#L485)):*
> ```Kotlin
>   /** Enqueue a computation with a delay. */
>   private suspend fun enqueueComputation(token: ComputationToken) {
>     // Exponential backoff
> ```

Extract the Herald.ExpotentialBackoff class to a common location and use that. You may need to add a public instance method to return the duration given an attempt number.

renjiezh (commented):
Thanks sanjay.

1. In current system, if a mill calls `enqueue()` without holding the lock  (that means the lock is preempted by other mills), the server will detect that by only checking whether the token is most recent. It is guaranteed by design that when a mill claims a task, the `lockOwner` is true and valid. So I think the ownership is implicitly checked. Please let me know if I omit something.

2. Either keep trying the oldest one or have a cooldown for the retries. This can be configured by the flag of this PR.

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/946#-:-NTFYsPHDrDrSrJQ-fQC:b9qzo8l)* status: 6 of 7 files reviewed, 2 unresolved discussions (waiting on @kungfucraig)

SanjayVas (commented):
> It is guaranteed by design that when a mill claims a task, the lockOwner is true and valid.

The token need not be from `claimTask`. The Mill can just call `getComputationToken` to get a new token for the Computation which may now be owned by someone else. Indeed, it does this right before the `enqueueComputation` call in `handleExceptions`.

> Either keep trying the oldest one or have a cooldown for the retries. This can be configured by the flag of this PR.

Or follow the YAGNI principle if we know we always want FIFO behavior. No need for flags that will never be used.

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/946#-:-NTFc8fB3ExgpgeF_-z2:b-oq20x3)* status: 6 of 7 files reviewed, 2 unresolved discussions (waiting on @kungfucraig and @renjiezh)

kungfucraig (commented):
My review isn't going to add any extra value. I'll take myself off the reviewers list.

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/946#-:-NTFhBme0w1LcYI3J1dH:b-x33sqy)* status: 6 of 7 files reviewed, 2 unresolved discussions (waiting on @renjiezh and @SanjayVas)

PR #952, opened 2023-04-18
Fix BigQuery timestamp conversions.

SanjayVas:
BigQuery timestamps are represented as microseconds since epoch.


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/952)

PR #951, opened 2023-04-17
v2alpha metrics service -- ListMetrics

riemanli:

PR #959, opened 2023-04-20
Add getMetric and batchGetMetric.

riemanli:


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/959)

riemanli:
* main
  * **PR #951** <a href="https://app.graphite.dev/github/pr/world-federation-of-advertisers/cross-media-measurement/951" target="_blank">Graphite</a> 
    * **PR #959** <a href="https://app.graphite.dev/github/pr/world-federation-of-advertisers/cross-media-measurement/959" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/world-federation-of-advertisers/cross-media-measurement/959?utm_source=stack-comment).

riemanli (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/959)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @SanjayVas and @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 1075 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/959#-NToH8oH74cNSzN3PUSC:-NToRU4J06BpmDFtgtu6:b-6gwb5z) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/55a74ae7d2005dfc9eccf444df74dbc83940985a/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L1075)):*

Inspect the status code and throw a StatusRuntimeException for known cases, e.g. NOT_FOUND.
</blockquote></details>

Thanks for pointing it out! The internal metrics service hasn't been implemented yet, so I am not sure what status codes will be returned. 

For other status exception catch, I added as much as I can.
___
*[`src/test/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsServiceTest.kt` line 4374 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/959#-NToJ8087cqf3tIlaUcJ:-NTp4wCn7VjXsh7XTCQO:b-ywdonf) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/55a74ae7d2005dfc9eccf444df74dbc83940985a/src/test/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsServiceTest.kt#L4374)):*

Since you're stubbing batchGetMetrics in this test method, you can use an argument matcher in the stubbing rather than capturing/verifying the arguments separately. If the stubbed service method isn't called as expected, the other verifications will fail.

More broadly, you should generally only perform interaction verification for functions which change state. See https://abseil.io/resources/swe-book/html/ch13.html#best_practices_for_interaction_testing (Google-internal version at go/mocking#interaction-state-changes)
</blockquote></details>

Thanks for suggesting and sharing the guide. I think there are two difficulties to use argument matcher for `batchGetMetrics`:
1. I need to verify how many times `batchGetMetrics` is called for some tests. This is listed as one of the reasons to use interaction test in the guide.
2. The behavior of `batchGetMetrics` can't be differentiated by the input argument. For example, `batchGetMetrics` can return a failed metric or a pending metric given the same external metric ID.

Please let me know if there is a way to make the argument matcher work. Appreciated it!

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/959)* status: all files reviewed, 2 unresolved discussions (waiting on @riemanli and @tristanvuong2021)

___
*[`src/test/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsServiceTest.kt` line 4374 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/959#-NToJ8087cqf3tIlaUcJ:-NTtbQoV5t8FmPCT4-ve:b-k0ownj) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/55a74ae7d2005dfc9eccf444df74dbc83940985a/src/test/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsServiceTest.kt#L4374)):*
> I need to verify how many times batchGetMetrics is called for some tests. 

Sure, though note that this is only needed in cases where e.g. you return the same result across those calls. If you return a different result on consecutive calls, then the test should fail for other reasons.

> The behavior of batchGetMetrics can't be differentiated by the input argument. For example, batchGetMetrics can return a failed metric or a pending metric given the same external metric ID.

Isn't this just consecutive calls with the same matcher? That's technically what you have right now in this method, but the matcher happens to be `any()`.
___
*[`src/test/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsServiceTest.kt` line 2758 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/959#-NTtau-J5bVLG5pjLyvd:-NTtau-J5bVLG5pjLyve:bo4yesf) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3d8b61158000a22aa359d9bd04a3c436edb5d548/src/test/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsServiceTest.kt#L2758)):*
> ```Kotlin
>         }
>       assertThat(exception.grpcStatusCode()).isEqualTo(Status.Code.INVALID_ARGUMENT)
>       val expectedExceptionDescription = "Required field unspecified or invalid."
> ```

Don't assert on actual exception messages unless you need that to disambiguate between multiple causes of the same exception. We want to be able to change these messages freely. When you do need to assert on the message, do the minimum to disambiguate.



_Code quote:_
```Kotlin
      val expectedExceptionDescription = "Required field unspecified or invalid."
      assertThat(exception.message).contains(expectedExceptionDescription)
```

riemanli (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/959)* status: 1 of 2 files reviewed, 2 unresolved discussions (waiting on @SanjayVas and @tristanvuong2021)

___
*[`src/test/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsServiceTest.kt` line 4374 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/959#-NToJ8087cqf3tIlaUcJ:-NTu3lo9-vupm8ALw32L:bjcngaa) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/55a74ae7d2005dfc9eccf444df74dbc83940985a/src/test/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsServiceTest.kt#L4374)):*
> though note that this is only needed in cases where e.g. you return the same result across those calls. If you return a different result on consecutive calls, then the test should fail for other reasons.

That's a good point! Removed the checks of verifying arguments when the behaviors can be captured by stubbed methods.

> Isn't this just consecutive calls with the same matcher? That's technically what you have right now in this method, but the matcher happens to be any().

I see what you meant now. Made the modification according to the suggestion for `batchGetMetrics` and other stubbed methods as well.
___
*[`src/test/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsServiceTest.kt` line 2758 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/959#-NTtau-J5bVLG5pjLyvd:-NTteXr3AdlLqWMpjARq:but4721) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3d8b61158000a22aa359d9bd04a3c436edb5d548/src/test/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsServiceTest.kt#L2758)):*

Don't assert on actual exception messages unless you need that to disambiguate between multiple causes of the same exception. We want to be able to change these messages freely. When you do need to assert on the message, do the minimum to disambiguate.
</blockquote></details>

That's a great point! Removed the checks of exception messages.

SanjayVas (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/959)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @tristanvuong2021)

stevenwarejones (changes requested):
> Reviewed 1 of 1 files at r4.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/959)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @Marco-Premier, @riemanli, @SanjayVas, and @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 1019 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/959#-NTzYEZgBAPEIeeSJtgM:-NTzYEZgBAPEIeeSJtgN:brbnj99) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f6ee60cc29e43382980f502ce2eee5d23101fc22/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L1019)):*
> ```Kotlin
>       metrics +=
>         if (anyMeasurementUpdated) {
>           batchGetInternalMetrics(principal.resourceKey.measurementConsumerId, externalMetricIds)
> ```

do we need to get get them all for only some measurements updates?
___
*[`src/test/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsServiceTest.kt` line 864 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/959#-NTzYt560npuYUvgnl6O:-NTzYt560npuYUvgnl6P:bn8fajo) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f6ee60cc29e43382980f502ce2eee5d23101fc22/src/test/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsServiceTest.kt#L864)):*
> ```Kotlin
>   measurementPublicKey = MEASUREMENT_CONSUMER_PUBLIC_KEY.toByteString()
> 
>   nonceHashes.addAll(
> ```

```
nonceHashes += listOf(...)
```

riemanli (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/959)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @Marco-Premier, @SanjayVas, @stevenwarejones, and @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 1019 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/959#-NTzYEZgBAPEIeeSJtgM:-NTziq5p5qa_FPlno7fB:bmwy77y) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f6ee60cc29e43382980f502ce2eee5d23101fc22/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L1019)):*

do we need to get get them all for only some measurements updates?
</blockquote></details>

We don't. It's just easier to get the latest metrics in this way, without bookkeeping the mapping and maintaining the order, and I assume the table read is fairly efficient with a batch method.
Let me know if you think partial retrieval is preferred.
___
*[`src/test/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsServiceTest.kt` line 864 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/959#-NTzYt560npuYUvgnl6O:-NTzgDm73B6JWYIBULAa:b-tii5l4) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f6ee60cc29e43382980f502ce2eee5d23101fc22/src/test/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsServiceTest.kt#L864)):*

```
nonceHashes += listOf(...)
```

</blockquote></details>

Thanks for pointing it out! Fixed this one and other similar mistakes.

Marco-Premier (commented):
> Reviewed 1 of 2 files at r5, 1 of 1 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/959)* status: all files reviewed, 4 unresolved discussions (waiting on @riemanli, @SanjayVas, @stevenwarejones, and @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 925 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/959#-NU0YNo4AdkqVPljgeUW:-NU0YNo54CfsN0mSX0Gh:bxcz0gl) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/9582c5ef8830c1c5c82cbbd659d02f74fbdaa9eb/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L925)):*
> ```Kotlin
>     val principal: ReportingPrincipal = principalFromCurrentContext
>     when (principal) {
>       is MeasurementConsumerPrincipal -> {
> ```

Here we only checks that the MeasurementConsumer owns the metric. 
What if the principal firing this API is DataProviderPrincipal for instance? Shall we handle that case and return a proper error (i.e. ```PERMISSION_DENIED```)?



_Code quote:_
```Kotlin
MeasurementConsumerPrincipal
```
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 973 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/959#-NU0Ze-dCbLd4SafBLDv:-NU0Ze-eDqxIjO9YQ-Sk:b-6vumu5) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/9582c5ef8830c1c5c82cbbd659d02f74fbdaa9eb/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L973)):*
> ```Kotlin
> 
>     when (principal) {
>       is MeasurementConsumerPrincipal -> {
> ```

Same as above comment



_Code quote:_
```Kotlin
MeasurementConsumerPrincipal
```

Marco-Premier (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/959)* status: all files reviewed, 4 unresolved discussions (waiting on @riemanli, @SanjayVas, @stevenwarejones, and @tristanvuong2021)

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/959)* status: all files reviewed, 3 unresolved discussions (waiting on @Marco-Premier, @riemanli, and @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 925 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/959#-NU0YNo4AdkqVPljgeUW:-NU2U5fVBurGUoj_6mdc:b-eawx66) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/9582c5ef8830c1c5c82cbbd659d02f74fbdaa9eb/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L925)):*

+1
</blockquote></details>

`ReportingPrincipal` is a sealed interface. Therefore this should be an "exhaustive `when`" such that it will not compile if new principals get added. There's no need for an `else` condition.

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/959)* status: all files reviewed, 3 unresolved discussions (waiting on @Marco-Premier, @riemanli, and @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 1019 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/959#-NTzYEZgBAPEIeeSJtgM:-NU2ybMn7dA4vPw7QMNc:b-chjfz9) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f6ee60cc29e43382980f502ce2eee5d23101fc22/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L1019)):*

Currently, we set the maximum number as 1000. It's hard to tell how much the performance can be improved without a proper test. The query itself might not take much longer, but the response will be larger proportionally. Note that the maximum number could change after we load test the system. If we are ok to wait until the test after deployment, I will add a TODO comment to indicate a potential future improvement.
</blockquote></details>

Why not add a TODO now?
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 925 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/959#-NU0YNo4AdkqVPljgeUW:-NU2yTnH6LNPCRZfh5jl:by2il5v) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/9582c5ef8830c1c5c82cbbd659d02f74fbdaa9eb/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L925)):*

Thanks Sanjay for the explanation. Note that right now we only have [`MeasurementConsumerPrincipal`](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/main/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/ReportingPrincipal.kt#L46) in `ReportingPrincipal`. If the user uses a principal other than the principal listed in ReportingPrincipal.kt, it will be caught [here](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/main/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/PrincipalServerInterceptor.kt#L41).
</blockquote></details>

you still need an else right?

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/959)* status: all files reviewed, 3 unresolved discussions (waiting on @Marco-Premier, @riemanli, and @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 925 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/959#-NU0YNo4AdkqVPljgeUW:-NU3CRXR45Ih_-PnJJZ6:ba0ytii) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/9582c5ef8830c1c5c82cbbd659d02f74fbdaa9eb/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L925)):*

you still need an else right?
</blockquote></details>

Adding an `else` would make it so the `when` is non-exhaustive, i.e. it won't be a compiler error if a principal type is added but not handled here.

riemanli (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/959)* status: 1 of 2 files reviewed, 3 unresolved discussions (waiting on @Marco-Premier, @stevenwarejones, and @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 1019 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/959#-NTzYEZgBAPEIeeSJtgM:-NU2yq8I4KlsOhEyD674:beeb81k) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f6ee60cc29e43382980f502ce2eee5d23101fc22/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L1019)):*

Why not add a TODO now?
</blockquote></details>

I was checking again to see if you'd have different thoughts after my previous reply :) 
Added TODOs.

riemanli (commented):
> Dismissed @Marco-Premier from 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/959)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Marco-Premier and @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 925 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/959#-NU0YNo4AdkqVPljgeUW:-NUN67Pa9nig1pCzCEEw:bnxvak5) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/9582c5ef8830c1c5c82cbbd659d02f74fbdaa9eb/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L925)):*

i guess this depends on the desired behavior. If all other principal types are not supposed to have access, then an else could be added to reject. I'm open to doing this later when new types are added.
</blockquote></details>

Sounds good. Thanks Steven!

riemanli (commented):
> Reviewed 1 of 1 files at r6, 1 of 1 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/959)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Marco-Premier and @tristanvuong2021)

riemanli:
@riemanli queued this pull request to merge with [Graphite](https://app.graphite.dev/github/pr/world-federation-of-advertisers/cross-media-measurement/959).

riemanli:
@riemanli merged this pull request with [Graphite](https://app.graphite.dev/github/pr/world-federation-of-advertisers/cross-media-measurement/959).

PR #972, opened 2023-04-27
Rewrite Terraform configuration to be more modular.

SanjayVas:
This temporarily drops GitHub Actions workflows related to Terraform, as those also need to be rewritten.

https://rally1.rallydev.com/#/?detail=/task/700460303861&fdp=true


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/972)

SanjayVas:
Plan command for halo-cmm-dev:

```shell
terraform -chdir=src/main/terraform/gcloud/cmms plan -var 'cluster_location=us-central1-c' -var 'spanner_instance_name=dev-instance' -var 'key_ring_name=test-key-ring' -var 'key_ring_location=us-central1'
```

bdomen-ggl:
Didn't look in detail since it's a draft, but generally looks like how I'd start. Main suggestion would to break it up the resources into `modules` and `environments` folders like in this example: https://getbetterdevops.io/terraform-create-infrastructure-in-multiple-environments/. Then from GHA, you can call the main.tf in the appropriate environment folder.

SanjayVas:
> Main suggestion would to break it up the resources into `modules` and `environments` folders like in this example: https://getbetterdevops.io/terraform-create-infrastructure-in-multiple-environments/. Then from GHA, you can call the main.tf in the appropriate environment folder.

We don't want to actually hard-code the environments in here. These come from [GitHub Environments](https://github.com/world-federation-of-advertisers/cross-media-measurement/settings/environments) instead, meaning everything about the particular environment will be passed in as vars. All of our test environments (dev, QA, nightly) would use the `cmms` root module.

As far as having a `modules` directory, my understanding was that is normally a subdirectory of the root module. Furthermore, this subdirectory is intended for modules that wouldn't be used outside of that root module. In this case, all the modules except for `cmms` are intended to be externally reusable. The `cmms` module is the only root module.

bdomen-ggl:
Ah sorry my misunderstanding. If that's the case, then yeah the directory structure is fine. The idea would have been to make it clear which folders would be for QA vs Nightly vs Kingdom only vs Duchy only vs modules. I'll take another look with that in mind.

SanjayVas:
Ended up moving the reusable modules into a `modules` directory. Apparently it's fine to put reusable modules in there, and the convention is those with `README` files are meant to be reused externally.

bdomen-ggl:
High level looks good. Based on your comments earlier, we'll want to add configuration passed in from vars to create the different environments or a single Kingdom or a single Duchy. Depending on how configurable we want this to be (specifically handling the creation of resources like the KMS key outside of TF) we can add some more configurations. If you're ready and want some help I can add some commits to show how that would look. We can use the `count` parameter to determine if a resource is created or not (if not, we can have the opposite condition use a data block instead).

SanjayVas:
> Based on your comments earlier, we'll want to add configuration passed in from vars to create the different environments or a single Kingdom or a single Duchy.

What I'd do in a future PR is just add examples for setting up a single component.

bdomen-ggl (approved):
> Reviewed 12 of 28 files at r1, 12 of 20 files at r2, 19 of 22 files at r3, 2 of 4 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/972)* status: 45 of 49 files reviewed, 2 unresolved discussions (waiting on @SanjayVas)

___
*[`src/main/terraform/gcloud/cmms/duchies.tf` line 63 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/972#-NUgzR0WCJfh9aWpiWbQ:-NUgzR0WCJfh9aWpiWbR:b-82yh05) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/b7dda36cf2f4f629a77fd72ca7d4902f8880107d/src/main/terraform/gcloud/cmms/duchies.tf#L63)):*
> ```HCL
> }
> 
> provider "kubernetes" {
> ```

Some people like putting providers in a separate file `providers.tf`. It's probably fine to leave it like this.
___
*[`src/main/terraform/gcloud/modules/duchy/variables.tf` line 29 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/972#-NUh-S2OBrLkaF0NWY1Q:-NUh-S2P4hbtVbNRx5Qj:byytdhk) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/b7dda36cf2f4f629a77fd72ca7d4902f8880107d/src/main/terraform/gcloud/modules/duchy/variables.tf#L29)):*
> ```HCL
>   description = "`google_storage_bucket` for the system."
>   type = object({
>     name = string
> ```

Since there's only a name, can just make the variable like `storage_bucket_name`. If you think you'll need more variables for the storage bucket and would like them grouped, then it's fine to leave it like this.

Same with other variables files below.

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/972)* status: 45 of 49 files reviewed, 2 unresolved discussions (waiting on @bdomen-ggl)

___
*[`src/main/terraform/gcloud/cmms/duchies.tf` line 63 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/972#-NUgzR0WCJfh9aWpiWbQ:-NUh4qW68rEciWx61gD3:b-xg4jnb) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/b7dda36cf2f4f629a77fd72ca7d4902f8880107d/src/main/terraform/gcloud/cmms/duchies.tf#L63)):*

Some people like putting providers in a separate file `providers.tf`. It's probably fine to leave it like this.
</blockquote></details>

I think I prefer it this way since the provider configuration for these in particular are not static and only exist for the Duchies.
___
*[`src/main/terraform/gcloud/modules/duchy/variables.tf` line 29 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/972#-NUh-S2OBrLkaF0NWY1Q:-NUh5XlL3a4YoIlE5IGN:b-fnm0vn) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/b7dda36cf2f4f629a77fd72ca7d4902f8880107d/src/main/terraform/gcloud/modules/duchy/variables.tf#L29)):*

Since there's only a name, can just make the variable like `storage_bucket_name`. If you think you'll need more variables for the storage bucket and would like them grouped, then it's fine to leave it like this.

Same with other variables files below.
</blockquote></details>

I intentionally use the objects for these so it's clear that the resource/data must already exist. The ones that are just names are for when the module itself will include the resource/data.

SanjayVas (commented):
> Dismissed @bdomen-ggl from 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/972)* status: 45 of 49 files reviewed, all discussions resolved (waiting on @bdomen-ggl and @stevenwarejones)

Marco-Premier (approved):
> Reviewed 13 of 28 files at r1, 12 of 20 files at r2, 20 of 22 files at r3, 4 of 4 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/972)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @bdomen-ggl and @stevenwarejones)

PR #976, opened 2023-05-02
Add MultithreadingHelper to refactor multithreaded LLV2

renjiezh:
## Background
To multithread the execution of LLV2. Currently all for-loops are single-threaded but iterations are actually independent. They can be refactored into a multi-threading pattern.
Most operations rely on [ProtocolCryptor](https://sourcegraph.com/github.com/world-federation-of-advertisers/cross-media-measurement/-/blob/src/main/cc/wfa/measurement/common/crypto/protocol_cryptor.cc). Unfortunately, the class ProtocolCryptor is not thread-safe. Thus, multiple identical ProtocolCryptors must be created and each thread should have a dedicated ProtocolCryptor. 
Current iterations append results directly to the output variable. This is not a good pattern for multiple threading if the order matters because the execution order of iterations is not deterministic.

## Design 
The general idea is to have a helper class assigning tasks to multiple threads and collect output after execution.
### Parameters
The helper has three primary parameters: data, # iterations, and a lambda expression.
The lambda expression reuses current code as the task to run. Given data and # iterations, the helper execute the lambda expression with index from 0 to (# iterations -1). Thus, the helper can split tasks even if the input is in batch as a variable `data`.
Capture list of the lambda expression has all the variables to execute iterations. No major change required from current for-loop block. 
### Results
Lambda expression is responsible for handling results from iterations and making sure it is thread-safe.
### Cryptors
The ProtocolCryptor is not thread-safe. Each thread should have its own ProtocolCryptor to execute lambda expressions.
The helper can be reused multiple times with different types of input and output. ProtocolCryptor only needs to be initialized once. For example SetupPhase, it has multiple for-loops to add various types of noise. Only one helper need to create and reuse.
Some procedure does not require ProtocolCryptor thus can skip `setupCryptors`.

## Class Interface
See the header file.

## Usage
See the unit test.


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/976)

SanjayVas (changes requested):
> Reviewed 3 of 7 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/976)* status: 3 of 7 files reviewed, 2 unresolved discussions (waiting on @kungfucraig, @ple13, @renjiezh, and @stevenwarejones)

___
*[`src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/multithreading_helper.h` line 19 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/976#-NV6EUYp7dXwsV9NBGGI:-NV6EUYp7dXwsV9NBGGJ:ba7xhrn) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/cf46a011bd3391cbc1c625dca77d09bc7fab7f5f/src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/multithreading_helper.h#L19)):*
> ```objc
> 
> #include <memory>
> #include <thread>  // NOLINT [build/c++11]
> ```

I believe the more common syntax is `NOLINT(category)`



_Suggestion:_
```objc
NOLINT(build/c++11)
```
___
*[`src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/multithreading_helper.h` line 19 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/976#-NV6GjG0CKn7pcOsQszh:-NV6GjG0CKn7pcOsQszi:bq7swm8) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/cf46a011bd3391cbc1c625dca77d09bc7fab7f5f/src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/multithreading_helper.h#L19)):*
> ```objc
> 
> #include <memory>
> #include <thread>  // NOLINT [build/c++11]
> ```

Since we're already depending on Riegeli, I recommend using the parallelism primitives that have been open-sourced as part of that library (i.e. ThreadPool). https://github.com/google/riegeli/blob/master/riegeli/base/parallelism.h



_Code quote:_
```objc
#include <thread>
```

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/976)* status: 3 of 7 files reviewed, 2 unresolved discussions (waiting on @kungfucraig, @ple13, @SanjayVas, and @stevenwarejones)

___
*[`src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/multithreading_helper.h` line 19 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/976#-NV6GjG0CKn7pcOsQszh:-NV78-Xd8sBrK0c9obl7:bpecadj) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/cf46a011bd3391cbc1c625dca77d09bc7fab7f5f/src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/multithreading_helper.h#L19)):*

Since we're already depending on Riegeli, I recommend using the parallelism primitives that have been open-sourced as part of that library (i.e. ThreadPool). https://github.com/google/riegeli/blob/master/riegeli/base/parallelism.h
</blockquote></details>

Hi sanjay, there are two things seems Riegeli does not support. 
1. We need a thread_local variable `cryptor`, seems the library does not support that. 
2. The ThreadPool from it use unlimited number of threads. But we want to configure it as a fixed number.
Besides, Riegeli is also using std::thread directly. I feel the threading mechanism of MultithreadingHelper is pretty straightforward and having Status to take care of errors.
Please let me know your thought.
___
*[`src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/multithreading_helper.h` line 19 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/976#-NV6EUYp7dXwsV9NBGGI:-NV79toR-Shs_BwRzmKY:bq5ke66) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/cf46a011bd3391cbc1c625dca77d09bc7fab7f5f/src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/multithreading_helper.h#L19)):*

I believe the more common syntax is `NOLINT(category)`
</blockquote></details>

Nice catch. Guess I followed the wrong examples.

SanjayVas (changes requested):
> Reviewed 1 of 7 files at r1.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/976)* status: 3 of 7 files reviewed, 5 unresolved discussions (waiting on @kungfucraig, @ple13, @renjiezh, and @stevenwarejones)

___
*[`src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/multithreading_helper.h` line 19 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/976#-NV6GjG0CKn7pcOsQszh:-NV7Ja8-C3_etvFu29Di:bdboo7k) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/cf46a011bd3391cbc1c625dca77d09bc7fab7f5f/src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/multithreading_helper.h#L19)):*

Hi sanjay, there are two things seems Riegeli does not support. 
1. We need a thread_local variable `cryptor`, seems the library does not support that. 
2. The ThreadPool from it use unlimited number of threads. But we want to configure it as a fixed number.
Besides, Riegeli is also using std::thread directly. I feel the threading mechanism of MultithreadingHelper is pretty straightforward and having Status to take care of errors.
Please let me know your thought.
</blockquote></details>

I'm not seeing any variable with the `thread_local` storage duration specifier. That said, (2) is a bigger concern. I didn't realize that it didn't allow for specifying a max number of threads.

My main concern was with trying to reuse threads since threads can be relatively expensive (hence why we use thread pools in JVM code). I realized however that because this is called via JNI there's no way to actually maintain the threads across calls. Therefore using a thread pool on the C++ side doesn't really give us any benefits.
___
*[`src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/multithreading_helper.h` line 49 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/976#-NV7Kp5f66p8wTF1j9GN:-NV7Kp5f66p8wTF1j9GO:bbk45m1) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/cf46a011bd3391cbc1c625dca77d09bc7fab7f5f/src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/multithreading_helper.h#L49)):*
nit: We don't use any "Hungarian" or other type-based notation aside from the `k` prefix for constants. Just spell out `num`.



_Suggestion:_
```objc
num_threads
```
___
*[`src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/multithreading_helper.h` line 55 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/976#-NV7MTNtBP2rikHVGIoC:-NV7MTNtBP2rikHVGIoD:bqmzg5u) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/cf46a011bd3391cbc1c625dca77d09bc7fab7f5f/src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/multithreading_helper.h#L55)):*
Since you have a factory function you can handle any complex initialization in there without needing an Init method. Make the constructor private to force use of the factory.

See https://abseil.io/tips/42
___
*[`src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/multithreading_helper.h` line 64 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/976#-NV7MrSK4I9meomVya2_:-NV7MrSK4I9meomVya2a:banupo1) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/cf46a011bd3391cbc1c625dca77d09bc7fab7f5f/src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/multithreading_helper.h#L64)):*
Follow the Options pattern to hold all of the cryptor options.
___
*[`src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/multithreading_helper.h` line 74 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/976#-NV7LtqL-JuRG1jESEnz:-NV7LtqL-JuRG1jESEo-:b-hw0ghg) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/cf46a011bd3391cbc1c625dca77d09bc7fab7f5f/src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/multithreading_helper.h#L74)):*
What is the `AndBlocking` part of the name about? Why not just `Execute`? I see there's another method named that for each thread. Shouldn't that just be a private method `ExecuteCryptorFunction` or something?



_Code quote:_
```objc
ExecuteAndBlocking
```

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/976)* status: 3 of 7 files reviewed, 4 unresolved discussions (waiting on @kungfucraig, @ple13, @SanjayVas, and @stevenwarejones)

___
*[`src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/multithreading_helper.h` line 55 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/976#-NV7MTNtBP2rikHVGIoC:-NV81Ljk-cG7YWCATXDJ:bjd5ebd) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/cf46a011bd3391cbc1c625dca77d09bc7fab7f5f/src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/multithreading_helper.h#L55)):*

Since you have a factory function you can handle any complex initialization in there without needing an Init method. Make the constructor private to force use of the factory.

See https://abseil.io/tips/42
</blockquote></details>

Thanks. Very great advice.
___
*[`src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/multithreading_helper.h` line 64 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/976#-NV7MrSK4I9meomVya2_:-NV82AWu1rcb99K0MDi0:bg39lkf) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/cf46a011bd3391cbc1c625dca77d09bc7fab7f5f/src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/multithreading_helper.h#L64)):*

Follow the Options pattern to hold all of the cryptor options.
</blockquote></details>

Do you mean adding the `struct CryptorOptions{ ElGamalCiphertext local_el_gamal_public_key ; //...//};`?
In this way, seems it has to copy the value of each field. Here we are using reference. Is there anyway to use the option struct without copy strings?
___
*[`src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/multithreading_helper.h` line 74 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/976#-NV7LtqL-JuRG1jESEnz:-NV827Mu7mpULKtD5fdb:bt1b19p) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/cf46a011bd3391cbc1c625dca77d09bc7fab7f5f/src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/multithreading_helper.h#L74)):*

What is the `AndBlocking` part of the name about? Why not just `Execute`? I see there's another method named that for each thread. Shouldn't that just be a private method `ExecuteCryptorFunction` or something?
</blockquote></details>

Renamed.

kungfucraig (changes requested):
> Reviewed 4 of 7 files at r1, 1 of 3 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/976)* status: 5 of 7 files reviewed, 7 unresolved discussions (waiting on @ple13, @renjiezh, @SanjayVas, and @stevenwarejones)

___
*[`src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/multithreading_helper.h` line 53 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/976#-NVAXPBU96LlwTchVgUv:-NVAXPBU96LlwTchVgUw:b-3u36xd) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/95c6a9ae6337fd60aefc89c15dcfdb617d429a38/src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/multithreading_helper.h#L53)):*
> ```objc
> 
>   int num_threads_;
>   std::vector<std::thread> threads_;
> ```

It seems like this can be a local variable in "Execute."
___
*[`src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/multithreading_helper.h` line 56 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/976#-NVAXgkwC8Lb098E79vu:-NVAXgkwC8Lb098E79vv:b-jc0tmc) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/95c6a9ae6337fd60aefc89c15dcfdb617d429a38/src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/multithreading_helper.h#L56)):*
> ```objc
>   std::vector<std::unique_ptr<ProtocolCryptor>> cryptors_;
> 
>   int num_iterations = 0;
> ```

It seems like this can be a local local variable in Execute and passed into ExecuteCryptorTask.
___
*[`src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/multithreading_helper.h` line 107 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/976#-NVAYD9OAAaAmydi33lM:-NVAYD9OAAaAmydi33lN:bysrxmx) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/95c6a9ae6337fd60aefc89c15dcfdb617d429a38/src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/multithreading_helper.h#L107)):*
> ```objc
>         absl::MutexLock lock(&mutex_);
> 
>         if (iteration_index_ < num_iterations && !failure_status_.has_value()) {
> ```

Is there any value in num iterations being different than the number of threads?

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/976)* status: 5 of 7 files reviewed, 7 unresolved discussions (waiting on @kungfucraig, @ple13, @SanjayVas, and @stevenwarejones)

___
*[`src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/multithreading_helper.h` line 53 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/976#-NVAXPBU96LlwTchVgUv:-NVAdh6K6GT-OcJW8W1w:b-3tbf09) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/95c6a9ae6337fd60aefc89c15dcfdb617d429a38/src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/multithreading_helper.h#L53)):*

It seems like this can be a local variable in "Execute."
</blockquote></details>

Yeah, they can be local variables and be passed by references ( num_threads_, num_iterations_, mutex_, failure_status_).  
However, from the object-oriented perspective(intuitively), theses variables are fields guarded by the mutex within the object. Making them into local variables could lead the function params list very long.
Do you suggest to use more local variable as possible?
___
*[`src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/multithreading_helper.h` line 56 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/976#-NVAXgkwC8Lb098E79vu:-NVAfrYSBCMSOzeYQAL-:bjgvfcz) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/95c6a9ae6337fd60aefc89c15dcfdb617d429a38/src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/multithreading_helper.h#L56)):*

It seems like this can be a local local variable in Execute and passed into ExecuteCryptorTask.
</blockquote></details>

Redirect to discussion above.
___
*[`src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/multithreading_helper.h` line 107 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/976#-NVAYD9OAAaAmydi33lM:-NVAd6aI9pm4kvEG0rI3:b-a04ct9) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/95c6a9ae6337fd60aefc89c15dcfdb617d429a38/src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/multithreading_helper.h#L107)):*

Is there any value in num iterations being different than the number of threads?
</blockquote></details>

num_threads is the parallelism(e.g. 4 or 10 based on cores of cpu). Num_iterations is the index limit in the for-loop.

SanjayVas (changes requested):
> Reviewed 1 of 3 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/976)* status: 5 of 7 files reviewed, 8 unresolved discussions (waiting on @kungfucraig, @ple13, @renjiezh, and @stevenwarejones)

___
*[`src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/multithreading_helper.h` line 64 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/976#-NV7MrSK4I9meomVya2_:-NVApbMSAi2EESEMdaLV:b-xurbyl) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/cf46a011bd3391cbc1c625dca77d09bc7fab7f5f/src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/multithreading_helper.h#L64)):*

Do you mean adding the `struct CryptorOptions{ ElGamalCiphertext local_el_gamal_public_key ; //...//};`?
In this way, seems it has to copy the value of each field. Here we are using reference. Is there anyway to use the option struct without copy strings?
</blockquote></details>

https://abseil.io/tips/173

See the section on temporaries under Caveats in particular.

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/976)* status: 4 of 10 files reviewed, 8 unresolved discussions (waiting on @kungfucraig, @ple13, @SanjayVas, and @stevenwarejones)

___
*[`src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/multithreading_helper.h` line 64 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/976#-NV7MrSK4I9meomVya2_:-NVBjtOsEJTcX9o1KCyc:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/cf46a011bd3391cbc1c625dca77d09bc7fab7f5f/src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/multithreading_helper.h#L64)):*

https://abseil.io/tips/173

See the section on temporaries under Caveats in particular.
</blockquote></details>

Done.

SanjayVas (changes requested):
> Reviewed 2 of 4 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/976)* status: 6 of 10 files reviewed, 9 unresolved discussions (waiting on @kungfucraig, @ple13, @renjiezh, and @stevenwarejones)

___
*[`src/main/cc/wfa/measurement/common/crypto/protocol_cryptor.h` line 112 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/976#-NVCPdUzDcEHY3XnpvUx:-NVCPdV-Ba1ch3lLylfZ:bxqdzwb) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6450ccda185f8eee362427d1abf4c21efc30b880/src/main/cc/wfa/measurement/common/crypto/protocol_cryptor.h#L112)):*
> ```objc
> };
> 
> struct ProtocolCryptorKeys {
> ```

nit: Consider naming this Options to show we're following that well-known pattern.



_Suggestion:_
```objc
ProtocolCryptorOptions
```
___
*[`src/main/cc/wfa/measurement/common/crypto/protocol_cryptor.h` line 113 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/976#-NVCOkUEDeYPZubvVhBH:-NVCOkUEDeYPZubvVhBI:b-fvjdwb) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6450ccda185f8eee362427d1abf4c21efc30b880/src/main/cc/wfa/measurement/common/crypto/protocol_cryptor.h#L113)):*
> ```objc
> 
> struct ProtocolCryptorKeys {
>   explicit ProtocolCryptorKeys(
> ```

Use designated initializer rather than regular constructor as described in https://abseil.io/tips/173
___
*[`src/main/cc/wfa/measurement/common/crypto/protocol_cryptor.h` line 135 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/976#-NVCPlZy5coIVVUE74er:-NVCPlZy5coIVVUE74es:buwlnls) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6450ccda185f8eee362427d1abf4c21efc30b880/src/main/cc/wfa/measurement/common/crypto/protocol_cryptor.h#L135)):*
> ```objc
> 
> // Create a ProtocolCryptor using keys required for internal ciphers.
> absl::StatusOr<std::unique_ptr<ProtocolCryptor>> CreateProtocolCryptorWithKeys(
> ```

Why the "WithKeys" suffix?



_Suggestion:_
```objc
CreateProtocolCryptor
```

ple13 (changes requested):
There is one caveat: the ProtocolCryptors are only almost identical. Each of them has a different context that is initialized with a different random generator.

> Reviewed 4 of 7 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/976#-:-NVGkcuu0NDUPePmKRid:bmkt131)* status: 6 of 10 files reviewed, 12 unresolved discussions (waiting on @kungfucraig, @renjiezh, @SanjayVas, and @stevenwarejones)

___
*[`src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/multithreading_helper.cc` line 49 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/976#-NVGHnMu4CBYaBv5Hsa4:-NVGHnMu4CBYaBv5Hsa5:undefined) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/9ea0c9f61129937a6895542364915a4d88b56749/src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/multithreading_helper.cc#L49)):*
> ```cpp
> absl::StatusOr<std::vector<std::unique_ptr<ProtocolCryptor>>>
> MultithreadingHelper::CreateCryptors(
>     int num, int curve_id, const ElGamalCiphertext& local_el_gamal_public_key,
> ```





_Suggestion:_
```cpp
int num_threads
```
___
*[`src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/multithreading_helper.cc` line 56 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/976#-NVGTe5mAHAfn0vb1Lto:-NVGTe5mAHAfn0vb1Ltp:blzrulc) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/9ea0c9f61129937a6895542364915a4d88b56749/src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/multithreading_helper.cc#L56)):*
> ```cpp
>   std::vector<std::unique_ptr<ProtocolCryptor>> cryptors;
>   for (size_t i = 0; i < num; i++) {
>     ProtocolCryptorKeys keys(
> ```

Can we move the key out of the for loop?
___
*[`src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/multithreading_helper.cc` line 100 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/976#-NVGQ00M3ch8kGdMgwzx:-NVGQ00N-XN6znR9N1uk:undefined) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/9ea0c9f61129937a6895542364915a4d88b56749/src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/multithreading_helper.cc#L100)):*
> ```cpp
> 
>   for (size_t i = 0; i < count; i++) {
>     size_t current_index = start_index + i;
> ```





_Suggestion:_
```cpp
  for (size_t current_index = start_index; current_index < start_index + count; current_index++) {
```

kungfucraig (changes requested):
> Reviewed 3 of 4 files at r5, 1 of 3 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/976)* status: 8 of 10 files reviewed, 10 unresolved discussions (waiting on @ple13, @renjiezh, and @stevenwarejones)

___
*[`src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/multithreading_helper.h` line 53 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/976#-NVAXPBU96LlwTchVgUv:-NVBOQvfDCt8lr_zgnw1:b6a4z08) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/95c6a9ae6337fd60aefc89c15dcfdb617d429a38/src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/multithreading_helper.h#L53)):*

Yeah, `threads_` don't need to be protect by mutex.
But I guess it cannot be marked as const.
</blockquote></details>

Right, it can't be const, but because it's completely local to Execute, it doesn't need to be a member variable. 

Also, the way it is now, if Execute gets called more than once you'll create another set of threads.
___
*[`src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/multithreading_helper.cc` line 49 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/976#-NVGHnMu4CBYaBv5Hsa4:-NVV2o_e7z-lCJwLW8Dv:b-vzhvzq) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/9ea0c9f61129937a6895542364915a4d88b56749/src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/multithreading_helper.cc#L49)):*
> ```cpp
> absl::StatusOr<std::vector<std::unique_ptr<ProtocolCryptor>>>
> MultithreadingHelper::CreateCryptors(
>     int num, int curve_id, const ElGamalCiphertext& local_el_gamal_public_key,
> ```

"num" is actually more correct here as it is indicating the number of cryptors to create.
___
*[`src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/multithreading_helper.cc` line 77 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/976#-NVV3WxR6-aweKF4N4dp:-NVV3WxR6-aweKF4N4dq:b99a57l) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/9ea0c9f61129937a6895542364915a4d88b56749/src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/multithreading_helper.cc#L77)):*
> ```cpp
>       count = num_iterations / num_threads_;
>     }
>     threads_.emplace_back(std::thread(&MultithreadingHelper::ExecuteCryptorTask,
> ```

Please make "threads_" a local variable. The reason is that if you call this function multiple times you'll either 1) need to clear threads_; 2) suffer from residue building up in threads_ due to previous calls. Making it local removes both of these concerns.
___
*[`src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/multithreading_helper.cc` line 86 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/976#-NVV4QmU-nLOYMlJ-0Pf:-NVV4QmU-nLOYMlJ-0Pg:b-zfbq0d) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/9ea0c9f61129937a6895542364915a4d88b56749/src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/multithreading_helper.cc#L86)):*
> ```cpp
>   }
> 
>   for (auto& failure : failures_) {
> ```

This has a similar problem with "threads_" in that you'll need to clear it after each call to Execute, which is error prone especially for future maintainers.

Alternatively

Make a local failures array..
std::vector<std::optional<absl::Status>> failures;

And have ExecuteCryptorTask accept another argument (std::optional<absl::Status>* ) that references the appropriate object in the local failures array and use it as an output parameter.
___
*[`src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/multithreading_helper.cc` line 100 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/976#-NVGQ00M3ch8kGdMgwzx:-NVV43977GIS3cuX4Umr:b32v57) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/9ea0c9f61129937a6895542364915a4d88b56749/src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/multithreading_helper.cc#L100)):*
> ```cpp
> 
>   for (size_t i = 0; i < count; i++) {
>     size_t current_index = start_index + i;
> ```

+1

kungfucraig (approved):
> Reviewed 1 of 5 files at r7, 3 of 4 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/976)* status: 8 of 10 files reviewed, 8 unresolved discussions (waiting on @ple13, @renjiezh, @SanjayVas, and @stevenwarejones)

___
*[`src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/liquid_legions_v2_encryption_utility.cc` line 599 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/976#-NVW-iML76txnPn4eWQ5:-NVW-iML76txnPn4eWQ6:b9puou5) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/93043f49201ad1f07e71efbb36e8f4dc110c7a47/src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/liquid_legions_v2_encryption_utility.cc#L599)):*
> ```cpp
> 
>     ProtocolCryptorOptions protocol_cryptor_keys{
> ```

nit: the variable name is still "keys" after the type rename.
___
*[`src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/multithreading_helper.h` line 54 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/976#-NVW-t3W06I-9n4Y98ez:-NVW-t3W06I-9n4Y98f-:b-vn6rwk) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/93043f49201ad1f07e71efbb36e8f4dc110c7a47/src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/multithreading_helper.h#L54)):*
> ```objc
> 
>   const int num_threads_;
>   std::vector<std::unique_ptr<ProtocolCryptor>> cryptors_;
> ```

You can mark this const too.
___
*[`src/test/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/multithreading_helper_test.cc` line 47 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/976#-NVW0U4F2VqJiIVO7DZB:-NVW0U4F2VqJiIVO7DZC:b-4rrcdm) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/93043f49201ad1f07e71efbb36e8f4dc110c7a47/src/test/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/multithreading_helper_test.cc#L47)):*
> ```cpp
>   return el_gamal_ciphertext;
> }
> ```

You should add a test that uses the same MultithreadingHelper to run a second function (or the same function twice is fine too).

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/976)* status: 4 of 10 files reviewed, 7 unresolved discussions (waiting on @kungfucraig, @ple13, @SanjayVas, and @stevenwarejones)

___
*[`src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/multithreading_helper.h` line 54 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/976#-NVW-t3W06I-9n4Y98ez:-NVWKx_62IENJGqLJUXa:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/93043f49201ad1f07e71efbb36e8f4dc110c7a47/src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/multithreading_helper.h#L54)):*

You can mark this const too.
</blockquote></details>

Done.
___
*[`src/test/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/multithreading_helper_test.cc` line 47 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/976#-NVW0U4F2VqJiIVO7DZB:-NVW74A04hB_vgBDRP4D:bagrte4) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/93043f49201ad1f07e71efbb36e8f4dc110c7a47/src/test/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/multithreading_helper_test.cc#L47)):*

You should add a test that uses the same MultithreadingHelper to run a second function (or the same function twice is fine too).
</blockquote></details>

Nice suggestion.

SanjayVas (changes requested):
> Reviewed 3 of 4 files at r8, 3 of 6 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/976)* status: 7 of 10 files reviewed, 10 unresolved discussions (waiting on @kungfucraig, @ple13, @renjiezh, and @stevenwarejones)

___
*[`src/main/cc/wfa/measurement/common/crypto/protocol_cryptor.cc` line 317 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/976#-NVWGokX3nev7RoJ33MI:-NVWGokX3nev7RoJ33MJ:b2ni8l0) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/93043f49201ad1f07e71efbb36e8f4dc110c7a47/src/main/cc/wfa/measurement/common/crypto/protocol_cryptor.cc#L317)):*
> ```cpp
> 
> absl::StatusOr<std::unique_ptr<ProtocolCryptor>> CreateProtocolCryptor(
>     const ProtocolCryptorOptions& keys) {
> ```

nit



_Suggestion:_
```cpp
options
```
___
*[`src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/multithreading_helper.h` line 64 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/976#-NVWIHNqETvY6u399ykj:-NVWIHNqETvY6u399ykk:b-gpmv1f) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/93043f49201ad1f07e71efbb36e8f4dc110c7a47/src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/multithreading_helper.h#L64)):*
> ```objc
>       absl::string_view local_pohlig_hellman_private_key,
>       const ElGamalCiphertext& composite_el_gamal_public_key,
>       const ElGamalCiphertext& partial_composite_el_gamal_public_key);
> ```

Take in a `ProtocolCryptorOptions`.



_Code quote:_
```objc
      int num_threads, int curve_id,
      const ElGamalCiphertext& local_el_gamal_public_key,
      absl::string_view local_el_gamal_private_key,
      absl::string_view local_pohlig_hellman_private_key,
      const ElGamalCiphertext& composite_el_gamal_public_key,
      const ElGamalCiphertext& partial_composite_el_gamal_public_key);
```
___
*[`src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/multithreading_helper.h` line 43 at r9](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/976#-NVWU9jqEIxOtkNbkQQH:-NVWU9jqEIxOtkNbkQQI:b-e6i4th) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8f07a3849aa4a8b19986570acbd20fe5c7da3751/src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/multithreading_helper.h#L43)):*
> ```objc
>   // Create [ProtocolCryptor] for threads.
>   //
>   // As [ProtocolCryptor] is not thread-safe, each thread has its own cryptor to
> ```

It's not? I'm looking at `ProtocolCryptorImpl` and it specifically has a mutex to ensure thread safety. Aside from that mutex, it looks like all of its members are `const`.



_Code quote:_
```objc
As [ProtocolCryptor] is not thread-safe
```
___
*[`src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/multithreading_helper.cc` line 27 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/976#-NVWPeRnBq8V8y8cKcE9:-NVWPeRnBq8V8y8cKcEA:br762n) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/93043f49201ad1f07e71efbb36e8f4dc110c7a47/src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/multithreading_helper.cc#L27)):*
> ```cpp
> using ::wfa::measurement::common::crypto::ProtocolCryptorOptions;
> 
> absl::StatusOr<std::unique_ptr<MultithreadingHelper>>
> ```

Does this need to be heap-allocated? As a reminder, we avoid heap allocation unless it's necessary (e.g. lifetime considerations or the object has a large memory footprint). See https://abseil.io/tips/123



_Suggestion:_
```cpp
absl::StatusOr<MultithreadingHelper>
```
___
*[`src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/multithreading_helper.cc` line 46 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/976#-NVWOhiiBIHlbjOcyN-8:-NVWOhiiBIHlbjOcyN-9:b-90u1ic) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/93043f49201ad1f07e71efbb36e8f4dc110c7a47/src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/multithreading_helper.cc#L46)):*
> ```cpp
>   std::unique_ptr<MultithreadingHelper> helper = absl::WrapUnique(
>       new MultithreadingHelper(num_threads, std::move(cryptors)));
>   return {std::move(helper)};
> ```

1. absl::WrapUnique is only for cases where we're calling legacy code that uses raw pointers where we also want type inference. Our code should never use `new`. See https://abseil.io/tips/126
2. Using std::move in a return statement is rarely what you want. It kills RVO. Explanation (Google-internal): http://go/totw/labs/should-i-return-std-move



_Suggestion:_
```cpp
  return std::make_unique<MultithreadingHelper>(num_threads, std::move(cryptors));
```

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/976)* status: 6 of 10 files reviewed, 9 unresolved discussions (waiting on @kungfucraig, @ple13, @SanjayVas, and @stevenwarejones)

___
*[`src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/multithreading_helper.h` line 64 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/976#-NVWIHNqETvY6u399ykj:-NVWgv9o5mdSO3_lJdgx:bldouxa) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/93043f49201ad1f07e71efbb36e8f4dc110c7a47/src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/multithreading_helper.h#L64)):*

Take in a `ProtocolCryptorOptions`.
</blockquote></details>

Done.
That's right.
___
*[`src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/multithreading_helper.h` line 43 at r9](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/976#-NVWU9jqEIxOtkNbkQQH:-NVWWLShAtAe-2hcMqmL:b-glo2r6) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8f07a3849aa4a8b19986570acbd20fe5c7da3751/src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/multithreading_helper.h#L43)):*

It's not? I'm looking at `ProtocolCryptorImpl` and it specifically has a mutex to ensure thread safety. Aside from that mutex, it looks like all of its members are `const`.
</blockquote></details>

Thanks for looking into it.
The mutex used for its function is pretty brutal force that it lock the object until the entire function finishes. It is required because the underlying library `private_join_and_compute` is not thread-safe. However, this impact the parallelism of execution if we really want to run cryptor operations in parallel. 
From the suggestion of team `private_join_and_compute`, using multiple instances of `private_join_and_compute` is the way they adopted already.
___
*[`src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/multithreading_helper.cc` line 27 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/976#-NVWPeRnBq8V8y8cKcE9:-NVWZQ9f79G0pJNlJp1a:b-e11eyr) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/93043f49201ad1f07e71efbb36e8f4dc110c7a47/src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/multithreading_helper.cc#L27)):*

Does this need to be heap-allocated? As a reminder, we avoid heap allocation unless it's necessary (e.g. lifetime considerations or the object has a large memory footprint). See https://abseil.io/tips/123
</blockquote></details>

Thanks. I was trying to return a pure MultithreadingHelper instance. However the compiler was complaining regarding the vector<unique_ptr<ProtocolCryptor>> if I just return MultithreadingHelper.

error: static_assert failed due to requirement 'is_constructible<std::unique_ptr<wfa::measurement::common::crypto::ProtocolCryptor, std::default_delete<wfa::measurement::common::crypto::ProtocolCryptor>>, const std::unique_ptr<wfa::measurement::common::crypto::ProtocolCryptor, std::default_delete<wfa::measurement::common::crypto::ProtocolCryptor>> &>::value' "result type must be constructible from input type"

Do you see any solution?
___
*[`src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/multithreading_helper.cc` line 46 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/976#-NVWOhiiBIHlbjOcyN-8:-NVWY35D0SWM_hSGUh38:b-gxrvlu) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/93043f49201ad1f07e71efbb36e8f4dc110c7a47/src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/multithreading_helper.cc#L46)):*

1. absl::WrapUnique is only for cases where we're calling legacy code that uses raw pointers where we also want type inference. Our code should never use `new`. See https://abseil.io/tips/126
2. Using std::move in a return statement is rarely what you want. It kills RVO. Explanation (Google-internal): http://go/totw/labs/should-i-return-std-move

</blockquote></details>

As the constructor of MultithreadingHelper is private, it is not allowed to create it by std::make_unique. The way to bypass that is using `new` with absl::WrapUnique.

ple13 (approved):
> Reviewed 3 of 6 files at r9, 3 of 3 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/976)* status: all files reviewed, 6 unresolved discussions (waiting on @kungfucraig, @SanjayVas, and @stevenwarejones)

stevenwarejones (commented):
> Reviewed 4 of 7 files at r1, 3 of 6 files at r9, 2 of 3 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/976)* status: all files reviewed, 4 unresolved discussions (waiting on @SanjayVas)

___
*[`src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/multithreading_helper.h` line 43 at r9](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/976#-NVWU9jqEIxOtkNbkQQH:-NV_5msPC61jLD943OUe:b3i82kg) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8f07a3849aa4a8b19986570acbd20fe5c7da3751/src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/multithreading_helper.h#L43)):*

Thanks for looking into it.
The mutex used for its function is pretty brutal force that it lock the object until the entire function finishes. It is required because the underlying library `private_join_and_compute` is not thread-safe. However, this impact the parallelism of execution if we really want to run cryptor operations in parallel. 
From the suggestion of team `private_join_and_compute`, using multiple instances of `private_join_and_compute` is the way they adopted already.
</blockquote></details>

Can you add this as a comment?

SanjayVas (approved):
> Reviewed 1 of 7 files at r1, 1 of 6 files at r9, 3 of 3 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/976)* status: all files reviewed, 2 unresolved discussions (waiting on @kungfucraig and @renjiezh)

___
*[`src/main/cc/wfa/measurement/common/crypto/protocol_cryptor.h` line 40 at r10](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/976#-NVa1r9t-longxCwfqSM:-NVa1r9t-longxCwfqSN:b-k0mlhr) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7154ed33622931581dffec443fab1c76d25aa0bf/src/main/cc/wfa/measurement/common/crypto/protocol_cryptor.h#L40)):*
> ```objc
> enum CompositeType { kFull, kPartial };
> 
> // A cryptor dealing with basic operations in various MPC protocols.
> ```

nit: Document that implementations are not thread safe.
___
*[`src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/multithreading_helper.cc` line 27 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/976#-NVWPeRnBq8V8y8cKcE9:-NVa22kX2tblsJb1RObt:bcjr42p) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/93043f49201ad1f07e71efbb36e8f4dc110c7a47/src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/multithreading_helper.cc#L27)):*

Thanks. I was trying to return a pure MultithreadingHelper instance. However the compiler was complaining regarding the vector<unique_ptr<ProtocolCryptor>> if I just return MultithreadingHelper.

error: static_assert failed due to requirement 'is_constructible<std::unique_ptr<wfa::measurement::common::crypto::ProtocolCryptor, std::default_delete<wfa::measurement::common::crypto::ProtocolCryptor>>, const std::unique_ptr<wfa::measurement::common::crypto::ProtocolCryptor, std::default_delete<wfa::measurement::common::crypto::ProtocolCryptor>> &>::value' "result type must be constructible from input type"

Do you see any solution?

</blockquote></details>

Ah, I think that member may make the type not copyable. That's why you need the unique_ptr here. Leave as-is.
___
*[`src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/multithreading_helper.cc` line 46 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/976#-NVWOhiiBIHlbjOcyN-8:-NVa2iUi77TQDYwoMJzE:b-rq0t2g) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/93043f49201ad1f07e71efbb36e8f4dc110c7a47/src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/multithreading_helper.cc#L46)):*

As the constructor of MultithreadingHelper is private, it is not allowed to create it by std::make_unique. The way to bypass that is using `new` with absl::WrapUnique.
</blockquote></details>

Would you mind adding that as a comment here? e.g.

```kotlin
// Use `new` rather than `make_unique` due to constructor being private.
```

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/976)* status: 7 of 10 files reviewed, all discussions resolved (waiting on @kungfucraig, @ple13, @SanjayVas, and @stevenwarejones)

___
*[`src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/multithreading_helper.h` line 43 at r9](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/976#-NVWU9jqEIxOtkNbkQQH:-NVaKx800u0ZZAqIoG7s:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8f07a3849aa4a8b19986570acbd20fe5c7da3751/src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/multithreading_helper.h#L43)):*

Can you add this as a comment?
</blockquote></details>

Done.

stevenwarejones (approved):
> Reviewed 3 of 3 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/976)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @kungfucraig)

___
*[`src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/multithreading_helper.h` line 43 at r9](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/976#-NVWU9jqEIxOtkNbkQQH:-NVaLUhz5oEsqFxbubYk:b-h8giev) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8f07a3849aa4a8b19986570acbd20fe5c7da3751/src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/multithreading_helper.h#L43)):*

Done.
</blockquote></details>

nit: `is a precaution`

SanjayVas (approved):
> Reviewed 3 of 3 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/976)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @kungfucraig)

renjiezh (commented):
> Reviewed 4 of 7 files at r1, 2 of 6 files at r9, 1 of 3 files at r10, 2 of 3 files at r11, 1 of 1 files at r12, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/976)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @kungfucraig)

PR #983, opened 2023-05-08
Add type registry and cel env caching.

tristanvuong2021:


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/983)

SanjayVas (changes requested):
> Reviewed 1 of 7 files at r1, 2 of 4 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/983)* status: 3 of 7 files reviewed, 5 unresolved discussions (waiting on @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/common/server/ReportingApiServerFlags.kt` line 36 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/983#-NUwQcAo9tlEvLgM1tSC:-NUwQcAp1cCwjnB_WLR6:b-9yj64i) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/13c33222ac9452c29615593d730ac91ddfed02a7/src/main/kotlin/org/wfanet/measurement/reporting/deploy/common/server/ReportingApiServerFlags.kt#L36)):*
> ```Kotlin
> 
>   @CommandLine.Option(
>     names = ["--list-event-groups-cache-refresh-interval"],
> ```

nit: maybe `--event-group-metadata-descriptor-cache-duration`, since this is really answering "How long do I cache EventGroupMetadataDescriptors for (before refreshing)?"
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v1alpha/EventGroupsService.kt` line 229 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/983#-NUwRg3kBjbYbZi9IqJz:-NUwRg3kBjbYbZi9IqK-:b-udv4tv) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/13c33222ac9452c29615593d730ac91ddfed02a7/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v1alpha/EventGroupsService.kt#L229)):*
> ```Kotlin
>   }
> 
>   private class ListEventGroupsCache(
> ```

This is technically a cache of the CEL environment (and by extension, the TypeRegistry used to build the environment which in turn comes from EventGroupMetadataDescriptors). The caching functionality isn't the salient point to the service, but rather the fact that it provides the environment.

I'd refactor this as follows: Extract an interface for the CEL env provider that this class implements. The service should take in an instance of the interface, since that's all it cares about. This means moving the implementation so it's not a nested class and the caller of the service constructor can pick the implementation.

Finally, write a test for the caching provider implementation.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v1alpha/EventGroupsService.kt` line 238 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/983#-NUwTvg_ALZFPqe9cYTX:-NUwTvg_ALZFPqe9cYTY:b-sc7aqf) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/13c33222ac9452c29615593d730ac91ddfed02a7/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v1alpha/EventGroupsService.kt#L238)):*
> ```Kotlin
>     )
> 
>     private lateinit var typeRegistryAndEnv: TypeRegistryAndEnv
> ```

Access to this needs to be guarded for thread safety so e.g. it doesn't get a ready in the middle of being set. A [kotlinx.coroutines.sync.Mutex](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.sync/-mutex/) should do the job. See https://kotlinlang.org/docs/shared-mutable-state-and-concurrency.html#mutual-exclusion
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v1alpha/EventGroupsService.kt` line 241 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/983#-NUwTjnmCmNzC6eyQL73:-NUwTjnmCmNzC6eyQL74:b8ajalh) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/13c33222ac9452c29615593d730ac91ddfed02a7/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v1alpha/EventGroupsService.kt#L241)):*
> ```Kotlin
> 
>     init {
>       CoroutineScope(Dispatchers.Default + SupervisorJob()).launch {
> ```

Take in the CoroutineContext (dispatcher) as a constructor parameter.



_Code quote:_
```Kotlin
Dispatchers.Default
```
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v1alpha/EventGroupsService.kt` line 242 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/983#-NUwVDKWAnrOPvlwV3Sp:-NUwVDKWAnrOPvlwV3Sq:b3ald4l) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/13c33222ac9452c29615593d730ac91ddfed02a7/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v1alpha/EventGroupsService.kt#L242)):*
> ```Kotlin
>     init {
>       CoroutineScope(Dispatchers.Default + SupervisorJob()).launch {
>         MinimumIntervalThrottler(Clock.systemUTC(), cacheRefreshInterval).loopOnReady {
> ```

Take in the Clock instance as a constructor parameter.



_Code quote:_
```Kotlin
Clock.systemUTC()
```

tristanvuong2021 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/983)* status: 0 of 15 files reviewed, 4 unresolved discussions (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v1alpha/EventGroupsService.kt` line 229 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/983#-NUwRg3kBjbYbZi9IqJz:-NUyWOm30SKcPvfrPMeL:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/13c33222ac9452c29615593d730ac91ddfed02a7/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v1alpha/EventGroupsService.kt#L229)):*

This is technically a cache of the CEL environment (and by extension, the TypeRegistry used to build the environment which in turn comes from EventGroupMetadataDescriptors). The caching functionality isn't the salient point to the service, but rather the fact that it provides the environment.

I'd refactor this as follows: Extract an interface for the CEL env provider that this class implements. The service should take in an instance of the interface, since that's all it cares about. This means moving the implementation so it's not a nested class and the caller of the service constructor can pick the implementation.

Finally, write a test for the caching provider implementation.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v1alpha/EventGroupsService.kt` line 238 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/983#-NUwTvg_ALZFPqe9cYTX:-NUwm7yV5Q_9rO4TO9VL:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/13c33222ac9452c29615593d730ac91ddfed02a7/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v1alpha/EventGroupsService.kt#L238)):*

Access to this needs to be guarded for thread safety so e.g. it doesn't get a ready in the middle of being set. A [kotlinx.coroutines.sync.Mutex](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.sync/-mutex/) should do the job. See https://kotlinlang.org/docs/shared-mutable-state-and-concurrency.html#mutual-exclusion
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v1alpha/EventGroupsService.kt` line 241 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/983#-NUwTjnmCmNzC6eyQL73:-NUwf2FM4pn8lHSqiyvU:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/13c33222ac9452c29615593d730ac91ddfed02a7/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v1alpha/EventGroupsService.kt#L241)):*

Take in the CoroutineContext (dispatcher) as a constructor parameter.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v1alpha/EventGroupsService.kt` line 242 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/983#-NUwVDKWAnrOPvlwV3Sp:-NUwkS9m07co5K0_Ox_o:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/13c33222ac9452c29615593d730ac91ddfed02a7/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v1alpha/EventGroupsService.kt#L242)):*

Take in the Clock instance as a constructor parameter.
</blockquote></details>

Done.

tristanvuong2021 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/983)* status: 9 of 15 files reviewed, 4 unresolved discussions (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/CelEnvProvider.kt` line 65 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/983#-NV0SB5S7E-eBd_9ofKk:-NVoT_MUCrUPgeDQeiJy:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e2e25503b1967e3eb6755bc381b830d0d5b7e402/src/main/kotlin/org/wfanet/measurement/reporting/service/api/CelEnvProvider.kt#L65)):*

The caller should be able to pass in an entire context. It just happens that the main useful thing to pass is the dispatcher, which is a context element and therefore also implements CoroutineContext.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/CelEnvProvider.kt` line 80 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/983#-NV0T_Yw1V8c_FzkwgMx:-NVoTa_p6QpFM3Lzpchx:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e2e25503b1967e3eb6755bc381b830d0d5b7e402/src/main/kotlin/org/wfanet/measurement/reporting/service/api/CelEnvProvider.kt#L80)):*

It's unsafe to just retry on any runtime exception. Actually check whether this is an exception that we'd consider retriable (this may involve wrapping the StatusException with our own Exception type which indicates this).

See https://grpc.github.io/grpc/core/md_doc_statuscodes.html for information on which codes are safe to retry on. At first glance since we're calling an idempotent method I'd say UNAVAILABLE and DEADLINE_EXCEEDED.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/CelEnvProvider.kt` line 182 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/983#-NV0TloM84Q2BFhVLgKq:-NVoTdOA3o6AjL5uV7bg:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e2e25503b1967e3eb6755bc381b830d0d5b7e402/src/main/kotlin/org/wfanet/measurement/reporting/service/api/CelEnvProvider.kt#L182)):*

Only service methods throw StatusRuntimeException. This is being called in a background coroutine and not directly by a service method.
</blockquote></details>

Done.
___
*[`src/test/kotlin/org/wfanet/measurement/reporting/service/api/CelEnvProviderTest.kt` line 171 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/983#-NV0Vc3SBKbbIzCSGvR7:-NW8zKGNCEw36MQm_asi:bttuo77) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e2e25503b1967e3eb6755bc381b830d0d5b7e402/src/test/kotlin/org/wfanet/measurement/reporting/service/api/CelEnvProviderTest.kt#L171)):*

Avoid introducing real delays in test code. See https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-test/kotlinx.coroutines.test/run-test.html
</blockquote></details>

Got rid of delays and can get the tests to work individually, but they time out when the entire test class is run at once.

SanjayVas (changes requested):
> Reviewed 1 of 3 files at r5, 3 of 5 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/983)* status: 13 of 15 files reviewed, 2 unresolved discussions (waiting on @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/CelEnvProvider.kt` line 65 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/983#-NV0SB5S7E-eBd_9ofKk:-NW9_dR25q8OQzBd10st:bnl900y) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e2e25503b1967e3eb6755bc381b830d0d5b7e402/src/main/kotlin/org/wfanet/measurement/reporting/service/api/CelEnvProvider.kt#L65)):*

Done.
</blockquote></details>

Not the CoroutineScope, but the CoroutineContext. The scope should be managed by `CelEnvCacheProvider`, which should implement `AutoCloseable` and cancel the coroutine scope on `close`.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/CelEnvProvider.kt` line 80 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/983#-NV0T_Yw1V8c_FzkwgMx:-NW9afY28VYslUI3iBF-:b-g5k736) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e2e25503b1967e3eb6755bc381b830d0d5b7e402/src/main/kotlin/org/wfanet/measurement/reporting/service/api/CelEnvProvider.kt#L80)):*

Done.
</blockquote></details>

StatusException should always be caught at the RPC site. Check the conditions there and wrap in a custom exception to indicate that it's retriable.

Another way to think of this is that the caller of `setTypeRegistryAndEnv` has no idea what RPCs may be called and therefore whether a given status code is retriable.

tristanvuong2021 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/983)* status: 10 of 15 files reviewed, 2 unresolved discussions (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/CelEnvProvider.kt` line 65 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/983#-NV0SB5S7E-eBd_9ofKk:-NWAJXjU1-xs_Pvkp-FN:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e2e25503b1967e3eb6755bc381b830d0d5b7e402/src/main/kotlin/org/wfanet/measurement/reporting/service/api/CelEnvProvider.kt#L65)):*

Not the CoroutineScope, but the CoroutineContext. The scope should be managed by `CelEnvCacheProvider`, which should implement `AutoCloseable` and cancel the coroutine scope on `close`.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/CelEnvProvider.kt` line 80 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/983#-NV0T_Yw1V8c_FzkwgMx:-NWAJWo8-KnlZvxsTFm8:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e2e25503b1967e3eb6755bc381b830d0d5b7e402/src/main/kotlin/org/wfanet/measurement/reporting/service/api/CelEnvProvider.kt#L80)):*

StatusException should always be caught at the RPC site. Check the conditions there and wrap in a custom exception to indicate that it's retriable.

Another way to think of this is that the caller of `setTypeRegistryAndEnv` has no idea what RPCs may be called and therefore whether a given status code is retriable.
</blockquote></details>

Done.

tristanvuong2021 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/983)* status: 13 of 15 files reviewed, 2 unresolved discussions (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/CelEnvProvider.kt` line 81 at r11](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/983#-NWJ-DXx9khM8zVNWl5I:-NWJw3p355aJ6yilRAIY:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/cb21427eee4f0e88964f16bf0506ab7c8af1b2fd/src/main/kotlin/org/wfanet/measurement/reporting/service/api/CelEnvProvider.kt#L81)):*

Don't need to do `async`/`await` here with a job since we're already in a suspending block.

I think what you want is something like

```kotlin
private val initialSyncJob: CompletableJob = Job()
```

Inside of the loopOnReady block you can call either `initialSyncJob.complete()` or `initialSyncJob.completeExceptionally(e)` depending on whether it succeeded or failed. On failure you should also throw the exception here to ensure that the coroutine scope gets cancelled. You can then call `initialSyncJob.join()` from `getTypeRegistryAndEnv`.

Also I'd only do retries on the initial sync (check `initialSyncJob.isCompleted`). After that it's okay to just log a warning on failure and let the next iteration try.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/CelEnvProvider.kt` line 117 at r11](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/983#-NWIyno84Ax91q8wQuR8:-NWJw4RVAaPM4igd4-gw:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/cb21427eee4f0e88964f16bf0506ab7c8af1b2fd/src/main/kotlin/org/wfanet/measurement/reporting/service/api/CelEnvProvider.kt#L117)):*

I assume you're still working on this part? Otherwise

1. Shouldn't need delays
2. Shouldn't need multiple attempts.

The logic should just be to suspend until the initial sync is complete and then return the value.
</blockquote></details>

Done.

SanjayVas (changes requested):
> Reviewed 2 of 2 files at r12, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/983)* status: all files reviewed, 8 unresolved discussions (waiting on @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/CelEnvProvider.kt` line 95 at r12](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/983#-NWO04v_9hOyiQ4MMc1I:-NWO04v_9hOyiQ4MMc1J:b-5b4mm8) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d421eb1b64bbed96accd5bb6564e1e7357c22fac/src/main/kotlin/org/wfanet/measurement/reporting/service/api/CelEnvProvider.kt#L95)):*
> ```Kotlin
>           }
>         } else {
>           val updateFlow = flow<Unit> { setTypeRegistryAndEnv() }
> ```

No need to use a flow here. Just call the suspending function.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/CelEnvProvider.kt` line 99 at r12](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/983#-NWO0GMYDEsHlSEQhMEa:-NWO0GMZ6W4VijkhNShI:undefined) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d421eb1b64bbed96accd5bb6564e1e7357c22fac/src/main/kotlin/org/wfanet/measurement/reporting/service/api/CelEnvProvider.kt#L99)):*
> ```Kotlin
>             updateFlow.collect {}
>           } catch (e: Exception) {
>             logger.warning(e.stackTraceToString())
> ```





_Suggestion:_
```Kotlin
logger.log(Level.WARNING, e) { "Error updating CEL env cache" }
```
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/CelEnvProvider.kt` line 116 at r12](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/983#-NWO0dx3-KAmxYVmOa0x:-NWO0dx3-KAmxYVmOa0y:bjr2mqh) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d421eb1b64bbed96accd5bb6564e1e7357c22fac/src/main/kotlin/org/wfanet/measurement/reporting/service/api/CelEnvProvider.kt#L116)):*
> ```Kotlin
>       initialSyncJob.join()
>       typeRegistryAndEnv
>     }
> ```

You can just unconditionally join the initial sync job.



_Suggestion:_
```Kotlin
    initialSyncJob.join()
    return typeRegistryAndEnv
```
___
*[`src/test/kotlin/org/wfanet/measurement/reporting/service/api/CelEnvProviderTest.kt` line 67 at r12](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/983#-NWO17yb-7U9mxgrHe2R:-NWO17yb-7U9mxgrHe2S:btkpnuy) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d421eb1b64bbed96accd5bb6564e1e7357c22fac/src/test/kotlin/org/wfanet/measurement/reporting/service/api/CelEnvProviderTest.kt#L67)):*
> ```Kotlin
>     mockService {
>       onBlocking { listEventGroupMetadataDescriptors(any()) }
>         .thenThrow(Status.DEADLINE_EXCEEDED.asRuntimeException())
> ```

Generally when we configure the stubbing inside of `mockService` it's because we have some default stubbing that will work for most if not all test methods. This looks like it only applies to one test method.
___
*[`src/test/kotlin/org/wfanet/measurement/reporting/service/api/CelEnvProviderTest.kt` line 103 at r12](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/983#-NWO25gY4lxqopm2CapB:-NWO25gY4lxqopm2CapC:byqtioo) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d421eb1b64bbed96accd5bb6564e1e7357c22fac/src/test/kotlin/org/wfanet/measurement/reporting/service/api/CelEnvProviderTest.kt#L103)):*
> ```Kotlin
> 
>             verify(cmmsEventGroupMetadataDescriptorsServiceMock, times(2))
>               .listEventGroupMetadataDescriptors(eventGroupMetadataDescriptorsCaptor.capture())
> ```

You're not doing anything with the captured requests. You're also not verifying the result of `getTypeRegistryAndEnv()`



_Code quote:_
```Kotlin
            verify(cmmsEventGroupMetadataDescriptorsServiceMock, times(2))
              .listEventGroupMetadataDescriptors(eventGroupMetadataDescriptorsCaptor.capture())
```
___
*[`src/test/kotlin/org/wfanet/measurement/reporting/service/api/CelEnvProviderTest.kt` line 109 at r12](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/983#-NWO1e7-1q1tjVyUIfCY:-NWO1e709lyaidCWicHa:b6036aj) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d421eb1b64bbed96accd5bb6564e1e7357c22fac/src/test/kotlin/org/wfanet/measurement/reporting/service/api/CelEnvProviderTest.kt#L109)):*
> ```Kotlin
>       }
>     } catch (e: Throwable) {
>       println(e.stackTraceToString())
> ```

You shouldn't need to catch exceptions in a test method.



_Code quote:_
```Kotlin
    } catch (e: Throwable) {
      println(e.stackTraceToString())
```
___
*[`src/test/kotlin/org/wfanet/measurement/reporting/service/api/CelEnvProviderTest.kt` line 111 at r12](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/983#-NWO1s3p5-nPt9FVR8AV:-NWO1s3p5-nPt9FVR8AW:bmt9o1s) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d421eb1b64bbed96accd5bb6564e1e7357c22fac/src/test/kotlin/org/wfanet/measurement/reporting/service/api/CelEnvProviderTest.kt#L111)):*
> ```Kotlin
>       println(e.stackTraceToString())
>     } finally {
>       assertThat(verified).isTrue()
> ```

You shouldn't need this. `runTest` works like `runBlocking`, so the code inside that block will have run by this point.



_Code quote:_
```Kotlin
assertThat(verified).isTrue()
```
___
*[`src/test/kotlin/org/wfanet/measurement/reporting/service/api/CelEnvProviderTest.kt` line 153 at r12](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/983#-NWOKjG6Dr7r2XswgRbb:-NWOKjG6Dr7r2XswgRbc:b-ccxijn) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d421eb1b64bbed96accd5bb6564e1e7357c22fac/src/test/kotlin/org/wfanet/measurement/reporting/service/api/CelEnvProviderTest.kt#L153)):*
> ```Kotlin
>       assertThat(verified).isTrue()
>     }
>   }
> ```

Add a test to make sure it refreshes again after the interval.

tristanvuong2021 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/983)* status: all files reviewed, 8 unresolved discussions (waiting on @SanjayVas)

___
*[`src/test/kotlin/org/wfanet/measurement/reporting/service/api/CelEnvProviderTest.kt` line 109 at r12](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/983#-NWO1e7-1q1tjVyUIfCY:-NWOQ3ilEv-nWtaATl2K:b-d1lrhc) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d421eb1b64bbed96accd5bb6564e1e7357c22fac/src/test/kotlin/org/wfanet/measurement/reporting/service/api/CelEnvProviderTest.kt#L109)):*

You shouldn't need to catch exceptions in a test method.
</blockquote></details>

The problem is that calling close with use also cancels the TestScope so an exception is always thrown on close here. Either I have use with a try catch, or I call close after everything is done and don't use use.
___
*[`src/test/kotlin/org/wfanet/measurement/reporting/service/api/CelEnvProviderTest.kt` line 153 at r12](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/983#-NWOKjG6Dr7r2XswgRbb:-NWOQKvd3XOqWpy0mpf2:blrl9he) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d421eb1b64bbed96accd5bb6564e1e7357c22fac/src/test/kotlin/org/wfanet/measurement/reporting/service/api/CelEnvProviderTest.kt#L153)):*

Add a test to make sure it refreshes again after the interval.
</blockquote></details>

I couldn't figure out how to do that without a delay so I removed it. If I kill it on the initial sync when there is an exception, then that means the initial sync has to succeed. If it succeeds, then there is no way to wait for the next run without a delay.

tristanvuong2021 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/983)* status: all files reviewed, 8 unresolved discussions (waiting on @SanjayVas)

___
*[`src/test/kotlin/org/wfanet/measurement/reporting/service/api/CelEnvProviderTest.kt` line 109 at r12](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/983#-NWO1e7-1q1tjVyUIfCY:-NWOVSOqAwMUzqoe-faA:b-xqswj8) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d421eb1b64bbed96accd5bb6564e1e7357c22fac/src/test/kotlin/org/wfanet/measurement/reporting/service/api/CelEnvProviderTest.kt#L109)):*

The problem is that calling close with use also cancels the TestScope so an exception is always thrown on close here. Either I have use with a try catch, or I call close after everything is done and don't use use.
</blockquote></details>

Maybe I could just wrap the use block in a try catch

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/983)* status: all files reviewed, 9 unresolved discussions (waiting on @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/CelEnvProvider.kt` line 120 at r12](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/983#-NWOWb3uEAcU3AcKxTAS:-NWOWb3uEAcU3AcKxTAT:bsxznel) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d421eb1b64bbed96accd5bb6564e1e7357c22fac/src/main/kotlin/org/wfanet/measurement/reporting/service/api/CelEnvProvider.kt#L120)):*
> ```Kotlin
> 
>   private suspend fun setTypeRegistryAndEnv() {
>     mutex.withLock {
> ```

I realized we don't actually need this as long as `setTypeRegistryAndEnv` is only ever called by MinimumIntervalThrottler, since that has its own mutex.



_Code quote:_
```Kotlin
mutex.withLock {
```
___
*[`src/test/kotlin/org/wfanet/measurement/reporting/service/api/CelEnvProviderTest.kt` line 109 at r12](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/983#-NWO1e7-1q1tjVyUIfCY:-NWOScGQ0WRN9Ua-8Ksx:b-5p9yox) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d421eb1b64bbed96accd5bb6564e1e7357c22fac/src/test/kotlin/org/wfanet/measurement/reporting/service/api/CelEnvProviderTest.kt#L109)):*

The problem is that calling close with use also cancels the TestScope so an exception is always thrown on close here. Either I have use with a try catch, or I call close after everything is done and don't use use.
</blockquote></details>

I think you accidentally removed the SupervisorJob() from when you construct the CoroutineScope. That should prevent the cancellation of the cache sync scope from also cancelling the parent test scope.
___
*[`src/test/kotlin/org/wfanet/measurement/reporting/service/api/CelEnvProviderTest.kt` line 153 at r12](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/983#-NWOKjG6Dr7r2XswgRbb:-NWOTiBq5iDnjiydjpNO:b-q0swm4) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d421eb1b64bbed96accd5bb6564e1e7357c22fac/src/test/kotlin/org/wfanet/measurement/reporting/service/api/CelEnvProviderTest.kt#L153)):*

I couldn't figure out how to do that without a delay so I removed it. If I kill it on the initial sync when there is an exception, then that means the initial sync has to succeed. If it succeeds, then there is no way to wait for the next run without a delay.
</blockquote></details>

It should work with the following steps:

1. Stub the service so it succeeds on two sequential calls, returning different results for each.
1. Call getTypeRegistryAndEnv(), verifying that you get the first result.
1. Call `advanceTimeBy` so that the throttler's `delay` returns
    * You might need a `yield()` here as well, but I don't think so as we're using the same dispatcher
1. Call getTypeRegistryAndEnv() again, verifying that you get the second result

You already verified the request params in the first test method, so no need to do that again.

tristanvuong2021 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/983)* status: 13 of 15 files reviewed, 9 unresolved discussions (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/CelEnvProvider.kt` line 95 at r12](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/983#-NWO04v_9hOyiQ4MMc1I:-NWOu3RY5GVzNoQdBgOP:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d421eb1b64bbed96accd5bb6564e1e7357c22fac/src/main/kotlin/org/wfanet/measurement/reporting/service/api/CelEnvProvider.kt#L95)):*

No need to use a flow here. Just call the suspending function.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/CelEnvProvider.kt` line 99 at r12](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/983#-NWO0GMYDEsHlSEQhMEa:-NWOu34S2-2T6mLG08u7:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d421eb1b64bbed96accd5bb6564e1e7357c22fac/src/main/kotlin/org/wfanet/measurement/reporting/service/api/CelEnvProvider.kt#L99)):*
> ```Kotlin
>             updateFlow.collect {}
>           } catch (e: Exception) {
>             logger.warning(e.stackTraceToString())
> ```

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/CelEnvProvider.kt` line 116 at r12](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/983#-NWO0dx3-KAmxYVmOa0x:-NWOu-DtAuDmREvlY1MD:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d421eb1b64bbed96accd5bb6564e1e7357c22fac/src/main/kotlin/org/wfanet/measurement/reporting/service/api/CelEnvProvider.kt#L116)):*

You can just unconditionally join the initial sync job.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/CelEnvProvider.kt` line 120 at r12](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/983#-NWOWb3uEAcU3AcKxTAS:-NWOtzFm0qZsmDxRFF40:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d421eb1b64bbed96accd5bb6564e1e7357c22fac/src/main/kotlin/org/wfanet/measurement/reporting/service/api/CelEnvProvider.kt#L120)):*

I realized we don't actually need this as long as `setTypeRegistryAndEnv` is only ever called by MinimumIntervalThrottler, since that has its own mutex.
</blockquote></details>

Done.
___
*[`src/test/kotlin/org/wfanet/measurement/reporting/service/api/CelEnvProviderTest.kt` line 67 at r12](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/983#-NWO17yb-7U9mxgrHe2R:-NWPJe109n3m_yKx3oW8:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d421eb1b64bbed96accd5bb6564e1e7357c22fac/src/test/kotlin/org/wfanet/measurement/reporting/service/api/CelEnvProviderTest.kt#L67)):*

Generally when we configure the stubbing inside of `mockService` it's because we have some default stubbing that will work for most if not all test methods. This looks like it only applies to one test method.
</blockquote></details>

Done.
___
*[`src/test/kotlin/org/wfanet/measurement/reporting/service/api/CelEnvProviderTest.kt` line 103 at r12](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/983#-NWO25gY4lxqopm2CapB:-NWPJegV7ahUvSmqXBaN:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d421eb1b64bbed96accd5bb6564e1e7357c22fac/src/test/kotlin/org/wfanet/measurement/reporting/service/api/CelEnvProviderTest.kt#L103)):*

You're not doing anything with the captured requests. You're also not verifying the result of `getTypeRegistryAndEnv()`
</blockquote></details>

Done.
___
*[`src/test/kotlin/org/wfanet/measurement/reporting/service/api/CelEnvProviderTest.kt` line 109 at r12](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/983#-NWO1e7-1q1tjVyUIfCY:-NWOtxN44TXbC-mPhh5J:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d421eb1b64bbed96accd5bb6564e1e7357c22fac/src/test/kotlin/org/wfanet/measurement/reporting/service/api/CelEnvProviderTest.kt#L109)):*

I think you accidentally removed the SupervisorJob() from when you construct the CoroutineScope. That should prevent the cancellation of the cache sync scope from also cancelling the parent test scope.
</blockquote></details>

Done.
___
*[`src/test/kotlin/org/wfanet/measurement/reporting/service/api/CelEnvProviderTest.kt` line 111 at r12](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/983#-NWO1s3p5-nPt9FVR8AV:-NWOtwyl1Fx9YTF0_pV-:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d421eb1b64bbed96accd5bb6564e1e7357c22fac/src/test/kotlin/org/wfanet/measurement/reporting/service/api/CelEnvProviderTest.kt#L111)):*

You shouldn't need this. `runTest` works like `runBlocking`, so the code inside that block will have run by this point.
</blockquote></details>

Done.
___
*[`src/test/kotlin/org/wfanet/measurement/reporting/service/api/CelEnvProviderTest.kt` line 153 at r12](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/983#-NWOKjG6Dr7r2XswgRbb:-NWPJgg3BtoJs1n8Uo16:b-mvsvdw) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d421eb1b64bbed96accd5bb6564e1e7357c22fac/src/test/kotlin/org/wfanet/measurement/reporting/service/api/CelEnvProviderTest.kt#L153)):*

It should work with the following steps:

1. Stub the service so it succeeds on two sequential calls, returning different results for each.
1. Call getTypeRegistryAndEnv(), verifying that you get the first result.
1. Call `advanceTimeBy` so that the throttler's `delay` returns
    * You might need a `yield()` here as well, but I don't think so as we're using the same dispatcher
1. Call getTypeRegistryAndEnv() again, verifying that you get the second result

You already verified the request params in the first test method, so no need to do that again.
</blockquote></details>

This doesn't work. Besides the first get call, nothing waits on the result. AdvanceTimeBy just skips delays, but the rest still needs to run. Sometimes enough time passes, sometimes not enough time passes.

tristanvuong2021 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/983)* status: 14 of 15 files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

___
*[`src/test/kotlin/org/wfanet/measurement/reporting/service/api/CelEnvProviderTest.kt` line 153 at r12](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/983#-NWOKjG6Dr7r2XswgRbb:-NXLWG-yBfF_PVc7O_fV:bj3b3uj) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d421eb1b64bbed96accd5bb6564e1e7357c22fac/src/test/kotlin/org/wfanet/measurement/reporting/service/api/CelEnvProviderTest.kt#L153)):*

Okay, so then we have a behavioral question: Should `getTypeRegistryAndEnv()` always wait for any pending sync operations to complete? If so, then it's simple: just wrap a mutex around both the set and get.

If we don't want this behavior (i.e. `getTypeRegistryAndEnv()` should just return whatever it can without waiting aside from the initial sync), then we'll need a separate method that specifically waits for any pending sync to complete.
</blockquote></details>

Ending up having to use real delays for the third test.

SanjayVas (approved):
> Reviewed 3 of 3 files at r15, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/983)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @tristanvuong2021)

___
*[`src/test/kotlin/org/wfanet/measurement/reporting/service/api/CelEnvProviderTest.kt` line 153 at r12](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/983#-NWOKjG6Dr7r2XswgRbb:-NXNOSBq9QSVfVinvc1J:b-5btr3) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d421eb1b64bbed96accd5bb6564e1e7357c22fac/src/test/kotlin/org/wfanet/measurement/reporting/service/api/CelEnvProviderTest.kt#L153)):*

Ending up having to use real delays for the third test.
</blockquote></details>

I pushed a commit that gets it working without real delays. I ended up dropping MinimumIntervalThrottler so I could access the Mutex. I figured we don't need the minimum interval guarantee and can just wait the full interval each time.

stevenwarejones (approved):
> Reviewed 1 of 7 files at r1, 7 of 14 files at r3, 1 of 3 files at r15, 3 of 3 files at r16, 2 of 3 files at r17, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/983)* status: 14 of 15 files reviewed, all discussions resolved (waiting on @Marco-Premier and @SanjayVas)

stevenwarejones (approved):
> Reviewed 1 of 3 files at r17, 1 of 1 files at r18, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/983)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Marco-Premier)

Marco-Premier (approved):
> Reviewed 1 of 7 files at r1, 7 of 14 files at r3, 1 of 3 files at r15, 2 of 3 files at r16, 1 of 1 files at r18, 3 of 3 files at r19, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/983)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @tristanvuong2021)

tristanvuong2021 (commented):
> Reviewed 1 of 7 files at r1, 6 of 14 files at r3, 1 of 3 files at r15, 2 of 3 files at r16, 1 of 1 files at r18, 3 of 3 files at r19, 1 of 1 files at r20, 2 of 2 files at r21, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/983)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @tristanvuong2021)

PR #1030, opened 2023-05-26
Adding GRPC for the Reporting BFF

bdomen-ggl:
Adding GRPC for the UI server - Reporting BFF. The backend to frontend mapping isn't done since we're still working out the details of the API.


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1030)

SanjayVas (commented):
> Reviewed 10 of 10 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1030)* status: all files reviewed, 3 unresolved discussions (waiting on @bdomen-ggl)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/bff/deploy/V1AlphaPublicApiServer.kt` line 96 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1030#-NWmll3d7xIwIYpAIJkY:-NWmll3d7xIwIYpAIJkZ:b3wkgtj) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/428229c5be78ed36ffb28db0d8618af4ac4da61a/src/main/kotlin/org/wfanet/measurement/reporting/bff/deploy/V1AlphaPublicApiServer.kt#L96)):*
> ```Kotlin
>     private set
> 
>   @CommandLine.Option(
> ```

Hmm, just an observation that because the Reporting API expects MC end-user auth (i.e. a client cert bound to a specific MC) then if the BFF calls that API then it needs to hold the private keys to those certs. This makes me wonder if we want to effectively have an auth bypass for the BFF, e.g. running a different Reporting server binary with different auth interceptors that can only be accessed from the same cluster.

@kungfucraig WDYT?
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/bff/service/api/v1alpha/BUILD.bazel` line 23 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1030#-NWmlb1n0ZlVROCYKTif:-NWmlb1n0ZlVROCYKTig:bbv4ahe) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/428229c5be78ed36ffb28db0d8618af4ac4da61a/src/main/kotlin/org/wfanet/measurement/reporting/bff/service/api/v1alpha/BUILD.bazel#L23)):*
> ```Starlark
> )
> 
> kt_jvm_library(
> ```

I think you have the server targets in two places.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/bff/service/api/v1alpha/ReportsService.kt` line 10 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1030#-NWmoP6N1dC_o_V2OCh_:-NWmoP6N1dC_o_V2OCha:bjne5iy) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/428229c5be78ed36ffb28db0d8618af4ac4da61a/src/main/kotlin/org/wfanet/measurement/reporting/bff/service/api/v1alpha/ReportsService.kt#L10)):*
> ```Kotlin
> 
> class ReportsService(
>     private val reportsStub: ReportsGrpcKt.ReportsCoroutineStub
> ```

This is the wrong stub. This is one for the BFF API, not for the Reporting API.



_Code quote:_
```Kotlin
ReportsGrpcKt.ReportsCoroutineStub
```
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/bff/service/api/v1alpha/ReportsService.kt` line 18 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1030#-NWmoX9m8tAQK_i8YVar:-NWmoX9m8tAQK_i8YVas:boqt9a7) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/428229c5be78ed36ffb28db0d8618af4ac4da61a/src/main/kotlin/org/wfanet/measurement/reporting/bff/service/api/v1alpha/ReportsService.kt#L18)):*
> ```Kotlin
> 
>         // Map internal response to UI response
>         val results = ListReportsResponse.newBuilder()
> ```

New code should always use Kotlin DSL builders.

bdomen-ggl (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1030)* status: all files reviewed, 3 unresolved discussions (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/bff/deploy/V1AlphaPublicApiServer.kt` line 96 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1030#-NWmll3d7xIwIYpAIJkY:-NWnDRZQ2d4VKBg4d03-:b-82f8fl) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/428229c5be78ed36ffb28db0d8618af4ac4da61a/src/main/kotlin/org/wfanet/measurement/reporting/bff/deploy/V1AlphaPublicApiServer.kt#L96)):*

Hmm, just an observation that because the Reporting API expects MC end-user auth (i.e. a client cert bound to a specific MC) then if the BFF calls that API then it needs to hold the private keys to those certs. This makes me wonder if we want to effectively have an auth bypass for the BFF, e.g. running a different Reporting server binary with different auth interceptors that can only be accessed from the same cluster.

@kungfucraig WDYT?
</blockquote></details>

Yeah, that's an interesting idea. For the short term that works, but for the long term we could even have everything run through the same service but just make permissions more flexible (eg. you have access if you either have the certs or are in a whitelist through IAM).
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/bff/service/api/v1alpha/BUILD.bazel` line 23 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1030#-NWmlb1n0ZlVROCYKTif:-NXAylB541xY_N1xWZnZ:b-pueros) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/428229c5be78ed36ffb28db0d8618af4ac4da61a/src/main/kotlin/org/wfanet/measurement/reporting/bff/service/api/v1alpha/BUILD.bazel#L23)):*

I think you have the server targets in two places.
</blockquote></details>

Yep thanks, I moved some things around and didn't delete this.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/bff/service/api/v1alpha/ReportsService.kt` line 10 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1030#-NWmoP6N1dC_o_V2OCh_:-NWnE7Ao-ao7XRm9SdcE:bp118e3) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/428229c5be78ed36ffb28db0d8618af4ac4da61a/src/main/kotlin/org/wfanet/measurement/reporting/bff/service/api/v1alpha/ReportsService.kt#L10)):*

This is the wrong stub. This is one for the BFF API, not for the Reporting API.
</blockquote></details>

Ah thanks. I actually missed that one.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/bff/service/api/v1alpha/ReportsService.kt` line 18 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1030#-NWmoX9m8tAQK_i8YVar:-NWnEBxQ8OLrBwJXIw9X:bgbb9k2) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/428229c5be78ed36ffb28db0d8618af4ac4da61a/src/main/kotlin/org/wfanet/measurement/reporting/bff/service/api/v1alpha/ReportsService.kt#L18)):*

New code should always use Kotlin DSL builders.
</blockquote></details>

Yeah. I just had this as a placeholder for the draft. I'll update this.

I didn't finish doing all the mapping yet either. I'll most likely do the mapping along with the proto message updates in another PR when we know more what exactly we want to show in the UI.

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1030)* status: all files reviewed, 4 unresolved discussions (waiting on @kungfucraig)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/bff/deploy/V1AlphaPublicApiServer.kt` line 96 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1030#-NWmll3d7xIwIYpAIJkY:-NXBmz3S-4EDT1SleJub:b-yitmfw) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/428229c5be78ed36ffb28db0d8618af4ac4da61a/src/main/kotlin/org/wfanet/measurement/reporting/bff/deploy/V1AlphaPublicApiServer.kt#L96)):*

Yeah, that's an interesting idea. For the short term that works, but for the long term we could even have everything run through the same service but just make permissions more flexible (eg. you have access if you either have the certs or are in a whitelist through IAM).
</blockquote></details>

From offline discussion, the decision was to just have the UI bound to a single MC for now and not bother with any client auth to the BFF. What this means for this PR is that you don't need these flags and instead just need flags pointing to the MC TLS certificate and private key.

kungfucraig (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1030)* status: all files reviewed, 4 unresolved discussions (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/bff/deploy/V1AlphaPublicApiServer.kt` line 96 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1030#-NWmll3d7xIwIYpAIJkY:-NXCf7-i8Qh7Gt2JU4bJ:b-x0bm01) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/428229c5be78ed36ffb28db0d8618af4ac4da61a/src/main/kotlin/org/wfanet/measurement/reporting/bff/deploy/V1AlphaPublicApiServer.kt#L96)):*

From offline discussion, the decision was to just have the UI bound to a single MC for now and not bother with any client auth to the BFF. What this means for this PR is that you don't need these flags and instead just need flags pointing to the MC TLS certificate and private key.
</blockquote></details>

I'd be happy with just having the BFF stood up for a single MC for now. This is the simplest thing we can do both for UI usage and implementation. For UI usage, it means no one has to install a cert, just hit the UI. Likely we'll just have a single MC that is well provisioned with Event Groups anyway. For implementation you don't have to deal with multiplexing the MC.

Of course eventually we'll need to deal with the authn/authz problem, but this seems like the simplest way to punt it for now.

PR #1061, opened 2023-06-12
Implement internal reports service read methods for reporting v2

tristanvuong2021:


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1061)

riemanli (commented):
> Reviewed 3 of 10 files at r1, 5 of 7 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1061)* status: 8 of 10 files reviewed, 2 unresolved discussions (waiting on @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/postgres/writers/CreateMetrics.kt` line 267 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1061#-NYRKfy846GfioLdt9PS:-NYRKfy97MjInwv6Da37:bpi48td) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/fbf8289368cd1e9a207841ab9e20003cb214a371/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/postgres/writers/CreateMetrics.kt#L267)):*
> ```Kotlin
>       ) {
>         metricCalculationSpecReportingMetricsBinders.forEach { addBinding(it) }
>       }
> ```

What will happen if there is no corresponding row in `MetricCalculationSpecReportingMetrics` table?



_Code quote:_
```Kotlin
    val metricCalculationSpecReportingMetricsStatement =
      boundStatement(
        """
        UPDATE MetricCalculationSpecReportingMetrics SET MetricId = $1
        WHERE MeasurementConsumerId = $2 AND CreateMetricRequestId = $3
        """
      ) {
        metricCalculationSpecReportingMetricsBinders.forEach { addBinding(it) }
      }
```
___
*[`src/main/proto/wfa/measurement/reporting/v2alpha/page_token.proto` line 39 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1061#-NYRLaUx35pqX_P-rX4o:-NYRLaUx35pqX_P-rX4p:btdsk8t) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/fbf8289368cd1e9a207841ab9e20003cb214a371/src/main/proto/wfa/measurement/reporting/v2alpha/page_token.proto#L39)):*
> ```proto
>   message PreviousPageEnd {
>     string cmms_measurement_consumer_id = 1;
>     google.protobuf.Timestamp create_time = 2;
> ```

Why do we need `create_time` in the filter?

tristanvuong2021 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1061)* status: 8 of 10 files reviewed, 2 unresolved discussions (waiting on @riemanli)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/postgres/writers/CreateMetrics.kt` line 267 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1061#-NYRKfy846GfioLdt9PS:-NYUexmQ8zR9tDYeSaBa:b6mwfr4) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/980c2d5a5666bda8cd29105c0a16cc208b970b64/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/postgres/writers/CreateMetrics.kt#L267)):*

What will happen if there is no corresponding row in `MetricCalculationSpecReportingMetrics` table?
</blockquote></details>

I have a check here: https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/980c2d5a5666bda8cd29105c0a16cc208b970b64/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/postgres/writers/CreateMetrics.kt#L355
___
*[`src/main/proto/wfa/measurement/reporting/v2alpha/page_token.proto` line 39 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1061#-NYRLaUx35pqX_P-rX4o:-NYUfFhpEMbQm7BRkXAS:b-qlxvnf) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/980c2d5a5666bda8cd29105c0a16cc208b970b64/src/main/proto/wfa/measurement/reporting/v2alpha/page_token.proto#L39)):*

Why do we need `create_time` in the filter?
</blockquote></details>

I forgot to add create_time filter to the public list reports too and not just the page token, but the idea is that the user can get the latest reports.

tristanvuong2021 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1061)* status: 8 of 10 files reviewed, 2 unresolved discussions (waiting on @riemanli)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/postgres/writers/CreateMetrics.kt` line 267 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1061#-NYRKfy846GfioLdt9PS:-NYUfZefECi7h4Y_gf3f:bkrdc71) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/980c2d5a5666bda8cd29105c0a16cc208b970b64/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/postgres/writers/CreateMetrics.kt#L267)):*

I have a check here: https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/980c2d5a5666bda8cd29105c0a16cc208b970b64/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/postgres/writers/CreateMetrics.kt#L355
</blockquote></details>

And if 0 rows are updated, nothing happens

tristanvuong2021 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1061)* status: 8 of 10 files reviewed, 2 unresolved discussions (waiting on @riemanli)

___
*[`src/main/proto/wfa/measurement/reporting/v2alpha/page_token.proto` line 39 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1061#-NYRLaUx35pqX_P-rX4o:-NYUgHLJ9nX59_LQA8Kq:btznxod) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/980c2d5a5666bda8cd29105c0a16cc208b970b64/src/main/proto/wfa/measurement/reporting/v2alpha/page_token.proto#L39)):*

I forgot to add create_time filter to the public list reports too and not just the page token, but the idea is that the user can get the latest reports.
</blockquote></details>

Alternative, by create_time is the default and just need to document it.

riemanli (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1061)* status: 8 of 10 files reviewed, 1 unresolved discussion (waiting on @tristanvuong2021)

___
*[`src/main/proto/wfa/measurement/reporting/v2alpha/page_token.proto` line 39 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1061#-NYRLaUx35pqX_P-rX4o:-NYUrEeO25oAQ4rfxLOW:b-fpe3ho) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/980c2d5a5666bda8cd29105c0a16cc208b970b64/src/main/proto/wfa/measurement/reporting/v2alpha/page_token.proto#L39)):*

Alternative, by create_time is the default and just need to document it.
</blockquote></details>

Got it thanks. Could you also make the corresponding changes for the public server?

riemanli (changes requested):
> Reviewed 2 of 7 files at r2, 3 of 5 files at r3, 2 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1061)* status: all files reviewed, 2 unresolved discussions (waiting on @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/internal/testing/v2/ReportsServiceTest.kt` line 877 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1061#-NYV_8kb-bMUkwLk7Xrf:-NYV_8kb-bMUkwLk7Xrg:b8d2f0e) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/b302b38e8b16ef8fc53068058d510954b3150432/src/main/kotlin/org/wfanet/measurement/reporting/service/internal/testing/v2/ReportsServiceTest.kt#L877)):*
> ```Kotlin
> 
>   @Test
>   fun `streamReports returns reports with id after id filter`(): Unit = runBlocking {
> ```

Could you add one more `streamReports` test that is with both an id after filter and a create time after filter?
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/internal/testing/v2/ReportsServiceTest.kt` line 1118 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1061#-NYVZeB18nfCWa5BBd6b:-NYVZeB18nfCWa5BBd6c:bakoigv) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/b302b38e8b16ef8fc53068058d510954b3150432/src/main/kotlin/org/wfanet/measurement/reporting/service/internal/testing/v2/ReportsServiceTest.kt#L1118)):*
> ```Kotlin
>         }
> 
>       val reportingMetricCalculationSpec2 =
> ```

Is there any particular reason to have two `reportingMetricCalculationSpec`s?

tristanvuong2021 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1061)* status: 10 of 11 files reviewed, 2 unresolved discussions (waiting on @riemanli and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/internal/testing/v2/ReportsServiceTest.kt` line 877 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1061#-NYV_8kb-bMUkwLk7Xrf:-NYZIiOM8tz8wjkzGY0j:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/b302b38e8b16ef8fc53068058d510954b3150432/src/main/kotlin/org/wfanet/measurement/reporting/service/internal/testing/v2/ReportsServiceTest.kt#L877)):*

Could you add one more `streamReports` test that is with both an id after filter and a create time after filter?
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/internal/testing/v2/ReportsServiceTest.kt` line 1118 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1061#-NYVZeB18nfCWa5BBd6b:-NYZIicX1qIXEN2OLTGs:b-jfhmi4) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/b302b38e8b16ef8fc53068058d510954b3150432/src/main/kotlin/org/wfanet/measurement/reporting/service/internal/testing/v2/ReportsServiceTest.kt#L1118)):*

Is there any particular reason to have two `reportingMetricCalculationSpec`s?
</blockquote></details>

I changed it to just one and added an explicit test for it instead.

riemanli (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1061)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @SanjayVas)

SanjayVas (changes requested):
> Reviewed 1 of 10 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1061)* status: all files reviewed, 2 unresolved discussions (waiting on @Marco-Premier and @tristanvuong2021)

___
*[`src/main/proto/wfa/measurement/reporting/v2alpha/page_token.proto` line 24 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1061#-NY_Kmk85io1oML2dfoe:-NY_Kmk85io1oML2dfof:b-c5rmti) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ec4ad4407fea53fa1bf87350e62cf11c5888d9e9/src/main/proto/wfa/measurement/reporting/v2alpha/page_token.proto#L24)):*
Can be handled in a different PR, but these page token messages aren't really part of the public API. They may still be bound to a given API version, but having them here makes them part of the public API surface.

From [AIP-158](https://google.aip.dev/158#opacity):

>Page tokens provided by APIs **must** be opaque (but URL-safe) strings, and
**must not** be user-parseable. This is because if users are able to
deconstruct these, _they will do so_. This effectively makes the implementation
details of your API's pagination become part of the API surface, and it becomes
impossible to update those details without breaking users.
>
> **Warning:** Base-64 encoding an otherwise-transparent page token is **not** a
sufficient obfuscation mechanism.
> 
> For page tokens which do not need to be stored in a database, and which do not
contain sensitive data, an API **may** obfuscate the page token by defining an
internal protocol buffer message with any data needed, and send the serialized
proto, base-64 encoded.

Of course the project being open source means that users can find the internal message definition if they are determined, but at least it should be clearer that doing so is at their own risk (and we won't provide any libraries that include it).
___
*[`src/main/proto/wfa/measurement/reporting/v2alpha/page_token.proto` line 37 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1061#-NY_6h2d8nulIEgwdX0c:-NY_6h2d8nulIEgwdX0d:bgyrk12) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ec4ad4407fea53fa1bf87350e62cf11c5888d9e9/src/main/proto/wfa/measurement/reporting/v2alpha/page_token.proto#L37)):*
> ```proto
>   int32 page_size = 1;
>   string cmms_measurement_consumer_id = 2;
>   message PreviousPageEnd {
> ```

For the way we handle page tokens (not requiring any DB persistence), the order of fields in here needs to match how the List responses are ordered (where later fields become tie-breakers for earlier ones). There also needs to be enough information to guarantee a total ordering. So in this case it should be (`create_time`, `cmms_measurement_consumer_id`, `external_report_id`), which should also be what's in the ORDER BY clause of the SQL query.

Marco-Premier (approved):
> Reviewed 3 of 7 files at r2, 2 of 5 files at r3, 2 of 2 files at r4, 1 of 1 files at r6, 4 of 4 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1061)* status: all files reviewed, 2 unresolved discussions (waiting on @tristanvuong2021)

tristanvuong2021 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1061)* status: all files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

___
*[`src/main/proto/wfa/measurement/reporting/v2alpha/page_token.proto` line 37 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1061#-NY_6h2d8nulIEgwdX0c:-NYcp2dH0hz5Y2ZiYq2x:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ec4ad4407fea53fa1bf87350e62cf11c5888d9e9/src/main/proto/wfa/measurement/reporting/v2alpha/page_token.proto#L37)):*

For the way we handle page tokens (not requiring any DB persistence), the order of fields in here needs to match how the List responses are ordered (where later fields become tie-breakers for earlier ones). There also needs to be enough information to guarantee a total ordering. So in this case it should be (`create_time`, `cmms_measurement_consumer_id`, `external_report_id`), which should also be what's in the ORDER BY clause of the SQL query.
</blockquote></details>

Done.

SanjayVas (changes requested):
> Reviewed 1 of 7 files at r2, 1 of 4 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1061)* status: all files reviewed, 4 unresolved discussions (waiting on @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/postgres/readers/ReportReader.kt` line 59 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1061#-NYe1bb37sJB-9YcNnfX:-NYe1bb37sJB-9YcNnfY:bbu0i2f) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/b0839aed945b82cc6be7557091fe0fa41fa7d913/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/postgres/readers/ReportReader.kt#L59)):*
> ```Kotlin
>     val createTime: Timestamp,
>     val timeIntervals: MutableSet<TimeInterval>,
>     // Key is externalReportingSetId.
> ```

Properties use KDoc syntax and follow all the normal documentation rules (https://developer.android.com/kotlin/style-guide#documentation)



_Code quote:_
```Kotlin
// Key is externalReportingSetId.
```
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/postgres/readers/ReportReader.kt` line 113 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1061#-NYe1yc40AOvwUWHTUR_:-NYe1yc40AOvwUWHTURa:b-byuocg) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/b0839aed945b82cc6be7557091fe0fa41fa7d913/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/postgres/readers/ReportReader.kt#L113)):*
> ```Kotlin
>         FROM MeasurementConsumers
>           JOIN Reports USING(MeasurementConsumerId)
>         """ +
> ```

Every multiline string using this syntax that has leading whitespace should have a `trimIndent()`.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/postgres/readers/ReportReader.kt` line 205 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1061#-NYe3T08-AmzQshpsgti:-NYe3T08-AmzQshpsgtj:b-86g0c7) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/b0839aed945b82cc6be7557091fe0fa41fa7d913/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/postgres/readers/ReportReader.kt#L205)):*
> ```Kotlin
>   private fun createResultFlow(statement: BoundStatement): Flow<Result> {
>     return flow {
>       val reportInfoCollection = buildResultMap(statement)
> ```

Isn't this basically reading in the entire result into memory first? That defeats one of the big benefits of using flows.

I think you want to internally call something different depending on whether it's a batch get vs. a list, as the two have different ordering requirements.

tristanvuong2021 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1061)* status: 11 of 12 files reviewed, 4 unresolved discussions (waiting on @Marco-Premier and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/postgres/readers/ReportReader.kt` line 59 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1061#-NYe1bb37sJB-9YcNnfX:-NYe9Mnu2xkkmy_TycrA:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/b0839aed945b82cc6be7557091fe0fa41fa7d913/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/postgres/readers/ReportReader.kt#L59)):*

Properties use KDoc syntax and follow all the normal documentation rules (https://developer.android.com/kotlin/style-guide#documentation)
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/postgres/readers/ReportReader.kt` line 113 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1061#-NYe1yc40AOvwUWHTUR_:-NYe9NJK3aKTmQsuPLol:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/b0839aed945b82cc6be7557091fe0fa41fa7d913/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/postgres/readers/ReportReader.kt#L113)):*

Every multiline string using this syntax that has leading whitespace should have a `trimIndent()`.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/postgres/readers/ReportReader.kt` line 205 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1061#-NYe3T08-AmzQshpsgti:-NYe9NsE5vFq4dxVP1IS:b-q327qw) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/b0839aed945b82cc6be7557091fe0fa41fa7d913/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/postgres/readers/ReportReader.kt#L205)):*

Isn't this basically reading in the entire result into memory first? That defeats one of the big benefits of using flows.

I think you want to internally call something different depending on whether it's a batch get vs. a list, as the two have different ordering requirements.
</blockquote></details>

It does read the whole result into memory first, but it allows me to use joins to reduce the number of queries.

I do have a way to handle ordering depending on if it is a batch get or a list, but in this case there is only a list.

tristanvuong2021 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1061)* status: 11 of 12 files reviewed, 4 unresolved discussions (waiting on @Marco-Premier and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/postgres/readers/ReportReader.kt` line 205 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1061#-NYe3T08-AmzQshpsgti:-NYeFLrT6ndIKb9iYzzv:baw53hh) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/b0839aed945b82cc6be7557091fe0fa41fa7d913/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/postgres/readers/ReportReader.kt#L205)):*

It does read the whole result into memory first, but it allows me to use joins to reduce the number of queries.

I do have a way to handle ordering depending on if it is a batch get or a list, but in this case there is only a list.
</blockquote></details>

Hmm, I think I have a way to do it without having it all in memory at once.

tristanvuong2021 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1061)* status: 11 of 12 files reviewed, 4 unresolved discussions (waiting on @Marco-Premier and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/postgres/readers/ReportReader.kt` line 205 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1061#-NYe3T08-AmzQshpsgti:-NYsEnhr0KDqO7znfs9V:bb6un1j) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/b0839aed945b82cc6be7557091fe0fa41fa7d913/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/postgres/readers/ReportReader.kt#L205)):*

Hmm, I think I have a way to do it without having it all in memory at once.
</blockquote></details>

Done

SanjayVas (changes requested):
> Reviewed 1 of 7 files at r2, 1 of 5 files at r3, 1 of 4 files at r7, 1 of 1 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1061)* status: all files reviewed, 6 unresolved discussions (waiting on @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/postgres/readers/ReportReader.kt` line 51 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1061#-NYsPBnADnQ0IzQPNEuJ:-NYsPBnADnQ0IzQPNEuK:b-vg7ow) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1ed89fab10ed93e88c52cb67e7e2a0aac34632d9/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/postgres/readers/ReportReader.kt#L51)):*
> ```Kotlin
>   )
> 
>   private data class ResultWrapper(val result: Result?)
> ```

It looks like this is used just because `consume` expects a non-null `T`. Maybe use an `Optional` so it's clearer, and document the reason with a TODO to allow nulls for `consume`. Allowing nulls would allow us to potentially use [`filterNotNull`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/filter-not-null.html).
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/postgres/readers/ReportReader.kt` line 61 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1061#-NYsOAKX2Y4bVw-Z_wnR:-NYsOAKX2Y4bVw-Z_wnS:b-9wcrk5) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1ed89fab10ed93e88c52cb67e7e2a0aac34632d9/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/postgres/readers/ReportReader.kt#L61)):*
> ```Kotlin
>     val createTime: Timestamp,
>     val timeIntervals: MutableSet<TimeInterval>,
>     /** Key is externalReportingSetId. */
> ```

First line of KDoc should be a summary fragment, meaning the comment should describe what it is before going into details. e.g. "Map of external reporting set ID to [ReportingMetricCalculationSpecInfo]"
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/postgres/readers/ReportReader.kt` line 316 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1061#-NYsZ0fF9YDBoxpffI82:-NYsZ0fF9YDBoxpffI83:b3ge2dc) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1ed89fab10ed93e88c52cb67e7e2a0aac34632d9/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/postgres/readers/ReportReader.kt#L316)):*
> ```Kotlin
> 
>     return flow {
>       readContext.executeQuery(statement).consume(translate).collect {
> ```

So it looks like what you're trying to do here is both emit some intermediate results and also accumulate some final result. There are flow operators for this kind of thing, e.g. [runningReduce](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/running-reduce.html) and [runningFold](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/running-fold.html).

My concern is mainly the fact that `translate` has a side effect of mutating `reportInfo`. We'd want anything like that to be done very narrowly in a clear way. In the case of the flow operators I mentioned, it does this by explicitly passing the mutable part as the `accumulator`.

I'm thinking something along the lines of using `runningFold` and `filter`, or at the very least separating out the mutable portion and calling it `accumulator` so readers may be able to recognize the `reduce` pattern.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/postgres/writers/CreateMetrics.kt` line 222 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1061#-NYsjOw654ul9ht31Q1D:-NYsjOw654ul9ht31Q1E:bahc74j) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1ed89fab10ed93e88c52cb67e7e2a0aac34632d9/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/postgres/writers/CreateMetrics.kt#L222)):*
> ```Kotlin
>                   bind("$3", createMetricRequestUuid)
>                 }
>               } catch (_: IllegalArgumentException) {}
> ```

This looks odd. Why are we catching and ignoring this? If it's really necessary, document why.



_Code quote:_
```Kotlin
catch (_: IllegalArgumentException) {}
```
___
*[`src/main/proto/wfa/measurement/reporting/v2alpha/reports_service.proto` line 85 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1061#-NYsjt1fBwFXmK2656xr:-NYsjt1fBwFXmK2656xs:b9pwso) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1ed89fab10ed93e88c52cb67e7e2a0aac34632d9/src/main/proto/wfa/measurement/reporting/v2alpha/reports_service.proto#L85)):*
> ```proto
> 
>   // Filter criteria. Repeated fields are treated as logical ORs, and multiple
>   // fields specified as logical ANDs.
> ```

Unlike the CMMS API, the Reporting API follows the filtering part of AIP-132, with the exception that we use CEL rather than the grammar from AIP-160. This means that the `filter` field must be a string.

tristanvuong2021 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1061)* status: all files reviewed, 6 unresolved discussions (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/postgres/readers/ReportReader.kt` line 316 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1061#-NYsZ0fF9YDBoxpffI82:-NYsv1vH7Xco3rYKIfz2:b-cum986) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1ed89fab10ed93e88c52cb67e7e2a0aac34632d9/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/postgres/readers/ReportReader.kt#L316)):*

So it looks like what you're trying to do here is both emit some intermediate results and also accumulate some final result. There are flow operators for this kind of thing, e.g. [runningReduce](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/running-reduce.html) and [runningFold](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/running-fold.html).

My concern is mainly the fact that `translate` has a side effect of mutating `reportInfo`. We'd want anything like that to be done very narrowly in a clear way. In the case of the flow operators I mentioned, it does this by explicitly passing the mutable part as the `accumulator`.

I'm thinking something along the lines of using `runningFold` and `filter`, or at the very least separating out the mutable portion and calling it `accumulator` so readers may be able to recognize the `reduce` pattern.
</blockquote></details>

It doesn't fit that pattern. It doesn't accumulate from start to finish based on an initial value. It accumulates then resets, accumulates then resets.

tristanvuong2021 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1061)* status: 6 of 12 files reviewed, 4 unresolved discussions (waiting on @Marco-Premier, @riemanli, and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/postgres/readers/ReportReader.kt` line 51 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1061#-NYsPBnADnQ0IzQPNEuJ:-NYt-HCvD7_elMofqO3V:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1ed89fab10ed93e88c52cb67e7e2a0aac34632d9/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/postgres/readers/ReportReader.kt#L51)):*

It looks like this is used just because `consume` expects a non-null `T`. Maybe use an `Optional` so it's clearer, and document the reason with a TODO to allow nulls for `consume`. Allowing nulls would allow us to potentially use [`filterNotNull`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/filter-not-null.html).
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/postgres/readers/ReportReader.kt` line 61 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1061#-NYsOAKX2Y4bVw-Z_wnR:-NYt-Hc95yagMwl3SMr0:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1ed89fab10ed93e88c52cb67e7e2a0aac34632d9/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/postgres/readers/ReportReader.kt#L61)):*

First line of KDoc should be a summary fragment, meaning the comment should describe what it is before going into details. e.g. "Map of external reporting set ID to [ReportingMetricCalculationSpecInfo]"
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/postgres/writers/CreateMetrics.kt` line 222 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1061#-NYsjOw654ul9ht31Q1D:-NYt-D-r1hRW0V09We5R:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1ed89fab10ed93e88c52cb67e7e2a0aac34632d9/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/postgres/writers/CreateMetrics.kt#L222)):*

This looks odd. Why are we catching and ignoring this? If it's really necessary, document why.
</blockquote></details>

Done.
___
*[`src/main/proto/wfa/measurement/reporting/v2alpha/reports_service.proto` line 85 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1061#-NYsjt1fBwFXmK2656xr:-NYt-Dho4Y80RBoUDxv-:b-8jmbz5) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1ed89fab10ed93e88c52cb67e7e2a0aac34632d9/src/main/proto/wfa/measurement/reporting/v2alpha/reports_service.proto#L85)):*

Unlike the CMMS API, the Reporting API follows the filtering part of AIP-132, with the exception that we use CEL rather than the grammar from AIP-160. This means that the `filter` field must be a string.
</blockquote></details>

Removed it. Will add filter later.

SanjayVas (changes requested):
> Reviewed 1 of 7 files at r2, 1 of 4 files at r7, 6 of 6 files at r9.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1061)* status: 11 of 12 files reviewed, 2 unresolved discussions (waiting on @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/postgres/writers/CreateMetrics.kt` line 224 at r9](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1061#-NYt2wIxAriwVxO3rcXo:-NYt2wIxAriwVxO3rcXp:b-q2ulzg) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c1e45000c3ba87a269ae48256694661da7381bf6/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/postgres/writers/CreateMetrics.kt#L224)):*
> ```Kotlin
>               } catch (_: IllegalArgumentException) {
>                 // Request IDs for metrics generated for a report are UUIDs. If the request ID isn't
>                 // a UUID, it is still a valid metric so the exception is caught instead of thrown.
> ```

If this is the reason, the code can be rewritten to make this clearer. We also generally want the try block around the narrowest part of the code that can throw this exception to avoid inadvertently catching an exception for another case (e.g. if this was thrown from `bind` instead, we'd be masking some other error).

```kotlin
if (it.requestId.isNotEmpty()) {
  val createMetricRequestUuid: UUID? = try {
    UUID.fromString(it.requestId)
  } catch (_: IllegalArgumentException) {
    // Non-Report Metrics may not use a UUID.
    null
  }

  if (createMetricRequestUuid != null) {
    // ...
  }
}
```



_Code quote:_
```Kotlin
                // Request IDs for metrics generated for a report are UUIDs. If the request ID isn't
                // a UUID, it is still a valid metric so the exception is caught instead of thrown.
```
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/ProtoConversions.kt` line 675 at r9](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1061#-NYt524o44CMQRrp0Nrn:-NYt524o44CMQRrp0Nro:bxy5ct5) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c1e45000c3ba87a269ae48256694661da7381bf6/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/ProtoConversions.kt#L675)):*
> ```Kotlin
>       StreamReportsRequestKt.filter {
>         cmmsMeasurementConsumerId = source.cmmsMeasurementConsumerId
>         after =
> ```

Shouldn't `after` only be set if `source.hasLastReport`? The stream implementation should be able to check `hasAfter`.

tristanvuong2021 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1061)* status: 8 of 12 files reviewed, 2 unresolved discussions (waiting on @Marco-Premier, @riemanli, and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/postgres/writers/CreateMetrics.kt` line 224 at r9](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1061#-NYt2wIxAriwVxO3rcXo:-NYtGfkE8aigH_vYIqjd:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c1e45000c3ba87a269ae48256694661da7381bf6/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/postgres/writers/CreateMetrics.kt#L224)):*

If this is the reason, the code can be rewritten to make this clearer. We also generally want the try block around the narrowest part of the code that can throw this exception to avoid inadvertently catching an exception for another case (e.g. if this was thrown from `bind` instead, we'd be masking some other error).

```kotlin
if (it.requestId.isNotEmpty()) {
  val createMetricRequestUuid: UUID? = try {
    UUID.fromString(it.requestId)
  } catch (_: IllegalArgumentException) {
    // Non-Report Metrics may not use a UUID.
    null
  }

  if (createMetricRequestUuid != null) {
    // ...
  }
}
```
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/ProtoConversions.kt` line 675 at r9](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1061#-NYt524o44CMQRrp0Nrn:-NYtGgTF790yug4tE7RH:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c1e45000c3ba87a269ae48256694661da7381bf6/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/ProtoConversions.kt#L675)):*

Shouldn't `after` only be set if `source.hasLastReport`? The stream implementation should be able to check `hasAfter`.
</blockquote></details>

Done.

SanjayVas (approved):
> Reviewed 1 of 1 files at r6, 1 of 4 files at r7, 1 of 1 files at r10, 3 of 3 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1061)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @tristanvuong2021)

riemanli (approved):
> Reviewed 3 of 4 files at r7, 3 of 6 files at r9, 1 of 1 files at r10, 3 of 3 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1061)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @tristanvuong2021)

PR #1070, opened 2023-06-19
Implement PostgresComputationStatsService

YuhongWang-Amazon:
This implements the PostgresComputationStatsService and partial PostgresComputationsService that is enough to run the PostgresComputationStatsServiceTest


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1070)

renjiezh (changes requested):
> Reviewed 9 of 13 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1070)* status: 9 of 13 files reviewed, 3 unresolved discussions (waiting on @YuhongWang-Amazon)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/postgres/PostgresComputationsService.kt` line 100 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1070#-NYPH4zF8m6UGwt_qvwP:-NYPH4zF8m6UGwt_qvwQ:bozzpmm) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3a2d7fe01d0b5a4855ba319361d8daa4e7969491/src/main/kotlin/org/wfanet/measurement/duchy/deploy/postgres/PostgresComputationsService.kt#L100)):*
> ```Kotlin
>         ComputationReader(protocolStageEnumHelper)
>           .readComputationToken(client.singleUse(), request.globalComputationId)
>           ?: failGrpc(Status.UNKNOWN) { "Created computation not found." }
> ```

`Status.NOT_FOUND` might be more accurate here.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/postgres/PostgresComputationsService.kt` line 179 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1070#-NYOWRN72uZoEBZTCdz_:-NYOWRN72uZoEBZTCdza:b-9704sz) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3a2d7fe01d0b5a4855ba319361d8daa4e7969491/src/main/kotlin/org/wfanet/measurement/duchy/deploy/postgres/PostgresComputationsService.kt#L179)):*
> ```Kotlin
>     attempt: Long = 0L
>   ): CreateComputationLogEntryRequest {
>     return CreateComputationLogEntryRequest.newBuilder()
> ```

Let's get rid of builders and use dsl.
___
*[`src/test/kotlin/org/wfanet/measurement/duchy/deploy/postgres/PostgresComputationStatsServiceTest.kt` line 74 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1070#-NYOV73y5DS4_sq_bMu3:-NYOV73y5DS4_sq_bMu4:b-gss1dv) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3a2d7fe01d0b5a4855ba319361d8daa4e7969491/src/test/kotlin/org/wfanet/measurement/duchy/deploy/postgres/PostgresComputationStatsServiceTest.kt#L74)):*
> ```Kotlin
>       client,
>       idGenerator,
>       ALSACE,
> ```

Can we denote the parameter name like `computationTypeEnumHelper = ComputationTypes,`
Besides, do we need a fixed duchy name here?

YuhongWang-Amazon (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1070)* status: 8 of 13 files reviewed, 3 unresolved discussions (waiting on @renjiezh and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/postgres/PostgresComputationsService.kt` line 100 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1070#-NYPH4zF8m6UGwt_qvwP:-NYUQNUE9nrBMPdXpqjX:b-7oxsna) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3a2d7fe01d0b5a4855ba319361d8daa4e7969491/src/main/kotlin/org/wfanet/measurement/duchy/deploy/postgres/PostgresComputationsService.kt#L100)):*

`Status.NOT_FOUND` might be more accurate here.
</blockquote></details>

I was thinking about the NOT_FOUND as well. But exception will be raised from the createComputation request and after a computation that was supposed to be successfully created in a previous command. Hence, it is a bit weird to raise a NOT_FOUND exception here. 
I decide to use `UNKOWN` here, because I don't know in which scenario this error could occur.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/postgres/PostgresComputationsService.kt` line 179 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1070#-NYOWRN72uZoEBZTCdz_:-NYUJxZ6DGDjV3SLyJwy:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3a2d7fe01d0b5a4855ba319361d8daa4e7969491/src/main/kotlin/org/wfanet/measurement/duchy/deploy/postgres/PostgresComputationsService.kt#L179)):*

Let's get rid of builders and use dsl.
</blockquote></details>

Done.
___
*[`src/test/kotlin/org/wfanet/measurement/duchy/deploy/postgres/PostgresComputationStatsServiceTest.kt` line 74 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1070#-NYOV73y5DS4_sq_bMu3:-NYURKLiAww8YnQG0hit:b-lh0lp5) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3a2d7fe01d0b5a4855ba319361d8daa4e7969491/src/test/kotlin/org/wfanet/measurement/duchy/deploy/postgres/PostgresComputationStatsServiceTest.kt#L74)):*

Can we denote the parameter name like `computationTypeEnumHelper = ComputationTypes,`
Besides, do we need a fixed duchy name here?
</blockquote></details>

Added the parameter names.

For duchy name, it is used when sending update to kingdom. But it was never verified in any of the integration tests. So either a fixed string or a randomly generated name could work. Do you have any preference?

SanjayVas (changes requested):
> Reviewed 4 of 13 files at r1, 1 of 4 files at r4, 1 of 4 files at r5.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1070)* status: 9 of 16 files reviewed, 12 unresolved discussions (waiting on @renjiezh and @YuhongWang-Amazon)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/postgres/PostgresComputationStatsService.kt` line 36 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1070#-NYUgS8CCyww83WO0J2h:-NYUgS8D84iRBMT9YyU2:b5myqrb) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ebf9d08143c07c8aad15df56516558ca9d0d1c03/src/main/kotlin/org/wfanet/measurement/duchy/deploy/postgres/PostgresComputationStatsService.kt#L36)):*
> ```Kotlin
>     val localComputationId = request.localComputationId
>     val metricName = request.metricName
>     grpcRequire(localComputationId != 0L) { "Missing computation ID" }
> ```

nit: Reference proto field names in these error messages



_Suggestion:_
```Kotlin
local_computation_id
```
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/postgres/PostgresComputationsService.kt` line 100 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1070#-NYPH4zF8m6UGwt_qvwP:-NYUiFUA0y_aM-v8F102:bidvn2j) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3a2d7fe01d0b5a4855ba319361d8daa4e7969491/src/main/kotlin/org/wfanet/measurement/duchy/deploy/postgres/PostgresComputationsService.kt#L100)):*

I was thinking about the NOT_FOUND as well. But exception will be raised from the createComputation request and after a computation that was supposed to be successfully created in a previous command. Hence, it is a bit weird to raise a NOT_FOUND exception here. 
I decide to use `UNKOWN` here, because I don't know in which scenario this error could occur.
</blockquote></details>

`INTERNAL` might be a better fit here, as this is indicating a violation of a system invariant

Ideally this would all be in one transaction so it couldn't possibly happen.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/postgres/PostgresComputationsService.kt` line 83 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1070#-NYUgt1o8ryHKW7lAfir:-NYUgt1o8ryHKW7lAfis:bxpymp1) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ebf9d08143c07c8aad15df56516558ca9d0d1c03/src/main/kotlin/org/wfanet/measurement/duchy/deploy/postgres/PostgresComputationsService.kt#L83)):*
> ```Kotlin
>   ): CreateComputationResponse {
>     grpcRequire(request.globalComputationId.isNotEmpty()) {
>       "globalComputationId is not specified."
> ```

nit: protobuf field name rather than language-dependent version



_Suggestion:_
```Kotlin
global_computation_id 
```
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/postgres/PostgresComputationsService.kt` line 86 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1070#-NYUh63OBpBtnR5Tz7-N:-NYUh63OBpBtnR5Tz7-O:bkpg39f) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ebf9d08143c07c8aad15df56516558ca9d0d1c03/src/main/kotlin/org/wfanet/measurement/duchy/deploy/postgres/PostgresComputationsService.kt#L86)):*
> ```Kotlin
>     }
> 
>     return try {
> ```

nit: keep the `try` around the narrowest block that might throw that exception

e.g.

```kotlin
val result: ComputationReader.ReadComputationTokenResult = try {
  // ... create computation ...

  ComputationReader(protocolStageEnumHelper)
    .readComputationToken(client.singleUse(), request.globalComputationId)
    ?: failGrpc(Status.UNKNOWN) { "Created computation not found." }
```
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/postgres/PostgresComputationsService.kt` line 140 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1070#-NYUks9560EUecsFVfBL:-NYUks9560EUecsFVfBM:b-j53xkk) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ebf9d08143c07c8aad15df56516558ca9d0d1c03/src/main/kotlin/org/wfanet/measurement/duchy/deploy/postgres/PostgresComputationsService.kt#L140)):*
> ```Kotlin
>           else -> throw ex
>         }
>       }
> ```

It's more idiomatic to use catch blocks for the specific exception types, even if there's a tiny amount of repetition.

That said, don't we expect the stage attempt details to exist in this case? Maybe that condition should be an INTERNAL, or not even have a corresponding DuchyInternalException that we catch (e.g. just IllegalStateException).



_Suggestion:_
```Kotlin
      } catch (e: ComputationNotFoundException) {
        throw e.asStatusRuntimeException(Status.Code.NOT_FOUND)
      } catch (e: ComputationDetailsNotFoundException) {
        throw e.asStatusRuntimeException(Status.Code.INTERNAL)
      }
```
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/postgres/readers/ComputationReader.kt` line 55 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1070#-NYUk5HM-yfVcnyDjG8W:-NYUk5HM-yfVcnyDjG8X:b-nozhbj) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ebf9d08143c07c8aad15df56516558ca9d0d1c03/src/main/kotlin/org/wfanet/measurement/duchy/deploy/postgres/readers/ComputationReader.kt#L55)):*
> ```Kotlin
> 
>   /** A wrapper data class for ComputationToken query result */
>   data class ReadComputationTokenResult(val computationToken: ComputationToken)
> ```

nit: we'd primarily been using Result wrappers when we've needed additional fields, e.g. returning an internal ID.

If you want to keep the wrapper, maybe rename it since it should always be referenced in the context of its enclosing type.



_Suggestion:_
```Kotlin
Result
```
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/postgres/readers/ComputationReader.kt` line 57 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1070#-NYUiyvW3aBVKEaW3wN9:-NYUiyvW3aBVKEaW3wNA:b5l1ehf) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ebf9d08143c07c8aad15df56516558ca9d0d1c03/src/main/kotlin/org/wfanet/measurement/duchy/deploy/postgres/readers/ComputationReader.kt#L57)):*
> ```Kotlin
>   data class ReadComputationTokenResult(val computationToken: ComputationToken)
> 
>   private fun buildComputationToken(row: ResultRow) =
> ```

Expression functions are only really cases for when it's either a one-liner or a block builder (e.g. `runBlocking`, `coroutineScope`, `flow`). Convert this to a block body instead. This will also allow you to pull the intermediate local variables out of the DSL builder scope.



_Code quote:_
```Kotlin
 =
```
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/postgres/readers/ComputationReader.kt` line 157 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1070#-NYUoDNu-tzCKzLT53C2:-NYUoDNu-tzCKzLT53C3:b-mbyp2f) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ebf9d08143c07c8aad15df56516558ca9d0d1c03/src/main/kotlin/org/wfanet/measurement/duchy/deploy/postgres/readers/ComputationReader.kt#L157)):*
> ```Kotlin
>         cs.Details AS StageDetails,
>         (
>           SELECT ARRAY_AGG(
> ```

Note that we still don't know whether this is a good way to do this. For background, this is trying to emulate the Spanner `SELECT AS STRUCT` subquery which has additional support from the Spanner Java client library to easily parse arrays of structs (it uses the same data structure for a top-level row as for a struct subquery element). See the [issue](https://github.com/pgjdbc/r2dbc-postgresql/issues/519) I filed with r2dbc-postgresql for additional context.

I believe the more "standard" way for other SQL dialects is to either do multiple queries in the same transaction or do JOINs (depending on whether the priority is to reduce redundant data in the response vs. reduce number of queries).

It's not strictly necessary to do it differently now, just something I wanted to point out as a potential issue (e.g. if using JSON arrays like this leads to some other problems such as parsing performance or potentially hitting a result row size limit).
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/postgres/readers/ComputationReader.kt` line 256 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1070#-NYUkIEU9NIV3ueJqk0W:-NYUkIEU9NIV3ueJqk0X:btlnx47) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ebf9d08143c07c8aad15df56516558ca9d0d1c03/src/main/kotlin/org/wfanet/measurement/duchy/deploy/postgres/readers/ComputationReader.kt#L256)):*
> ```Kotlin
>   }
> 
>   /**
> ```

Drop this since you've now got the function from common-jvm.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/postgres/writers/ClaimWork.kt` line 63 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1070#-NYUrdXYAqVkuy3CXRP5:-NYUrdXZ-kauriIVs9yj:b-p12660) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ebf9d08143c07c8aad15df56516558ca9d0d1c03/src/main/kotlin/org/wfanet/measurement/duchy/deploy/postgres/writers/ClaimWork.kt#L63)):*
> ```Kotlin
>   ComputationProtocolStageDetailsHelper<
>     ProtocolT, StageT, StageDT, ComputationDT
>   > by computationProtocolStageDetailsHelper {
> ```

I don't think you need to copy this from the Spanner impl. AFAICT this is a hack to get all the "helper" functions into the class scope, which is entirely unnecessary.

We had previously discussed how there are a lot of IMO bad abstractions in the DB layer. I personally would consider this a great opportunity to implement new abstractions and not use anything in the`org.wfanet.measurement.duchy.db.computation` package, but I understand if you don't want to tackle that.



_Code quote:_
```Kotlin
  ComputationTypeEnumHelper<ProtocolT> by computationTypeEnumHelper,
  ComputationProtocolStagesEnumHelper<ProtocolT, StageT> by computationProtocolStagesEnumHelper,
  ComputationProtocolStageDetailsHelper<
    ProtocolT, StageT, StageDT, ComputationDT
  > by computationProtocolStageDetailsHelper {
```

YuhongWang-Amazon (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1070)* status: 8 of 17 files reviewed, 9 unresolved discussions (waiting on @renjiezh and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/postgres/PostgresComputationsService.kt` line 100 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1070#-NYPH4zF8m6UGwt_qvwP:-NYYkZkm7o9ChVA0Y5yd:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3a2d7fe01d0b5a4855ba319361d8daa4e7969491/src/main/kotlin/org/wfanet/measurement/duchy/deploy/postgres/PostgresComputationsService.kt#L100)):*

`INTERNAL` might be a better fit here, as this is indicating a violation of a system invariant

Ideally this would all be in one transaction so it couldn't possibly happen.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/postgres/PostgresComputationsService.kt` line 140 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1070#-NYUks9560EUecsFVfBL:-NYZ6fLK8nWb1wkb3YA2:b-mr7h7j) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ebf9d08143c07c8aad15df56516558ca9d0d1c03/src/main/kotlin/org/wfanet/measurement/duchy/deploy/postgres/PostgresComputationsService.kt#L140)):*

It's more idiomatic to use catch blocks for the specific exception types, even if there's a tiny amount of repetition.

That said, don't we expect the stage attempt details to exist in this case? Maybe that condition should be an INTERNAL, or not even have a corresponding DuchyInternalException that we catch (e.g. just IllegalStateException).
</blockquote></details>

IllegalStatException is probably better. Since this error is for the scenario when someone is holding the computation lock but there is NO computation stage details associated with it.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/postgres/readers/ComputationReader.kt` line 55 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1070#-NYUk5HM-yfVcnyDjG8W:-NYZ906Z2n6WIYpSCVdi:b-dzl44v) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ebf9d08143c07c8aad15df56516558ca9d0d1c03/src/main/kotlin/org/wfanet/measurement/duchy/deploy/postgres/readers/ComputationReader.kt#L55)):*

nit: we'd primarily been using Result wrappers when we've needed additional fields, e.g. returning an internal ID.

If you want to keep the wrapper, maybe rename it since it should always be referenced in the context of its enclosing type.
</blockquote></details>

Removed the wrapper. 
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/postgres/readers/ComputationReader.kt` line 57 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1070#-NYUiyvW3aBVKEaW3wN9:-NYZA2Zl3M60PG61ypLr:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ebf9d08143c07c8aad15df56516558ca9d0d1c03/src/main/kotlin/org/wfanet/measurement/duchy/deploy/postgres/readers/ComputationReader.kt#L57)):*

Expression functions are only really cases for when it's either a one-liner or a block builder (e.g. `runBlocking`, `coroutineScope`, `flow`). Convert this to a block body instead. This will also allow you to pull the intermediate local variables out of the DSL builder scope.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/postgres/readers/ComputationReader.kt` line 157 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1070#-NYUoDNu-tzCKzLT53C2:-NYZa_324a4pjPXOqa00:b-hclvu2) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ebf9d08143c07c8aad15df56516558ca9d0d1c03/src/main/kotlin/org/wfanet/measurement/duchy/deploy/postgres/readers/ComputationReader.kt#L157)):*

Note that we still don't know whether this is a good way to do this. For background, this is trying to emulate the Spanner `SELECT AS STRUCT` subquery which has additional support from the Spanner Java client library to easily parse arrays of structs (it uses the same data structure for a top-level row as for a struct subquery element). See the [issue](https://github.com/pgjdbc/r2dbc-postgresql/issues/519) I filed with r2dbc-postgresql for additional context.

I believe the more "standard" way for other SQL dialects is to either do multiple queries in the same transaction or do JOINs (depending on whether the priority is to reduce redundant data in the response vs. reduce number of queries).

It's not strictly necessary to do it differently now, just something I wanted to point out as a potential issue (e.g. if using JSON arrays like this leads to some other problems such as parsing performance or potentially hitting a result row size limit).
</blockquote></details>

Thanks for the context.
I think this JSON approach is not very elegant; especially that extra step of encoding and decoding of the byte array data.
I updated this reader into multiple queries. Please help take another look.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/postgres/readers/ComputationReader.kt` line 256 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1070#-NYUkIEU9NIV3ueJqk0W:-NYZBAfI4DKZN4FTeq74:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ebf9d08143c07c8aad15df56516558ca9d0d1c03/src/main/kotlin/org/wfanet/measurement/duchy/deploy/postgres/readers/ComputationReader.kt#L256)):*

Drop this since you've now got the function from common-jvm.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/postgres/writers/ClaimWork.kt` line 63 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1070#-NYUrdXYAqVkuy3CXRP5:-NYZDOFOA3Wz7sWFQAgm:b7r4z0p) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ebf9d08143c07c8aad15df56516558ca9d0d1c03/src/main/kotlin/org/wfanet/measurement/duchy/deploy/postgres/writers/ClaimWork.kt#L63)):*

I don't think you need to copy this from the Spanner impl. AFAICT this is a hack to get all the "helper" functions into the class scope, which is entirely unnecessary.

We had previously discussed how there are a lot of IMO bad abstractions in the DB layer. I personally would consider this a great opportunity to implement new abstractions and not use anything in the`org.wfanet.measurement.duchy.db.computation` package, but I understand if you don't want to tackle that.
</blockquote></details>

I am interested into creating a better abstractions for the computation data. I looked into it a bit, but found it quite complicated to rebuild all those. Hence I think it would be better to create a design doc before the implementation.

Since I planned to finish the implementation of all Duchy Postgres services by the end of this month. I prefer to defer that work to after all duchy postgres service implementations.

SanjayVas (changes requested):
> Reviewed 4 of 13 files at r1, 1 of 4 files at r4, 1 of 4 files at r5, 1 of 1 files at r6, 6 of 6 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1070)* status: all files reviewed, 6 unresolved discussions (waiting on @renjiezh and @YuhongWang-Amazon)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/postgres/PostgresComputationsService.kt` line 144 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1070#-NYdcYvI2nox8VBjUtX9:-NYdcYvI2nox8VBjUtXA:b-t0ozs5) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5d4e42b71ad3800fc85c07161e9c3eb605acbef0/src/main/kotlin/org/wfanet/measurement/duchy/deploy/postgres/PostgresComputationsService.kt#L144)):*
> ```Kotlin
>       )
>       token.toClaimWorkResponse()
>     } else ClaimWorkResponse.getDefaultInstance()
> ```

If you use braces on one part of an if block, use it on all parts. (Basically only omit braces when the whole `if` expression fits on one line.



_Code quote:_
```Kotlin
 ClaimWorkResponse.getDefaultInstance()
```
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/postgres/readers/ComputationReader.kt` line 106 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1070#-NYddPnv4sMzLBp13ju3:-NYddPnv4sMzLBp13ju4:b9tp16j) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5d4e42b71ad3800fc85c07161e9c3eb605acbef0/src/main/kotlin/org/wfanet/measurement/duchy/deploy/postgres/readers/ComputationReader.kt#L106)):*
> ```Kotlin
>   ): ComputationToken {
>     return computationToken {
>       this.globalComputationId = computation.globalComputationId
> ```

nit: I believe this is an unnecessary qualification since there's no other variable in scope with the same name



_Code quote:_
```Kotlin
this.
```
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/postgres/readers/ComputationReader.kt` line 249 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1070#-NYdfM8g6rFZdWXpxpLu:-NYdfM8g6rFZdWXpxpLv:bt4maic) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5d4e42b71ad3800fc85c07161e9c3eb605acbef0/src/main/kotlin/org/wfanet/measurement/duchy/deploy/postgres/readers/ComputationReader.kt#L249)):*
> ```Kotlin
>     val readContext = client.readTransaction()
>     try {
>       val computation = readComputation(readContext, globalComputationId) ?: return null
> ```

Specify types explicitly where not obvious (e.g. not calling a constructor or clear factory/builder).



_Code quote:_
```Kotlin
val computation
```

YuhongWang-Amazon (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1070)* status: 15 of 17 files reviewed, 5 unresolved discussions (waiting on @renjiezh and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/postgres/PostgresComputationsService.kt` line 144 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1070#-NYdcYvI2nox8VBjUtX9:-NYs8Ln2AR1tWfsmhULU:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5d4e42b71ad3800fc85c07161e9c3eb605acbef0/src/main/kotlin/org/wfanet/measurement/duchy/deploy/postgres/PostgresComputationsService.kt#L144)):*

If you use braces on one part of an if block, use it on all parts. (Basically only omit braces when the whole `if` expression fits on one line.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/postgres/readers/ComputationReader.kt` line 249 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1070#-NYdfM8g6rFZdWXpxpLu:-NYs91q89okS4s4-YkWc:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5d4e42b71ad3800fc85c07161e9c3eb605acbef0/src/main/kotlin/org/wfanet/measurement/duchy/deploy/postgres/readers/ComputationReader.kt#L249)):*

Specify types explicitly where not obvious (e.g. not calling a constructor or clear factory/builder).
</blockquote></details>

Done.

SanjayVas (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1070)* status: all files reviewed, 3 unresolved discussions (waiting on @renjiezh)

YuhongWang-Amazon (commented):
> Reviewed 1 of 13 files at r1.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1070)* status: all files reviewed, 3 unresolved discussions (waiting on @renjiezh)

___
*[`src/test/kotlin/org/wfanet/measurement/duchy/deploy/postgres/PostgresComputationStatsServiceTest.kt` line 74 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1070#-NYOV73y5DS4_sq_bMu3:-NYstBCJ12ZMkBx4CQzl:b-9jt1ym) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3a2d7fe01d0b5a4855ba319361d8daa4e7969491/src/test/kotlin/org/wfanet/measurement/duchy/deploy/postgres/PostgresComputationStatsServiceTest.kt#L74)):*

Added the parameter names.

For duchy name, it is used when sending update to kingdom. But it was never verified in any of the integration tests. So either a fixed string or a randomly generated name could work. Do you have any preference?



</blockquote></details>

@renjiezh do you have any other thoughts on this? If not, could you help resolve the feedbacks? Appreciated!

renjiezh (approved):
:lgtm:

> Reviewed 1 of 4 files at r4, 1 of 4 files at r5, 1 of 1 files at r6, 4 of 6 files at r7, 2 of 2 files at r8, 1 of 1 files at r9.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1070#-:-NYyCkAEBTdscT29jlic:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @YuhongWang-Amazon)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/postgres/PostgresComputationsService.kt` line 134 at r9](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1070#-NYyAzy81C9fKhDlg2ul:-NYyAzy81C9fKhDlg2um:bnazxmb) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a946bf2b90bb43955f9a9ab359dbd253401873c9/src/main/kotlin/org/wfanet/measurement/duchy/deploy/postgres/PostgresComputationsService.kt#L134)):*
> ```Kotlin
>       val token =
>         ComputationReader(protocolStageEnumHelper).readComputationToken(client, claimed)
>           ?: failGrpc(Status.UNKNOWN) { "Claimed computation $claimed not found." }
> ```

nit: As we discussed in another thread, "INTERNAL" or "NOT_FOUND" might be better.
___
*[`src/test/kotlin/org/wfanet/measurement/duchy/deploy/postgres/PostgresComputationStatsServiceTest.kt` line 74 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1070#-NYOV73y5DS4_sq_bMu3:-NYy9wFu5JNU2qdIGm2D:brxp9sq) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3a2d7fe01d0b5a4855ba319361d8daa4e7969491/src/test/kotlin/org/wfanet/measurement/duchy/deploy/postgres/PostgresComputationStatsServiceTest.kt#L74)):*

@renjiezh do you have any other thoughts on this? If not, could you help resolve the feedbacks? Appreciated!
</blockquote></details>

Sounds good.

YuhongWang-Amazon (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1070)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @YuhongWang-Amazon)

PR #1081, opened 2023-06-26
Remove unused dependencies from //src/main/kotlin/org/wfanet/panelmatch/common.

SanjayVas:
In general Bazel targets should follow "strict deps", meaning each target includes all dependencies that are used without relying on transitive dependencies. This prevents errors when transitive dependencies change. Eventually this will be enforced by Kotlin rules.

Furthermore, Bazel targets should not include unused deps. This prevents unnecessary rebuilding of targets.


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1081)

Marco-Premier (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1081)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @SanjayVas)

PR #1082, opened 2023-06-27
Add validation for simulator synthetic data specification

jcorilla:


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1082)

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1082)* status: 0 of 5 files reviewed, 7 unresolved discussions (waiting on @jcorilla)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/validation/SimulatorSyntheticDataSpecValidator.kt` line 26 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1082#-NYz0iyN2zo_oowxpTsI:-NYz0iyN2zo_oowxpTsJ:bq8as7a) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/92fa1c911c2dbf4746768f60cebcd63678078be0/src/main/kotlin/org/wfanet/measurement/loadtest/validation/SimulatorSyntheticDataSpecValidator.kt#L26)):*
> ```Kotlin
> import org.wfanet.measurement.common.toLocalDate
> 
> class SyntheticDataSpecValidationException(message: String? = null, cause: Throwable? = null) :
> ```

nit: potentially more accurate name that follows existing patterns



_Suggestion:_
```Kotlin
InvalidSyntheticDataSpecException
```
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/validation/SimulatorSyntheticDataSpecValidator.kt` line 30 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1082#-NYz-sDi6IR5XYfn0Ezg:-NYz-sDi6IR5XYfn0Ezh:b7ntqtw) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/92fa1c911c2dbf4746768f60cebcd63678078be0/src/main/kotlin/org/wfanet/measurement/loadtest/validation/SimulatorSyntheticDataSpecValidator.kt#L30)):*
> ```Kotlin
> 
> fun SimulatorSyntheticDataSpec.validate() {
>   val source = this
> ```

We only use this pattern in proto conversion extensions where we need to disambiguate `this` references.



_Code quote:_
```Kotlin
val source = this
```
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/validation/SimulatorSyntheticDataSpecValidator.kt` line 34 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1082#-NYz0EIh1jpd4SEWju-1:-NYz0EIh1jpd4SEWju-2:ba89ski) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/92fa1c911c2dbf4746768f60cebcd63678078be0/src/main/kotlin/org/wfanet/measurement/loadtest/validation/SimulatorSyntheticDataSpecValidator.kt#L34)):*
> ```Kotlin
>     throw SyntheticDataSpecValidationException(
>       "Invalid simulatorSyntheticDataSpec",
>       IllegalArgumentException("simulatorSyntheticDataSpec does not define a population.")
> ```

Don't create another exception just to have a cause here. The cause mechanism is to keep original context such as the original stack trace. Additionally, the exception type already indicates that it's an invalid spec. Just use the more specific message.
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/validation/SimulatorSyntheticDataSpecValidator.kt` line 61 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1082#-NYz1_J-CXxFOvjEuyNG:-NYz1_J-CXxFOvjEuyNH:bc2elhg) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/92fa1c911c2dbf4746768f60cebcd63678078be0/src/main/kotlin/org/wfanet/measurement/loadtest/validation/SimulatorSyntheticDataSpecValidator.kt#L61)):*
> ```Kotlin
>     val populationFieldValues = mutableSetOf<String>()
>     for (subPopulation in sortedSubPopulations) {
>       if (populationFieldValues.contains(subPopulation.populationFieldsValuesMap.toString())) {
> ```

The `toString()` value of a map is meant for human readability only. It shouldn't be used for anything else.

I'm not even certain what this is trying to accomplish. By the error message, I'm guessing this is meant to be checking whether there are any overlapping map entries, i.e. any entry in one sub-population fields map having the same (key, value) as an entry in another sub-population field map?



_Code quote:_
```Kotlin
subPopulation.populationFieldsValuesMap.toString()
```
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/validation/SimulatorSyntheticDataSpecValidator.kt` line 72 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1082#-NYz14FKCGaYxswGysG3:-NYz14FKCGaYxswGysG4:b-1e9xij) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/92fa1c911c2dbf4746768f60cebcd63678078be0/src/main/kotlin/org/wfanet/measurement/loadtest/validation/SimulatorSyntheticDataSpecValidator.kt#L72)):*
> ```Kotlin
> }
> 
> fun SyntheticPopulationSpec.SubPopulation.validate(syntheticPopulation: SyntheticPopulationSpec) {
> ```

Only the method for the top-level spec message should be public. All of these sub-field methods should be private.
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/validation/SimulatorSyntheticDataSpecValidator.kt` line 79 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1082#-NYz31A_AVvHGoxf8bDw:-NYz31A_AVvHGoxf8bDx:b-mcp9i4) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/92fa1c911c2dbf4746768f60cebcd63678078be0/src/main/kotlin/org/wfanet/measurement/loadtest/validation/SimulatorSyntheticDataSpecValidator.kt#L79)):*
> ```Kotlin
>       "Invalid subpopulation",
>       IllegalArgumentException("invalid population field values.")
>     )
> ```

This message can be more specific. Something like "Keys of sub-population map must match fields in population spec"



_Code quote:_
```Kotlin
    throw SyntheticDataSpecValidationException(
      "Invalid subpopulation",
      IllegalArgumentException("invalid population field values.")
    )
```
___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata/testing/BUILD.bazel` line 39 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1082#-NYz3S-7-9xjy5II0a3R:-NYz3S-7-9xjy5II0a3S:b-4dwbrk) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/92fa1c911c2dbf4746768f60cebcd63678078be0/src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata/testing/BUILD.bazel#L39)):*
> ```Starlark
> proto_library(
>     name = "simulator_synthetic_data_spec_proto",
>     testonly = False,
> ```

If it's in a testing package, it must be testonly. If it's not testonly, it should be in another package.

kungfucraig:
Abandoned

PR #1083, opened 2023-06-27
Add generation of events from SimulatorSyntheticDataSpec.

tristanvuong2021:


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1083)

SanjayVas (changes requested):
> Reviewed 1 of 4 files at r1, 1 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1083)* status: 2 of 4 files reviewed, 8 unresolved discussions (waiting on @kungfucraig and @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SimulatorSyntheticDataSpecEventQuery.kt` line 40 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1083#-NZ3E-lK0zu6VBs5pVSG:-NZ3E-lLEauNctiZO0XO:b-tec2fp) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6e05d15d09aad5848ac874fb85bcbaf35ec98eb9/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SimulatorSyntheticDataSpecEventQuery.kt#L40)):*
> ```Kotlin
>    * * with the [Descriptor].
>    */
>   fun generateEvents(
> ```

Split the generation into a separate type so it can be used independently of the EventQuery implementation. The EventQuery implementation can just call the generator.
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SimulatorSyntheticDataSpecEventQuery.kt` line 48 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1083#-NZ3F7B_6V1fUCkKADWM:-NZ3F7B_6V1fUCkKADWN:blny38i) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6e05d15d09aad5848ac874fb85bcbaf35ec98eb9/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SimulatorSyntheticDataSpecEventQuery.kt#L48)):*
> ```Kotlin
> 
>     return sequence {
>       simulatorSyntheticDataSpec.eventGroupSpecList.forEach { syntheticEventGroupSpec ->
> ```

Prefer regular for-in loops over `forEach` except when chaining. This is *especially* true in cases like this where the element is named.



_Suggestion:_
```Kotlin
for (syntheticEventGroupSpec in simulatorSyntheticDataSpec.eventGroupSpecList) {
```
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SimulatorSyntheticDataSpecEventQuery.kt` line 59 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1083#-NZ3FwyK-ao22hppX_le:-NZ3FwyK-ao22hppX_lf:b-23d6m5) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6e05d15d09aad5848ac874fb85bcbaf35ec98eb9/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SimulatorSyntheticDataSpecEventQuery.kt#L59)):*
> ```Kotlin
> 
>               population.populationFieldsList.forEach {
>                 val subPopulationFieldValue: String? = subPopulation.populationFieldsValuesMap[it]
> ```

Isn't every field in the spec supposed to have an associated value in the map?



_Suggestion:_
```Kotlin
val subPopulationFieldValue: String = subPopulation.populationFieldsValuesMap.getValue(it)
```
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SimulatorSyntheticDataSpecEventQuery.kt` line 66 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1083#-NZ3GSeo5pg5dsjqoFZQ:-NZ3GSeo5pg5dsjqoFZR:b-z9qjsb) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6e05d15d09aad5848ac874fb85bcbaf35ec98eb9/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SimulatorSyntheticDataSpecEventQuery.kt#L66)):*
> ```Kotlin
>                       fieldPath,
>                       subPopulationFieldValue,
>                       descriptor.findFieldByName(fieldPath.first())
> ```

I assume the NPE comes from using this? If so, explicitly check whether the return value of this is null rather than using a try/catch for the NPE.



_Code quote:_
```Kotlin
descriptor.findFieldByName(fieldPath.first())
```
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SimulatorSyntheticDataSpecEventQuery.kt` line 69 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1083#-NZ3GeaX7HY54-7uavQL:-NZ3GeaX7HY54-7uavQM:bfc9lkn) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6e05d15d09aad5848ac874fb85bcbaf35ec98eb9/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SimulatorSyntheticDataSpecEventQuery.kt#L69)):*
> ```Kotlin
>                     )
>                   } catch (_: NullPointerException) {
>                     throw IllegalArgumentException()
> ```

If we end up keeping this try/catch, the IllegalArgumentException should specify the NPE as the cause.



_Code quote:_
```Kotlin
throw IllegalArgumentException()
```
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SimulatorSyntheticDataSpecEventQuery.kt` line 96 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1083#-NZ3Gy_50P4K9UK6QBS2:-NZ3Gy_50P4K9UK6QBS3:b2ni8l0) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6e05d15d09aad5848ac874fb85bcbaf35ec98eb9/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SimulatorSyntheticDataSpecEventQuery.kt#L96)):*
> ```Kotlin
>               for (i in 1..numEvents) {
>                 yield(event)
>               }
> ```

nit



_Suggestion:_
```Kotlin
              repeat(numEvents) {
                yield(event)
              }
```
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SimulatorSyntheticDataSpecEventQuery.kt` line 140 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1083#-NZ3M9HvDhOfTLJqOiZI:-NZ3M9HvDhOfTLJqOiZJ:bjdwyqa) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6e05d15d09aad5848ac874fb85bcbaf35ec98eb9/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SimulatorSyntheticDataSpecEventQuery.kt#L140)):*
> ```Kotlin
>     if (fieldPath.size == 1) {
>       @Suppress("WHEN_ENUM_CAN_BE_NULL_IN_JAVA") // Proto enum fields are never null.
>       val fieldValue: Any =
> ```

I mentioned this in one of our design discussions. I'd rather update the spec message definition so that the map value is a message with a oneof with fields of the various types we support. That way the only type we need to convert is enum and we don't have to worry about precision issues.
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SimulatorSyntheticDataSpecEventQuery.kt` line 146 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1083#-NZ3Lxuw1ABlopLrz6aO:-NZ3Lxuw1ABlopLrz6aP:b-wqbs32) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6e05d15d09aad5848ac874fb85bcbaf35ec98eb9/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SimulatorSyntheticDataSpecEventQuery.kt#L146)):*
> ```Kotlin
>           FieldDescriptor.Type.BYTES -> value.toByteStringUtf8()
>           FieldDescriptor.Type.ENUM ->
>             curFieldDescriptor.enumType.findValueByName(value.uppercase(Locale.getDefault()))
> ```

Require enum values to be specified as integers.



_Code quote:_
```Kotlin
curFieldDescriptor.enumType.findValueByName(value.uppercase(Locale.getDefault()))
```

tristanvuong2021 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1083)* status: 1 of 5 files reviewed, 7 unresolved discussions (waiting on @kungfucraig and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SimulatorSyntheticDataSpecEventQuery.kt` line 40 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1083#-NZ3E-lK0zu6VBs5pVSG:-NZ6e7EF3vOnMf4_DTS9:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6e05d15d09aad5848ac874fb85bcbaf35ec98eb9/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SimulatorSyntheticDataSpecEventQuery.kt#L40)):*

Split the generation into a separate type so it can be used independently of the EventQuery implementation. The EventQuery implementation can just call the generator.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SimulatorSyntheticDataSpecEventQuery.kt` line 48 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1083#-NZ3F7B_6V1fUCkKADWM:-NZ6fuM85aDJ9xwyHPvE:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6e05d15d09aad5848ac874fb85bcbaf35ec98eb9/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SimulatorSyntheticDataSpecEventQuery.kt#L48)):*

Prefer regular for-in loops over `forEach` except when chaining. This is *especially* true in cases like this where the element is named.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SimulatorSyntheticDataSpecEventQuery.kt` line 59 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1083#-NZ3FwyK-ao22hppX_le:-NZ6fupL3WLMHP3_qT6Q:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6e05d15d09aad5848ac874fb85bcbaf35ec98eb9/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SimulatorSyntheticDataSpecEventQuery.kt#L59)):*

Isn't every field in the spec supposed to have an associated value in the map?
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SimulatorSyntheticDataSpecEventQuery.kt` line 66 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1083#-NZ3GSeo5pg5dsjqoFZQ:-NZ6fvT9CAMV3-P0dPc0:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6e05d15d09aad5848ac874fb85bcbaf35ec98eb9/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SimulatorSyntheticDataSpecEventQuery.kt#L66)):*

I assume the NPE comes from using this? If so, explicitly check whether the return value of this is null rather than using a try/catch for the NPE.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SimulatorSyntheticDataSpecEventQuery.kt` line 69 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1083#-NZ3GeaX7HY54-7uavQL:-NZ6fvwO1cPvjvO7Lkrg:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6e05d15d09aad5848ac874fb85bcbaf35ec98eb9/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SimulatorSyntheticDataSpecEventQuery.kt#L69)):*

If we end up keeping this try/catch, the IllegalArgumentException should specify the NPE as the cause.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SimulatorSyntheticDataSpecEventQuery.kt` line 140 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1083#-NZ3M9HvDhOfTLJqOiZI:-NZ6jdRW7kzZnDGDOJAO:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6e05d15d09aad5848ac874fb85bcbaf35ec98eb9/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SimulatorSyntheticDataSpecEventQuery.kt#L140)):*

I mentioned this in one of our design discussions. I'd rather update the spec message definition so that the map value is a message with a oneof with fields of the various types we support. That way the only type we need to convert is enum and we don't have to worry about precision issues.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SimulatorSyntheticDataSpecEventQuery.kt` line 146 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1083#-NZ3Lxuw1ABlopLrz6aO:-NZ6i4c_5pefgiQywwyd:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6e05d15d09aad5848ac874fb85bcbaf35ec98eb9/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SimulatorSyntheticDataSpecEventQuery.kt#L146)):*

Require enum values to be specified as integers.
</blockquote></details>

Done.

SanjayVas (changes requested):
> Reviewed 2 of 3 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1083)* status: 3 of 5 files reviewed, 10 unresolved discussions (waiting on @kungfucraig and @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/BUILD.bazel` line 16 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1083#-NZ83AT1D-jHkU7FiUxO:-NZ83AT1D-jHkU7FiUxP:bb0h41o) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/b7ff7b1fccd15279758a17bf1c7830bdec1f9fcb/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/BUILD.bazel#L16)):*
> ```Starlark
> )
> 
> kt_jvm_library(
> ```

Can be in a separate PR, but we'll want more granular libraries here since we'll want the test harness to be able to depend on a specific EventQuery impl. As it is, the edp_simulator binary should not directly depend on any EventQuery impls. It should just be dependent on the library for the EventQuery interface.

(e.g. we shouldn't have any BigQuery client library dependencies for a simulator that doesn't call BigQuery)
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SimulatorSyntheticDataSpecEventQuery.kt` line 40 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1083#-NZ3E-lK0zu6VBs5pVSG:-NZ81JQDDuf4cLjAbGW7:bst1smb) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6e05d15d09aad5848ac874fb85bcbaf35ec98eb9/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SimulatorSyntheticDataSpecEventQuery.kt#L40)):*

Done.
</blockquote></details>

Sorry, what I meant was an object in a separate file. Since the EventQuery interface isn't being implemented in this PR, the file and EventQuery class need not yet exist.

e.g. `SyntheticDataGeneration.kt` with `object SyntheticDataGeneration`.
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SimulatorSyntheticDataSpecEventQuery.kt` line 128 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1083#-NZ85cbEAX09pfRvpLea:-NZ85cbEAX09pfRvpLeb:b-w9ygp1) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/b7ff7b1fccd15279758a17bf1c7830bdec1f9fcb/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SimulatorSyntheticDataSpecEventQuery.kt#L128)):*
> ```Kotlin
>      * Throws
>      * * [IllegalArgumentException] if field is [FieldDescriptor.Type.MESSAGE].
>      * * [NullPointerException] if field can't be found.
> ```

Use `@throws` tag for these



_Code quote:_
```Kotlin
     * Throws
     * * [IllegalArgumentException] if field is [FieldDescriptor.Type.MESSAGE].
     * * [NullPointerException] if field can't be found.
```
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SimulatorSyntheticDataSpecEventQuery.kt` line 130 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1083#-NZ87Sqa1xVJ8SVWMw1l:-NZ87Sqa1xVJ8SVWMw1m:buybhne) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/b7ff7b1fccd15279758a17bf1c7830bdec1f9fcb/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SimulatorSyntheticDataSpecEventQuery.kt#L130)):*
> ```Kotlin
>      * * [NullPointerException] if field can't be found.
>      */
>     private fun Message.Builder.setField(
> ```

Wouldn't it be simpler to have this helper written as `private fun Message.Builder.setField(path: Collection<String>, value: FieldValue)`, and do the field lookup inside the helper?

Something like

```kotlin
private fun Message.Builder.setField(path: Collection<String>, value: FieldValue) {
  val fieldDescriptor = getDescriptorForType().findFieldByName(path.first()) ?: throw IllegalArgumentException()

  // Base case.
  if (path.size == 1) {
    // Assume field is a singular scalar field, since FieldValue doesn't support repeated values.
    setField(fieldDescriptor, getValue(value, fieldDescriptor))
    return
  }

  val nestedBuilder = getFieldBuilder(fieldDescriptor) ?: throw IllegalArgumentExeption()
  nestedBuilder.setField(path.drop(1), value)
}
```

If we want to support repeated fields, then FieldValue should have repeated versions of each field type. e.g. `repeated int32 int32_values = 13`
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SimulatorSyntheticDataSpecEventQuery.kt` line 140 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1083#-NZ85hKyDSHDxCD61cu5:-NZ85hKyDSHDxCD61cu6:bboygfd) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/b7ff7b1fccd15279758a17bf1c7830bdec1f9fcb/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SimulatorSyntheticDataSpecEventQuery.kt#L140)):*
> ```Kotlin
>         @Suppress("WHEN_ENUM_CAN_BE_NULL_IN_JAVA") // Proto enum fields are never null.
>         val value: Any =
>           when (curFieldDescriptor.type) {
> ```

Switch this check to use `fieldValue.valueCase`. This way you get an error if the field types mismatch rather than silently using a default value.

e.g. if the `double` field is set on FieldValue but the descriptor says it's supposed to be an `int32`, we'll get an error instead of silently using the default value in the unset `int32` field.
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SimulatorSyntheticDataSpecEventQuery.kt` line 176 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1083#-NZ8AzCNAtU9IiTOf5on:-NZ8AzCNAtU9IiTOf5oo:b37srv8) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/b7ff7b1fccd15279758a17bf1c7830bdec1f9fcb/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SimulatorSyntheticDataSpecEventQuery.kt#L176)):*
Not sure what's going on here. The merging should happen automatically. From the description of `getFieldBuilder`[^1]: 

> The original message field (if exist) will be merged into the field builder, which will then be nested into its parent builder.

[^1]: https://protobuf.dev/reference/java/api-docs/com/google/protobuf/Message.Builder.html#getFieldBuilder-com.google.protobuf.Descriptors.FieldDescriptor-
___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata/testing/simulator_synthetic_data_spec.proto` line 39 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1083#-NZ85-y0351g3H4mjiK9:-NZ85-y0351g3H4mjiKA:b29zbb6) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/b7ff7b1fccd15279758a17bf1c7830bdec1f9fcb/src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata/testing/simulator_synthetic_data_spec.proto#L39)):*
> ```proto
> message FieldValue {
>   oneof value {
>     string string = 1;
> ```

nit: Maybe suffix these with `_value` to avoid conflicting with reserved names in different languages?



_Suggestion:_
```proto
string_value
```

tristanvuong2021 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1083)* status: 0 of 5 files reviewed, 8 unresolved discussions (waiting on @kungfucraig and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SimulatorSyntheticDataSpecEventQuery.kt` line 40 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1083#-NZ3E-lK0zu6VBs5pVSG:-NZ8QNWAAT7ED8-9cFYV:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6e05d15d09aad5848ac874fb85bcbaf35ec98eb9/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SimulatorSyntheticDataSpecEventQuery.kt#L40)):*

Sorry, what I meant was an object in a separate file. Since the EventQuery interface isn't being implemented in this PR, the file and EventQuery class need not yet exist.

e.g. `SyntheticDataGeneration.kt` with `object SyntheticDataGeneration`.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SimulatorSyntheticDataSpecEventQuery.kt` line 128 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1083#-NZ85cbEAX09pfRvpLea:-NZ8QU734D6a_z9Kctye:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/b7ff7b1fccd15279758a17bf1c7830bdec1f9fcb/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SimulatorSyntheticDataSpecEventQuery.kt#L128)):*

Use `@throws` tag for these
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SimulatorSyntheticDataSpecEventQuery.kt` line 130 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1083#-NZ87Sqa1xVJ8SVWMw1l:-NZ8Txy84SaESXVrNME2:bw5hwin) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/b7ff7b1fccd15279758a17bf1c7830bdec1f9fcb/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SimulatorSyntheticDataSpecEventQuery.kt#L130)):*

Wouldn't it be simpler to have this helper written as `private fun Message.Builder.setField(path: Collection<String>, value: FieldValue)`, and do the field lookup inside the helper?

Something like

```kotlin
private fun Message.Builder.setField(path: Collection<String>, value: FieldValue) {
  val fieldDescriptor = getDescriptorForType().findFieldByName(path.first()) ?: throw IllegalArgumentException()

  // Base case.
  if (path.size == 1) {
    // Assume field is a singular scalar field, since FieldValue doesn't support repeated values.
    setField(fieldDescriptor, getValue(value, fieldDescriptor))
    return
  }

  val nestedBuilder = getFieldBuilder(fieldDescriptor) ?: throw IllegalArgumentExeption()
  nestedBuilder.setField(path.drop(1), value)
}
```

If we want to support repeated fields, then FieldValue should have repeated versions of each field type. e.g. `repeated int32 int32_values = 13`
</blockquote></details>

It has to be a message. I will just not support repeated now and will do it later if someone needs it.
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SimulatorSyntheticDataSpecEventQuery.kt` line 140 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1083#-NZ85hKyDSHDxCD61cu5:-NZ8T_CO5YzOTD2Whb_A:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/b7ff7b1fccd15279758a17bf1c7830bdec1f9fcb/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SimulatorSyntheticDataSpecEventQuery.kt#L140)):*

Switch this check to use `fieldValue.valueCase`. This way you get an error if the field types mismatch rather than silently using a default value.

e.g. if the `double` field is set on FieldValue but the descriptor says it's supposed to be an `int32`, we'll get an error instead of silently using the default value in the unset `int32` field.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SimulatorSyntheticDataSpecEventQuery.kt` line 176 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1083#-NZ8AzCNAtU9IiTOf5on:-NZ8WB118HyBVCnueUeP:b-t2hjuz) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/b7ff7b1fccd15279758a17bf1c7830bdec1f9fcb/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SimulatorSyntheticDataSpecEventQuery.kt#L176)):*

Not sure what's going on here. The merging should happen automatically. From the description of `getFieldBuilder`[^1]: 

> The original message field (if exist) will be merged into the field builder, which will then be nested into its parent builder.

[^1]: https://protobuf.dev/reference/java/api-docs/com/google/protobuf/Message.Builder.html#getFieldBuilder-com.google.protobuf.Descriptors.FieldDescriptor-
</blockquote></details>

Didn't notice that. Fixed it.

SanjayVas (changes requested):
> Reviewed 5 of 5 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1083)* status: all files reviewed, 4 unresolved discussions (waiting on @kungfucraig and @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SimulatorSyntheticDataSpecEventQuery.kt` line 128 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1083#-NZ85cbEAX09pfRvpLea:-NZ8bHhL8RSBV6AOatVo:bcg224p) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/b7ff7b1fccd15279758a17bf1c7830bdec1f9fcb/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SimulatorSyntheticDataSpecEventQuery.kt#L128)):*

Done.
</blockquote></details>

The format for the `@throws` tag is as follows:

```kotlin
  /**
   * Helper function for setting a field value in a [Message.Builder].
   *
   * @throws IllegalArgumentException if field is [FieldDescriptor.Type.MESSAGE]
   * @throws NullPointerException if field can't be found
   */
```
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SyntheticDataGeneration.kt` line 140 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1083#-NZ8c4Rq320799Qqx1cX:-NZ8c4Rq320799Qqx1cY:b-yk2w7q) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/dba6629db3d442a817cf7644605ea8dd17daa9ce/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SyntheticDataGeneration.kt#L140)):*
> ```Kotlin
>     val traversedFieldPath = fieldPath.drop(1)
>     nestedBuilder.setField(traversedFieldPath, fieldValue)
>     builder.setField(curFieldDescriptor, nestedBuilder.build())
> ```

This line is also unnecessary.



_Code quote:_
```Kotlin
 builder.setField(curFieldDescriptor, nestedBuilder.build())
```
___
*[`src/test/kotlin/org/wfanet/measurement/loadtest/dataprovider/SyntheticDataGenerationTest.kt` line 288 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1083#-NZ8cJpr8f9EhFBqw2hv:-NZ8cJpr8f9EhFBqw2hw:b-mbindr) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/dba6629db3d442a817cf7644605ea8dd17daa9ce/src/test/kotlin/org/wfanet/measurement/loadtest/dataprovider/SyntheticDataGenerationTest.kt#L288)):*
> ```Kotlin
>     }
> 
>     assertThat(testEventList).containsExactlyElementsIn(expectedTestEvents).inOrder()
> ```

There's no contract about the ordering of the generated data.



_Code quote:_
```Kotlin
.inOrder()
```

tristanvuong2021 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1083)* status: 3 of 5 files reviewed, 4 unresolved discussions (waiting on @kungfucraig and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SimulatorSyntheticDataSpecEventQuery.kt` line 128 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1083#-NZ85cbEAX09pfRvpLea:-NZ8jSJzDpdQTMkR4iJD:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/b7ff7b1fccd15279758a17bf1c7830bdec1f9fcb/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SimulatorSyntheticDataSpecEventQuery.kt#L128)):*

The format for the `@throws` tag is as follows:

```kotlin
  /**
   * Helper function for setting a field value in a [Message.Builder].
   *
   * @throws IllegalArgumentException if field is [FieldDescriptor.Type.MESSAGE]
   * @throws NullPointerException if field can't be found
   */
```

</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SyntheticDataGeneration.kt` line 140 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1083#-NZ8c4Rq320799Qqx1cX:-NZ8jTzV8T7Y9f-aJhRA:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/dba6629db3d442a817cf7644605ea8dd17daa9ce/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SyntheticDataGeneration.kt#L140)):*

This line is also unnecessary.
</blockquote></details>

Done.
___
*[`src/test/kotlin/org/wfanet/measurement/loadtest/dataprovider/SyntheticDataGenerationTest.kt` line 288 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1083#-NZ8cJpr8f9EhFBqw2hv:-NZ8jQOi8Et8nisoOp37:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/dba6629db3d442a817cf7644605ea8dd17daa9ce/src/test/kotlin/org/wfanet/measurement/loadtest/dataprovider/SyntheticDataGenerationTest.kt#L288)):*

There's no contract about the ordering of the generated data.
</blockquote></details>

Done.

SanjayVas (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1083)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @kungfucraig)

Marco-Premier (approved):
> Reviewed 3 of 5 files at r6, 2 of 2 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1083)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @kungfucraig and @stevenwarejones)

PR #1087, opened 2023-06-27
Use batch DDL for update-vid-model-foreign-keys-and-indexes.sql.

SanjayVas:
Without batch DDL, this changeset takes minutes to apply resulting in the update workflow timing out.


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1087)

PR #1094, opened 2023-07-06
Replace use of deprecated event_field message option in test templates.

SanjayVas:


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1094)

uakyol (approved):

PR #1095, opened 2023-07-07
Pass EventGroup to EventQuery.

SanjayVas:


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1095)

tristanvuong2021 (approved):
> Reviewed 9 of 9 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1095)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @SanjayVas)

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1095)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt` line 425 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1095#-N_cyRNPBNI_1GWzNLWc:-N_eDbEl9M5lDP1Rld0W:b-un95ra) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/42f8e675196984887c14c418ee9b2a112634e090/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt#L425)):*

NIT: Shall we be more specific on this error message? We throw SPEC_INVALID even if eventGroup is not found. May be hard to understand what the problem is
</blockquote></details>

The `orEmpty()` on the message is just to make the compiler happy, as the `message` property of `Throwable` is nullable. In practice, it will always be non-null.

In the case of an EventGroup not being found, the message will be "EventGroup <name> not found". The unit test verifies that the message contains the EventGroup name.

PR #1096, opened 2023-07-10
Add the state machine and the encryption methods proto for the reach …

ple13:
…only llv2 protocol. The state machine is very similar to that of the reach and frequency protocol, except that the execution phase has only one round.


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1096)

renjiezh (approved):
Looks good to me. I guess the `methods.proto` will be updated as needed during writing the c++ code.

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1096#-:-N_1GGkABelTaSJqZTLc:b-vt7q85)* status: all files reviewed, 3 unresolved discussions (waiting on @kungfucraig and @ple13)

ple13 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1096)* status: all files reviewed, 3 unresolved discussions (waiting on @kungfucraig and @SanjayVas)

___
*[`src/main/proto/wfa/measurement/internal/duchy/crypto.proto` line 65 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1096#-N_02lRVDdcTuk2Wi7ch:-N_2qTwW9jI3uExYPmAr:bvcwrik) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c9079781ca5b235aba45d339afb0f0dcec5a186d/src/main/proto/wfa/measurement/internal/duchy/crypto.proto#L65)):*

I can't find an obvious standard named this. The protobuf docs refer to a Ruby type used for large integers which would be equivalent to something like BigInteger in Java, but Wikipedia mentions this as an arbitrary precision type{^1] (e.g. BigDecimal in Java).

Looking up the Paillier cryptosystem, it looks like this should be an integer type. Is this supposed to be an integer that can be >64 bits? If so, define the encoding here. e.g. "An arbitrary-length unsigned integer in big endian byte order" or "An arbitrary-length signed integer in two's complement in big endian byte order"

[^https://en.wikipedia.org/wiki/Arbitrary-precision_arithmetic}
</blockquote></details>

The BigNum is the class that represents arbitrary length integers and is defined at https://github.com/google/private-join-and-compute/blob/master/private_join_and_compute/crypto/big_num.h. BigNum is currently being used in our code (for example, https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/main/src/main/cc/wfa/measurement/common/crypto/protocol_cryptor.h). The encoding and decoding are handled by the BigNum and Context class (https://github.com/google/private-join-and-compute/blob/master/private_join_and_compute/crypto/context.h), so explicitly specifying how the BigNum number is encoded would not be useful here. For Paillier, n has at least 4096 bits.
___
*[`src/main/proto/wfa/measurement/internal/duchy/crypto.proto` line 74 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1096#-N_01kmJ5LCxtDykqEn-:-N_3-if5Bi5B9evhFnMz:b-lpkhkg) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c9079781ca5b235aba45d339afb0f0dcec5a186d/src/main/proto/wfa/measurement/internal/duchy/crypto.proto#L74)):*

nit: comments should be on each field individually
</blockquote></details>

I agree that the comments should be on each field individually. But in this case, p and q are symmetric and the relationship n=pq should be specified. So, it's easier to have the comments as it is now.
___
*[`src/main/proto/wfa/measurement/internal/duchy/protocol/reach_only_liquid_legions_sketch_aggregation_v2.proto` line 157 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1096#-N_05Ok10w6JQmPCQXmR:-N_3-4XC2O2MxKKHv3vL:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c9079781ca5b235aba45d339afb0f0dcec5a186d/src/main/proto/wfa/measurement/internal/duchy/protocol/reach_only_liquid_legions_sketch_aggregation_v2.proto#L157)):*

Use documented TODO format referencing a GitHub issue or GitHub username: https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/main/docs/code-style.md#format
</blockquote></details>

Done.

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1096)* status: all files reviewed, 2 unresolved discussions (waiting on @kungfucraig and @ple13)

___
*[`src/main/proto/wfa/measurement/internal/duchy/crypto.proto` line 65 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1096#-N_02lRVDdcTuk2Wi7ch:-N_4pTjxAB2HqgJLwaKM:byw1nwi) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c9079781ca5b235aba45d339afb0f0dcec5a186d/src/main/proto/wfa/measurement/internal/duchy/crypto.proto#L65)):*
> The encoding and decoding are handled by the BigNum and Context class ([https://github.com/google/private-join-and-compute/blob/master/private\_join\_and\_compute/crypto/context.h](https://github.com/google/private-join-and-compute/blob/master/private_join_and_compute/crypto/context.h)), so explicitly specifying how the BigNum number is encoded would not be useful here.

One shouldn't assume that a reader of the API definition has any knowledge of the implementation. The API definition should have enough information such that one could implement the underlying services themselves.  That is to say it should be as implementation-agnostic as possible. At least mentioning that the encoding is from the C++ class `private_join_and_compute.BigNum` would be slightly better. Going further and indicating that it's a representation of an OpenSSL `BIGNUM` and linking [that documentation](https://www.openssl.org/docs/man1.0.2/man3/BN_bn2bin.html) would be better still. Avoiding referencing the implementation at all would be best.
___
*[`src/main/proto/wfa/measurement/internal/duchy/crypto.proto` line 74 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1096#-N_01kmJ5LCxtDykqEn-:-N_4pXG_34C5h1mtUStK:b-cb7t2k) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c9079781ca5b235aba45d339afb0f0dcec5a186d/src/main/proto/wfa/measurement/internal/duchy/crypto.proto#L74)):*

I agree that the comments should be on each field individually. But in this case, p and q are symmetric and the relationship n=pq should be specified. So, it's easier to have the comments as it is now.
</blockquote></details>

Note that part of the reason for the individual comments is for tooling. Some protobuf implementations copy leading comments to the respective field accessors. As a result, someone reading the documentation for the field accessor for `q` would find nothing. Giving the same description to both fields would be fine. e.g. "One of the two large prime factors of the Paillier modulus `n`."

kungfucraig (changes requested):
> Reviewed 2 of 4 files at r1, 1 of 1 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1096)* status: all files reviewed, 7 unresolved discussions (waiting on @ple13 and @SanjayVas)

___
*[`src/main/proto/wfa/measurement/internal/duchy/crypto.proto` line 65 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1096#-N_02lRVDdcTuk2Wi7ch:-N_4dmKl0uwjlPGIDtLh:b-j0pm4e) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c9079781ca5b235aba45d339afb0f0dcec5a186d/src/main/proto/wfa/measurement/internal/duchy/crypto.proto#L65)):*

> The encoding and decoding are handled by the BigNum and Context class ([https://github.com/google/private-join-and-compute/blob/master/private\_join\_and\_compute/crypto/context.h](https://github.com/google/private-join-and-compute/blob/master/private_join_and_compute/crypto/context.h)), so explicitly specifying how the BigNum number is encoded would not be useful here.

One shouldn't assume that a reader of the API definition has any knowledge of the implementation. The API definition should have enough information such that one could implement the underlying services themselves.  That is to say it should be as implementation-agnostic as possible. At least mentioning that the encoding is from the C++ class `private_join_and_compute.BigNum` would be slightly better. Going further and indicating that it's a representation of an OpenSSL `BIGNUM` and linking [that documentation](https://www.openssl.org/docs/man1.0.2/man3/BN_bn2bin.html) would be better still. Avoiding referencing the implementation at all would be best.
</blockquote></details>

As this isn't a public API, pointing to big_num.h in a comment seems reasonable.
___
*[`src/main/proto/wfa/measurement/internal/duchy/crypto.proto` line 74 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1096#-N_01kmJ5LCxtDykqEn-:-N_5IoJr5Ss11iiE4bVe:b5j7dqu) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c9079781ca5b235aba45d339afb0f0dcec5a186d/src/main/proto/wfa/measurement/internal/duchy/crypto.proto#L74)):*

Note that part of the reason for the individual comments is for tooling. Some protobuf implementations copy leading comments to the respective field accessors. As a result, someone reading the documentation for the field accessor for `q` would find nothing. Giving the same description to both fields would be fine. e.g. "One of the two large prime factors of the Paillier modulus `n`."
</blockquote></details>

+1 a little bit of redundancy in the commenting to reap the benefits of automation seems worthy.
___
*[`src/main/proto/wfa/measurement/internal/duchy/protocol/reach_only_liquid_legions_sketch_aggregation_v2.proto` line 1 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1096#-N_5JFNm3Hcg0AbAJqhq:-N_5JFNm3Hcg0AbAJqhr:b-2k6mxs) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/db288aadaff74969fe40390e112e3671912421cb/src/main/proto/wfa/measurement/internal/duchy/protocol/reach_only_liquid_legions_sketch_aggregation_v2.proto#L1)):*
> ```proto
> // Copyright 2020 The Cross-Media Measurement Authors
> ```

Update date
___
*[`src/main/proto/wfa/measurement/internal/duchy/protocol/reach_only_liquid_legions_sketch_aggregation_v2.proto` line 31 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1096#-N_5JVJXE4R8MoK3a7Hu:-N_5JVJXE4R8MoK3a7Hv:b-gi63ao) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/db288aadaff74969fe40390e112e3671912421cb/src/main/proto/wfa/measurement/internal/duchy/protocol/reach_only_liquid_legions_sketch_aggregation_v2.proto#L31)):*
> ```proto
>   enum Stage {
>     // The computation stage is unknown. This is never set intentionally.
>     STAGE_UNKNOWN = 0;
> ```

Rename to: STAGE_UNSPECIFIED

See: https://google.aip.dev/126
___
*[`src/main/proto/wfa/measurement/internal/duchy/protocol/reach_only_liquid_legions_sketch_aggregation_v2.proto` line 89 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1096#-N_5PKdHCY_R00aeHiJh:-N_5PKdHCY_R00aeHiJi:bcomkie) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/db288aadaff74969fe40390e112e3671912421cb/src/main/proto/wfa/measurement/internal/duchy/protocol/reach_only_liquid_legions_sketch_aggregation_v2.proto#L89)):*
> ```proto
>     EXECUTION_PHASE = 8;
> 
>     // The computation is done the worker can remove BLOBs that are no longer
> ```

Should we zero out the noise contribution in the DB as well?
___
*[`src/main/proto/wfa/measurement/internal/duchy/protocol/reach_only_liquid_legions_v2_encryption_methods.proto` line 1 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1096#-N_5JK241HxJjTPNmi5B:-N_5JK241HxJjTPNmi5C:b-2k6mxs) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/db288aadaff74969fe40390e112e3671912421cb/src/main/proto/wfa/measurement/internal/duchy/protocol/reach_only_liquid_legions_v2_encryption_methods.proto#L1)):*
> ```proto
> // Copyright 2020 The Cross-Media Measurement Authors
> ```

Update date
___
*[`src/main/proto/wfa/measurement/internal/duchy/protocol/reach_only_liquid_legions_v2_encryption_methods.proto` line 128 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1096#-N_5QDa36vZwEaMc2Qr4:-N_5QDa36vZwEaMc2Qr5:bii0ien) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/db288aadaff74969fe40390e112e3671912421cb/src/main/proto/wfa/measurement/internal/duchy/protocol/reach_only_liquid_legions_v2_encryption_methods.proto#L128)):*
> ```proto
>   bytes combined_register_vector = 1;
>   // The Paillier public key.
>   PaillierPublicKey paillier_public_key = 2;
> ```

Given this was present in the request it doesn't seem necessary to return it.

ple13 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1096)* status: 3 of 4 files reviewed, 7 unresolved discussions (waiting on @kungfucraig, @renjiezh, and @SanjayVas)

___
*[`src/main/proto/wfa/measurement/internal/duchy/crypto.proto` line 65 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1096#-N_02lRVDdcTuk2Wi7ch:-N_5Gt_X4JzzyM2zgECh:b221h9l) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c9079781ca5b235aba45d339afb0f0dcec5a186d/src/main/proto/wfa/measurement/internal/duchy/crypto.proto#L65)):*

As this isn't a public API, pointing to big_num.h in a comment seems reasonable.
</blockquote></details>

Agree. BigNum is replaced with private_joint_and_compute::BigNum.
___
*[`src/main/proto/wfa/measurement/internal/duchy/crypto.proto` line 74 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1096#-N_01kmJ5LCxtDykqEn-:-N_5H8OkE5db-hjr3EPh:bt2vr1b) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c9079781ca5b235aba45d339afb0f0dcec5a186d/src/main/proto/wfa/measurement/internal/duchy/crypto.proto#L74)):*

+1 a little bit of redundancy in the commenting to reap the benefits of automation seems worthy.
</blockquote></details>

Fixed.
___
*[`src/main/proto/wfa/measurement/internal/duchy/protocol/reach_only_liquid_legions_sketch_aggregation_v2.proto` line 1 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1096#-N_5JFNm3Hcg0AbAJqhq:-N_5RHenE7KCWoLMM-RL:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/db288aadaff74969fe40390e112e3671912421cb/src/main/proto/wfa/measurement/internal/duchy/protocol/reach_only_liquid_legions_sketch_aggregation_v2.proto#L1)):*

Update date
</blockquote></details>

Done.
___
*[`src/main/proto/wfa/measurement/internal/duchy/protocol/reach_only_liquid_legions_sketch_aggregation_v2.proto` line 31 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1096#-N_5JVJXE4R8MoK3a7Hu:-N_5RL2qDNOK9aEtbLLS:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/db288aadaff74969fe40390e112e3671912421cb/src/main/proto/wfa/measurement/internal/duchy/protocol/reach_only_liquid_legions_sketch_aggregation_v2.proto#L31)):*

Rename to: STAGE_UNSPECIFIED

See: https://google.aip.dev/126
</blockquote></details>

Done.
___
*[`src/main/proto/wfa/measurement/internal/duchy/protocol/reach_only_liquid_legions_sketch_aggregation_v2.proto` line 89 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1096#-N_5PKdHCY_R00aeHiJh:-N_5RrW69NVuVMOUQ5Eq:bxj1nbg) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/db288aadaff74969fe40390e112e3671912421cb/src/main/proto/wfa/measurement/internal/duchy/protocol/reach_only_liquid_legions_sketch_aggregation_v2.proto#L89)):*

Should we zero out the noise contribution in the DB as well?
</blockquote></details>

Yes. Comment has been updated.
___
*[`src/main/proto/wfa/measurement/internal/duchy/protocol/reach_only_liquid_legions_v2_encryption_methods.proto` line 1 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1096#-N_5JK241HxJjTPNmi5B:-N_5RUcH3SjYYO1F6W4j:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/db288aadaff74969fe40390e112e3671912421cb/src/main/proto/wfa/measurement/internal/duchy/protocol/reach_only_liquid_legions_v2_encryption_methods.proto#L1)):*

Update date
</blockquote></details>

Done.
___
*[`src/main/proto/wfa/measurement/internal/duchy/protocol/reach_only_liquid_legions_v2_encryption_methods.proto` line 128 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1096#-N_5QDa36vZwEaMc2Qr4:-N_5RY6U5V17OynfWZGp:b5ek4yh) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/db288aadaff74969fe40390e112e3671912421cb/src/main/proto/wfa/measurement/internal/duchy/protocol/reach_only_liquid_legions_v2_encryption_methods.proto#L128)):*

Given this was present in the request it doesn't seem necessary to return it.
</blockquote></details>

That's right. Updated.

kungfucraig (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1096)* status: all files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

___
*[`src/main/proto/wfa/measurement/internal/duchy/crypto.proto` line 65 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1096#-N_02lRVDdcTuk2Wi7ch:-N_5YUtt49GhOYNs63rZ:b-8oxedo) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c9079781ca5b235aba45d339afb0f0dcec5a186d/src/main/proto/wfa/measurement/internal/duchy/crypto.proto#L65)):*

Agree. BigNum is replaced with private_joint_and_compute::BigNum.
</blockquote></details>

Link to header file?

ple13 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1096)* status: all files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

___
*[`src/main/proto/wfa/measurement/internal/duchy/crypto.proto` line 65 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1096#-N_02lRVDdcTuk2Wi7ch:-N_5_GS7EHpwZncOkOdM:b2s11wh) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c9079781ca5b235aba45d339afb0f0dcec5a186d/src/main/proto/wfa/measurement/internal/duchy/crypto.proto#L65)):*

Link to header file?
</blockquote></details>

Added the link to the header file.

SanjayVas (changes requested):
> Reviewed 1 of 1 files at r5.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1096)* status: all files reviewed, 1 unresolved discussion (waiting on @kungfucraig and @ple13)

___
*[`src/main/proto/wfa/measurement/internal/duchy/crypto.proto` line 65 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1096#-N_02lRVDdcTuk2Wi7ch:-N_5kfgL0--lFdPjFTex:b-d36ydk) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c9079781ca5b235aba45d339afb0f0dcec5a186d/src/main/proto/wfa/measurement/internal/duchy/crypto.proto#L65)):*
> As this isn't a public API, pointing to big_num.h in a comment seems reasonable.

The "internal" vs. "public" distinction has no bearing on this. The comments I made were intended to apply to API definitions in general. Note that our "internal" distinction is really only to indicate that it's an API that wraps a persistence layer, which is separate from the "public" API to allow them to evolve independently.

It's not difficult to just paraphrase the definition from `BigNum::ToBytes` here to avoid referencing a specific implementation, which indicates that this is just an unsigned integer in big-endian form.

ple13 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1096)* status: 3 of 4 files reviewed, 1 unresolved discussion (waiting on @kungfucraig and @SanjayVas)

___
*[`src/main/proto/wfa/measurement/internal/duchy/crypto.proto` line 65 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1096#-N_02lRVDdcTuk2Wi7ch:-N_7-r1FEEAe_aXgNxpO:bzi61kl) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c9079781ca5b235aba45d339afb0f0dcec5a186d/src/main/proto/wfa/measurement/internal/duchy/crypto.proto#L65)):*

> As this isn't a public API, pointing to big_num.h in a comment seems reasonable.

The "internal" vs. "public" distinction has no bearing on this. The comments I made were intended to apply to API definitions in general. Note that our "internal" distinction is really only to indicate that it's an API that wraps a persistence layer, which is separate from the "public" API to allow them to evolve independently.

It's not difficult to just paraphrase the definition from `BigNum::ToBytes` here to avoid referencing a specific implementation, which indicates that this is just an unsigned integer in big-endian form.
</blockquote></details>

I think that the reference to BigNum in the comment is actually not useful and may cause confusion. I've updated the protobuf to remove BigNum reference in the comment and declare the PaillierPublicKey and PaillierPrivateKey the same way we declared ElGamal keys.

kungfucraig (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1096)* status: all files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

SanjayVas (approved):
> Reviewed 2 of 3 files at r4, 1 of 1 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1096)* status: all files reviewed, 1 unresolved discussion (waiting on @kungfucraig and @ple13)

___
*[`src/main/proto/wfa/measurement/internal/duchy/crypto.proto` line 65 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1096#-N_02lRVDdcTuk2Wi7ch:-N_B1nxJ83tWSB_Db_9V:b5kwiik) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c9079781ca5b235aba45d339afb0f0dcec5a186d/src/main/proto/wfa/measurement/internal/duchy/crypto.proto#L65)):*

I think that the reference to BigNum in the comment is actually not useful and may cause confusion. I've updated the protobuf to remove BigNum reference in the comment and declare the PaillierPublicKey and PaillierPrivateKey the same way we declared ElGamal keys.
</blockquote></details>

I still think that mentioning the format (big-endian unsigned integer) would be a good idea (on ElGamalPublicKey too), but not going to block on that.

I mention this because we've encountered an issue before related to assumed byte ordering (Guava's HashCode implementation outputs little-endian, but Java ByteBuffers use big-endian by default). As a result, we updated the documentation for our Requisition nonce hashes.

kungfucraig (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1096)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @ple13)

ple13 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1096)* status: 3 of 4 files reviewed, all discussions resolved (waiting on @kungfucraig, @Marco-Premier, and @SanjayVas)

___
*[`src/main/proto/wfa/measurement/internal/duchy/crypto.proto` line 65 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1096#-N_02lRVDdcTuk2Wi7ch:-N_F_V6HDNMQX8Ds4oYp:bd4s0eq) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c9079781ca5b235aba45d339afb0f0dcec5a186d/src/main/proto/wfa/measurement/internal/duchy/crypto.proto#L65)):*

I still think that mentioning the format (big-endian unsigned integer) would be a good idea (on ElGamalPublicKey too), but not going to block on that.

I mention this because we've encountered an issue before related to assumed byte ordering (Guava's HashCode implementation outputs little-endian, but Java ByteBuffers use big-endian by default). As a result, we updated the documentation for our Requisition nonce hashes.
</blockquote></details>

Thanks. I think we have many places like that..

SanjayVas (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1096)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Marco-Premier)

Marco-Premier (approved):
> Reviewed 1 of 1 files at r2, 1 of 3 files at r4, 1 of 1 files at r6, 1 of 1 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1096)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @ple13)

PR #1104, opened 2023-07-14
Add postgres computation readers

YuhongWang-Amazon:
Implement postgres computation readers


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1104)

YuhongWang-Amazon (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1104)* status: 0 of 6 files reviewed, 1 unresolved discussion

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/readers/ComputationReader.kt` line 40 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1104#-N_L6sQE8Ec7Gf_HK3hb:-N_L6sQGCAGDvJWGyg__:bkej54w) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/2c9ed62a560df8b890a6c7224415643eb3e8b8fd/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/readers/ComputationReader.kt#L40)):*
> ```Kotlin
> 
> /**
>  * Performs read operations on Computations and ComputationStages tables
> ```

@renjiezh to follow up on this [comment](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1089#-N_AxSGS7kcy5j7-RBcZ). I moved the requisition and blob queries to its own reader, and reference them from the computation reader. Since we always read Computations and ComputationStages tables together with JOIN statements. Hence, that one is left here.

YuhongWang-Amazon (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1104)* status: 4 of 6 files reviewed, 2 unresolved discussions (waiting on @renjiezh)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/readers/ComputationBlobReferenceReader.kt` line 76 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1104#-N_Zqye1E_jC6G7u0GDB:-N__FEb20p1th_ND-7HQ:ba0m2vk) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/2c9ed62a560df8b890a6c7224415643eb3e8b8fd/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/readers/ComputationBlobReferenceReader.kt#L76)):*

I wonder if we can rename it as `readBlobIdToPathMap`. I feel `ByDepType` here is kinda confusing because the key parameters are actually localId and stage. 
</blockquote></details>

Done. Thanks. That's a much better function name.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/readers/ComputationReader.kt` line 40 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1104#-N_L6sQE8Ec7Gf_HK3hb:-N__FMl10fHJZVHfa4fP:b-7rl0jq) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/2c9ed62a560df8b890a6c7224415643eb3e8b8fd/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/readers/ComputationReader.kt#L40)):*

Thanks. Yeah, you are right. Since we always read as ComputationToken so it has to link with the current ComputationStage. In the comment, maybe `with ComputationStages tables` is more precise.
</blockquote></details>

done!
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/readers/ComputationReader.kt` line 207 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1104#-N_Ztdiv3RRxlsVgSjnM:-N__Mkdb2fLPhfHpM52d:bqdvnu7) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/2c9ed62a560df8b890a6c7224415643eb3e8b8fd/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/readers/ComputationReader.kt#L207)):*

This way to read and build ComputationToken is much easier to understand thus less error-pruning.  However, I wonder if it has worse performance than the older way that using a large single SQL query joining three tables. 
Note that, readComputationToken is the most frequently used function so we want to make it as fast as we can. Could you share your insight of the performance impact?
</blockquote></details>

I don't have much experience with Postgres performances. Since the duchy data schema is relatively simple and the size of the duchy database should not be too large, I think the performance difference between a large single query and multiple smaller queries is probably the overhead of making database requests. (e.g. network latencies).
From my understanding the throughput of the duchy data services will not be too large, hence this overhead should not have much impact on the overall performance.

Since the smaller queries and functions are easier to read and manage, I prefer to keep it this way for now. We can always update this to JOINed queries, if there are any significant performance regressions.

renjiezh (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1104)* status: all files reviewed, 1 unresolved discussion (waiting on @YuhongWang-Amazon)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/readers/ComputationReader.kt` line 207 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1104#-N_Ztdiv3RRxlsVgSjnM:-N__USG29n8-3SZBm3ga:b5d2lqb) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/2c9ed62a560df8b890a6c7224415643eb3e8b8fd/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/readers/ComputationReader.kt#L207)):*

I don't have much experience with Postgres performances. Since the duchy data schema is relatively simple and the size of the duchy database should not be too large, I think the performance difference between a large single query and multiple smaller queries is probably the overhead of making database requests. (e.g. network latencies).
From my understanding the throughput of the duchy data services will not be too large, hence this overhead should not have much impact on the overall performance.

Since the smaller queries and functions are easier to read and manage, I prefer to keep it this way for now. We can always update this to JOINed queries, if there are any significant performance regressions.
</blockquote></details>

Sounds good to me.

renjiezh (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1104)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @YuhongWang-Amazon)

SanjayVas (changes requested):
> Reviewed 1 of 6 files at r1, 3 of 5 files at r2, 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1104)* status: all files reviewed, 3 unresolved discussions (waiting on @YuhongWang-Amazon)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/readers/ComputationBlobReferenceReader.kt` line 30 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1104#-N_epBh1EHNv3PscBwcv:-N_epBh1EHNv3PscBwcw:b2ni8l0) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ba5e04b6793379b4be178b625a65c47c01422449/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/readers/ComputationBlobReferenceReader.kt#L30)):*
> ```Kotlin
> 
>   /**
>    * Gets the [ComputationBlobDependency] of a computation.
> ```

nit



_Suggestion:_
```Kotlin
Reads 
```
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/readers/ComputationBlobReferenceReader.kt` line 113 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1104#-N_epJb65KULK7trQnxw:-N_epJb65KULK7trQnxx:b-3kevz6) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ba5e04b6793379b4be178b625a65c47c01422449/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/readers/ComputationBlobReferenceReader.kt#L113)):*
> ```Kotlin
>    * @param readContext The transaction context for reading from the Postgres database.
>    * @param localComputationId A local identifier for a computation
>    * @return A list of computation blob keys
> ```

FYI the `@return` line isn't required if it's obvious, e.g. from the summary fragment. Of course, it's never incorrect to have it.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/readers/ComputationBlobReferenceReader.kt` line 175 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1104#-N_eq5pd8LfXNQGHfR-3:-N_eq5pd8LfXNQGHfR-4:bxr2zho) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ba5e04b6793379b4be178b625a65c47c01422449/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/readers/ComputationBlobReferenceReader.kt#L175)):*
> ```Kotlin
>       dependencyType = row.getProtoEnum("DependencyType", ComputationBlobDependency::forNumber)
>     }
>   }
> ```

nit: prefer the more conventional `if (namedVariable != null) { ... }` over `namedVariable?.let { ... }` when smart casts can ensure that `namedVariable` is non-`null` inside the `then` branch.

This case is using it to avoid an extra named variable as well, so the choice is yours. I personally think that it's worth the extra variable in this case.



_Suggestion:_
```Kotlin
  private fun buildBlobMetadata(row: ResultRow): ComputationStageBlobMetadata {
    val path = row.get<String?>("PathToBlob")
    return computationStageBlobMetadata {
      blobId = row["BlobId"]
      if (path != null) {
        this.path = path
      }
      dependencyType = row.getProtoEnum("DependencyType", ComputationBlobDependency::forNumber)
    }
  }
```
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/readers/ComputationReader.kt` line 278 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1104#-N_evXumCNOJ8Ma27COy:-N_evXumCNOJ8Ma27COz:b-lrmgjn) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ba5e04b6793379b4be178b625a65c47c01422449/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/readers/ComputationReader.kt#L278)):*
> ```Kotlin
>         bind("$1", computationTypes[0])
>         updatedBefore?.let { bind("$2", it) }
>       }
> ```

Do separate bound statements rather than one with conditional binding when there's a difference in the number of bound variables. Note that it might be clearer to avoid having any bound variables in `baseSql`. If you look at the Kingdom's Spanner readers, `baseSql` tends not to include the `WHERE` clause at all.

Also prefer the more conventional `if (namedVariable != null) { ... }` over `namedVariable?.let { ... }` when smart casts can ensure that `namedVariable` is non-`null` inside the appropriate branch.



_Suggestion:_
```Kotlin
    val query =
      if (updatedBefore == null) {
        boundStatement(baseSql) {
          bind("$1", computationTypes[0])
        }
      } else {
        boundStatement(baseSql + " AND UpdatedTime <= $2") {
          bind("$1", computationTypes[0])
          bind("$2", updatedBefore)
        }
      }
```

YuhongWang-Amazon (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1104)* status: 2 of 25 files reviewed, 1 unresolved discussion (waiting on @renjiezh and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/readers/ComputationBlobReferenceReader.kt` line 113 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1104#-N_epJb65KULK7trQnxw:-N_iQutH4O5ZQ-JkViLq:b-m9exdc) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ba5e04b6793379b4be178b625a65c47c01422449/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/readers/ComputationBlobReferenceReader.kt#L113)):*

FYI the `@return` line isn't required if it's obvious, e.g. from the summary fragment. Of course, it's never incorrect to have it.
</blockquote></details>

Thanks. I will keep these returns here but keep that in mind for future works.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/readers/ComputationReader.kt` line 278 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1104#-N_evXumCNOJ8Ma27COy:-N_iUS-M8pQO3ESpRfZ9:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ba5e04b6793379b4be178b625a65c47c01422449/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/readers/ComputationReader.kt#L278)):*

Do separate bound statements rather than one with conditional binding when there's a difference in the number of bound variables. Note that it might be clearer to avoid having any bound variables in `baseSql`. If you look at the Kingdom's Spanner readers, `baseSql` tends not to include the `WHERE` clause at all.

Also prefer the more conventional `if (namedVariable != null) { ... }` over `namedVariable?.let { ... }` when smart casts can ensure that `namedVariable` is non-`null` inside the appropriate branch.
</blockquote></details>

Done.

SanjayVas (approved):
> Reviewed 3 of 23 files at r4, 19 of 20 files at r5, 1 of 1 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1104)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @YuhongWang-Amazon)

PR #1105, opened 2023-07-14
Add postgres computation writers

YuhongWang-Amazon:
Implement postgres computation writers


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1105)

YuhongWang-Amazon (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1105)* status: 0 of 16 files reviewed, 2 unresolved discussions

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/writers/AdvanceComputationStage.kt` line 58 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1105#-N_LBqJB3I3FvHUYUkUL:-N_LBqJB3I3FvHUYUkUM:bd2qbz1) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/579b18747e22d2a188189086a67b50f8814fc37b/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/writers/AdvanceComputationStage.kt#L58)):*
> ```Kotlin
>  */
> class AdvanceComputationStage<ProtocolT, StageT, StageDT : Message>(
>   private val localId: Long,
> ```

@renjiezh @SanjayVas following up on the discussion about expanding [ComputationEditToken](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0875884eab0ff7368d8350b15e11aa70aeec9e1e/src/main/kotlin/org/wfanet/measurement/duchy/db/computation/ComputationsDatabase.kt#L134C4-L134C4) into multiple parameters. 
My original intention is to avoid using anything from the db.computation package, so that it would be clearer on what are the required parameters. And would be easier for future refactoring.

But it turns out that I will be way too much efforts to re-create the enum-helper and [details-helper](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0875884eab0ff7368d8350b15e11aa70aeec9e1e/src/main/kotlin/org/wfanet/measurement/duchy/db/computation/ComputationProtocolStageDetailsHelper.kt) classes.

After look into all of its use cases, I noticed most of them only need the computationId and editVersion from the token. Hence I decided to keep this version of implementation. 

I don't mind change it back to the editToken version for the clean code purpose. Let me know what do you think?
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/writers/ComputationMutations.kt` line 30 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1105#-N_L882r7K7Rk2EZOKgW:-N_L882r7K7Rk2EZOKgX:bviuyq9) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/579b18747e22d2a188189086a67b50f8814fc37b/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/writers/ComputationMutations.kt#L30)):*
> ```Kotlin
> 
> /** Insert a new row into the Postgres Computations table. */
> suspend fun PostgresWriter.TransactionScope.insertComputation(
> ```

@renjiezh following up this [comment](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1089#-N_FQG-T7hnVhMttabC6). My understanding of the writer class is that each writer will only expose an execute method which will perform one or multiple queries in the same transaction.

Hence in my implementation I created one writer for each computationService method with the same name, as you mentioned. (e.g. [CreateComputation](src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/writers/CreateComputation.kt:49)).

However, there are some queries that are shared by multiple writers. (e.g. [updateComputation](src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/writers/ComputationMutations.kt:88)). I didn't implement these queries as a full PostgresWriter because they will not be executed alone by any service handlers. Therefore, I implemented them as extension functions of the TransactionScope.
I found it is better to logically group them into the mutation classes like this for each table.

SanjayVas (commented):
Waiting for Renjie to take a pass before reviewing.

> Reviewed 1 of 33 files at r3, 1 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1105#-:-N_tef7YBWqG06S9FyzL:b-vio5jl)* status: 2 of 50 files reviewed, 2 unresolved discussions (waiting on @renjiezh)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/writers/AdvanceComputationStage.kt` line 58 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1105#-N_LBqJB3I3FvHUYUkUL:-N_teAq39l340mXirhbM:b-wwphah) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/579b18747e22d2a188189086a67b50f8814fc37b/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/writers/AdvanceComputationStage.kt#L58)):*

@renjiezh @SanjayVas following up on the discussion about expanding [ComputationEditToken](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0875884eab0ff7368d8350b15e11aa70aeec9e1e/src/main/kotlin/org/wfanet/measurement/duchy/db/computation/ComputationsDatabase.kt#L134C4-L134C4) into multiple parameters. 
My original intention is to avoid using anything from the db.computation package, so that it would be clearer on what are the required parameters. And would be easier for future refactoring.

But it turns out that I will be way too much efforts to re-create the enum-helper and [details-helper](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0875884eab0ff7368d8350b15e11aa70aeec9e1e/src/main/kotlin/org/wfanet/measurement/duchy/db/computation/ComputationProtocolStageDetailsHelper.kt) classes.

After look into all of its use cases, I noticed most of them only need the computationId and editVersion from the token. Hence I decided to keep this version of implementation. 

I don't mind change it back to the editToken version for the clean code purpose. Let me know what do you think?
</blockquote></details>

So this is about whether it's okay to decompose and pass the necessary parts of `ComputationEditToken` vs. passing the token directly? I don't have a strong opinion here.

renjiezh (changes requested):
> Reviewed 14 of 16 files at r1, 2 of 36 files at r2, 32 of 33 files at r3, 2 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1105)* status: all files reviewed, 5 unresolved discussions (waiting on @YuhongWang-Amazon)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/writers/AdvanceComputationStage.kt` line 58 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1105#-N_LBqJB3I3FvHUYUkUL:-N_pWuxf8LMlei22O_w-:b-xx5x4q) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/579b18747e22d2a188189086a67b50f8814fc37b/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/writers/AdvanceComputationStage.kt#L58)):*

@renjiezh @SanjayVas following up on the discussion about expanding [ComputationEditToken](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0875884eab0ff7368d8350b15e11aa70aeec9e1e/src/main/kotlin/org/wfanet/measurement/duchy/db/computation/ComputationsDatabase.kt#L134C4-L134C4) into multiple parameters. 
My original intention is to avoid using anything from the db.computation package, so that it would be clearer on what are the required parameters. And would be easier for future refactoring.

But it turns out that I will be way too much efforts to re-create the enum-helper and [details-helper](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0875884eab0ff7368d8350b15e11aa70aeec9e1e/src/main/kotlin/org/wfanet/measurement/duchy/db/computation/ComputationProtocolStageDetailsHelper.kt) classes.

After look into all of its use cases, I noticed most of them only need the computationId and editVersion from the token. Hence I decided to keep this version of implementation. 

I don't mind change it back to the editToken version for the clean code purpose. Let me know what do you think?
</blockquote></details>

I think it is fine to save the effort as the impact is minor.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/writers/AdvanceComputationStage.kt` line 174 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1105#-N_p_N3IAnXrp2fZe1VX:-N_p_N3IAnXrp2fZe1VY:bdj7top) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/2200a40400b778fd008f10da7011f2300a063c6a/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/writers/AdvanceComputationStage.kt#L174)):*
> ```Kotlin
>         localId = localId,
>         stage = nextStageLong,
>         blobId = index.toLong(),
> ```

`blobId = index.toLong() + inputBlobPaths.size`
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/writers/AdvanceComputationStage.kt` line 176 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1105#-N_pZs7cD7CHCh6zByUB:-N_pZs7cD7CHCh6zByUC:b-fg5fjv) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/2200a40400b778fd008f10da7011f2300a063c6a/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/writers/AdvanceComputationStage.kt#L176)):*
> ```Kotlin
>         blobId = index.toLong(),
>         pathToBlob = path,
>         dependencyType = ComputationBlobDependency.INPUT
> ```

I guess this should be `ComputationBlobDependency.PASS_THOUGH`.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/writers/ComputationMutations.kt` line 30 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1105#-N_L882r7K7Rk2EZOKgW:-N_pQHjaC9CBSHPUrE3f:b-ba7jcz) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/579b18747e22d2a188189086a67b50f8814fc37b/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/writers/ComputationMutations.kt#L30)):*

@renjiezh following up this [comment](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1089#-N_FQG-T7hnVhMttabC6). My understanding of the writer class is that each writer will only expose an execute method which will perform one or multiple queries in the same transaction.

Hence in my implementation I created one writer for each computationService method with the same name, as you mentioned. (e.g. [CreateComputation](src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/writers/CreateComputation.kt:49)).

However, there are some queries that are shared by multiple writers. (e.g. [updateComputation](src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/writers/ComputationMutations.kt:88)). I didn't implement these queries as a full PostgresWriter because they will not be executed alone by any service handlers. Therefore, I implemented them as extension functions of the TransactionScope.
I found it is better to logically group them into the mutation classes like this for each table.
</blockquote></details>

Sounds good to me.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/writers/ComputationMutations.kt` line 58 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1105#-N_pRuTX9_hqIzPk5tDW:-N_pRuTX9_hqIzPk5tDX:bywkafl) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/2200a40400b778fd008f10da7011f2300a063c6a/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/writers/ComputationMutations.kt#L58)):*
> ```Kotlin
>           CreationTime
>         )
>       VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9::jsonb, $10)
> ```

I wonder if it is possible to avoid using `$n` as variable names? As there are a few variables, it is hard to map them. only by order. 
If it can be changed, I think using explicit names for SQL quries is better.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/writers/FinishComputation.kt` line 112 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1105#-N_pdstO6_mnAbKwAW2R:-N_pdstO6_mnAbKwAW2S:b-g7jmqk) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/2200a40400b778fd008f10da7011f2300a063c6a/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/writers/FinishComputation.kt#L112)):*
> ```Kotlin
>       val reason =
>         if (
>           unfinished.protocol == protocol &&
> ```

I am surprised that it needs to check the protocol given that a Computation with a localId must have a unique protocol. 
Consider remove the check for protocol. Or add a TODO comment if we want to play safe.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/writers/FinishComputation.kt` line 168 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1105#-N_peQ5D0m4Y7a99O1q1:-N_peQ5ECri-Z5ShTZmf:bs1liex) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/2200a40400b778fd008f10da7011f2300a063c6a/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/writers/FinishComputation.kt#L168)):*
> ```Kotlin
>       SELECT s.ComputationId, s.ComputationStage, s.Attempt, s.Details, c.Protocol
>       FROM ComputationStageAttempts as s
>       JOIN Computations AS c
> ```

As mentioned above, if the protocol is useless, no need to join Computations.
It will be better to move this function to ComputationStageAttemptReader as well.

YuhongWang-Amazon (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1105)* status: 47 of 50 files reviewed, 5 unresolved discussions (waiting on @renjiezh)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/writers/AdvanceComputationStage.kt` line 58 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1105#-N_LBqJB3I3FvHUYUkUL:-N_twIs76WTC-oJVJBsC:bu16rnt) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/579b18747e22d2a188189086a67b50f8814fc37b/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/writers/AdvanceComputationStage.kt#L58)):*

I think it is fine to save the effort as the impact is minor.
</blockquote></details>

Updated the FinishComputation and AdvantageComputationStage to use back the ComputationEditToken. The other use cases only needs the `localId` or `editVersion` from the token, which are simple enough to just pass in those two params.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/writers/AdvanceComputationStage.kt` line 174 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1105#-N_p_N3IAnXrp2fZe1VX:-N_txy4Z7Bv6a2v_-4kZ:b-t4jypw) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/2200a40400b778fd008f10da7011f2300a063c6a/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/writers/AdvanceComputationStage.kt#L174)):*

`blobId = index.toLong() + inputBlobPaths.size`
</blockquote></details>

Good catch! Thanks!
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/writers/AdvanceComputationStage.kt` line 176 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1105#-N_pZs7cD7CHCh6zByUB:-N_txzzd1SH088hPWeju:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/2200a40400b778fd008f10da7011f2300a063c6a/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/writers/AdvanceComputationStage.kt#L176)):*

I guess this should be `ComputationBlobDependency.PASS_THOUGH`.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/writers/ComputationMutations.kt` line 58 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1105#-N_pRuTX9_hqIzPk5tDW:-N_u0SOk6THutC7ObwRn:b-buhzmd) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/2200a40400b778fd008f10da7011f2300a063c6a/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/writers/ComputationMutations.kt#L58)):*

I wonder if it is possible to avoid using `$n` as variable names? As there are a few variables, it is hard to map them. only by order. 
If it can be changed, I think using explicit names for SQL quries is better.
</blockquote></details>

I also found this numbered variables are hard to manage, especially when I want to change the order of those. I didn't some research however, I didn't find any good workaround for this.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/writers/FinishComputation.kt` line 112 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1105#-N_pdstO6_mnAbKwAW2R:-N_u0gst68ce-vlK44gB:bymhv1b) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/2200a40400b778fd008f10da7011f2300a063c6a/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/writers/FinishComputation.kt#L112)):*

I am surprised that it needs to check the protocol given that a Computation with a localId must have a unique protocol. 
Consider remove the check for protocol. Or add a TODO comment if we want to play safe.
</blockquote></details>

I think you are right. The `protocol` from that unfinishedAttempt query was joined from the Computations table and filtered by `localComputationId`. hence there is no reason for it to mismatch. Deleted!
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/writers/FinishComputation.kt` line 168 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1105#-N_peQ5D0m4Y7a99O1q1:-N_tyyD03GeHp_-1iWBK:b-896fja) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/2200a40400b778fd008f10da7011f2300a063c6a/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/writers/FinishComputation.kt#L168)):*

As mentioned above, if the protocol is useless, no need to join Computations.
It will be better to move this function to ComputationStageAttemptReader as well.
</blockquote></details>

Done!

renjiezh (approved):
> Reviewed 3 of 3 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1105)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @YuhongWang-Amazon)

SanjayVas (approved):
> Reviewed 13 of 16 files at r1, 2 of 36 files at r2, 30 of 33 files at r3, 3 of 3 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1105)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @YuhongWang-Amazon)

PR #1122, opened 2023-07-19
Avoid unnecessary joins when updating Requisitions to UNFULFILLED.

SanjayVas:


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1122)

PR #1129, opened 2023-07-22
Implementation of the initialization, setup, and execution phase of the reach-only protocol.

ple13:
Implement the phases of the new sparse reach only protocol:

1. Initialization phase: Duchies sample local El Gamal keypair.
2. Setup phase:
a) Non-aggregators add noise registers to the crv and shuffle the modified crv. They encrypt their excessive noise using the composite El Gamal public key. They send the crv and the excessive noise ciphertext to the aggregator.
b) Aggregator: Waits for the crv and the excessive noise ciphertexts from non-aggregators. It adds noise to the crv, shuffles it, and encrypts its excessive noise with the composite El Gamal key. It then combines the excessive noise ciphertexts by adding them together. It sends the modified crv and the excessive noise ciphertext to the next worker.
3. Execution phase: Duchies collaborate to decrypt, randomize the register indices, and decrypt the excessive noise ciphertext. The aggregator can count the number of distinct registers, obtain the total amount of excessive noise, then estimate the reach based on the available information.


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1129)

SanjayVas (changes requested):
Reviewing for C++ style rather than protocol correctness.

> Reviewed 1 of 16 files at r1, 1 of 5 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1129#-:-Na7mzmW9daBzTWdymle:b-e16m60)* status: 2 of 16 files reviewed, 3 unresolved discussions (waiting on @kungfucraig, @ple13, and @renjiezh)

___
*[`src/main/cc/wfa/measurement/common/crypto/encryption_utility_helper.h` line 49 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1129#-Na7kOd7E8Aa8_x6vrLL:-Na7kOd8DV37Yn6Yx0y8:bdl9ftl) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7ed4d9b90feed62cb037f138c8a8a21706d3b556/src/main/cc/wfa/measurement/common/crypto/encryption_utility_helper.h#L49)):*
> ```objc
> // the deterministically encrypted results.
> absl::StatusOr<std::vector<std::string>> GetBlindedRegisterIndexes(
>     absl::string_view data, ProtocolCryptor& protocol_cryptor);
> ```

Do all of these methods only use a non-const reference to `ProtocolCryptor` because the methods are marked as non-const solely due to needing to mutate `mutex_`? That's one of the few cases where the `mutable` keyword is acceptable.

This can be handled in a separate PR as it doesn't directly relate to this one, but I'd recommend fixing in a "pre-factoring" PR before this one so we can avoid introducing new functions that unnecessarily have the ability to mutate the ProtocolCryptor instance.
___
*[`src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/liquid_legions_v2_encryption_utility_helper.cc` line 25 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1129#-Na7mH3JAd7bwoIdvwxv:-Na7mH3JAd7bwoIdvwxw:b-rnz51c) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7ed4d9b90feed62cb037f138c8a8a21706d3b556/src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/liquid_legions_v2_encryption_utility_helper.cc#L25)):*
> ```cpp
> using ::wfa::measurement::internal::duchy::ElGamalPublicKey;
> 
> ::wfa::any_sketch::crypto::ElGamalPublicKey ToAnysketchElGamalKey(
> ```

nit: "AnySketch" is treated as two words (e.g. the repo name is any-sketch).



_Suggestion:_
```cpp
ToAnySketchElGamalKey
```
___
*[`src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/liquid_legions_v2_encryption_utility_helper.cc` line 33 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1129#-Na7lpY-CcpksEl5O9H-:-Na7lpY-CcpksEl5O9H0:b-xbtupm) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7ed4d9b90feed62cb037f138c8a8a21706d3b556/src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/liquid_legions_v2_encryption_utility_helper.cc#L33)):*
> ```cpp
> }
> 
> ElGamalPublicKey ToCmmsElGamalKey(
> ```

This naming implies to me that it's converting to a CMMS public API `ElGamalPublicKey`, whereas this is converting to a Duchy-internal message.



_Suggestion:_
```cpp
ToDuchyElGamalKey
```

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1129)* status: 2 of 16 files reviewed, 3 unresolved discussions (waiting on @kungfucraig, @ple13, and @SanjayVas)

___
*[`src/main/cc/wfa/measurement/common/crypto/encryption_utility_helper.h` line 49 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1129#-Na7kOd7E8Aa8_x6vrLL:-Na8XlcS4XUUJ0j8aMXj:bac849f) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7ed4d9b90feed62cb037f138c8a8a21706d3b556/src/main/cc/wfa/measurement/common/crypto/encryption_utility_helper.h#L49)):*

Do all of these methods only use a non-const reference to `ProtocolCryptor` because the methods are marked as non-const solely due to needing to mutate `mutex_`? That's one of the few cases where the `mutable` keyword is acceptable.

This can be handled in a separate PR as it doesn't directly relate to this one, but I'd recommend fixing in a "pre-factoring" PR before this one so we can avoid introducing new functions that unnecessarily have the ability to mutate the ProtocolCryptor instance.
</blockquote></details>

The methods in `ProtocolCryptor` are intrinsically non-const because the implementation are based on operations from `private_join_and_compute`, which always change the `context_` for randomness.  (correct me if I am wrong or not precise.)
So I guess it is not feasible to use `const &ProtocolCryptor`.

kungfucraig (changes requested):
> Reviewed 2 of 16 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1129)* status: 4 of 16 files reviewed, 5 unresolved discussions (waiting on @ple13, @renjiezh, and @SanjayVas)

___
*[`src/main/proto/wfa/measurement/internal/duchy/protocol/reach_only_liquid_legions_v2_encryption_methods.proto` line 65 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1129#-Na7HapH39OULIAyLnaE:-Na7HapH39OULIAyLnaF:bvmdq3o) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7ed4d9b90feed62cb037f138c8a8a21706d3b556/src/main/proto/wfa/measurement/internal/duchy/protocol/reach_only_liquid_legions_v2_encryption_methods.proto#L65)):*
> ```proto
>   // Public Key of the composite ElGamal cipher. Used to encrypt the excessive
>   // noise (which is zero) when noise_parameters is not available.
>   ElGamalPublicKey composite_el_gamal_public_key = 5;
> ```

Why is this called "composite key?" It's just the aggregator's key, right?
___
*[`src/main/proto/wfa/measurement/internal/duchy/protocol/reach_only_liquid_legions_v2_encryption_methods.proto` line 157 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1129#-Na8bYQO9ue0WX6LNWcE:-Na8bYQO9ue0WX6LNWcF:b-nfbrhj) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7ed4d9b90feed62cb037f138c8a8a21706d3b556/src/main/proto/wfa/measurement/internal/duchy/protocol/reach_only_liquid_legions_v2_encryption_methods.proto#L157)):*
> ```proto
>   RegisterNoiseGenerationParameters noise_parameters = 8;
>   // The mechanism used to generate noise in previous phases.
>   LiquidLegionsV2NoiseConfig.NoiseMechanism noise_mechanism = 9;
> ```

Why are these two fields being added vs. the aggregator adding it's noise at the end of the setup phase?

ple13 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1129)* status: 3 of 16 files reviewed, 4 unresolved discussions (waiting on @kungfucraig, @renjiezh, and @SanjayVas)

___
*[`src/main/cc/wfa/measurement/common/crypto/encryption_utility_helper.h` line 49 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1129#-Na7kOd7E8Aa8_x6vrLL:-Na8g79O6GV2mXI4HFZp:b8jgp4y) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7ed4d9b90feed62cb037f138c8a8a21706d3b556/src/main/cc/wfa/measurement/common/crypto/encryption_utility_helper.h#L49)):*

The methods in `ProtocolCryptor` are intrinsically non-const because the implementation are based on operations from `private_join_and_compute`, which always change the `context_` for randomness.  (correct me if I am wrong or not precise.)
So I guess it is not feasible to use `const &ProtocolCryptor`.
</blockquote></details>

I tried to add the const keyword, then the compiler complained about the mutex_. The context_ will stay the same after it has been initialized, and does not need to be changed. The same context_ is used to generate different randomness.
___
*[`src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/liquid_legions_v2_encryption_utility_helper.cc` line 25 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1129#-Na7mH3JAd7bwoIdvwxv:-Na8gakn8yzAovObyPMg:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7ed4d9b90feed62cb037f138c8a8a21706d3b556/src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/liquid_legions_v2_encryption_utility_helper.cc#L25)):*

nit: "AnySketch" is treated as two words (e.g. the repo name is any-sketch).
</blockquote></details>

Done.
___
*[`src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/liquid_legions_v2_encryption_utility_helper.cc` line 33 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1129#-Na7lpY-CcpksEl5O9H-:-Na8gbp68cCreNI5vwSW:bbjdsct) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7ed4d9b90feed62cb037f138c8a8a21706d3b556/src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/liquid_legions_v2_encryption_utility_helper.cc#L33)):*

This naming implies to me that it's converting to a CMMS public API `ElGamalPublicKey`, whereas this is converting to a Duchy-internal message.
</blockquote></details>

This is the full ElGamal public key, which is shared across all computation parties (EDPs, Kingdom, Duchies). But you're right that the current usage only involves the duchies  (as the function is currently only used in the unit duchy tests).
___
*[`src/main/proto/wfa/measurement/internal/duchy/protocol/reach_only_liquid_legions_v2_encryption_methods.proto` line 65 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1129#-Na7HapH39OULIAyLnaE:-Na8ihwjB8hnIV7ymFtx:beqkmcg) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7ed4d9b90feed62cb037f138c8a8a21706d3b556/src/main/proto/wfa/measurement/internal/duchy/protocol/reach_only_liquid_legions_v2_encryption_methods.proto#L65)):*

Why is this called "composite key?" It's just the aggregator's key, right?
</blockquote></details>

This is the combined key from all the duchies (non-aggregator + aggregator). Using the composite key, a non-aggregator can send the pair (crv, noise_ciphertext) to the aggregator in the setup phase without leaking any information.
___
*[`src/main/proto/wfa/measurement/internal/duchy/protocol/reach_only_liquid_legions_v2_encryption_methods.proto` line 157 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1129#-Na8bYQO9ue0WX6LNWcE:-Na8k4KvDmajt5FJs9OA:bgv2hhy) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7ed4d9b90feed62cb037f138c8a8a21706d3b556/src/main/proto/wfa/measurement/internal/duchy/protocol/reach_only_liquid_legions_v2_encryption_methods.proto#L157)):*

Why are these two fields being added vs. the aggregator adding it's noise at the end of the setup phase?
</blockquote></details>

The aggregator needs to know the noise_mechanism and the noise_parameters to compute a) the max blind histogram noise (to generate the lookup table) and b) the number of baseline reach noise to subtract it from the register count.

renjiezh (approved):
> Reviewed 11 of 16 files at r1, 5 of 5 files at r2, 5 of 5 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1129)* status: all files reviewed, 3 unresolved discussions (waiting on @kungfucraig and @SanjayVas)

SanjayVas (changes requested):
> Reviewed 7 of 16 files at r1, 3 of 5 files at r2, 5 of 5 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1129)* status: all files reviewed, 6 unresolved discussions (waiting on @kungfucraig and @ple13)

___
*[`src/main/cc/wfa/measurement/common/crypto/encryption_utility_helper.h` line 49 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1129#-Na7kOd7E8Aa8_x6vrLL:-NaD2g3kARl8qMa8ZifQ:bsd78he) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7ed4d9b90feed62cb037f138c8a8a21706d3b556/src/main/cc/wfa/measurement/common/crypto/encryption_utility_helper.h#L49)):*
You would need to add the [`mutable`](https://en.cppreference.com/w/cpp/language/cv) specifier to the `mutex_` field to indicate that it can be changed even by `const` methods. Use of this specifier is rare, but a mutex field is one of the few cases where it would be valid.

> Mutable is used to specify that the member does not affect the externally visible state of the class (as often used for mutexes, memo caches, lazy evaluation, and access instrumentation).
___
*[`src/main/cc/wfa/measurement/common/crypto/encryption_utility_helper.cc` line 147 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1129#-NaD4d1Y2sSibkk8byF6:-NaD4d1Y2sSibkk8byF7:b-55tgn3) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c453df7ef9e72de352a7fefe504db48412e85882/src/main/cc/wfa/measurement/common/crypto/encryption_utility_helper.cc#L147)):*
> ```cpp
> absl::StatusOr<ElGamalEcPointPair> GetEcPointPairFromString(
>     absl::string_view str, int curve_id) {
>   std::unique_ptr<Context> context(new Context);
> ```

Prefer `make_unique` over using raw `new`. See https://abseil.io/tips/126

That said, it looks like the Context doesn't outlive this function scope since its ownership is never passed. You may be able to stack allocate it instead.

```cc
Context context;
ASSIGN_OR_RETURN(ECGroup ec_group, ECGroup::Create(curve_id, &context));
```



_Suggestion:_
```cpp
auto context = std::make_unique<Context>();
```
___
*[`src/main/cc/wfa/measurement/common/crypto/protocol_cryptor.h` line 75 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1129#-NaDFJKLA00HX6_hBpys:-NaDFJKLA00HX6_hBpyt:bpju43x) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c453df7ef9e72de352a7fefe504db48412e85882/src/main/cc/wfa/measurement/common/crypto/protocol_cryptor.h#L75)):*
> ```objc
>   // Encrypts an integer with the full composite ElGamal Key.
>   virtual absl::StatusOr<std::string>
>   EncryptIntegerWithCompositElGamalAndWriteToString(int64_t value) = 0;
> ```

nit: "Write" seems to imply that one would pass the string to be written to as opposed to returning the string. For consistency with other methods in this class, leave out the "with". Finally, you don't need to indicate the parameter type in the name (though you may wish to leave it in this case for consistency).



_Suggestion:_
```objc
EncryptToStringCompositeElGamal
```
___
*[`src/main/cc/wfa/measurement/common/crypto/protocol_cryptor.cc` line 196 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1129#-NaDJJ2PCxkXcLZFWalA:-NaDJJ2PCxkXcLZFWalB:bi6jdcb) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c453df7ef9e72de352a7fefe504db48412e85882/src/main/cc/wfa/measurement/common/crypto/protocol_cryptor.cc#L196)):*
> ```cpp
>     ciphertext.replace(0, kBytesPerEcPoint, temp);
>     ASSIGN_OR_RETURN(temp, zero_ec.e.ToBytesCompressed());
>     ciphertext.replace(kBytesPerEcPoint, kBytesPerEcPoint, temp);
> ```

nit: prefer scoped initializer to avoid variable reuse[^1]. This pattern is generally preferred over the ASSIGN_OR_RETURN macro, but we already use that macro broadly so consider consistency.

[^1]: https://abseil.io/tips/181



_Suggestion:_
```cpp
    if (absl::StatusOr<std::string> result = zero_ec.u.ToBytesCompressed(); result.ok()) {
      ciphertext.replace(0, kBytesPerEcPoint, *result);
    else {
      return result.status();
    }
    if (absl::StatusOr<std::string> result = zero_ec.e.ToBytesCompressed(); result.ok()) {
      ciphertext.replace(kBytesPerEcPoint, kBytesPerEcPoint, *result);
    } else {
      return result.status();
    }
    
```
___
*[`src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/liquid_legions_v2_encryption_utility_helper.cc` line 33 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1129#-Na7lpY-CcpksEl5O9H-:-NaDLGqY-qxPHvLn9By1:b-b5kqnq) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7ed4d9b90feed62cb037f138c8a8a21706d3b556/src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/liquid_legions_v2_encryption_utility_helper.cc#L33)):*

This is the full ElGamal public key, which is shared across all computation parties (EDPs, Kingdom, Duchies). But you're right that the current usage only involves the duchies  (as the function is currently only used in the unit duchy tests).
</blockquote></details>

"Duchy" here is referring to the namespace of the type, which parallels the use of "AnySketch" in the other function. You could also go more verbose and use "DuchyInternal" if you think that's clearer.

kungfucraig (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1129)* status: all files reviewed, 5 unresolved discussions (waiting on @ple13)

___
*[`src/main/proto/wfa/measurement/internal/duchy/protocol/reach_only_liquid_legions_v2_encryption_methods.proto` line 157 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1129#-Na8bYQO9ue0WX6LNWcE:-NaDVVSd2QrEKqZUUbhE:bdxx1ky) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7ed4d9b90feed62cb037f138c8a8a21706d3b556/src/main/proto/wfa/measurement/internal/duchy/protocol/reach_only_liquid_legions_v2_encryption_methods.proto#L157)):*

The aggregator needs to know the noise_mechanism and the noise_parameters to compute a) the max blind histogram noise (to generate the lookup table) and b) the number of baseline reach noise to subtract it from the register count.
</blockquote></details>

Got it! Thanks.

kungfucraig (changes requested):
> Reviewed 6 of 16 files at r1, 3 of 5 files at r2, 5 of 5 files at r3.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1129)* status: all files reviewed, 13 unresolved discussions (waiting on @ple13 and @SanjayVas)

___
*[`src/main/cc/wfa/measurement/common/crypto/encryption_utility_helper.cc` line 71 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1129#-NaDdQQ2-f2wrnnAiFPK:-NaDdQQ36sU4KLJ8Xkh9:b-gbg9xd) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c453df7ef9e72de352a7fefe504db48412e85882/src/main/cc/wfa/measurement/common/crypto/encryption_utility_helper.cc#L71)):*
> ```cpp
> }
> 
> absl::StatusOr<std::vector<std::string>> GetRollv2BlindedRegisterIndexes(
> ```

Should this use the multithread helper?

cc: @renjiezh 
___
*[`src/main/cc/wfa/measurement/common/crypto/encryption_utility_helper.cc` line 147 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1129#-NaD4d1Y2sSibkk8byF6:-NaDdbBZBu3Yfq1RR-_m:b-6vch1i) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c453df7ef9e72de352a7fefe504db48412e85882/src/main/cc/wfa/measurement/common/crypto/encryption_utility_helper.cc#L147)):*

Prefer `make_unique` over using raw `new`. See https://abseil.io/tips/126

That said, it looks like the Context doesn't outlive this function scope since its ownership is never passed. You may be able to stack allocate it instead.

```cc
Context context;
ASSIGN_OR_RETURN(ECGroup ec_group, ECGroup::Create(curve_id, &context));
```
</blockquote></details>

If possible, prefer the stack allocation.
___
*[`src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/BUILD.bazel` line 12 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1129#-NaDhd0D04iMYx9tHX9J:-NaDhd0D04iMYx9tHX9K:btrup6o) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c453df7ef9e72de352a7fefe504db48412e85882/src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/BUILD.bazel#L12)):*
> ```Starlark
> _INCLUDE_PREFIX = "/src/main/cc"
> 
> cc_library(
> ```

Can we mark this target test only?

https://bazel.build/reference/be/common-definitions#common.testonly
___
*[`src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/liquid_legions_v2_encryption_utility_helper.h` line 1 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1129#-NaDhFz10FQe1P5LHFH5:-NaDhFz10FQe1P5LHFH6:b-jqgo4m) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c453df7ef9e72de352a7fefe504db48412e85882/src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/liquid_legions_v2_encryption_utility_helper.h#L1)):*
> ```objc
> ```

It looks like these functions are only used in tests. Can you rename this file to end with "test_helper"? 

@SanjayVas is that the convention you want to use?
___
*[`src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/reach_only_liquid_legions_v2_encryption_utility.cc` line 184 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1129#-NaDuW4E4qK8l62IjfZn:-NaDuW4E4qK8l62IjfZo:bfwmd2p) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c453df7ef9e72de352a7fefe504db48412e85882/src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/reach_only_liquid_legions_v2_encryption_utility.cc#L184)):*
> ```cpp
>   ASSIGN_OR_RETURN(int64_t noise_registers_count,
>                    distributed_noiser.GenerateNoiseComponent());
>   // Make sure that there is at least one publisher noise added.
> ```

Can you comment in the file as to why this is important?
___
*[`src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/reach_only_liquid_legions_v2_encryption_utility.cc` line 573 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1129#-NaDy8EI0Y_was1gmvyA:-NaDy8EI0Y_was1gmvyB:bunbpwz) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c453df7ef9e72de352a7fefe504db48412e85882/src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/reach_only_liquid_legions_v2_encryption_utility.cc#L573)):*
> ```cpp
>         break;
>       }
>     }
> ```

Can we have a check that will throw an error if we do not successfully lookup the value?
___
*[`src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/reach_only_liquid_legions_v2_encryption_utility.cc` line 594 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1129#-NaDyUZL8275VJhsls1d:-NaDyUZL8275VJhsls1e:b-5qh9) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c453df7ef9e72de352a7fefe504db48412e85882/src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/reach_only_liquid_legions_v2_encryption_utility.cc#L594)):*
> ```cpp
>   // there exist publisher noise and padding noise.
>     non_empty_register_count -= 2;
> ```

Weren't there 2 registers per Duchy added? And weren't those factored into the excessive noise count anyway? 
___
*[`src/main/proto/wfa/measurement/internal/duchy/protocol/reach_only_liquid_legions_v2_encryption_methods.proto` line 67 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1129#-NaDbMsMBKOM6WAQFWdK:-NaDbMsMBKOM6WAQFWdL:bwv994m) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c453df7ef9e72de352a7fefe504db48412e85882/src/main/proto/wfa/measurement/internal/duchy/protocol/reach_only_liquid_legions_v2_encryption_methods.proto#L67)):*
> ```proto
>   ElGamalPublicKey composite_el_gamal_public_key = 5;
>   // The attached encrypted excessive noises. Only for the aggregator.
>   bytes serialized_excessive_noise_ciphertext = 6;
> ```

Can you mention that there will be one noise element for each non-aggregator worker?
___
*[`src/test/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/reach_only_liquid_legions_v2_encryption_utility_test.cc` line 96 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1129#-NaDqFvs23MOLm0tvjyY:-NaDqFvs23MOLm0tvjyZ:b-6ip1po) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c453df7ef9e72de352a7fefe504db48412e85882/src/test/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/reach_only_liquid_legions_v2_encryption_utility_test.cc#L96)):*
> ```cpp
> }
> 
> // The ReachOnlyTest generates cipher keys for 3 duchies, and the combined
> ```

Can you note in a comment which duchy is the aggregator?
___
*[`src/test/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/reach_only_liquid_legions_v2_encryption_utility_test.cc` line 253 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1129#-NaDqZCb8LVmzJd8Lv1o:-NaDqZCb8LVmzJd8Lv1p:b-65sxe3) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c453df7ef9e72de352a7fefe504db48412e85882/src/test/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/reach_only_liquid_legions_v2_encryption_utility_test.cc#L253)):*
> ```cpp
>                      CompleteReachOnlySetupPhaseAtAggregator(
>                          complete_reach_only_setup_phase_request_3));
>     EXPECT_THAT(
> ```

Should you check something about the excessive noise?

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1129)* status: all files reviewed, 13 unresolved discussions (waiting on @kungfucraig and @ple13)

___
*[`src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/liquid_legions_v2_encryption_utility_helper.h` line 1 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1129#-NaDhFz10FQe1P5LHFH5:-NaEIbJz5DrjznQWiDqm:b-vsq06q) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c453df7ef9e72de352a7fefe504db48412e85882/src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/liquid_legions_v2_encryption_utility_helper.h#L1)):*

It looks like these functions are only used in tests. Can you rename this file to end with "test_helper"? 

@SanjayVas is that the convention you want to use?
</blockquote></details>

The naming need not have "test" in it so long as it's under a `testing` Bazel package and the BUILD target has the `testonly` attribute set to `True`.

ple13 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1129)* status: all files reviewed, 13 unresolved discussions (waiting on @kungfucraig, @renjiezh, and @SanjayVas)

___
*[`src/main/cc/wfa/measurement/common/crypto/encryption_utility_helper.h` line 49 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1129#-Na7kOd7E8Aa8_x6vrLL:-NaDoxqN16a8uEmiuVOy:bs6bugf) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7ed4d9b90feed62cb037f138c8a8a21706d3b556/src/main/cc/wfa/measurement/common/crypto/encryption_utility_helper.h#L49)):*

You would need to add the [`mutable`](https://en.cppreference.com/w/cpp/language/cv) specifier to the `mutex_` field to indicate that it can be changed even by `const` methods. Use of this specifier is rare, but a mutex field is one of the few cases where it would be valid.

> Mutable is used to specify that the member does not affect the externally visible state of the class (as often used for mutexes, memo caches, lazy evaluation, and access instrumentation).
</blockquote></details>

Thanks. Let me file a bug to refactor the class ProtocolCryptor. 
___
*[`src/main/cc/wfa/measurement/common/crypto/encryption_utility_helper.cc` line 71 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1129#-NaDdQQ2-f2wrnnAiFPK:-NaIpKg2370Kg2w84Po1:b-pq5u5p) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c453df7ef9e72de352a7fefe504db48412e85882/src/main/cc/wfa/measurement/common/crypto/encryption_utility_helper.cc#L71)):*

Should this use the multithread helper?

cc: @renjiezh 
</blockquote></details>

Yes, this should be done with multithread helper.  The function has been updated.
___
*[`src/main/cc/wfa/measurement/common/crypto/encryption_utility_helper.cc` line 147 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1129#-NaD4d1Y2sSibkk8byF6:-NaDpyDA7kPglA6VIkhl:bfis22s) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c453df7ef9e72de352a7fefe504db48412e85882/src/main/cc/wfa/measurement/common/crypto/encryption_utility_helper.cc#L147)):*

If possible, prefer the stack allocation.
</blockquote></details>

Replaced it with stack allocation instead of pointer.
___
*[`src/main/cc/wfa/measurement/common/crypto/protocol_cryptor.h` line 75 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1129#-NaDFJKLA00HX6_hBpys:-NaDqvhF9j20p3U5fPOV:b-flvjh3) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c453df7ef9e72de352a7fefe504db48412e85882/src/main/cc/wfa/measurement/common/crypto/protocol_cryptor.h#L75)):*

nit: "Write" seems to imply that one would pass the string to be written to as opposed to returning the string. For consistency with other methods in this class, leave out the "with". Finally, you don't need to indicate the parameter type in the name (though you may wish to leave it in this case for consistency).
</blockquote></details>

Fixed. I used EncryptIntegerToStringCompositeElGamal to make it explicit that we are encrypting an integer.
___
*[`src/main/cc/wfa/measurement/common/crypto/protocol_cryptor.cc` line 196 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1129#-NaDJJ2PCxkXcLZFWalA:-NaDrii6Chz2sbpv9q63:bt2vr1b) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c453df7ef9e72de352a7fefe504db48412e85882/src/main/cc/wfa/measurement/common/crypto/protocol_cryptor.cc#L196)):*

nit: prefer scoped initializer to avoid variable reuse[^1]. This pattern is generally preferred over the ASSIGN_OR_RETURN macro, but we already use that macro broadly so consider consistency.

[^1]: https://abseil.io/tips/181
</blockquote></details>

Fixed.
___
*[`src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/BUILD.bazel` line 12 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1129#-NaDhd0D04iMYx9tHX9J:-NaIofiZ07KScLhMuWsu:bvof08h) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c453df7ef9e72de352a7fefe504db48412e85882/src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/BUILD.bazel#L12)):*

Can we mark this target test only?

https://bazel.build/reference/be/common-definitions#common.testonly
</blockquote></details>

Done. This has been moved into testing folder and marked with testonly attribute.
___
*[`src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/liquid_legions_v2_encryption_utility_helper.h` line 1 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1129#-NaDhFz10FQe1P5LHFH5:-NaIoO0q4nPeB77RZHPb:b-lynqdq) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c453df7ef9e72de352a7fefe504db48412e85882/src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/liquid_legions_v2_encryption_utility_helper.h#L1)):*

The naming need not have "test" in it so long as it's under a `testing` Bazel package and the BUILD target has the `testonly` attribute set to `True`.
</blockquote></details>

The files are now moved into the testing folder and has testonly attribute.
___
*[`src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/liquid_legions_v2_encryption_utility_helper.cc` line 33 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1129#-Na7lpY-CcpksEl5O9H-:-NaDrl5F-MNvWdqeZo3M:bt2vr1b) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7ed4d9b90feed62cb037f138c8a8a21706d3b556/src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/liquid_legions_v2_encryption_utility_helper.cc#L33)):*

"Duchy" here is referring to the namespace of the type, which parallels the use of "AnySketch" in the other function. You could also go more verbose and use "DuchyInternal" if you think that's clearer.
</blockquote></details>

Fixed.
___
*[`src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/reach_only_liquid_legions_v2_encryption_utility.cc` line 184 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1129#-NaDuW4E4qK8l62IjfZn:-NaIgyl_AI8DPnvBUlCC:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c453df7ef9e72de352a7fefe504db48412e85882/src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/reach_only_liquid_legions_v2_encryption_utility.cc#L184)):*

Can you comment in the file as to why this is important?
</blockquote></details>

Done.
___
*[`src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/reach_only_liquid_legions_v2_encryption_utility.cc` line 573 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1129#-NaDy8EI0Y_was1gmvyA:-NaISmk4CMMb0KyXJCK5:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c453df7ef9e72de352a7fefe504db48412e85882/src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/reach_only_liquid_legions_v2_encryption_utility.cc#L573)):*

Can we have a check that will throw an error if we do not successfully lookup the value?
</blockquote></details>

Done.
___
*[`src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/reach_only_liquid_legions_v2_encryption_utility.cc` line 594 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1129#-NaDyUZL8275VJhsls1d:-NaIQI5b2G2tunpqR31h:b-nn2f1m) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c453df7ef9e72de352a7fefe504db48412e85882/src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/reach_only_liquid_legions_v2_encryption_utility.cc#L594)):*

Weren't there 2 registers per Duchy added? And weren't those factored into the excessive noise count anyway? 
</blockquote></details>

They add noises with the same fixed ID (kPublisherNoiseRegisterId and kPaddingNoiseRegisterId), so the number of unique register IDs won't increase.
___
*[`src/main/proto/wfa/measurement/internal/duchy/protocol/reach_only_liquid_legions_v2_encryption_methods.proto` line 67 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1129#-NaDbMsMBKOM6WAQFWdK:-NaIQEi1AOq-waK1JHIg:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c453df7ef9e72de352a7fefe504db48412e85882/src/main/proto/wfa/measurement/internal/duchy/protocol/reach_only_liquid_legions_v2_encryption_methods.proto#L67)):*

Can you mention that there will be one noise element for each non-aggregator worker?
</blockquote></details>

Done.
___
*[`src/test/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/reach_only_liquid_legions_v2_encryption_utility_test.cc` line 96 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1129#-NaDqFvs23MOLm0tvjyY:-NaIRCMk6wMxxZqY7jEH:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c453df7ef9e72de352a7fefe504db48412e85882/src/test/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/reach_only_liquid_legions_v2_encryption_utility_test.cc#L96)):*

Can you note in a comment which duchy is the aggregator?
</blockquote></details>

Done.
___
*[`src/test/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/reach_only_liquid_legions_v2_encryption_utility_test.cc` line 253 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1129#-NaDqZCb8LVmzJd8Lv1o:-NaISWe1EVHzOKMN5pQS:bmxk0gt) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c453df7ef9e72de352a7fefe504db48412e85882/src/test/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/reach_only_liquid_legions_v2_encryption_utility_test.cc#L253)):*

Should you check something about the excessive noise? 
</blockquote></details>

I added the size check for the excessive noise.

ple13 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1129)* status: 3 of 23 files reviewed, 13 unresolved discussions (waiting on @kungfucraig, @renjiezh, and @SanjayVas)

___
*[`src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/reach_only_liquid_legions_v2_encryption_utility.cc` line 594 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1129#-NaDyUZL8275VJhsls1d:-NaJUUo6BESddcXjx7_B:b-8x7wmx) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c453df7ef9e72de352a7fefe504db48412e85882/src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/reach_only_liquid_legions_v2_encryption_utility.cc#L594)):*

They add noises with the same fixed ID (kPublisherNoiseRegisterId and kPaddingNoiseRegisterId), so the number of unique register IDs won't increase.
</blockquote></details>

I'm sorry I misread it. Currently we only add noise in the setup phase if the noise_parameters are set. So we check the noise_parameters before subtracting 2. However, I think the second check is not very intuitive and it's better to add 1 additional publisher/padding noise in any case, and just subtract 2 at the end. I'll update the code to reflect this change.

kungfucraig (approved):
> Reviewed 19 of 20 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1129)* status: 22 of 23 files reviewed, 5 unresolved discussions (waiting on @renjiezh and @SanjayVas)

SanjayVas (changes requested):
> Reviewed 20 of 20 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1129)* status: all files reviewed, 2 unresolved discussions (waiting on @ple13)

___
*[`src/main/cc/wfa/measurement/common/crypto/encryption_utility_helper.h` line 31 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1129#-NaORj03CO1RUcQWM0WX:-NaORj03CO1RUcQWM0WY:bimoa4f) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a9664fe8105c2cc91d1a1e5f52ab4e0a31c4cd9e/src/main/cc/wfa/measurement/common/crypto/encryption_utility_helper.h#L31)):*
> ```objc
> 
> using ::wfa::measurement::common::crypto::CompositeType;
> using ::wfa::measurement::internal::duchy::protocol::liquid_legions_v2::
> ```

A common crypto package shouldn't depend on an internal Duchy package. Move one or the other.
___
*[`src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/reach_only_liquid_legions_v2_encryption_utility.cc` line 577 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1129#-NaOS9Xv4jtXbAQBwZ-m:-NaOS9Xv4jtXbAQBwZ-n:b-ew82bc) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a9664fe8105c2cc91d1a1e5f52ab4e0a31c4cd9e/src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/reach_only_liquid_legions_v2_encryption_utility.cc#L577)):*
> ```cpp
>       }
>     }
>     // Throws an error if the decryption fails.
> ```

nit: we never throw anything



_Suggestion:_
```cpp
Returns
```

ple13 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1129)* status: 16 of 23 files reviewed, 1 unresolved discussion (waiting on @kungfucraig and @SanjayVas)

___
*[`src/main/cc/wfa/measurement/common/crypto/encryption_utility_helper.h` line 31 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1129#-NaORj03CO1RUcQWM0WX:-NaPD4Y95jWbAfl2AIil:b-ex615s) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a9664fe8105c2cc91d1a1e5f52ab4e0a31c4cd9e/src/main/cc/wfa/measurement/common/crypto/encryption_utility_helper.h#L31)):*

A common crypto package shouldn't depend on an internal Duchy package. Move one or the other.
</blockquote></details>

I moved the functions Get[Rollv2]BlindedRegisterIndexes into the [reach_only_]liquid_ligions_v2_encryption_utility.cc respectively.
___
*[`src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/reach_only_liquid_legions_v2_encryption_utility.cc` line 577 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1129#-NaOS9Xv4jtXbAQBwZ-m:-NaPCAqQ7pFiu1QGDck1:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a9664fe8105c2cc91d1a1e5f52ab4e0a31c4cd9e/src/main/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2/reach_only_liquid_legions_v2_encryption_utility.cc#L577)):*

nit: we never throw anything
</blockquote></details>

Done.

SanjayVas (approved):
> Reviewed 9 of 9 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1129)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @ple13)

stevenwarejones (approved):
> Reviewed 2 of 16 files at r1, 12 of 20 files at r4, 8 of 9 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1129)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @ple13)

benjaminkreuter:
This looks OK to me as far as protocol correctness goes.

benjaminkreuter (approved):
> Reviewed 2 of 16 files at r1, 12 of 20 files at r4, 9 of 9 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1129)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @ple13)

ple13 (commented):
Thanks Ben!

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1129#-:-Nb0n6RK9iAmbKeEvEXE:b-y53l14)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @ple13)

ple13 (commented):
> Reviewed 1 of 16 files at r1, 9 of 20 files at r4, 8 of 9 files at r5, 5 of 5 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1129)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @ple13)

PR #1130, opened 2023-07-24
Example on how to parse a row of the synthetic data and convert it to a Uk template 

uakyol:
This is just an example to convert a row of the data to the template. Other changes are likely needed to make this compile and work like : 
1. Specifying the correct columns to select from
2. Specifying types of the columns
3. The multi event group changes Origin needs to make to take the run based on the spec
4. Changing the DataproviderIds either in the synthetic data or the EDPSimulators to match each other.


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1130)

PR #1125, opened 2023-07-20
Duchy Api and System Api support Reach-Only Llv2 protocol.

renjiezh:

PR #1123, opened 2023-07-20
Update cross-media-measurement-api version

Marco-Premier:
Replace Timestamps with google.type.Date in ModelRolloutService Fix broken tests

Internal proto resources remained the same. What changed is the way public proto resources are converted into internal ones and viceversa.

PR #1136, opened 2023-07-26
Add postgres computations service

YuhongWang-Amazon:
This adds the postgres computations service and unit tests for it.


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1136)

renjiezh (changes requested):
> Reviewed 20 of 21 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1136)* status: all files reviewed, 7 unresolved discussions (waiting on @YuhongWang-Amazon)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/db/computation/ComputationTypes.kt` line 33 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1136#-NaNMqY99kEsNUyFmzvY:-NaNMqY99kEsNUyFmzvZ:bkb1y78) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/33e19735d28a450e3c023616a6096502dc59defd/src/main/kotlin/org/wfanet/measurement/duchy/db/computation/ComputationTypes.kt#L33)):*
> ```Kotlin
> }
> 
> fun ComputationStage.toComputationType() =
> ```

[This PR](https://github.com/world-federation-of-advertisers/cross-media-measurement/pull/1125) updates duchy to support REACH_ONLY_LIQUID_LEGIONS_V2 protocol. Some change might need to add for postgres's implementation.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/PostgresComputationsService.kt` line 305 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1136#-NaNOMSsERjX4Uhw4An_:-NaNOMSsERjX4Uhw4Ana:b-jzse0u) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/33e19735d28a450e3c023616a6096502dc59defd/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/PostgresComputationsService.kt#L305)):*
> ```Kotlin
>         endingStage = request.endingComputationStage,
>         endComputationReason =
>           when (val it = request.reason) {
> ```

nit: what about
```
when (request.reason) {
  ...
  else -> error("Unknown CompletedReason ${request.reason}")
}
```
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/PostgresComputationsService.kt` line 327 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1136#-NaNPEv_4O-W2f6Y0QQy:-NaNPEv_4O-W2f6Y0QQz:b7v68fv) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/33e19735d28a450e3c023616a6096502dc59defd/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/PostgresComputationsService.kt#L327)):*
> ```Kotlin
>     val token =
>       computationReader.readComputationToken(client, request.token.globalComputationId)
>         ?: failGrpc(Status.INTERNAL) {
> ```

might be Status.NOT_FOUND. 
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/PostgresComputationsService.kt` line 351 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1136#-NaNPIC75RoXjbzv6R2K:-NaNPIC75RoXjbzv6R2L:b-7ropln) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/33e19735d28a450e3c023616a6096502dc59defd/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/PostgresComputationsService.kt#L351)):*
> ```Kotlin
>     val token =
>       computationReader.readComputationToken(client, request.token.globalComputationId)
>         ?: failGrpc(Status.INTERNAL) {
> ```

ditto
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/PostgresComputationsService.kt` line 373 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1136#-NaNPJIA8OVlP7oBGOb7:-NaNPJIA8OVlP7oBGOb8:b-7ropln) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/33e19735d28a450e3c023616a6096502dc59defd/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/PostgresComputationsService.kt#L373)):*
> ```Kotlin
>     val token =
>       computationReader.readComputationToken(client, request.token.globalComputationId)
>         ?: failGrpc(Status.INTERNAL) {
> ```

ditto
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/PostgresComputationsService.kt` line 428 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1136#-NaNPRCW3H5irMWG63Qd:-NaNPRCW3H5irMWG63Qe:bqvqzbf) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/33e19735d28a450e3c023616a6096502dc59defd/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/PostgresComputationsService.kt#L428)):*
> ```Kotlin
>   ): GetComputationIdsResponse {
>     val ids = computationReader.readGlobalComputationIds(client.singleUse(), request.stagesList)
>     return GetComputationIdsResponse.newBuilder().addAllGlobalIds(ids).build()
> ```

Please avoid using builder.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/PostgresComputationsService.kt` line 460 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1136#-NaNPj6OBQcjfFjLqcU4:-NaNPj6OBQcjfFjLqcU5:b-rqxvo7) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/33e19735d28a450e3c023616a6096502dc59defd/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/PostgresComputationsService.kt#L460)):*
> ```Kotlin
>     val token =
>       computationReader.readComputationToken(client, request.token.globalComputationId)
>         ?: failGrpc(Status.INTERNAL) {
> ```

ditto.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/service/internal/testing/ComputationsServiceTest.kt` line 532 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1136#-NaNTIs3DLUKUhoyxtZ-:-NaNTIs3DLUKUhoyxtZ0:b79tyx6) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/33e19735d28a450e3c023616a6096502dc59defd/src/main/kotlin/org/wfanet/measurement/duchy/service/internal/testing/ComputationsServiceTest.kt#L532)):*
> ```Kotlin
> 
>       val nextStage = computationStage {
>         liquidLegionsSketchAggregationV2 = Stage.WAIT_REQUISITIONS_AND_KEY_SET
> ```

Nit:To make the unit test more realistic, this stage could be `Stage.CONFIRMATION_PHASE`.
Check the Stage to AfterTransition map [here](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/83512a46c81f0cda45d20439e45e86089eab77bc/src/main/kotlin/org/wfanet/measurement/duchy/db/computation/LiquidLegionsSketchAggregationV2Protocol.kt#L105-L127)

YuhongWang-Amazon (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1136)* status: 20 of 21 files reviewed, 5 unresolved discussions (waiting on @renjiezh)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/db/computation/ComputationTypes.kt` line 33 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1136#-NaNMqY99kEsNUyFmzvY:-NaRicGb6GLclyL2T7xQ:b-4r0hr1) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/33e19735d28a450e3c023616a6096502dc59defd/src/main/kotlin/org/wfanet/measurement/duchy/db/computation/ComputationTypes.kt#L33)):*

[This PR](https://github.com/world-federation-of-advertisers/cross-media-measurement/pull/1125) updates duchy to support REACH_ONLY_LIQUID_LEGIONS_V2 protocol. Some change might need to add for postgres's implementation.
</blockquote></details>

Thanks for the heads up. When do you plan to merge that PR? I can rebase after you merged that one.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/PostgresComputationsService.kt` line 327 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1136#-NaNPEv_4O-W2f6Y0QQy:-NaRjVvw2EWFdUSY_QaL:b-heumjf) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/33e19735d28a450e3c023616a6096502dc59defd/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/PostgresComputationsService.kt#L327)):*

might be Status.NOT_FOUND. 
</blockquote></details>

We had this [discussion last time](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1070#-NYPH4zF8m6UGwt_qvwP). Since in this use case, a null computation token should only be caused by system invariant, we decided to use `Status.INTERNAL`. 
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/PostgresComputationsService.kt` line 428 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1136#-NaNPRCW3H5irMWG63Qd:-NaRk1DO6o5Ft9CZC1yX:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/33e19735d28a450e3c023616a6096502dc59defd/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/PostgresComputationsService.kt#L428)):*

Please avoid using builder.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/service/internal/testing/ComputationsServiceTest.kt` line 532 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1136#-NaNTIs3DLUKUhoyxtZ-:-NaRrgqvC35AVBBhJ0mv:b-90yngk) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/33e19735d28a450e3c023616a6096502dc59defd/src/main/kotlin/org/wfanet/measurement/duchy/service/internal/testing/ComputationsServiceTest.kt#L532)):*

Nit:To make the unit test more realistic, this stage could be `Stage.CONFIRMATION_PHASE`.
Check the Stage to AfterTransition map [here](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/83512a46c81f0cda45d20439e45e86089eab77bc/src/main/kotlin/org/wfanet/measurement/duchy/db/computation/LiquidLegionsSketchAggregationV2Protocol.kt#L105-L127)
</blockquote></details>

It was not trivial to change this stage to CONFIRMATION_PHASE. Since the only valid transition for `INITIALIZATION_PHASE` is the `WAIT_REQUISITIONS_AND_KEY_SET`. Many setup are required to advance a computation from `INITIALIZATION_PHASE` to `CONFIRMATION_PHASE`.

Is it ok to just use this for a unit test? Or are there any better ways to prepare a ready to be advanced `WAIT_REQUISITIONS_AND_KEY_SET` phase computation?

SanjayVas (commented):
Waiting for @renjiezh to complete his review first.

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1136#-:-NaTUi_WESOnzfI7ljM1:bn4g280)* status: 20 of 21 files reviewed, 5 unresolved discussions (waiting on @renjiezh and @YuhongWang-Amazon)

renjiezh (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1136#-:-NaTVea01Ca9CfpPMNwd:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @YuhongWang-Amazon)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/db/computation/ComputationTypes.kt` line 33 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1136#-NaNMqY99kEsNUyFmzvY:-NaTUxzqE9l-9sz9G83I:b-s2ayvu) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/33e19735d28a450e3c023616a6096502dc59defd/src/main/kotlin/org/wfanet/measurement/duchy/db/computation/ComputationTypes.kt#L33)):*

Thanks for the heads up. When do you plan to merge that PR? I can rebase after you merged that one.
</blockquote></details>

I was blocked by another one (https://github.com/world-federation-of-advertisers/cross-media-measurement/pull/1123). I will merge it once this blocker solved,
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/service/internal/testing/ComputationsServiceTest.kt` line 532 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1136#-NaNTIs3DLUKUhoyxtZ-:-NaTVKVe7Z8O35uYndHG:b-ts707) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/33e19735d28a450e3c023616a6096502dc59defd/src/main/kotlin/org/wfanet/measurement/duchy/service/internal/testing/ComputationsServiceTest.kt#L532)):*

It was not trivial to change this stage to CONFIRMATION_PHASE. Since the only valid transition for `INITIALIZATION_PHASE` is the `WAIT_REQUISITIONS_AND_KEY_SET`. Many setup are required to advance a computation from `INITIALIZATION_PHASE` to `CONFIRMATION_PHASE`.

Is it ok to just use this for a unit test? Or are there any better ways to prepare a ready to be advanced `WAIT_REQUISITIONS_AND_KEY_SET` phase computation?
</blockquote></details>

Makes sense. Let's save the effort.

SanjayVas (changes requested):
> Reviewed 5 of 21 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1136)* status: all files reviewed, 4 unresolved discussions (waiting on @YuhongWang-Amazon)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/PostgresComputationsService.kt` line 118 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1136#-NaTZ2jo-cjUhnqi04p0:-NaTZ2jo-cjUhnqi04p1:b-djeqwu) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/313ffcdd9d1e4b28a41796d7e377781cda99cf3e/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/PostgresComputationsService.kt#L118)):*
> ```Kotlin
>   private val idGenerator: IdGenerator,
>   private val duchyName: String,
>   private val computationStorageClient: ComputationStore,
> ```

`ComputationStore` implements the `Store` interface and wraps a `StorageClient` instance.



_Suggestion:_
```Kotlin
computationStore
```
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/PostgresComputationsService.kt` line 227 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1136#-NaTZfyV5YdjI4kc5haE:-NaTZfyV5YdjI4kc5haF:b-hzra90) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/313ffcdd9d1e4b28a41796d7e377781cda99cf3e/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/PostgresComputationsService.kt#L227)):*
> ```Kotlin
>       try {
>         computationStorageClient.get(blobKey)?.delete()
>       } catch (e: StatusException) {
> ```

A `Store` shouldn't be leaking gRPC `StatusException`, especially since not every underlying implementation even uses gRPC. If the blob isn't found `get` should just return null.



_Code quote:_
```Kotlin
```
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/PostgresComputationsService.kt` line 273 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1136#-NaTaBtb9vLLZVcX54m2:-NaTaBtb9vLLZVcX54m3:b-xnfi0w) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/313ffcdd9d1e4b28a41796d7e377781cda99cf3e/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/PostgresComputationsService.kt#L273)):*
> ```Kotlin
> 
>         if (!isTerminated(computationStageEnum)) {
>           FinishComputation(
> ```

@renjiezh This behavior doesn't look correct to me. I thought we only delete finished computations? My understanding was that we have separate TTLs for finished pending computations than for deleting finished computations? We shouldn't be terminating and deleting the same computation in the same RPC.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/PostgresComputationsService.kt` line 326 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1136#-NaTbQ0lCqK8Dqpc_hJP:-NaTbQ0lCqK8Dqpc_hJQ:bcnnqzg) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/313ffcdd9d1e4b28a41796d7e377781cda99cf3e/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/PostgresComputationsService.kt#L326)):*
> ```Kotlin
>     )
> 
>     val token =
> ```

It looks like this may be repeating an issue that exists in the Spanner implementation: the token isn't read in the same transaction that performs the mutation.

The FinishComputation operation should return the updated token.

renjiezh (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1136)* status: all files reviewed, 4 unresolved discussions (waiting on @SanjayVas and @YuhongWang-Amazon)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/PostgresComputationsService.kt` line 273 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1136#-NaTaBtb9vLLZVcX54m2:-NaTcpSUE5DZ-CltaYcQ:bfx1oo0) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/313ffcdd9d1e4b28a41796d7e377781cda99cf3e/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/PostgresComputationsService.kt#L273)):*

@renjiezh This behavior doesn't look correct to me. I thought we only delete finished computations? My understanding was that we have separate TTLs for finished pending computations than for deleting finished computations? We shouldn't be terminating and deleting the same computation in the same RPC.
</blockquote></details>

purgeComputations in the service is capable to deletes any Computations with the respect to the TTL. However, the request has the field `stages` to restrict t stage of Computations to delete. The caller is the cronjob ComputationCleaner, which only request to delete COMPLETE  Computations. 
The FinishComputation here is to handle purge requests for unfinished computations. However, it is never executed in our system yet. So I think it is fine.

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1136)* status: all files reviewed, 4 unresolved discussions (waiting on @renjiezh and @YuhongWang-Amazon)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/PostgresComputationsService.kt` line 273 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1136#-NaTaBtb9vLLZVcX54m2:-NaTgAdB1BxiCNNMOyAr:b-sovv3p) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/313ffcdd9d1e4b28a41796d7e377781cda99cf3e/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/PostgresComputationsService.kt#L273)):*

purgeComputations in the service is capable to deletes any Computations with the respect to the TTL. However, the request has the field `stages` to restrict t stage of Computations to delete. The caller is the cronjob ComputationCleaner, which only request to delete COMPLETE  Computations. 
The FinishComputation here is to handle purge requests for unfinished computations. However, it is never executed in our system yet. So I think it is fine.
</blockquote></details>

My point still stands that the same Computation should never be finished and deleted in the same RPC. You should have to make two separate RPCs: one to finish and one to delete.

renjiezh (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1136)* status: all files reviewed, 4 unresolved discussions (waiting on @SanjayVas and @YuhongWang-Amazon)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/PostgresComputationsService.kt` line 273 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1136#-NaTaBtb9vLLZVcX54m2:-NaTh-4GCqFNydlXSr5Q:b-fztfqh) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/313ffcdd9d1e4b28a41796d7e377781cda99cf3e/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/PostgresComputationsService.kt#L273)):*

My point still stands that the same Computation should never be finished and deleted in the same RPC. You should have to make two separate RPCs: one to finish and one to delete.
</blockquote></details>

Suppose the scenario that, a Computation never gets the requisitions fulfilled (something wrong from the EDP side). Thus the Computation will be PENDING_REQUISITION forever. Currently we don't have any method to dealing with this case. One solution is to rely on the cronjob to force clean very old Computations(not enabled yet). This is why I keep the feature here.
Could you let me know your consideration why computations cannot be finished and deleted in the same RPC?

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1136)* status: all files reviewed, 4 unresolved discussions (waiting on @renjiezh and @YuhongWang-Amazon)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/PostgresComputationsService.kt` line 273 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1136#-NaTaBtb9vLLZVcX54m2:-NaTjhDv3svKizca73HN:bfk6jc2) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/313ffcdd9d1e4b28a41796d7e377781cda99cf3e/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/PostgresComputationsService.kt#L273)):*
The idea is to allow time for processing changes before things get deleted. We follow this pattern in the Kingdom: Measurements are either transitioned from  pending->cancelled or from cancelled->deleted transitions, with different TTLs for each operation.

For Duchy computations, isn't the normal flow that the Measurement gets marked as completed (SUCCEEDED or FAILED) at the Kingdom and then the Herald picks that up to delete it?

> Suppose the scenario that, a Computation never gets the requisitions fulfilled (something wrong from the EDP side). Thus the Computation will be PENDING_REQUISITION forever.

In practice, this case is handled at the Kingdom (it has a cron job for cancelling Measurements that have been pending too long).

renjiezh (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1136)* status: all files reviewed, 4 unresolved discussions (waiting on @YuhongWang-Amazon)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/PostgresComputationsService.kt` line 273 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1136#-NaTaBtb9vLLZVcX54m2:-NaTlXpL-wAdf2IjvF12:b-gh3nz8) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/313ffcdd9d1e4b28a41796d7e377781cda99cf3e/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/PostgresComputationsService.kt#L273)):*

The idea is to allow time for processing changes before things get deleted. We follow this pattern in the Kingdom: Measurements are either transitioned from  pending->cancelled or from cancelled->deleted transitions, with different TTLs for each operation.

For Duchy computations, isn't the normal flow that the Measurement gets marked as completed (SUCCEEDED or FAILED) at the Kingdom and then the Herald picks that up to delete it?

> Suppose the scenario that, a Computation never gets the requisitions fulfilled (something wrong from the EDP side). Thus the Computation will be PENDING_REQUISITION forever.

In practice, this case is handled at the Kingdom (it has a cron job for cancelling Measurements that have been pending too long).
</blockquote></details>

Sounds good. Then we change the purge function to allow only terminal (Complete, Failed) Computation to be deleted. 
@YuhongWang-Amazon Could you please make the change in the PR so that it ignore non-terminated stages in `request.stages`.

YuhongWang-Amazon (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1136)* status: all files reviewed, 4 unresolved discussions (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/PostgresComputationsService.kt` line 273 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1136#-NaTaBtb9vLLZVcX54m2:-NaTqMkT-p4iFrxqb48s:bxdw537) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/313ffcdd9d1e4b28a41796d7e377781cda99cf3e/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/PostgresComputationsService.kt#L273)):*

Sounds good. Then we change the purge function to allow only terminal (Complete, Failed) Computation to be deleted. 
@YuhongWang-Amazon Could you please make the change in the PR so that it ignore non-terminated stages in `request.stages`. 
</blockquote></details>

Sure, I can make the update. Since this has the same logic as spanner implementation, I will also update that.

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1136)* status: 11 of 28 files reviewed, 4 unresolved discussions (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/PostgresComputationsService.kt` line 273 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1136#-NaTaBtb9vLLZVcX54m2:-NahMwrJ1Qn7qMuuEB2E:b-gfvvlf) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/313ffcdd9d1e4b28a41796d7e377781cda99cf3e/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/PostgresComputationsService.kt#L273)):*

Updated both postgres and spanner implementations to filter out the non-terminal stages from the requests.
I am wondering if it is better to failGrpc when there are non-terminal stages in the request.
</blockquote></details>

That is a good idea to failGrpc for that. Otherwise the caller might be confused that some of the Computations still exist.

YuhongWang-Amazon (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1136)* status: 10 of 28 files reviewed, 4 unresolved discussions (waiting on @renjiezh and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/PostgresComputationsService.kt` line 273 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1136#-NaTaBtb9vLLZVcX54m2:-NahVGl1DjIP9FxiyK0P:b-wm3ce1) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/313ffcdd9d1e4b28a41796d7e377781cda99cf3e/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/PostgresComputationsService.kt#L273)):*

That is a good idea to failGrpc for that. Otherwise the caller might be confused that some of the Computations still exist.
</blockquote></details>

Sounds good. Updated with test coverage.

SanjayVas (changes requested):
> Reviewed 5 of 21 files at r1, 11 of 17 files at r4, 3 of 4 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1136)* status: 24 of 28 files reviewed, 5 unresolved discussions (waiting on @renjiezh and @YuhongWang-Amazon)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/PostgresComputationsService.kt` line 227 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1136#-NaTZfyV5YdjI4kc5haE:-Nai0YYg8ohV3yFWVimb:b-yhpf1c) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/313ffcdd9d1e4b28a41796d7e377781cda99cf3e/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/PostgresComputationsService.kt#L227)):*

Im trying to be consistent with the spanner implementation. Should I remove the try catch block from both places? https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/main/src/main/kotlin/org/wfanet/measurement/duchy/service/internal/computations/ComputationsService.kt#L137-L152
</blockquote></details>

Yes. Basically if a StatusException can possibly be thrown by this then it means there's something wrong with the `Store` or `StorageClient` impl.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/PostgresComputationsService.kt` line 273 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1136#-NaTaBtb9vLLZVcX54m2:-Nai2mxx36mi6e9oX8vA:b865bj5) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/313ffcdd9d1e4b28a41796d7e377781cda99cf3e/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/PostgresComputationsService.kt#L273)):*

Sounds good. Updated with test coverage.
</blockquote></details>

I'm still seeing this "if not terminated then finish" logic? If we're throwing an error for non-terminal states then this can never be hit.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/readers/ComputationReader.kt` line 222 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1136#-Nai079XEyxrnHvlsmAv:-Nai079XEyxrnHvlsmAw:bm8pf2t) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e647b4821d2183c08aa20c279f9d8e3385cbacc1/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/readers/ComputationReader.kt#L222)):*
> ```Kotlin
> 
>   /**
>    * Reads a [ComputationToken] by globalComputationId.
> ```

nit: maybe make the difference between this and the other overload more clear?



_Suggestion:_
```Kotlin
Reads a [ComputationToken] by [globalComputationId] in a new transaction.
```
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/readers/ComputationReader.kt` line 377 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1136#-Nai-MVX3DzDhT3HPxMW:-Nai-MVX3DzDhT3HPxMX:bbvwaxg) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e647b4821d2183c08aa20c279f9d8e3385cbacc1/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/readers/ComputationReader.kt#L377)):*
> ```Kotlin
>   }
> 
>   suspend fun readLockOwner(readContext: ReadContext, computationId: Long): LockOwnerQueryResult? {
> ```

Document that this returns null when the *Computation* is not found (as opposed to the lock owner not being found as a result).

Alternatively if this is always an error case, throw a `ComputationNotFoundException`.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/writers/ClaimWork.kt` line 71 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1136#-NahzhL1BmaGbiAvXNj_:-NahzhL1BmaGbiAvXNja:b4v8dl9) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e647b4821d2183c08aa20c279f9d8e3385cbacc1/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/writers/ClaimWork.kt#L71)):*
> ```Kotlin
>       ?.let {
>         computationReader.readComputationToken(transactionContext, it.globalId)
>           ?: throw UnknownDataError()
> ```

Shouldn't this be a more specific `ComputationNotFoundException`? The calling code can decide what to do with that, e.g. whether it should result in a gRPC status with code INTERNAL.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/service/internal/DuchyInternalException.kt` line 105 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1136#-NahyxDQ7hlCXbcdqj9p:-NahyxDQ7hlCXbcdqj9q:b-xdb077) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e647b4821d2183c08aa20c279f9d8e3385cbacc1/src/main/kotlin/org/wfanet/measurement/duchy/service/internal/DuchyInternalException.kt#L105)):*
> ```Kotlin
> }
> 
> class UnknownDataError(
> ```

`Error` and `Exception` are different types in Java, and this is one of the latter.



_Suggestion:_
```Kotlin
UnknownDataException
```

YuhongWang-Amazon (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1136)* status: 24 of 28 files reviewed, 4 unresolved discussions (waiting on @renjiezh and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/PostgresComputationsService.kt` line 227 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1136#-NaTZfyV5YdjI4kc5haE:-Naj2Pus3_qDkfMXYYw4:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/313ffcdd9d1e4b28a41796d7e377781cda99cf3e/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/PostgresComputationsService.kt#L227)):*

Yes. Basically if a StatusException can possibly be thrown by this then it means there's something wrong with the `Store` or `StorageClient` impl.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/PostgresComputationsService.kt` line 273 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1136#-NaTaBtb9vLLZVcX54m2:-NajPkje6y_0zvnw0DFD:bss4a1y) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/313ffcdd9d1e4b28a41796d7e377781cda99cf3e/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/PostgresComputationsService.kt#L273)):*

I'm still seeing this "if not terminated then finish" logic? If we're throwing an error for non-terminal states then this can never be hit.
</blockquote></details>

Removed!
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/readers/ComputationReader.kt` line 377 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1136#-Nai-MVX3DzDhT3HPxMW:-NajE-5-DZQpZnhCjnUr:b-og2iy4) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e647b4821d2183c08aa20c279f9d8e3385cbacc1/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/readers/ComputationReader.kt#L377)):*

Document that this returns null when the *Computation* is not found (as opposed to the lock owner not being found as a result).

Alternatively if this is always an error case, throw a `ComputationNotFoundException`.
</blockquote></details>

Added the document.

This scenario is similar to the other places where I throws `UnknownDataError` (changed to `DataCorruptedException`).
Currently, `readLockOwner` is only called in ClaimWork writer, where the previous `listUnclaimedTasks` confirmed that the computation exists. Hence, I returned `null` here and let the caller to throw `UnknownDataError`.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/writers/ClaimWork.kt` line 71 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1136#-NahzhL1BmaGbiAvXNj_:-Naj4zXc6gPedUwW4HgU:b-kdbq3) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e647b4821d2183c08aa20c279f9d8e3385cbacc1/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/writers/ClaimWork.kt#L71)):*

Shouldn't this be a more specific `ComputationNotFoundException`? The calling code can decide what to do with that, e.g. whether it should result in a gRPC status with code INTERNAL.
</blockquote></details>

I defined a new `UnknownDataError` here to differentiate from the `ComputationNotFoundException`.
In this scenario, the previous code just modified the same computation within the same transaction, hence the reader is not expected to get a null, unless the database is corrupted or something.
On the other hand, the `ComputationNotFoundException` is expected to be raised when a transaction could not find the target computation in the first place.

We could actually replace this exception with `!!` if we don't expect this scenario to happen (at least it should be very rare). Do you have any opinion which approach is better?
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/service/internal/DuchyInternalException.kt` line 105 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1136#-NahyxDQ7hlCXbcdqj9p:-NajI_0O10YEUpnb55Ke:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e647b4821d2183c08aa20c279f9d8e3385cbacc1/src/main/kotlin/org/wfanet/measurement/duchy/service/internal/DuchyInternalException.kt#L105)):*

`Error` and `Exception` are different types in Java, and this is one of the latter.
</blockquote></details>

Done.

SanjayVas (changes requested):
> Reviewed 1 of 21 files at r1, 13 of 13 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1136)* status: all files reviewed, 2 unresolved discussions (waiting on @renjiezh and @YuhongWang-Amazon)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/PostgresComputationsService.kt` line 266 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1136#-NamvJDP-ACMSN1ADP1H:-NamvJDQ9D5oH0sgT-uu:b-xe5dup) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d19b2cd46a5e6cbc991fd09a407b0a2f41f9fe9a/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/PostgresComputationsService.kt#L266)):*
> ```Kotlin
>       for (globalId in globalIds) {
>         val token = computationReader.readComputationToken(client, globalId) ?: continue
>         DeleteComputation(token.localComputationId).execute(client, idGenerator)
> ```

Each service method should be a single transaction.



_Code quote:_
```Kotlin
        val token = computationReader.readComputationToken(client, globalId) ?: continue
        DeleteComputation(token.localComputationId).execute(client, idGenerator)
```
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/writers/ClaimWork.kt` line 71 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1136#-NahzhL1BmaGbiAvXNj_:-NamvoeS9gl2boa-uTbF:b-lue6wr) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e647b4821d2183c08aa20c279f9d8e3385cbacc1/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/writers/ClaimWork.kt#L71)):*
> I defined a new `UnknownDataError` here to differentiate from the `ComputationNotFoundException`.

I don't see a need for this.

> In this scenario, the previous code just modified the same computation within the same transaction, hence the reader is not expected to get a null, 

If it's violating an invariant, throw an `IllegalStateException` (there's a very convenient `checkNotNull` function for this). No need for a special exception type.

> unless the database is corrupted or something.

More likely to be "or something", as Postgres should be guaranteeing this for us. That is to say that we don't know if the cause is data corruption. All we know is that something that should be impossible happened.

YuhongWang-Amazon (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1136)* status: 20 of 31 files reviewed, 2 unresolved discussions (waiting on @renjiezh and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/PostgresComputationsService.kt` line 266 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1136#-NamvJDP-ACMSN1ADP1H:-Naqp8Sv4O4YrwCytKhx:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d19b2cd46a5e6cbc991fd09a407b0a2f41f9fe9a/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/PostgresComputationsService.kt#L266)):*

Each service method should be a single transaction.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/writers/ClaimWork.kt` line 71 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1136#-NahzhL1BmaGbiAvXNj_:-Naqp9ki4ouFd5uy8uGK:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e647b4821d2183c08aa20c279f9d8e3385cbacc1/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/writers/ClaimWork.kt#L71)):*

> I defined a new `UnknownDataError` here to differentiate from the `ComputationNotFoundException`.

I don't see a need for this.

> In this scenario, the previous code just modified the same computation within the same transaction, hence the reader is not expected to get a null, 

If it's violating an invariant, throw an `IllegalStateException` (there's a very convenient `checkNotNull` function for this). No need for a special exception type.

> unless the database is corrupted or something.

More likely to be "or something", as Postgres should be guaranteeing this for us. That is to say that we don't know if the cause is data corruption. All we know is that something that should be impossible happened.
</blockquote></details>

Done.

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1136)* status: 18 of 31 files reviewed, 2 unresolved discussions (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/service/internal/computations/ComputationsService.kt` line 372 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1136#-Nar1pFQBMWcRPlaEqh2:-NarCd-xEWCRoKXGO4Ce:b98o08i) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/15f5362f6b74dd450605e9364c885408e68f1061/src/main/kotlin/org/wfanet/measurement/duchy/service/internal/computations/ComputationsService.kt#L372)):*

@renjiezh it seems that these two methods are not used, could you help confirm whether we can remove them or not?
</blockquote></details>

You're right. It is safe to remove both.

SanjayVas (changes requested):
> Reviewed 9 of 11 files at r7, 3 of 4 files at r8, 1 of 1 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1136)* status: all files reviewed, 2 unresolved discussions (waiting on @YuhongWang-Amazon)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/PostgresComputationsService.kt` line 266 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1136#-NamvJDP-ACMSN1ADP1H:-NarYyD_5UvoJUKIU0o-:b-qp8suu) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d19b2cd46a5e6cbc991fd09a407b0a2f41f9fe9a/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/PostgresComputationsService.kt#L266)):*

Done.
</blockquote></details>

The read of global IDs is still happening in a different transaction than the deletion here. Perhaps move that read into DeleteComputation and pass the `force` parameter to it, having it unconditionally return the sample.
___
*[`src/main/proto/wfa/measurement/internal/duchy/error_code.proto` line 44 at r9](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1136#-NarZfCX5Nz3aH0s5wM0:-NarZfCX5Nz3aH0s5wM1:b-l6hdjr) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3b74c02322307381c5159a6057a4bd86b0a57da3/src/main/proto/wfa/measurement/internal/duchy/error_code.proto#L44)):*
> ```proto
> 
>   /** Data corrupted for unknown reasons. */
>   DATA_CORRUPTED = 7;
> ```

revert

YuhongWang-Amazon (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1136)* status: all files reviewed, 2 unresolved discussions (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/PostgresComputationsService.kt` line 266 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1136#-NamvJDP-ACMSN1ADP1H:-Nar_OQ_8a3tJyRTzSuj:bdxqpnc) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d19b2cd46a5e6cbc991fd09a407b0a2f41f9fe9a/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/PostgresComputationsService.kt#L266)):*

The read of global IDs is still happening in a different transaction than the deletion here. Perhaps move that read into DeleteComputation and pass the `force` parameter to it, having it unconditionally return the sample.
</blockquote></details>

I didn't make this a single transaction because I was thinking about the partial purges. E.g 10 computations should be purged, but somehow it failed at deleting the 6th one. In this implementation, the first 5 computations will still be deleted. But in a transactional implementation, it will be either all 10 are deleted or non.

Since the exceptions are caught and logged, the returned `purgeCount` could also reflect the actually deleted rows. I am wondering if there is the need to make this one transaction.

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1136)* status: all files reviewed, 2 unresolved discussions (waiting on @YuhongWang-Amazon)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/PostgresComputationsService.kt` line 266 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1136#-NamvJDP-ACMSN1ADP1H:-Narxjzt454p9qtforVX:b-tzn0tx) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d19b2cd46a5e6cbc991fd09a407b0a2f41f9fe9a/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/PostgresComputationsService.kt#L266)):*
> in a transactional implementation, it will be either all 10 are deleted or non.

That would be the preferred behavior. It's how we handle similar batch deletions in the CMMS public API.

YuhongWang-Amazon (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1136)* status: 28 of 32 files reviewed, 2 unresolved discussions (waiting on @renjiezh and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/PostgresComputationsService.kt` line 266 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1136#-NamvJDP-ACMSN1ADP1H:-NasZBirEmz7bqWdCyth:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d19b2cd46a5e6cbc991fd09a407b0a2f41f9fe9a/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/PostgresComputationsService.kt#L266)):*

> in a transactional implementation, it will be either all 10 are deleted or non.

That would be the preferred behavior. It's how we handle similar batch deletions in the CMMS public API.
</blockquote></details>

Done.
___
*[`src/main/proto/wfa/measurement/internal/duchy/error_code.proto` line 44 at r9](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1136#-NarZfCX5Nz3aH0s5wM0:-NasZDns9JqyMJuL09SM:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3b74c02322307381c5159a6057a4bd86b0a57da3/src/main/proto/wfa/measurement/internal/duchy/error_code.proto#L44)):*

revert
</blockquote></details>

Done.

SanjayVas (approved):
> Reviewed 4 of 4 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1136)* status: all files reviewed, 1 unresolved discussion (waiting on @YuhongWang-Amazon)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/PostgresComputationsService.kt` line 260 at r10](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1136#-Naw_0bP0tHJG1UJ74fI:-Naw_0bP0tHJG1UJ74fJ:bgad3uw) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ccf5922a52a3865d9b146a21bb9d4c341ae44d38/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/PostgresComputationsService.kt#L260)):*
> ```Kotlin
>     return purgeComputationsResponse {
>       purgeCount = purgeResult.purgeCount
>       purgeResult.purgeSamples?.forEach { purgeSample += it }
> ```

nit: Use a regular conditional, relying on smart casts (the `+=` operator works for both "add" and "add all")

Alternatively, just have purgeResult.purgeSamples be the empty set (a `repeated` field with no elements is equivalent to one that hasn't been set).



_Suggestion:_
```Kotlin
if (purgeResult.purgeSamples != null) {
  purgeSamples += purgeResult.purgeSamples
}
```

YuhongWang-Amazon (commented):
> Reviewed 2 of 2 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1136)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @YuhongWang-Amazon)

PR #1143, opened 2023-07-31
Bdomen react api utils

bdomen-ggl:
Adding the UI models and API code.
Notes:

- Not hooking up to the backend yet (BFF not yet implemented) so using a Fake Api.
- Can easily switch to the Real Api (and finish implementing it) when the BFF server is ready.
- Models are written in TypeScript to help with type safety in the UI. They are at the moment arbitrarily made to make the UI easier to render. The idea would be the BFF would emit a similar patter, but still experimenting with this.
- Starting to add some things like memoization. Not needed immediately with fake data, but will be needed when hooking up to the BFF. Can still use some enhancements before then.


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1143)

bdomen-ggl (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1143)* status: 0 of 16 files reviewed, 5 unresolved discussions (waiting on @bdomen-ggl and @SanjayVas)

___
*[`experimental/reporting-ui/src/main/react/reporting-ui/library/utils/api/memoize.ts` line 18 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1143#-NahQiIy6pwBoq4NGw4s:-NahQiIy6pwBoq4NGw4t:b-b1zssl) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/320575ec11bce12e6ea212238c1c4b73b63d8456/experimental/reporting-ui/src/main/react/reporting-ui/library/utils/api/memoize.ts#L18)):*
> ```XML
> export const memoizePromiseFn = fn => {
>   return (...args) => {
>     const key = JSON.stringify(args);
> ```

May want to add the function name or some other base key in addition to the args later.
___
*[`experimental/reporting-ui/src/main/react/reporting-ui/library/utils/api/realApi.ts` line 24 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1143#-NahOpP3AUMd2c_NZ9uR:-NahOpP3AUMd2c_NZ9uS:b-tvjzy6) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/320575ec11bce12e6ea212238c1c4b73b63d8456/experimental/reporting-ui/src/main/react/reporting-ui/library/utils/api/realApi.ts#L24)):*
> ```XML
> import path from 'path';
> 
> export class RealApi implements IReportingApi {
> ```

Just stubbing this for now until the BFF is ready for use.
___
*[`experimental/reporting-ui/src/main/react/reporting-ui/library/utils/api/reportingApi.ts` line 25 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1143#-NahOxTYDgjtdn6toY4x:-NahOxTYDgjtdn6toY4y:b-chgen8) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/320575ec11bce12e6ea212238c1c4b73b63d8456/experimental/reporting-ui/src/main/react/reporting-ui/library/utils/api/reportingApi.ts#L25)):*
> ```XML
> import {IReportingApi} from './IReportingApi';
> 
> export class ReportingApi {
> ```

This will ultimately be the class to be called. It will handle common tasks like memoization (below) and logging (future) so they don't have to be reimplemented with every implementation.

Alternatively this can be removed/replaced with the real version as we won't realistically have multiple implementations and can remove a layer of abstraction.
___
*[`experimental/reporting-ui/src/main/react/reporting-ui/library/utils/api/reportingApi.ts` line 39 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1143#-NahPneg04-qmNFGl90c:-NahPneg04-qmNFGl90d:b7m5khz) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/320575ec11bce12e6ea212238c1c4b73b63d8456/experimental/reporting-ui/src/main/react/reporting-ui/library/utils/api/reportingApi.ts#L39)):*
> ```XML
> 
>   getReport(req: GetReportRequest): Promise<GetReportResponse> {
>     const getCached = memoizePromiseFn(this.api.getReport);
> ```

Reports are immutable. If we ask for the same one, we'll always get the same result. This will be useful in the UI when multiple widgets will want to ultimately call Get Report. This will ensure the request is only made once over the wire.

We can add memoization to other methods like List Reports, but the results can change with the same request and adding a refresh feature is excessive at the moment. May add memoization with a refresh button in the future.
___
*[`experimental/reporting-ui/src/main/react/reporting-ui/library/utils/api/models/InitApiProps.ts` line 16 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1143#-NahPZ2d-45m1ApGFWdT:-NahPZ2d-45m1ApGFWdU:b-q4wuam) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/320575ec11bce12e6ea212238c1c4b73b63d8456/experimental/reporting-ui/src/main/react/reporting-ui/library/utils/api/models/InitApiProps.ts#L16)):*
> ```XML
> 
> export type InitApiProps = {
>   // eslint-disable-next-line node/no-unsupported-features/node-builtins
> ```

Working on updating the eslint configuration to allow this as it's supported in the environment we're using. Not high priority right now.

SanjayVas (changes requested):
> Reviewed 6 of 16 files at r1, 4 of 4 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1143)* status: 10 of 16 files reviewed, 13 unresolved discussions (waiting on @bdomen-ggl)

___
*[`experimental/reporting-ui/src/main/react/reporting-ui/library/utils/api/fakeApi.ts` line 1 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1143#-Nai8Ngm0-R4ijoWbBZJ:-Nai8Ngm0-R4ijoWbBZK:b-d4guiz) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/320575ec11bce12e6ea212238c1c4b73b63d8456/experimental/reporting-ui/src/main/react/reporting-ui/library/utils/api/fakeApi.ts#L1)):*
> ```XML
> ```

Fakes are test infrastructure, so they should go under a `testing` Bazel package with `default_testonly = True` to ensure that they can't be used from non-test code.
___
*[`experimental/reporting-ui/src/main/react/reporting-ui/library/utils/api/fakeApi.ts` line 1 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1143#-Nai8nGk0M7c1vxvS0F1:-Nai8nGl5A2Ctg4th6Ak:b-5s2l19) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/320575ec11bce12e6ea212238c1c4b73b63d8456/experimental/reporting-ui/src/main/react/reporting-ui/library/utils/api/fakeApi.ts#L1)):*
> ```XML
> ```

I think file names should be using `snake_case` for consistent style.
___
*[`experimental/reporting-ui/src/main/react/reporting-ui/library/utils/api/fakeApi.ts` line 305 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1143#-NaiEQjc844BgOWteuke:-NaiEQjc844BgOWteukf:b-wowsqm) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/320575ec11bce12e6ea212238c1c4b73b63d8456/experimental/reporting-ui/src/main/react/reporting-ui/library/utils/api/fakeApi.ts#L305)):*
> ```XML
>   async listReports(): Promise<ListReportsResponse> {
>     return new Promise(resolve =>
>       resolve({
> ```

[Freeze](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/freeze) the response object (anything representing a service response should be immutable).
___
*[`experimental/reporting-ui/src/main/react/reporting-ui/library/utils/api/fakeApi.ts` line 308 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1143#-NaiB3_j6THPpN0D5Vn9:-NaiB3_j6THPpN0D5VnA:b-hujwsu) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/320575ec11bce12e6ea212238c1c4b73b63d8456/experimental/reporting-ui/src/main/react/reporting-ui/library/utils/api/fakeApi.ts#L308)):*
> ```XML
>         reports: this.reports,
>       })
>     );
> ```

Assuming this is a standard ECMAScript Promise, then you should be able to just use `Promise.resolve()` to build an immediately-resolved Promise.



_Suggestion:_
```XML
    return Promise.resolve({reports: this.reports});
```
___
*[`experimental/reporting-ui/src/main/react/reporting-ui/library/utils/api/memoize.ts` line 15 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1143#-NaiCw0K15QZVi3OTH-o:-NaiCw0K15QZVi3OTH-p:bdtp519) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/320575ec11bce12e6ea212238c1c4b73b63d8456/experimental/reporting-ui/src/main/react/reporting-ui/library/utils/api/memoize.ts#L15)):*
> ```XML
> 
> const cache = new Map();
> ```

I don't think I like having a single global cache for everything. Each cache should have its properties, e.g. max size and expiration.
___
*[`experimental/reporting-ui/src/main/react/reporting-ui/library/utils/api/realApi.ts` line 35 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1143#-NaiDBazB95fs3DIDAzj:-NaiDBazB95fs3DIDAzk:b-4pikg3) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/320575ec11bce12e6ea212238c1c4b73b63d8456/experimental/reporting-ui/src/main/react/reporting-ui/library/utils/api/realApi.ts#L35)):*
> ```XML
> 
>   async listReports(): Promise<ListReportsResponse> {
>     return fetch(path.join(this.baseUrl.toString(), '/api/reports'))
> ```

The main benefit of using `async` functions is so you can use `await` within them rather than having to chain `then`s.
___
*[`experimental/reporting-ui/src/main/react/reporting-ui/library/utils/api/reportingApi.ts` line 39 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1143#-NahPneg04-qmNFGl90c:-NaiC9HS2odg5q97pQTO:bxo31es) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/320575ec11bce12e6ea212238c1c4b73b63d8456/experimental/reporting-ui/src/main/react/reporting-ui/library/utils/api/reportingApi.ts#L39)):*
> Reports are immutable. If we ask for the same one, we'll always get the same result

If you're talking about the resource from the Reporting API, this is not true. Reports have state. A Report only stops being mutable once it reaches a terminal state.
___
*[`experimental/reporting-ui/src/main/react/reporting-ui/library/utils/api/models/Demo.ts` line 15 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1143#-NaiDV1d3XpoyFppoCt5:-NaiDV1d3XpoyFppoCt6:b-yz4gp8) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/320575ec11bce12e6ea212238c1c4b73b63d8456/experimental/reporting-ui/src/main/react/reporting-ui/library/utils/api/models/Demo.ts#L15)):*
> ```XML
> 
> export type Demo = {
> ```

nit: Be more verbose. Is this supposed to be a demographic?



_Code quote:_
```XML
Demo
```
___
*[`experimental/reporting-ui/src/main/react/reporting-ui/library/utils/api/IReportingApi.ts` line 22 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1143#-Nai9HJR868rrUecRvb8:-Nai9HJR868rrUecRvb9:b-5co03f) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/320575ec11bce12e6ea212238c1c4b73b63d8456/experimental/reporting-ui/src/main/react/reporting-ui/library/utils/api/IReportingApi.ts#L22)):*
> ```XML
> } from './models';
> 
> export interface IReportingApi {
> ```

Are we following https://google.github.io/styleguide/tsguide.html? If so, then this shouldn't have an `I` prefix (unless that's particularly idiomatic to React TS). See https://google.github.io/styleguide/tsguide.html#naming-style

bdomen-ggl (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1143)* status: 10 of 16 files reviewed, 13 unresolved discussions (waiting on @SanjayVas)

___
*[`experimental/reporting-ui/src/main/react/reporting-ui/library/utils/api/fakeApi.ts` line 1 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1143#-Nai8Ngm0-R4ijoWbBZJ:-NalKtJb8pCMRN5e8n-H:b3uuyfs) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/320575ec11bce12e6ea212238c1c4b73b63d8456/experimental/reporting-ui/src/main/react/reporting-ui/library/utils/api/fakeApi.ts#L1)):*

Fakes are test infrastructure, so they should go under a `testing` Bazel package with `default_testonly = True` to ensure that they can't be used from non-test code.
</blockquote></details>

Sure. This is intended to be used so I can build out more of the UI service code while still being able to render data in the UI without hardcoding it since we don't have a backend server to hit yet.

Would you suggest I make "real" API calls instead (despite the backend doesn't exist yet) and visualize my views in another tool like Storybook? https://storybook.js.org/ I opted against this at first because we're trying to push out some UI code and setting up Storybook would be another piece that would take a while to add. Also, the nice thing about the fake data in the app vs something like Story Book is that I can go through user workflows (granted we only have two views) and observe how API calls might be optimized.

I could build out the backend server first but was asked to push a UI with fake data so I started with a Fake Api. We can revisit this discussion if you would prefer this route.

I could also just leave the Fake Api local only (not check it in), but would be worried I might accidentally delete it and have to re-write it.
___
*[`experimental/reporting-ui/src/main/react/reporting-ui/library/utils/api/fakeApi.ts` line 1 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1143#-Nai8nGk0M7c1vxvS0F1:-NalMu1X7eonhh1TRhkE:b-4tlm4u) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/320575ec11bce12e6ea212238c1c4b73b63d8456/experimental/reporting-ui/src/main/react/reporting-ui/library/utils/api/fakeApi.ts#L1)):*

I think file names should be using `snake_case` for consistent style.
</blockquote></details>

I can do that. Everywhere I've been at before used camel case for folders and pascal case for files in javascript/typescript projects so I just started with that. The idea being that the name of classes tend to be the same as the file name (some linting rules I've used enforced this).
___
*[`experimental/reporting-ui/src/main/react/reporting-ui/library/utils/api/fakeApi.ts` line 305 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1143#-NaiEQjc844BgOWteuke:-NalNvaKDTe-LrtrmA6k:b-ixxo6m) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/320575ec11bce12e6ea212238c1c4b73b63d8456/experimental/reporting-ui/src/main/react/reporting-ui/library/utils/api/fakeApi.ts#L305)):*

[Freeze](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/freeze) the response object (anything representing a service response should be immutable).
</blockquote></details>

I can do this, but freeze is only a shallow freeze and there's lots of nested objects.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/freeze

Are you also suggesting I recursively go through every property that is an object and freeze them before assigning? Or is the shallow freeze good enough?
___
*[`experimental/reporting-ui/src/main/react/reporting-ui/library/utils/api/fakeApi.ts` line 308 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1143#-NaiB3_j6THPpN0D5Vn9:-NalPn0r5nVkt73CZH-2:b-8o2acm) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/320575ec11bce12e6ea212238c1c4b73b63d8456/experimental/reporting-ui/src/main/react/reporting-ui/library/utils/api/fakeApi.ts#L308)):*

Assuming this is a standard ECMAScript Promise, then you should be able to just use `Promise.resolve()` to build an immediately-resolved Promise.
</blockquote></details>

Yes I can. I was initially doing it this way since I was expecting there to be some more logic. I simplified some assumptions so that's no longer the case.
___
*[`experimental/reporting-ui/src/main/react/reporting-ui/library/utils/api/memoize.ts` line 15 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1143#-NaiCw0K15QZVi3OTH-o:-NalRcbD7p-O-Yx7vQP3:b-y39xbf) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/320575ec11bce12e6ea212238c1c4b73b63d8456/experimental/reporting-ui/src/main/react/reporting-ui/library/utils/api/memoize.ts#L15)):*

I don't think I like having a single global cache for everything. Each cache should have its properties, e.g. max size and expiration.
</blockquote></details>

That will probably be true. There's only one use with Get Report right now. I was going to wait until we needed to use this more to see what the requirements will be before adding more complexity. It's technically not even a cache (in the sense of a fully featured cache) since it's just a map.

I can change it to a cache and make it specific to the caller to fine tune caching policies. I'll likely hardcode the policies for now and add an app config in a future PR. I'll need other things anyways like Reporting BFF API URL.
___
*[`experimental/reporting-ui/src/main/react/reporting-ui/library/utils/api/realApi.ts` line 35 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1143#-NaiDBazB95fs3DIDAzj:-NalQpki5QEOASxlYaKz:bboiqp5) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/320575ec11bce12e6ea212238c1c4b73b63d8456/experimental/reporting-ui/src/main/react/reporting-ui/library/utils/api/realApi.ts#L35)):*

The main benefit of using `async` functions is so you can use `await` within them rather than having to chain `then`s.
</blockquote></details>

Yes. Either async/await or just using thens is fine. I copied fetch from somewhere and didn't update the thens. There's some subtle differences between the two but I don't think they really matter in this case.
___
*[`experimental/reporting-ui/src/main/react/reporting-ui/library/utils/api/reportingApi.ts` line 39 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1143#-NahPneg04-qmNFGl90c:-NalRE0B5N_AP-gPwYQa:b-cong1l) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/320575ec11bce12e6ea212238c1c4b73b63d8456/experimental/reporting-ui/src/main/react/reporting-ui/library/utils/api/reportingApi.ts#L39)):*

> Reports are immutable. If we ask for the same one, we'll always get the same result

If you're talking about the resource from the Reporting API, this is not true. Reports have state. A Report only stops being mutable once it reaches a terminal state.
</blockquote></details>

Ah yes that's technically right. The non-terminal state will likely lend itself to another view since all the charts and graphs won't have anything. I'll take a look at this case.
___
*[`experimental/reporting-ui/src/main/react/reporting-ui/library/utils/api/models/Demo.ts` line 15 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1143#-NaiDV1d3XpoyFppoCt5:-NalRX9-BA8MTjqzZY0m:b9fsxkx) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/320575ec11bce12e6ea212238c1c4b73b63d8456/experimental/reporting-ui/src/main/react/reporting-ui/library/utils/api/models/Demo.ts#L15)):*

nit: Be more verbose. Is this supposed to be a demographic?
</blockquote></details>

Yes, I can change the name.
___
*[`experimental/reporting-ui/src/main/react/reporting-ui/library/utils/api/IReportingApi.ts` line 22 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1143#-Nai9HJR868rrUecRvb8:-NalQ1sJ55FdRJL8xSMP:beao4cn) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/320575ec11bce12e6ea212238c1c4b73b63d8456/experimental/reporting-ui/src/main/react/reporting-ui/library/utils/api/IReportingApi.ts#L22)):*

Are we following https://google.github.io/styleguide/tsguide.html? If so, then this shouldn't have an `I` prefix (unless that's particularly idiomatic to React TS). See https://google.github.io/styleguide/tsguide.html#naming-style
</blockquote></details>

I threw this up last second and copied an existing example I saw. I can switch it to the style guide you linked.

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1143)* status: 10 of 16 files reviewed, 13 unresolved discussions (waiting on @bdomen-ggl)

___
*[`experimental/reporting-ui/src/main/react/reporting-ui/library/utils/api/fakeApi.ts` line 1 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1143#-Nai8Ngm0-R4ijoWbBZJ:-Nam-7cO1S9bal3nipJM:b-v2nw29) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/320575ec11bce12e6ea212238c1c4b73b63d8456/experimental/reporting-ui/src/main/react/reporting-ui/library/utils/api/fakeApi.ts#L1)):*

Sure. This is intended to be used so I can build out more of the UI service code while still being able to render data in the UI without hardcoding it since we don't have a backend server to hit yet.

Would you suggest I make "real" API calls instead (despite the backend doesn't exist yet) and visualize my views in another tool like Storybook? https://storybook.js.org/ I opted against this at first because we're trying to push out some UI code and setting up Storybook would be another piece that would take a while to add. Also, the nice thing about the fake data in the app vs something like Story Book is that I can go through user workflows (granted we only have two views) and observe how API calls might be optimized.

I could build out the backend server first but was asked to push a UI with fake data so I started with a Fake Api. We can revisit this discussion if you would prefer this route.

I could also just leave the Fake Api local only (not check it in), but would be worried I might accidentally delete it and have to re-write it.
</blockquote></details>

I'm not saying the fake is a bad idea. You can still use it, just make sure it's in the right place.

The way we do this in server-side code is to have all service clients/stubs be provided to the top-level constructor. This way something at the top can decide whether to use the fake. I don't know what the top level is for a React app (is there some App type?) but you'd have a testing version of that which also goes under a `testing` Bazel package.
___
*[`experimental/reporting-ui/src/main/react/reporting-ui/library/utils/api/fakeApi.ts` line 1 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1143#-Nai8nGk0M7c1vxvS0F1:-Nam-YxD4ZHGDkcg_E1s:bechgeb) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/320575ec11bce12e6ea212238c1c4b73b63d8456/experimental/reporting-ui/src/main/react/reporting-ui/library/utils/api/fakeApi.ts#L1)):*

I can do that. Everywhere I've been at before used camel case for folders and pascal case for files in javascript/typescript projects so I just started with that. The idea being that the name of classes tend to be the same as the file name (some linting rules I've used enforced this).
</blockquote></details>

This is coming from the [Google JS style guide](https://google.github.io/styleguide/jsguide.html#file-name). It doesn't appear to be mentioned in the TS style guide, but I assume falling back to the JS style guide makes sense when the TS one doesn't specify.

If we're looking for other style guides, it looks like React is relatively unopinionated on the matter and there isn't a popular style guide that covers this. FWIW the Angular style guide uses lowercase with hyphens, so either way I don't think lowerCamelCase is the right choice.
___
*[`experimental/reporting-ui/src/main/react/reporting-ui/library/utils/api/fakeApi.ts` line 305 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1143#-NaiEQjc844BgOWteuke:-Nam1mHZAm1Fng6F1IWQ:b7c1fm0) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/320575ec11bce12e6ea212238c1c4b73b63d8456/experimental/reporting-ui/src/main/react/reporting-ui/library/utils/api/fakeApi.ts#L305)):*

I can do this, but freeze is only a shallow freeze and there's lots of nested objects.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/freeze

Are you also suggesting I recursively go through every property that is an object and freeze them before assigning? Or is the shallow freeze good enough?
</blockquote></details>

Shallow is fine. It's mostly about intent, since JS can't really stop you from doing most things.
___
*[`experimental/reporting-ui/src/main/react/reporting-ui/library/utils/api/memoize.ts` line 15 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1143#-NaiCw0K15QZVi3OTH-o:-NamA9otCokOqEb7vyAT:b-yb85ht) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/320575ec11bce12e6ea212238c1c4b73b63d8456/experimental/reporting-ui/src/main/react/reporting-ui/library/utils/api/memoize.ts#L15)):*

That will probably be true. There's only one use with Get Report right now. I was going to wait until we needed to use this more to see what the requirements will be before adding more complexity. It's technically not even a cache (in the sense of a fully featured cache) since it's just a map.

I can change it to a cache and make it specific to the caller to fine tune caching policies. I'll likely hardcode the policies for now and add an app config in a future PR. I'll need other things anyways like Reporting BFF API URL.
</blockquote></details>

A map is a fine implementation of a cache (not every cache needs to handle all the policies/features internally). I also know that you were trying to share caching across the real and fake service clients, but I don't think that's necessary.
___
*[`experimental/reporting-ui/src/main/react/reporting-ui/library/utils/api/IReportingApi.ts` line 22 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1143#-Nai9HJR868rrUecRvb8:-NamBakA2qyaQU8RaBPa:btjj1xa) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/320575ec11bce12e6ea212238c1c4b73b63d8456/experimental/reporting-ui/src/main/react/reporting-ui/library/utils/api/IReportingApi.ts#L22)):*

I threw this up last second and copied an existing example I saw. I can switch it to the style guide you linked.
</blockquote></details>

Maybe also call it `ReportingClient` instead, since it's really an API client not an API. I also don't think it makes sense to be in a `utils` package, as generally utils are free functions.

bdomen-ggl:
*[`experimental/reporting-ui/src/main/react/reporting-ui/library/utils/api/models/Demo.ts` line 15 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1143#-NaiDV1d3XpoyFppoCt5:-NamCEbGCL2oOEcsJe5a:bb6un1j) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/320575ec11bce12e6ea212238c1c4b73b63d8456/experimental/reporting-ui/src/main/react/reporting-ui/library/utils/api/models/Demo.ts#L15)):*

Yes, I can change the name.
</blockquote></details>

Done

bdomen-ggl:
*[`experimental/reporting-ui/src/main/react/reporting-ui/library/utils/api/reportingApi.ts` line 39 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1143#-NahPneg04-qmNFGl90c:-NamCH8vD0EfD4jOXQSK:bb6un1j) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/320575ec11bce12e6ea212238c1c4b73b63d8456/experimental/reporting-ui/src/main/react/reporting-ui/library/utils/api/reportingApi.ts#L39)):*

Ah yes that's technically right. The non-terminal state will likely lend itself to another view since all the charts and graphs won't have anything. I'll take a look at this case.
</blockquote></details>

Done

bdomen-ggl:
*[`experimental/reporting-ui/src/main/react/reporting-ui/library/utils/api/realApi.ts` line 35 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1143#-NaiDBazB95fs3DIDAzj:-Nal__E5-_X9c8GlxFSJ:b-5gox7a) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/320575ec11bce12e6ea212238c1c4b73b63d8456/experimental/reporting-ui/src/main/react/reporting-ui/library/utils/api/realApi.ts#L35)):*

Yes. Either async/await or just using thens is fine. I copied fetch from somewhere and didn't update the thens. There's some subtle differences between the two but I don't think they really matter in this case.
</blockquote></details>

Done
Removed async since this is only doing a fetch and I want to return a promise anyways. This allows the API code to be called in a UI in case someone wants to access it directly instead of through the helpers.

Specifically, someone could call the API directly in a React component using `useMemo` (React built-in memoization method, but only for a React component). Didn't use this because we'd still need one lower down the call stack.

bdomen-ggl:
*[`experimental/reporting-ui/src/main/react/reporting-ui/library/utils/api/realApi.ts` line 35 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1143#-NaiDBazB95fs3DIDAzj:-Nar7eavE0nBzPA3ZggK:b930u5w) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/320575ec11bce12e6ea212238c1c4b73b63d8456/experimental/reporting-ui/src/main/react/reporting-ui/library/utils/api/realApi.ts#L35)):*

Done
Removed async since this is only doing a fetch and I want to return a promise anyways. This allows the API code to be called in a UI in case someone wants to access it directly instead of through the helpers.

Specifically, someone could call the API directly in a React component using `useMemo` (React built-in memoization method, but only for a React component). Didn't use this because we'd still need one lower down the call stack.
</blockquote></details>

Actually, left async/await and removed thens. It was an autosuggestion and I didn't realize how that behavior worked.

bdomen-ggl:
*[`experimental/reporting-ui/src/main/react/reporting-ui/library/utils/api/memoize.ts` line 15 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1143#-NaiCw0K15QZVi3OTH-o:-Nam2kz38tSxZGN3ovnC:b-7whst5) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/320575ec11bce12e6ea212238c1c4b73b63d8456/experimental/reporting-ui/src/main/react/reporting-ui/library/utils/api/memoize.ts#L15)):*

A map is a fine implementation of a cache (not every cache needs to handle all the policies/features internally). I also know that you were trying to share caching across the real and fake service clients, but I don't think that's necessary.
</blockquote></details>

DONE
Gotcha. I moved this all into a class so it's not global anymore. And yeah, I was trying to share the caching. Since the widgets are independent of each other, they can make many of the same calls to Get Report.

Also, I looked into the cache API, but it must use URLs as keys, which I don't think we necessarily want to enforce since there may be other operations that would be useful to memoize as well. Also when using the fake data class, there isn't a URL being called.

bdomen-ggl:
*[`experimental/reporting-ui/src/main/react/reporting-ui/library/utils/api/IReportingApi.ts` line 22 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1143#-Nai9HJR868rrUecRvb8:-NarKTQC0aLf1QawIMGQ:b-cs9n7p) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/320575ec11bce12e6ea212238c1c4b73b63d8456/experimental/reporting-ui/src/main/react/reporting-ui/library/utils/api/IReportingApi.ts#L22)):*

Maybe also call it `ReportingClient` instead, since it's really an API client not an API. I also don't think it makes sense to be in a `utils` package, as generally utils are free functions.
</blockquote></details>

Done
Yeah, that makes sense. I was having a rough time with naming this one since there's similar things.
Renamed to ReportingClient and moved everything out of utils.

bdomen-ggl:
*[`experimental/reporting-ui/src/main/react/reporting-ui/library/utils/api/fakeApi.ts` line 308 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1143#-NaiB3_j6THPpN0D5Vn9:-NarP21iCAwKXhBsmyj_:bb6un1j) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/320575ec11bce12e6ea212238c1c4b73b63d8456/experimental/reporting-ui/src/main/react/reporting-ui/library/utils/api/fakeApi.ts#L308)):*

Yes I can. I was initially doing it this way since I was expecting there to be some more logic. I simplified some assumptions so that's no longer the case.
</blockquote></details>

Done

bdomen-ggl:
*[`experimental/reporting-ui/src/main/react/reporting-ui/library/utils/api/fakeApi.ts` line 305 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1143#-NaiEQjc844BgOWteuke:-NarPv9QDb1_byC-AI2_:b-5966e1) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/320575ec11bce12e6ea212238c1c4b73b63d8456/experimental/reporting-ui/src/main/react/reporting-ui/library/utils/api/fakeApi.ts#L305)):*

Shallow is fine. It's mostly about intent, since JS can't really stop you from doing most things.
</blockquote></details>

Done
Sounds good. Just wanted to make sure.

bdomen-ggl:
*[`experimental/reporting-ui/src/main/react/reporting-ui/library/utils/api/fakeApi.ts` line 1 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1143#-Nai8nGk0M7c1vxvS0F1:-NarSEvX794Dm_yV75ab:b7ovllk) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/320575ec11bce12e6ea212238c1c4b73b63d8456/experimental/reporting-ui/src/main/react/reporting-ui/library/utils/api/fakeApi.ts#L1)):*

This is coming from the [Google JS style guide](https://google.github.io/styleguide/jsguide.html#file-name). It doesn't appear to be mentioned in the TS style guide, but I assume falling back to the JS style guide makes sense when the TS one doesn't specify.

If we're looking for other style guides, it looks like React is relatively unopinionated on the matter and there isn't a popular style guide that covers this. FWIW the Angular style guide uses lowercase with hyphens, so either way I don't think lowerCamelCase is the right choice.
</blockquote></details>

Done
React is fairly un-opinionated on the topic. They do have some base suggestions from the create react app repo (nothing written, just how the app is created). It's also confusing because they have a different set of styles for JS vs TS. I'll stick with the style guide you linked.

SanjayVas (changes requested):
> Reviewed 9 of 22 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1143)* status: 9 of 22 files reviewed, 8 unresolved discussions (waiting on @bdomen-ggl)

*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1143#-NasQ3HgAM_849vZPSIh:-NasQ3HgAM_849vZPSIi:b-gho8hr) (no related file):*
This is missing BUILD.bazel files. Each folder with source files should be a Bazel package and therefore have a BUILD.bazel file. A source file should generally not be referenced outside of a Bazel package directly, and instead should be accessed via an appropriate target (usually using a `<lang>_library` rule.

Bazel packages (meaning folder names) should also all be lowercase.

___
___
*[`experimental/reporting-ui/src/main/react/reporting-ui/library/reportingClient/real_api.ts` line 36 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1143#-NasPduoAKB1DoXnEe55:-NasPduoAKB1DoXnEe56:b-mgnnlw) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1697b75d1e062b5c5c1e740132d7b74cf50580b7/experimental/reporting-ui/src/main/react/reporting-ui/library/reportingClient/real_api.ts#L36)):*
> ```XML
>   async listReports(): Promise<ListReportsResponse> {
>     const res = await fetch(path.join(this.baseUrl.toString(), '/api/reports'));
>     const reports = await res.json();
> ```

I don't think this returns a Promise. See https://developer.mozilla.org/en-US/docs/Web/API/Response/json_static



_Suggestion:_
```XML
res.json();
```
___
*[`experimental/reporting-ui/src/main/react/reporting-ui/library/utils/api/memoize.ts` line 15 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1143#-NaiCw0K15QZVi3OTH-o:-NasQTAu0S7nhsPtZGe0:b-6btiv3) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/320575ec11bce12e6ea212238c1c4b73b63d8456/experimental/reporting-ui/src/main/react/reporting-ui/library/utils/api/memoize.ts#L15)):*
> Gotcha. I moved this all into a class so it's not global anymore. And yeah, I was trying to share the caching.

Sorry, the "global" part I was concerned with wasn't the fact that the functions were unscoped (though that too), but that the cache is shared. If you're trying to cache RPC responses, that should be handled in the client impl.

> Since the widgets are independent of each other, they can make many of the same calls to Get Report.

That doesn't seem right. AFAIK all of the common MV* design patterns use model objects. Some single thing (e.g. a button) should trigger loading of a Report, e.g. updating or replacing some Report model object. How that gets propagated to all the components is dependent on the pattern.

bdomen-ggl (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1143)* status: 7 of 30 files reviewed, 8 unresolved discussions (waiting on @SanjayVas)

*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1143#-NasQ3HgAM_849vZPSIh:-NavpQcB13nTgbwlohTW:b-6xhczy) (no related file):*

This is missing BUILD.bazel files. Each folder with source files should be a Bazel package and therefore have a BUILD.bazel file. A source file should generally not be referenced outside of a Bazel package directly, and instead should be accessed via an appropriate target (usually using a `<lang>_library` rule.

Bazel packages (meaning folder names) should also all be lowercase.
</blockquote></details>

Right, I was going to add it later when the library starts to get used as I add the views. I can add BUILD.bazel now.

By the way, when you say "each folder with source files," do you mean ALL folders including all subdirectories or just the top level ones? I was generally following the Aspect example here (https://github.com/bazelbuild/rules_nodejs/tree/stable/examples/create-react-app) and they only have the BUILD file at the workspace root. Although they're building in a very different way using `react_scripts`. I thought it would be nice to go another level and add BUILD files for things like src and public since they contain different kinds of files, like they could be different packages.

UPDATE: Talked with Sanjay. The example I was following is a bad Bazel pattern (using a glob to load "**/*.ts" for example) and I should not do it that way. And yes, each and every folder should have a BUILD.bazel file. Got the library/BUILD file, but need to add all the rest and update the first one to NOT use the bad pattern and reference the subdirectories.

___
___
*[`experimental/reporting-ui/src/main/react/reporting-ui/library/reportingClient/real_api.ts` line 36 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1143#-NasPduoAKB1DoXnEe55:-Navl-Le8CMLiQyAKiFa:b-neknxg) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1697b75d1e062b5c5c1e740132d7b74cf50580b7/experimental/reporting-ui/src/main/react/reporting-ui/library/reportingClient/real_api.ts#L36)):*

I don't think this returns a Promise. See https://developer.mozilla.org/en-US/docs/Web/API/Response/json_static
</blockquote></details>

Ah weird.
What I had was the suggestion from the fetch API docs:
https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch.
Mouseover in my environment also shows `.json()` returns a promise. Same when I use a JavaScript playground online. I'll look into this some more.
It does seem weird that `.json()` would return a promise.

UPDATE: Ah nvm, you are looking at the static method `Response.json(obj)` but I'm using the instance method `response.json()`. The prior returns an object and the later returns a promise.
___
*[`experimental/reporting-ui/src/main/react/reporting-ui/library/utils/api/fakeApi.ts` line 1 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1143#-Nai8Ngm0-R4ijoWbBZJ:-NamDDXnAvhtLdh1H-Ss:b3mhq5c) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/320575ec11bce12e6ea212238c1c4b73b63d8456/experimental/reporting-ui/src/main/react/reporting-ui/library/utils/api/fakeApi.ts#L1)):*

I'm not saying the fake is a bad idea. You can still use it, just make sure it's in the right place.

The way we do this in server-side code is to have all service clients/stubs be provided to the top-level constructor. This way something at the top can decide whether to use the fake. I don't know what the top level is for a React app (is there some App type?) but you'd have a testing version of that which also goes under a `testing` Bazel package.
</blockquote></details>

WIP
React kind of does but kind of doesn't have a top-level constructor. There's the index.tsx which is the entry point for webpack bundling where we can probably add configurations and the like. I'm having an annoying time since JavaScript doesn't handle DI terribly nicely. So trying to find a way to inject even like a class map at the top level.

UPDATE: Going to copy index.tsx to the test folder and update the imports to the main folder. Then before the app gets rendered, I can bootstrap the appropriate configurations (dev/test/prod) including setting up the reporting client.
___
*[`experimental/reporting-ui/src/main/react/reporting-ui/library/utils/api/memoize.ts` line 15 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1143#-NaiCw0K15QZVi3OTH-o:-Navm-wrDHyttXLclA8p:b-ftrqi3) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/320575ec11bce12e6ea212238c1c4b73b63d8456/experimental/reporting-ui/src/main/react/reporting-ui/library/utils/api/memoize.ts#L15)):*

> Gotcha. I moved this all into a class so it's not global anymore. And yeah, I was trying to share the caching.

Sorry, the "global" part I was concerned with wasn't the fact that the functions were unscoped (though that too), but that the cache is shared. If you're trying to cache RPC responses, that should be handled in the client impl.

> Since the widgets are independent of each other, they can make many of the same calls to Get Report.

That doesn't seem right. AFAIK all of the common MV* design patterns use model objects. Some single thing (e.g. a button) should trigger loading of a Report, e.g. updating or replacing some Report model object. How that gets propagated to all the components is dependent on the pattern.
</blockquote></details>

Talked on a call about the MV[something] pattern:
I was under the impression we wanted to have 100% self-contained widgets (they would know how to render and fetch data given a base reporting api url) and they would be their own MV[something] pattern. This is not the desired behavior. The upper level "something" (ie. the get report view business logic) should handle fetching the data, turning it into the right model (if not automatically done by the api) and then passed to the widget that will just render the model based on it's template. The widget should not know how to populate its own data.

Don't think there's much else to change in this PR since it's only memoizing the call to the reporting api.

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1143)* status: 0 of 35 files reviewed, 12 unresolved discussions (waiting on @bdomen-ggl)

___
*[`experimental/reporting-ui/src/main/react/reporting-ui/library/BUILD.bazel` line 1 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1143#-Nb1PoaE0pMOP05GAxDj:-Nb1PoaE0pMOP05GAxDk:be45jjg) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d337f472e630598eabc4b85b622e417cd5abad57/experimental/reporting-ui/src/main/react/reporting-ui/library/BUILD.bazel#L1)):*
> ```Starlark
> load("@aspect_rules_js//js:defs.bzl", "js_library")
> ```

It seems odd to call this package `library` when most Bazel packages will expose library targets. What is it a library *of*?
___
*[`experimental/reporting-ui/src/main/react/reporting-ui/library/BUILD.bazel` line 15 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1143#-Nb1PRo--uKt-TwNTlPJ:-Nb1PRo--uKt-TwNTlPK:b-quciy1) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d337f472e630598eabc4b85b622e417cd5abad57/experimental/reporting-ui/src/main/react/reporting-ui/library/BUILD.bazel#L15)):*
> ```Starlark
> ]
> 
> js_library(
> ```

What's the benefit to having both a js_library and a ts_project (assuming the latter is also a library rule)? It looks like ts_library can take in TS source files, and that the js_library target isn't use outside of the package anyway (and should therefore have private visibility).
___
*[`experimental/reporting-ui/src/main/react/reporting-ui/library/BUILD.bazel` line 22 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1143#-Nb1Mcr-0E7x_Qof-rX4:-Nb1Mcr-0E7x_Qof-rX5:b573f1h) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d337f472e630598eabc4b85b622e417cd5abad57/experimental/reporting-ui/src/main/react/reporting-ui/library/BUILD.bazel#L22)):*
> ```Starlark
> )
> 
> ts_project(
> ```

So is `ts_project` effectively a library rule?

For background, most Bazel packages tend to expose `<lang>_library` targets for that language. Then there's a `<lang>_binary` target that does the compilation.
___
*[`experimental/reporting-ui/src/main/react/reporting-ui/library/BUILD.bazel` line 23 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1143#-Nb1OuA1AFez7Z5simSn:-Nb1OuA1AFez7Z5simSo:bfysa7i) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d337f472e630598eabc4b85b622e417cd5abad57/experimental/reporting-ui/src/main/react/reporting-ui/library/BUILD.bazel#L23)):*
> ```Starlark
> 
> ts_project(
>     name = "src_ts",
> ```

Generally if you have a single library target for a Bazel package you name it the same thing as the package to indicate it's the "default" target.

Using more granular targets is preferred over globbing all source files in a package, but not necessary if all of the sources get used together anyway.



_Code quote:_
```Starlark
src_ts
```

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1143)* status: 0 of 35 files reviewed, 12 unresolved discussions (waiting on @bdomen-ggl)

___
*[`experimental/reporting-ui/src/main/react/reporting-ui/library/BUILD.bazel` line 22 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1143#-Nb1Mcr-0E7x_Qof-rX4:-NbG75rz3QVSu9hE8beJ:b-9jymn0) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d337f472e630598eabc4b85b622e417cd5abad57/experimental/reporting-ui/src/main/react/reporting-ui/library/BUILD.bazel#L22)):*

ts_project wraps the tsc cli:
https://github.com/aspect-build/rules_ts/blob/main/docs/rules.md

If you're suggesting to then take that output and put it in a js_library or something I can do that with this:
https://github.com/aspect-build/rules_js/blob/main/docs/js_library.md
</blockquote></details>

I'm just trying to figure out the intended usage, meaning what rules are intended to consume `ts_project` output. Looking at the inputs/outputs, it looks like this is meant to take in TS (and optionally JS) source files as `srcs`, and other targets that produce TS typings as `deps`. It outputs a JsInfo provider. Therefore it looks to me like a TS replacement for `js_library`.

bdomen-ggl (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1143)* status: 0 of 54 files reviewed, 13 unresolved discussions (waiting on @SanjayVas)

*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1143#-NbH0nPMAJWUwnw1ARsi:-NbH0nPMAJWUwnw1ARsj:bkbu14c) (no related file):*
@SanjayVas Latest is ready. Updated the file structure and folder names and updated all the BUILD files.

___
___
*[`experimental/reporting-ui/src/main/react/reporting-ui/library/BUILD.bazel` line 22 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1143#-Nb1Mcr-0E7x_Qof-rX4:-NbGgB1aDIKanBn7p9Mi:b62jcfd) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d337f472e630598eabc4b85b622e417cd5abad57/experimental/reporting-ui/src/main/react/reporting-ui/library/BUILD.bazel#L22)):*

I'm just trying to figure out the intended usage, meaning what rules are intended to consume `ts_project` output. Looking at the inputs/outputs, it looks like this is meant to take in TS (and optionally JS) source files as `srcs`, and other targets that produce TS typings as `deps`. It outputs a JsInfo provider. Therefore it looks to me like a TS replacement for `js_library`.
</blockquote></details>

Talked off line. ts_project was an artifact from some build rules I realized I'm not longer using. Can just get rid of it. But yes, it can take TS and JS files as input and will output something similar to a js_library, but with any TS files compiled into JS.

SanjayVas (changes requested):
> Reviewed 1 of 11 files at r4, 1 of 31 files at r5, 3 of 15 files at r6, 45 of 49 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1143)* status: 50 of 54 files reviewed, 13 unresolved discussions (waiting on @bdomen-ggl)

___
*[`experimental/reporting-ui/src/main/react/reporting-ui/BUILD.bazel` line 8 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1143#-NbM8i1t-urMCFjUArdp:-NbM8i1t-urMCFjUArdq:bz73zqy) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ec5641fcd83f79971c1cbc4c1be78240150b8fd1/experimental/reporting-ui/src/main/react/reporting-ui/BUILD.bazel#L8)):*
> ```Starlark
> 
> js_library(
>     name = "src",
> ```

Prefer more descriptive names.



_Suggestion:_
```Starlark
app
```
___
*[`experimental/reporting-ui/src/main/react/reporting-ui/BUILD.bazel` line 24 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1143#-NbM8vM81q8oimf_TG5m:-NbM8vM81q8oimf_TG5n:boxvvp1) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ec5641fcd83f79971c1cbc4c1be78240150b8fd1/experimental/reporting-ui/src/main/react/reporting-ui/BUILD.bazel#L24)):*
> ```Starlark
> 
> js_library(
>     name = "build_node_deps",
> ```

Not sure what the "build" is about. Perhaps just `node_deps` or `node_modules`?



_Suggestion:_
```Starlark
node_modules
```
___
*[`experimental/reporting-ui/src/test/react/reporting-ui/fake/src/BUILD.bazel` line 9 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1143#-NbMDkW5CX259FW4nI4D:-NbMDkW5CX259FW4nI4E:b-oj29r9) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ec5641fcd83f79971c1cbc4c1be78240150b8fd1/experimental/reporting-ui/src/test/react/reporting-ui/fake/src/BUILD.bazel#L9)):*
> ```Starlark
> )
> 
> js_library(
> ```

This should be marked testonly and in a `testing` package. `reporting-ui/client/reporting/testing` seems like a good place, since this is a fake implementation of `ReportingClient` for testing.
___
*[`experimental/reporting-ui/src/main/react/reporting-ui/client/reporting/client_impl.ts` line 33 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1143#-NbMAzAd-zODyr6JrDDn:-NbMAYlp6NfznoNdhFkh:bwariqr) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ec5641fcd83f79971c1cbc4c1be78240150b8fd1/experimental/reporting-ui/src/main/react/reporting-ui/client/reporting/client_impl.ts#L33)):*
> ```XML
> 
>   getReport(req: GetReportRequest): Promise<GetReportResponse> {
>     const getCached = this.memoizer.memoizePromiseFn(this.api.getReport);
> ```

1. Any caching should be specific to the implementation rather than attempting to share it across implementations (e.g. a real implementation may want to cache, a fake won't).
2. It's only safe to permanently cache Reports in terminal states, meaning caching here can't be implemented as simple memoization.

This means that you shouldn't have a base `ReportingClientImpl` that's shared across implementations, and you can freely use that name for the real client impl.
___
*[`experimental/reporting-ui/src/main/react/reporting-ui/client/reporting/real_api.ts` line 23 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1143#-NbMCkhCEhZYlAyBvKMt:-NbMCkhCEhZYlAyBvKMu:bvjwmii) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ec5641fcd83f79971c1cbc4c1be78240150b8fd1/experimental/reporting-ui/src/main/react/reporting-ui/client/reporting/real_api.ts#L23)):*
> ```XML
> import {ReportingClient} from './client';
> 
> export class RealApi implements ReportingClient {
> ```

The real client impl can just be named `ReportingClientImpl`.

See comment on `client_impl.ts`



_Code quote:_
```XML
RealApi
```

bdomen-ggl (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1143)* status: 39 of 55 files reviewed, 13 unresolved discussions (waiting on @SanjayVas)

*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1143#-NbH0nPMAJWUwnw1ARsi:-NbQp6JCA0vWFjN8TWH6:bmbmm0) (no related file):*

@SanjayVas Latest is ready. Updated the file structure and folder names and updated all the BUILD files.
</blockquote></details>

Next round of comments addressed.

___
___
*[`experimental/reporting-ui/src/main/react/reporting-ui/BUILD.bazel` line 8 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1143#-NbM8i1t-urMCFjUArdp:-NbPUfcSDW9FbjVLVW4k:bb6un1j) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ec5641fcd83f79971c1cbc4c1be78240150b8fd1/experimental/reporting-ui/src/main/react/reporting-ui/BUILD.bazel#L8)):*

Prefer more descriptive names.
</blockquote></details>

Done
___
*[`experimental/reporting-ui/src/main/react/reporting-ui/BUILD.bazel` line 24 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1143#-NbM8vM81q8oimf_TG5m:-NbPUgC6E4ZwoaX5XoXj:bb6un1j) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ec5641fcd83f79971c1cbc4c1be78240150b8fd1/experimental/reporting-ui/src/main/react/reporting-ui/BUILD.bazel#L24)):*

Not sure what the "build" is about. Perhaps just `node_deps` or `node_modules`?
</blockquote></details>

Done
___
*[`experimental/reporting-ui/src/test/react/reporting-ui/fake/src/BUILD.bazel` line 9 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1143#-NbMDkW5CX259FW4nI4D:-NbPV8fiBDQVef3sZnUn:b-vtciwe) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ec5641fcd83f79971c1cbc4c1be78240150b8fd1/experimental/reporting-ui/src/test/react/reporting-ui/fake/src/BUILD.bazel#L9)):*

This should be marked testonly and in a `testing` package. `reporting-ui/client/reporting/testing` seems like a good place, since this is a fake implementation of `ReportingClient` for testing.
</blockquote></details>

Ah sorry. Forgot to mark testonly. And gotcha, thought when you said "testing package" before to put it in the test folder.
___
*[`experimental/reporting-ui/src/main/react/reporting-ui/client/reporting/client_impl.ts` line 33 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1143#-NbMAzAd-zODyr6JrDDn:-NbPmXaY2TIhH0NsOuj0:bb6un1j) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ec5641fcd83f79971c1cbc4c1be78240150b8fd1/experimental/reporting-ui/src/main/react/reporting-ui/client/reporting/client_impl.ts#L33)):*

1. Any caching should be specific to the implementation rather than attempting to share it across implementations (e.g. a real implementation may want to cache, a fake won't).
2. It's only safe to permanently cache Reports in terminal states, meaning caching here can't be implemented as simple memoization.

This means that you shouldn't have a base `ReportingClientImpl` that's shared across implementations, and you can freely use that name for the real client impl.
</blockquote></details>

Done
___
*[`experimental/reporting-ui/src/main/react/reporting-ui/client/reporting/real_api.ts` line 23 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1143#-NbMCkhCEhZYlAyBvKMt:-NbPmZLX3Cm5fKqnnDSi:bb6un1j) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ec5641fcd83f79971c1cbc4c1be78240150b8fd1/experimental/reporting-ui/src/main/react/reporting-ui/client/reporting/real_api.ts#L23)):*

The real client impl can just be named `ReportingClientImpl`.

See comment on `client_impl.ts`
</blockquote></details>

Done

SanjayVas (changes requested):
> Reviewed 3 of 49 files at r7, 13 of 13 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1143)* status: all files reviewed, 11 unresolved discussions (waiting on @bdomen-ggl)

___
*[`experimental/reporting-ui/src/main/react/reporting-ui/client/reporting/client_impl.ts` line 33 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1143#-NbMAzAd-zODyr6JrDDn:-NbR5mp93YMYxiQPXE_Z:bdpxzf0) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ec5641fcd83f79971c1cbc4c1be78240150b8fd1/experimental/reporting-ui/src/main/react/reporting-ui/client/reporting/client_impl.ts#L33)):*

Done
</blockquote></details>

I still see this doing unconditional caching.
___
*[`experimental/reporting-ui/src/main/react/reporting-ui/client/reporting/testing/fake_api.ts` line 45 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1143#-NbR60tr5wZC1rFt53ql:-NbR60ts11d_X2YyhDV4:b45h) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e7b4d571c23cdd7ec7f758d6398514b6897f6790/experimental/reporting-ui/src/main/react/reporting-ui/client/reporting/testing/fake_api.ts#L45)):*
> ```XML
> }
> 
> export class FakeApi implements ReportingClient {
> ```





_Suggestion:_
```XML
FakeReportingClient
```

bdomen-ggl (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1143)* status: 46 of 56 files reviewed, 11 unresolved discussions (waiting on @SanjayVas)

___
*[`experimental/reporting-ui/src/main/react/reporting-ui/client/reporting/client_impl.ts` line 33 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1143#-NbMAzAd-zODyr6JrDDn:-NbRAT6YDtoRUNnTeTl4:b356d9s) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ec5641fcd83f79971c1cbc4c1be78240150b8fd1/experimental/reporting-ui/src/main/react/reporting-ui/client/reporting/client_impl.ts#L33)):*

I still see this doing unconditional caching.
</blockquote></details>

Sorry, was focusing too much on #1 and forgot #2.

bdomen-ggl (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1143)* status: 46 of 56 files reviewed, 11 unresolved discussions (waiting on @SanjayVas)

___
*[`experimental/reporting-ui/src/main/react/reporting-ui/client/reporting/testing/fake_api.ts` line 45 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1143#-NbR60tr5wZC1rFt53ql:-NbVQCRz7e2n0iFoZV4l:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e7b4d571c23cdd7ec7f758d6398514b6897f6790/experimental/reporting-ui/src/main/react/reporting-ui/client/reporting/testing/fake_api.ts#L45)):*
> ```XML
> }
> 
> export class FakeApi implements ReportingClient {
> ```

Done.

bdomen-ggl:
*[`experimental/reporting-ui/src/main/react/reporting-ui/client/reporting/client_impl.ts` line 33 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1143#-NbMAzAd-zODyr6JrDDn:-NbVcq29DJlw6_6gIN7v:b-9v5u3f) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ec5641fcd83f79971c1cbc4c1be78240150b8fd1/experimental/reporting-ui/src/main/react/reporting-ui/client/reporting/client_impl.ts#L33)):*

Sorry, was focusing too much on #1 and forgot #2.
</blockquote></details>

OK... not sure what those links are, meant to be number one and number two.

bdomen-ggl (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1143)* status: 46 of 56 files reviewed, 11 unresolved discussions (waiting on @SanjayVas)

*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1143#-NbH0nPMAJWUwnw1ARsi:-NbVci5b07KzD6cDeWuz:be985h6) (no related file):*

Next round of comments addressed.
</blockquote></details>

In case I missed it, next round ready for review. Think that's the last of them so far.

___

SanjayVas (changes requested):
> Reviewed 10 of 10 files at r9, 3 of 3 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1143)* status: all files reviewed, 7 unresolved discussions (waiting on @bdomen-ggl)

___
*[`experimental/reporting-ui/src/main/react/reporting-ui/model/reporting/report_state.ts` line 22 at r10](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1143#-Nbp3KBS2Eb1U456FkMF:-Nbp3KBS2Eb1U456FkMG:b-pxwtgj) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6ea972d878c44f2c3bd33184822c537b859a62dd/experimental/reporting-ui/src/main/react/reporting-ui/model/reporting/report_state.ts#L22)):*
> ```XML
> };
> 
> export const terminatingStates = Object.freeze([ReportState.SUCCEEDED, ReportState.FAILED]);
> ```

I assume constants should be in `SCREAMING_SNAKE_CASE`. Also these are called terminal states, not terminating.



_Suggestion:_
```XML
TERMINAL_STATES
```

bdomen-ggl (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1143)* status: 53 of 56 files reviewed, 7 unresolved discussions (waiting on @SanjayVas)

___
*[`experimental/reporting-ui/src/main/react/reporting-ui/model/reporting/report_state.ts` line 22 at r10](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1143#-Nbp3KBS2Eb1U456FkMF:-Nbyf97rCWzRqDQfy0ki:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6ea972d878c44f2c3bd33184822c537b859a62dd/experimental/reporting-ui/src/main/react/reporting-ui/model/reporting/report_state.ts#L22)):*

I assume constants should be in `SCREAMING_SNAKE_CASE`. Also these are called terminal states, not terminating.
</blockquote></details>

Done.

SanjayVas (approved):
> Reviewed 1 of 1 files at r11, 2 of 2 files at r12, 1 of 1 files at r13, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1143)* status: all files reviewed, 5 unresolved discussions (waiting on @bdomen-ggl)

PR #1146, opened 2023-08-02
EdpSimulator fulfills Reach-Only LiquidLegionsV2 requisitions.

renjiezh:


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1146)

kungfucraig (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1146)* status: all files reviewed, 1 unresolved discussion (waiting on @iverson52000, @ple13, @renjiezh, and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt` line 767 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1146#-NasA7QK60XBCfLUlCeR:-Nas9zV_05xdc70i20CE:b-ujuacv) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/2e3ea24217f2fae5be1d0df393c0c52c5b214cd0/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt#L767)):*
> ```Kotlin
>    * and send to the duchy.
>    */
>   private suspend fun fulfillRequisitionForReachOnlyLiquidLegionsV2Measurement(
> ```

The copy paste reuse is a bit bothersome, but I'm not sure there's a great alternative that's not going to be either 1) a pain to implement; 2) hard to read. :(

ple13 (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1146)* status: all files reviewed, 1 unresolved discussion (waiting on @iverson52000, @renjiezh, and @SanjayVas)

iverson52000 (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1146)* status: all files reviewed, 1 unresolved discussion (waiting on @renjiezh and @SanjayVas)

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1146)* status: all files reviewed, 4 unresolved discussions (waiting on @kungfucraig and @renjiezh)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt` line 442 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1146#-Nb0uWJH5CYJKjbUtvlN:-Nb0uWJH5CYJKjbUtvlO:b-f824id) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/2e3ea24217f2fae5be1d0df393c0c52c5b214cd0/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt#L442)):*
> ```Kotlin
>         "Consent signaling verification failed for ${requisition.name}"
>       }
>       refuseRequisition(
> ```

This goes a bit against the existing pattern, where the refusals for up-front spec/consent validation happen directly in `executeRequisitionFulfillingWorkflow`.

It makes me think it might be better to have a local `RequisitionRefusalException` that we throw from all the code that wants to trigger refusal and handle that in the top-level loop. This way we also avoid having to sprinkle `continue` everywhere in that loop.
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt` line 567 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1146#-Nb0v6hq5pCoiJbjtT33:-Nb0v6hq5pCoiJbjtT34:bem94aw) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/2e3ea24217f2fae5be1d0df393c0c52c5b214cd0/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt#L567)):*
> ```Kotlin
>             )
>             continue
>           } else if (protocols.any { it.hasReachOnlyLiquidLegionsV2() }) {
> ```

Just noting that this inside of the case for `REACH` *or* `REACH_AND_FREQUENCY`. Right now the Direct protocol is valid for all measurement types. R/F LLv2 can technically work for either `REACH` or `REACH_AND_FREQUENCY`. RO LLv2 only works for `REACH`.
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt` line 767 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1146#-NasA7QK60XBCfLUlCeR:-Nb0yWxrCzVZHQ6NeSkk:bpih166) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/2e3ea24217f2fae5be1d0df393c0c52c5b214cd0/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt#L767)):*

The copy paste reuse is a bit bothersome, but I'm not sure there's a great alternative that's not going to be either 1) a pain to implement; 2) hard to read. :(
</blockquote></details>

You can possibly share more code if you make a Kotlin class hierarchy that can take advantage of inheritance, which protobuf doesn't have. e.g. having a sealed class hierarchy for LLv2 protocol, where the base class is for RO. Then you can have code that safely operates on both variants since it can do compiler-enforced exhaustive `when` expressions.
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SketchEncrypter.kt` line 89 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1146#-Nb11i6tAJaAtk_vdFM1:-Nb11i6tAJaAtk_vdFM2:b22asoo) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/2e3ea24217f2fae5be1d0df393c0c52c5b214cd0/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SketchEncrypter.kt#L89)):*
> ```Kotlin
>     sketch: Sketch,
>     encryptionKey: ElGamalPublicKey,
>     protocol: ProtocolConfig.Protocol,
> ```

This library intentionally doesn't reference the public API.

I'd suggest making a separate overload for the `encrypt` method in the interface for reach-only.

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1146)* status: all files reviewed, 4 unresolved discussions (waiting on @kungfucraig, @renjiezh, and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt` line 442 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1146#-Nb0uWJH5CYJKjbUtvlN:-Nb1N0Zt0ljyqcIcXco4:b-hhge1o) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/2e3ea24217f2fae5be1d0df393c0c52c5b214cd0/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt#L442)):*

This goes a bit against the existing pattern, where the refusals for up-front spec/consent validation happen directly in `executeRequisitionFulfillingWorkflow`.

It makes me think it might be better to have a local `RequisitionRefusalException` that we throw from all the code that wants to trigger refusal and handle that in the top-level loop. This way we also avoid having to sprinkle `continue` everywhere in that loop.
</blockquote></details>

In this file, there are already two local Exception classes and another PrivacyBudgetManagerException from PBM. If I understand correctly, I will let each iteration catch these 3 types of exception and call refuseRequisition. So that the function refuseRequisition only occurs in one catch block.
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt` line 567 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1146#-Nb0v6hq5pCoiJbjtT33:-Nb1OLdm5M35vUKq_bGh:b-51p0ep) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/2e3ea24217f2fae5be1d0df393c0c52c5b214cd0/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt#L567)):*

Just noting that this inside of the case for `REACH` *or* `REACH_AND_FREQUENCY`. Right now the Direct protocol is valid for all measurement types. R/F LLv2 can technically work for either `REACH` or `REACH_AND_FREQUENCY`. RO LLv2 only works for `REACH`.
</blockquote></details>

Yeah. But seems no easy way to reduce the lines of code here.
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt` line 767 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1146#-NasA7QK60XBCfLUlCeR:-Nb1OZhL7iLOB9GvGy3g:bkuf15s) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/2e3ea24217f2fae5be1d0df393c0c52c5b214cd0/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt#L767)):*

You can possibly share more code if you make a Kotlin class hierarchy that can take advantage of inheritance, which protobuf doesn't have. e.g. having a sealed class hierarchy for LLv2 protocol, where the base class is for RO. Then you can have code that safely operates on both variants since it can do compiler-enforced exhaustive `when` expressions.
</blockquote></details>

Guess I will keep it as-is for readability?
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SketchEncrypter.kt` line 89 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1146#-Nb11i6tAJaAtk_vdFM1:-Nb1Ofw_EarHuQe7wjPK:bhy9ot6) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/2e3ea24217f2fae5be1d0df393c0c52c5b214cd0/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SketchEncrypter.kt#L89)):*

This library intentionally doesn't reference the public API.

I'd suggest making a separate overload for the `encrypt` method in the interface for reach-only.
</blockquote></details>

Will do.

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1146)* status: all files reviewed, 4 unresolved discussions (waiting on @kungfucraig and @renjiezh)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt` line 442 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1146#-Nb0uWJH5CYJKjbUtvlN:-Nb1RKq-4knD-quHUt6U:b-o9frjc) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/2e3ea24217f2fae5be1d0df393c0c52c5b214cd0/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt#L442)):*

In this file, there are already two local Exception classes and another PrivacyBudgetManagerException from PBM. If I understand correctly, I will let each iteration catch these 3 types of exception and call refuseRequisition. So that the function refuseRequisition only occurs in one catch block.
</blockquote></details>

I think you can wrap all of those in `RequisitionRefusalException` so there's only one type you have to handle up top.
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt` line 567 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1146#-Nb0v6hq5pCoiJbjtT33:-Nb1QAqNBbnuvtFGROpc:b-7obmq4) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/2e3ea24217f2fae5be1d0df393c0c52c5b214cd0/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt#L567)):*

Yeah. But seems no easy way to reduce the lines of code here.
</blockquote></details>

Making it clear is more important than avoiding duplication, though I think the duplication can still be kept minimal with a little refactoring. One option is to invert the checks, where you look at protocol first and then measurement type within each protocol (note that some of the methods appear to check the measurement type again, e.g. despite the method name, `fulfillDirectReachAndFrequencyMeasurement` is used for either REACH or REACH_AND_FREQUENCY`).

The other option is to duplicate some logic in both the `REACH` and `REACH_AND_FREQUENCY` branches.

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1146)* status: 1 of 5 files reviewed, 4 unresolved discussions (waiting on @iverson52000, @kungfucraig, @ple13, and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt` line 442 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1146#-Nb0uWJH5CYJKjbUtvlN:-NbG0pqZ0sb6QtDr1ISc:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/2e3ea24217f2fae5be1d0df393c0c52c5b214cd0/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt#L442)):*

I think you can wrap all of those in `RequisitionRefusalException` so there's only one type you have to handle up top.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt` line 567 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1146#-Nb0v6hq5pCoiJbjtT33:-NbG0qVW9eM8K_7YF6dH:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/2e3ea24217f2fae5be1d0df393c0c52c5b214cd0/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt#L567)):*

Making it clear is more important than avoiding duplication, though I think the duplication can still be kept minimal with a little refactoring. One option is to invert the checks, where you look at protocol first and then measurement type within each protocol (note that some of the methods appear to check the measurement type again, e.g. despite the method name, `fulfillDirectReachAndFrequencyMeasurement` is used for either REACH or REACH_AND_FREQUENCY`).

The other option is to duplicate some logic in both the `REACH` and `REACH_AND_FREQUENCY` branches.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SketchEncrypter.kt` line 89 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1146#-Nb11i6tAJaAtk_vdFM1:-NbG0tTd3rmdn5gtCAn7:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/2e3ea24217f2fae5be1d0df393c0c52c5b214cd0/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SketchEncrypter.kt#L89)):*

Will do.
</blockquote></details>

Done.

SanjayVas (changes requested):
> Reviewed 4 of 4 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1146)* status: all files reviewed, 6 unresolved discussions (waiting on @kungfucraig and @renjiezh)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt` line 549 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1146#-NbG2lKE2kJraWlxJJ6N:-NbG2lKE2kJraWlxJJ6O:bbcyn3c) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/28dfcd91c07632010dee6c4c039bbe10a3591566/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt#L549)):*
> ```Kotlin
>             fulfillImpressionMeasurement(requisition, requisitionSpec, measurementSpec)
>           } else {
>             logger.log(
> ```

Should refuse rather than just log I think, otherwise you'll keep reprocessing this Requisition. I'd consider this a case where the spec is invalid.

Basically one of the following actions should be taken for each Requisition that's meant for this (EDP, MC) pair: fulfill or refuse.
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt` line 555 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1146#-NbG3aQTAzs93nqnsGxw:-NbG3aQU0qlH0Xaa9n54:bf4wkw7) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/28dfcd91c07632010dee6c4c039bbe10a3591566/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt#L555)):*
> ```Kotlin
>           }
>         } else if (protocols.any { it.hasLiquidLegionsV2() }) {
>           require(measurementSpec.hasReach() || measurementSpec.hasReachAndFrequency()) {
> ```

Should refuse rather than throw a runtime exception, otherwise you'll crash loop trying to re-process this Requisition. I'd consider this a case where the spec is invalid.
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt` line 568 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1146#-NbG3opy2DpMOxaUjgiI:-NbG3opy2DpMOxaUjgiJ:bf4wkw7) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/28dfcd91c07632010dee6c4c039bbe10a3591566/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt#L568)):*
> ```Kotlin
>           )
>         } else if (protocols.any { it.hasReachOnlyLiquidLegionsV2() }) {
>           require(measurementSpec.hasReach()) {
> ```

Should refuse rather than throw a runtime exception, otherwise you'll crash loop trying to re-process this Requisition. I'd consider this a case where the spec is invalid.
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt` line 584 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1146#-NbG3sVO1B3irSUWWv0f:-NbG3sVO1B3irSUWWv0g:bfttg0f) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/28dfcd91c07632010dee6c4c039bbe10a3591566/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt#L584)):*
> ```Kotlin
>           )
>         } else {
>           logger.log(
> ```

Should refuse rather than just log, otherwise you'll crash loop trying to re-process this Requisition. I'd consider this a case where the spec is invalid.

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1146)* status: 4 of 5 files reviewed, 6 unresolved discussions (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt` line 549 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1146#-NbG2lKE2kJraWlxJJ6N:-NbLOuCQ5FaaivVw2AgG:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/28dfcd91c07632010dee6c4c039bbe10a3591566/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt#L549)):*

Refusal makes sense here.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt` line 555 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1146#-NbG3aQTAzs93nqnsGxw:-NbLOxL7CxKkD-Mb453O:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/28dfcd91c07632010dee6c4c039bbe10a3591566/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt#L555)):*

Agreed. In general if we cannot fulfill for any permanent reason we should refuse the requisition (i.e. fail fast).
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt` line 568 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1146#-NbG3opy2DpMOxaUjgiI:-NbLOy7vEY9VkSFvdR55:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/28dfcd91c07632010dee6c4c039bbe10a3591566/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt#L568)):*

Should refuse rather than throw a runtime exception, otherwise you'll crash loop trying to re-process this Requisition. I'd consider this a case where the spec is invalid.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt` line 584 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1146#-NbG3sVO1B3irSUWWv0f:-NbLOzApEsZh7Fo1gAv4:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/28dfcd91c07632010dee6c4c039bbe10a3591566/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt#L584)):*

+1
</blockquote></details>

Done.

SanjayVas (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1146)* status: all files reviewed, 1 unresolved discussion (waiting on @renjiezh)

kungfucraig (approved):
> Reviewed 3 of 4 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1146)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Marco-Premier)

YuhongWang-Amazon (approved):
> Reviewed 1 of 5 files at r1, 3 of 4 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1146)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Marco-Premier)

renjiezh (commented):
> Reviewed 2 of 4 files at r2, 3 of 3 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1146)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Marco-Premier)

PR #1151, opened 2023-08-03
Update maximum per user in Metric Spec for Reporting V2

tristanvuong2021:


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1151)

PR #1183, opened 2023-08-22
Update cross-media-measurement-api dep for maximum frequency specification change.

SanjayVas:
Update the cross-media-measurement-api dependency for world-federation-of-advertisers/cross-media-measurement-api#165.

This also fixes an issue where the EDP simulator was not respecting the maximum frequency specified in the `MeasurementSpec` for R/F measurements using the `Direct` protocol.


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1183)

SanjayVas:
@renjiezh for Duchy changes

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1183)* status: 0 of 39 files reviewed, all discussions resolved (waiting on @renjiezh and @riemanli)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/liquidlegionsv2/LiquidLegionsV2Mill.kt` line 996 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1183#-NcU7MVBBHbnqlJNGDhU:-NcU7MVC4-jWIRJBMmtK:b-s2b1ll) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0c6eccb7539947ca26a5db5f8a4b6288c9df26c7/src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/liquidlegionsv2/LiquidLegionsV2Mill.kt#L996)):*
> ```Kotlin
>       if (measurementSpec.hasReach()) 1
>       else measurementSpec.reachAndFrequency.maximumFrequencyPerUser
>     if (measurementSpecMaximumFrequency > 0) {
> ```

This was a bug. `maximum_frequency_per_user` should have had no effect on Duchy computations. If it were supported, it would need to be handled on the EDP end.
___
*[`src/main/resources/reporting/postgres/create-v2-reporting-schema.sql` line 242 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1183#-NcU6Z6GAQdNMLDSNWuz:-NcU6Z6GAQdNMLDSNWv-:b-cnk83l) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/61fcb7f1441a933737979420aebebad05eb3208e/src/main/resources/reporting/postgres/create-v2-reporting-schema.sql#L242)):*
> ```TSQL
> 
>   -- Must not be NULL if MetricType is FREQUENCY_HISTOGRAM
>   MaximumFrequency bigint,
> ```

This should be okay as nobody has deployed Reporting v2 anywhere yet.
___
*[`src/test/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/MeasurementsServiceTest.kt` line 906 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1183#-NcU7-__1cDSlWGgSqNR:-NcU7-_a1UpHZ9zXbtdK:b-kex188) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/61fcb7f1441a933737979420aebebad05eb3208e/src/test/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/MeasurementsServiceTest.kt#L906)):*
> ```Kotlin
> 
>   @Test
>   fun `createMeasurement throws INVALID_ARGUMENT when RF spec is missing max frequency`() {
> ```

This is the new test. The other changes to INVALID_ARGUMENT tests are because many of them were underspecified and were actually triggering "missing parent" errors rather than the error they were testing for.

riemanli (commented):
> Reviewed 36 of 39 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1183)* status: 36 of 39 files reviewed, all discussions resolved (waiting on @renjiezh and @tristanvuong2021)

___
*[`src/main/resources/reporting/postgres/create-v2-reporting-schema.sql` line 242 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1183#-NcU6Z6GAQdNMLDSNWuz:-NcYRylO4L88hh2-h_wu:bcvap3o) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/61fcb7f1441a933737979420aebebad05eb3208e/src/main/resources/reporting/postgres/create-v2-reporting-schema.sql#L242)):*

This should be okay as nobody has deployed Reporting v2 anywhere yet.
</blockquote></details>

@tristanvuong2021 has deployed for integration tests. Is that still ok?

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1183)* status: 36 of 39 files reviewed, all discussions resolved (waiting on @renjiezh and @tristanvuong2021)

___
*[`src/main/resources/reporting/postgres/create-v2-reporting-schema.sql` line 242 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1183#-NcU6Z6GAQdNMLDSNWuz:-NcYZxfi4VnnHjorm6JB:bi2du94) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/61fcb7f1441a933737979420aebebad05eb3208e/src/main/resources/reporting/postgres/create-v2-reporting-schema.sql#L242)):*

@tristanvuong2021 has deployed for integration tests. Is that still ok?
</blockquote></details>

The GitHub workflow for updating the dev environment looks like it still only deploys Reporting v1. Therefore this shouldn't break any release testing. The process for switching to Reporting v2 would involve Terraforming a new DB.

Any other deployments were done manually so there's no guarantees on clean updates.

tristanvuong2021 (commented):
> Reviewed 27 of 39 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1183)* status: 36 of 39 files reviewed, all discussions resolved (waiting on @renjiezh)

riemanli (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1183)* status: 36 of 39 files reviewed, all discussions resolved (waiting on @renjiezh)

tristanvuong2021 (commented):
> Reviewed 9 of 39 files at r1.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1183)* status: 37 of 39 files reviewed, all discussions resolved (waiting on @renjiezh)

tristanvuong2021 (commented):
> Reviewed 2 of 39 files at r1.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1183)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @renjiezh)

tristanvuong2021 (changes requested):
> Reviewed 1 of 39 files at r1.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1183)* status: all files reviewed, 1 unresolved discussion (waiting on @renjiezh and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v1alpha/ReportsService.kt` line 204 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1183#-NcZ5Vu-8NnMLqDQbtQa:-NcZ5Vu0147q_JBnynls:b-s35jyb) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/61fcb7f1441a933737979420aebebad05eb3208e/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v1alpha/ReportsService.kt#L204)):*
> ```Kotlin
> private const val REACH_FREQUENCY_REACH_EPSILON = 0.0033
> private const val REACH_FREQUENCY_FREQUENCY_EPSILON = 0.115
> private const val MAXIMUM_FREQUENCY = 10
> ```

Why is the user input ignored and a constant used instead?

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1183)* status: 37 of 41 files reviewed, 1 unresolved discussion (waiting on @renjiezh, @riemanli, and @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v1alpha/ReportsService.kt` line 204 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1183#-NcZ5Vu-8NnMLqDQbtQa:-NcZEQOaBd86SP0TnXlC:b-kj4fdd) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/61fcb7f1441a933737979420aebebad05eb3208e/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v1alpha/ReportsService.kt#L204)):*

Why is the user input ignored and a constant used instead?
</blockquote></details>

The Reporting v1 API doesn't have this field. We're no longer making changes to that API.

tristanvuong2021 (approved):
> Reviewed 7 of 7 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1183)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @renjiezh and @riemanli)

tristanvuong2021 (approved):
> Reviewed 1 of 2 files at r3, 3 of 3 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1183)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @renjiezh and @riemanli)

renjiezh (approved):
> Reviewed 32 of 39 files at r1, 7 of 7 files at r2, 1 of 2 files at r3, 3 of 3 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1183)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @riemanli)

riemanli (commented):
> Reviewed 2 of 7 files at r2, 1 of 2 files at r3.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1183)* status: all files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/postgres/writers/SetMeasurementResult.kt` line 398 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1183#-NcitBo77pRgvil7Xyai:-NcitBo77pRgvil7Xyaj:bdbewdg) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/95cc6b31be8f24db56c63080ca8910a87a24d7cc/src/main/kotlin/org/wfanet/measurement/reporting/deploy/postgres/writers/SetMeasurementResult.kt#L398)):*
> ```Kotlin
>             this.columnHeader = columnHeader
>             for (frequencyHistogram in frequencyHistograms) {
>               for (frequency in 1..largestFrequency) {
> ```

Just want to make the behavior change is intentional. The new code will show the results up to the maximum frequency of seen frequencies, whereas before will show all results up to the maximum frequency set by the user even they are 0s at the last few frequencies.

This probably doesn't matter a lot, but I would like to point it out in case the user build the UI based on the previous behavior.

tristanvuong2021 (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1183)* status: all files reviewed, 1 unresolved discussion (waiting on @riemanli)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/postgres/writers/SetMeasurementResult.kt` line 398 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1183#-NcitBo77pRgvil7Xyai:-NcivlDjAL0w0tbL0zM7:bql4wx8) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/95cc6b31be8f24db56c63080ca8910a87a24d7cc/src/main/kotlin/org/wfanet/measurement/reporting/deploy/postgres/writers/SetMeasurementResult.kt#L398)):*

Just want to make the behavior change is intentional. The new code will show the results up to the maximum frequency of seen frequencies, whereas before will show all results up to the maximum frequency set by the user even they are 0s at the last few frequencies.

This probably doesn't matter a lot, but I would like to point it out in case the user build the UI based on the previous behavior.
</blockquote></details>

There is an issue with there being more results than row headers because row headers aren't created based on the results. This would actually match up results with row headers.

riemanli (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1183)* status: all files reviewed, 1 unresolved discussion (waiting on @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/postgres/writers/SetMeasurementResult.kt` line 398 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1183#-NcitBo77pRgvil7Xyai:-NcixeXU-F1gnNG7C8Rh:b-vu3mtw) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/95cc6b31be8f24db56c63080ca8910a87a24d7cc/src/main/kotlin/org/wfanet/measurement/reporting/deploy/postgres/writers/SetMeasurementResult.kt#L398)):*
> There is an issue with there being more results than row headers because row headers aren't created based on the results.

Is the issue in the current system? Based on what I saw in https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1ddf3a8ff410cdb41d30be10058e25b62ccbb74a/src/main/kotlin/org/wfanet/measurement/reporting/deploy/postgres/writers/SetMeasurementResult.kt#L388-L396 and https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1ddf3a8ff410cdb41d30be10058e25b62ccbb74a/src/main/kotlin/org/wfanet/measurement/reporting/deploy/postgres/writers/SetMeasurementResult.kt#L373-L375 the headers and results seem matched with range 1~maxFreq.

tristanvuong2021 (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1183)* status: all files reviewed, 1 unresolved discussion (waiting on @riemanli)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/postgres/writers/SetMeasurementResult.kt` line 398 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1183#-NcitBo77pRgvil7Xyai:-Ncj-uXP8UKYcqsWXtL7:b34ysqj) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/95cc6b31be8f24db56c63080ca8910a87a24d7cc/src/main/kotlin/org/wfanet/measurement/reporting/deploy/postgres/writers/SetMeasurementResult.kt#L398)):*

> There is an issue with there being more results than row headers because row headers aren't created based on the results.

Is the issue in the current system? Based on what I saw in https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1ddf3a8ff410cdb41d30be10058e25b62ccbb74a/src/main/kotlin/org/wfanet/measurement/reporting/deploy/postgres/writers/SetMeasurementResult.kt#L388-L396 and https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1ddf3a8ff410cdb41d30be10058e25b62ccbb74a/src/main/kotlin/org/wfanet/measurement/reporting/deploy/postgres/writers/SetMeasurementResult.kt#L373-L375 the headers and results seem matched with range 1~maxFreq.
</blockquote></details>

The latest commit you see fixes an issue, but it still doesn't help with more results than row headers. The change here sets the table based on the actual results. It will be easier to see what is going on with the results with this.

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1183)* status: 39 of 44 files reviewed, all discussions resolved (waiting on @renjiezh, @riemanli, and @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/postgres/writers/SetMeasurementResult.kt` line 52 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1183#-NcZkPxaD9heZO3bG5Wf:-NcZkPxaD9heZO3bG5Wg:b-3frwm8) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/95cc6b31be8f24db56c63080ca8910a87a24d7cc/src/main/kotlin/org/wfanet/measurement/reporting/deploy/postgres/writers/SetMeasurementResult.kt#L52)):*
> ```Kotlin
>  * * [MeasurementStateInvalidException] Measurement does not have PENDING state.
>  */
> class SetMeasurementResult(private val request: SetMeasurementResultRequest) :
> ```

@tristanvuong2021 I had to make changes here to fix the fact that `maximum_frequency_per_user` was being treated as `maxumum_frequency`. PTAL.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/postgres/writers/SetMeasurementResult.kt` line 398 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1183#-NcitBo77pRgvil7Xyai:-NcxKfN75Dr0Vlz3jRC_:b549x3l) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/95cc6b31be8f24db56c63080ca8910a87a24d7cc/src/main/kotlin/org/wfanet/measurement/reporting/deploy/postgres/writers/SetMeasurementResult.kt#L398)):*
> Just want to make the behavior change is intentional. The new code will show the results up to the maximum frequency of seen frequencies, whereas before will show all results up to the maximum frequency set by the user even they are 0s at the last few frequencies.

Yes, this is intentional. It makes the output agnostic to the input and purely based on the results. This means that we properly display the actual result even in the case where the result provider (e.g. the EDP for a direct measurement) did not respect the input max frequency.

> This probably doesn't matter a lot, but I would like to point it out in case the user build the UI based on the previous behavior.

The behavior you're mentioning is recent and hasn't even made it into a release.

tristanvuong2021 (approved):
> Reviewed 5 of 5 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1183)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @renjiezh and @riemanli)

riemanli (changes requested):
> Reviewed 3 of 7 files at r2, 3 of 3 files at r4, 5 of 5 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1183)* status: all files reviewed, 1 unresolved discussion (waiting on @renjiezh and @SanjayVas)

___
*[`src/main/k8s/local/metric_spec_config.textproto` line 18 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1183#-Ncy2bTf3Npp3T-Q1ssq:-Ncy2bTf3Npp3T-Q1ssr:b-ioa8ge) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8d0ffec21d8214116313f4c65d609bab84ad6f09/src/main/k8s/local/metric_spec_config.textproto#L18)):*
> ```
>     delta: 1.0E-12
>   }
>   maximum_frequency_per_user: 10
> ```

Should we instead set the default value of `maximum_frequency` in the frequency histogram params?

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1183)* status: 43 of 44 files reviewed, 1 unresolved discussion (waiting on @renjiezh, @riemanli, and @tristanvuong2021)

___
*[`src/main/k8s/local/metric_spec_config.textproto` line 18 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1183#-Ncy2bTf3Npp3T-Q1ssq:-NcyWk6453nFZk6ra13J:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8d0ffec21d8214116313f4c65d609bab84ad6f09/src/main/k8s/local/metric_spec_config.textproto#L18)):*

Should we instead set the default value of `maximum_frequency` in the frequency histogram params?
</blockquote></details>

Done.

riemanli (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1183)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @renjiezh and @tristanvuong2021)

stevenwarejones (commented):
Can you update the title to `Update cross-media-measurement for maximum frequency specification change.`

> Reviewed 4 of 39 files at r1, 1 of 1 files at r6.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1183#-:-Nd7eiWH-RallT5pGFR8:b-tcih1k)* status: 43 of 45 files reviewed, all discussions resolved (waiting on @renjiezh, @riemanli, and @tristanvuong2021)

SanjayVas (commented):
The title is reflecting the fact that this is updating the cross-media-measurement-api dependency, and is consistent with other PRs that have updated a dep (e.g. "Update <dep> to 0.x.y"). I've added the word "dep" to make this clearer.

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1183#-:-Nd7iZG99SR8QbY9hc5S:b-vnkq9t)* status: 43 of 45 files reviewed, all discussions resolved (waiting on @renjiezh, @riemanli, and @tristanvuong2021)

SanjayVas (commented):
> consistent with other PRs that have updated a dep (e.g. "Update <dep> to 0.x.y")

Markdown ate my example title. It should read "Update &lt;dep> to 0.x.y"

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1183#-:-Nd7j9Hf3o4F69AXhtUP:b-a9av2z)* status: 43 of 45 files reviewed, all discussions resolved (waiting on @renjiezh, @riemanli, and @tristanvuong2021)

riemanli (approved):
> Reviewed 2 of 2 files at r7, 1 of 3 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1183)* status: 43 of 45 files reviewed, all discussions resolved (waiting on @renjiezh and @tristanvuong2021)

stevenwarejones (changes requested):
> Reviewed 2 of 39 files at r1, 2 of 2 files at r7, 2 of 3 files at r8, 1 of 5 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1183)* status: 43 of 51 files reviewed, 5 unresolved discussions (waiting on @renjiezh, @riemanli, @SanjayVas, and @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/LiquidLegionsV2Starter.kt` line 333 at r11](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1183#-NdHTTGDBghpwE5hilyS:-NdHTTGDBghpwE5hilyT:b-o2426a) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1ee0d8287516537151ebfc1befa9dfdc54986bc2/src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/LiquidLegionsV2Starter.kt#L333)):*
> ```Kotlin
>         maximumFrequency = protocolConfig.maximumFrequency
>       }
>       require(maximumFrequency > 0) { "Maximum frequency must be greater than 0" }
> ```

it should be greater than 1. If its just 1, they should be doing a reach-only measurement.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/liquidlegionsv2/LiquidLegionsV2Mill.kt` line 996 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1183#-NcU7MVBBHbnqlJNGDhU:-NdHU6bF2TLO6FPxp6Kq:b-dw6ypz) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0c6eccb7539947ca26a5db5f8a4b6288c9df26c7/src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/liquidlegionsv2/LiquidLegionsV2Mill.kt#L996)):*

This was a bug. `maximum_frequency_per_user` should have had no effect on Duchy computations. If it were supported, it would need to be handled on the EDP end.
</blockquote></details>

not to mention that secret sharing would not have supported it
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/liquidlegionsv2/LiquidLegionsV2Mill.kt` line 560 at r11](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1183#-NdHTfSY5D-qQO_qhMd_:-NdHTfSY5D-qQO_qhMda:b-7ropln) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1ee0d8287516537151ebfc1befa9dfdc54986bc2/src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/liquidlegionsv2/LiquidLegionsV2Mill.kt#L560)):*
> ```Kotlin
>     val llv2Parameters = llv2Details.parameters
>     require(AGGREGATOR == llv2Details.role) { "invalid role for this function." }
>     val maximumRequestedFrequency = llv2Parameters.maximumFrequency.coerceAtLeast(1)
> ```

ditto
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/liquidlegionsv2/LiquidLegionsV2Mill.kt` line 659 at r11](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1183#-NdHTm4B6xpFtAfTpTB3:-NdHTm4B6xpFtAfTpTB4:b-7ropln) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1ee0d8287516537151ebfc1befa9dfdc54986bc2/src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/liquidlegionsv2/LiquidLegionsV2Mill.kt#L659)):*
> ```Kotlin
>     }
>     var reach = 0L
>     val maximumRequestedFrequency = llv2Parameters.maximumFrequency.coerceAtLeast(1)
> ```

ditto
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/liquidlegionsv2/LiquidLegionsV2Mill.kt` line 764 at r11](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1183#-NdHTpz48Y8tzO8J6bfL:-NdHTpz48Y8tzO8J6bfM:b-7ropln) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1ee0d8287516537151ebfc1befa9dfdc54986bc2/src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/liquidlegionsv2/LiquidLegionsV2Mill.kt#L764)):*
> ```Kotlin
>     val llv2Parameters = llv2Details.parameters
>     require(NON_AGGREGATOR == llv2Details.role) { "invalid role for this function." }
>     val maximumRequestedFrequency = llv2Parameters.maximumFrequency.coerceAtLeast(1)
> ```

ditto
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/liquidlegionsv2/LiquidLegionsV2Mill.kt` line 805 at r11](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1183#-NdHTumvCgbNR_8RMHHO:-NdHTumvCgbNR_8RMHHP:b-q9h4u9) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1ee0d8287516537151ebfc1befa9dfdc54986bc2/src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/liquidlegionsv2/LiquidLegionsV2Mill.kt#L805)):*
> ```Kotlin
> 
>     // If this is a reach-only computation, then our job is done.
>     if (maximumRequestedFrequency == 1) {
> ```

imo, we should never get here if its =1 - that should have been requested as a reach only measurement

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1183)* status: 43 of 51 files reviewed, 5 unresolved discussions (waiting on @renjiezh, @riemanli, @stevenwarejones, and @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/LiquidLegionsV2Starter.kt` line 333 at r11](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1183#-NdHTTGDBghpwE5hilyS:-NdHakOH9cq2gi30niYx:b-wk8yvf) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1ee0d8287516537151ebfc1befa9dfdc54986bc2/src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/LiquidLegionsV2Starter.kt#L333)):*

it should be greater than 1. If its just 1, they should be doing a reach-only measurement.
</blockquote></details>

We still allow this, even though it's not optimal. When the`REACH` Measurement type was added, we didn't want to break existing callers who were using the `REACH_AND_FREQUENCY` Measurement type for "reach-only" measurements by specifying a max frequency of 1. These existing callers include Reporting v1.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/liquidlegionsv2/LiquidLegionsV2Mill.kt` line 560 at r11](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1183#-NdHTfSY5D-qQO_qhMd_:-NdHb11KEmHt0xnrbDnp:b-cjqbf2) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1ee0d8287516537151ebfc1befa9dfdc54986bc2/src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/liquidlegionsv2/LiquidLegionsV2Mill.kt#L560)):*

ditto
</blockquote></details>

There are two valid protocols for the `REACH` Measurement type: the existing R/F LLv2 and the new RO LLv2. Which protocol is actually used is controlled by the Kingdom flag. Therefore, this Mill needs to be able to run Measurements using a max frequency of 1.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/liquidlegionsv2/LiquidLegionsV2Mill.kt` line 659 at r11](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1183#-NdHTm4B6xpFtAfTpTB3:-NdHbLFg6-B8Cs1L1OZD:b-ctrph2) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1ee0d8287516537151ebfc1befa9dfdc54986bc2/src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/liquidlegionsv2/LiquidLegionsV2Mill.kt#L659)):*

ditto
</blockquote></details>

See above comment
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/liquidlegionsv2/LiquidLegionsV2Mill.kt` line 764 at r11](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1183#-NdHTpz48Y8tzO8J6bfL:-NdHbQtb0n0-Qn0DiD_Q:b-ctrph2) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1ee0d8287516537151ebfc1befa9dfdc54986bc2/src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/liquidlegionsv2/LiquidLegionsV2Mill.kt#L764)):*

ditto
</blockquote></details>

See above comment
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/liquidlegionsv2/LiquidLegionsV2Mill.kt` line 805 at r11](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1183#-NdHTumvCgbNR_8RMHHO:-NdHbRzhECiajddwJR-J:b-ctrph2) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1ee0d8287516537151ebfc1befa9dfdc54986bc2/src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/liquidlegionsv2/LiquidLegionsV2Mill.kt#L805)):*

imo, we should never get here if its =1 - that should have been requested as a reach only measurement
</blockquote></details>

See above comment

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1183)* status: 43 of 51 files reviewed, 5 unresolved discussions (waiting on @renjiezh, @riemanli, @stevenwarejones, and @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/liquidlegionsv2/LiquidLegionsV2Mill.kt` line 996 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1183#-NcU7MVBBHbnqlJNGDhU:-NdHcCdQAlYMW9Rj5kUL:bp0icvn) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0c6eccb7539947ca26a5db5f8a4b6288c9df26c7/src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/liquidlegionsv2/LiquidLegionsV2Mill.kt#L996)):*

not to mention that secret sharing would not have supported it
</blockquote></details>

It turns out this was okay, just not clearly documented. Basically the max frequency in the LLv2 protocol config was acting as the maximum possible value for maximum frequency. Therefore the protocol implementation was actually respecting the max frequency specified by the MC in the request, but only as long as it didn't exceed the protocol config value. In that case, it would silently coerce it to the lower value.

riemanli (approved):
> Reviewed 1 of 39 files at r1, 2 of 3 files at r8, 5 of 5 files at r9, 5 of 5 files at r10, 5 of 5 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1183)* status: all files reviewed, 5 unresolved discussions (waiting on @renjiezh, @stevenwarejones, and @tristanvuong2021)

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1183)* status: all files reviewed, 3 unresolved discussions (waiting on @renjiezh, @SanjayVas, and @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/LiquidLegionsV2Starter.kt` line 333 at r11](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1183#-NdaFBk17j5Bwjhzp3JL:-NdaFBk17j5Bwjhzp3JM:b-7gsk8v) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1ee0d8287516537151ebfc1befa9dfdc54986bc2/src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/LiquidLegionsV2Starter.kt#L333)):*
> ```Kotlin
>         maximumFrequency = protocolConfig.maximumFrequency
>       }
>       require(maximumFrequency > 0) { "Maximum frequency must be greater than 0" }
> ```

wouldn't it be better for the kingdom to internally convert reach_and_frequency measurements with a max frequency of 1 to reach_only measurements?

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1183)* status: all files reviewed, 3 unresolved discussions (waiting on @renjiezh, @stevenwarejones, and @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/LiquidLegionsV2Starter.kt` line 333 at r11](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1183#-NdaFBk17j5Bwjhzp3JL:-NdahXVD7PJhGtxbLjP6:btwsl3c) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1ee0d8287516537151ebfc1befa9dfdc54986bc2/src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/LiquidLegionsV2Starter.kt#L333)):*

wouldn't it be better for the kingdom to internally convert reach_and_frequency measurements with a max frequency of 1 to reach_only measurements?
</blockquote></details>

The MeasurementSpec is signed by the MC. The Kingdom cannot tamper with it.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/postgres/writers/SetMeasurementResult.kt` line 360 at r11](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1183#-NdaD_IsEBJW49N3-DMc:-NdagsXkA7P4_sQTX3se:be1r4n9) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1ee0d8287516537151ebfc1befa9dfdc54986bc2/src/main/kotlin/org/wfanet/measurement/reporting/deploy/postgres/writers/SetMeasurementResult.kt#L360)):*
> aggregating them to the correct maxFrequency in case the EDP used too high of a max frequency?

AFAIK, the Reporting system doesn't try to correct anything. Whether we even want it to is a separate discussion and outside the scope of this PR.
___
*[`src/main/proto/wfa/measurement/reporting/v1alpha/metric.proto` line 50 at r11](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1183#-NdaECOIA02rJm09tXbx:-NdaiBey-tiPsrtNWP_Z:br88l4n) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1ee0d8287516537151ebfc1befa9dfdc54986bc2/src/main/proto/wfa/measurement/reporting/v1alpha/metric.proto#L50)):*

the comment is a little confusing. if its not supported, then it should get moved to reserved
</blockquote></details>

We're avoiding forcing changes to Reporting v1 clients. This is also why this PR doesn't rename the field in `FrequencyHistogramParams` to `maximum_frequency` like it does for the Reporting v2 API.

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1183)* status: all files reviewed, 3 unresolved discussions (waiting on @renjiezh, @SanjayVas, and @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/LiquidLegionsV2Starter.kt` line 333 at r11](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1183#-NdaFBk17j5Bwjhzp3JL:-Ndb35ef878IXPYY-7iV:b-r7t57v) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1ee0d8287516537151ebfc1befa9dfdc54986bc2/src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/LiquidLegionsV2Starter.kt#L333)):*

The MeasurementSpec is signed by the MC. The Kingdom cannot tamper with it.
</blockquote></details>

good point. Or why not just alter the mpc workers to collaboratively treat the measurement like a reach_only measurement is max_frequency is 1?

kungfucraig (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1183)* status: all files reviewed, 3 unresolved discussions (waiting on @renjiezh, @stevenwarejones, and @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/LiquidLegionsV2Starter.kt` line 333 at r11](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1183#-NdaFBk17j5Bwjhzp3JL:-Ndb9ApqD4hDPNI_kNyz:br22dnz) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1ee0d8287516537151ebfc1befa9dfdc54986bc2/src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/LiquidLegionsV2Starter.kt#L333)):*

good point. Or why not just alter the mpc workers to collaboratively treat the measurement like a reach_only measurement is max_frequency is 1?
</blockquote></details>

Maybe have the Kingdom throw an error if max frequency <= 1, and make the client create a reach only measurement. 

The API implementation probably shouldn't play games by changing the type and the inputs that an EDP should supply are different in a reach-only scenario vs. a reach+frequency scenario.

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1183)* status: all files reviewed, 3 unresolved discussions (waiting on @renjiezh, @stevenwarejones, and @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/LiquidLegionsV2Starter.kt` line 333 at r11](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1183#-NdaFBk17j5Bwjhzp3JL:-Ndb7zNbAukD-0_PmXsU:bcpa8bz) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1ee0d8287516537151ebfc1befa9dfdc54986bc2/src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/LiquidLegionsV2Starter.kt#L333)):*

Maybe have the Kingdom throw an error if max frequency <= 1, and make the client create a reach only measurement. 

The API implementation probably shouldn't play games by changing the type and the inputs that an EDP should supply are different in a reach-only scenario vs. a reach+frequency scenario.
</blockquote></details>

I don't know if that's worth it to make that change now. Ideally we just get all callers to migrate to the `REACH` Measurement type. After that we can require max frequency to be > 1 and delete the code to handle that case in the Mill impl.

kungfucraig (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1183)* status: all files reviewed, 3 unresolved discussions (waiting on @renjiezh, @stevenwarejones, and @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/LiquidLegionsV2Starter.kt` line 333 at r11](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1183#-NdaFBk17j5Bwjhzp3JL:-NdbE59V42WeSsAjXGxA:bx0a5kl) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1ee0d8287516537151ebfc1befa9dfdc54986bc2/src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/LiquidLegionsV2Starter.kt#L333)):*

I don't know if that's worth it to make that change now. Ideally we just get all callers to migrate to the `REACH` Measurement type. After that we can require max frequency to be > 1 and delete the code to handle that case in the Mill impl.
</blockquote></details>

@SanjayVas Philosophically you and I are in the same spot. Your approach seems reasonable to me.

stevenwarejones (approved):
> Reviewed 1 of 3 files at r4, 12 of 12 files at r12, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1183)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @renjiezh and @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/postgres/writers/SetMeasurementResult.kt` line 360 at r11](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1183#-NdaD_IsEBJW49N3-DMc:-NdfxZnE7Lb_5A11vC4e:b-lzq7hm) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1ee0d8287516537151ebfc1befa9dfdc54986bc2/src/main/kotlin/org/wfanet/measurement/reporting/deploy/postgres/writers/SetMeasurementResult.kt#L360)):*

> aggregating them to the correct maxFrequency in case the EDP used too high of a max frequency?

AFAIK, the Reporting system doesn't try to correct anything. Whether we even want it to is a separate discussion and outside the scope of this PR.
</blockquote></details>

This means that the results of a direct vs mpc measurement could be different for single edp since the mpc does auto-capping.

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1183)* status: 50 of 51 files reviewed, all discussions resolved (waiting on @renjiezh, @stevenwarejones, and @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/postgres/writers/SetMeasurementResult.kt` line 360 at r11](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1183#-NdaD_IsEBJW49N3-DMc:-NdgJP4WCGGyRHF8L6p6:b-gpz5i2) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1ee0d8287516537151ebfc1befa9dfdc54986bc2/src/main/kotlin/org/wfanet/measurement/reporting/deploy/postgres/writers/SetMeasurementResult.kt#L360)):*
> This means that the results of a direct vs mpc measurement could be different for single edp

We dropped support for single-EDP reach via the LLv2 protocol. If there's only one EDP for a REACH or REACH_AND_FREQUENCY measurement, the only valid protocol is Direct.

>  since the mpc does auto-capping.

I'm not sure what you mean here. The LLv2 and direct protocols both currently use a fixed maximum frequency. The system makes no guarantees about the result if an EDP does not correctly follow the protocol.

SanjayVas (commented):
> Reviewed 1 of 1 files at r13, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1183)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @renjiezh, @stevenwarejones, and @tristanvuong2021)

SanjayVas (commented):
> Reviewed 1 of 12 files at r12.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1183)* status: 50 of 51 files reviewed, all discussions resolved (waiting on @renjiezh, @stevenwarejones, and @tristanvuong2021)

SanjayVas (commented):
> Reviewed 1 of 1 files at r14, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1183)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @renjiezh, @stevenwarejones, and @tristanvuong2021)

stevenwarejones (approved):
> Reviewed 1 of 1 files at r13, 1 of 1 files at r14, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1183)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @renjiezh and @tristanvuong2021)

SanjayVas (commented):
> Reviewed 1 of 7 files at r15, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1183)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @renjiezh, @riemanli, @stevenwarejones, and @tristanvuong2021)

SanjayVas (commented):
> Reviewed 1 of 1 files at r16.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1183)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @renjiezh, @riemanli, @stevenwarejones, and @tristanvuong2021)

PR #1154, opened 2023-08-05
Circulate direct computation methodology from protocol config to measurement result.

riemanli:
In this PR, the implementation of the [API changes in CMM-API](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/pull/163) is done. The main changes are 
1. new direct protocol config,
2. additional info in measurement results, and
3. new enum constants in `ProtocolConfig.NoiseMechanism`. 
4. EDP simulator choose the noise mechanism based on the overlap of its preferences and the noise mechanism options in the direct protocol config.

Any future measurement creations will contain certain direct protocol configs based on the measurement types. For the backward compatibility for existing measurements and requisitions, the empty internal direct protocol configs will be converted with the options that would have been valid at the time the Measurement was created.

EDPs are restricted to use continuous noise mechanisms by setting the validation in `V2alphaPublicApiServer`.

PR #1107, opened 2023-07-14
Add internal DP params proto

riemanli:
Part of the changes for implementing error bars in reporting server. Before adding an internal protocol config of measurements to the internal reporting server, we pulled the DP params message out of the internal metric message as later the internal protocol config needs it.

PR #1100, opened 2023-07-12
Store set operation compilation IDs

riemanli:
Part of the changes for implementing error bars in reporting server. To easily find the union of any two weighted measurements in a metric, we store the binary representation of each measurement from the binary representation of the weighed subset union created during set expression compilation of a reporting set.

PR #1184, opened 2023-08-22
Store measurement computation details in reporting for error bars

riemanli:
In addition to storing results in the internal measurements, we also store the computation details about the measurement creation. The structure of the internal measurement result is now same as the one in Kingdom -- measurement results are in a list to represent results from different data providers; noise mechanisms and computation methodologies are added to `Result`. Note that unlike the Kingdom `Result` which stores only direct noise mechanisms and direct methodologies, here we store both direct and MPC noise mechanisms and methodologies.

The additional computation details can be used to compute measurement variances and potentially used to identify measurements. No DB change is needed as the changes happened in `Details`.

### DIFF
--- a/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/postgres/writers/SetMeasurementResults.kt
+++ b/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/postgres/writers/SetMeasurementResults.kt
@@ -52,7 +52,7 @@ class SetMeasurementResults(private val request: BatchSetMeasurementResultsReque
       ) {
         val state = Measurement.State.SUCCEEDED
         request.measurementResultsList.forEach {
-          val details = MeasurementKt.details { result = it.result }
+          val details = MeasurementKt.details { results += it.resultsList }
           addBinding {
             bind("$1", details)
             bind("$2", details.toJson())
--- a/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt
+++ b/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt
@@ -96,6 +96,7 @@ import org.wfanet.measurement.internal.reporting.v2.BatchSetCmmsMeasurementIdsRe
 import org.wfanet.measurement.internal.reporting.v2.BatchSetCmmsMeasurementIdsResponse
 import org.wfanet.measurement.internal.reporting.v2.BatchSetCmmsMeasurementResultsResponse
 import org.wfanet.measurement.internal.reporting.v2.BatchSetMeasurementFailuresRequestKt.measurementFailure
+import org.wfanet.measurement.internal.reporting.v2.BatchSetMeasurementResultsRequest
 import org.wfanet.measurement.internal.reporting.v2.BatchSetMeasurementResultsRequestKt.measurementResult
 import org.wfanet.measurement.internal.reporting.v2.CreateMetricRequest as InternalCreateMetricRequest
 import org.wfanet.measurement.internal.reporting.v2.Measurement as InternalMeasurement
@@ -770,15 +771,11 @@ class MetricsService(
         cmmsMeasurementConsumerId = principal.resourceKey.measurementConsumerId
         measurementResults +=
           succeededMeasurementsList.map { measurement ->
-            measurementResult {
-              cmmsMeasurementId = MeasurementKey.fromName(measurement.name)!!.measurementId
-              result =
-                buildInternalMeasurementResult(
-                  measurement,
-                  apiAuthenticationKey,
-                  principal.resourceKey.toName()
-                )
-            }
+            buildInternalMeasurementResult(
+              measurement,
+              apiAuthenticationKey,
+              principal.resourceKey.toName()
+            )
           }
       }
 
@@ -839,7 +836,7 @@ class MetricsService(
       measurement: Measurement,
       apiAuthenticationKey: String,
       principalName: String,
-    ): InternalMeasurement.Result {
+    ): BatchSetMeasurementResultsRequest.MeasurementResult {
       val measurementSpec = MeasurementSpec.parseFrom(measurement.measurementSpec.data)
       val encryptionPrivateKeyHandle =
         encryptionKeyPairStore.getPrivateKeyHandle(
@@ -850,13 +847,22 @@ class MetricsService(
             "Encryption private key not found for the measurement ${measurement.name}."
           }
 
-      return aggregateResults(
-        measurement.resultsList
-          .map {
-            decryptMeasurementResultPair(it, encryptionPrivateKeyHandle, apiAuthenticationKey)
+      val decryptedMeasurementResults: List<Measurement.Result> =
+        measurement.resultsList.map {
+          decryptMeasurementResultPair(it, encryptionPrivateKeyHandle, apiAuthenticationKey)
+        }
+
+      return measurementResult {
+        cmmsMeasurementId = MeasurementKey.fromName(measurement.name)!!.measurementId
+        results +=
+          decryptedMeasurementResults.map {
+            try {
+              it.toInternal(measurement.protocolConfig)
+            } catch (e: Throwable) {
+              failGrpc(Status.UNKNOWN) { e.message ?: "Unable to read measurement result." }
+            }
           }
-          .map(Measurement.Result::toInternal)
-      )
+      }
     }
 
     /** Decrypts a [Measurement.ResultPair] to [Measurement.Result] */
@@ -903,66 +909,6 @@ class MetricsService(
         throw Exception("Measurement result signature is invalid", e)
       }
       return Measurement.Result.parseFrom(signedResult.data)
-    }
-
-    /** Aggregates a list of [InternalMeasurement.Result]s to a [InternalMeasurement.Result] */
-    private fun aggregateResults(
-      internalMeasurementResults: List<InternalMeasurement.Result>
-    ): InternalMeasurement.Result {
-      if (internalMeasurementResults.isEmpty()) {
-        error("No measurement result.")
-      }
-      var reachValue = 0L
-      var impressionValue = 0L
-      val frequencyDistribution = mutableMapOf<Long, Double>()
-      var watchDurationValue = duration {
-        seconds = 0
-        nanos = 0
-      }
-
-      // Aggregation
-      for (result in internalMeasurementResults) {
-        if (result.hasFrequency()) {
-          if (!result.hasReach()) {
-            error("Missing reach measurement in the Reach-Frequency measurement.")
-          }
-          for ((frequency, percentage) in result.frequency.relativeFrequencyDistributionMap) {
-            val previousTotalReachCount =
-              frequencyDistribution.getOrDefault(frequency, 0.0) * reachValue
-            val currentReachCount = percentage * result.reach.value
-            frequencyDistribution[frequency] =
-              (previousTotalReachCount + currentReachCount) / (reachValue + result.reach.value)
-          }
-        }
-        if (result.hasReach()) {
-          reachValue += result.reach.value
-        }
-        if (result.hasImpression()) {
-          impressionValue += result.impression.value
-        }
-        if (result.hasWatchDuration()) {
-          watchDurationValue += result.watchDuration.value
-        }
-      }
-
-      return InternalMeasurementKt.result {
-        if (internalMeasurementResults.first().hasReach()) {
-          this.reach = InternalMeasurementKt.ResultKt.reach { value = reachValue }
-        }
-        if (internalMeasurementResults.first().hasFrequency()) {
-          this.frequency =
-            InternalMeasurementKt.ResultKt.frequency {
-              relativeFrequencyDistribution.putAll(frequencyDistribution)
-            }
-        }
-        if (internalMeasurementResults.first().hasImpression()) {
-          this.impression = InternalMeasurementKt.ResultKt.impression { value = impressionValue }
-        }
-        if (internalMeasurementResults.first().hasWatchDuration()) {
-          this.watchDuration =
-            InternalMeasurementKt.ResultKt.watchDuration { value = watchDurationValue }
-        }
-      }
     }
   }
 
@@ -1080,6 +1026,7 @@ class MetricsService(
         }
     }
   }
+
   override suspend fun listMetrics(request: ListMetricsRequest): ListMetricsResponse {
     val parentKey =
       grpcRequireNotNull(MeasurementConsumerKey.fromName(request.parent)) {
@@ -1527,6 +1474,66 @@ private fun buildMetricResult(metric: InternalMetric): MetricResult {
   }
 }
 
+/** Aggregates a list of [InternalMeasurement.Result]s to a [InternalMeasurement.Result] */
+private fun aggregateResults(
+  internalMeasurementResults: List<InternalMeasurement.Result>
+): InternalMeasurement.Result {
+  if (internalMeasurementResults.isEmpty()) {
+    error("No measurement result.")
+  }
+  var reachValue = 0L
+  var impressionValue = 0L
+  val frequencyDistribution = mutableMapOf<Long, Double>()
+  var watchDurationValue = duration {
+    seconds = 0
+    nanos = 0
+  }
+
+  // Aggregation
+  for (result in internalMeasurementResults) {
+    if (result.hasFrequency()) {
+      if (!result.hasReach()) {
+        error("Missing reach measurement in the Reach-Frequency measurement.")
+      }
+      for ((frequency, percentage) in result.frequency.relativeFrequencyDistributionMap) {
+        val previousTotalReachCount =
+          frequencyDistribution.getOrDefault(frequency, 0.0) * reachValue
+        val currentReachCount = percentage * result.reach.value
+        frequencyDistribution[frequency] =
+          (previousTotalReachCount + currentReachCount) / (reachValue + result.reach.value)
+      }
+    }
+    if (result.hasReach()) {
+      reachValue += result.reach.value
+    }
+    if (result.hasImpression()) {
+      impressionValue += result.impression.value
+    }
+    if (result.hasWatchDuration()) {
+      watchDurationValue += result.watchDuration.value
+    }
+  }
+
+  return InternalMeasurementKt.result {
+    if (internalMeasurementResults.first().hasReach()) {
+      this.reach = InternalMeasurementKt.ResultKt.reach { value = reachValue }
+    }
+    if (internalMeasurementResults.first().hasFrequency()) {
+      this.frequency =
+        InternalMeasurementKt.ResultKt.frequency {
+          relativeFrequencyDistribution.putAll(frequencyDistribution)
+        }
+    }
+    if (internalMeasurementResults.first().hasImpression()) {
+      this.impression = InternalMeasurementKt.ResultKt.impression { value = impressionValue }
+    }
+    if (internalMeasurementResults.first().hasWatchDuration()) {
+      this.watchDuration =
+        InternalMeasurementKt.ResultKt.watchDuration { value = watchDurationValue }
+    }
+  }
+}
+
 /** Calculates the watch duration result by summing up [WeightedMeasurement]s. */
 private fun calculateWatchDurationResults(
   weightedMeasurements: List<WeightedMeasurement>
@@ -1535,11 +1542,12 @@ private fun calculateWatchDurationResults(
     val watchDuration: Duration =
       weightedMeasurements
         .map { weightedMeasurement ->
-          if (!weightedMeasurement.measurement.details.result.hasWatchDuration()) {
+          if (weightedMeasurement.measurement.details.resultsList.any { !it.hasWatchDuration() }) {
             error("Watch duration measurement is missing.")
           }
-          weightedMeasurement.measurement.details.result.watchDuration.value *
-            weightedMeasurement.weight
+          aggregateResults(weightedMeasurement.measurement.details.resultsList)
+            .watchDuration
+            .value * weightedMeasurement.weight
         }
         .reduce { sum, element -> sum + element }
     value = watchDuration.seconds + (watchDuration.nanos.toDouble() / NANOS_PER_SECOND)
@@ -1553,10 +1561,11 @@ private fun calculateImpressionResults(
   return impressionCountResult {
     value =
       weightedMeasurements.sumOf { weightedMeasurement ->
-        if (!weightedMeasurement.measurement.details.result.hasImpression()) {
+        if (weightedMeasurement.measurement.details.resultsList.any { !it.hasImpression() }) {
           error("Impression measurement is missing.")
         }
-        weightedMeasurement.measurement.details.result.impression.value * weightedMeasurement.weight
+        aggregateResults(weightedMeasurement.measurement.details.resultsList).impression.value *
+          weightedMeasurement.weight
       }
   }
 }
@@ -1569,10 +1578,14 @@ private fun calculateFrequencyHistogramResults(
   val aggregatedFrequencyHistogramMap: MutableMap<Long, Double> =
     weightedMeasurements
       .map { weightedMeasurement ->
-        val result = weightedMeasurement.measurement.details.result
-        if (!result.hasFrequency() || !result.hasReach()) {
+        if (
+          weightedMeasurement.measurement.details.resultsList.any {
+            !it.hasReach() || !it.hasFrequency()
+          }
+        ) {
           error("Reach-Frequency measurement is missing.")
         }
+        val result = aggregateResults(weightedMeasurement.measurement.details.resultsList)
         val reach = result.reach.value
         result.frequency.relativeFrequencyDistributionMap.mapValues { (_, rate) ->
           rate * weightedMeasurement.weight * reach
@@ -1613,10 +1626,11 @@ private fun calculateReachResults(
   return reachResult {
     value =
       weightedMeasurements.sumOf { weightedMeasurement ->
-        if (!weightedMeasurement.measurement.details.result.hasReach()) {
+        if (weightedMeasurement.measurement.details.resultsList.any { !it.hasReach() }) {
           error("Reach measurement is missing.")
         }
-        weightedMeasurement.measurement.details.result.reach.value * weightedMeasurement.weight
+        aggregateResults(weightedMeasurement.measurement.details.resultsList).reach.value *
+          weightedMeasurement.weight
       }
   }
 }
--- a/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/ProtoConversions.kt
+++ b/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/ProtoConversions.kt
@@ -19,20 +19,34 @@ package org.wfanet.measurement.reporting.service.api.v2alpha
 import com.google.protobuf.util.Timestamps
 import com.google.type.Interval
 import com.google.type.interval
+import org.wfanet.measurement.api.v2alpha.CustomDirectMethodology
 import org.wfanet.measurement.api.v2alpha.DifferentialPrivacyParams
 import org.wfanet.measurement.api.v2alpha.EventGroupKey as CmmsEventGroupKey
+import org.wfanet.measurement.api.v2alpha.LiquidLegionsCountDistinct
+import org.wfanet.measurement.api.v2alpha.LiquidLegionsDistribution
 import org.wfanet.measurement.api.v2alpha.Measurement
 import org.wfanet.measurement.api.v2alpha.MeasurementConsumerKey
 import org.wfanet.measurement.api.v2alpha.MeasurementSpec
 import org.wfanet.measurement.api.v2alpha.MeasurementSpec.VidSamplingInterval
 import org.wfanet.measurement.api.v2alpha.MeasurementSpecKt
+import org.wfanet.measurement.api.v2alpha.ProtocolConfig
 import org.wfanet.measurement.api.v2alpha.differentialPrivacyParams
 import org.wfanet.measurement.config.reporting.MetricSpecConfig
+import org.wfanet.measurement.internal.reporting.v2.CustomDirectMethodology as InternalCustomDirectMethodology
+import org.wfanet.measurement.internal.reporting.v2.DeterministicCount
+import org.wfanet.measurement.internal.reporting.v2.DeterministicCountDistinct
+import org.wfanet.measurement.internal.reporting.v2.DeterministicDistribution
+import org.wfanet.measurement.internal.reporting.v2.DeterministicSum
+import org.wfanet.measurement.internal.reporting.v2.LiquidLegionsCountDistinct as InternalLiquidLegionsCountDistinct
+import org.wfanet.measurement.internal.reporting.v2.LiquidLegionsDistribution as InternalLiquidLegionsDistribution
+import org.wfanet.measurement.internal.reporting.v2.LiquidLegionsV2
 import org.wfanet.measurement.internal.reporting.v2.Measurement as InternalMeasurement
 import org.wfanet.measurement.internal.reporting.v2.MeasurementKt as InternalMeasurementKt
 import org.wfanet.measurement.internal.reporting.v2.MetricSpec as InternalMetricSpec
 import org.wfanet.measurement.internal.reporting.v2.MetricSpecKt as InternalMetricSpecKt
+import org.wfanet.measurement.internal.reporting.v2.NoiseMechanism
 import org.wfanet.measurement.internal.reporting.v2.PeriodicTimeInterval as InternalPeriodicTimeInterval
+import org.wfanet.measurement.internal.reporting.v2.ReachOnlyLiquidLegionsV2
 import org.wfanet.measurement.internal.reporting.v2.Report as InternalReport
 import org.wfanet.measurement.internal.reporting.v2.ReportingSet as InternalReportingSet
 import org.wfanet.measurement.internal.reporting.v2.StreamMetricsRequest
@@ -42,8 +56,15 @@ import org.wfanet.measurement.internal.reporting.v2.StreamReportingSetsRequestKt
 import org.wfanet.measurement.internal.reporting.v2.StreamReportsRequest
 import org.wfanet.measurement.internal.reporting.v2.StreamReportsRequestKt
 import org.wfanet.measurement.internal.reporting.v2.TimeIntervals as InternalTimeIntervals
+import org.wfanet.measurement.internal.reporting.v2.customDirectMethodology
+import org.wfanet.measurement.internal.reporting.v2.liquidLegionsCountDistinct
+import org.wfanet.measurement.internal.reporting.v2.liquidLegionsDistribution
+import org.wfanet.measurement.internal.reporting.v2.liquidLegionsSketchParams
+import org.wfanet.measurement.internal.reporting.v2.liquidLegionsV2
 import org.wfanet.measurement.internal.reporting.v2.metricSpec as internalMetricSpec
 import org.wfanet.measurement.internal.reporting.v2.periodicTimeInterval as internalPeriodicTimeInterval
+import org.wfanet.measurement.internal.reporting.v2.reachOnlyLiquidLegionsSketchParams
+import org.wfanet.measurement.internal.reporting.v2.reachOnlyLiquidLegionsV2
 import org.wfanet.measurement.internal.reporting.v2.streamMetricsRequest
 import org.wfanet.measurement.internal.reporting.v2.streamReportingSetsRequest
 import org.wfanet.measurement.internal.reporting.v2.streamReportsRequest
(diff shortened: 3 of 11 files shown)


riemanli:
* main
  * **PR #1154** <a href="https://app.graphite.dev/github/pr/world-federation-of-advertisers/cross-media-measurement/1154" target="_blank">Graphite</a> 
    * **PR #1100** <a href="https://app.graphite.dev/github/pr/world-federation-of-advertisers/cross-media-measurement/1100" target="_blank">Graphite</a> 
      * **PR #1107** <a href="https://app.graphite.dev/github/pr/world-federation-of-advertisers/cross-media-measurement/1107" target="_blank">Graphite</a> 
        * **PR #1184** <a href="https://app.graphite.dev/github/pr/world-federation-of-advertisers/cross-media-measurement/1184" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/world-federation-of-advertisers/cross-media-measurement/1184?utm_source=stack-comment).

wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1184)

tristanvuong2021 (approved):
> Reviewed 5 of 22 files at r1, 17 of 17 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1184)* status: all files reviewed, 1 unresolved discussion (waiting on @riemanli)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 858 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1184#-Ne8zOIGAgupgpGELYZ2:-Ne8zOIGAgupgpGELYZ3:beyyc3x) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7398ca8319da2ee5f03bad112b8a437cf496778e/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L858)):*
> ```Kotlin
>           }
> 
>       val decryptMeasurementResults =
> ```

nit: val decryptedMeasurementResults

riemanli (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1184)* status: 21 of 22 files reviewed, 1 unresolved discussion (waiting on @Marco-Premier, @stevenwarejones, and @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 858 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1184#-Ne8zOIGAgupgpGELYZ2:-Ne9hYbh86hjG-Wrwl2b:bfhz2u7) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7398ca8319da2ee5f03bad112b8a437cf496778e/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L858)):*

nit: val decryptedMeasurementResults
</blockquote></details>

Thanks for catching it. Done.

stevenwarejones (changes requested):
> Reviewed 1 of 17 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1184)* status: 21 of 22 files reviewed, 5 unresolved discussions (waiting on @Marco-Premier, @riemanli, and @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 890 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1184#-NeA5Yop8gsXslryRc9q:-NeA5Yop8gsXslryRc9r:b-j4qg4w) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/fe942223a298cc79c28d20ce18d72788657ebcdf/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L890)):*
> ```Kotlin
>                    *   methodology.
>                    */
>                   Measurement.Result.Reach.MethodologyCase.METHODOLOGY_NOT_SET,
> ```

i don't think this should default to deterministic. it just shouldn't compute error bars if not present
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 927 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1184#-NeA5gmo8pcPh5tEUbY9:-NeA5gmo8pcPh5tEUbYA:b-7ropln) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/fe942223a298cc79c28d20ce18d72788657ebcdf/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L927)):*
> ```Kotlin
>                    *   methodology.
>                    */
>                   Measurement.Result.Frequency.MethodologyCase.METHODOLOGY_NOT_SET,
> ```

ditto
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 964 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1184#-NeA5nQ-2iMuRst_lDfD:-NeA5nQ-2iMuRst_lDfE:b-7ropln) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/fe942223a298cc79c28d20ce18d72788657ebcdf/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L964)):*
> ```Kotlin
>                    *   methodology.
>                    */
>                   Measurement.Result.Impression.MethodologyCase.METHODOLOGY_NOT_SET,
> ```

ditto
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 985 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1184#-NeA5oA18POQI3sBSQP0:-NeA5oA18POQI3sBSQP1:b-7ropln) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/fe942223a298cc79c28d20ce18d72788657ebcdf/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L985)):*
> ```Kotlin
>                    *   methodology.
>                    */
>                   Measurement.Result.WatchDuration.MethodologyCase.METHODOLOGY_NOT_SET,
> ```

ditto
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/ProtoConversions.kt` line 662 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1184#-NeA60OI6NWqNpjRH3VK:-NeA60OI6NWqNpjRH3VL:bwd9m1y) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/fe942223a298cc79c28d20ce18d72788657ebcdf/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/ProtoConversions.kt#L662)):*
> ```Kotlin
>     ProtocolConfig.NoiseMechanism.DISCRETE_GAUSSIAN -> NoiseMechanism.DISCRETE_GAUSSIAN
>     /** TODO(@riemanli): once EDPs incorporate the new API, raise an error for unspecified. */
>     ProtocolConfig.NoiseMechanism.NOISE_MECHANISM_UNSPECIFIED,
> ```

we should have an internal unspecified as well that this maps to

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1184)* status: 21 of 22 files reviewed, 5 unresolved discussions (waiting on @Marco-Premier, @riemanli, @stevenwarejones, and @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 890 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1184#-NeA5Yop8gsXslryRc9q:-NeAD66A6_BYzuj0JjZs:b-i0kwvl) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/fe942223a298cc79c28d20ce18d72788657ebcdf/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L890)):*

i don't think this should default to deterministic. it just shouldn't compute error bars if not present
</blockquote></details>

So the point of the API change was to:

1. Make what EDPs should already be doing explicit, i.e. what methodology and noise mechanism every EDP was already assumed to be using.
   * This is why the API change documented what values are assumed if each field is unset.
2. Allow EDPs to choose between different methodologies and noise mechanisms going forward.

Whether the EDP was actually following the assumptions is irrelevant. You'll get the same inaccuracy in variance if the EDP simply states they are using one methodology when they are in fact using another.

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1184)* status: 21 of 22 files reviewed, 5 unresolved discussions (waiting on @Marco-Premier, @riemanli, @SanjayVas, and @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 890 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1184#-NeA5Yop8gsXslryRc9q:-NeJNk464_MKs6gqOpPa:b-7inl1n) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/fe942223a298cc79c28d20ce18d72788657ebcdf/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L890)):*

So the point of the API change was to:

1. Make what EDPs should already be doing explicit, i.e. what methodology and noise mechanism every EDP was already assumed to be using.
   * This is why the API change documented what values are assumed if each field is unset.
2. Allow EDPs to choose between different methodologies and noise mechanisms going forward.

Whether the EDP was actually following the assumptions is irrelevant. You'll get the same inaccuracy in variance if the EDP simply states they are using one methodology when they are in fact using another.
</blockquote></details>

i'm saying we should not be setting the `deterministicCountDistinct` field in this case

tristanvuong2021 (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1184)* status: all files reviewed, 5 unresolved discussions (waiting on @Marco-Premier, @riemanli, and @SanjayVas)

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1184)* status: all files reviewed, 5 unresolved discussions (waiting on @Marco-Premier, @riemanli, and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 890 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1184#-NeA5Yop8gsXslryRc9q:-NeL4CoB8oyXGBdLa1o_:bxqp76r) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/fe942223a298cc79c28d20ce18d72788657ebcdf/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L890)):*
> i'm saying we should not be setting the `deterministicCountDistinct` field in this case

Documentation for the [`methodology` field in `Measurement.Result.Impression`](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/39034b41b30ed85cabf8afe6388bfb956f07af25/src/main/proto/wfa/measurement/api/v2alpha/measurement.proto#L200):

>    // The computation methodology done by data provider.
      //
      // Currently, only direct measurement would populate it. Before all EDPs
      // incorporate the new API, if no methodology is set for the impression
      // result, it is assumed that DeterministicCount is used.

Meaning we've explicitly stated that if the methodology is unspecified it is assumed to be `DeterministicCount`. This, being the only valid methodology, is what all EDPs should have already been doing. It can only possibly be another methodology once we add more values to that oneof.

There is intentionally no way to capture that the EDP used an invalid methodology. We just assume they've been doing it correctly with the understanding that the result (and the corresponding variance) would be inaccurate in that case.

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1184)* status: all files reviewed, 5 unresolved discussions (waiting on @Marco-Premier and @riemanli)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 890 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1184#-NeA5Yop8gsXslryRc9q:-NedYXbxCEHJqLFzfTVo:b-xrp26u) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/fe942223a298cc79c28d20ce18d72788657ebcdf/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L890)):*

> i'm saying we should not be setting the `deterministicCountDistinct` field in this case

Documentation for the [`methodology` field in `Measurement.Result.Impression`](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/39034b41b30ed85cabf8afe6388bfb956f07af25/src/main/proto/wfa/measurement/api/v2alpha/measurement.proto#L200):

>    // The computation methodology done by data provider.
      //
      // Currently, only direct measurement would populate it. Before all EDPs
      // incorporate the new API, if no methodology is set for the impression
      // result, it is assumed that DeterministicCount is used.

Meaning we've explicitly stated that if the methodology is unspecified it is assumed to be `DeterministicCount`. This, being the only valid methodology, is what all EDPs should have already been doing. It can only possibly be another methodology once we add more values to that oneof.

There is intentionally no way to capture that the EDP used an invalid methodology. We just assume they've been doing it correctly with the understanding that the result (and the corresponding variance) would be inaccurate in that case.
</blockquote></details>

as discussed, lets map to a custom methodology

riemanli (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1184)* status: all files reviewed, 5 unresolved discussions (waiting on @kungfucraig, @Marco-Premier, and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/ProtoConversions.kt` line 662 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1184#-NeA60OI6NWqNpjRH3VK:-Nee1_b650TUJdJb1RzN:b-cvouec) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/fe942223a298cc79c28d20ce18d72788657ebcdf/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/ProtoConversions.kt#L662)):*

we should have an internal unspecified as well that this maps to
</blockquote></details>

This is similar to the issue we had for methodology fields. The [document in the current API](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/main/src/main/proto/wfa/measurement/api/v2alpha/measurement.proto#L133-L137) states that if unspecified the noise mechanism is `CONTINUOUS_LAPLACE` by default.

If we don't want this, we can remove that document and let the downstream services decide what to do when the mechanism is unspecified. For reporting server, the conversion method here will map unspecified to unspecified, but later the caller will use `CONTINUOUS_LAPLACE` when unspecified is detected.

Wdyt?

CC @kungfucraig @SanjayVas

kungfucraig (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1184)* status: all files reviewed, 5 unresolved discussions (waiting on @Marco-Premier, @riemanli, and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/ProtoConversions.kt` line 662 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1184#-NeA60OI6NWqNpjRH3VK:-Nee6oVDEfURemCMpXmr:b-nsmwn5) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/fe942223a298cc79c28d20ce18d72788657ebcdf/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/ProtoConversions.kt#L662)):*

This is similar to the issue we had for methodology fields. The [document in the current API](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/main/src/main/proto/wfa/measurement/api/v2alpha/measurement.proto#L133-L137) states that if unspecified the noise mechanism is `CONTINUOUS_LAPLACE` by default.

If we don't want this, we can remove that document and let the downstream services decide what to do when the mechanism is unspecified. For reporting server, the conversion method here will map unspecified to unspecified, but later the caller will use `CONTINUOUS_LAPLACE` when unspecified is detected.

Wdyt?

CC @kungfucraig @SanjayVas
</blockquote></details>

If we pursue the NaN default for the methodologies, then mapping this to UNSPECIFIED would definitely make sense. Basically we're saying, if you don't set this we don't know what you did, and we won't use it for anything (e.g. we will not compute a variance for you). 

Practically speaking, from the point of view of the Measurement Result we'd expect either both or neither of the methodology and nose mechanism to be set.

Anyway, once all EDPs are using this and methodology both then we'll make them both required.

riemanli (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1184)* status: 18 of 22 files reviewed, 5 unresolved discussions (waiting on @Marco-Premier, @SanjayVas, @stevenwarejones, and @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 890 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1184#-NeA5Yop8gsXslryRc9q:-NeeGI-h9wd9eAmpcKbN:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/fe942223a298cc79c28d20ce18d72788657ebcdf/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L890)):*

as discussed, lets map to a custom methodology
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 927 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1184#-NeA5gmo8pcPh5tEUbY9:-NeeGK6M701sm12JCpG8:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/fe942223a298cc79c28d20ce18d72788657ebcdf/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L927)):*

ditto
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 964 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1184#-NeA5nQ-2iMuRst_lDfD:-NeeGKfm1fBABS9cKT9P:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/fe942223a298cc79c28d20ce18d72788657ebcdf/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L964)):*

ditto
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 985 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1184#-NeA5oA18POQI3sBSQP0:-NeeGLBA6ZsbMIUtji6q:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/fe942223a298cc79c28d20ce18d72788657ebcdf/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L985)):*

ditto
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/ProtoConversions.kt` line 662 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1184#-NeA60OI6NWqNpjRH3VK:-NeeEZd7CuPcUMltTzWW:bakqfz4) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/fe942223a298cc79c28d20ce18d72788657ebcdf/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/ProtoConversions.kt#L662)):*

If we pursue the NaN default for the methodologies, then mapping this to UNSPECIFIED would definitely make sense. Basically we're saying, if you don't set this we don't know what you did, and we won't use it for anything (e.g. we will not compute a variance for you). 

Practically speaking, from the point of view of the Measurement Result we'd expect either both or neither of the methodology and nose mechanism to be set.

Anyway, once all EDPs are using this and methodology both then we'll make them both required.
</blockquote></details>

Got it. I will skip the variance calculation if either the noise mechanism or the methodology is unspecified.

Fixed the conversion mapping.

stevenwarejones (approved):
> Reviewed 3 of 22 files at r1, 1 of 17 files at r2, 3 of 4 files at r4, 1 of 1 files at r5.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1184)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Marco-Premier and @SanjayVas)

riemanli (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1184)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Marco-Premier and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 890 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1184#-NeA5Yop8gsXslryRc9q:-NekhNyX4P3Em0hrFpFB:b-3vv0pf) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/fe942223a298cc79c28d20ce18d72788657ebcdf/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L890)):*

Done.
</blockquote></details>

After another round of discussion, we decided to keep unspecified and won't do variance calculation for it.

kungfucraig (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1184)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Marco-Premier and @SanjayVas)

stevenwarejones (approved):
> Reviewed 2 of 3 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1184)* status: 21 of 22 files reviewed, all discussions resolved (waiting on @Marco-Premier, @SanjayVas, and @tristanvuong2021)

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1184)* status: 21 of 22 files reviewed, 1 unresolved discussion (waiting on @Marco-Premier, @riemanli, and @tristanvuong2021)

___
*[`src/main/proto/wfa/measurement/internal/reporting/v2/measurement.proto` line 99 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1184#-NetD6ra4h3ayD4dunO_:-NetD6ra4h3ayD4dunOa:bfmw50o) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5af075ee7b2f523f2faeed647db5856da0c4ca97/src/main/proto/wfa/measurement/internal/reporting/v2/measurement.proto#L99)):*
> ```proto
>     // Computation configs of the `Measurement` derived from the CMMS public
>     // API.
>     repeated ComputationConfig computation_configs = 3;
> ```

I thought the plan was to avoid having to effectively duplicate the protocol config and methodology info here by just computing whatever we needed when we get each Measurement result. Is that not feasible?

kungfucraig (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1184)* status: 21 of 22 files reviewed, 1 unresolved discussion (waiting on @Marco-Premier, @stevenwarejones, and @tristanvuong2021)

___
*[`src/main/proto/wfa/measurement/internal/reporting/v2/measurement.proto` line 99 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1184#-NetD6ra4h3ayD4dunO_:-NetbHNkBpSg4xgn0FEH:bgcmfq5) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5af075ee7b2f523f2faeed647db5856da0c4ca97/src/main/proto/wfa/measurement/internal/reporting/v2/measurement.proto#L99)):*

That is a great question. We did discuss that before. However, I rethought about it and have the following reason to store the info:

Theoretically, what you mentioned is feasible for variance calculation. We can compute and save measurement variances when the measurements are succeeded, and later we use the measurement variances to compute metric variances. However, this approach is under some assumptions:
1. `VidSamplingInterval` of every `Measurement` under the same `Metric` is always the same. For example, we can't allow random start of  `VidSamplingInterval` for measurement specs.
2. Decay rate and sketch size specified in `MetricSpec` and so in `MeasurementSpec` are always honored. EDPs can't use values other than what specified in the spec. 
3. Metric variance calculations won't need any additional information other than measurement variances and the parameters mentioned in (1) and (2).

For (1), I remember we mentioned we might want to do it for better privacy budget consumptions. For (2), @stevenwarejones, @kungfucraig and I have discussed this before in this [thread](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/pull/163#pullrequestreview-1594657860) about the possibility that EDPs won't follow the values in spec.

With those consideration, I decided to store computation configs which leaves us a way if any assumption changes.

In worst case if we really don't need it, remove this line in `details` is simple as there is no DB schema change.
</blockquote></details>

Why not take the other approach, which would be, if we end up needing this add it when we do? You know, YAGNI, and all of that... :)

riemanli (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1184)* status: 21 of 22 files reviewed, 1 unresolved discussion (waiting on @Marco-Premier, @stevenwarejones, and @tristanvuong2021)

___
*[`src/main/proto/wfa/measurement/internal/reporting/v2/measurement.proto` line 99 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1184#-NetD6ra4h3ayD4dunO_:-NetdFw172OpnEPn8R4L:b-l82i1k) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5af075ee7b2f523f2faeed647db5856da0c4ca97/src/main/proto/wfa/measurement/internal/reporting/v2/measurement.proto#L99)):*

Why not take the other approach, which would be, if we end up needing this add it when we do? You know, YAGNI, and all of that... :)
</blockquote></details>

I'm sorry. After inspecting the formulas again, I also found that I really need to know what methodology is used, so I do need to store the methodology info at least.

In this case, do we still want to compute measurement variances upfront? The difference now is storing noise mechanism and parameters of methodologies or not, which I think is pretty minor.

kungfucraig (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1184)* status: 21 of 22 files reviewed, 1 unresolved discussion (waiting on @Marco-Premier, @stevenwarejones, and @tristanvuong2021)

___
*[`src/main/proto/wfa/measurement/internal/reporting/v2/measurement.proto` line 99 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1184#-NetD6ra4h3ayD4dunO_:-NetoFxv95cqzFBPejU_:b-vl55d2) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5af075ee7b2f523f2faeed647db5856da0c4ca97/src/main/proto/wfa/measurement/internal/reporting/v2/measurement.proto#L99)):*

I'm sorry. After inspecting the formulas again, I also found that I really need to know what methodology is used, so I do need to store the methodology info at least.

In this case, do we still want to compute measurement variances upfront? The difference now is storing noise mechanism and parameters of methodologies or not, which I think is pretty minor.
</blockquote></details>

Yeah it's minor. I don't feel strongly about this. If you want to keep it, that's fine.

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1184)* status: 21 of 22 files reviewed, 1 unresolved discussion (waiting on @Marco-Premier, @stevenwarejones, and @tristanvuong2021)

___
*[`src/main/proto/wfa/measurement/internal/reporting/v2/measurement.proto` line 99 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1184#-NetD6ra4h3ayD4dunO_:-NetqF8x70u6nVWXurPY:b3koakx) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5af075ee7b2f523f2faeed647db5856da0c4ca97/src/main/proto/wfa/measurement/internal/reporting/v2/measurement.proto#L99)):*
> the possibility that EDPs won't follow the values in spec.

If we want to allow this then we'd need to make API change (even if it's just documentation). As currently written, EDPs **must** conform to the spec.

> After inspecting the formulas again, I also found that I really need to know what methodology is used, so I do need to store the methodology info at least.

Still confused by this. Why would we need to store it if we compute it as soon as we get the Measurement result? We should have the entire CMMS Measurement at that point.

That all said, I don't have a strong opinion on storing this if there are other benefits. I was just figuring we could avoid having to duplicate information that's already on the CMMS Measurement resource.

riemanli (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1184)* status: 21 of 22 files reviewed, 1 unresolved discussion (waiting on @Marco-Premier, @stevenwarejones, and @tristanvuong2021)

___
*[`src/main/proto/wfa/measurement/internal/reporting/v2/measurement.proto` line 99 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1184#-NetD6ra4h3ayD4dunO_:-Netx2q2ExO45KLqy2F2:b-ypgk5h) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5af075ee7b2f523f2faeed647db5856da0c4ca97/src/main/proto/wfa/measurement/internal/reporting/v2/measurement.proto#L99)):*

> the possibility that EDPs won't follow the values in spec.

If we want to allow this then we'd need to make API change (even if it's just documentation). As currently written, EDPs **must** conform to the spec.

> After inspecting the formulas again, I also found that I really need to know what methodology is used, so I do need to store the methodology info at least.

Still confused by this. Why would we need to store it if we compute it as soon as we get the Measurement result? We should have the entire CMMS Measurement at that point.

That all said, I don't have a strong opinion on storing this if there are other benefits. I was just figuring we could avoid having to duplicate information that's already on the CMMS Measurement resource.
</blockquote></details>

Thanks Sanjay for the feedback and digging into this.

One of the steps to compute metric variances is computing the covariance of every two measurements. As we can see in [the functions of covariances](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/main/src/main/kotlin/org/wfanet/measurement/measurementconsumer/stats/Covariances.kt), depending on the methodologies of the two measurements, we need to use different functions. Of course we can call `getMeasurement`s again later when computing metric variances to get the info from CMMS, but I think saving this to avoid rpcs would be more beneficial.

SanjayVas (changes requested):
> Reviewed 1 of 22 files at r1, 1 of 17 files at r2, 1 of 4 files at r4, 1 of 3 files at r6.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1184)* status: 21 of 22 files reviewed, 3 unresolved discussions (waiting on @Marco-Premier, @riemanli, @stevenwarejones, and @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 859 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1184#-Neu4vc4EIGSHRpRBDQ6:-Neu4vc4EIGSHRpRBDQ7:blpvf2g) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5af075ee7b2f523f2faeed647db5856da0c4ca97/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L859)):*
> ```Kotlin
>           }
> 
>       val decryptedMeasurementResults =
> ```

nit: explicit types where not obvious
___
*[`src/main/proto/wfa/measurement/internal/reporting/v2/measurement.proto` line 99 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1184#-NetD6ra4h3ayD4dunO_:-Neu4d8c0K9I_7-X8336:b-3r22r2) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5af075ee7b2f523f2faeed647db5856da0c4ca97/src/main/proto/wfa/measurement/internal/reporting/v2/measurement.proto#L99)):*

Thanks Sanjay for the feedback and digging into this.

One of the steps to compute metric variances is computing the covariance of every two measurements. As we can see in [the functions of covariances](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/main/src/main/kotlin/org/wfanet/measurement/measurementconsumer/stats/Covariances.kt), depending on the methodologies of the two measurements, we need to use different functions. Of course we can call `getMeasurement`s again later when computing metric variances to get the info from CMMS, but I think saving this to avoid rpcs would be more beneficial.
</blockquote></details>

Sounds like this qualifies under "if there are other benefits" then :)
___
*[`src/main/proto/wfa/measurement/internal/reporting/v2/measurement.proto` line 99 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1184#-Neu5s6OAEaI5rW1Mxml:-Neu5s6OAEaI5rW1Mxmm:b6pqrzf) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5af075ee7b2f523f2faeed647db5856da0c4ca97/src/main/proto/wfa/measurement/internal/reporting/v2/measurement.proto#L99)):*
> ```proto
>     // Computation configs of the `Measurement` derived from the CMMS public
>     // API.
>     repeated ComputationConfig computation_configs = 3;
> ```

Why is this repeated? Is this to handle the case where we're combining multiple Direct results, e.g. for impressions? If so, how do we know which config each result is mapped to? Should we maybe avoid combining the results in this internal message so that we can put the config alongside the result?
___
*[`src/main/proto/wfa/measurement/internal/reporting/v2/measurement.proto` line 104 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1184#-Neu5fy766VY7dpXWqWj:-Neu5fy766VY7dpXWqWk:bf61x74) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5af075ee7b2f523f2faeed647db5856da0c4ca97/src/main/proto/wfa/measurement/internal/reporting/v2/measurement.proto#L104)):*
> ```proto
> }
> 
> // The mechanism used to generate noise in computations.
> ```

nit: mind pulling these out into a separate file or two?

riemanli (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1184)* status: 14 of 23 files reviewed, 1 unresolved discussion (waiting on @Marco-Premier, @SanjayVas, @stevenwarejones, and @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 859 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1184#-Neu4vc4EIGSHRpRBDQ6:-NeuA7pfAYgo62ITNjvz:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5af075ee7b2f523f2faeed647db5856da0c4ca97/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L859)):*

nit: explicit types where not obvious
</blockquote></details>

Done.
___
*[`src/main/proto/wfa/measurement/internal/reporting/v2/measurement.proto` line 99 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1184#-Neu5s6OAEaI5rW1Mxml:-NeuISW49mdFLLcVfVkG:b-39jw80) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5af075ee7b2f523f2faeed647db5856da0c4ca97/src/main/proto/wfa/measurement/internal/reporting/v2/measurement.proto#L99)):*

Why is this repeated? Is this to handle the case where we're combining multiple Direct results, e.g. for impressions? If so, how do we know which config each result is mapped to? Should we maybe avoid combining the results in this internal message so that we can put the config alongside the result?
</blockquote></details>

Thanks for pointing it out! It's an oversight.

I fixed it as you suggested by refactoring the internal reporting measurement to store a list of `Result`s. In addition, the structure of  the internal `Result` is modified to be the same as the one in the Kingdom.
___
*[`src/main/proto/wfa/measurement/internal/reporting/v2/measurement.proto` line 104 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1184#-Neu5fy766VY7dpXWqWj:-NeuIR2sCI2PMZl-4Gy4:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5af075ee7b2f523f2faeed647db5856da0c4ca97/src/main/proto/wfa/measurement/internal/reporting/v2/measurement.proto#L104)):*

nit: mind pulling these out into a separate file or two?
</blockquote></details>

Done.

tristanvuong2021 (approved):
> Reviewed 1 of 4 files at r4, 9 of 9 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1184)* status: all files reviewed, 2 unresolved discussions (waiting on @Marco-Premier, @riemanli, @SanjayVas, and @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/internal/reporting/v2/noise_mechanism.proto` line 36 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1184#-NfCaeyx3GV-xKvtQYsC:-NfCaeyx3GV-xKvtQYsD:b-730rjm) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a3cd3732732a71436d04ebf7ffd837de878995fe/src/main/proto/wfa/measurement/internal/reporting/v2/noise_mechanism.proto#L36)):*
> ```proto
>   // Continuous Gaussian.
>   CONTINUOUS_GAUSSIAN = 5;
> }
> ```

nit: new line

kungfucraig (commented):
> Reviewed 1 of 22 files at r1, 2 of 17 files at r2, 3 of 9 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1184)* status: all files reviewed, 2 unresolved discussions (waiting on @Marco-Premier, @riemanli, @SanjayVas, and @stevenwarejones)

SanjayVas (approved):
> Reviewed 12 of 17 files at r2, 9 of 9 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1184)* status: all files reviewed, 2 unresolved discussions (waiting on @Marco-Premier, @riemanli, and @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/internal/reporting/v2/measurement.proto` line 128 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1184#-NfCqthe2F2P-Bid8Lxj:-NfCqthf4PcqeWCwqBpq:b-eu4gx7) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a3cd3732732a71436d04ebf7ffd837de878995fe/src/main/proto/wfa/measurement/internal/reporting/v2/measurement.proto#L128)):*
> ```proto
>     //
>     // For reach and frequency, there is only one result. For impression and
>     // watch duration, the result is per data provider.
> ```

nit: this is technically dependent on the protocol rather than the measurement type. LLv2 will give a single result, whereas Direct will give one per DataProvider.



_Code quote:_
```proto
    // For reach and frequency, there is only one result. For impression and
    // watch duration, the result is per data provider.
```

riemanli (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1184)* status: 21 of 23 files reviewed, all discussions resolved (waiting on @kungfucraig, @Marco-Premier, @SanjayVas, @stevenwarejones, and @tristanvuong2021)

___
*[`src/main/proto/wfa/measurement/internal/reporting/v2/measurement.proto` line 128 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1184#-NfCqthe2F2P-Bid8Lxj:-NfDU4kZ-rhtb8Va1D4b:b-hsmsbi) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a3cd3732732a71436d04ebf7ffd837de878995fe/src/main/proto/wfa/measurement/internal/reporting/v2/measurement.proto#L128)):*

nit: this is technically dependent on the protocol rather than the measurement type. LLv2 will give a single result, whereas Direct will give one per DataProvider.
</blockquote></details>

Done by explicitly listing the result types. 
___
*[`src/main/proto/wfa/measurement/internal/reporting/v2/noise_mechanism.proto` line 36 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1184#-NfCaeyx3GV-xKvtQYsC:-NfCdw_w2FDP-dbIdgN4:b-896fja) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a3cd3732732a71436d04ebf7ffd837de878995fe/src/main/proto/wfa/measurement/internal/reporting/v2/noise_mechanism.proto#L36)):*

nit: new line
</blockquote></details>

Done!

SanjayVas (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1184)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @kungfucraig, @Marco-Premier, and @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/internal/reporting/v2/measurement.proto` line 128 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1184#-NfCqthe2F2P-Bid8Lxj:-NfDcNI1BYVUkvx6yBlh:b-8zf5e2) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a3cd3732732a71436d04ebf7ffd837de878995fe/src/main/proto/wfa/measurement/internal/reporting/v2/measurement.proto#L128)):*

Done by explicitly listing the result types.
</blockquote></details>

I think it's better not to mention measurement types at all and just stick to mentioning the protocols (LLv2 vs. Direct). Note that "one per DataProvider" is still true for Direct reach or R/F currently; it just happens to be that the Kingdom only uses that protocol when there's a single DataProvider.

riemanli (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1184)* status: 22 of 23 files reviewed, all discussions resolved (waiting on @kungfucraig, @Marco-Premier, @SanjayVas, and @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/internal/reporting/v2/measurement.proto` line 128 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1184#-NfCqthe2F2P-Bid8Lxj:-NfDiSZX6g79N-bSWDqH:b-896fja) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a3cd3732732a71436d04ebf7ffd837de878995fe/src/main/proto/wfa/measurement/internal/reporting/v2/measurement.proto#L128)):*

I think it's better not to mention measurement types at all and just stick to mentioning the protocols (LLv2 vs. Direct). Note that "one per DataProvider" is still true for Direct reach or R/F currently; it just happens to be that the Kingdom only uses that protocol when there's a single DataProvider.
</blockquote></details>

Done!

SanjayVas (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1184)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @kungfucraig, @Marco-Premier, and @stevenwarejones)

stevenwarejones (approved):
> Reviewed 12 of 17 files at r2, 7 of 9 files at r7, 1 of 2 files at r8, 1 of 1 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1184)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @kungfucraig and @Marco-Premier)

PR #1188, opened 2023-08-29
Command Line Tool Supports Creation of Reach-Only Measurements.

renjiezh:


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1188)

SanjayVas (changes requested):
> Reviewed 1 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1188)* status: 1 of 3 files reviewed, 2 unresolved discussions (waiting on @renjiezh)

___
*[`src/main/kotlin/org/wfanet/measurement/api/v2alpha/tools/MeasurementSystem.kt` line 653 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1188#-Nd7ZLZW3ub0hSJBrcve:-Nd7ZLZX-TyYuNGs7vnw:b-knp8ij) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a5e332b9e97e800338f7c4d18e4c865baa33d7d3/src/main/kotlin/org/wfanet/measurement/api/v2alpha/tools/MeasurementSystem.kt#L653)):*
> ```Kotlin
>     class ReachOnlyParams {
>       @Option(
>         names = ["--reach-only"],
> ```

We can just reference "reach" here, as that's the underlying Measurement type. The "reach only" distinction is only to differentiate the new protocol from the existing R/F LLv2 protocol.



_Suggestion:_
```Kotlin
--reach
```
___
*[`src/main/kotlin/org/wfanet/measurement/api/v2alpha/tools/MeasurementSystem.kt` line 659 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1188#-Nd7_9QYCS5zg7A5qivO:-Nd7_9QYCS5zg7A5qivP:bl3cp63) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a5e332b9e97e800338f7c4d18e4c865baa33d7d3/src/main/kotlin/org/wfanet/measurement/api/v2alpha/tools/MeasurementSystem.kt#L659)):*
> ```Kotlin
>       var selected = false
> 
>       @set:Option(
> ```

What happens if you factor these options into another class and use `@Mixin` to put it inside of both `ReachParms` and `ReachAndFrequencyParams`?

renjiezh:
Replaced by https://github.com/world-federation-of-advertisers/cross-media-measurement/pull/1254

PR #1190, opened 2023-08-29
Stop defaulting imagePullPolicy on K8s Jobs to `always`.

SanjayVas:
This should prevent some unnecessary pulls when the image is unchanged.


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1190)

PR #1192, opened 2023-08-30
Update cross-media-measurement-api for max watch duration.

SanjayVas:
Update the cross-media-measurement-api dependency for world-federation-of-advertisers/cross-media-measurement-api#167.

PR #1197, opened 2023-08-31
Use participant count instead of gRPC stubs count in LLv2 mills.

### LINKED ISSUE #1194: LLv2 Mill incorrectly uses number of worker gRPC stubs to compute contributor count
The number of Duchies that are involved in a computation may be a subset of the Duchies that a given Mill has gRPC stubs for. This is because the Kingdom selects the Duchy computation participants.

The Liquid Legions v2 Mill should instead use the count of the `computation_participants` field in the `Computation` resource of the system API.

https://rally1.rallydev.com/#/?detail=/defect/720940790287&fdp=true

SanjayVas:
Fixes #1194


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1197)

SanjayVas:
Depends on #1192

Marco-Premier (approved):
> Reviewed 3 of 7 files at r1, 4 of 4 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1197)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @SanjayVas)

PR #1229, opened 2023-09-19
Fix internal GetMeasurementConsumer possibly returning incorrect Certificate

uakyol:


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1229)

SanjayVas:
I have #1230, which include more preventatives/testing.

PR #1232, opened 2023-09-19
Add new config to reporting workflow

tristanvuong2021:


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1232)

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1232)* status: all files reviewed, 1 unresolved discussion (waiting on @tristanvuong2021)

___
*[`.github/workflows/configure-reporting.yml` line 131 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1232#-NejJlhZ46u4PxmIPPc1:-NejJlhZ46u4PxmIPPc2:ba1v72r) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6ba30ff90df641f23eccd20388b31758a9a4dc6a/.github/workflows/configure-reporting.yml#L131)):*
> ```YAML
>     - name: Write measurement spec config
>       env:
>         MEASUREMENT_SPEC_CONFIG: ${{ vars.MEASUREMENT_SPEC_CONFIG }}
> ```

I don't think we need to write this from an environment var since it doesn't actually change per environment. Just write a static config.

PR #1238, opened 2023-09-20
Fix Mill infinite retry if Computation is already failed.

SanjayVas:


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1238)

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1238)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/MillBase.kt` line 259 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1238#-NeryQA8Do5pgoIG0zEs:-NeryQA8Do5pgoIG0zEt:biw6ph5) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ae7695532b17c153eadf22402ef660eec86fb93f/src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/MillBase.kt#L259)):*
> ```Kotlin
>     } catch (e: StatusException) {
>       when (e.status.code) {
>         Status.Code.FAILED_PRECONDITION -> {
> ```

can you add documentation somewhere what Grpc errors we expect to throw under which conditions?

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1238)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/MillBase.kt` line 259 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1238#-NeryQA8Do5pgoIG0zEs:-Nesbnb82v6GETktrl8E:b-wjsaq7) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ae7695532b17c153eadf22402ef660eec86fb93f/src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/MillBase.kt#L259)):*

can you add documentation somewhere what Grpc errors we expect to throw under which conditions?
</blockquote></details>

We're just following the [gRPC docs](https://github.com/grpc/grpc/blob/master/doc/statuscodes.md) right now. The TODO here is because we're making an assumption on the cause of the error, as opposed to passing error details which can be used to get the additional detail as described in [AIP-193](https://google.aip.dev/193). Once we're passing those additional details, then we'll of course have that information documented.

stevenwarejones (dismissed):
can't errors like this pop-up if the certificates are not set up correctly? should we really be doing infinite retry in that case?

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1238#-:-NfH2xaiEHhCoN0Yw9Rz:b-xcc6xq)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @SanjayVas)

SanjayVas (commented):
I'm not sure what you mean. This PR is *fixing* a case where we were getting stuck in a retry loop when the Computation is already marked as FAILED in the Kingdom.

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1238#-:-NfJ0A26EDTW8o8ESbWX:b-q99vmy)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @SanjayVas)

PR #1244, opened 2023-09-21
s3 beam cleanup

stevenwarejones:
The previous code passed sensitive credentials to beam via pipelineOptions
It also would not have worked if someone had wanted to connect to multiple s3 buckets from beam from Google DataFlow

### DIFF
--- a/src/main/kotlin/org/wfanet/panelmatch/client/deploy/ExchangeWorkflowDaemon.kt
+++ b/src/main/kotlin/org/wfanet/panelmatch/client/deploy/ExchangeWorkflowDaemon.kt
@@ -35,6 +35,11 @@ import org.wfanet.panelmatch.common.Timeout
 import org.wfanet.panelmatch.common.certificates.CertificateManager
 import org.wfanet.panelmatch.common.secrets.SecretMap
 import org.wfanet.panelmatch.common.storage.StorageFactory
+import software.amazon.awssdk.auth.credentials.AwsCredentialsProvider
+import software.amazon.awssdk.auth.credentials.AwsSessionCredentials
+import software.amazon.awssdk.auth.credentials.DefaultCredentialsProvider
+import software.amazon.awssdk.services.sts.auth.StsAssumeRoleCredentialsProvider
+import software.amazon.awssdk.services.sts.model.AssumeRoleRequest
 
 /** Runs ExchangeWorkflows. */
 abstract class ExchangeWorkflowDaemon : Runnable {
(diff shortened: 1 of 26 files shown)


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1244)

jonmolle (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1244)* status: 0 of 26 files reviewed, 1 unresolved discussion (waiting on @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/panelmatch/client/deploy/ExchangeWorkflowDaemon.kt` line 98 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1244#-NfHYqa36ZDHl2_psGgY:-NfHYqa36ZDHl2_psGgZ:b-rmjv3j) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/97c72a8ee96b1e550b3738944d119c115bcebf5d/src/main/kotlin/org/wfanet/panelmatch/client/deploy/ExchangeWorkflowDaemon.kt#L98)):*
> ```Kotlin
> 
>   /** A map of [RecurringExchange] to [AwsSessionCredentials] for use by beam. */
>   protected val awsSessionsCredentialsProvider: Map<String, AwsSessionCredentials> by lazy {
> ```

This is more of a question than anything specific issue with your code:

Why have this here rather than in the S3StorageFactory? I could see that you might want to have access to some other AWS resources, but this is specifically tagged as "S3". I don't see anything here that prevents you from using either the AWS equivalent of Workload Identity Federation (adding a GCP service account as a web identity for an IAM role) or at least assuming the role using the default credentials supplied in the environment if you choose to go that way.

My thoughts are twofold, and if I'm missing something please let me know, I'm happy to be convinced:
1. I want to avoid GCP and AWS-specific code in this file. Mostly as a way to reduce the required dependency graph. I know if you're using both that's impossible, but keeping aws and gcp sdks out of this file seems like a good idea.
2. Does still work if you need multiple AWS credentials in the same exchange? It looks like you're stuck with one set of credentials per recurring exchange.
3. It looks like GCP and AWS flows (assuming you want to avoid storing credentials) are different, so separating them might make debugging and coding easier.

wdyt? I'll try to write a bit of an example on the GCP side as I think the AWS side is covered by your previous PR adding the assumed role to S3. Making more general code might be helpful in case we need to generate credentials for other services aside from S3 going forward, but that's probably overthinking things right now.

stevenwarejones (commented):
do you have bandwidth to test removing the s3frombeam code altogether @jonmolle and testing the new assume role code from data flow accessing s3?

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1244#-:-NfWOt-l1z1SZzPVZOL7:bkb6fbu)* status: 0 of 26 files reviewed, 1 unresolved discussion (waiting on @jonmolle)

PR #1257, opened 2023-09-27
Fix the bug in the llv2 setup phase that causes the aggregator to call DestroyKeysAndCounts when max frequency is one. Add unit test for llv2 when the max frequency is one.

ple13:


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1257)

renjiezh (commented):
Could you add a comment in CompleteSetupPhase in .cc file around `request.maximum_frequency() == 1` to explain the new (but correct) behavior.

> Reviewed 8 of 12 files at r1, 2 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1257#-:-NfKgRFo-87UdHeR8fXZ:b-9mtiyb)* status: 10 of 12 files reviewed, all discussions resolved (waiting on @kungfucraig and @SanjayVas)

kungfucraig (changes requested):
> Reviewed 9 of 12 files at r1, 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1257)* status: all files reviewed, 2 unresolved discussions (waiting on @ple13 and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/liquidlegionsv2/LiquidLegionsV2Mill.kt` line 961 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1257#-NfM7VbdAFHjeSlVClGn:-NfM7VbdAFHjeSlVClGo:b-qztxtp) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a227171b2903266264e0a61a1c72855b4b178132/src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/liquidlegionsv2/LiquidLegionsV2Mill.kt#L961)):*
> ```Kotlin
>     totalRequisitionsCount: Int,
>     participantCount: Int,
>   ): CompleteSetupPhaseRequest {
> ```

Can this method call the aggregator version and then just set the max frequency?  It would save some duplication.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/liquidlegionsv2/LiquidLegionsV2Mill.kt` line 987 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1257#-NfM7KQ4CruQbpF9MIFU:-NfM7KQ4CruQbpF9MIFV:b-9le132) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a227171b2903266264e0a61a1c72855b4b178132/src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/liquidlegionsv2/LiquidLegionsV2Mill.kt#L987)):*
> ```Kotlin
>     llv2Details: LiquidLegionsSketchAggregationV2.ComputationDetails,
>     totalRequisitionsCount: Int,
>     participantCount: Int,
> ```

Can you note in a comment here that the main difference between this and the previous method is that you do not set max frequency and that this causes a different behavior in the crypto code?

SanjayVas (approved):
> Reviewed 9 of 12 files at r1, 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1257)* status: all files reviewed, 3 unresolved discussions (waiting on @ple13)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/liquidlegionsv2/LiquidLegionsV2Mill.kt` line 957 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1257#-NfMizU49_tWatihLSCe:-NfMizU49_tWatihLSCf:b-rx53p4) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a227171b2903266264e0a61a1c72855b4b178132/src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/liquidlegionsv2/LiquidLegionsV2Mill.kt#L957)):*
As an aside (don't need to fix in this PR), these conversion extensions go against [our style guide](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/main/docs/code-style.md#kotlin):

> -   Avoid defining extensions on common types when it may appear that they could work on all instances of that type.
>       -   For example, `String.toFoo()` where the String must be a specific serialization of a `Foo` to work. This is version of the "stringly typed" anti-pattern.

It would be better to have a `buildCompleteSetupPhaseRequest` that took the combined register vector (it looks like that's what the receiver param is) as a parameter.

ple13 (commented):
Done.

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1257#-:-NfNnZR63z2zuQVaiR6W:b-896fix)* status: 7 of 12 files reviewed, 2 unresolved discussions (waiting on @kungfucraig, @renjiezh, and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/liquidlegionsv2/LiquidLegionsV2Mill.kt` line 957 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1257#-NfMizU49_tWatihLSCe:-NfNlYBX2En_LZs-QGcX:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a227171b2903266264e0a61a1c72855b4b178132/src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/liquidlegionsv2/LiquidLegionsV2Mill.kt#L957)):*

As an aside (don't need to fix in this PR), these conversion extensions go against [our style guide](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/main/docs/code-style.md#kotlin):

> -   Avoid defining extensions on common types when it may appear that they could work on all instances of that type.
>       -   For example, `String.toFoo()` where the String must be a specific serialization of a `Foo` to work. This is version of the "stringly typed" anti-pattern.

It would be better to have a `buildCompleteSetupPhaseRequest` that took the combined register vector (it looks like that's what the receiver param is) as a parameter.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/liquidlegionsv2/LiquidLegionsV2Mill.kt` line 961 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1257#-NfM7VbdAFHjeSlVClGn:-NfNlZwhE9OYQib2EfUo:b-rlfm2j) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a227171b2903266264e0a61a1c72855b4b178132/src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/liquidlegionsv2/LiquidLegionsV2Mill.kt#L961)):*

Can this method call the aggregator version and then just set the max frequency?  It would save some duplication.
</blockquote></details>

I've updated the logic. Now we don't need two different functions any more.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/liquidlegionsv2/LiquidLegionsV2Mill.kt` line 987 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1257#-NfM7KQ4CruQbpF9MIFU:-NfNlmIu18k6Q-1scz0v:btjqlcs) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a227171b2903266264e0a61a1c72855b4b178132/src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/liquidlegionsv2/LiquidLegionsV2Mill.kt#L987)):*

Can you note in a comment here that the main difference between this and the previous method is that you do not set max frequency and that this causes a different behavior in the crypto code?
</blockquote></details>

After our discussion, the frequency will be set for all duchies.

renjiezh (approved):
> Reviewed 5 of 5 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1257)* status: all files reviewed, 2 unresolved discussions (waiting on @kungfucraig)

kungfucraig (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1257)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @ple13)

SanjayVas (approved):
> Reviewed 5 of 5 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1257)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @ple13)

PR #1270, opened 2023-10-03
Change List Reports Response order to create_time desc

tristanvuong2021:


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1270)

tristanvuong2021 (commented):
This makes it easier to get the latest Reports. It is possible to use the CEL filter to get Reports created after a certain time, but that method will not work as well when there are a lot of Reports.

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1270#-:-NfrJkiR0NMI9O_NLAYM:bsgujeq)* status: 0 of 4 files reviewed, all discussions resolved

tristanvuong2021 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1270)* status: 3 of 4 files reviewed, 2 unresolved discussions (waiting on @riemanli)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/internal/testing/v2/ReportsServiceTest.kt` line 1274 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1270#-NfsINQ30pjeOV8FstND:-NfsMSPd0Fjo5Hh_U5T8:bipyndn) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/700f5608dfd97d6bb4dd50cf91d840b1c0fd68e9/src/main/kotlin/org/wfanet/measurement/reporting/service/internal/testing/v2/ReportsServiceTest.kt#L1274)):*

Why is `createdReport2` included? Isn't the filter exclusive?
</blockquote></details>

The filter is meant to represent a specific Report so any Reports that follow based on the sorting order are returned. In this case, the external id in the filter is "" so any Report with the same create_time, but a non-blank external id will be returned. In practice, it is highly unlikely any two Reports will have the exact same create_time, but this is just in case it somehow occurs.
___
*[`src/main/proto/wfa/measurement/internal/reporting/v2/reports_service.proto` line 48 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1270#-NfsI66JDiJTnarj3fIC:-NfsM2948p7JOZn_zdVL:bjeu03v) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/700f5608dfd97d6bb4dd50cf91d840b1c0fd68e9/src/main/proto/wfa/measurement/internal/reporting/v2/reports_service.proto#L48)):*

Should the name become BeforeFilter?
</blockquote></details>

The intention is that based on the sorting order, Reports that follow the Report represented in the AfterFilter are returned so After is still appropriate.

riemanli (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1270)* status: 3 of 4 files reviewed, all discussions resolved

Marco-Premier (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1270)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @tristanvuong2021)

PR #1271, opened 2023-10-03
Refactoring the LiquidLegionsV2Mill and ReachOnlyLiquidLegionsV2Mill

ple13:
Currently, the two classes LiquidLegionsV2Mill and ReachOnlyLiquidLegionsV2Mill share some common code (e.g. the functions verifyEdpParticipation, verifyDuchySignature, failComputationAtConfirmationPhase, sendResultToKingdom, nextDuchyStub, aggregatorDuchyStub, ComputationToken.participantCount). A base class, called LiquidLegionsV2MillBase, will be created to own these common pieces.

On top of the above functions, the two mills behave similarly in the initialization and confirmation phase. On the surface, it makes sense to also unify these behaviors inside the LiquidLegionsV2MillBase class. However, the effort to achieve this goal is high. As the protobufs, crypto code, and kotlin code of the two protocols are almost separated, having an abstract function for initialization/confirmation phase would require unifying the protobufs, crypto code, and kotlin code.

The tests for the two protocols remain unchanged to reflect the existing implementation.


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1271)

SanjayVas (commented):
> Reviewed 1 of 10 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1271)* status: 1 of 10 files reviewed, 1 unresolved discussion (waiting on @kungfucraig, @ple13, and @renjiezh)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/liquidlegionsv2/LiquidLegionsV2MillBase.kt` line 75 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1271#-NfvqZFL7hwIxlv4T3yd:-NfvqZFL7hwIxlv4T3ye:b-fz9mto) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8bf1de7694b9ba02be8701835a3825fa0e70264c/src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/liquidlegionsv2/LiquidLegionsV2MillBase.kt#L75)):*
> ```Kotlin
>  *   computationControlClients, used for passing computation to other duchies.
>  */
> abstract class LiquidLegionsV2MillBase(
> ```

nit: instead of calling this `LiquidLegionsV2MillBase`, call it `LiquidLegionsV2Mill` and rename the one for the R/F protocol `ReachFrequencyLiquidLegionsV2Mill`. This gives us parallelism in naming relative to `ReachOnlyLiquidLegionsV2Mill`.



_Code quote:_
```Kotlin
LiquidLegionsV2MillBase
```

SanjayVas (approved):
> Reviewed 9 of 10 files at r1.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1271)* status: all files reviewed, 1 unresolved discussion (waiting on @ple13 and @renjiezh)

renjiezh (changes requested):
> Reviewed 10 of 10 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1271)* status: all files reviewed, 2 unresolved discussions (waiting on @ple13)

___
*[`src/main/proto/wfa/measurement/internal/duchy/protocol/reach_only_liquid_legions_sketch_aggregation_v2.proto` line 113 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1271#-Nfvurz788tZfV2t20oH:-Nfvurz788tZfV2t20oI:b-4a49r3) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c98743fc690ec1e50f577945ab0de64ca85951c5/src/main/proto/wfa/measurement/internal/duchy/protocol/reach_only_liquid_legions_sketch_aggregation_v2.proto#L113)):*
> ```proto
>     Parameters parameters = 2;
> 
>     message ComputationParticipant {
> ```

Comparing Llv2.ComputationDetails with RoLlv2.ComputationDetails, we can see the only difference is the extra field `max_frequency` for Llv2. I wonder if it is possible to subtract the ComputationDetails into a standalone .proto(not necessary, but the purpose is to share the same class) and keep field `max_frequency` as optional. In this way, both protocols are using the same ComputationDetails. A good comment around `max_frequency` is needed to avoid trouble.
WDYT?

ple13 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1271)* status: all files reviewed, 2 unresolved discussions (waiting on @renjiezh and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/liquidlegionsv2/LiquidLegionsV2MillBase.kt` line 75 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1271#-NfvqZFL7hwIxlv4T3yd:-Ng5ODTx6CWFaPxJWkVy:b-6jysgr) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8bf1de7694b9ba02be8701835a3825fa0e70264c/src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/liquidlegionsv2/LiquidLegionsV2MillBase.kt#L75)):*

nit: instead of calling this `LiquidLegionsV2MillBase`, call it `LiquidLegionsV2Mill` and rename the one for the R/F protocol `ReachFrequencyLiquidLegionsV2Mill`. This gives us parallelism in naming relative to `ReachOnlyLiquidLegionsV2Mill`.
</blockquote></details>

We currently use the `llv2` prefix for the LiquidLegionsV2 protocol. I think we should change the llv2 prefix to rfllv2 as well. WDYT?
___
*[`src/main/proto/wfa/measurement/internal/duchy/protocol/reach_only_liquid_legions_sketch_aggregation_v2.proto` line 113 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1271#-Nfvurz788tZfV2t20oH:-Ng5Pa1n1FDYEvSdAAH1:b29bgs2) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c98743fc690ec1e50f577945ab0de64ca85951c5/src/main/proto/wfa/measurement/internal/duchy/protocol/reach_only_liquid_legions_sketch_aggregation_v2.proto#L113)):*

Comparing Llv2.ComputationDetails with RoLlv2.ComputationDetails, we can see the only difference is the extra field `max_frequency` for Llv2. I wonder if it is possible to subtract the ComputationDetails into a standalone .proto(not necessary, but the purpose is to share the same class) and keep field `max_frequency` as optional. In this way, both protocols are using the same ComputationDetails. A good comment around `max_frequency` is needed to avoid trouble.
WDYT?
</blockquote></details>

This can be done, and we could also do the same for many other things such as llv2 and rollv2 configs. However, I think this should have low priority.

renjiezh (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1271)* status: all files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

ple13 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1271)* status: 6 of 18 files reviewed, 1 unresolved discussion (waiting on @renjiezh and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/liquidlegionsv2/LiquidLegionsV2MillBase.kt` line 75 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1271#-NfvqZFL7hwIxlv4T3yd:-NgKjoKRBDVDJs0T1ZWZ:bkwmuvj) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8bf1de7694b9ba02be8701835a3825fa0e70264c/src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/liquidlegionsv2/LiquidLegionsV2MillBase.kt#L75)):*

> We currently use the `llv2` prefix for the LiquidLegionsV2 protocol.

Where do you mean specifically?

If you're talking about referencing LLv2 in general, I don't think we're entirely consistent: sometimes LLv2 refers to a sketch type, sometimes to the family of protocols (both RO and R/F variants), and sometimes specifically to the original (R/F) protocol. We don't need to change all cases where the protocols are referenced, as some of that would be disruptive to API consumers.
</blockquote></details>

Thanks. I'll only change the LiquidLegionsV2Mill to ReachFrequencyLiquidLegionsV2Mill.

SanjayVas (approved):
> Reviewed 11 of 13 files at r2, 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1271)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @stevenwarejones)

stevenwarejones (approved):
> Reviewed 5 of 10 files at r1, 11 of 13 files at r2, 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1271)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @ple13)

PR #1273, opened 2023-10-04
Marcopremier panelexchange kindtest

Marco-Premier:
This PR does not cover documentation updates and neither:

- FullWithPreprocessingTest
- FullWorkflowTest
- MiniWorkflowTest
- SingleStepTest

It contains DoubleBlindWorkflowTest.
Remaining tests will be added on a following PR.


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1273)

stevenwarejones (changes requested):
> Reviewed 31 of 39 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1273)* status: 31 of 39 files reviewed, 5 unresolved discussions (waiting on @Marco-Premier and @SanjayVas)

___
*[`src/main/k8s/panelmatch/testing/data/BUILD.bazel` line 7 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1273#-Ng-EHZ027wz4SsZSdD_:-Ng-EHZ027wz4SsZSdDa:b-k28z9) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/42b5367737c245b5eaa3ad812dd2e37a34e9ce64/src/main/k8s/panelmatch/testing/data/BUILD.bazel#L7)):*
> ```Starlark
> 
> filegroup(
>     name = "exchange_workflow",
> ```

why not use the existing workflows filegroup?
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/panelmatchresourcesetup/PanelMatchResourceSetup.kt` line 62 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1273#-Ng-EeCb3pbWRtK-gHui:-Ng-EeCb3pbWRtK-gHuj:b-vub9hy) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/42b5367737c245b5eaa3ad812dd2e37a34e9ce64/src/main/kotlin/org/wfanet/measurement/loadtest/panelmatchresourcesetup/PanelMatchResourceSetup.kt#L62)):*
> ```Kotlin
> ) {
> 
>   /*data class DataProviderAndKey(
> ```

i think you meant to delete this
___
*[`src/main/kotlin/org/wfanet/panelmatch/client/storage/forwarded/ForwardedStorageFactory.kt` line 28 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1273#-Ng-Ex0q8BiZ2w0ce3GS:-Ng-Ex0q8BiZ2w0ce3GT:b-kvz4g) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/42b5367737c245b5eaa3ad812dd2e37a34e9ce64/src/main/kotlin/org/wfanet/panelmatch/client/storage/forwarded/ForwardedStorageFactory.kt#L28)):*
> ```Kotlin
> import org.wfanet.panelmatch.common.storage.StorageFactory
> 
> class ForwardedStorageFactory(
> ```

add a kdoc
___
*[`src/main/proto/wfa/measurement/loadtest/panelmatch/resourcesetup/resources.proto` line 19 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1273#-Ng-FA5e9XO93V-iSG6z:-Ng-FA5e9XO93V-iSG7-:b-rtagwg) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/42b5367737c245b5eaa3ad812dd2e37a34e9ce64/src/main/proto/wfa/measurement/loadtest/panelmatch/resourcesetup/resources.proto#L19)):*
> ```proto
> package wfa.measurement.loadtest.panelmatch.resourcesetup;
> 
> // import "wfa/measurement/api/v2alpha/event_annotations.proto";
> ```

delete
___
*[`src/main/proto/wfa/panelmatch/client/storage/storage_details.proto` line 97 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1273#-Ng-FF-93j_0ihBnxYXF:-Ng-FF-93j_0ihBnxYXG:b-bs1h1v) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/42b5367737c245b5eaa3ad812dd2e37a34e9ce64/src/main/proto/wfa/panelmatch/client/storage/storage_details.proto#L97)):*
> ```proto
>   Visibility visibility = 5;
> 
>   message ForwardedStorage {
> ```

This really shouldn't be part of storageDetails. It should be in its own file and only used for loadtest.

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1273)* status: 31 of 39 files reviewed, 9 unresolved discussions (waiting on @Marco-Premier and @SanjayVas)

___
*[`src/test/kotlin/org/wfanet/panelmatch/integration/k8s/AbstractPanelMatchCorrectnessTest.kt` line 89 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1273#-Ng-IPQ5Ch9U_aP-G-iR:-Ng-IPQ5Ch9U_aP-G-iS:bjpxy9i) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/42b5367737c245b5eaa3ad812dd2e37a34e9ce64/src/test/kotlin/org/wfanet/panelmatch/integration/k8s/AbstractPanelMatchCorrectnessTest.kt#L89)):*
> ```Kotlin
> import org.wfanet.panelmatch.common.storage.toByteString
> 
> abstract class AbstractPanelMatchCorrectnessTest {
> ```

`AbstractK8sPanelMatchCorrectnessTest` ?
___
*[`src/test/kotlin/org/wfanet/panelmatch/integration/k8s/AbstractPanelMatchCorrectnessTest.kt` line 95 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1273#-Ng-K-WA5di_1rtjPh1v:-Ng-K-WA5di_1rtjPh1w:b-xh84k9) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/42b5367737c245b5eaa3ad812dd2e37a34e9ce64/src/test/kotlin/org/wfanet/panelmatch/integration/k8s/AbstractPanelMatchCorrectnessTest.kt#L95)):*
> ```Kotlin
>   private lateinit var exchangeKey: ExchangeKey
>   protected abstract val initialDataProviderInputs: Map<String, ByteString>
>   protected abstract val initialModelProviderInputs: Map<String, ByteString>
> ```

you should also pass in expected outputs for both parties that you verify
___
*[`src/test/kotlin/org/wfanet/panelmatch/integration/k8s/AbstractPanelMatchCorrectnessTest.kt` line 175 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1273#-Ng-JELT8XIwb2JMp7pf:-Ng-JELT8XIwb2JMp7pg:b-nwisw9) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/42b5367737c245b5eaa3ad812dd2e37a34e9ce64/src/test/kotlin/org/wfanet/panelmatch/integration/k8s/AbstractPanelMatchCorrectnessTest.kt#L175)):*
> ```Kotlin
>           .also { channels.add(it) }
>           .withDefaultDeadline(DEFAULT_RPC_DEADLINE)
>       val dpForwardedStorage =
> ```

i think storage should be abstract here since someone might want to run an cloud test using s3 gcp or whatever
___
*[`src/test/kotlin/org/wfanet/panelmatch/integration/k8s/DoubleBlindWorkflowTest.kt` line 111 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1273#-Ng-KAG092RXD_Q7ATV8:-Ng-KAG092RXD_Q7ATV9:b-t4d165) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/42b5367737c245b5eaa3ad812dd2e37a34e9ce64/src/test/kotlin/org/wfanet/panelmatch/integration/k8s/DoubleBlindWorkflowTest.kt#L111)):*
> ```Kotlin
> 
>       val result = mpForwardedStorage.getBlob("mp-decrypted-join-keys")
>       JoinKeyAndIdCollection.parseFrom(result?.toByteString())
> ```

see my other comment. outputs' contents should be verified against expected outputs. This may be problematic if any outputs are not-deterministic. But you should be able to avoid that as long as there are no discarded queries.

SanjayVas (changes requested):
> Reviewed 35 of 39 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1273)* status: 38 of 39 files reviewed, 18 unresolved discussions (waiting on @Marco-Premier and @stevenwarejones)

___
*[`.github/workflows/build-test.yml` line 47 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1273#-Ng-y6v8-qa28uMYrSo_:-Ng-y6v8-qa28uMYrSoa:b1v8ecx) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/42b5367737c245b5eaa3ad812dd2e37a34e9ce64/.github/workflows/build-test.yml#L47)):*
> ```YAML
>     needs: wait-for-main
>     env:
>       MEASUREMENT_CLUSTER_LOGS_PATH: measurement-cluster-logs
> ```

There will be one set of cluster logs since you're keeping the cluster up. Stick with the single export and upload.
___
*[`src/main/docker/panel_exchange_client/images.bzl` line 40 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1273#-Ng-yyL9-m9xu_paC2hm:-Ng-yyL9-m9xu_paC2hn:b-9p0iq9) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/42b5367737c245b5eaa3ad812dd2e37a34e9ce64/src/main/docker/panel_exchange_client/images.bzl#L40)):*
> ```Starlark
> ]
> 
> FORWARDED_STORAGE_IMAGES = [
> ```

nit: this is called LOCAL_IMAGES in the other set to parallel GCloud/GKE and AWS/EKS.



_Code quote:_
```Starlark
FORWARDED_STORAGE_IMAGES
```
___
*[`src/main/k8s/local/BUILD.bazel` line 66 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1273#-Ng-zYlh7ByGdfNvX5YQ:-Ng-zYlh7ByGdfNvX5YR:b-mh9u4b) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/42b5367737c245b5eaa3ad812dd2e37a34e9ce64/src/main/k8s/local/BUILD.bazel#L66)):*
> ```Starlark
> 
> cue_dump(
>     name = "panelmatch_storage_emulators",
> ```

Drop the "storage" since this contains a Spanner emulator, which is DB not (blob) storage.



_Suggestion:_
```Starlark
panelmatch_emulators
```
___
*[`src/main/k8s/panelmatch/local/forwarded_storage_mp_exchange_workflow_daemon.cue` line 39 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1273#-Ng0-UlGBEYqn6ZmYWgr:-Ng0-UlGBEYqn6ZmYWgs:bx9zp35) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/42b5367737c245b5eaa3ad812dd2e37a34e9ce64/src/main/k8s/panelmatch/local/forwarded_storage_mp_exchange_workflow_daemon.cue#L39)):*
> ```CUE
> deployments: {
> 	"model-provider-panel-exchange-daemon": {
> 		_jvmFlags:   "-Xmx3584m" // 4GiB - 512MiB overhead.
> ```

The GitHub Actions runner is quite memory-constrained. Do we need this much for local testing?



_Code quote:_
```CUE
"-Xmx3584m" // 4GiB - 512MiB overhead.
```
___
*[`src/main/k8s/panelmatch/testing/secretfiles/edp1_cs_cert.der` line 0 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1273#-Ng00Cyx5MNRTupGQs9A:-Ng00Cyx5MNRTupGQs9B:bohhm63) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/42b5367737c245b5eaa3ad812dd2e37a34e9ce64/src/main/k8s/panelmatch/testing/secretfiles/edp1_cs_cert.der#L0)):*
I can't tell from the Reviewable UI, but this should just be a symlink to `src/main/k8s/testing/secretfiles/edp1_cs_cert.der`
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/panelmatchresourcesetup/InProcessPanelMatchResourceSetup.kt` line 52 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1273#-Ng01Nb18H0JIU05lEeG:-Ng01Nb22RZSVc6PWNEv:bxiob8o) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/42b5367737c245b5eaa3ad812dd2e37a34e9ce64/src/main/kotlin/org/wfanet/measurement/loadtest/panelmatchresourcesetup/InProcessPanelMatchResourceSetup.kt#L52)):*
> ```Kotlin
>  *   in ResourceSetup.
>  */
> class InProcessPanelMatchResourceSetup(
> ```

What's the "InProcess" part about here? This takes in a channel to an existing server, and doesn't appear to care whether it's an in-process server. I don't believe we need a separate "in-process" version of ResourceSetup, so we shouldn't need one for PanelMatchResourceSetup.



_Code quote:_
```Kotlin
InProcessPanelMatchResourceSetup
```
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/panelmatchresourcesetup/Output.kt` line 0 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1273#-Ng032no5udqmNdwjsqD:-Ng032np8W3Ba2f5Ou5A:b-rgts7l) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/42b5367737c245b5eaa3ad812dd2e37a34e9ce64/src/main/kotlin/org/wfanet/measurement/loadtest/panelmatchresourcesetup/Output.kt#L0)):*
This appears to be a direct copy of `src/main/kotlin/org/wfanet/measurement/loadtest/resourcesetup/Output.kt`. We really shouldn't just add another copy.
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/panelmatchresourcesetup/PanelMatchResourceSetup.kt` line 54 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1273#-Ng040I40AZr1MkvlvNp:-Ng040I40AZr1MkvlvNq:b18py7d) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/42b5367737c245b5eaa3ad812dd2e37a34e9ce64/src/main/kotlin/org/wfanet/measurement/loadtest/panelmatchresourcesetup/PanelMatchResourceSetup.kt#L54)):*
> ```Kotlin
> private val API_VERSION = Version.V2_ALPHA
> 
> class PanelMatchResourceSetup(
> ```

See other comment re:reconciling the differences between this and InProcessPanelMatchResourceSetup.
___
*[`src/test/kotlin/org/wfanet/panelmatch/integration/k8s/AbstractPanelMatchCorrectnessTest.kt` line 89 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1273#-Ng-IPQ5Ch9U_aP-G-iR:-Ng04ZWb4SSWSJ6SwLcp:b-ogvozx) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/42b5367737c245b5eaa3ad812dd2e37a34e9ce64/src/test/kotlin/org/wfanet/panelmatch/integration/k8s/AbstractPanelMatchCorrectnessTest.kt#L89)):*

`AbstractK8sPanelMatchCorrectnessTest` ?
</blockquote></details>

I believe this is just copying the naming from `src/test/kotlin/org/wfanet/measurement/integration/k8s/AbstractCorrectnessTest.kt`. It's already in a `k8s` package and shouldn't be visible outside the Bazel package, so no need to include that in the name.
___
*[`src/test/kotlin/org/wfanet/panelmatch/integration/k8s/AbstractPanelMatchCorrectnessTest.kt` line 119 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1273#-Ng0586X98GMx2w2u3Jh:-Ng0586X98GMx2w2u3Ji:b6ubvg7) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/42b5367737c245b5eaa3ad812dd2e37a34e9ce64/src/test/kotlin/org/wfanet/panelmatch/integration/k8s/AbstractPanelMatchCorrectnessTest.kt#L119)):*
> ```Kotlin
>   private val DEFAULT_RPC_DEADLINE = Duration.ofSeconds(30)
> 
>   protected suspend fun runResourceSetup(
> ```

This should only be in the "empty cluster" version of the test, so shouldn't be in the abstract base class. We will need a version of the test that can hit existing deployed instances in a cloud environment, but that can be added in a later PR.

stevenwarejones (commented):
> Reviewed 3 of 39 files at r1.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1273)* status: 38 of 39 files reviewed, 17 unresolved discussions (waiting on @Marco-Premier and @SanjayVas)

Marco-Premier (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1273)* status: 16 of 44 files reviewed, 16 unresolved discussions (waiting on @SanjayVas and @stevenwarejones)

___
*[`.github/workflows/build-test.yml` line 47 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1273#-Ng-y6v8-qa28uMYrSo_:-Ng33xGIAmmMbBkPPTeI:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/42b5367737c245b5eaa3ad812dd2e37a34e9ce64/.github/workflows/build-test.yml#L47)):*

There will be one set of cluster logs since you're keeping the cluster up. Stick with the single export and upload.
</blockquote></details>

Done.
___
*[`src/main/k8s/local/BUILD.bazel` line 66 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1273#-Ng-zYlh7ByGdfNvX5YQ:-Ng352NR4xpVkxJmISdl:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/42b5367737c245b5eaa3ad812dd2e37a34e9ce64/src/main/k8s/local/BUILD.bazel#L66)):*

Drop the "storage" since this contains a Spanner emulator, which is DB not (blob) storage.
</blockquote></details>

Done.
___
*[`src/main/k8s/panelmatch/local/forwarded_storage_mp_exchange_workflow_daemon.cue` line 39 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1273#-Ng0-UlGBEYqn6ZmYWgr:-Ng3ahEc6rnjr1aYMYZG:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/42b5367737c245b5eaa3ad812dd2e37a34e9ce64/src/main/k8s/panelmatch/local/forwarded_storage_mp_exchange_workflow_daemon.cue#L39)):*

The GitHub Actions runner is quite memory-constrained. Do we need this much for local testing?
</blockquote></details>

Done.
___
*[`src/main/k8s/panelmatch/testing/data/BUILD.bazel` line 7 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1273#-Ng-EHZ027wz4SsZSdD_:-Ng35k3H47WJwmc5ykdN:bbyggdo) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/42b5367737c245b5eaa3ad812dd2e37a34e9ce64/src/main/k8s/panelmatch/testing/data/BUILD.bazel#L7)):*

why not use the existing workflows filegroup?
</blockquote></details>

TL;DR -> to make kind test and previous existing test independent by each other



Do you mean the ```config``` filegroup defined inside the test folder:

https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/main/src/test/kotlin/org/wfanet/panelmatch/integration/config/BUILD.bazel

? 

If so, I left that there for the existing tests and then I just followed the same pattern we have for the EmptyClusterCorrectnessTest where data is placed under ```testing/data/``` and referenced by bazel (https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/marcopremier-panelexchange-kindtest/src/test/kotlin/org/wfanet/panelmatch/integration/k8s/BUILD.bazel#L17)
___
*[`src/main/k8s/panelmatch/testing/secretfiles/edp1_cs_cert.der` line  at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1273#-Ng00Cyx5MNRTupGQs9A:-Ng3_2vL3UuXSUrF7E9x:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/42b5367737c245b5eaa3ad812dd2e37a34e9ce64/src/main/k8s/panelmatch/testing/secretfiles/edp1_cs_cert.der#L)):*

I can't tell from the Reviewable UI, but this should just be a symlink to `src/main/k8s/testing/secretfiles/edp1_cs_cert.der`
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/panelmatchresourcesetup/PanelMatchResourceSetup.kt` line 54 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1273#-Ng040I40AZr1MkvlvNp:-NgJEwHn9e7acDAsSsvQ:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/42b5367737c245b5eaa3ad812dd2e37a34e9ce64/src/main/kotlin/org/wfanet/measurement/loadtest/panelmatchresourcesetup/PanelMatchResourceSetup.kt#L54)):*

See other comment re:reconciling the differences between this and InProcessPanelMatchResourceSetup.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/panelmatchresourcesetup/PanelMatchResourceSetup.kt` line 62 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1273#-Ng-EeCb3pbWRtK-gHui:-Ng3KWF7CABBL1IOQg03:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/42b5367737c245b5eaa3ad812dd2e37a34e9ce64/src/main/kotlin/org/wfanet/measurement/loadtest/panelmatchresourcesetup/PanelMatchResourceSetup.kt#L62)):*

i think you meant to delete this
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/panelmatch/client/storage/forwarded/ForwardedStorageFactory.kt` line 28 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1273#-Ng-Ex0q8BiZ2w0ce3GS:-Ng3ddK57t_8ezRlVfD0:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/42b5367737c245b5eaa3ad812dd2e37a34e9ce64/src/main/kotlin/org/wfanet/panelmatch/client/storage/forwarded/ForwardedStorageFactory.kt#L28)):*

add a kdoc
</blockquote></details>

Done.
___
*[`src/main/proto/wfa/measurement/loadtest/panelmatch/resourcesetup/resources.proto` line 19 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1273#-Ng-FA5e9XO93V-iSG6z:-Ng37_idC31de2SFxSff:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/42b5367737c245b5eaa3ad812dd2e37a34e9ce64/src/main/proto/wfa/measurement/loadtest/panelmatch/resourcesetup/resources.proto#L19)):*

delete
</blockquote></details>

Done.
___
*[`src/test/kotlin/org/wfanet/panelmatch/integration/k8s/AbstractPanelMatchCorrectnessTest.kt` line 95 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1273#-Ng-K-WA5di_1rtjPh1v:-NgJJ0Oj302ZX22A8fnl:b-1xqcuh) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/42b5367737c245b5eaa3ad812dd2e37a34e9ce64/src/test/kotlin/org/wfanet/panelmatch/integration/k8s/AbstractPanelMatchCorrectnessTest.kt#L95)):*

you should also pass in expected outputs for both parties that you verify
</blockquote></details>

I think this would add a bit of complexity since each test uses different types to test against:

- DoubleBlind (no output check)
- MiniWorkflow (ByteString)
- SingleStep (ByteString)
- FullWithPreprocessing (List<String>)

Wouldn't be better is the test class implementation is responsible to check whatever it needs to?

LMK
___
*[`src/test/kotlin/org/wfanet/panelmatch/integration/k8s/AbstractPanelMatchCorrectnessTest.kt` line 119 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1273#-Ng0586X98GMx2w2u3Jh:-NgJFCurCoZM_XUW_qHh:b-g8lkzh) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/42b5367737c245b5eaa3ad812dd2e37a34e9ce64/src/test/kotlin/org/wfanet/panelmatch/integration/k8s/AbstractPanelMatchCorrectnessTest.kt#L119)):*

This should only be in the "empty cluster" version of the test, so shouldn't be in the abstract base class. We will need a version of the test that can hit existing deployed instances in a cloud environment, but that can be added in a later PR.
</blockquote></details>

Ok, I've added a TODO
___
*[`src/test/kotlin/org/wfanet/panelmatch/integration/k8s/AbstractPanelMatchCorrectnessTest.kt` line 175 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1273#-Ng-JELT8XIwb2JMp7pf:-NgJHmwc3tfwoijDcihi:b-mn8oob) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/42b5367737c245b5eaa3ad812dd2e37a34e9ce64/src/test/kotlin/org/wfanet/panelmatch/integration/k8s/AbstractPanelMatchCorrectnessTest.kt#L175)):*

i think storage should be abstract here since someone might want to run an cloud test using s3 gcp or whatever
</blockquote></details>

Agree. Do you mind if I add a TODO for this and add it later once refactoring code to move ```runResourceSetup``` away from the abstract base class? Otherwise I should refactor this twice (creating channels, retrieving the POD etc)
___
*[`src/test/kotlin/org/wfanet/panelmatch/integration/k8s/DoubleBlindWorkflowTest.kt` line 111 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1273#-Ng-KAG092RXD_Q7ATV8:-Ng3AJPL0INtEKs1acYo:bgk22e6) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/42b5367737c245b5eaa3ad812dd2e37a34e9ce64/src/test/kotlin/org/wfanet/panelmatch/integration/k8s/DoubleBlindWorkflowTest.kt#L111)):*

see my other comment. outputs' contents should be verified against expected outputs. This may be problematic if any outputs are not-deterministic. But you should be able to avoid that as long as there are no discarded queries.
</blockquote></details>

Can this be added when the remaining tests are added too @stevenwarejones ?
In this case the only thing that's being tested is that the output can be parsed without exception. I don't think we are checking any outcome here
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/panelmatchresourcesetup/InProcessPanelMatchResourceSetup.kt` line 52 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1273#-Ng01Nb18H0JIU05lEeG:-NgJIB4z-Nvbvs6KxmOS:b-2nvand) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/42b5367737c245b5eaa3ad812dd2e37a34e9ce64/src/main/kotlin/org/wfanet/measurement/loadtest/panelmatchresourcesetup/InProcessPanelMatchResourceSetup.kt#L52)):*

What's the "InProcess" part about here? This takes in a channel to an existing server, and doesn't appear to care whether it's an in-process server. I don't believe we need a separate "in-process" version of ResourceSetup, so we shouldn't need one for PanelMatchResourceSetup.
</blockquote></details>

This class has been removed
___
*[`src/main/proto/wfa/panelmatch/client/storage/storage_details.proto` line 97 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1273#-Ng-FF-93j_0ihBnxYXF:-Ng3Jp497LHeWx0-OYT0:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/42b5367737c245b5eaa3ad812dd2e37a34e9ce64/src/main/proto/wfa/panelmatch/client/storage/storage_details.proto#L97)):*

This really shouldn't be part of storageDetails. It should be in its own file and only used for loadtest.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/panelmatchresourcesetup/Output.kt` line  at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1273#-Ng032no5udqmNdwjsqD:-Ng3Xm196TBvcYBNPIL2:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/42b5367737c245b5eaa3ad812dd2e37a34e9ce64/src/main/kotlin/org/wfanet/measurement/loadtest/panelmatchresourcesetup/Output.kt#L)):*

This appears to be a direct copy of `src/main/kotlin/org/wfanet/measurement/loadtest/resourcesetup/Output.kt`. We really shouldn't just add another copy.
</blockquote></details>

Done.

stevenwarejones (changes requested):
> Reviewed 1 of 39 files at r1, 25 of 27 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1273)* status: 42 of 44 files reviewed, 14 unresolved discussions (waiting on @Marco-Premier and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/panelmatch/client/deploy/example/forwarded/ForwardedStorageExchangeWorkflowDaemon.kt` line 70 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1273#-NgU20j0Bh1p93HM5Gdv:-NgU20j0Bh1p93HM5Gdw:bnij4fv) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4668e3b4f6e96bed4a30547a56611e0bc525e698/src/main/kotlin/org/wfanet/panelmatch/client/deploy/example/forwarded/ForwardedStorageExchangeWorkflowDaemon.kt#L70)):*
> ```Kotlin
>     get() = TODO("Not yet implemented")
> 
>   override val certificateManager: CertificateManager = TestCertificateManager
> ```

can you add a TODO to have each daemon use a separate root. I believe they both use the same root now which might not catch some edge defect use cases.
___
*[`src/test/kotlin/org/wfanet/panelmatch/integration/k8s/AbstractPanelMatchCorrectnessTest.kt` line 95 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1273#-Ng-K-WA5di_1rtjPh1v:-NgU00EH575lInFPKFH4:b-r2b10y) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/42b5367737c245b5eaa3ad812dd2e37a34e9ce64/src/test/kotlin/org/wfanet/panelmatch/integration/k8s/AbstractPanelMatchCorrectnessTest.kt#L95)):*

I think this would add a bit of complexity since each test uses different types to test against:

- DoubleBlind (no output check)
- MiniWorkflow (ByteString)
- SingleStep (ByteString)
- FullWithPreprocessing (List<String>)

Wouldn't be better is the test class implementation is responsible to check whatever it needs to?

LMK
</blockquote></details>

I don't think its that complicated.... just pass in a map of `blobName:String to expectedOutput: Message`- parse the byteString blob based on the Message descriptor and proto compares. This is fine to leave as a TODO as long as you check in the individual tests for now.
___
*[`src/test/kotlin/org/wfanet/panelmatch/integration/k8s/AbstractPanelMatchCorrectnessTest.kt` line 175 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1273#-Ng-JELT8XIwb2JMp7pf:-NgU0vMb8qWOAmI3qdNH:bz8kpeg) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/42b5367737c245b5eaa3ad812dd2e37a34e9ce64/src/test/kotlin/org/wfanet/panelmatch/integration/k8s/AbstractPanelMatchCorrectnessTest.kt#L175)):*

Agree. Do you mind if I add a TODO for this and add it later once refactoring code to move ```runResourceSetup``` away from the abstract base class? Otherwise I should refactor this twice (creating channels, retrieving the POD etc)
</blockquote></details>

yep - add a TODO
___
*[`src/test/kotlin/org/wfanet/panelmatch/integration/k8s/DoubleBlindWorkflowTest.kt` line 111 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1273#-Ng-KAG092RXD_Q7ATV8:-NgU1HCd8WJjvX-H8XwT:b-45v48o) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/42b5367737c245b5eaa3ad812dd2e37a34e9ce64/src/test/kotlin/org/wfanet/panelmatch/integration/k8s/DoubleBlindWorkflowTest.kt#L111)):*

Can this be added when the remaining tests are added too @stevenwarejones ?
In this case the only thing that's being tested is that the output can be parsed without exception. I don't think we are checking any outcome here
</blockquote></details>

I guess what you have is parity with the existing test: https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/192c5ee4aefec3238e0a9ef4e5b30b46493a6a3f/src/test/kotlin/org/wfanet/panelmatch/integration/DoubleBlindWorkflowTest.kt#L57

Its not great, however. It really should be verifying the parsed outputs. Please add a TODO to check the specific parsed outputs.

Marco-Premier (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1273)* status: 40 of 44 files reviewed, 14 unresolved discussions (waiting on @SanjayVas and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/panelmatch/client/deploy/example/forwarded/ForwardedStorageExchangeWorkflowDaemon.kt` line 70 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1273#-NgU20j0Bh1p93HM5Gdv:-NgUhENi7Uein4RwWBeQ:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4668e3b4f6e96bed4a30547a56611e0bc525e698/src/main/kotlin/org/wfanet/panelmatch/client/deploy/example/forwarded/ForwardedStorageExchangeWorkflowDaemon.kt#L70)):*

can you add a TODO to have each daemon use a separate root. I believe they both use the same root now which might not catch some edge defect use cases.
</blockquote></details>

Done.
___
*[`src/test/kotlin/org/wfanet/panelmatch/integration/k8s/AbstractPanelMatchCorrectnessTest.kt` line 95 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1273#-Ng-K-WA5di_1rtjPh1v:-NgUkWHIBuGTPBHtc1R8:b-y0ve1d) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/42b5367737c245b5eaa3ad812dd2e37a34e9ce64/src/test/kotlin/org/wfanet/panelmatch/integration/k8s/AbstractPanelMatchCorrectnessTest.kt#L95)):*

I don't think its that complicated.... just pass in a map of `blobName:String to expectedOutput: Message`- parse the byteString blob based on the Message descriptor and proto compares. This is fine to leave as a TODO as long as you check in the individual tests for now.
</blockquote></details>

Ok adding a TODO here and will address it with others in a following PR
___
*[`src/test/kotlin/org/wfanet/panelmatch/integration/k8s/AbstractPanelMatchCorrectnessTest.kt` line 175 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1273#-Ng-JELT8XIwb2JMp7pf:-NgUiKR-6BSSRiogWWGN:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/42b5367737c245b5eaa3ad812dd2e37a34e9ce64/src/test/kotlin/org/wfanet/panelmatch/integration/k8s/AbstractPanelMatchCorrectnessTest.kt#L175)):*

yep - add a TODO
</blockquote></details>

Done.
___
*[`src/test/kotlin/org/wfanet/panelmatch/integration/k8s/DoubleBlindWorkflowTest.kt` line 111 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1273#-Ng-KAG092RXD_Q7ATV8:-NgUkIf2DFBHgurs_CFB:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/42b5367737c245b5eaa3ad812dd2e37a34e9ce64/src/test/kotlin/org/wfanet/panelmatch/integration/k8s/DoubleBlindWorkflowTest.kt#L111)):*

I guess what you have is parity with the existing test: https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/192c5ee4aefec3238e0a9ef4e5b30b46493a6a3f/src/test/kotlin/org/wfanet/panelmatch/integration/DoubleBlindWorkflowTest.kt#L57

Its not great, however. It really should be verifying the parsed outputs. Please add a TODO to check the specific parsed outputs.
</blockquote></details>

Done.

SanjayVas (changes requested):
> Reviewed 1 of 39 files at r1, 26 of 27 files at r2, 3 of 3 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1273)* status: all files reviewed, 10 unresolved discussions (waiting on @Marco-Premier and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/integration/common/InProcessKingdom.kt` line 193 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1273#-NgUvWRj3pBojo7M1G3Q:-NgUvWRj3pBojo7M1G3R:b2ckv03) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3554bc82b8f2c573d03529195ace8aaa5c3e21f7/src/main/kotlin/org/wfanet/measurement/integration/common/InProcessKingdom.kt#L193)):*
> ```Kotlin
>     get() = systemApiServer.channel
> 
>   /** Provides a InProcessPanelMatchResourceSetup instance with the Kingdom's internal API. */
> ```

nit: no longer accurate. Also the summary fragment should be a noun phrase since this is a property and not a function.

Speaking of which, perhaps this should be written in a way that doesn't give a new instance on each access?



_Suggestion:_
```Kotlin
/** [PanelMatchResourceSetup] instance with the Kingdom's internal API. */
```
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/panelmatchresourcesetup/PanelMatchResourceSetup.kt` line 67 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1273#-NgUwSKwAGgV7BkV_bGv:-NgUwSKwAGgV7BkV_bGw:b36bb12) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3554bc82b8f2c573d03529195ace8aaa5c3e21f7/src/main/kotlin/org/wfanet/measurement/loadtest/panelmatchresourcesetup/PanelMatchResourceSetup.kt#L67)):*
> ```Kotlin
> ) {
> 
>   /** The Channel can be used in the in-process integration test. */
> ```

Follow [style guide](https://developer.android.com/kotlin/style-guide#documentation) for documenting functions. This means having a summary fragment followed by optional param documentation.
___
*[`src/main/proto/wfa/panelmatch/client/loadtest/forwarded_storage.proto` line 23 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1273#-NgUsb4n0awscbxuqqCR:-NgUsb4n0awscbxuqqCS:b-y28qq7) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3554bc82b8f2c573d03529195ace8aaa5c3e21f7/src/main/proto/wfa/panelmatch/client/loadtest/forwarded_storage.proto#L23)):*
> ```proto
> option java_outer_classname = "ForwardedStorageProto";
> 
> message ForwardedStorage {
> ```

nit: shouldn't this be called ForwardedStorageConfig to mirror other similar message types?
___
*[`src/test/kotlin/org/wfanet/panelmatch/integration/k8s/DoubleBlindWorkflowTest.kt` line 48 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1273#-NgUtB1z7nkT9jIHo6qY:-NgUtB1z7nkT9jIHo6qZ:bx61lvp) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3554bc82b8f2c573d03529195ace8aaa5c3e21f7/src/test/kotlin/org/wfanet/panelmatch/integration/k8s/DoubleBlindWorkflowTest.kt#L48)):*
> ```Kotlin
> 
> @RunWith(JUnit4::class)
> class DoubleBlindWorkflowTest : AbstractPanelMatchCorrectnessTest() {
> ```

We don't want a whole separate instance of the test for each workflow type, as that means doing all the setup to populate the cluster each time. If you want to test different workflow types, each should just be another test method in the same class. This may mean having to refactor how resource setup is done. Of course if you only want to test one workflow then this is fine.

As a reminder, resource setup exists just to do the one-time setup of API resources that we need in order to write config files. e.g. the Kingdom needs to have the AKID mapping file on startup, so we need to do a bootstrapping pass to generate the config files.



_Code quote:_
```Kotlin
DoubleBlindWorkflowTest
```

Marco-Premier (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1273)* status: all files reviewed, 8 unresolved discussions (waiting on @SanjayVas and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/panelmatchresourcesetup/PanelMatchResourceSetup.kt` line 67 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1273#-NgUwSKwAGgV7BkV_bGv:-NgXlBJP4A4P4-bQHcw5:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3554bc82b8f2c573d03529195ace8aaa5c3e21f7/src/main/kotlin/org/wfanet/measurement/loadtest/panelmatchresourcesetup/PanelMatchResourceSetup.kt#L67)):*

Follow [style guide](https://developer.android.com/kotlin/style-guide#documentation) for documenting functions. This means having a summary fragment followed by optional param documentation.
</blockquote></details>

Done.
___
*[`src/test/kotlin/org/wfanet/panelmatch/integration/k8s/DoubleBlindWorkflowTest.kt` line 48 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1273#-NgUtB1z7nkT9jIHo6qY:-NgXmFhB0EO00CBD-i-h:bs8ojs3) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3554bc82b8f2c573d03529195ace8aaa5c3e21f7/src/test/kotlin/org/wfanet/panelmatch/integration/k8s/DoubleBlindWorkflowTest.kt#L48)):*

We don't want a whole separate instance of the test for each workflow type, as that means doing all the setup to populate the cluster each time. If you want to test different workflow types, each should just be another test method in the same class. This may mean having to refactor how resource setup is done. Of course if you only want to test one workflow then this is fine.

As a reminder, resource setup exists just to do the one-time setup of API resources that we need in order to write config files. e.g. the Kingdom needs to have the AKID mapping file on startup, so we need to do a bootstrapping pass to generate the config files.
</blockquote></details>

Yes indeed, thanks Sanjay. I'm aware of that. Due to the size of the PR I preferred to collect feedback for a single test first.
My idea is to merge this and then (in a following PR), when adding other tests refactor this part; meaning that:
1) I'll move resourceSetup in a EmptyCluster version of the test in which I will create both ```DataProvider``` and ```ModelProvider``` and setup AKID mapping.
2) I will create a class ```PanelMatchSimulator``` similar to ```MeasurementConsumerSimulator``` that takes care of creating new RecurringExchange and injecting workflows for each test (```DoubleBlindWorkflowTest```, ```FullWorkflowTest``` etc) so that it would conform to Existing Correctness tests

how does that sound?

stevenwarejones (approved):
> Reviewed 1 of 27 files at r2, 3 of 3 files at r3, 6 of 6 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1273)* status: all files reviewed, 3 unresolved discussions (waiting on @SanjayVas)

SanjayVas (approved):
> Reviewed 6 of 6 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1273)* status: all files reviewed, 1 unresolved discussion (waiting on @stevenwarejones)

___
*[`src/test/kotlin/org/wfanet/panelmatch/integration/k8s/DoubleBlindWorkflowTest.kt` line 48 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1273#-NgUtB1z7nkT9jIHo6qY:-Ng_3p4z1jQnTciunJlM:bb73djm) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3554bc82b8f2c573d03529195ace8aaa5c3e21f7/src/test/kotlin/org/wfanet/panelmatch/integration/k8s/DoubleBlindWorkflowTest.kt#L48)):*

Yes indeed, thanks Sanjay. I'm aware of that. Due to the size of the PR I preferred to collect feedback for a single test first.
My idea is to merge this and then (in a following PR), when adding other tests refactor this part; meaning that:
1) I'll move resourceSetup in a EmptyCluster version of the test in which I will create both ```DataProvider``` and ```ModelProvider``` and setup AKID mapping.
2) I will create a class ```PanelMatchSimulator``` similar to ```MeasurementConsumerSimulator``` that takes care of creating new RecurringExchange and injecting workflows for each test (```DoubleBlindWorkflowTest```, ```FullWorkflowTest``` etc) so that it would conform to Existing Correctness tests

how does that sound?
</blockquote></details>

SGTM

stevenwarejones (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1273)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Marco-Premier)

Issue #984, opened 2023-05-08
R/F Measurement result sometimes has incorrect reach value of 1


SanjayVas:
Observed in runs of the correctness test on halo-cmm-dev. This uses a [fixed dataset](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/main/src/main/k8s/testing/data/synthetic-labelled-events.csv) and query parameters (time, filters, etc.), and therefore should give the same result every time aside from DP noise (reach: 1296). Instead, we frequently (maybe 10% of runs?) get a reach value of 1.

From @jiayu-google:
> Ok, with vidSamplingWidth = 1, the error is not that large, Probability(observing 1296 | truth = 1) < 1e-50. So almost impossible.

## Sample

Expected:

```textproto
reach {
  value: 1296
}
frequency {
  relative_frequency_distribution {
    key: 1
    value: 0.3078817733990148
  }
  relative_frequency_distribution {
    key: 2
    value: 0.6042692939244664
  }
  relative_frequency_distribution {
    key: 3
    value: 0.031198686371100164
  }
  relative_frequency_distribution {
    key: 4
    value: 0.05008210180623974
  }
  relative_frequency_distribution {
    key: 5
    value: 0.004105090311986864
  }
  relative_frequency_distribution {
    key: 6
    value: 0.0024630541871921183
  }
}
```

Actual:

```textproto
reach {
  value: 1
}
frequency {
  relative_frequency_distribution {
    key: 1
    value: 0.0738255033557047
  }
  relative_frequency_distribution {
    key: 4
    value: 0.6510067114093959
  }
  relative_frequency_distribution {
    key: 6
    value: 0.026845637583892617
  }
  relative_frequency_distribution {
    key: 7
    value: 0.03355704697986577
  }
  relative_frequency_distribution {
    key: 8
    value: 0.21476510067114093
  }
}
```

SanjayVas:
I was able to reproduce this intermittently in local tests. The one common factor is that every time this error occurs, I see an "editVersionMismatch" error in the logs. This appears to be related to #680.

PR #1290, opened 2023-10-12
Fix bug on rollv2 that outputs zero reach if the execution phase crashes right after writing to the output blob.

ple13:
In the execution phase of reach only protocols, the aggregator first initializes the reach to 0. It then runs an existingOutputOr block to compute the next token. If the output blob is empty, it updates the reach and writes stuff to the output blob. Otherwise, it skips the block, reuses the cache results and sends the reach to the kingdom.

There is a problem when the output blob exists, but the aggregator crashes before sending the reach to the kingdom. As the existingOutputOr block is skipped, the reach will not be updated and will stay at the initialized value (i.e. 0). As a result, the kingdom will receive 0.


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1290)

SanjayVas (commented):
I wonder if this behavior was the other part of #984.

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1290#-:-Ng_42Al9Gh5nPo7VIOM:b62x078)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @renjiezh)

renjiezh (commented):
I happen to see a suspicious case that yields reach=1 which might be related. 
I was running multiple Measurements one by one for the accuracy test. The last Measurement was about to finish. However, at this moment the timeout force components to shutdown. Then the measurement didn't fully failed but gave me a result of reach=1. 
I guess it somehow produced the scenario of crashes. Unfortunately I cannot reproduce it . 

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1290#-:-Ng_5jpSE9bx14gNZmz1:b-m60tcx)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @SanjayVas)

renjiezh (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1290)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on @ple13 and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/liquidlegionsv2/ReachOnlyLiquidLegionsV2Mill.kt` line 528 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1290#-Ng_FKf89YXOr846oH1N:-Ng_FKf89YXOr846oH1O:brruiyy) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/32f43681f2dc3b17725fea3f22efe36014940b0d/src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/liquidlegionsv2/ReachOnlyLiquidLegionsV2Mill.kt#L528)):*
> ```Kotlin
>         reach = cryptoResult.reach
>         cryptoResult.toByteString()
>       }
> ```

What if we simply move ` sendResultToKingdom(token, ReachResult(reach))` into the scope of  `existingOutputOr`?
If I understand correctly, it guarantees that the reach result is sent to Kingdom before store the blob. If it crashes after store blob but before `completeComputation`, it will not rerun but reuse the token from the blob for `completeComputation`.
One purpose of it is to avoid to write the result into blob/db in duchies.

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1290)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on @renjiezh)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/liquidlegionsv2/ReachOnlyLiquidLegionsV2Mill.kt` line 528 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1290#-Ng_FKf89YXOr846oH1N:-Ng_gxW1AYgRT-gUQbTW:bwhtqex) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/32f43681f2dc3b17725fea3f22efe36014940b0d/src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/liquidlegionsv2/ReachOnlyLiquidLegionsV2Mill.kt#L528)):*

What if we simply move ` sendResultToKingdom(token, ReachResult(reach))` into the scope of  `existingOutputOr`?
If I understand correctly, it guarantees that the reach result is sent to Kingdom before store the blob. If it crashes after store blob but before `completeComputation`, it will not rerun but reuse the token from the blob for `completeComputation`.
One purpose of it is to avoid to write the result into blob/db in duchies.
</blockquote></details>

I like the idea of using the existing blob check if that's effectively the same. It also means existingOutputOr can continue to return the appropriate next token.

ple13 (commented):
I'm not sure if they are related. This bug sets reach = 0 in the llv2 computationDetails, but we have reach = 1 in the above scenarios.

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1290#-:-NgedMOyDqDr1iYv4w9_:bwnpik6)* status: 0 of 5 files reviewed, 1 unresolved discussion (waiting on @renjiezh)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/liquidlegionsv2/ReachOnlyLiquidLegionsV2Mill.kt` line 528 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1290#-Ng_FKf89YXOr846oH1N:-Nged5ob2vpJ-LrcJJvu:bpfo18z) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/32f43681f2dc3b17725fea3f22efe36014940b0d/src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/liquidlegionsv2/ReachOnlyLiquidLegionsV2Mill.kt#L528)):*

I like the idea of using the existing blob check if that's effectively the same. It also means existingOutputOr can continue to return the appropriate next token.
</blockquote></details>

This is a good idea. I moved the sendResultToKingdom inside the existingOutputOr block and don't update the computationDetails.

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1290)* status: 0 of 5 files reviewed, 2 unresolved discussions (waiting on @ple13 and @renjiezh)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/MillBase.kt` line 470 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1290#-NgtJE4oDBQm8-xAvhT-:-NgtJE4p1K8s5tDH3KJv:b-djzfg9) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7763c4ebba4e4c466c7b8f17e26a5a22dab5d681/src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/MillBase.kt#L470)):*
> ```Kotlin
>    * otherwise compute the new result by executing [block].
>    */
>   protected suspend fun executeBlock(
> ```

This name is far too generic, as the pattern of executing the lambda parameter is very common. I'm also slightly confused, as the name indicates that `block` will be unconditionally executed but the description somehow implies that it will both always be executed and but also only executed in the "otherwise" clause? I thought we were explicitly relying on the behavior that `block` is not always executed.



_Code quote:_
```Kotlin
executeBlock
```

ple13 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1290)* status: 0 of 5 files reviewed, 2 unresolved discussions (waiting on @renjiezh and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/MillBase.kt` line 470 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1290#-NgtJE4oDBQm8-xAvhT-:-Nh3_09X6RfUY3q1Rddq:b-16w3b8) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7763c4ebba4e4c466c7b8f17e26a5a22dab5d681/src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/MillBase.kt#L470)):*

This name is far too generic, as the pattern of executing the lambda parameter is very common. I'm also slightly confused, as the name indicates that `block` will be unconditionally executed but the description somehow implies that it will both always be executed and but also only executed in the "otherwise" clause? I thought we were explicitly relying on the behavior that `block` is not always executed.
</blockquote></details>

I've changed it to existingOutputAnd. Please let me know what you think. As we log the runtime in the millbase, I want to keep the function declared there so that we still have the log.

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1290)* status: 0 of 5 files reviewed, 2 unresolved discussions (waiting on @renjiezh)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/MillBase.kt` line 470 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1290#-NgtJE4oDBQm8-xAvhT-:-Nh3abDKAyelF4bmPp9Q:b-vln9l8) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7763c4ebba4e4c466c7b8f17e26a5a22dab5d681/src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/MillBase.kt#L470)):*
> I thought we were explicitly relying on the behavior that `block` is not always executed.

What about this point? I thought this was the crux of @renjiezh's suggestion.

ple13 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1290)* status: 0 of 5 files reviewed, 2 unresolved discussions (waiting on @renjiezh)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/MillBase.kt` line 470 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1290#-NgtJE4oDBQm8-xAvhT-:-Nh3edJ414hMTykHPmxr:bq6soxx) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7763c4ebba4e4c466c7b8f17e26a5a22dab5d681/src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/MillBase.kt#L470)):*

> I thought we were explicitly relying on the behavior that `block` is not always executed.

What about this point? I thought this was the crux of @renjiezh's suggestion.
</blockquote></details>

Do you mean the idea of sending results to the Kingdom inside the block?

ple13 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1290)* status: 0 of 5 files reviewed, 2 unresolved discussions (waiting on @renjiezh)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/MillBase.kt` line 470 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1290#-NgtJE4oDBQm8-xAvhT-:-Nh3ud9u8HRIONH4RhJj:b-12q03g) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7763c4ebba4e4c466c7b8f17e26a5a22dab5d681/src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/MillBase.kt#L470)):*

Do you mean the idea of sending results to the Kingdom inside the block?
</blockquote></details>

Do we always assume that the block may not be executed for lambda block? Or is it just for the function existingOutputOr?

SanjayVas (approved):
> Reviewed 3 of 5 files at r3, 2 of 2 files at r4.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1290)* status: all files reviewed, 2 unresolved discussions (waiting on @renjiezh)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/MillBase.kt` line 470 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1290#-NgtJE4oDBQm8-xAvhT-:-Nh7mKoE2V9BxA3de_fi:bnpidxj) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7763c4ebba4e4c466c7b8f17e26a5a22dab5d681/src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/MillBase.kt#L470)):*

Do we always assume that the block may not be executed for lambda block? Or is it just for the function existingOutputOr?
</blockquote></details>

It was for the existing function. I thought that @renjiezh's suggestion involved conditionally sending the result to the Kingdom. I'm going to unblock this for now as long as this PR actually solves the problem.

renjiezh (approved):
> Reviewed 3 of 5 files at r3, 2 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1290)* status: all files reviewed, 1 unresolved discussion (waiting on @ple13)

ple13 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1290)* status: all files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/MillBase.kt` line 470 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1290#-NgtJE4oDBQm8-xAvhT-:-Nh8X0pHA6xvVL_JvnXc:b-6gvmib) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7763c4ebba4e4c466c7b8f17e26a5a22dab5d681/src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/MillBase.kt#L470)):*

It was for the existing function. I thought that @renjiezh's suggestion involved conditionally sending the result to the Kingdom. I'm going to unblock this for now as long as this PR actually solves the problem.
</blockquote></details>

Thanks. For rollv2, Renjie's idea works and the fix is done that way (as reach is the only thing that will be sent to the Kingdom).

stevenwarejones (approved):
> Reviewed 3 of 5 files at r3, 2 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1290)* status: all files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

PR #1307, opened 2023-10-23
Fix internal DeleteEventGroup not clearing details.

SanjayVas:
The EventGroupDetails column in the EventGroups table is nullable, and GetEventGroup correctly handles the null case by not having the details field set in the EventGroup message. DeleteEventGroup was incorrectly setting both the column and the response value to an empty EventGroup.Details.


Marco-Premier (approved):
> Reviewed 2 of 2 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1307)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @SanjayVas)

PR #1313, opened 2023-10-27
Allow data provider to provide separate result consent signaling certificate

stevenwarejones:


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1313)

SanjayVas (changes requested):
> Reviewed 16 of 19 files at r2.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1313)* status: 11 of 19 files reviewed, 5 unresolved discussions (waiting on @Marco-Premier and @stevenwarejones)

___
*[`build/repositories.bzl` line 41 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1313#-Ni_t2jG6oLefDKgSVgG:-Ni_t2jH4aHHmb8qSP7T:b-u82d3k) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8558ffc784e66584248d581255c8c03dd74b6724/build/repositories.bzl#L41)):*
> ```Starlark
>     )
> 
>     # TODO: Update with latest version before merge
> ```

I believe you've already created the release for this.
___
*[`src/main/kotlin/org/wfanet/measurement/integration/common/InProcessEdpSimulator.kt` line 134 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1313#-Ni_thrS-oREuW9yg-kY:-Ni_thrS-oREuW9yg-kZ:b7e64rj) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8558ffc784e66584248d581255c8c03dd74b6724/src/main/kotlin/org/wfanet/measurement/integration/common/InProcessEdpSimulator.kt#L134)):*
> ```Kotlin
> 
>   private val edpResultCsSigningKey =
>     if (Files.exists(SECRET_FILES_PATH.resolve("${displayName}_result_cs_cert.der"))) {
> ```

You don't need this check. Local tests can rely on the disposition of files that are under version control. You should therefore also not need SECRET_FILES_PATH.
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/RequisitionsService.kt` line 228 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1313#-Ni_vPSs9uVStkFnZGU4:-Ni_vPSs9uVStkFnZGU5:b-nfhtde) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8558ffc784e66584248d581255c8c03dd74b6724/src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/RequisitionsService.kt#L228)):*
> ```Kotlin
>         encryptedData = request.encryptedData
>         val dataProviderCertificateKey = DataProviderCertificateKey.fromName(request.certificate)
>         if (dataProviderCertificateKey !== null) {
> ```

If the value is set (string field is not empty) then we should throw an INVALID_ARGUMENT if it doesn't parse correctly.
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt` line 149 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1313#-Ni_ymGa8zH6OOlgAAqa:-Ni_ymGa8zH6OOlgAAqb:bm461p8) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8558ffc784e66584248d581255c8c03dd74b6724/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt#L149)):*
Replace this with a map of certificate resource name to `SigningKeyHandle`, as well as a single resource name of the result certificate. This would resolve the code comment in `fulfillDirectMeasurement`:

> ```kotlin
> // Verify that the signing key we have matches the one from the Requisition. A real EDP would
> // look up the matching signing key, but the simulator is only configured with one.
> ```
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt` line 179 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1313#-Ni_zLidEXgxp40NzAAC:-Ni_yUBk-IpFeP-wMH0w:b-b7n0lh) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8558ffc784e66584248d581255c8c03dd74b6724/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt#L179)):*
> ```Kotlin
> 
>   /** Ensures that an appropriate [Certificate] exists for the [DataProvider]. */
>   suspend fun ensureCertificate(): Certificate? {
> ```

Rather than having the EDP simulator creating certificates, require that the Certificate resource already exists. The caller should pass in the resource name.

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1313)* status: 8 of 44 files reviewed, 5 unresolved discussions (waiting on @Marco-Premier and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/integration/common/InProcessEdpSimulator.kt` line 134 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1313#-Ni_thrS-oREuW9yg-kY:-Nia9wB5CfvxrK05n9FB:b-xhja5u) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8558ffc784e66584248d581255c8c03dd74b6724/src/main/kotlin/org/wfanet/measurement/integration/common/InProcessEdpSimulator.kt#L134)):*

You don't need this check. Local tests can rely on the disposition of files that are under version control. You should therefore also not need SECRET_FILES_PATH.
</blockquote></details>

I'm not sure what this comment was referring to since the context is lost. However, I the code above for `separateResultSigningKey` is null if a test edp doesn't have that present.
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/RequisitionsService.kt` line 228 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1313#-Ni_vPSs9uVStkFnZGU4:-Nia9qGl1xp1VNnw_I9B:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8558ffc784e66584248d581255c8c03dd74b6724/src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/RequisitionsService.kt#L228)):*

If the value is set (string field is not empty) then we should throw an INVALID_ARGUMENT if it doesn't parse correctly.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt` line 149 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1313#-Ni_ymGa8zH6OOlgAAqa:-NjEWdYf5tW5ING2hReT:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8558ffc784e66584248d581255c8c03dd74b6724/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt#L149)):*

Replace this with a map of certificate resource name to `SigningKeyHandle`, as well as a single resource name of the result certificate. This would resolve the code comment in `fulfillDirectMeasurement`:

> ```kotlin
> // Verify that the signing key we have matches the one from the Requisition. A real EDP would
> // look up the matching signing key, but the simulator is only configured with one.
> ```
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt` line 179 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1313#-Ni_zLidEXgxp40NzAAC:-NjEWfcNDiX7wLFqZLrT:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8558ffc784e66584248d581255c8c03dd74b6724/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt#L179)):*

Rather than having the EDP simulator creating certificates, require that the Certificate resource already exists. The caller should pass in the resource name.
</blockquote></details>

Done.
___
*[`build/repositories.bzl` line 41 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1313#-Ni_t2jG6oLefDKgSVgG:-Nia7AgD1YFeThjZYWQA:bihzsey) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8558ffc784e66584248d581255c8c03dd74b6724/build/repositories.bzl#L41)):*

I believe you've already created the release for this.
</blockquote></details>

This was already updated to the new release.

SanjayVas (changes requested):
> Reviewed 1 of 19 files at r2, 2 of 12 files at r3, 25 of 37 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1313)* status: 33 of 44 files reviewed, 10 unresolved discussions (waiting on @Marco-Premier and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/api/v2alpha/tools/BUILD.bazel` line 23 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1313#-NjFHtca1gmX3HtObZEE:-NjFHtca1gmX3HtObZEF:b-pwk7r0) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f1c095d77a8d73ddc334680fe45db2df07b02685/src/main/kotlin/org/wfanet/measurement/api/v2alpha/tools/BUILD.bazel#L23)):*
> ```Starlark
> )
> 
> java_binary(
> ```

These shouldn't be separate binaries and should instead be subcommands of `MeasurementSystem`. That CLI tool is meant to cover methods from the CMMS public API.
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt` line 149 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1313#-Ni_ymGa8zH6OOlgAAqa:-NjFS-1OB38T4yVsheg1:bnt06cu) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8558ffc784e66584248d581255c8c03dd74b6724/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt#L149)):*

Done.
</blockquote></details>

Sorry, I guess we don't actually need this to be a map if we only ever use one certificate-key pair for signing. I was thinking about the case of having two certificates, but that can just be exercised by picking a different certificate here than the one in the DataProvider PublicKey resource.
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt` line 155 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1313#-NjFKuGcBKgmIviS2jA2:-NjFKuGcBKgmIviS2jA3:b-kirmpc) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f1c095d77a8d73ddc334680fe45db2df07b02685/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt#L155)):*
> ```Kotlin
>    * name from the requisition will be used.
>    */
>   val resultSigningResource: DataProviderCertificateKey?
> ```

Since the EDP should always specify the result certificate, I think we want this to be non-nullable. If we really don't want the caller to have to specify it, we can default it to be the first certificate in `registeredCertificates`. 



_Code quote:_
```Kotlin
  /**
   * The CertificateKey to use for result signing. If not specified, the data provider certificate
   * name from the requisition will be used.
   */
  val resultSigningResource: DataProviderCertificateKey?
```
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt` line 379 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1313#-NjFLY1i69PYUmtcsviT:-NjFLY1jBGOh_QVhzFhD:bg4qrg5) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f1c095d77a8d73ddc334680fe45db2df07b02685/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt#L379)):*
> ```Kotlin
>     val measurementSpec: MeasurementSpec = requisition.measurementSpec.message.unpack()
> 
>     // TODO: A real EDP should verify the signed `data_provider_public_key`.
> ```

This is not necessary, and furthermore not even feasible after https://github.com/world-federation-of-advertisers/cross-media-measurement-api/pull/187.

The key is signed so that the MC can verify it came from the EDP.



_Code quote:_
```Kotlin
 // TODO: A real EDP should verify the signed `data_provider_public_key`.
```
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt` line 381 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1313#-NjFMSDD0Y3vsAfkVvai:-NjFMSDD0Y3vsAfkVvaj:be4odko) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f1c095d77a8d73ddc334680fe45db2df07b02685/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt#L381)):*
> ```Kotlin
>     // TODO: A real EDP should verify the signed `data_provider_public_key`.
>     // TODO: A real EDP should map the `data_provider_public_key` to a known private key for
>     // decryption.
> ```

This shouldn't be a TODO, since we don't necessarily plan on implementing it for the EDP simulator. We're just trying to indicate something that the simulator does not do but a real EDP should.



_Code quote:_
```Kotlin
    // TODO: A real EDP should map the `data_provider_public_key` to a known private key for
    // decryption.
```
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/resourcesetup/ResourceSetup.kt` line 217 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1313#-NjFNw_k3vCcWIuxsqzt:-NjFNw_k3vCcWIuxsqzu:b5bops6) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f1c095d77a8d73ddc334680fe45db2df07b02685/src/main/kotlin/org/wfanet/measurement/loadtest/resourcesetup/ResourceSetup.kt#L217)):*
> ```Kotlin
>           Resources.Resource.ResourceCase.DATA_PROVIDER -> {
>             val displayName = resource.dataProvider.displayName
>             writer.appendLine("build:$configName --define=${displayName}_name=${resource.name}")
> ```

Also output the EDP certificate resource name definition here.
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/resourcesetup/ResourceSetup.kt` line 243 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1313#-NjFQ9OtACebT-wo9VU3:-NjFQ9OtACebT-wo9VU4:bkzph2k) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f1c095d77a8d73ddc334680fe45db2df07b02685/src/main/kotlin/org/wfanet/measurement/loadtest/resourcesetup/ResourceSetup.kt#L243)):*
> ```Kotlin
>   suspend fun createInternalDataProvider(
>     dataProviderContent: EntityContent
>   ): DataProviderResources {
> ```

Why not just return the internal `DataProvider`?



_Code quote:_
```Kotlin
DataProviderResources
```
___
*[`src/main/proto/wfa/measurement/internal/kingdom/requisitions_service.proto` line 86 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1313#-NjFQZJ53s0dhmIn1IXx:-NjFQZJ67Cei2MkWQAGh:b-9bi2a1) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f1c095d77a8d73ddc334680fe45db2df07b02685/src/main/proto/wfa/measurement/internal/kingdom/requisitions_service.proto#L86)):*
> ```proto
>     bytes encrypted_data = 2;
>     string api_version = 3;
>     // Certificate used to sign the encrypted `SignedData`.
> ```

nit: certificates are used to verify signatures, not to sign



_Suggestion:_
```proto
verify
```
___
*[`src/test/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulatorTest.kt` line 2192 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1313#-NjFRHKvBd-bK9fRZX_y:-NjFRHKvBd-bK9fRZX_z:bqhxhfz) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f1c095d77a8d73ddc334680fe45db2df07b02685/src/test/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulatorTest.kt#L2192)):*
> ```Kotlin
>         "${EDP_DISPLAY_NAME}_result_cs_cert.der",
>         "${EDP_DISPLAY_NAME}_result_cs_private.der"
>       )
> ```

nit: consistent naming since these are both signing keys used for consent signaling.



_Suggestion:_
```Kotlin
    private val EDP_SIGNING_KEY =
      loadSigningKey("${EDP_DISPLAY_NAME}_cs_cert.der", "${EDP_DISPLAY_NAME}_cs_private.der")
    private val EDP_RESULT_SIGNING_KEY =
      loadSigningKey(
        "${EDP_DISPLAY_NAME}_result_cs_cert.der",
        "${EDP_DISPLAY_NAME}_result_cs_private.der"
      )
```

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1313)* status: 11 of 62 files reviewed, 9 unresolved discussions (waiting on @Marco-Premier and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/integration/common/InProcessEdpSimulator.kt` line 134 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1313#-Ni_thrS-oREuW9yg-kY:-NjJ56JnBX1IVekObTHb:b-1ipjqf) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8558ffc784e66584248d581255c8c03dd74b6724/src/main/kotlin/org/wfanet/measurement/integration/common/InProcessEdpSimulator.kt#L134)):*

I'm not sure what this comment was referring to since the context is lost. However, I the code above for `separateResultSigningKey` is null if a test edp doesn't have that present.
</blockquote></details>

okay. removed.
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt` line 149 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1313#-Ni_ymGa8zH6OOlgAAqa:-NjI7mOL8_Wm0ThSgEGC:b-nt3y8k) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8558ffc784e66584248d581255c8c03dd74b6724/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt#L149)):*

Sorry, I guess we don't actually need this to be a map if we only ever use one certificate-key pair for signing. I was thinking about the case of having two certificates, but that can just be exercised by picking a different certificate here than the one in the DataProvider PublicKey resource.
</blockquote></details>

A map allows more flexibility in the future in case an EDP wants to use different signing keys for different MCs or some other use case we can't think of yet. As it is, an EDP will want to often register two certificates - one for the signing of the public key and one for measurement result consent. It makes since to pass in both here. it also allows the EDP to input all the current certs it views as valid.
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt` line 155 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1313#-NjFKuGcBKgmIviS2jA2:-NjI8164C-CHEImIM1tu:bb7f2mf) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f1c095d77a8d73ddc334680fe45db2df07b02685/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt#L155)):*

Since the EDP should always specify the result certificate, I think we want this to be non-nullable. If we really don't want the caller to have to specify it, we can default it to be the first certificate in `registeredCertificates`.
</blockquote></details>

done
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt` line 379 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1313#-NjFLY1i69PYUmtcsviT:-NjI8EcX9SoVkDoXam0b:b-dl4j7g) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f1c095d77a8d73ddc334680fe45db2df07b02685/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt#L379)):*

This is not necessary, and furthermore not even feasible after https://github.com/world-federation-of-advertisers/cross-media-measurement-api/pull/187.

The key is signed so that the MC can verify it came from the EDP.
</blockquote></details>

i agree that its temporary. We can drop it after 187 is merged.
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt` line 381 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1313#-NjFMSDD0Y3vsAfkVvai:-NjI8fMoDn8v3xlTOqCL:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f1c095d77a8d73ddc334680fe45db2df07b02685/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt#L381)):*

This shouldn't be a TODO, since we don't necessarily plan on implementing it for the EDP simulator. We're just trying to indicate something that the simulator does not do but a real EDP should.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/resourcesetup/ResourceSetup.kt` line 217 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1313#-NjFNw_k3vCcWIuxsqzt:-NjI97kw361ZybPal1wp:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f1c095d77a8d73ddc334680fe45db2df07b02685/src/main/kotlin/org/wfanet/measurement/loadtest/resourcesetup/ResourceSetup.kt#L217)):*

Also output the EDP certificate resource name definition here.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/resourcesetup/ResourceSetup.kt` line 243 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1313#-NjFQ9OtACebT-wo9VU3:-NjIKxQhA9LA-gYggtSF:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f1c095d77a8d73ddc334680fe45db2df07b02685/src/main/kotlin/org/wfanet/measurement/loadtest/resourcesetup/ResourceSetup.kt#L243)):*

Why not just return the internal `DataProvider`?
</blockquote></details>

Done.
___
*[`src/main/proto/wfa/measurement/internal/kingdom/requisitions_service.proto` line 86 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1313#-NjFQZJ53s0dhmIn1IXx:-NjIyiUs0Ey7ygogxOI1:bb7p5eq) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f1c095d77a8d73ddc334680fe45db2df07b02685/src/main/proto/wfa/measurement/internal/kingdom/requisitions_service.proto#L86)):*

nit: certificates are used to verify signatures, not to sign
</blockquote></details>

thx!
___
*[`src/main/kotlin/org/wfanet/measurement/api/v2alpha/tools/BUILD.bazel` line 23 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1313#-NjFHtca1gmX3HtObZEE:-NjKWh81CUbzBhcfh2Of:bqezvc3) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f1c095d77a8d73ddc334680fe45db2df07b02685/src/main/kotlin/org/wfanet/measurement/api/v2alpha/tools/BUILD.bazel#L23)):*

These shouldn't be separate binaries and should instead be subcommands of `MeasurementSystem`. That CLI tool is meant to cover methods from the CMMS public API.
</blockquote></details>

didnt realize this. thx! done.

SanjayVas (changes requested):
> Reviewed 2 of 51 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1313)* status: 13 of 62 files reviewed, 10 unresolved discussions (waiting on @Marco-Premier and @stevenwarejones)

*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1313#-NjiB1JOC0zo2-6IVnu4:-NjiB1JP6pDihArWjp50:b-6vvsog) (no related file):*
I left multiple comments for this, but summarizing it here: The change is just so that the EDP specifies which cert is used to verify the result. There is no need for our tests to have more than one cert registered for an EDP, since the same cert can be used for multiple purposes. Registering another cert just adds unnecessary complexity.

A single internal API service test is sufficient to verify that the cert returned with the Measurement resource is the one that the EDP specified when fulfilling.

___
___
*[`build/variables.bzl` line 47 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1313#-Nji9cSI4jB-vXy0i2sB:-Nji9cSI4jB-vXy0i2sC:b-cmia6p) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/961c641339c598017c816e8f296de01a6fd320df/build/variables.bzl#L47)):*
> ```Starlark
>     edp1_name = "$(edp1_name)",
>     edp1_cert_name = "$(edp1_cert_name)",
>     edp1_result_cert_name = "$(edp1_result_cert_name)",
> ```

Revert this change. We only need one EDP cert resource for testing.
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt` line 149 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1313#-Ni_ymGa8zH6OOlgAAqa:-Nji7lzIAFHl4WtU-uPl:b43heuf) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8558ffc784e66584248d581255c8c03dd74b6724/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt#L149)):*
Real EDPs will have many certificates as they rotate. The simulator however only ever signs the Measurement result, so it doesn't need any other signing keys or certificates. Put another way, the simulator will only ever use one entry in the map.

> As it is, an EDP will want to often register two certificates - one for the signing of the public key and one for measurement result consent.

This isn't necessary. An EDP *can* use different certs, but that's entirely optional. Mainly it's that the EDP needs to be able to specify which cert they want to use for the result rather than being bound by the MC choice of the pre-shared key.
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt` line 379 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1313#-NjFLY1i69PYUmtcsviT:-Nji8efd0wA3IfTMj7Kl:b-z2p30g) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f1c095d77a8d73ddc334680fe45db2df07b02685/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt#L379)):*

i agree that its temporary. We can drop it after 187 is merged.
</blockquote></details>

That PR has already been merged.

It also wasn't necessary even before that PR was merged. The EDP never needed to verify this signature since it's the entity that generated the signature in the first place. This is why the PR drops it.
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/resourcesetup/ResourceSetup.kt` line 163 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1313#-NjiATSw24FhPO7aUEFd:-NjiATSw24FhPO7aUEFe:b8th6jj) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/961c641339c598017c816e8f296de01a6fd320df/src/main/kotlin/org/wfanet/measurement/loadtest/resourcesetup/ResourceSetup.kt#L163)):*
> ```Kotlin
>               displayName = it.displayName
>               certificate = externalDataProviderCertificateKeyName
>               resultCertificate =
> ```

Why did this change? Resource setup doesn't need to generate another EDP cert, as that is entirely unnecessary for our tests. The only change that was needed here was to output the name of the EDP cert.
___
*[`src/main/proto/wfa/measurement/loadtest/resourcesetup/resources.proto` line 37 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1313#-NjiCvHp8r0UlvAB8I18:-NjiCvHp8r0UlvAB8I19:b-cmia6p) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/961c641339c598017c816e8f296de01a6fd320df/src/main/proto/wfa/measurement/loadtest/resourcesetup/resources.proto#L37)):*
> ```proto
>       string display_name = 2;
>       string certificate = 3;
>       string result_certificate = 4;
> ```

Revert this change. We only need one EDP cert resource for testing.

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1313)* status: 13 of 62 files reviewed, 10 unresolved discussions (waiting on @Marco-Premier and @SanjayVas)

*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1313#-NjiB1JOC0zo2-6IVnu4:-NjnJHi8-uXfYJ9VOg0X:bbb1n1c) (no related file):*

I left multiple comments for this, but summarizing it here: The change is just so that the EDP specifies which cert is used to verify the result. There is no need for our tests to have more than one cert registered for an EDP, since the same cert can be used for multiple purposes. Registering another cert just adds unnecessary complexity.

A single internal API service test is sufficient to verify that the cert returned with the Measurement resource is the one that the EDP specified when fulfilling.
</blockquote></details>

Sanjay - the EDP doesn't need to specify two certificates. They just need to specify which resource to use for their data provider registration and which certificate resource to use for result signing. They can be the same resource. I think this behavior is fine - not requiring an EDP to have two separate certificates, but, instead, require them to explicitly set the resource to be used for each use case.

___
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/resourcesetup/ResourceSetup.kt` line 163 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1313#-NjiATSw24FhPO7aUEFd:-NjnJvbj0PFi4b1GIeUx:b-4ly1ak) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/961c641339c598017c816e8f296de01a6fd320df/src/main/kotlin/org/wfanet/measurement/loadtest/resourcesetup/ResourceSetup.kt#L163)):*

Why did this change? Resource setup doesn't need to generate another EDP cert, as that is entirely unnecessary for our tests. The only change that was needed here was to output the name of the EDP cert.
</blockquote></details>

See my top comment. This change doesn't require the EDP to generate a 2nd certificate. It just requires them to explicitly set which certificate to use for result signing, even if its the same resource name as the one they registered initially.

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1313)* status: 13 of 62 files reviewed, 10 unresolved discussions (waiting on @Marco-Premier and @stevenwarejones)

*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1313#-NjiB1JOC0zo2-6IVnu4:-NjndvyX9GE3f94TO-qm:b-l59ej8) (no related file):*
> the EDP doesn't need to specify two certificates. They just need to specify which resource to use for their data provider registration and which certificate resource to use for result signing.

A real EDP needs to do this. AFAICT the correctness test just needs a single certificate resource and private key pair. The EDP simulator only needs to be able to sign the result and has no other use for a certificate resource.

> They can be the same resource.

And should be for the correctness test. This means that resource setup should result in only one DataProvider Certificate resource.

___
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/resourcesetup/ResourceSetup.kt` line 163 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1313#-NjiATSw24FhPO7aUEFd:-Njngcbo927D2H36-RGn:b2c1frx) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/961c641339c598017c816e8f296de01a6fd320df/src/main/kotlin/org/wfanet/measurement/loadtest/resourcesetup/ResourceSetup.kt#L163)):*

See my top comment. This change doesn't require the EDP to generate a 2nd certificate. It just requires them to explicitly set which certificate to use for result signing, even if its the same resource name as the one they registered initially.
</blockquote></details>

My point is that resource setup should never call CreateDataProviderCertificate. Only one DataProvider certificate resource is needed for the correctness test, and it's automatically created as part of CreateDataProvider. Furthermore the use case of the certificate resource is irrelevant to resource setup - consumers of the resources that it creates decide what to use them for.

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1313)* status: 13 of 62 files reviewed, 10 unresolved discussions (waiting on @Marco-Premier and @SanjayVas)

*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1313#-NjiB1JOC0zo2-6IVnu4:-Njoqlb12c3eIOkdP6IW:bavpdu4) (no related file):*

> the EDP doesn't need to specify two certificates. They just need to specify which resource to use for their data provider registration and which certificate resource to use for result signing.

A real EDP needs to do this. AFAICT the correctness test just needs a single certificate resource and private key pair. The EDP simulator only needs to be able to sign the result and has no other use for a certificate resource.

> They can be the same resource.

And should be for the correctness test. This means that resource setup should result in only one DataProvider Certificate resource.
</blockquote></details>

Resource setup will result in only one DataProvider Certificate Resource if that data provider doesn't pass a separate result certificate. Both resource fields will be populated but with the same resource name.

___

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1313)* status: 13 of 62 files reviewed, 10 unresolved discussions (waiting on @Marco-Premier and @stevenwarejones)

*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1313#-NjiB1JOC0zo2-6IVnu4:-NjsACe97L-a9BNCAEtw:bz9nasy) (no related file):*
> Resource setup will result in only one DataProvider Certificate Resource if that data provider doesn't pass a separate result certificate.

I'm not sure what you mean by "if that data provider doesn't pass". The data provider has no involvement with resource setup. The only function of resource setup is to set up the minimal resources needed for the correctness test. The code path for creating an additional DataProvider certificate will never be used and should be reverted.

___

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1313)* status: 13 of 62 files reviewed, 10 unresolved discussions (waiting on @Marco-Premier and @stevenwarejones)

*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1313#-NjiB1JOC0zo2-6IVnu4:-NkGUN_4-TXHFFokvVTt:bjeqtou) (no related file):*
> What I meant is that in the config, additional certificates are passed in as flags for each EDP

This is the part that is unnecessary and should be reverted. The EDP simulator only uses one certificate-key pair: the one used to sign the result. It never interacts with and needs no knowledge of the certificate associated with the encryption public key.

> Shouldn't the correctness test exercise the new code?

It in fact exclusively uses the new code path. It always passes a certificate to FulfillDirectRequisition.

I think you're conflating two things:

1. Updating the EDP simulator and therefore the correctness test for the API change
2. Verifying that the certificate passed to FulfillDirectRequisition is indeed used for the Measurement result and *not* the certificate associated with the DataProvider PublicKey resource by specifying different certificates for each.

(2) can be sufficiently covered by a single test method in the internal service unit test. It need not and should not be covered in the correctness test, which only needs to cover a small set of end-to-end cases. We especially want resource setup to be minimal and not create any more resources than necessary. This PR was mostly done at r3.

___

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1313)* status: 10 of 62 files reviewed, 9 unresolved discussions (waiting on @Marco-Premier and @SanjayVas)

*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1313#-NjiB1JOC0zo2-6IVnu4:-Nkayess0jDuiLQhYcut:b-896fix) (no related file):*

> What I meant is that in the config, additional certificates are passed in as flags for each EDP

This is the part that is unnecessary and should be reverted. The EDP simulator only uses one certificate-key pair: the one used to sign the result. It never interacts with and needs no knowledge of the certificate associated with the encryption public key.

> Shouldn't the correctness test exercise the new code?

It in fact exclusively uses the new code path. It always passes a certificate to FulfillDirectRequisition.

I think you're conflating two things:

1. Updating the EDP simulator and therefore the correctness test for the API change
2. Verifying that the certificate passed to FulfillDirectRequisition is indeed used for the Measurement result and *not* the certificate associated with the DataProvider PublicKey resource by specifying different certificates for each.

(2) can be sufficiently covered by a single test method in the internal service unit test. It need not and should not be covered in the correctness test, which only needs to cover a small set of end-to-end cases. We especially want resource setup to be minimal and not create any more resources than necessary. This PR was mostly done at r3.
</blockquote></details>

Done.

___
___
*[`build/variables.bzl` line 47 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1313#-Nji9cSI4jB-vXy0i2sB:-Nkayg16AiR27FaA-Ap6:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/961c641339c598017c816e8f296de01a6fd320df/build/variables.bzl#L47)):*

Revert this change. We only need one EDP cert resource for testing.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt` line 149 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1313#-Ni_ymGa8zH6OOlgAAqa:-NkayjDn6Fmo7RdXKqnB:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8558ffc784e66584248d581255c8c03dd74b6724/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt#L149)):*

Real EDPs will have many certificates as they rotate. The simulator however only ever signs the Measurement result, so it doesn't need any other signing keys or certificates. Put another way, the simulator will only ever use one entry in the map.

> As it is, an EDP will want to often register two certificates - one for the signing of the public key and one for measurement result consent.

This isn't necessary. An EDP *can* use different certs, but that's entirely optional. Mainly it's that the EDP needs to be able to specify which cert they want to use for the result rather than being bound by the MC choice of the pre-shared key.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt` line 379 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1313#-NjFLY1i69PYUmtcsviT:-Nkayk3PDmSxr7ycr-Qq:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f1c095d77a8d73ddc334680fe45db2df07b02685/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt#L379)):*

That PR has already been merged.

It also wasn't necessary even before that PR was merged. The EDP never needed to verify this signature since it's the entity that generated the signature in the first place. This is why the PR drops it.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/resourcesetup/ResourceSetup.kt` line 163 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1313#-NjiATSw24FhPO7aUEFd:-Nkayn-a1R7XII1AfHAe:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/961c641339c598017c816e8f296de01a6fd320df/src/main/kotlin/org/wfanet/measurement/loadtest/resourcesetup/ResourceSetup.kt#L163)):*

My point is that resource setup should never call CreateDataProviderCertificate. Only one DataProvider certificate resource is needed for the correctness test, and it's automatically created as part of CreateDataProvider. Furthermore the use case of the certificate resource is irrelevant to resource setup - consumers of the resources that it creates decide what to use them for.
</blockquote></details>

Done.
___
*[`src/main/proto/wfa/measurement/loadtest/resourcesetup/resources.proto` line 37 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1313#-NjiCvHp8r0UlvAB8I18:-NkayqbX03fmeE8zAfXC:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/961c641339c598017c816e8f296de01a6fd320df/src/main/proto/wfa/measurement/loadtest/resourcesetup/resources.proto#L37)):*

Revert this change. We only need one EDP cert resource for testing.
</blockquote></details>

Done.

SanjayVas (changes requested):
> Reviewed 2 of 37 files at r4, 14 of 51 files at r5, 19 of 27 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1313)* status: 43 of 62 files reviewed, 8 unresolved discussions (waiting on @Marco-Premier and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/api/v2alpha/tools/MeasurementSystem.kt` line 992 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1313#-NkbABQxAgvPikx1ydPI:-NkbABQy66SsuP7dDVWF:bs37emo) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f651876137de62b743a1af14122d6b4ce390e459/src/main/kotlin/org/wfanet/measurement/api/v2alpha/tools/MeasurementSystem.kt#L992)):*
> ```Kotlin
>   @Command(name = "get", description = ["gets a DataProvider"])
>   fun getDataProvider(
>     @Option(names = ["--out", "-o"], description = ["Output file"], required = true) outFile: File,
> ```

This is inconsistent with the rest of the commands which output to stdout.

Note that this tool is just for debugging and is not intended to be used for scripting, so the output need not be machine-readable to anything other than the unit test.



_Code quote:_
```Kotlin
@Option(names = ["--out", "-o"], description = ["Output file"], required = true) outFile: File,
```
___
*[`src/main/kotlin/org/wfanet/measurement/integration/common/InProcessCmmsComponents.kt` line 88 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1313#-NkbAqMW20FCPSdJjYxR:-NkbAqMW20FCPSdJjYxS:b-pcz8jj) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f651876137de62b743a1af14122d6b4ce390e459/src/main/kotlin/org/wfanet/measurement/integration/common/InProcessCmmsComponents.kt#L88)):*
> ```Kotlin
>         DataProviderCertificateKey.fromName(resource.dataProvider.certificate)!!
>       val validCertificates = mapOf(dataProviderCertificateKey to edpContent.signingKey)
>       val certificateKey = DataProviderCertificateKey.fromName(resource.dataProvider.certificate)!!
> ```

`certificateKey` and `dataProviderCertificateKey` appear to be the same
___
*[`src/main/kotlin/org/wfanet/measurement/integration/common/InProcessLifeOfAnEventGroupIntegrationTest.kt` line 108 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1313#-NkbR6FWE2aZVCxvJFaI:-NkbR6FWE2aZVCxvJFaJ:b-fxdbyg) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f651876137de62b743a1af14122d6b4ce390e459/src/main/kotlin/org/wfanet/measurement/integration/common/InProcessLifeOfAnEventGroupIntegrationTest.kt#L108)):*
> ```Kotlin
>     val externalDataProviderId = externalIdToApiId(internalDataProvider.externalDataProviderId)
>     val externalDataProviderResourceName = DataProviderKey(externalDataProviderId).toName()
>     edpResourceName = externalDataProviderResourceName
> ```

nit: inconsistent naming and unnecessary variable



_Suggestion:_
```Kotlin
    val dataProviderId = externalIdToApiId(internalDataProvider.externalDataProviderId)
    edpResourceName = DataProviderKey(dataProviderId).toName()
```
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt` line 149 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1313#-Ni_ymGa8zH6OOlgAAqa:-NkbBMIEBxo51gwMAngM:b-r6tt3j) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8558ffc784e66584248d581255c8c03dd74b6724/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt#L149)):*

Done.
</blockquote></details>

This is marked as "Done" but appears unchanged (there's still a map despite there only ever being one certificate-key pair used by the EDP simulator)
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/resourcesetup/ResourceSetup.kt` line 154 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1313#-NkbO4fz1wYeHQXN-BUX:-NkbO4fz1wYeHQXN-BUY:bwm8n93) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f651876137de62b743a1af14122d6b4ce390e459/src/main/kotlin/org/wfanet/measurement/loadtest/resourcesetup/ResourceSetup.kt#L154)):*
> ```Kotlin
>         externalIdToApiId(internalDataProvider.certificate.externalCertificateId)
>       val externalDataProviderCertificateKeyName =
>         DataProviderCertificateKey(externalDataProviderId, externalCertificateId).toName()
> ```

nit: Naming consistency. Resource IDs in the API are different from external IDs, despite one being an encoding of the other.

Also specify types explicitly when not obvious. See https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/main/docs/code-style.md#kotlin



_Suggestion:_
```Kotlin
      val internalDataProvider: InternalDataProvider = createInternalDataProvider(it)
      val dataProviderId: String = externalIdToApiId(internalDataProvider.externalDataProviderId)
      val dataProviderResourceName: String = DataProviderKey(dataProviderId).toName()
      val certificateId: String =
        externalIdToApiId(internalDataProvider.certificate.externalCertificateId)
      val dataProviderCertificateKeyName: String =
        DataProviderCertificateKey(dataProviderId, certificateId).toName()
```
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/resourcesetup/ResourceSetup.kt` line 430 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1313#-NkbQ57e055ZDbCXc9DW:-NkbPbL29-ud9P4s6qvM:bkohetg) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f651876137de62b743a1af14122d6b4ce390e459/src/main/kotlin/org/wfanet/measurement/loadtest/resourcesetup/ResourceSetup.kt#L430)):*
> ```Kotlin
>   /** The consent signaling encryption key. */
>   val encryptionPublicKey: EncryptionPublicKey,
>   /** The private encryption key. */
> ```

This isn't created by ResourceSetup.
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/resourcesetup/ResourceSetupFlags.kt` line 73 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1313#-NkbPyJR7s_h5aOEzDSY:-NkbPyJR7s_h5aOEzDSZ:b8w3vzw) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f651876137de62b743a1af14122d6b4ce390e459/src/main/kotlin/org/wfanet/measurement/loadtest/resourcesetup/ResourceSetupFlags.kt#L73)):*
> ```Kotlin
>     required = false
>   )
>   lateinit var edpEncryptionPrivateKeysets: Map<String, File>
> ```

AFAICT ResourceSetup doesn't need this
___
*[`src/test/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulatorTest.kt` line 2205 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1313#-NkbQcHV4GTMVrRuLQGe:-NkbQcHV4GTMVrRuLQGf:b54obiq) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f651876137de62b743a1af14122d6b4ce390e459/src/test/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulatorTest.kt#L2205)):*
> ```Kotlin
>         EDP_DISPLAY_NAME,
>         loadEncryptionPrivateKey("${EDP_DISPLAY_NAME}_enc_private.tink"),
>         mapOf(
> ```

EDP simulator only uses one signing key, so we don't need this.

SanjayVas (changes requested):
> Reviewed 4 of 27 files at r6, 3 of 3 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1313)* status: 48 of 62 files reviewed, 5 unresolved discussions (waiting on @Marco-Premier and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/api/v2alpha/tools/MeasurementSystem.kt` line 992 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1313#-NkbABQxAgvPikx1ydPI:-NkvwE6Q4vVOqQjzB_yb:bw6lc54) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f651876137de62b743a1af14122d6b4ce390e459/src/main/kotlin/org/wfanet/measurement/api/v2alpha/tools/MeasurementSystem.kt#L992)):*
> would you prefer the entire protobuf just get echoed?

Yes.

> EncryptionPublicKeys.kt in this same directory outputs to a file.

I was referring to the subcommands of this command. That outputs a file because it's outputting in binary format to be used as an input for other calls.
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt` line 149 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1313#-Ni_ymGa8zH6OOlgAAqa:-Nkvyq7A0vud_uMj0OGF:b-b6uzgm) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8558ffc784e66584248d581255c8c03dd74b6724/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt#L149)):*

I think a map is the appropriate input type here. It doesn't seem so overly complex that its overkill.
</blockquote></details>

It's not a question of being overkill but rather that it is misleading. It makes it appear that the EDP simulator uses more than one SigningKeyHandle when this is never the case. That is to say that the EDP simulator doesn't need a collection of valid certificates, but rather just a SigningKeyHandle that matches `certificateKey`.
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/resourcesetup/ResourceSetup.kt` line 430 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1313#-NkbQ57e055ZDbCXc9DW:-NkvzSCi-4qDUVa-voqH:bi57wzg) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f651876137de62b743a1af14122d6b4ce390e459/src/main/kotlin/org/wfanet/measurement/loadtest/resourcesetup/ResourceSetup.kt#L430)):*

I don't think I understand this comment. The signingKey is also not created by ResourceSetup.
</blockquote></details>

Sorry, what I meant is that ResourceSetup doesn't use this field. (EntityContent is a type used to provide input to ResourceSetup).
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/resourcesetup/ResourceSetupFlags.kt` line 73 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1313#-NkbPyJR7s_h5aOEzDSY:-Nkw0aTHDEqLkqwQU9cw:b-ovgur5) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f651876137de62b743a1af14122d6b4ce390e459/src/main/kotlin/org/wfanet/measurement/loadtest/resourcesetup/ResourceSetupFlags.kt#L73)):*

I prefer it this way vs hard-coding it in the InProcessEdpSimulator
</blockquote></details>

I'm not sure what you mean by "preferring it this way". The only flags to ResourceSetup should be for things that ResourceSetup actually needs, i.e. inputs to create the API resources.

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1313)* status: 41 of 62 files reviewed, 5 unresolved discussions (waiting on @Marco-Premier and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/api/v2alpha/tools/MeasurementSystem.kt` line 992 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1313#-NkbABQxAgvPikx1ydPI:-Nl08MV_6IvyVxzBblUy:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f651876137de62b743a1af14122d6b4ce390e459/src/main/kotlin/org/wfanet/measurement/api/v2alpha/tools/MeasurementSystem.kt#L992)):*

> would you prefer the entire protobuf just get echoed?

Yes.

> EncryptionPublicKeys.kt in this same directory outputs to a file.

I was referring to the subcommands of this command. That outputs a file because it's outputting in binary format to be used as an input for other calls.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt` line 149 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1313#-Ni_ymGa8zH6OOlgAAqa:-Nl08dGy2H8xj6eop-ap:b-16d2yg) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8558ffc784e66584248d581255c8c03dd74b6724/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt#L149)):*

It's not a question of being overkill but rather that it is misleading. It makes it appear that the EDP simulator uses more than one SigningKeyHandle when this is never the case. That is to say that the EDP simulator doesn't need a collection of valid certificates, but rather just a SigningKeyHandle that matches `certificateKey`.
</blockquote></details>

removed the map
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/resourcesetup/ResourceSetup.kt` line 430 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1313#-NkbQ57e055ZDbCXc9DW:-Nl0eJ4GAz-fi573wqDo:ba7u6a) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f651876137de62b743a1af14122d6b4ce390e459/src/main/kotlin/org/wfanet/measurement/loadtest/resourcesetup/ResourceSetup.kt#L430)):*

Sorry, what I meant is that ResourceSetup doesn't use this field. (EntityContent is a type used to provide input to ResourceSetup).
</blockquote></details>

so, maybe there's a better way of going about this, but I didn't like how the InProcessEdpSimulator has hard-coded paths to its encryptionKey and signingKey. 

Instead, I pass those values through the constructor. I thought the best way to populate that data was to put it in EntityContent

Lmk what you think would be better.

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1313)* status: 41 of 62 files reviewed, 5 unresolved discussions (waiting on @Marco-Premier and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/resourcesetup/ResourceSetupFlags.kt` line 73 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1313#-NkbPyJR7s_h5aOEzDSY:-Nl3rN_O6IZDd_TdCsPC:by7sexd) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f651876137de62b743a1af14122d6b4ce390e459/src/main/kotlin/org/wfanet/measurement/loadtest/resourcesetup/ResourceSetupFlags.kt#L73)):*

I'm not sure what you mean by "preferring it this way". The only flags to ResourceSetup should be for things that ResourceSetup actually needs, i.e. inputs to create the API resources.
</blockquote></details>

The ResourceSetupRunner currently does more than just set up API resources. It also is builds the channels, makes sure each party has the right cert. See my other comment about how the InProcessSimulator currently has a hard-coded path to certificates. These changes allow these paths to get passed in via flags.

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1313)* status: 33 of 64 files reviewed, 3 unresolved discussions (waiting on @Marco-Premier and @SanjayVas)

___
*[`src/main/k8s/testing/secretfiles/BUILD.bazel` line 159 at r9](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1313#-Nl4_bf66C3ReeagRT2I:-Nl8oKEx5IEKmUB5aSJb:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a48b464f7c59e8a254045472df5ceee2fd357ca0/src/main/k8s/testing/secretfiles/BUILD.bazel#L159)):*

What are all these extra result certs for? We only need one additional cert-key pair for the internal service test.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/resourcesetup/ResourceSetup.kt` line 430 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1313#-NkbQ57e055ZDbCXc9DW:-Nl8omwX7kIls_WVusRX:bfndu7s) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f651876137de62b743a1af14122d6b4ce390e459/src/main/kotlin/org/wfanet/measurement/loadtest/resourcesetup/ResourceSetup.kt#L430)):*

I don't see a problem with the private key being loaded in InProcessEdpSimulator with hard-coded paths. It's mirroring what the EDP simulator runner should do in terms of reading the private key from a flag.

The problem with the design in this PR is that it's passing input to ResourceSetup that it doesn't use. The private key input is for EdpSimulator, *not* ResourceSetup.
</blockquote></details>

okay. I'll revert this for now. We can revisit later.
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/resourcesetup/ResourceSetupFlags.kt` line 73 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1313#-NkbPyJR7s_h5aOEzDSY:-Nl8oivbDRHd06SPxhzP:bfndu7s) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f651876137de62b743a1af14122d6b4ce390e459/src/main/kotlin/org/wfanet/measurement/loadtest/resourcesetup/ResourceSetupFlags.kt#L73)):*

> The ResourceSetupRunner currently does more than just set up API resources.

Not really.

> It also is builds the channels

No it doesn't. The client stubs are passed into the constructor, meaning the channels are created by the caller.

> makes sure each party has the right cert

Only insofar as what is needed for resource creation.

> InProcessSimulator currently has a hard-coded path to certificates. These changes allow these paths to get passed in via flags.

Not sure what you mean. InProcessEdpSimulator doesn't have any command-line flags. If you mean you'd rather have them passed in as constructor args to InProcessEdpSimulator rather than having InProcessEdpSimulator load them itself then that's fine, but you don't need to pass the keys through ResourceSetup to make that change.

As an aside, that it would make it inconsistent with all of the other InProcess components, which load their own test keys. Basically in the current design InProcessX takes on the same role as XRunner, i.e. indicating what files to use.
</blockquote></details>

okay. I'll revert this for now. We can revisit later.

SanjayVas (changes requested):
> Reviewed 24 of 31 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1313)* status: 57 of 64 files reviewed, 2 unresolved discussions (waiting on @Marco-Premier and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt` line 150 at r10](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1313#-NlPSFo14RfpVXqNSSx4:-NlPSFo14RfpVXqNSSx5:b-mnvdf6) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a6aa37a946d2e843a78d06677b456a557e0e5af1/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt#L150)):*
> ```Kotlin
>   val displayName: String,
>   /** The EDP's encryption key. */
>   val publicEncryptionKey: PublicKeyHandle,
> ```

The private key contains the public key. You don't need to pass it separately.

See the `publicKey` property of `PrivateKeyHandle`.

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1313)* status: 55 of 64 files reviewed, 2 unresolved discussions (waiting on @Marco-Premier and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt` line 150 at r10](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1313#-NlPSFo14RfpVXqNSSx4:-NlQ0cQsCWn-U8MRXAtH:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a6aa37a946d2e843a78d06677b456a557e0e5af1/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt#L150)):*

The private key contains the public key. You don't need to pass it separately.

See the `publicKey` property of `PrivateKeyHandle`.
</blockquote></details>

Done.

SanjayVas (approved):
> Reviewed 3 of 31 files at r10, 6 of 6 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1313)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Marco-Premier)

SanjayVas:
This introduced a new compiler warning:

> ```
> src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt:1517:9: warning: variable 'requisitionCertificateKey' is never used
>     val requisitionCertificateKey =
> ```

PR #1319, opened 2023-11-01
Limit the number of responses from StreamActiveComputations calls.

SanjayVas:
Fixes #1318

PR #1321, opened 2023-11-06
Limit the number of responses from StreamActiveComputations calls. (#1319)

SanjayVas:
This eliminates the issue of a high number of rows being returned, which should be sufficient to address the latency issue. Moving the Duchy ID filtration into the WHERE clause should also help with the number of rows scanned, but there is still potential to have a high number due to the ORDER BY clause.

(cherry picked from commit e8dbecb4c398ebfa20be2a993ff14cb1362ed768)


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1321)

PR #1324, opened 2023-11-07
Move K8s service account creation for GKE out of Terraform.

SanjayVas:
Due to Terraform limitations, the Kubernetes provider may give errors when making changes to the underlying cluster resource. The provider documentation indicates that Kubernetes resources should not be configured in the same root module or apply step as the cluster resources.


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1324)

SanjayVas (commented):
Possibly. The basic issue is if your Kubernetes provider is configured using interpolation from other resources. It looks like the AWS modules do that as well. I ran into an error when running `terraform plan` tweaking something in the cluster config. It couldn't read the endpoint from the cluster resource, so the Kubernetes provider defaulted to `localhost` and failed.

See [Stacking with managed Kubernetes cluster resources](https://registry.terraform.io/providers/hashicorp/kubernetes/latest/docs#stacking-with-managed-kubernetes-cluster-resources)

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1324#-:-NipHTET4ctuMc5t5AQS:bcehop2)* status: all files reviewed, 1 unresolved discussion (waiting on @YuhongWang-Amazon)

___
*[`src/main/k8s/dev/edp_simulator_gke.cue` line 55 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1324#-NiohDj1B5YalqOT54Vo:-NipH1_E02XrD1tn8iY4:b-s1sv9d) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0316c233ec870b1fac0444041f7dfaf9ac0f90c0/src/main/k8s/dev/edp_simulator_gke.cue#L55)):*

It seems that the following spec will be removed after this change. Could it cause the pod to be scheduled on wrong nodes?
```
  nodeSelector:    iam.gke.io/gke-metadata-server-enabled: "true"
```
</blockquote></details>

It means the pod is free to be scheduled on nodes regardless of whether they have the GKE metadata server (necessary for workload identity). All of our node pools have this enabled anyway (as specified in the Terraform config).

YuhongWang-Amazon (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1324)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @SanjayVas)

PR #1340, opened 2023-11-17
initial commit

bdomen-ggl:


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1340)

SanjayVas (changes requested):
> Reviewed 2 of 4 files at r1, 1 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1340)* status: 3 of 4 files reviewed, 2 unresolved discussions (waiting on @bdomen-ggl)

___
*[`experimental/reporting-ui/src/main/kotlin/org/wfanet/measurement/reporting/bff/service/api/v1alpha/ReportsService.kt` line 43 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1340#-NkWgozDCFWuTErzCHmW:-NkWgozDCFWuTErzCHmX:b-grl76a) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d2a23fcd00178828f91012c09593cbbb8dd53741/experimental/reporting-ui/src/main/kotlin/org/wfanet/measurement/reporting/bff/service/api/v1alpha/ReportsService.kt#L43)):*
> ```Kotlin
>     val resp = runBlocking(Dispatchers.IO) { haloReportsStub.listReports(haloRequest) }
> 
>     // Not doing pagination right now
> ```

Add a TODO and throw an error if the caller specifies a page_size or page_token for now. Might also want to log a warning if the Reporting response has a next_page_token since we're not handling that.
___
*[`src/main/proto/wfa/measurement/reporting/bff/v1alpha/list_report_detail.proto` line 28 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1340#-NkWk3dV1lexBKoGvoYz:-NkWk3dV1lexBKoGvoZ-:b-8euy1u) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d2a23fcd00178828f91012c09593cbbb8dd53741/src/main/proto/wfa/measurement/reporting/bff/v1alpha/list_report_detail.proto#L28)):*
> ```proto
> option go_package = "wfa/measurement/reporting/bff/v1alpha/reportingpb";
> 
> message ListReportDetail {
> ```

For the BFF API you should still use Report if your method is ListReports. You can use either field masks or views to limit what is returned.

For the UI, my naming recommendation would be ReportListItem.



_Code quote:_
```proto
ListReportDetail
```

bdomen-ggl (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1340)* status: 0 of 4 files reviewed, 2 unresolved discussions (waiting on @SanjayVas)

___
*[`experimental/reporting-ui/src/main/kotlin/org/wfanet/measurement/reporting/bff/service/api/v1alpha/ReportsService.kt` line 43 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1340#-NkWgozDCFWuTErzCHmW:-NkXCDeo8AG_7Dg4e9Ax:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d2a23fcd00178828f91012c09593cbbb8dd53741/experimental/reporting-ui/src/main/kotlin/org/wfanet/measurement/reporting/bff/service/api/v1alpha/ReportsService.kt#L43)):*

Add a TODO and throw an error if the caller specifies a page_size or page_token for now. Might also want to log a warning if the Reporting response has a next_page_token since we're not handling that.
</blockquote></details>

Done.
___
*[`src/main/proto/wfa/measurement/reporting/bff/v1alpha/list_report_detail.proto` line 28 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1340#-NkWk3dV1lexBKoGvoYz:-NkX7DVU3bKtvhYAjDq1:b35y8hu) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d2a23fcd00178828f91012c09593cbbb8dd53741/src/main/proto/wfa/measurement/reporting/bff/v1alpha/list_report_detail.proto#L28)):*

For the BFF API you should still use Report if your method is ListReports. You can use either field masks or views to limit what is returned.

For the UI, my naming recommendation would be ReportListItem.
</blockquote></details>

For my clarity, the BFF is calling to list reports from Halo (potentially looking into using the field masks if we have time), which returns a `Report`. The BFF then will transform the Halo `Report` into a more UI friendly `ReportListItem` the UI can then easily use to display a list of Reports.

My takeaway for this PR is to rename `ListReportDetail` -> `ReportListItem`.

bdomen-ggl (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1340)* status: 1 of 5 files reviewed, 6 unresolved discussions (waiting on @SanjayVas)

___
*[`experimental/reporting-ui/src/main/kotlin/org/wfanet/measurement/reporting/bff/service/api/v1alpha/ReportsService.kt` line 36 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1340#-Nkaqfh-51azLIHcdXhD:-Nkb7G_S2C6799xcu9xn:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5df64c7f51f9fe0d99cb963353c4bebd2978262d/experimental/reporting-ui/src/main/kotlin/org/wfanet/measurement/reporting/bff/service/api/v1alpha/ReportsService.kt#L36)):*

This annotation is only for compatibility with a Java caller. I don't think we expect anyone to be hand-writing Java code that calls this service impl.
</blockquote></details>

Done.
___
*[`experimental/reporting-ui/src/main/kotlin/org/wfanet/measurement/reporting/bff/service/api/v1alpha/ReportsService.kt` line 41 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1340#-Nkar4zJ3VQN_S2joCDe:-Nkb7JHtDYrzcCj8yrBi:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5df64c7f51f9fe0d99cb963353c4bebd2978262d/experimental/reporting-ui/src/main/kotlin/org/wfanet/measurement/reporting/bff/service/api/v1alpha/ReportsService.kt#L41)):*

I think your pageSize condition is backwards.
</blockquote></details>

Done.
___
*[`experimental/reporting-ui/src/main/kotlin/org/wfanet/measurement/reporting/bff/service/api/v1alpha/ReportsService.kt` line 42 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1340#-NkarF6V3b-nhIlBRZQs:-Nkb7JYb4pP1EJSBa6lK:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5df64c7f51f9fe0d99cb963353c4bebd2978262d/experimental/reporting-ui/src/main/kotlin/org/wfanet/measurement/reporting/bff/service/api/v1alpha/ReportsService.kt#L42)):*

Use a gRPC error.
</blockquote></details>

Done.
___
*[`experimental/reporting-ui/src/main/kotlin/org/wfanet/measurement/reporting/bff/service/api/v1alpha/ReportsService.kt` line 52 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1340#-NkaqY9x0CXkgwIqeQAA:-Nkb7K-F5iL_8rdDacCL:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5df64c7f51f9fe0d99cb963353c4bebd2978262d/experimental/reporting-ui/src/main/kotlin/org/wfanet/measurement/reporting/bff/service/api/v1alpha/ReportsService.kt#L52)):*
> ```Kotlin
>     val resp = runBlocking(Dispatchers.IO) { haloReportsStub.listReports(haloRequest) }
> 
>     if (resp.nextPageToken.length > 0) {
> ```

Done.
___
*[`src/main/proto/wfa/measurement/reporting/bff/v1alpha/reports_service.proto` line 77 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1340#-Nkaq33mAnxi6Ni4RO-A:-Nkb7RKb-p3sU7zsRavY:b-je6gu0) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5df64c7f51f9fe0d99cb963353c4bebd2978262d/src/main/proto/wfa/measurement/reporting/bff/v1alpha/reports_service.proto#L77)):*

Don't use the optional keyword here. Use the proto3 default behavior where "unset" is the same as "set to default value".

Use of the optional keyword should be rare in proto3 APIs and only for cases where you need to differentiate the two cases (e.g. when 0 or empty string is a valid value that it distinct from "unset")
</blockquote></details>

Oops yes, sorry was bouncing between a couple thoughts. Ended up not wanting to use it but forgot to remove it.
___
*[`src/main/proto/wfa/measurement/reporting/bff/v1alpha/list_report_detail.proto` line 28 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1340#-NkWk3dV1lexBKoGvoYz:-Nkb3RWB6XQpzz4rfb5X:b6jto9) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d2a23fcd00178828f91012c09593cbbb8dd53741/src/main/proto/wfa/measurement/reporting/bff/v1alpha/list_report_detail.proto#L28)):*

The BFF API is just another resource-oriented API that follows the AIPs.

> My takeaway for this PR is to rename `ListReportDetail` -> `ReportListItem`.

If that's the case, then the method should be `ListReportListItems`, meaning you're having a separate resource in the BFF API for `ReportListItem` with a separate pattern (`measurementConsumers/{measurement_consumer}/reportListItems/{report_list_item}`) where the `report_list_item` ID segment happens to use the same IDs as the `Report` resource. While this is valid, are you sure it's what you want?

My assumption was that you instead want to have different views for Report in the BFF API: one that you use by default when showing a list of reports, and another that you use for showing the full Report. In this case, the "ReportListItem" distinction would be in the UI as a different model/view.
</blockquote></details>

So I think I want the first option you said - a different resource. The reason is a specific consideration is to NOT just have different views for listing and showing the charts and tables, but to return different objects. Specifically we don't want to return the whole report object as over time this could cause the app to respond poorly and even crash with the amount of data we could be sending.

Also there were suggestions that the ListReportListItem could also have some processed data to show other specific summary data instead of just the basics name and status. Going with the pattern that we want the UI to be a thin client, I think it makes sense to have the backend define the ListReportListItem as an API. That way we're not sending extra data to both the list page and the get page.

bdomen-ggl:
*[`experimental/reporting-ui/src/main/react/reporting-ui/client/reporting/client_impl.ts` line 34 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1340#-Nkq31YyE61uFyYNcP5N:-Nkq31YyE61uFyYNcP5O:b8ienxo) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f62a9c308613a4210b2e9cf74312a1157b95b1a4/experimental/reporting-ui/src/main/react/reporting-ui/client/reporting/client_impl.ts#L34)):*
> ```XML
> 
>   async listReports(): Promise<ListReportsResponse> {
>     const res = await fetch(this.baseUrl.toString() + '/api/reportListItems');
> ```

Yes, this technically isn't correct. There is a follow up Draft PR to correctly connect the UI to the BFF.

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1340)* status: 1 of 5 files reviewed, 7 unresolved discussions (waiting on @bdomen-ggl)

___
*[`src/main/proto/wfa/measurement/reporting/bff/v1alpha/list_report_detail.proto` line 28 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1340#-NkWk3dV1lexBKoGvoYz:-Nkqthm9ADaJ6qNLjwDY:b-afra7a) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d2a23fcd00178828f91012c09593cbbb8dd53741/src/main/proto/wfa/measurement/reporting/bff/v1alpha/list_report_detail.proto#L28)):*

So I think I want the first option you said - a different resource. The reason is a specific consideration is to NOT just have different views for listing and showing the charts and tables, but to return different objects. Specifically we don't want to return the whole report object as over time this could cause the app to respond poorly and even crash with the amount of data we could be sending.

Also there were suggestions that the ListReportListItem could also have some processed data to show other specific summary data instead of just the basics name and status. Going with the pattern that we want the UI to be a thin client, I think it makes sense to have the backend define the ListReportListItem as an API. That way we're not sending extra data to both the list page and the get page.
</blockquote></details>

Note that I'm referring to a [view enumeration](https://google.aip.dev/157#view-enumeration) for partial responses in the API here, not to UI views. This allows for exactly what you are indicating: limiting the amount of data returned by the server.

bdomen-ggl (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1340)* status: 0 of 6 files reviewed, 8 unresolved discussions (waiting on @bdomen-ggl and @SanjayVas)

*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1340#-Nkv8zTG9QX5uNeakEBM:-Nkv8zTG9QX5uNeakEBN:b-ju9dnd) (no related file):*
@SanjayVas Ready for next review. Reverted the API resource change and added the view. I've not implemented one before, does that look right? Also gutted the report model. It was a copy of the V1Alpha report but will look much different. I just added a few pieces that the list call can use.

___
___
*[`src/main/proto/wfa/measurement/reporting/bff/v1alpha/list_report_detail.proto` line 28 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1340#-NkWk3dV1lexBKoGvoYz:-Nkumzlg8ki9A2_LpPmm:b-ehpvi8) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d2a23fcd00178828f91012c09593cbbb8dd53741/src/main/proto/wfa/measurement/reporting/bff/v1alpha/list_report_detail.proto#L28)):*

Note that I'm referring to a [view enumeration](https://google.aip.dev/157#view-enumeration) for partial responses in the API here, not to UI views. This allows for exactly what you are indicating: limiting the amount of data returned by the server.
</blockquote></details>

Had a discussion offline. Will go back to ListReports and add the view enumeration.

SanjayVas (changes requested):
> Reviewed 5 of 6 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1340)* status: 5 of 6 files reviewed, 7 unresolved discussions (waiting on @bdomen-ggl)

___
*[`experimental/reporting-ui/src/main/kotlin/org/wfanet/measurement/reporting/bff/service/api/v1alpha/ReportsService.kt` line 88 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1340#-Nkw1pxH8brJ68oACLjW:-Nkw1pxIBOYqOzJEvhE0:b5rerrf) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/be7d74b1a8740bebbb465d9cbf334e233df8c309/experimental/reporting-ui/src/main/kotlin/org/wfanet/measurement/reporting/bff/service/api/v1alpha/ReportsService.kt#L88)):*
> ```Kotlin
>     }
> 
>   private fun convertHaloReportToBasicReport(haloReport: HaloReport): Report {
> ```

nit: We generally implement these conversions as extensions (receiver methods) on the source type. e.g.

```kotlin
private fun HaloReport.toBasicReport(): Report {
  val source = this // Pattern we use to avoid `this` ambiguity if needed
  return report {
    name = source.name
    // ...
  }
}
```
___
*[`src/main/proto/wfa/measurement/reporting/bff/v1alpha/report.proto` line 39 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1340#-Nkw3_Mj8jamNvTaPHcY:-Nkw3_Mj8jamNvTaPHcZ:blap1tn) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/be7d74b1a8740bebbb465d9cbf334e233df8c309/src/main/proto/wfa/measurement/reporting/bff/v1alpha/report.proto#L39)):*
> ```proto
>   string id = 1 [(google.api.field_behavior) = OUTPUT_ONLY];
> 
>   string name = 2 [(google.api.field_behavior) = OUTPUT_ONLY];
> ```

* `name` generally goes first.
* `name` is not inherently OUTPUT_ONLY, e.g. if the standard Update method is implemented then the `name` field is used as input.
* The ID field name should be prefixed with the resource type.



_Suggestion:_
```proto
  string name = 1;

  string report_id = 2 [(google.api.field_behavior) = OUTPUT_ONLY];
```
___
*[`src/main/proto/wfa/measurement/reporting/bff/v1alpha/reports_service.proto` line 54 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1340#-Nkw4FPs4DhOz5PLh6We:-Nkw4FPs4DhOz5PLh6Wf:undefined) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/be7d74b1a8740bebbb465d9cbf334e233df8c309/src/main/proto/wfa/measurement/reporting/bff/v1alpha/reports_service.proto#L54)):*
> ```proto
>   REPORT_VIEW_UNSPECIFIED = 0;
> 
>   // Include basic metadata about the book, but not the full contents.
> ```





_Suggestion:_
```proto
report
```
___
*[`src/main/proto/wfa/measurement/reporting/bff/v1alpha/reports_service.proto` line 63 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1340#-Nkw4KHICx0-cNkO9O0r:-Nkw4KHICx0-cNkO9O0s:bkjneaq) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/be7d74b1a8740bebbb465d9cbf334e233df8c309/src/main/proto/wfa/measurement/reporting/bff/v1alpha/reports_service.proto#L63)):*
> ```proto
> // Request message for `GetReport` method.
> message GetReportRequest {
>   ReportView view = 1;
> ```

nit: name should go first as field 1 since it's a mandatory field for Get request message types.
___
*[`src/main/proto/wfa/measurement/reporting/bff/v1alpha/reports_service.proto` line 63 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1340#-Nkw5bu-7sL5kfzX2tnX:-Nkw5bu04psQ0Wa3X1V2:b-or5j7l) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/be7d74b1a8740bebbb465d9cbf334e233df8c309/src/main/proto/wfa/measurement/reporting/bff/v1alpha/reports_service.proto#L63)):*
> ```proto
> // Request message for `GetReport` method.
> message GetReportRequest {
>   ReportView view = 1;
> ```

Document the defaults. Unlike the example in the AIP, I assume we'd use FULL as the default for Get but BASIC as the default for List.

bdomen-ggl (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1340)* status: 2 of 6 files reviewed, 5 unresolved discussions (waiting on @bdomen-ggl and @SanjayVas)

___
*[`experimental/reporting-ui/src/main/kotlin/org/wfanet/measurement/reporting/bff/service/api/v1alpha/ReportsService.kt` line 88 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1340#-Nkw1pxH8brJ68oACLjW:-Nkwi3Wj2r48xbkcPLDv:bsuj9x1) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/be7d74b1a8740bebbb465d9cbf334e233df8c309/experimental/reporting-ui/src/main/kotlin/org/wfanet/measurement/reporting/bff/service/api/v1alpha/ReportsService.kt#L88)):*

nit: We generally implement these conversions as extensions (receiver methods) on the source type. e.g.

```kotlin
private fun HaloReport.toBasicReport(): Report {
  val source = this // Pattern we use to avoid `this` ambiguity if needed
  return report {
    name = source.name
    // ...
  }
}
```
</blockquote></details>

Done.

Also should I do this for the request objects too? I'm translating the BFF requests to the Halo requests. Might be OK like that since that direction is 1:1?
___
*[`src/main/proto/wfa/measurement/reporting/bff/v1alpha/report.proto` line 39 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1340#-Nkw3_Mj8jamNvTaPHcY:-NkwNcoE4vK6tdlr1Ao3:bfr0ruk) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/be7d74b1a8740bebbb465d9cbf334e233df8c309/src/main/proto/wfa/measurement/reporting/bff/v1alpha/report.proto#L39)):*

* `name` generally goes first.
* `name` is not inherently OUTPUT_ONLY, e.g. if the standard Update method is implemented then the `name` field is used as input.
* The ID field name should be prefixed with the resource type.
</blockquote></details>

Ah yes, copied from the previously made list item which was immutable.

Done.
___
*[`src/main/proto/wfa/measurement/reporting/bff/v1alpha/reports_service.proto` line 54 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1340#-Nkw4FPs4DhOz5PLh6We:-NkwNiJE0QCXJw15kuzc:b3l9a0g) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/be7d74b1a8740bebbb465d9cbf334e233df8c309/src/main/proto/wfa/measurement/reporting/bff/v1alpha/reports_service.proto#L54)):*
> ```proto
>   REPORT_VIEW_UNSPECIFIED = 0;
> 
>   // Include basic metadata about the book, but not the full contents.
> ```

Ah I missed a copypasta. Thanks

Done.
___
*[`src/main/proto/wfa/measurement/reporting/bff/v1alpha/reports_service.proto` line 63 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1340#-Nkw4KHICx0-cNkO9O0r:-NkwRNZS9V5hTg0jviWm:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/be7d74b1a8740bebbb465d9cbf334e233df8c309/src/main/proto/wfa/measurement/reporting/bff/v1alpha/reports_service.proto#L63)):*

nit: name should go first as field 1 since it's a mandatory field for Get request message types.
</blockquote></details>

Done.
___
*[`src/main/proto/wfa/measurement/reporting/bff/v1alpha/reports_service.proto` line 63 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1340#-Nkw5bu-7sL5kfzX2tnX:-NkwRP-O7VmFB4cbW779:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/be7d74b1a8740bebbb465d9cbf334e233df8c309/src/main/proto/wfa/measurement/reporting/bff/v1alpha/reports_service.proto#L63)):*

Document the defaults. Unlike the example in the AIP, I assume we'd use FULL as the default for Get but BASIC as the default for List.
</blockquote></details>

Done.

bdomen-ggl (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1340)* status: 4 of 6 files reviewed, 1 unresolved discussion (waiting on @bdomen-ggl and @SanjayVas)

*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1340#-Nl4cOIN34Xo-mJbVQ3q:-Nl4cOIN34Xo-mJbVQ3r:blf7tcx) (no related file):*
@SanjayVas Oops, I made the changes, but looks like something is still pending.

___

bdomen-ggl (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1340)* status: all files reviewed, 2 unresolved discussions (waiting on @bdomen-ggl and @SanjayVas)

___
*[`src/main/proto/wfa/measurement/reporting/bff/v1alpha/report.proto` line 39 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1340#-Nl4cwm663gdGc5mS93J:-Nl4lgwq-gsLxMlXMDp_:b-181upg) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e688a45983fc706484e18f25f45d16c4f7c0a6f1/src/main/proto/wfa/measurement/reporting/bff/v1alpha/report.proto#L39)):*

Why did you drop the OUTPUT_ONLY here?

From AIP-122:

> -   Resources **may** provide the resource ID as a separate field (e.g. `book_id`). This field **must** apply the [`OUTPUT_ONLY`](https://google.aip.dev/203#output-only) field behavior classification.
</blockquote></details>

Oh dang thanks. I was having some build issues and didn't put it back.

SanjayVas (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1340)* status: all files reviewed, 1 unresolved discussion (waiting on @bdomen-ggl)

SanjayVas:
This introduced a new compiler warning:

> ```
> INFO: From KotlinCompile //experimental/reporting-ui/src/main/kotlin/org/wfanet/measurement/reporting/bff/service/api/v1alpha:reports_service { kt: 1, java: 0, srcjars: 0 } for k8:
> experimental/reporting-ui/src/main/kotlin/org/wfanet/measurement/reporting/bff/service/api/v1alpha/ReportsService.kt:61:22: warning: 'getter for tags: (Mutable)Map<String!, String!>!' is deprecated. Deprecated in Java
>         .filter { it.tags.containsKey("ui.halo-cmm.org") }
>                      ^
> ```

PR #1343, opened 2023-11-17
Add and Implementing List Event Groups in the BFF

bdomen-ggl:


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1343)

PR #1344, opened 2023-11-18
Add Create Report to the BFF

bdomen-ggl:
Adding the Create Report API stub to the BFF.

Also updating the BFF API protos (http annotations and proto messages) to exclude measurementConsumer/parent since that will be built into the service. The protobufs aren't currently used so rearranging the values won't break anything.

### DIFF
--- a/experimental/reporting-ui/src/main/kotlin/org/wfanet/measurement/reporting/bff/deploy/common/server/ReportingApiServerFlags.kt
+++ b/experimental/reporting-ui/src/main/kotlin/org/wfanet/measurement/reporting/bff/deploy/common/server/ReportingApiServerFlags.kt
@@ -45,4 +45,11 @@ class ReportingApiServerFlags {
   )
   var debugVerboseGrpcClientLogging by Delegates.notNull<Boolean>()
     private set
+
+  @set:CommandLine.Option(
+    names = ["--measurement-consumer"],
+    description = ["measurement consumer id of the service owner"],
+    required = true,
+  )
+  lateinit var measurementConsumer: String
 }
--- a/experimental/reporting-ui/src/main/kotlin/org/wfanet/measurement/reporting/bff/deploy/common/server/V1AlphaPublicApiServer.kt
+++ b/experimental/reporting-ui/src/main/kotlin/org/wfanet/measurement/reporting/bff/deploy/common/server/V1AlphaPublicApiServer.kt
@@ -53,7 +53,13 @@ private fun run(
       .withVerboseLogging(reportingApiServerFlags.debugVerboseGrpcClientLogging)
 
   val services: List<ServerServiceDefinition> =
-    listOf(ReportsService(HaloReportsGrpcKt.ReportsCoroutineStub(channel)).bindService())
+    listOf(
+      ReportsService(
+          HaloReportsGrpcKt.ReportsCoroutineStub(channel),
+          reportingApiServerFlags.measurementConsumer
+        )
+        .bindService()
+    )
   CommonServer.fromFlags(commonServerFlags, SERVER_NAME, services).start().blockUntilShutdown()
 }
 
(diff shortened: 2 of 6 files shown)


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1344)

bdomen-ggl (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1344)* status: 0 of 6 files reviewed, 1 unresolved discussion

___
*[`experimental/reporting-ui/src/main/kotlin/org/wfanet/measurement/reporting/bff/service/api/v1alpha/ReportsService.kt` line 95 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1344#-Nl5WnKt7YG9LUC6f5bj:-Nl5WnKt7YG9LUC6f5bk:b-6mpaor) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/91dc85be6b3928500a824847ef9eb488fc370157/experimental/reporting-ui/src/main/kotlin/org/wfanet/measurement/reporting/bff/service/api/v1alpha/ReportsService.kt#L95)):*
> ```Kotlin
>     }
> 
>     try {
> ```

@SanjayVas I've seen public APIs and error handling implemented in multiple ways. How would you suggest we handle errors here? At the end of the day, just want a way to let the user know if the create request was successfully submitted or not, which all these methods can do.
1. Throw any errors directly from the backend service. Normally not a fan of this, but since we're just wrapping the public Reporting service, I think this is fine. (IE. no try/catch)
2. Catch errors, wrap them, and rethrow them. Usually this is to hide internal error messages that might expose security/architecture. Again since we're wrapping a public API, don't think we need to do this. We may get transient errors on connection failures though.
3. Public APIs shouldn't throw errors, return an error code/message instead. I think either throwing errors or not is fine, but behavior should be documented so caller knows what to do.

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1344)* status: 0 of 6 files reviewed, 2 unresolved discussions (waiting on @bdomen-ggl)

___
*[`experimental/reporting-ui/src/main/kotlin/org/wfanet/measurement/reporting/bff/service/api/v1alpha/ReportsService.kt` line 95 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1344#-Nl5WnKt7YG9LUC6f5bj:-NlPVoh9D1InvQFItW4M:b-1yw20f) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/91dc85be6b3928500a824847ef9eb488fc370157/experimental/reporting-ui/src/main/kotlin/org/wfanet/measurement/reporting/bff/service/api/v1alpha/ReportsService.kt#L95)):*
From our [style guide](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/main/docs/code-style.md#kotlin):

> -   Catch `StatusException` from RPCs immediately at the call site.
>      -   This prevents gRPC statuses from being incorrectly propagated from gRPC servers that are also gRPC clients.
>      -   If you need to re-throw the exception to be handled at a higher level, wrap it in another exception type.

Basically we should be following the AIPs when it comes to errors. In practice this means throwing our own StatusRuntimeExceptions. If our service impl is calling another gRPC service as we're doing here, the correct thing to do is catch StatusException specifically and inspect it to determine which Status code to use in the resulting StatusRuntimeException, falling back to UNKNOWN when we don't know enough to give more info.
___
*[`experimental/reporting-ui/src/main/kotlin/org/wfanet/measurement/reporting/bff/service/api/v1alpha/ReportsService.kt` line 96 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1344#-NlPWGh29BfSrkmNVPQA:-NlPWGh29BfSrkmNVPQB:bjmrybg) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/91dc85be6b3928500a824847ef9eb488fc370157/experimental/reporting-ui/src/main/kotlin/org/wfanet/measurement/reporting/bff/service/api/v1alpha/ReportsService.kt#L96)):*
> ```Kotlin
> 
>     try {
>       val resp = runBlocking(Dispatchers.IO) { backendReportsStub.createReport(backendRequest) }
> ```

Never use runBlocking from coroutine code. Also no need to switch dispatchers here, as the stub is also coroutine-aware. We generally only need to do this (e.g. using `withContext`) when directly calling blocking code such as filesystem I/O from a coroutine.



_Code quote:_
```Kotlin
runBlocking
```

PR #1347, opened 2023-11-22
Add a metric reuse mechanism based on timed ReportingMetricEntries

riemanli:
Given the fact that `MetricCalculationSpec` is now an individual resource and can be reused, we can utilize this to identify a set of metrics across reports. Note that the granularity is "a set of metrics" that is associated with `MetricCalculationSpec`.

When a `MetricCalculationSpec` is reused in a different report, by identifying the pair of `MetricCalculationSpec` and the `ReportingSet` and the `TimeInterval`s from the report, we are able to identify the existing set of metrics created from the previous reports. This could be done during the internal report creation by the following:
1. Left join the table `Metrics` and `MetricCalculationSpecReportingMetrics` by `MeasurementConsumerId` and `MetricId`.
2. Find the set of metrics given `MeasurementConsumerId`, `ReportingSetId`, `MetricCalculationSpecId`, a set of `TimeIntervalStart` and `TimeIntervalEndExclusive`.
3. Repeat the process for all `ReportingMetricEntries`, i.e. pairs of `MetricCalculationSpecs` and `ReportingSets`.

At the step (2) above, the reason why we could get a set of metrics is because a `MetricCalculationSpec` could correspond to multiple metrics due to the fan-out of metric types, groupings, and time intervals.

### DIFF
--- a/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/postgres/readers/MetricReader.kt
+++ b/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/postgres/readers/MetricReader.kt
@@ -21,6 +21,7 @@ import com.google.type.Interval
 import com.google.type.interval
 import io.r2dbc.postgresql.codec.Interval as PostgresInterval
 import java.time.Instant
+import java.time.ZoneOffset
 import java.util.UUID
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.emptyFlow
@@ -30,6 +31,7 @@ import org.wfanet.measurement.common.db.r2dbc.ReadContext
 import org.wfanet.measurement.common.db.r2dbc.ResultRow
 import org.wfanet.measurement.common.db.r2dbc.boundStatement
 import org.wfanet.measurement.common.identity.InternalId
+import org.wfanet.measurement.common.toInstant
 import org.wfanet.measurement.common.toProtoDuration
 import org.wfanet.measurement.common.toProtoTime
 import org.wfanet.measurement.internal.reporting.v2.BatchGetMetricsRequest
@@ -50,6 +52,21 @@ class MetricReader(private val readContext: ReadContext) {
     val metricId: InternalId,
     val createMetricRequestId: String,
     val metric: Metric
+  )
+
+  data class ReportingMetricKey(
+    val reportingSetId: InternalId,
+    val metricCalculationSpecId: InternalId,
+    val timeInterval: Interval,
+  )
+
+  data class ReportingMetric(
+    val reportingMetricKey: ReportingMetricKey,
+    val metricId: InternalId,
+    val createMetricRequestId: String,
+    val externalMetricId: String,
+    val metricSpec: MetricSpec,
+    val metricDetails: Metric.Details
   )
 
   private data class MetricInfo(
@@ -159,9 +176,10 @@ class MetricReader(private val readContext: ReadContext) {
       StringBuilder(
         """
           $baseSqlSelect
-          FROM MeasurementConsumers
+          FROM
+            MeasurementConsumers
             JOIN Metrics USING(MeasurementConsumerId)
-          $baseSqlJoins
+            $baseSqlJoins
           WHERE Metrics.MeasurementConsumerId = $1
             AND CreateMetricRequestId IN
         """
@@ -204,6 +222,264 @@ class MetricReader(private val readContext: ReadContext) {
         )
       }
     }
+  }
+
+  fun readReportingMetricsByReportingMetricKey(
+    measurementConsumerId: InternalId,
+    reportingMetricKeys: Collection<ReportingMetricKey>,
+  ): Flow<ReportingMetric> {
+    if (reportingMetricKeys.isEmpty()) {
+      return emptyFlow()
+    }
+
+    val sqlSelect: String =
+      """
+      SELECT
+        CmmsMeasurementConsumerId,
+        Metrics.MeasurementConsumerId,
+        Metrics.CreateMetricRequestId,
+        ReportingSets.ReportingSetId as MetricsReportingSetId,
+        MetricCalculationSpecReportingMetrics.MetricCalculationSpecId,
+        Metrics.MetricId,
+        Metrics.ExternalMetricId,
+        Metrics.TimeIntervalStart AS MetricsTimeIntervalStart,
+        Metrics.TimeIntervalEndExclusive AS MetricsTimeIntervalEndExclusive,
+        Metrics.MetricType,
+        Metrics.DifferentialPrivacyEpsilon,
+        Metrics.DifferentialPrivacyDelta,
+        Metrics.FrequencyDifferentialPrivacyEpsilon,
+        Metrics.FrequencyDifferentialPrivacyDelta,
+        Metrics.MaximumFrequency,
+        Metrics.MaximumFrequencyPerUser,
+        Metrics.MaximumWatchDurationPerUser,
+        Metrics.VidSamplingIntervalStart,
+        Metrics.VidSamplingIntervalWidth,
+        Metrics.MetricDetails
+      """
+        .trimIndent()
+
+    val sqlJoins: String =
+      """
+      JOIN Metrics USING(MeasurementConsumerId)
+      JOIN ReportingSets USING(MeasurementConsumerId, ReportingSetId)
+      LEFT JOIN MetricCalculationSpecReportingMetrics ON
+        Metrics.MeasurementConsumerId = MetricCalculationSpecReportingMetrics.MeasurementConsumerId
+        AND Metrics.MetricId = MetricCalculationSpecReportingMetrics.MetricId
+      """
+        .trimIndent()
+
+    val sql =
+      StringBuilder(
+        """
+          $sqlSelect
+          FROM
+            MeasurementConsumers
+            $sqlJoins
+          WHERE Metrics.MeasurementConsumerId = $1
+        """
+          .trimIndent()
+      )
+
+    // The index in `sql` ends at $1.
+    var offset = 2
+    val reportingSetIdBindingMap = mutableMapOf<InternalId, String>()
+    val metricCalculationSpecIdBindingMap = mutableMapOf<InternalId, String>()
+    val timeStampBindingMap = mutableMapOf<Timestamp, String>()
+    val sqlWhereConditions =
+      reportingMetricKeys.joinToString(separator = " OR ", prefix = "\nAND (", postfix = ")") {
+        timedReportingMetricKey ->
+        val reportingSetIdIndex =
+          reportingSetIdBindingMap.getOrPut(timedReportingMetricKey.reportingSetId) {
+            "$${offset++}"
+          }
+        val metricCalculationSpecIdIndex =
+          metricCalculationSpecIdBindingMap.getOrPut(
+            timedReportingMetricKey.metricCalculationSpecId
+          ) {
+            "$${offset++}"
+          }
+        val timeIntervalStartIndex =
+          timeStampBindingMap.getOrPut(timedReportingMetricKey.timeInterval.startTime) {
+            "$${offset++}"
+          }
+        val timeIntervalEndExclusiveIndex =
+          timeStampBindingMap.getOrPut(timedReportingMetricKey.timeInterval.endTime) {
+            "$${offset++}"
+          }
+        "(Metrics.ReportingSetId = $reportingSetIdIndex AND " +
+          "MetricCalculationSpecReportingMetrics.MetricCalculationSpecId = $metricCalculationSpecIdIndex AND " +
+          "Metrics.TimeIntervalStart = $timeIntervalStartIndex AND " +
+          "Metrics.TimeIntervalEndExclusive = $timeIntervalEndExclusiveIndex)"
+      }
+
+    sql.append(sqlWhereConditions)
+
+    val statement =
+      boundStatement(sql.toString()) {
+        bind("$1", measurementConsumerId)
+        reportingMetricKeys.forEach { timedReportingMetricKey ->
+          val reportingSetIdIndex =
+            reportingSetIdBindingMap.getValue(timedReportingMetricKey.reportingSetId)
+          val metricCalculationSpecIdIndex =
+            metricCalculationSpecIdBindingMap.getValue(
+              timedReportingMetricKey.metricCalculationSpecId
+            )
+          val timeIntervalStartIndex =
+            timeStampBindingMap.getValue(timedReportingMetricKey.timeInterval.startTime)
+          val timeIntervalEndExclusiveIndex =
+            timeStampBindingMap.getValue(timedReportingMetricKey.timeInterval.endTime)
+          bind(reportingSetIdIndex, timedReportingMetricKey.reportingSetId)
+          bind(metricCalculationSpecIdIndex, timedReportingMetricKey.metricCalculationSpecId)
+          bind(
+            timeIntervalStartIndex,
+            timedReportingMetricKey.timeInterval.startTime.toInstant().atOffset(ZoneOffset.UTC)
+          )
+          bind(
+            timeIntervalEndExclusiveIndex,
+            timedReportingMetricKey.timeInterval.endTime.toInstant().atOffset(ZoneOffset.UTC)
+          )
+        }
+      }
+
+    return flow {
+      val reportingMetricMap: Map<String, ReportingMetric> = buildReportingMetricMap(statement)
+      for (entry in reportingMetricMap) {
+        emit(entry.value)
+      }
+    }
+  }
+
+  /**
+   * Returns a map that maintains the order of the query result for timed reporting metric entry.
+   */
+  private suspend fun buildReportingMetricMap(
+    statement: BoundStatement
+  ): Map<String, ReportingMetric> {
+    // Key is externalMetricId.
+    val reportingMetricMap: MutableMap<String, ReportingMetric> = linkedMapOf()
+
+    val translate: (row: ResultRow) -> Unit = { row: ResultRow ->
+      val createMetricRequestId: String =
+        requireNotNull(row["CreateMetricRequestId"]) {
+          "Metric that is associated with a MetricCalculationSpec must have createMetricRequestId"
+        }
+      val reportingSetId: InternalId = row["MetricsReportingSetId"]
+      val metricCalculationSpecId: InternalId = row["MetricCalculationSpecId"]
+      val metricId: InternalId = row["MetricId"]
+      val externalMetricId: String = row["ExternalMetricId"]
+      val metricTimeIntervalStart: Instant = row["MetricsTimeIntervalStart"]
+      val metricTimeIntervalEnd: Instant = row["MetricsTimeIntervalEndExclusive"]
+      val metricType: MetricSpec.TypeCase = MetricSpec.TypeCase.forNumber(row["MetricType"])
+      val differentialPrivacyEpsilon: Double = row["DifferentialPrivacyEpsilon"]
+      val differentialPrivacyDelta: Double = row["DifferentialPrivacyDelta"]
+      val frequencyDifferentialPrivacyEpsilon: Double? = row["FrequencyDifferentialPrivacyEpsilon"]
+      val frequencyDifferentialPrivacyDelta: Double? = row["FrequencyDifferentialPrivacyDelta"]
+      val maximumFrequency: Int? = row["MaximumFrequency"]
+      val maximumFrequencyPerUser: Int? = row["MaximumFrequencyPerUser"]
+      val maximumWatchDurationPerUser: PostgresInterval? = row["MaximumWatchDurationPerUser"]
+      val vidSamplingStart: Float = row["VidSamplingIntervalStart"]
+      val vidSamplingWidth: Float = row["VidSamplingIntervalWidth"]
+      val metricDetails: Metric.Details =
+        row.getProtoMessage("MetricDetails", Metric.Details.parser())
+
+      reportingMetricMap.computeIfAbsent(externalMetricId) {
+        val metricTimeInterval = interval {
+          startTime = metricTimeIntervalStart.toProtoTime()
+          endTime = metricTimeIntervalEnd.toProtoTime()
+        }
+
+        val vidSamplingInterval =
+          MetricSpecKt.vidSamplingInterval {
+            start = vidSamplingStart
+            width = vidSamplingWidth
+          }
+
+        val metricSpec = metricSpec {
+          when (metricType) {
+            MetricSpec.TypeCase.REACH ->
+              reach =
+                MetricSpecKt.reachParams {
+                  privacyParams =
+                    MetricSpecKt.differentialPrivacyParams {
+                      epsilon = differentialPrivacyEpsilon
+                      delta = differentialPrivacyDelta
+                    }
+                }
+            MetricSpec.TypeCase.REACH_AND_FREQUENCY -> {
+              if (
+                frequencyDifferentialPrivacyDelta == null ||
+                  frequencyDifferentialPrivacyEpsilon == null ||
+                  maximumFrequency == null
+              ) {
+                throw IllegalStateException()
+              }
+
+              reachAndFrequency =
+                MetricSpecKt.reachAndFrequencyParams {
+                  reachPrivacyParams =
+                    MetricSpecKt.differentialPrivacyParams {
+                      epsilon = differentialPrivacyEpsilon
+                      delta = differentialPrivacyDelta
+                    }
+                  frequencyPrivacyParams =
+                    MetricSpecKt.differentialPrivacyParams {
+                      epsilon = frequencyDifferentialPrivacyEpsilon
+                      delta = frequencyDifferentialPrivacyDelta
+                    }
+                  this.maximumFrequency = maximumFrequency
+                }
+            }
+            MetricSpec.TypeCase.IMPRESSION_COUNT -> {
+              if (maximumFrequencyPerUser == null) {
+                throw IllegalStateException()
+              }
+
+              impressionCount =
+                MetricSpecKt.impressionCountParams {
+                  privacyParams =
+                    MetricSpecKt.differentialPrivacyParams {
+                      epsilon = differentialPrivacyEpsilon
+                      delta = differentialPrivacyDelta
+                    }
+                  this.maximumFrequencyPerUser = maximumFrequencyPerUser
+                }
+            }
+            MetricSpec.TypeCase.WATCH_DURATION -> {
+              watchDuration =
+                MetricSpecKt.watchDurationParams {
+                  privacyParams =
+                    MetricSpecKt.differentialPrivacyParams {
+                      epsilon = differentialPrivacyEpsilon
+                      delta = differentialPrivacyDelta
+                    }
+                  this.maximumWatchDurationPerUser =
+                    checkNotNull(maximumWatchDurationPerUser).duration.toProtoDuration()
+                }
+            }
+            MetricSpec.TypeCase.TYPE_NOT_SET -> throw IllegalStateException()
+          }
+          this.vidSamplingInterval = vidSamplingInterval
+        }
+
+        ReportingMetric(
+          reportingMetricKey =
+            ReportingMetricKey(
+              reportingSetId = reportingSetId,
+              metricCalculationSpecId = metricCalculationSpecId,
+              timeInterval = metricTimeInterval,
+            ),
+          createMetricRequestId = createMetricRequestId,
+          metricId = metricId,
+          externalMetricId = externalMetricId,
+          metricSpec = metricSpec,
+          metricDetails = metricDetails,
+        )
+      }
+    }
+
+    readContext.executeQuery(statement).consume(translate).collect {}
+
+    return reportingMetricMap
   }
 
   fun batchGetMetrics(
--- a/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/postgres/writers/CreateReport.kt
+++ b/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/postgres/writers/CreateReport.kt
@@ -30,9 +30,11 @@ import org.wfanet.measurement.common.toJson
 import org.wfanet.measurement.common.toProtoTime
 import org.wfanet.measurement.internal.reporting.v2.CreateReportRequest
 import org.wfanet.measurement.internal.reporting.v2.Report
+import org.wfanet.measurement.internal.reporting.v2.ReportKt
 import org.wfanet.measurement.internal.reporting.v2.copy
 import org.wfanet.measurement.reporting.deploy.v2.postgres.readers.MeasurementConsumerReader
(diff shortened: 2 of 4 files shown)


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1347)

riemanli:
* main
  * **PR #1347** <a href="https://app.graphite.dev/github/pr/world-federation-of-advertisers/cross-media-measurement/1347?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/world-federation-of-advertisers/cross-media-measurement/1347?utm_source=stack-comment).

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1347)* status: all files reviewed, 4 unresolved discussions (waiting on @riemanli)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/postgres/readers/MetricReader.kt` line 230 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1347#-NkH2Au-7KFKEU3hH2y6:-NkH2Au-7KFKEU3hH2y7:b-k669h2) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/491a555f7b45eab43ef4347bc59794a2e2b4cd8c/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/postgres/readers/MetricReader.kt#L230)):*
> ```Kotlin
>           LEFT JOIN MetricCalculationSpecReportingMetrics
>             ON Metrics.MeasurementConsumerId = MetricCalculationSpecReportingMetrics.MeasurementConsumerId
>             AND Metrics.MetricId = MetricCalculationSpecReportingMetrics.MetricId
> ```

nit: keep consistent indentation for FROM clause



_Suggestion:_
```Kotlin
          FROM
            MeasurementConsumers
            JOIN Metrics USING(MeasurementConsumerId)
            $baseSqlJoins
            LEFT JOIN MetricCalculationSpecReportingMetrics ON
              Metrics.MeasurementConsumerId = MetricCalculationSpecReportingMetrics.MeasurementConsumerId
              AND Metrics.MetricId = MetricCalculationSpecReportingMetrics.MetricId
```
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/postgres/readers/MetricReader.kt` line 234 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1347#-NkH2lZ9EN8cyg_9RWBs:-NkH2lZ9EN8cyg_9RWBt:b-u54b3z) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/491a555f7b45eab43ef4347bc59794a2e2b4cd8c/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/postgres/readers/MetricReader.kt#L234)):*
> ```Kotlin
>             AND Metrics.ReportingSetId = $2
>             AND MetricCalculationSpecReportingMetrics.MetricCalculationSpecId = $3
>             AND
> ```

nit: append the `AND` when you append the additional conjunct so this reads clearer. The hanging `AND` here looks like a bug at first glance without careful reading.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/postgres/writers/CreateReport.kt` line 118 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1347#-NkH439hDSm0DCy5DssT:-NkH439iBBGyDR3VrF-p:b-k34t1w) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/491a555f7b45eab43ef4347bc59794a2e2b4cd8c/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/postgres/writers/CreateReport.kt#L118)):*
> ```Kotlin
> 
>     // External ID to MetricCalculationSpecReader.Result
>     val metricCalculationSpecReaderResultMap: Map<String, MetricCalculationSpecReader.Result> =
> ```

nit: accurate naming can avoid the need for explanatory comments. Names also need not be maximally verbose (e.g. the "ReaderResult" part).



_Suggestion:_
```Kotlin
    val metricCalculationSpecsByExternalId: Map<String, MetricCalculationSpecReader.Result> =
```
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/internal/testing/v2/ReportsServiceTest.kt` line 505 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1347#-NkH5K7r6vRoXXo4bj7l:-NkH5K7r6vRoXXo4bj7m:b-c9oauz) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/491a555f7b45eab43ef4347bc59794a2e2b4cd8c/src/main/kotlin/org/wfanet/measurement/reporting/service/internal/testing/v2/ReportsServiceTest.kt#L505)):*
> ```Kotlin
> 
>   @Test
>   fun `createReport returns report after report metrics have been created`() = runBlocking {
> ```

This doesn't really tell me what the method is testing. I assume it's verifying that the existing metrics get reused?



_Suggestion:_
```Kotlin
reuses existing metrics
```

riemanli (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1347)* status: 0 of 3 files reviewed, 2 unresolved discussions (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/postgres/readers/MetricReader.kt` line 230 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1347#-NkH2Au-7KFKEU3hH2y6:-NkHAElDC4Pp8AqlAldt:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/491a555f7b45eab43ef4347bc59794a2e2b4cd8c/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/postgres/readers/MetricReader.kt#L230)):*

nit: keep consistent indentation for FROM clause
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/postgres/readers/MetricReader.kt` line 234 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1347#-NkH2lZ9EN8cyg_9RWBs:-NkHA_Fo0CYaZasLidrr:b3d58kq) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/491a555f7b45eab43ef4347bc59794a2e2b4cd8c/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/postgres/readers/MetricReader.kt#L234)):*

nit: append the `AND` when you append the additional conjunct so this reads clearer. The hanging `AND` here looks like a bug at first glance without careful reading.
</blockquote></details>

Got it. Done. 
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/internal/testing/v2/ReportsServiceTest.kt` line 505 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1347#-NkH5K7r6vRoXXo4bj7l:-NkHBpp-80dv-N7Ul-wK:bpu5ien) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/491a555f7b45eab43ef4347bc59794a2e2b4cd8c/src/main/kotlin/org/wfanet/measurement/reporting/service/internal/testing/v2/ReportsServiceTest.kt#L505)):*

This doesn't really tell me what the method is testing. I assume it's verifying that the existing metrics get reused?
</blockquote></details>

Thanks for pointing it out. Done!

SanjayVas (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1347)* status: all files reviewed, 1 unresolved discussion (waiting on @riemanli)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/postgres/writers/CreateReport.kt` line 118 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1347#-NkH439hDSm0DCy5DssT:-NkM6Q179MEJHqw3Myn9:b-1eipqe) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/491a555f7b45eab43ef4347bc59794a2e2b4cd8c/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/postgres/writers/CreateReport.kt#L118)):*

Thanks for the advice! Should the value name be singular or plural, i.e. metricCalculationSpec**s**ByExternalId vs metricCalculationSpecByExternalId?
</blockquote></details>

I've tended to go with plural as that indicates it's a collection.

kungfucraig (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1347)* status: all files reviewed, 2 unresolved discussions (waiting on @riemanli)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/internal/testing/v2/ReportsServiceTest.kt` line 616 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1347#-NkMEId_BF5ovGTfnoA7:-NkMEId_BF5ovGTfnoA8:bz1vryr) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7d4e4ec88c9fb291910304846a49f077dceba68b/src/main/kotlin/org/wfanet/measurement/reporting/service/internal/testing/v2/ReportsServiceTest.kt#L616)):*
> ```Kotlin
>       )
> 
>     for (entry in createdReportReusingMetrics.reportingMetricEntriesMap.entries) {
> ```

I'm trying to understand if this loop tests that the metrics were actually reused. If it does I'm not sure how it's achieving that. Could you add some comments to explain?

riemanli (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1347)* status: all files reviewed, 2 unresolved discussions (waiting on @kungfucraig)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/internal/testing/v2/ReportsServiceTest.kt` line 616 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1347#-NkMEId_BF5ovGTfnoA7:-NkMEkF01b6qB3LWgIGa:bch0o66) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7d4e4ec88c9fb291910304846a49f077dceba68b/src/main/kotlin/org/wfanet/measurement/reporting/service/internal/testing/v2/ReportsServiceTest.kt#L616)):*

I'm trying to understand if this loop tests that the metrics were actually reused. If it does I'm not sure how it's achieving that. Could you add some comments to explain?
</blockquote></details>

This loop doesn't check metric reuse but creation of a report instead. The check for metric reuse is the `assertThat` at the end.

kungfucraig (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1347)* status: all files reviewed, 2 unresolved discussions (waiting on @riemanli and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/internal/testing/v2/ReportsServiceTest.kt` line 616 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1347#-NkMEId_BF5ovGTfnoA7:-NkMFUCs6n_kkF3bUhif:b-ao7a5m) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7d4e4ec88c9fb291910304846a49f077dceba68b/src/main/kotlin/org/wfanet/measurement/reporting/service/internal/testing/v2/ReportsServiceTest.kt#L616)):*

This loop doesn't check metric reuse but creation of a report instead. The check for metric reuse is the `assertThat` at the end.
</blockquote></details>

Okay. I see, and the last check does it on the basis of this field (https://sourcegraph.com/github.com/world-federation-of-advertisers/cross-media-measurement@5327f5695b70fc2b62b8e1aa1b21544c38538749/-/blob/src/main/proto/wfa/measurement/reporting/v1alpha/report.proto?L87) being the same?

riemanli:
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/internal/testing/v2/ReportsServiceTest.kt` line 616 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1347#-NkMEId_BF5ovGTfnoA7:-NkMGMf15x_xxeh8GdLK:bcx28m5) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7d4e4ec88c9fb291910304846a49f077dceba68b/src/main/kotlin/org/wfanet/measurement/reporting/service/internal/testing/v2/ReportsServiceTest.kt#L616)):*

Okay. I see, and the last check does it on the basis of this field (https://sourcegraph.com/github.com/world-federation-of-advertisers/cross-media-measurement@5327f5695b70fc2b62b8e1aa1b21544c38538749/-/blob/src/main/proto/wfa/measurement/reporting/v1alpha/report.proto?L87) being the same?
</blockquote></details>

Something like that, but it's checking this [field](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/main/src/main/proto/wfa/measurement/internal/reporting/v2/report.proto#L74). Note that we don't add metric reuse for V1.

kungfucraig (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1347)* status: 1 of 3 files reviewed, 2 unresolved discussions (waiting on @riemanli and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/internal/testing/v2/ReportsServiceTest.kt` line 616 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1347#-NkMEId_BF5ovGTfnoA7:-NkNEvNp6LrGaOLNPic3:byprmlr) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7d4e4ec88c9fb291910304846a49f077dceba68b/src/main/kotlin/org/wfanet/measurement/reporting/service/internal/testing/v2/ReportsServiceTest.kt#L616)):*

Something like that, but it's checking this [field](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/main/src/main/proto/wfa/measurement/internal/reporting/v2/report.proto#L74). Note that we don't add metric reuse for V1.
</blockquote></details>

Checking reporting metric entires, doesn't help does it? That's telling you if reuse should happen, but it doesn't prove that reuse did happen. For that you'd need to see that the same set of Metrics was used in the results.

Tristan has a task to add Metric references to the Result object. Until that happens I don't see we have enough information in Report to prove Metrics were reused.

Maybe we can inspect the Metric API, or maybe we have to wait for the feature Tristan is adding to check this?

riemanli:
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/internal/testing/v2/ReportsServiceTest.kt` line 616 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1347#-NkMEId_BF5ovGTfnoA7:-NkNG6b_79_4KwYQbehL:bdy3spy) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7d4e4ec88c9fb291910304846a49f077dceba68b/src/main/kotlin/org/wfanet/measurement/reporting/service/internal/testing/v2/ReportsServiceTest.kt#L616)):*

Checking reporting metric entires, doesn't help does it? That's telling you if reuse should happen, but it doesn't prove that reuse did happen. For that you'd need to see that the same set of Metrics was used in the results.

Tristan has a task to add Metric references to the Result object. Until that happens I don't see we have enough information in Report to prove Metrics were reused.

Maybe we can inspect the Metric API, or maybe we have to wait for the feature Tristan is adding to check this?
</blockquote></details>

Great question. When create a report, the external metric IDs are not set until the metrics are created. Based on this, the testing flow is as follows:
1. Create `Report1`.
2. Submit create metric requests for all the metrics used in `Report1`.
3. Retrieve `Report1`. At this point, `Report1` should have all external metric IDs in `ReportingMetricEntries`.
4. Create `Report2` using the same report resource. At this point, if there is no metric reuse, `Report2` shouldn't have any external metric IDs and thus has different `ReportingMetricEntries` from what `Report1` has. If `Report1.ReportingMetricEntries` is equal to `Report2.ReportingMetricEntries`, it proves metric reuse happened during report creation.

riemanli:
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/postgres/writers/CreateReport.kt` line 146 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1347#-Nk_Yr5xDS7SBjsTHkZQ:-Nk_Yr5xDS7SBjsTHkZR:btzkojw) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/935df4e9b086465e8d386daceb4323a0c0e228d0/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/postgres/writers/CreateReport.kt#L146)):*
> ```Kotlin
>     // Find existing sets of metrics for each ReportingMetricEntry from
>     // report.reportingMetricEntriesMap.
>     val deferred =
> ```

@SanjayVas I tried to optimize the iterated queries by using async but the performance is not improved. Am I doing anything wrong with the async code below?

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1347)* status: 0 of 3 files reviewed, 3 unresolved discussions (waiting on @riemanli)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/postgres/writers/CreateReport.kt` line 146 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1347#-Nk_Yr5xDS7SBjsTHkZQ:-Nkam14JAr1R-5mC-aor:b-kuxxby) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/935df4e9b086465e8d386daceb4323a0c0e228d0/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/postgres/writers/CreateReport.kt#L146)):*
The async code looks like it should indeed make the calls to `readReportingMetricsByTimedReportingMetricEntry` in parallel, meaning the queries themselves should be done in parallel (up to the parallelism limit of the DB connections).

Note that this will result in `n` simultaneous connections, as the connection needs to stay open until the Flow is consumed. What you probably want is to have a single query that returns all of the results rather than making `n` separate queries.

> the performance is not improved

Are you sure this (executing the queries) is the bottleneck?

riemanli (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1347)* status: 0 of 4 files reviewed, 3 unresolved discussions (waiting on @kungfucraig and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/postgres/writers/CreateReport.kt` line 146 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1347#-Nk_Yr5xDS7SBjsTHkZQ:-Nkc-uPA3k68WyBSY0As:btvhaq7) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/935df4e9b086465e8d386daceb4323a0c0e228d0/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/postgres/writers/CreateReport.kt#L146)):*

The async code looks like it should indeed make the calls to `readReportingMetricsByTimedReportingMetricEntry` in parallel, meaning the queries themselves should be done in parallel (up to the parallelism limit of the DB connections).

Note that this will result in `n` simultaneous connections, as the connection needs to stay open until the Flow is consumed. What you probably want is to have a single query that returns all of the results rather than making `n` separate queries.

> the performance is not improved

Are you sure this (executing the queries) is the bottleneck?
</blockquote></details>

After a deeper investigation, I found two things
1. The iterated queries of metric reuse are no longer the bottleneck after I pruned the join statement in the query. It's the existing report creation taking a lot of time. A data point from my experiment: 
* Number of metrics in a report: 4320
* Number of metrics reused: 4320
* Execution time of metric reuse: 230 ms
* Execution time of report creation: 2 **secs**

2. In addition, I followed what you suggested to use a single query that returns all of the results. This further improves the performance a bit. 230ms => 200ms.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/postgres/writers/CreateReport.kt` line 382 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1347#-NkdD3fSA37BRbYGKKS1:-NkdD3fSA37BRbYGKKS2:b-3aagni) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/af434a995c812c75d19f953bc742e8c20f4a821b/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/postgres/writers/CreateReport.kt#L382)):*
> ```Kotlin
>           // Note that metric reuse mechanism won't produce the same order of `ReportingMetric`
>           // internally, but this won't impact any immutable fields in public resources.
>           reportingMetricMap.getValue(metricCalculationSpecReportingMetricKey).forEach {
> ```

Should've pointed this out earlier, but I hope this is ok: Due to the fact that we are reusing a set of metrics based on `MetricCalculationSpec`, we are not able to know the order of the internal `ReportingMetric`s. Note that `ReportingMetric` is only in internal resource and its order won't impact any immutable fields of public resources. 

@SanjayVas @kungfucraig Please let me know what you think.



_Code quote:_
```Kotlin
          // Note that metric reuse mechanism won't produce the same order of `ReportingMetric`
          // internally, but this won't impact any immutable fields in public resources.
          reportingMetricMap.getValue(metricCalculationSpecReportingMetricKey).forEach {
```
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/internal/testing/v2/ReportsServiceTest.kt` line 616 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1347#-NkMEId_BF5ovGTfnoA7:-NkdAs8oEKA9yGOEp7wa:b784mal) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7d4e4ec88c9fb291910304846a49f077dceba68b/src/main/kotlin/org/wfanet/measurement/reporting/service/internal/testing/v2/ReportsServiceTest.kt#L616)):*
> Can you have your test check for R1.RME = R2.RME?

I can do something slightly different because the order of `ReportingMetrics` is not guaranteed. PTAL at the new version.

SanjayVas (approved):
> Reviewed 4 of 4 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1347)* status: all files reviewed, 2 unresolved discussions (waiting on @kungfucraig)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/postgres/writers/CreateReport.kt` line 146 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1347#-Nk_Yr5xDS7SBjsTHkZQ:-NkqKSY1C1Kjp3tZmooj:bjqrjc) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/935df4e9b086465e8d386daceb4323a0c0e228d0/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/postgres/writers/CreateReport.kt#L146)):*

After a deeper investigation, I found two things
1. The iterated queries of metric reuse are no longer the bottleneck after I pruned the join statement in the query. It's the existing report creation taking a lot of time. A data point from my experiment: 
* Number of metrics in a report: 4320
* Number of metrics reused: 4320
* Execution time of metric reuse: 230 ms
* Execution time of report creation: 2 **secs**

2. In addition, I followed what you suggested to use a single query that returns all of the results. This further improves the performance a bit. 230ms => 200ms.
</blockquote></details>

This might be okay then. Tristan is working on improving Report creation, as he found that the remaining bottleneck is having to make multiple calls to the CMMS API. In theory switching to batch methods on the CMMS API should bring this down.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/postgres/writers/CreateReport.kt` line 382 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1347#-NkdD3fSA37BRbYGKKS1:-NkqKs383GkXBGs5-LGq:b-wiyppn) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/af434a995c812c75d19f953bc742e8c20f4a821b/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/postgres/writers/CreateReport.kt#L382)):*

Should've pointed this out earlier, but I hope this is ok: Due to the fact that we are reusing a set of metrics based on `MetricCalculationSpec`, we are not able to know the order of the internal `ReportingMetric`s. Note that `ReportingMetric` is only in internal resource and its order won't impact any immutable fields of public resources. 

@SanjayVas @kungfucraig Please let me know what you think.
</blockquote></details>

Totally fine as long as order isn't part of the API. I'd personally recommend documenting when this is or is not the case (e.g. the public APIs have an UNORDERED_LIST field behavior).

kungfucraig (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1347)* status: all files reviewed, 3 unresolved discussions (waiting on @riemanli)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/internal/testing/v2/ReportsServiceTest.kt` line 616 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1347#-NkMEId_BF5ovGTfnoA7:-NkpXtnp5IHhIRstD59C:b-kcm8m0) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7d4e4ec88c9fb291910304846a49f077dceba68b/src/main/kotlin/org/wfanet/measurement/reporting/service/internal/testing/v2/ReportsServiceTest.kt#L616)):*

> Can you have your test check for R1.RME = R2.RME?

I can do something slightly different because the order of `ReportingMetrics` is not guaranteed. PTAL at the new version.
</blockquote></details>

I think I'm still confused. Shouldn't we be checking that the Metrics 
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/internal/testing/v2/ReportsServiceTest.kt` line 679 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1347#-NkqYjTN3_1l_Xk1ODic:-NkqYjTN3_1l_Xk1ODid:bkgxv01) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/af434a995c812c75d19f953bc742e8c20f4a821b/src/main/kotlin/org/wfanet/measurement/reporting/service/internal/testing/v2/ReportsServiceTest.kt#L679)):*
> ```Kotlin
>     // internally. This won't impact any immutable fields in public resources.
>     for (entry in createdReportReusingMetrics.reportingMetricEntriesMap.entries) {
>       val expectedReportingMetricCalculationSpec =
> ```

expectedReportingMetrics ?

riemanli (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1347)* status: 3 of 4 files reviewed, 2 unresolved discussions (waiting on @kungfucraig and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/postgres/writers/CreateReport.kt` line 146 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1347#-Nk_Yr5xDS7SBjsTHkZQ:-NlNNG33BjBGCVb8qFit:b-qx665v) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/935df4e9b086465e8d386daceb4323a0c0e228d0/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/postgres/writers/CreateReport.kt#L146)):*

This might be okay then. Tristan is working on improving Report creation, as he found that the remaining bottleneck is having to make multiple calls to the CMMS API. In theory switching to batch methods on the CMMS API should bring this down.
</blockquote></details>

The analysis I did above is just for internal report creation. There is no call to CMM.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/internal/testing/v2/ReportsServiceTest.kt` line 616 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1347#-NkMEId_BF5ovGTfnoA7:-NlNPe_CA-v_BrPKVV4M:b-fy9efk) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7d4e4ec88c9fb291910304846a49f077dceba68b/src/main/kotlin/org/wfanet/measurement/reporting/service/internal/testing/v2/ReportsServiceTest.kt#L616)):*

I think I'm still confused. Shouldn't we be checking that the Metrics
</blockquote></details>

Resolved offline. 

Here is the explanation: When a report creation request is submitted through the public reports service by a MC, what the public RPC function does first is call the internal reports service to insert an initial internal report to the DB. Without any metric reuse, the internal reports service populates `createMetricRequestId` into this initial internal report, and the public reports service can then use the IDs to submit metric creation requests. If metric reuse happens during the insertion of the initial internal report, both `createMetricRequestId` and `externalMetricId` of the reused metric will be populated into the internal report. 

As a result, in the unit test I check the occurrence of metric reuse by checking whether `reportingMetricEntriesMap` of the second report is the same as the one in the first report.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/internal/testing/v2/ReportsServiceTest.kt` line 679 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1347#-NkqYjTN3_1l_Xk1ODic:-NlNPLx7DM8LPr7QiqES:b-zyokx) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/af434a995c812c75d19f953bc742e8c20f4a821b/src/main/kotlin/org/wfanet/measurement/reporting/service/internal/testing/v2/ReportsServiceTest.kt#L679)):*

expectedReportingMetrics ?
</blockquote></details>

`expectedReportingMetricCalculationSpec` is correct because the type of the value in the map is `ReportingMetricCalculationSpec`. See https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/af434a995c812c75d19f953bc742e8c20f4a821b/src/main/proto/wfa/measurement/internal/reporting/v2/report.proto.

I changed it to `expected` to be more concise and less confusing anyway.

stevenwarejones (approved):
> Reviewed 3 of 4 files at r6, 1 of 1 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1347)* status: all files reviewed, 2 unresolved discussions (waiting on @kungfucraig and @SanjayVas)

kungfucraig (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1347)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @SanjayVas)

PR #1353, opened 2023-11-30
Update proto and db schema to fulfill requisition with random seed.

renjiezh:


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1353)

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1353)* status: 0 of 9 files reviewed, 1 unresolved discussion (waiting on @kungfucraig, @ple13, and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/gcloud/spanner/computation/ComputationTokenProtoQuery.kt` line 160 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1353#-NkWlqtK9Y3IMRfbqEyH:-NkWlqtK9Y3IMRfbqEyI:b-j6zuca) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1953cbffefeed0e6793b38c11d878e745ac2999b/src/main/kotlin/org/wfanet/measurement/duchy/deploy/gcloud/spanner/computation/ComputationTokenProtoQuery.kt#L160)):*
> ```Kotlin
>             }
>             val dataPath = it.getString("PathToBlob")
>             if (!dataPath.isNullOrBlank()) {
> ```

I tried to use `isNull` here but `getString()` cannot be used for a null value from a sql query result. In the sql query above, it has IFNULL() providing blank string as default. 
And I checked the insertRequisition & updateRequisition, they guaranteed the empty string cannot be written. So I think it is safe that blank strings won't appear in column `PathToBlob`.

SanjayVas (approved):
> Reviewed 9 of 9 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1353)* status: all files reviewed, 1 unresolved discussion (waiting on @kungfucraig, @ple13, and @renjiezh)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/gcloud/spanner/computation/ComputationTokenProtoQuery.kt` line 160 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1353#-NkWlqtK9Y3IMRfbqEyH:-NkakdNzCqc2wLkL-8Zk:b-kcm8w6) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1953cbffefeed0e6793b38c11d878e745ac2999b/src/main/kotlin/org/wfanet/measurement/duchy/deploy/gcloud/spanner/computation/ComputationTokenProtoQuery.kt#L160)):*

I tried to use `isNull` here but `getString()` cannot be used for a null value from a sql query result. In the sql query above, it has IFNULL() providing blank string as default. 
And I checked the insertRequisition & updateRequisition, they guaranteed the empty string cannot be written. So I think it is safe that blank strings won't appear in column `PathToBlob`.
</blockquote></details>

With the `IFNULL` in the query this check should just be `isEmpty()`. Ideally we wouldn't use IFNULL in the query and instead separately call `it.isNull("PathToBlob")` to check whether it's NULL before calling `getString`.

ple13 (approved):
> Reviewed 9 of 9 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1353)* status: all files reviewed, 1 unresolved discussion (waiting on @kungfucraig and @renjiezh)

kungfucraig (approved):
> Reviewed 8 of 9 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1353)* status: all files reviewed, 2 unresolved discussions (waiting on @renjiezh)

___
*[`src/main/resources/duchy/postgres/add-requisition-random-seed.sql` line 21 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1353#-NkqXz6SAFW9m9rUxZS1:-NkqXz6SAFW9m9rUxZS2:bwb02j0) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7526cb0cf5ff473b5e433b7ac4ca2377ce2fefa3/src/main/resources/duchy/postgres/add-requisition-random-seed.sql#L21)):*
> ```TSQL
> ALTER TABLE Requisitions
>     -- A random seed to expand into the deterministic requisition blob.
>     ADD COLUMN RandomSeed bytea;
> ```

nit: add a new line

PR #1369, opened 2023-12-15
update list view style and fix some UI problems

bdomen-ggl:
Add a header to the list reports view. Not using the same one from get reports since most of the icons don't make sense. Also making it easier to modify when we get some style updates.

Fixing the react-router browser router. It wasn't working when you refresh the page (other than the root) and would fail if you directly navigate to a sub-link.


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1369)

SanjayVas (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1369)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @iverson52000)

PR #1386, opened 2024-01-03
Exchangestep Claiming limiting and exportTask fix

stevenwarejones:


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1386)

src/main/kotlin/org/wfanet/panelmatch/client/launcher/ExchangeStepLauncher.kt
```diff
@@ -24,22 +25,33 @@ import org.wfanet.panelmatch.client.launcher.InvalidExchangeStepException.Failur
...
 
+  private val semaphore = if (maxParallelExchangeSteps !== null) Semaphore(maxParallelExchangeSteps) else null
+
   /**
    * Finds a single ready Exchange Step and starts executing. If an Exchange Step is found,
    * validates it, and starts executing. If not found simply returns.
    */
   suspend fun findAndRunExchangeStep() {
-    val (exchangeStep, attemptKey) = apiClient.claimExchangeStep() ?: return
+    if (semaphore !== null) semaphore.acquire()
```
jonmolle, line 40: My one question here after reading more in kotlin docs is: should this be tryAcquire? Otherwise we'll queue up suspended requests much faster than we can clear them. The throttler should control the loop speed so we don't hammer this like a spinlock.

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1386)* status: 0 of 6 files reviewed, 1 unresolved discussion (waiting on @jonmolle and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/panelmatch/client/launcher/ExchangeStepLauncher.kt` line 39 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1386#gh-1442047417:-NnKUU3g0I3vU87C1SQt:bw5ugbx) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3dbf6d465220673987304e45c5e4f5983189a35b/src/main/kotlin/org/wfanet/panelmatch/client/launcher/ExchangeStepLauncher.kt#L39)):*
<details><summary><i>Previously, jonmolle wrote…</i></summary><blockquote>

My one question here after reading more in kotlin docs is: should this be tryAcquire? Otherwise we'll queue up suspended requests much faster than we can clear them. The throttler should control the loop speed so we don't hammer this like a spinlock.
</blockquote></details>

good call. changed this to tryAcquire. wdyt?

jonmolle (approved):
lgtm

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1386#-:-NnKVaUE5QAxvXeBHLJ_:bb7k0qh)* status: 0 of 6 files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

jonmolle (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1386)* status: 0 of 12 files reviewed, 3 unresolved discussions (waiting on @SanjayVas and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/panelmatch/client/launcher/GrpcApiClient.kt` line 43 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1386#-NnLH-v83H7wPLVGA0B5:-NnLH-v83H7wPLVGA0B6:bwq4h9i) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/222ecc9c75e24884d8ecc77a0f17de2c81b3be87/src/main/kotlin/org/wfanet/panelmatch/client/launcher/GrpcApiClient.kt#L43)):*
> ```Kotlin
> ) : ApiClient {
>   private val semaphore =
>     if (maxClaimedExchangeSteps != -1) Semaphore(maxClaimedExchangeSteps) else null
> ```

nitpick: perhaps > 0 ? Or do you want an error here?
___
*[`src/main/kotlin/org/wfanet/panelmatch/client/launcher/GrpcApiClient.kt` line 90 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1386#-NnLIS_a84GxHrOh7YZ7:-NnLIS_a84GxHrOh7YZ8:bb34qdx) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/222ecc9c75e24884d8ecc77a0f17de2c81b3be87/src/main/kotlin/org/wfanet/panelmatch/client/launcher/GrpcApiClient.kt#L90)):*
> ```Kotlin
>     }
>     exchangeStepAttemptsClient.finishExchangeStepAttempt(request)
>     if (semaphore !== null) semaphore.release()
> ```

I'm a little concerned how this will behave in the case of an API error. Do you think it is too much to put the rest of the function in a try block and this in a finally?

jonmolle (approved):
LGTM the update. I only have nits, so mostly whatever you choose sounds fine to me.

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1386#-:-NnLKrL85ZTLIvQpzShZ:b-unma39)* status: 0 of 12 files reviewed, 3 unresolved discussions (waiting on @SanjayVas and @stevenwarejones)

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1386)* status: 0 of 12 files reviewed, 3 unresolved discussions (waiting on @jonmolle and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/panelmatch/client/launcher/GrpcApiClient.kt` line 43 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1386#-NnLH-v83H7wPLVGA0B5:-NnLMENzBBqneOxz5HGw:b-9t30q8) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/222ecc9c75e24884d8ecc77a0f17de2c81b3be87/src/main/kotlin/org/wfanet/panelmatch/client/launcher/GrpcApiClient.kt#L43)):*
<details><summary><i>Previously, jonmolle wrote…</i></summary><blockquote>

nitpick: perhaps > 0 ? Or do you want an error here?
</blockquote></details>

I would throw an error either way. Semaphore will throw an error if its passed a value less than 1. I think this is acceptable behavior.
___
*[`src/main/kotlin/org/wfanet/panelmatch/client/launcher/GrpcApiClient.kt` line 90 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1386#-NnLIS_a84GxHrOh7YZ7:-NnLLsd69eB7TvF8A6Fk:b-y9ga70) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/222ecc9c75e24884d8ecc77a0f17de2c81b3be87/src/main/kotlin/org/wfanet/panelmatch/client/launcher/GrpcApiClient.kt#L90)):*
<details><summary><i>Previously, jonmolle wrote…</i></summary><blockquote>

I'm a little concerned how this will behave in the case of an API error. Do you think it is too much to put the rest of the function in a try block and this in a finally?
</blockquote></details>

if there are API issues, we shouldn't be looping either. I don't think the current client is very robust agains client errors.

jonmolle (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1386)* status: 0 of 12 files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/panelmatch/client/launcher/GrpcApiClient.kt` line 43 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1386#-NnLH-v83H7wPLVGA0B5:-NnLO3iX8BdPftGtz_3Q:b-sf8yw5) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/222ecc9c75e24884d8ecc77a0f17de2c81b3be87/src/main/kotlin/org/wfanet/panelmatch/client/launcher/GrpcApiClient.kt#L43)):*

I would throw an error either way. Semaphore will throw an error if its passed a value less than 1. I think this is acceptable behavior.
</blockquote></details>

sure, sgtm
___
*[`src/main/kotlin/org/wfanet/panelmatch/client/launcher/GrpcApiClient.kt` line 90 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1386#-NnLIS_a84GxHrOh7YZ7:-NnLNXU12I1k-G3v9FJG:b-m3mmuc) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/222ecc9c75e24884d8ecc77a0f17de2c81b3be87/src/main/kotlin/org/wfanet/panelmatch/client/launcher/GrpcApiClient.kt#L90)):*

if there are API issues, we shouldn't be looping either. I don't think the current client is very robust agains client errors.
</blockquote></details>

That's fair. I was thinking about it in a way that an error in one recurring exchange wouldn't block the others, but I believe if something was broken on the kingdom side of this it would be broken for the entire party.

stevenwarejones (commented):
> Reviewed 1 of 5 files at r1, 2 of 5 files at r2, 9 of 9 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1386)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @SanjayVas)

SanjayVas (commented):
> Reviewed 1 of 9 files at r5.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1386)* status: all files reviewed, 1 unresolved discussion

___
*[`src/main/kotlin/org/wfanet/panelmatch/client/deploy/ExchangeWorkflowDaemonFromFlags.kt` line 114 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1386#-NnfD1Q-3glsml7SdO1o:-NnfD1Q-3glsml7SdO1p:b-vvqfpy) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/222ecc9c75e24884d8ecc77a0f17de2c81b3be87/src/main/kotlin/org/wfanet/panelmatch/client/deploy/ExchangeWorkflowDaemonFromFlags.kt#L114)):*
> ```Kotlin
>   }
> 
>   private val maxParallelClaimedExchangeSteps: Int by lazy { flags.maxParallelClaimedExchangeSteps }
> ```

Why is this lazy? It can just delegate. Though I'm also not clear why this private property even needs to exist when the underlying one can be used directly.



_Suggestion:_
```Kotlin
private val maxParallelClaimedExchangeSteps: Int
  get() = flags.maxParallelClaimedExchangeSteps
```

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1386)* status: all files reviewed, 3 unresolved discussions

___
*[`src/main/kotlin/org/wfanet/panelmatch/client/deploy/ExchangeWorkflowFlags.kt` line 121 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1386#-NnfDbaQ4p-rgbSnrGhp:-NnfDbaRES2f5CIvgLDp:b87mlye) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/222ecc9c75e24884d8ecc77a0f17de2c81b3be87/src/main/kotlin/org/wfanet/panelmatch/client/deploy/ExchangeWorkflowFlags.kt#L121)):*
> ```Kotlin
>   @set:Option(
>     names = ["--max-parallel-claimed-exchange-steps"],
>     defaultValue = "-1",
> ```

Is `0` a valid value? If not, just make that the "unlimited" value. Alternatively, make this nullable so it's explicit.
___
*[`src/main/kotlin/org/wfanet/panelmatch/client/launcher/GrpcApiClient.kt` line 42 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1386#-NnfEHK7CEzy4Wf53MBJ:-NnfEHK7CEzy4Wf53MBK:bvaoh4m) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/222ecc9c75e24884d8ecc77a0f17de2c81b3be87/src/main/kotlin/org/wfanet/panelmatch/client/launcher/GrpcApiClient.kt#L42)):*
> ```Kotlin
>   private val maxClaimedExchangeSteps: Int,
> ) : ApiClient {
>   private val semaphore =
> ```

I wonder if a "no-op semaphore" impl might be better here than having this be conditionally instantiated.

SanjayVas:
Which commit(s) to the main branch were these cherry-picked from?

PR #1393, opened 2024-01-05
add test for complex expression without operative field

stevenwarejones:
For the Origin pilot, its my understanding their event template has no operative fields that line up with our PBM definition. I would expect the compileToNormalForm to return a True expression in this case.


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1393)

stevenwarejones:
closed in lieu of https://github.com/world-federation-of-advertisers/cross-media-measurement/pull/1396

PR #1399, opened 2024-01-08
debugging write to kms [don't merge]

stevenwarejones:
Uses a plaintext secret map instead of a kms to store private keys

Also uses this branch of common-jvm https://github.com/world-federation-of-advertisers/common-jvm/tree/stevenwarejones_kms_storage_client_logging

### DIFF
--- a/build/repositories.bzl
+++ b/build/repositories.bzl
@@ -25,10 +25,10 @@ def wfa_measurement_system_repositories():
     """Imports all direct dependencies for wfa_measurement_system."""
 
     wfa_repo_archive(
+        commit = "506734bfff557e54be5ff389214f96720389ed90",
         name = "wfa_common_jvm",
         repo = "common-jvm",
-        sha256 = "c0e39569064363ed61b9b780bf2aa37b99d0c81933a7a434b6b320c2e81063b4",
-        version = "0.65.0",
+        sha256 = "16381fdf06ee513661f3ec56769dfcd8ba462a2f930a7d86a418175e57a60622",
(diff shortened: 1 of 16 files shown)


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1399)

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1399)* status: 0 of 6 files reviewed, 1 unresolved discussion (waiting on @jonmolle)

___
*[`src/main/kotlin/org/wfanet/panelmatch/client/deploy/ExchangeWorkflowFlags.kt` line 131 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1399#-NoK3JbH5pyer_0NrCnR:-NoK3JbH5pyer_0NrCnS:b-nncl5a) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a34520b625006d93eaa323b98ed15ba3fd015827/src/main/kotlin/org/wfanet/panelmatch/client/deploy/ExchangeWorkflowFlags.kt#L131)):*
> ```Kotlin
>     names = ["--private-key-path"],
>     defaultValue = "",
>     description = ["Fixed path for a private key used in all exchanges"],
> ```

update this comment that this is a fixed path for the _encrypted_ private key (ie already encrypted previously by the kms)

PR #1405, opened 2024-01-09
cherry pick missing absl transitive dependency

stevenwarejones:


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1405)

SanjayVas:
Why do we need this in v0.4.x? Are we planning on cherry-picking another commit that upgrades absl?

stevenwarejones (commented):
Two reasons:
1. Best practice to include it directly. 
2. It would be convenient for us to have this so we don't have to patch the code.

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1405#-:-NnneEwJ6nj89oFsyXUo:b-gtkl5)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @SanjayVas)

SanjayVas (commented):
I'm aware of the best practice, but AFAICT this isn't fixing a bug that exists in 0.4.2. We'll also be releasing 0.5.0 soon.

> It would be convenient for us to have this so we don't have to patch the code.

I'm not sure what you mean here, unless there's some other pending cherry-pick to 0.4.x that depends on this.

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1405#-:-NnoJ65gCQQHSeU_ZINN:bm2vpxs)* status: 0 of 2 files reviewed, all discussions resolved

PR #1415, opened 2024-01-11
Task expiration spanner fix

jonmolle:
Specifically: if attempt has caused a temporary failure, we do not update the status of the task to READY_TO_RETRY if the attempt being marked as failed is not the latest attempt.

Justification: We have noticed that there are cases where multiple
IN_PROGRESS attempts exist for a single task. 
  - Whether those are active or not is unclear, but the result of this was if any fail or time out the task is set to READY_TO_RETRY while other attempts are currently IN_PROGRESS (which is not desirable). 
  - If there is somehow a newer attempt than the one being finished, we trust _that_ attempt will update the task state when it completes or times out and therefore the task is not, in fact, ready to retry.

### DIFF
--- a/src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/writers/CreateExchangesAndSteps.kt
+++ b/src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/writers/CreateExchangesAndSteps.kt
@@ -113,7 +113,7 @@ class CreateExchangesAndSteps(
         bind(Params.EXTERNAL_PARTY_ID to externalPartyId)
 
         appendClause("WHERE")
-        append(conjuncts.joinToString(" AND "))
+        appendClause(conjuncts.joinToString(" AND "))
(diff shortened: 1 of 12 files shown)


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1415)

stevenwarejones (commented):
do you want to cherry-pick this into 0.4 as well?

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1415#-:-NnuRR_95IR-pLnrKEf5:b-2myl6s)* status: 1 of 2 files reviewed, all discussions resolved

stevenwarejones (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1415#-:-Nnxr8oK1bVD1vkLP-fz:bnfp4nl)* status: 1 of 2 files reviewed, all discussions resolved

PR #1425, opened 2024-01-17
Support sampling for synthetic event group spec

uakyol:
**Consider the following problem** for generating synthetic data for 3 Edps that represents a realistic intersection
Using the current [VidRange field](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/119d42385f5f56da7d28693d0987a9aa0bdd53aa/src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata/testing/simulator_synthetic_data_spec.proto#L108)
Edp1 range -> (0,100)
Edp2 range -> (50,150) [At this point we have 50 intersection for Edp1 and 2]
Edp3 range -> ?? [How should we decide on this]

**Here are 2 possible solutions** :  
1. Sample Vids from those ranges given a sample size for each vidRange.
2. Another way is to sample n ranges from the larger range while creating the [SyntheticEventGroupSpec](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/119d42385f5f56da7d28693d0987a9aa0bdd53aa/src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata/testing/simulator_synthetic_data_spec.proto#L97) and approximate. 

This PR is the implementation of the first option


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1425)

SanjayVas (changes requested):
> Reviewed 2 of 3 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1425)* status: 2 of 3 files reviewed, 7 unresolved discussions (waiting on @uakyol)

*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1425#-NoNI4TgDrxHi9E00bjd:-NoNI4TgDrxHi9E00bje:b-eoap44) (no related file):*
Forgive my lack of understanding on how all the math works, but how does this interact with `VidSamplingInterval` from `MeasurementSpec`? Are they unrelated?

___
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SyntheticDataGeneration.kt` line 50 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1425#-NoNGi-f8mg7taz5UA2O:-NoNGi-gDxOu6Bl5xH0c:b-vn9z2w) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4bbe9913abe64b6ee149eb44633ba8ac307f795b/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SyntheticDataGeneration.kt#L50)):*
> ```Kotlin
>     populationSpec: SyntheticPopulationSpec,
>     syntheticEventGroupSpec: SyntheticEventGroupSpec,
>     randomSeed: Long = 0L,
> ```

Pass a Kotlin Random object defaulting to `Random` instead of passing a seed. This way the caller can pass a pre-seeded Random or some non-default Random impl.



_Code quote:_
```Kotlin
randomSeed: Long = 0L,
```
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SyntheticDataGeneration.kt` line 190 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1425#-NoNFZ0EDGsNDoj_-j3i:-NoNFZ0EDGsNDoj_-j3j:bhj90zv) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4bbe9913abe64b6ee149eb44633ba8ac307f795b/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SyntheticDataGeneration.kt#L190)):*
> ```Kotlin
> 
> private fun SyntheticEventGroupSpec.FrequencySpec.hasOverlaps(): Boolean {
>   return this.vidRangeSpecsList
> ```

nit: unnecessary qualification



_Code quote:_
```Kotlin
this.
```
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SyntheticDataGeneration.kt` line 191 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1425#-NoNFSCp2SvyFFxNUp2u:-NoNFSCq5Ixqrs-Ug_Ud:b8sijcc) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4bbe9913abe64b6ee149eb44633ba8ac307f795b/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SyntheticDataGeneration.kt#L191)):*
> ```Kotlin
> private fun SyntheticEventGroupSpec.FrequencySpec.hasOverlaps(): Boolean {
>   return this.vidRangeSpecsList
>     .toList()
> ```

Isn't this already a List?



_Code quote:_
```Kotlin
.toList()
```
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SyntheticDataGeneration.kt` line 194 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1425#-NoNH8Ot0fyIPBNA4oi0:-NoNH8Ot0fyIPBNA4oi1:bhe7ve5) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4bbe9913abe64b6ee149eb44633ba8ac307f795b/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SyntheticDataGeneration.kt#L194)):*
> ```Kotlin
>     .flatMap { vidRangeSpec: VidRangeSpec ->
>       listOf(
>         RangePoint(vidRangeSpec.vidRange.start, true),
> ```

I feel like there's a better way to implement this that can take advantage of Kotlin ranges. See https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.ranges/-open-end-range/ and how we [implemented our own `overlaps` method for a range of Instant](https://github.com/world-federation-of-advertisers/common-jvm/blob/ddedd5aa4362df505aec38f1156639dfb6dab1d4/src/main/kotlin/org/wfanet/measurement/common/OpenEndTimeRange.kt#L33).
___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata/testing/simulator_synthetic_data_spec.proto` line 116 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1425#-NoNG18F7BA5D7QZ31Rn:-NoNG18F7BA5D7QZ31Ro:blehh05) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4bbe9913abe64b6ee149eb44633ba8ac307f795b/src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata/testing/simulator_synthetic_data_spec.proto#L116)):*
> ```proto
>       // A map of `non_population_fields` from `SyntheticPopulationSpec` to
>       // their values.
>       map<string, FieldValue> non_population_field_values = 3;
> ```

Since this message is already in use the field should not be renumbered.



_Code quote:_
```proto
3
```
___
*[`src/test/kotlin/org/wfanet/measurement/loadtest/dataprovider/SyntheticDataGenerationTest.kt` line 386 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1425#-NoNHumY7Er6kaGBods0:-NoNHumY7Er6kaGBods1:b-lxipgq) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4bbe9913abe64b6ee149eb44633ba8ac307f795b/src/test/kotlin/org/wfanet/measurement/loadtest/dataprovider/SyntheticDataGenerationTest.kt#L386)):*
> ```Kotlin
>   }
> 
>   fun `generateEvents fails when overlapping vidRanges exist`() {
> ```

nit: be more specific than "fails"



_Suggestion:_
```Kotlin
generateEvents throws IllegalArgumentException when vid ranges overlap
```

uakyol:

They are completely unrelated. After these vids are generated here, they are [hashed](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4ee0287d534bbc9e5e75de969438d2928fbe680e/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt#L1178) to the [0,1] interval. 

`VidSamplingInterval` from `MeasurementSpec` is an interval in [0,1] it selects a bunch of them and based on the size of the interval, reach result is scaled to approximate what would have been if they weren't sampled.
</blockquote></details>

On second look, my sentence is hard to parse, let me rephrase : 

After these vids are generated here, they are [hashed](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4ee0287d534bbc9e5e75de969438d2928fbe680e/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt#L1178) to the \[0,1\] interval.

`VidSamplingInterval` from `MeasurementSpec` is an interval in \[0,1\] . The vids that are hashed, are filtered if they are in the `VidSamplingInterval`. This gives a provides a number of vids that can be interpreted as reach. But since they were uniformly hashed and only 1/n of them are selected where n=`VidSamplingInterval.width` we multiply the number of selected vids with n and approximate reach.

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1425)* status: all files reviewed, 4 unresolved discussions (waiting on @uakyol)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SyntheticDataGeneration.kt` line 194 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1425#-NoNH8Ot0fyIPBNA4oi0:-NoN_kkIBAIVR-IS20Oo:bcpcdjf) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4bbe9913abe64b6ee149eb44633ba8ac307f795b/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SyntheticDataGeneration.kt#L194)):*

ah, actually I didn't need what I did nor do I need Ranges. I simplified the implementation wdyt?
</blockquote></details>

Can still be a little clearer by having the overlapping predicate in `any` rather than a transformation in `zipWithNext`.

```kotlin
  vidRangeSpecsList
    .map { it.vidRange }
    .sortedBy { it.start }
    .zipWithNext()
    .any { (first, second) -> first.overlaps(second) }
```

Feel free to either add an extension for `overlaps` or inline the impl.
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SyntheticDataGeneration.kt` line 50 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1425#-NoNZPXS2fJeKkewbIkc:-NoNZPXS2fJeKkewbIkd:b-g9laph) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4a3968bcf674a49a969e95ce841d7d76786b1461/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SyntheticDataGeneration.kt#L50)):*
> ```Kotlin
>     populationSpec: SyntheticPopulationSpec,
>     syntheticEventGroupSpec: SyntheticEventGroupSpec,
>     random: Random = Random(0L),
> ```

Don't specify a default seed.



_Suggestion:_
```Kotlin
Random
```
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SyntheticDataGeneration.kt` line 50 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1425#-NoNZf3X54uH0ojCatXb:-NoNZf3X54uH0ojCatXc:bjkdcvz) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4a3968bcf674a49a969e95ce841d7d76786b1461/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SyntheticDataGeneration.kt#L50)):*
> ```Kotlin
>     populationSpec: SyntheticPopulationSpec,
>     syntheticEventGroupSpec: SyntheticEventGroupSpec,
>     random: Random = Random(0L),
> ```

Just a double-check on determinism here. These synthetic specs must always produce the same events regardless of even what language the generator is implemented in. I'm pretty sure that means that it cannot use randomness anywhere in its implementation as a result.



_Code quote:_
```Kotlin
 random: Random
```

uakyol (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1425)* status: 2 of 3 files reviewed, 4 unresolved discussions (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SyntheticDataGeneration.kt` line 194 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1425#-NoNH8Ot0fyIPBNA4oi0:-NoNeU5D0nV51D_C1MHR:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4bbe9913abe64b6ee149eb44633ba8ac307f795b/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SyntheticDataGeneration.kt#L194)):*

Can still be a little clearer by having the overlapping predicate in `any` rather than a transformation in `zipWithNext`.

```kotlin
  vidRangeSpecsList
    .map { it.vidRange }
    .sortedBy { it.start }
    .zipWithNext()
    .any { (first, second) -> first.overlaps(second) }
```

Feel free to either add an extension for `overlaps` or inline the impl.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SyntheticDataGeneration.kt` line 50 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1425#-NoNZf3X54uH0ojCatXb:-NoNd9yq1amgYhYA4bwC:b88zf03) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4a3968bcf674a49a969e95ce841d7d76786b1461/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SyntheticDataGeneration.kt#L50)):*

Just a double-check on determinism here. These synthetic specs must always produce the same events regardless of even what language the generator is implemented in. I'm pretty sure that means that it cannot use randomness anywhere in its implementation as a result.
</blockquote></details>

So, that basically is the whole purpose of this PR. This approach changes that assumption. The alternative is to generate a massive SyntheticEventGroupSpec that contains non overlapping vidranges that are sampled from a larger vid range. This new spec will be :

1. Much larger
2. Will only approximate uniform sampling because it samples ranges rather than individual vids
3. Will itself need to be generated randomly

We will need randomness to generate realistic intersections between EDPs, it is not feasible to hand craft realistic data and vid intersection between EDPs. If we don't do it here, than we will have to do it in another spec that will be parsed and randomly generates this. 

Fwiw, if you do not specify any sample_size, then there is no randomness. 
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SyntheticDataGeneration.kt` line 50 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1425#-NoNZPXS2fJeKkewbIkc:-NoNdUKv2UKDE7kZoGUz:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4a3968bcf674a49a969e95ce841d7d76786b1461/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SyntheticDataGeneration.kt#L50)):*

Don't specify a default seed.
</blockquote></details>

Done.

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1425)* status: 2 of 3 files reviewed, 4 unresolved discussions (waiting on @kungfucraig)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SyntheticDataGeneration.kt` line 50 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1425#-NoNZf3X54uH0ojCatXb:-NoNejPN3sDDBaGuLM4X:b-ujj6lf) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4a3968bcf674a49a969e95ce841d7d76786b1461/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SyntheticDataGeneration.kt#L50)):*
I personally think this is a no-go then. Synthetic generation was added as part of the larger effort to guarantee determinism for event queries, i.e. you can look at the spec and know exactly what events you will get.

> Will itself need to be generated randomly

I think this part is fine. As far as the other two points, this is why I was suggesting that we may need a new spec proto definition that e.g. does not rely on knowing the VID ranges from the population spec.

CC @kungfucraig to see if his opinion differs.

uakyol (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1425)* status: 2 of 3 files reviewed, 4 unresolved discussions (waiting on @kungfucraig and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SyntheticDataGeneration.kt` line 50 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1425#-NoNZf3X54uH0ojCatXb:-NoNgXWF1r91dIhPW3Sp:b-3y5dp7) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4a3968bcf674a49a969e95ce841d7d76786b1461/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SyntheticDataGeneration.kt#L50)):*

I personally think this is a no-go then. Synthetic generation was added as part of the larger effort to guarantee determinism for event queries, i.e. you can look at the spec and know exactly what events you will get.

> Will itself need to be generated randomly

I think this part is fine. As far as the other two points, this is why I was suggesting that we may need a new spec proto definition that e.g. does not rely on knowing the VID ranges from the population spec.

CC @kungfucraig to see if his opinion differs.
</blockquote></details>

To get on the same page : That new spec proto definition will have use randomness in its implementation.  My idea was to use this spec so that it will have randomness in its implementation if and only if the "sample_size" is set. Wdyt of documenting this more clearly in the proto and going with this? 

I am suggesting this because any SyntheticEventGroupSpec without sample_size will still satisfy your point of  `guarantee determinism for event queries, i.e. you can look at the spec and know exactly what events you will get.`

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1425)* status: 2 of 3 files reviewed, 4 unresolved discussions (waiting on @kungfucraig)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SyntheticDataGeneration.kt` line 50 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1425#-NoNZf3X54uH0ojCatXb:-NoNiolq1Aj4GJBnOWOj:b8kevsh) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4a3968bcf674a49a969e95ce841d7d76786b1461/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SyntheticDataGeneration.kt#L50)):*
> I am suggesting this because any SyntheticEventGroupSpec without sample_size will still satisfy your point

This isn't sufficient. Any synthetic event generation process **must** be deterministic always.

> That new spec proto definition will have use randomness in its implementation

To clarify what I mean by spec proto definition, I mean the definition of an alternative to the current SyntheticEventGroupSpec or alternative submessages (e.g. with oneofs). The implementation of such **must** be deterministic. Any randomness would have to be applied in generating the spec itself.

uakyol (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1425)* status: 2 of 3 files reviewed, 4 unresolved discussions (waiting on @kungfucraig and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SyntheticDataGeneration.kt` line 50 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1425#-NoNZf3X54uH0ojCatXb:-NoNk8NMCGBApY3_EkA5:b-2fbc90) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4a3968bcf674a49a969e95ce841d7d76786b1461/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SyntheticDataGeneration.kt#L50)):*


> This isn't sufficient. Any synthetic event generation process **must** be deterministic always.

Ah, gotcha on this



> The implementation of such **must** be deterministic. Any randomness would have to be applied in generating the spec itself.

I don't think that is possible. In order to do that, we would need to list the actual vids that the spec will generate in the spec itself. In other words, we need to generate the vids randomly and put them in a spec and then just read the spec and output those vids. Which makes it not a spec but the data itself.

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1425)* status: 2 of 3 files reviewed, 4 unresolved discussions (waiting on @kungfucraig)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SyntheticDataGeneration.kt` line 50 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1425#-NoNZf3X54uH0ojCatXb:-NoNlHyuDGT5qLByfjLW:b-w3d4b5) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4a3968bcf674a49a969e95ce841d7d76786b1461/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SyntheticDataGeneration.kt#L50)):*
>I don't think that is possible. In order to do that, we would need to list the actual vids that the spec will generate in the spec itself.

I'm not convinced yet that this is true. e.g. perhaps specifying multipliers or offsets. Regardless, the task is to have deterministic event generation from compressed specs. This means restricting what the synthetic generation is capable of.

kungfucraig (changes requested):
> Reviewed 1 of 3 files at r2.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1425)* status: 2 of 3 files reviewed, 5 unresolved discussions (waiting on @SanjayVas and @uakyol)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata/testing/simulator_synthetic_data_spec.proto` line 116 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1425#-NomBCJT3Qwqzd5tVgAq:-NomBCJT3Qwqzd5tVgAr:b-rj2v9h) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/cfdf5f51ee6c0c2e78edeeb3d9993d3d8d36726c/src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata/testing/simulator_synthetic_data_spec.proto#L116)):*
> ```proto
>       // Number of vids sampled uniformly without replacement from vid_range.
>       // If this is 0, no sampling is done and all the vids in range are taken.
>       int32 sample_size = 3;
> ```

Per offline conversation @uakyol will add a seed and rng type to the spec proto.

uakyol (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1425)* status: 0 of 3 files reviewed, 5 unresolved discussions (waiting on @kungfucraig and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SyntheticDataGeneration.kt` line 50 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1425#-NoNZf3X54uH0ojCatXb:-Np0XnZY1BmIdIZcawC1:b2qn496) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4a3968bcf674a49a969e95ce841d7d76786b1461/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SyntheticDataGeneration.kt#L50)):*

>I don't think that is possible. In order to do that, we would need to list the actual vids that the spec will generate in the spec itself.

I'm not convinced yet that this is true. e.g. perhaps specifying multipliers or offsets. Regardless, the task is to have deterministic event generation from compressed specs. This means restricting what the synthetic generation is capable of.
</blockquote></details>

updated based on offline conversation
___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata/testing/simulator_synthetic_data_spec.proto` line 116 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1425#-NomBCJT3Qwqzd5tVgAq:-Np0Xm7o-f1_gyuH_QkR:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/cfdf5f51ee6c0c2e78edeeb3d9993d3d8d36726c/src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata/testing/simulator_synthetic_data_spec.proto#L116)):*

Per offline conversation @uakyol will add a seed and rng type to the spec proto.
</blockquote></details>

Done.

SanjayVas (changes requested):
> Reviewed 1 of 3 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1425)* status: 1 of 3 files reviewed, 7 unresolved discussions (waiting on @kungfucraig and @uakyol)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SyntheticDataGeneration.kt` line 47 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1425#-Np0yyL91AFBy4u7ccQO:-Np0yyL91AFBy4u7ccQP:b-3lma6u) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0e63840ecde499978c7dc8c0d89fe3e8cb165316/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SyntheticDataGeneration.kt#L47)):*
> ```Kotlin
>    * @param populationSpec specification of the synthetic population
>    * @param syntheticEventGroupSpec specification of the synthetic event group
>    * @param random object to be used in sampling vids when sampleSize is provided in specification
> ```

This param no longer exists.
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SyntheticDataGeneration.kt` line 58 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1425#-Np0z7sx6WIEg6KA4pm4:-Np0z7sx6WIEg6KA4pm5:b48745u) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0e63840ecde499978c7dc8c0d89fe3e8cb165316/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SyntheticDataGeneration.kt#L58)):*
> ```Kotlin
>       "Expecting KOTLIN_RANDOM rng type, got ${syntheticEventGroupSpec.rngType}"
>     }
>     val random = Random(syntheticEventGroupSpec.randomSeed)
> ```

Don't we need to reinitialize this for each VidRangeSpec? The spec doesn't have any guarantee on the order that the ranges are sampled, and changing that order would affect the random output.

Regardless, whatever we go with needs to be documented in the spec definition so there's no ambiguity.
___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata/testing/simulator_synthetic_data_spec.proto` line 138 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1425#-Np0yKKMB-6sXr1Xr4uz:-Np0yKKMB-6sXr1Xr4v-:bmtupse) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0e63840ecde499978c7dc8c0d89fe3e8cb165316/src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata/testing/simulator_synthetic_data_spec.proto#L138)):*
> ```proto
>   repeated DateSpec date_specs = 2;
> 
>   // Random seed to be fed into the random number generator to sample vids.
> ```

Document that this is required if any VidRangeSpec specifies a sample_size.
___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata/testing/simulator_synthetic_data_spec.proto` line 149 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1425#-Np0yF0iD34t_ZYjLol5:-Np0yF0iD34t_ZYjLol6:br6p7vn) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0e63840ecde499978c7dc8c0d89fe3e8cb165316/src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata/testing/simulator_synthetic_data_spec.proto#L149)):*
> ```proto
>   }
> 
>   // Activation state of this `Account`. Output-only.
> ```

Wrong comment copy-pasted from somewhere. Also document that this is required if `random_seed` is specified.



_Code quote:_
```proto
  // Activation state of this `Account`. Output-only.
```

uakyol (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1425)* status: 0 of 3 files reviewed, 7 unresolved discussions (waiting on @kungfucraig and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SyntheticDataGeneration.kt` line 47 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1425#-Np0yyL91AFBy4u7ccQO:-Np16FHBEdaBcPFSdCIu:b-k73z76) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0e63840ecde499978c7dc8c0d89fe3e8cb165316/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SyntheticDataGeneration.kt#L47)):*

This param no longer exists.
</blockquote></details>

thanks
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SyntheticDataGeneration.kt` line 58 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1425#-Np0z7sx6WIEg6KA4pm4:-Np16drk9ajI4IU0FRo7:b-f1hdc7) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0e63840ecde499978c7dc8c0d89fe3e8cb165316/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SyntheticDataGeneration.kt#L58)):*

Don't we need to reinitialize this for each VidRangeSpec? The spec doesn't have any guarantee on the order that the ranges are sampled, and changing that order would affect the random output.

Regardless, whatever we go with needs to be documented in the spec definition so there's no ambiguity.
</blockquote></details>

ah, I would like to go with initing it here so we get variety even if VidRangeSpec s are the same but in different SyntheticDataSpecs that has other VidRangeSpec. 

Do you think it is enough to document this in the spec or we do you think we should add a field in the VidRangeSpec such as "id" or "orderNumber"?
___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata/testing/simulator_synthetic_data_spec.proto` line 138 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1425#-Np0yKKMB-6sXr1Xr4uz:-Np16UMrAkI5SUt0LH58:bsf6lwp) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0e63840ecde499978c7dc8c0d89fe3e8cb165316/src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata/testing/simulator_synthetic_data_spec.proto#L138)):*

Document that this is required if any VidRangeSpec specifies a sample_size.
</blockquote></details>

do you think we should check if the random seed != 0 when there is a sample size or we can let the default value (0) be used?
___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata/testing/simulator_synthetic_data_spec.proto` line 149 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1425#-Np0yF0iD34t_ZYjLol5:-Np1636c4MxN74akxLbB:b-m6fmbr) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0e63840ecde499978c7dc8c0d89fe3e8cb165316/src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata/testing/simulator_synthetic_data_spec.proto#L149)):*

Wrong comment copy-pasted from somewhere. Also document that this is required if `random_seed` is specified.
</blockquote></details>

thanks for spotting

SanjayVas (changes requested):
> Reviewed 1 of 3 files at r4, 2 of 2 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1425)* status: all files reviewed, 3 unresolved discussions (waiting on @kungfucraig and @uakyol)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SyntheticDataGeneration.kt` line 58 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1425#-Np0z7sx6WIEg6KA4pm4:-Np1JT7VCvWl8r6TVuQy:b-nr8j8g) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0e63840ecde499978c7dc8c0d89fe3e8cb165316/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SyntheticDataGeneration.kt#L58)):*

ah, I would like to go with initing it here so we get variety even if VidRangeSpec s are the same but in different SyntheticDataSpecs that has other VidRangeSpec. 

Do you think it is enough to document this in the spec or we do you think we should add a field in the VidRangeSpec such as "id" or "orderNumber"?
</blockquote></details>

I'm thinking two options:
1. Document in the protobuf that the sampling must be done for the VidRangeSpecs in the order that they appear.
2. Move the random_seed field to VidRangeSpec so it's effectively explicit in the API that we're re-initializing for each.

I personally prefer (2) as it makes the protobuf API a bit more self-documenting.
___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata/testing/simulator_synthetic_data_spec.proto` line 138 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1425#-Np0yKKMB-6sXr1Xr4uz:-Np1K5Wz5aX6alyXsjet:b-392fyw) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0e63840ecde499978c7dc8c0d89fe3e8cb165316/src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata/testing/simulator_synthetic_data_spec.proto#L138)):*

do you think we should check if the random seed != 0 when there is a sample size or we can let the default value (0) be used?
</blockquote></details>

Oh, if the default (`0`) is a valid value then you don't need to mark it as required.

kungfucraig (changes requested):
> Reviewed 1 of 3 files at r4, 2 of 2 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1425)* status: all files reviewed, 4 unresolved discussions (waiting on @SanjayVas and @uakyol)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata/testing/simulator_synthetic_data_spec.proto` line 116 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1425#-Np1UKq89Hfipphojalu:-Np1UKq89Hfipphojalv:bspvi5p) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/001cb321496af2645d9062acf73657fce93d9a55/src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata/testing/simulator_synthetic_data_spec.proto#L116)):*
> ```proto
>       // Number of vids sampled uniformly without replacement from vid_range.
>       // If this is 0, no sampling is done and all the vids in range are taken.
>       int32 sample_size = 3;
> ```

It occurs to me that expressing this as a fraction could be better. Don't feel too strongly, but thought I'd mention it.
___
*[`src/test/kotlin/org/wfanet/measurement/loadtest/dataprovider/SyntheticDataGenerationTest.kt` line 494 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1425#-Np1UyIg5lst28ziMH2y:-Np1UyIg5lst28ziMH2z:b60g47d) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/001cb321496af2645d9062acf73657fce93d9a55/src/test/kotlin/org/wfanet/measurement/loadtest/dataprovider/SyntheticDataGenerationTest.kt#L494)):*
> ```Kotlin
>     }
>   }
> ```

Test that fails for when RNG is not specified?

kungfucraig (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1425)* status: all files reviewed, 4 unresolved discussions (waiting on @SanjayVas and @uakyol)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SyntheticDataGeneration.kt` line 58 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1425#-Np0z7sx6WIEg6KA4pm4:-Np1Vz32238aM0JEIJha:b-hnl2rc) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0e63840ecde499978c7dc8c0d89fe3e8cb165316/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SyntheticDataGeneration.kt#L58)):*

I like #2 as well. More localized behavior. 

If you go this route, you should wrap sample_size and the rng in a sampling spec inside of VIDRangeSpec.
</blockquote></details>

Just to be clear, I like the second option as well.

uakyol (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1425)* status: 0 of 3 files reviewed, 4 unresolved discussions (waiting on @kungfucraig and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SyntheticDataGeneration.kt` line 58 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1425#-Np0z7sx6WIEg6KA4pm4:-Np62uM05H_QbyT3RGfb:btxpkmy) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0e63840ecde499978c7dc8c0d89fe3e8cb165316/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SyntheticDataGeneration.kt#L58)):*

Just to be clear, I like the second option as well.
</blockquote></details>

went with 2. thanks
___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata/testing/simulator_synthetic_data_spec.proto` line 138 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1425#-Np0yKKMB-6sXr1Xr4uz:-Np63JAF1pWCtZfPW6cL:beh4o2u) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0e63840ecde499978c7dc8c0d89fe3e8cb165316/src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata/testing/simulator_synthetic_data_spec.proto#L138)):*

Oh, if the default (`0`) is a valid value then you don't need to mark it as required.
</blockquote></details>

thanks, took out the required comment
___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata/testing/simulator_synthetic_data_spec.proto` line 116 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1425#-Np1UKq89Hfipphojalu:-Np64xpy1W2wQGWEu7eE:bg9w745) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/001cb321496af2645d9062acf73657fce93d9a55/src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata/testing/simulator_synthetic_data_spec.proto#L116)):*

It occurs to me that expressing this as a fraction could be better. Don't feel too strongly, but thought I'd mention it.
</blockquote></details>

thank you, I think it will add a layer of doing math. The use case I see is that we give the vidrangespec a large range of vids to sample from and just tweak the sample sizes. I think it might not be as good to tie the sample size calculation to the pool of vids in the VidRangeSpec itself.
___
*[`src/test/kotlin/org/wfanet/measurement/loadtest/dataprovider/SyntheticDataGenerationTest.kt` line 494 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1425#-Np1UyIg5lst28ziMH2y:-Np64u1x-Jj3blj9R_1V:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/001cb321496af2645d9062acf73657fce93d9a55/src/test/kotlin/org/wfanet/measurement/loadtest/dataprovider/SyntheticDataGenerationTest.kt#L494)):*

Test that fails for when RNG is not specified?
</blockquote></details>

Done.

SanjayVas (changes requested):
> Reviewed 3 of 3 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1425)* status: all files reviewed, 4 unresolved discussions (waiting on @kungfucraig and @uakyol)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata/testing/simulator_synthetic_data_spec.proto` line 149 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1425#-Np6QVyC6Ul76MhkGC80:-Np6QVyC6Ul76MhkGC81:b-ipjnc6) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8215f7ad9d1f826a095017304bf796d32a915130/src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata/testing/simulator_synthetic_data_spec.proto#L149)):*
> ```proto
>   }
> 
>   // Random Number Generator type for this `SyntheticEventGroupSpec`.
> ```

This still needs to be documented as required if sample_size is specified in any VidRangeSpec.

uakyol (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1425)* status: 0 of 3 files reviewed, 4 unresolved discussions (waiting on @kungfucraig and @SanjayVas)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata/testing/simulator_synthetic_data_spec.proto` line 149 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1425#-Np6QVyC6Ul76MhkGC80:-Np6TXA-4QbUxj1gSseV:b-tfu8bq) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8215f7ad9d1f826a095017304bf796d32a915130/src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata/testing/simulator_synthetic_data_spec.proto#L149)):*

This still needs to be documented as required if sample_size is specified in any VidRangeSpec.
</blockquote></details>

Thanks, also changed it up a little. Reflecting this comment, only checking rng type if sampling is required. this will allow us to keep the textprotos as is for the integration tests. 
We now double loop over the spec but performance implication will be negligible for the first loop since its not generating anything and just checking. 
wdyt?

SanjayVas (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1425)* status: all files reviewed, 3 unresolved discussions (waiting on @kungfucraig)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata/testing/simulator_synthetic_data_spec.proto` line 149 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1425#-Np6QVyC6Ul76MhkGC80:-Np6WtzD5DwZ2UrUkTmn:bb6ywmx) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8215f7ad9d1f826a095017304bf796d32a915130/src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata/testing/simulator_synthetic_data_spec.proto#L149)):*

Thanks, also changed it up a little. Reflecting this comment, only checking rng type if sampling is required. this will allow us to keep the textprotos as is for the integration tests. 
We now double loop over the spec but performance implication will be negligible for the first loop since its not generating anything and just checking. 
wdyt?
</blockquote></details>

LGTM

kungfucraig (approved):
> Reviewed 2 of 3 files at r6, 1 of 1 files at r7, 3 of 3 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1425)* status: all files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

stevenwarejones (changes requested):
> Reviewed 2 of 3 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1425)* status: all files reviewed, 3 unresolved discussions (waiting on @SanjayVas and @uakyol)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SyntheticDataGeneration.kt` line 121 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1425#-NpQnprh3HnA9AuwrQUi:-NpQnprh3HnA9AuwrQUj:best74z) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3c92dc8784b2fa169769466960bc7ad1bfbbacba/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SyntheticDataGeneration.kt#L121)):*
> ```Kotlin
>     val vidRangeSequence =
>       (vidRangeSpec.vidRange.start until vidRangeSpec.vidRange.endExclusive).asSequence()
>     if (vidRangeSpec.sampleSize == 0) {
> ```

```
if (vidRangeSpec.sampleSize <= vidRangeSequence.size())
```

___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata/testing/simulator_synthetic_data_spec.proto` line 147 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1425#-NpQmwgr2SOAd_6fTvtF:-NpQmwgr2SOAd_6fTvtG:b-3611ob) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3c92dc8784b2fa169769466960bc7ad1bfbbacba/src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata/testing/simulator_synthetic_data_spec.proto#L147)):*
> ```proto
>     RNG_TYPE_UNSPECIFIED = 0;
>     // Signals kotlin.random.Random should be used for sampling.
>     KOTLIN_RANDOM = 1;
> ```

are there any use cases where we might want to support a secure random? If not, why do we even need a random number generator type?

SanjayVas (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1425)* status: all files reviewed, 3 unresolved discussions (waiting on @stevenwarejones and @uakyol)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata/testing/simulator_synthetic_data_spec.proto` line 147 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1425#-NpQmwgr2SOAd_6fTvtF:-NpRs7_e4IiOGCa63BFX:b-c3j3q5) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3c92dc8784b2fa169769466960bc7ad1bfbbacba/src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata/testing/simulator_synthetic_data_spec.proto#L147)):*

are there any use cases where we might want to support a secure random? If not, why do we even need a random number generator type?
</blockquote></details>

The requirement is that the output be deterministic for a given spec definition. In order to guarantee that, we need to know which RNG impl is being used.

The next step wouldn't be adding a secure random, but a language-agnostic standard PRNG impl such as AES.

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1425)* status: all files reviewed, 2 unresolved discussions (waiting on @SanjayVas and @uakyol)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata/testing/simulator_synthetic_data_spec.proto` line 147 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1425#-NpQmwgr2SOAd_6fTvtF:-NpSBqZgCEVLhFSey9tQ:b-8hkut7) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3c92dc8784b2fa169769466960bc7ad1bfbbacba/src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata/testing/simulator_synthetic_data_spec.proto#L147)):*

The requirement is that the output be deterministic for a given spec definition. In order to guarantee that, we need to know which RNG impl is being used.

The next step wouldn't be adding a secure random, but a language-agnostic standard PRNG impl such as AES.
</blockquote></details>

Hmmm.... Should we just start with some language agnostic impl for now? Seems odd to tie a proto to a particular language then. Or maybe just roll out something simple ourselves:
https://www.codeproject.com/Articles/9187/A-fast-equivalent-for-System-Random
https://www.codeproject.com/Articles/164087/Random-Number-Generation

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1425)* status: all files reviewed, 3 unresolved discussions (waiting on @SanjayVas and @uakyol)

SanjayVas (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1425)* status: all files reviewed, 3 unresolved discussions (waiting on @stevenwarejones and @uakyol)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata/testing/simulator_synthetic_data_spec.proto` line 147 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1425#-NpQmwgr2SOAd_6fTvtF:-NpSE3GSCg6cHpfJ08Nj:bn2dzp6) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3c92dc8784b2fa169769466960bc7ad1bfbbacba/src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata/testing/simulator_synthetic_data_spec.proto#L147)):*

Hmmm.... Should we just start with some language agnostic impl for now? Seems odd to tie a proto to a particular language then. Or maybe just roll out something simple ourselves:
https://www.codeproject.com/Articles/9187/A-fast-equivalent-for-System-Random
https://www.codeproject.com/Articles/164087/Random-Number-Generation
</blockquote></details>

We actually have one in the works as we need it for HMSS, but we want to roll out with a single impl for now since this is blocking 0.5.0. We don't strictly need the language-agnostic one until someone wants to implement this in another language, perhaps when we're adding conformance testing.

uakyol:
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SyntheticDataGeneration.kt` line 121 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1425#-NpQnprh3HnA9AuwrQUi:-NpVQ2WDAQ_b3iS30DhS:blj70wd) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3c92dc8784b2fa169769466960bc7ad1bfbbacba/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SyntheticDataGeneration.kt#L121)):*

```
if (vidRangeSpec.sampleSize <= vidRangeSequence.size())
```

</blockquote></details>

sorry, I couldn't understand this comment. vidRangeSpec.sampleSize should be smaller than vidRangeSequence.size() for there to be sampling. This if statement covers the case for when sampling is  not requested. We treat if the sampleSize field is set to default value "0" then there should be no sampling.

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1425)* status: all files reviewed, 3 unresolved discussions (waiting on @SanjayVas and @uakyol)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SyntheticDataGeneration.kt` line 121 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1425#-NpQnprh3HnA9AuwrQUi:-NpVUK4_--lAiVmqonL7:b-8cn8qk) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3c92dc8784b2fa169769466960bc7ad1bfbbacba/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SyntheticDataGeneration.kt#L121)):*

sorry, I couldn't understand this comment. vidRangeSpec.sampleSize should be smaller than vidRangeSequence.size() for there to be sampling. This if statement covers the case for when sampling is  not requested. We treat if the sampleSize field is set to default value "0" then there should be no sampling.
</blockquote></details>

nvm. What I was saying is not actually relevant here as I take a second look.

Should we throw an error if sampleSize is larger than size of vidRangeSequence?
___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata/testing/simulator_synthetic_data_spec.proto` line 147 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1425#-NpQmwgr2SOAd_6fTvtF:-NpVVWJOBBTQlHXjbNM-:b-7xvhom) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3c92dc8784b2fa169769466960bc7ad1bfbbacba/src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata/testing/simulator_synthetic_data_spec.proto#L147)):*

We actually have one in the works as we need it for HMSS, but we want to roll out with a single impl for now since this is blocking 0.5.0. We don't strictly need the language-agnostic one until someone wants to implement this in another language, perhaps when we're adding conformance testing.
</blockquote></details>

I'd prefer to roll out our own since a language specific one will almost certainly be deprecated in the future.

SanjayVas (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1425)* status: all files reviewed, 3 unresolved discussions (waiting on @stevenwarejones and @uakyol)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata/testing/simulator_synthetic_data_spec.proto` line 147 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1425#-NpQmwgr2SOAd_6fTvtF:-NpVXWRpAfA2VIawdPbd:b8qfqlw) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3c92dc8784b2fa169769466960bc7ad1bfbbacba/src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata/testing/simulator_synthetic_data_spec.proto#L147)):*

I'd prefer to roll out our own since a language specific one will almost certainly be deprecated in the future.
</blockquote></details>

I agree that a language-agnostic one would likely fully supplant a language-specific one, but we don't actually need the language-agnostic one any time soon. We'd only need for conformance testing *if* we plan on using sampling for those specs. 

More importantly, waiting on the language-agnostic one would push 0.5.0 back a few weeks. Getting this in is critical for Origin.

uakyol (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1425)* status: 1 of 3 files reviewed, 3 unresolved discussions (waiting on @kungfucraig, @SanjayVas, and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SyntheticDataGeneration.kt` line 121 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1425#-NpQnprh3HnA9AuwrQUi:-NpVd4OW0WHw1WnEKO_f:bbx3y1l) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3c92dc8784b2fa169769466960bc7ad1bfbbacba/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SyntheticDataGeneration.kt#L121)):*

nvm. What I was saying is not actually relevant here as I take a second look.

Should we throw an error if sampleSize is larger than size of vidRangeSequence?
</blockquote></details>

Thanks for catching, I thought I was checking this. Added check and test

kungfucraig (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1425)* status: all files reviewed, 3 unresolved discussions (waiting on @SanjayVas and @stevenwarejones)

stevenwarejones (changes requested):
> Reviewed 1 of 2 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1425)* status: all files reviewed, 2 unresolved discussions (waiting on @SanjayVas and @uakyol)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata/testing/simulator_synthetic_data_spec.proto` line 147 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1425#-NpQmwgr2SOAd_6fTvtF:-NpWFLXRDixfMZZjbt2k:b-tb0bs9) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3c92dc8784b2fa169769466960bc7ad1bfbbacba/src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata/testing/simulator_synthetic_data_spec.proto#L147)):*

I agree that a language-agnostic one would likely fully supplant a language-specific one, but we don't actually need the language-agnostic one any time soon. We'd only need for conformance testing *if* we plan on using sampling for those specs. 

More importantly, waiting on the language-agnostic one would push 0.5.0 back a few weeks. Getting this in is critical for Origin.
</blockquote></details>

Lets not make it part of the proto then. We can add it to the proto later once we support a platform-independent method.

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1425)* status: all files reviewed, 2 unresolved discussions (waiting on @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata/testing/simulator_synthetic_data_spec.proto` line 147 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1425#-NpQmwgr2SOAd_6fTvtF:-NpWJm_L2RPaRJzFABv7:b-bqy71v) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3c92dc8784b2fa169769466960bc7ad1bfbbacba/src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata/testing/simulator_synthetic_data_spec.proto#L147)):*

Lets not make it part of the proto then. We can add it to the proto later once we support a platform-independent method.
</blockquote></details>

It's specifically *because* we intend to support different RNG implementations that it **must** be in the spec definition. Or more accurately, the fact that the RNG implementation affects the output it **must** be in the spec definition.

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1425)* status: all files reviewed, 2 unresolved discussions (waiting on @SanjayVas and @uakyol)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata/testing/simulator_synthetic_data_spec.proto` line 147 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1425#-NpQmwgr2SOAd_6fTvtF:-NpWgUyr0GCKsqJ5fbSC:b-hkrhva) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3c92dc8784b2fa169769466960bc7ad1bfbbacba/src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata/testing/simulator_synthetic_data_spec.proto#L147)):*

It's specifically *because* we intend to support different RNG implementations that it **must** be in the spec definition. Or more accurately, the fact that the RNG implementation affects the output it **must** be in the spec definition.
</blockquote></details>

Then I'd prefer to roll out something very simple - roll out our own - I don't think it should take weeks.

uakyol (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1425)* status: 0 of 3 files reviewed, 2 unresolved discussions (waiting on @kungfucraig, @SanjayVas, and @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata/testing/simulator_synthetic_data_spec.proto` line 147 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1425#-NpQmwgr2SOAd_6fTvtF:-Np_ERwl-r3SRXVmt2Md:bk0f5if) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3c92dc8784b2fa169769466960bc7ad1bfbbacba/src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata/testing/simulator_synthetic_data_spec.proto#L147)):*

Then I'd prefer to roll out something very simple - roll out our own - I don't think it should take weeks.
</blockquote></details>

updated based on offline conversation and used java.util.Random

stevenwarejones (approved):
> Reviewed 3 of 3 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1425)* status: all files reviewed, 1 unresolved discussion (waiting on @kungfucraig and @SanjayVas)

kungfucraig (approved):
> Reviewed 2 of 3 files at r10, 1 of 1 files at r12, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1425)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @SanjayVas and @stevenwarejones)

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1425)* status: all files reviewed, 1 unresolved discussion

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SyntheticDataGeneration.kt` line 135 at r12](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1425#-NpajxNM19jyRE8wTMlT:-NpajxNN3wLonuNPvGBC:bp02hj4) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/2aeb8cb080e58f9e5801bb5f0fc246855344149d/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SyntheticDataGeneration.kt#L135)):*
> ```Kotlin
>       return vidRangeSequence
>     }
>     return vidRangeSequence.shuffled(random).take(vidRangeSpec.sampleSize)
> ```

an integrator would need to know what shuffle algorithm to use as well. I'd prefer we just stick with `Collections.shuffle` for that

PR #1433, opened 2024-01-22
Allow retry of Report creation when Metric creation fails and request id is not used

tristanvuong2021:
This is an attempt at fixing the situation when only the report id is sent in the create report request, but it fails before the metrics are created. If that happens, retrying the report creation returns an ALREADY_EXISTS.

### DIFF
--- a/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/ReportsService.kt
+++ b/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/ReportsService.kt
@@ -317,34 +317,84 @@ class ReportsService(
       try {
         internalReportsStub.createReport(internalCreateReportRequest)
       } catch (e: StatusException) {
-        throw when (e.status.code) {
+        when (e.status.code) {
             Status.Code.DEADLINE_EXCEEDED ->
-              Status.DEADLINE_EXCEEDED.withDescription("Unable to create Report.")
-            Status.Code.CANCELLED -> Status.CANCELLED.withDescription("Unable to create Report.")
+              throw Status.DEADLINE_EXCEEDED.withDescription("Unable to create Report.")
+                .withCause(e)
+                .asRuntimeException()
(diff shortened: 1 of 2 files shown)


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1433)

SanjayVas (commented):
> This is an attempt at fixing the situation when only the report id is sent in the create report request, but it fails before the metrics are created. If that happens, retrying the report creation returns an ALREADY_EXISTS.

This sounds like expected behavior. CreateReport is only idempotent (retryable) when using a request ID.

From [AIP-133](https://google.aip.dev/133):

> If a user tries to create a resource with an ID that would result in a duplicate resource name, the service **must** error with `ALREADY_EXISTS`.

This is only modified by [AIP-155](https://google.aip.dev/155) with the specification of a request ID.

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1433#-:-NomSdxv0_8GE_HjtlcS:b4000zc)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @riemanli)

tristanvuong2021 (commented):
The problem I am thinking of is that from the user end the creation of the report failed, but they get an ALREADY_EXISTS>

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1433#-:-NomXQVo1ICuT8-oXNAj:b-awtp0z)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @riemanli)

tristanvuong2021 (commented):
when retrying after the failure

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1433#-:-NomXd1g5gEXyhCMHVKV:b-iy47v2)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @riemanli)

PR #1434, opened 2024-01-22
Add a default handler for uncaught exceptions

jonmolle:
This requires a more comprehensive rework, but that would be a bigger change. This at least gives visibility to the errors, but does not call back to the kingdom or identify which step the failure came from (the traceback should help at least).


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1434)

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1434)* status: all files reviewed, 2 unresolved discussions (waiting on @jonmolle and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/panelmatch/client/launcher/CoroutineLauncher.kt` line 39 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1434#-NomWOqY7fkxOmAew-_v:-NomWOqY7fkxOmAew-_w:b-ncq3rr) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f3713dd42d7e16a8a9bd4bf7a45d2febfb166708/src/main/kotlin/org/wfanet/panelmatch/client/launcher/CoroutineLauncher.kt#L39)):*
> ```Kotlin
>   private val semaphore = if (maxCoroutines !== null) Semaphore(maxCoroutines) else null
> 
>   override suspend fun execute(step: ValidatedExchangeStep, attemptKey: ExchangeStepAttemptKey) {
> ```

This code looks fishy.

1. There's a suspending function that does not wait for all created coroutines to complete. Generally functions that launch coroutines in other scopes are not `suspend` functions, meaning in general a function should *either* suspend *or* launch a new coroutine in a different scope but not both.
2. Nothing is keeping track of the Job. *Something* should be keeping a reference to this so it can call `join()`
___
*[`src/main/kotlin/org/wfanet/panelmatch/client/launcher/CoroutineLauncher.kt` line 51 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1434#-NomU9VE6C-iObmYfR40:-NomU9VE6C-iObmYfR41:b-opnqi) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f3713dd42d7e16a8a9bd4bf7a45d2febfb166708/src/main/kotlin/org/wfanet/panelmatch/client/launcher/CoroutineLauncher.kt#L51)):*
> ```Kotlin
>     private val logger by loggerFor()
> 
>     val defaultHandler: CoroutineExceptionHandler = CoroutineExceptionHandler { _, e ->
> ```

Note that this case represents a programming error, not a behavioral error. That is to say that our code should never result in uncaught exceptions, and instead should make proper use of structured concurrency to make sure that such exceptions are either handled or propagated.

At minimum add a comment to this effect. Perhaps go so far as adding an exitProcess to kill the JVM.

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1434)* status: all files reviewed, 2 unresolved discussions (waiting on @jonmolle and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/panelmatch/client/launcher/CoroutineLauncher.kt` line 39 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1434#-NomWOqY7fkxOmAew-_v:-Nombh7MBaoGR2FX0th7:b9jbp0o) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f3713dd42d7e16a8a9bd4bf7a45d2febfb166708/src/main/kotlin/org/wfanet/panelmatch/client/launcher/CoroutineLauncher.kt#L39)):*

This code looks fishy.

1. There's a suspending function that does not wait for all created coroutines to complete. Generally functions that launch coroutines in other scopes are not `suspend` functions, meaning in general a function should *either* suspend *or* launch a new coroutine in a different scope but not both.
2. Nothing is keeping track of the Job. *Something* should be keeping a reference to this so it can call `join()`
</blockquote></details>

I see from the comment that the fact that this requires a rework is known. For now, add a TODO.

To be honest I'm not sure why this class exists. It doesn't manage its own CoroutineScope or even manage the Jobs created in the scope that's passed in. Since the caller controls the scope, it can just launch the coroutine that calls the suspending function itself. The only useful bit I see in here is the semaphore.

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1434)* status: 0 of 6 files reviewed, 2 unresolved discussions (waiting on @jonmolle and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/panelmatch/client/launcher/CoroutineLauncher.kt` line 39 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1434#-NomWOqY7fkxOmAew-_v:-NomjHa14y7PD-N1VNx9:bkvz4uu) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f3713dd42d7e16a8a9bd4bf7a45d2febfb166708/src/main/kotlin/org/wfanet/panelmatch/client/launcher/CoroutineLauncher.kt#L39)):*

I see from the comment that the fact that this requires a rework is known. For now, add a TODO.

To be honest I'm not sure why this class exists. It doesn't manage its own CoroutineScope or even manage the Jobs created in the scope that's passed in. Since the caller controls the scope, it can just launch the coroutine that calls the suspending function itself. The only useful bit I see in here is the semaphore.
</blockquote></details>

Consider this simpler alternative: #1436

jonmolle (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1434)* status: 0 of 6 files reviewed, 2 unresolved discussions (waiting on @SanjayVas and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/panelmatch/client/launcher/CoroutineLauncher.kt` line 39 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1434#-NomWOqY7fkxOmAew-_v:-NomjLra5PFaxM3cZK-b:b-uaqp9p) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f3713dd42d7e16a8a9bd4bf7a45d2febfb166708/src/main/kotlin/org/wfanet/panelmatch/client/launcher/CoroutineLauncher.kt#L39)):*

Consider this simpler alternative: #1436
</blockquote></details>

Agreed there. The JobLauncher isn't helpful and is actually a blocker for this. I believe the only use case is testing iff you don't want to use coroutines.

I already completed a rework, but I'm going to do another one. I believe the suspend is due to up- and down-stream suspending dependencies (ApiClient calls in particular). I'll take a look at that while I rework this. The semaphore should probably be replaced by limiting parallelism in the scope's dispatcher, tbh.
___
*[`src/main/kotlin/org/wfanet/panelmatch/client/launcher/CoroutineLauncher.kt` line 51 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1434#-NomU9VE6C-iObmYfR40:-NomjgCL9L07df5D0Z50:b6cvdvx) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f3713dd42d7e16a8a9bd4bf7a45d2febfb166708/src/main/kotlin/org/wfanet/panelmatch/client/launcher/CoroutineLauncher.kt#L51)):*

Note that this case represents a programming error, not a behavioral error. That is to say that our code should never result in uncaught exceptions, and instead should make proper use of structured concurrency to make sure that such exceptions are either handled or propagated.

At minimum add a comment to this effect. Perhaps go so far as adding an exitProcess to kill the JVM.
</blockquote></details>

It does, this is replaced by a caught exception. I'm merging this into ExchangeTaskExecutor.execute, which it should have always been part of, but for clarity I was trying to work within the confines of existing code to minimize the impact on the dev branch of 0.4.

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1434)* status: 0 of 6 files reviewed, 3 unresolved discussions (waiting on @jonmolle and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/panelmatch/client/launcher/CoroutineLauncher.kt` line 66 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1434#-Nompya5AiIOhDZCQn6_:-Nompya5AiIOhDZCQn6a:b-pi1yse) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d12361ada94e8d784ec79aaed62120c2247b240d/src/main/kotlin/org/wfanet/panelmatch/client/launcher/CoroutineLauncher.kt#L66)):*
> ```Kotlin
> 
>     runBlocking {
>       apiClient?.finishExchangeStepAttempt(attemptKeyName, attemptState)
> ```

i think you should try to parse the attemptKeyName here. Otherwise, its going to cause grpc errors when you call the kingdom. If its not a call attemptKeyName, then we need another way to report this error to the kingdom.

SanjayVas (changes requested):
> Reviewed 6 of 6 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1434)* status: all files reviewed, 2 unresolved discussions (waiting on @jonmolle)

___
*[`src/main/kotlin/org/wfanet/panelmatch/client/launcher/CoroutineLauncher.kt` line 39 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1434#-NomWOqY7fkxOmAew-_v:-NonELIX6EFiR2YWY7BL:bfm6vq0) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f3713dd42d7e16a8a9bd4bf7a45d2febfb166708/src/main/kotlin/org/wfanet/panelmatch/client/launcher/CoroutineLauncher.kt#L39)):*
> I believe the only use case is testing iff you don't want to use coroutines.

That's not actually true, as the interface has a suspending function. Therefore it's using coroutines regardless.

> Agreed there. The JobLauncher isn't helpful and is actually a blocker for this.

Should I be expecting to see that in this PR, or is that a separate one? Just trying to figure out if this is ready for another round of review.

> The semaphore should probably be replaced by limiting parallelism in the scope's dispatcher, tbh.

Note that that API is is technically still listed as experimental, but it's been around long enough that it's probably fine.

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1434)* status: all files reviewed, 2 unresolved discussions (waiting on @jonmolle and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/panelmatch/client/launcher/CoroutineLauncher.kt` line 39 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1434#-NomWOqY7fkxOmAew-_v:-NonJKKU7ZncS3JADY5i:b-fwwbyh) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f3713dd42d7e16a8a9bd4bf7a45d2febfb166708/src/main/kotlin/org/wfanet/panelmatch/client/launcher/CoroutineLauncher.kt#L39)):*

> I believe the only use case is testing iff you don't want to use coroutines.

That's not actually true, as the interface has a suspending function. Therefore it's using coroutines regardless.

> Agreed there. The JobLauncher isn't helpful and is actually a blocker for this.

Should I be expecting to see that in this PR, or is that a separate one? Just trying to figure out if this is ready for another round of review.

> The semaphore should probably be replaced by limiting parallelism in the scope's dispatcher, tbh.

Note that that API is is technically still listed as experimental, but it's been around long enough that it's probably fine.
</blockquote></details>

yeah - that's why I didn't use it in the scope's dispatcher. I think we've agreed to avoid experimental.

stevenwarejones (approved):
> Reviewed 18 of 18 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1434)* status: all files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

PR #1438, opened 2024-01-24
Add Get Report to the UI and add style fixes from the Demo version

bdomen-ggl:
Core code to implement Get Report view:
* client_impl: Help with TS errors (I think) and update property name.
* d3_wrapper.css: Helps fix the bar styles to they're positioned correctly.
* d3_wrapper: Update fields used for drawing. Implement Draw Bar method (this one wasn't used before).
* frequencies: Added as another chart we want to see.
* impressions: Added as another chart we want to see.
* index: Pointing to BFF.
* report, ...
* charts: Updating to be what we want to show.
* terminal_report, report_view: Generally updating to match the new report model and charts component.
* report_view_model: Mapping the data from the BFF model to the UI model.

Style tweaks:
* app.css, app.tsx: Push the bottom margin of the page so the feedback button doesn't cover the bottom elements of the page.
* chart: Fix window resizing but charts not resizing. D3 doesn't auto-resize so we have to do that manually. Added a chart type we ended up not using for completeness. Set the max height so full widths charts aren't really tall.
* d3_wrapper: Fix chart resizing. Fix max height. Change ticks count so it makes more sense for both lots and few points. Added a legend. 
* total_reach, unique_reach_by_plat: Make name flexible for daily vs cumulative.
* index.css: Fixing font.
* index.html: Adding Google Analytics, updating title, and fixing js bundle path.
* formatting: Adding the rest of the powers (0-30) and simplifying the logic.
* header.css: Making it sticky so the banner can sit below and not scroll off the top.
* overview: Changing the headings and adding a card.
* overview_card: Allow for non-numbers.
* summary_table.css: Allow for scrolling on small windows/devices.
* summary_table, summary_table_row: Removed a column
* header: Changing branding name.
* report_list_view, report_view: Adding alert - "for demo purposes".
*


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1438)

bdomen-ggl:
Going to redo this PR to be two, one with the core Get Report changes and a second with the general style changes.

PR #1439, opened 2024-01-25
Adding Get Report to the UI and updating models from the BFF changes

bdomen-ggl:
Notes:
* There were some minor changes like updating the connection to the BFF, but the work is generally around making Get Report work.
* Notable work:
  * Updated the BFF Report model to match the latest BFF response. This won't affect List Reports since it doesn't go through the time intervals.
  * Added logic to consume the report and package the data to send to each chart.
  * Fixed a chart rendering bug - It wouldn't resize when the window resized. D3 doesn't detect that automatically.
  * Added a general multi-bar chart (the existing one has percentages on the Y-axes and is a single bar).
  * Added two chart wrappers (impressions, frequencies).
  * Added filter by demographic capability but still working on adding the UI for it.
* Minor work:
  * Updated most of the charts (some were already good) so they all take a ChartGroup instead of a custom object, making it easier to pass to D3.
  * Styled the charts a little better like max height.
  * Added a legend to the charts.
  * Changed Report State enum to a string because that's how it comes from the BFF despite being an enum.
  * Extended the number to string formatter to handle more cases and handle default better.
  * Created the sticky header for the warning/info banner.

Future PRs:
* Finish filter logic.
* Fix test data.
* Update some of the setX logic to handle the else case for non-cumulative data (where applicable).
* Update the chart displaying so it's more flexible to different data sets (eg. cumulative and non-cumulative).


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1439)

SanjayVas (changes requested):
We generally prefer more granular, focused PRs. e.g. separate PRs for GetReport BFF integration and styling/UX tweaks.

> Reviewed 3 of 49 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1439#-:-Np6ReEoBZYTVUS7EHQD:b-4n0axy)* status: 3 of 49 files reviewed, 1 unresolved discussion (waiting on @bdomen-ggl)

___
*[`experimental/reporting-ui/src/main/react/reporting-ui/index.tsx` line 28 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1439#-Np6R5eB6kLT3J6f6DGo:-Np6R5eC0KGvFPKhyukj:b-qwwnro) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6155988fd9ed5654c6b8a8ca7d9946e5710cc397/experimental/reporting-ui/src/main/react/reporting-ui/index.tsx#L28)):*
> ```XML
>   reportingClient: new ReportingClientImpl(
>     {
>       endpoint: new URL('http://34.41.59.152:8080'),
> ```

I'm assuming this is halo-cmm-dev? I'm not sure we want our checked-in config to point to that. Let's leave it at localhost until we can actually make it configurable, and we can point to dev when manually deploying.

It'll need to be configurable before we can do automated deployment.



_Code quote:_
```XML
34.41.59.152
```

bdomen-ggl (commented):
Yeah I was generally trying to do that but there were a lot of additions for the demo that were overlapping in the same files. I tried to pull out what I could from non-related files and this is what was left. I can redo it so there are fewer changes at a time.

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1439#-:-NpL3-2g3eyEt_5s4ayq:bm3krd9)* status: 3 of 49 files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

___
*[`experimental/reporting-ui/src/main/react/reporting-ui/index.tsx` line 28 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1439#-Np6R5eB6kLT3J6f6DGo:-NpL3te16RjLNSHLkI__:b-km9jyo) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6155988fd9ed5654c6b8a8ca7d9946e5710cc397/experimental/reporting-ui/src/main/react/reporting-ui/index.tsx#L28)):*

I'm assuming this is halo-cmm-dev? I'm not sure we want our checked-in config to point to that. Let's leave it at localhost until we can actually make it configurable, and we can point to dev when manually deploying.

It'll need to be configurable before we can do automated deployment.
</blockquote></details>

I'll put this back and look into configuration. From what I saw at a glance, it looks like people use environment variables when using Webpack.

PR #1440, opened 2024-01-25
Cache Certificates in Metrics Service

tristanvuong2021:
Without this change, the number of RPC calls to the Certificates service scales with the number of measurements. which can be a significant number.


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1440)

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1440)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @riemanli and @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 257 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1440#-NpWnAfWBRDvH5DEJHcH:-NpWnAfXDp6FBMqnqSeI:botethp) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0cd96d1ab6368af2d8ff2e4c51ff4b677d7062d9/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L257)):*
> ```Kotlin
>   ) {
>     // Map of resource name to Certificate.
>     private val certificateMap: ConcurrentHashMap<String, Certificate> = ConcurrentHashMap()
> ```

This looks like it's caching for the lifetime of the server. This means we only pick up revocation state changes when the certificate is first fetched (we are checking the revocation state, right?).

This should probably cache for some maximum amount of time (however long we're willing to tolerate not knowing about a recent revocation)

tristanvuong2021 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1440)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @riemanli and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 257 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1440#-NpWnAfWBRDvH5DEJHcH:-Np_vU2P73fn_8bNqYNb:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4e3f72233ca62f4e632dfd0893480deb744636bc/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L257)):*

This looks like it's caching for the lifetime of the server. This means we only pick up revocation state changes when the certificate is first fetched (we are checking the revocation state, right?).

This should probably cache for some maximum amount of time (however long we're willing to tolerate not knowing about a recent revocation)
</blockquote></details>

Done.

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1440)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @riemanli and @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 269 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1440#-NpaR4qN2zUcKyjTZIIo:-NpaR4qO7REA7nf91c6O:b-ab1nxp) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ea12afb26b090317ba865e60c4ea7656b382215e/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L269)):*
> ```Kotlin
>       LoadingCache<ResourceNameApiAuthenticationKey, Deferred<Certificate>> =
>       CacheBuilder.newBuilder()
>         .refreshAfterWrite(30L, TimeUnit.MINUTES)
> ```

Have this be a default specified in the constructor that can be overridden. e.g. `certificateCacheLifetimeMinutes`



_Code quote:_
```Kotlin
30L
```
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 273 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1440#-NpaS7Vp9KYAJb1MwqyY:-NpaRRHQ5tfMDkXLBeJj:b-fiobd7) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ea12afb26b090317ba865e60c4ea7656b382215e/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L273)):*
> ```Kotlin
>           object : CacheLoader<ResourceNameApiAuthenticationKey, Deferred<Certificate>>() {
>             override fun load(key: ResourceNameApiAuthenticationKey): Deferred<Certificate> {
>               return runBlocking {
> ```

Using runBlocking here means that you're waiting for the `async` to finish before returning, which somewhat kills the point of returning the `Deferred`.

I believe what you really want is to use the Caffeine AsyncLoadingCache. This is in fact recommended in JavaDoc for Guava's CacheBuilder. You can use the utils in [kotinx-coroutines-guava](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-guava/) to handle the conversion between Guava futures and Kotlin Deferreds.



_Code quote:_
```Kotlin
runBlocking {
```

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1440)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @riemanli and @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 273 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1440#-NpaS7Vp9KYAJb1MwqyY:-NpaU02Q-rWZW2y_RcqJ:b-3weoyg) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ea12afb26b090317ba865e60c4ea7656b382215e/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L273)):*

Using runBlocking here means that you're waiting for the `async` to finish before returning, which somewhat kills the point of returning the `Deferred`.

I believe what you really want is to use the Caffeine AsyncLoadingCache. This is in fact recommended in JavaDoc for Guava's CacheBuilder. You can use the utils in [kotinx-coroutines-guava](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-guava/) to handle the conversion between Guava futures and Kotlin Deferreds.
</blockquote></details>

Or I guess AsyncCacheLoader uses CompletableFuture, which is part of the Java stdlib. The utils for that are in kotlinx.coroutines.future

tristanvuong2021 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1440)* status: 0 of 10 files reviewed, 2 unresolved discussions (waiting on @riemanli and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 269 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1440#-NpaR4qN2zUcKyjTZIIo:-NpaiMXR11QCiyM5xNE4:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ea12afb26b090317ba865e60c4ea7656b382215e/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L269)):*

Have this be a default specified in the constructor that can be overridden. e.g. `certificateCacheLifetimeMinutes`
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 273 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1440#-NpaS7Vp9KYAJb1MwqyY:-NpfQKVCDK9sO6cZLziW:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ea12afb26b090317ba865e60c4ea7656b382215e/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L273)):*

Or I guess AsyncCacheLoader uses CompletableFuture, which is part of the Java stdlib. The utils for that are in kotlinx.coroutines.future
</blockquote></details>

Done.

SanjayVas (changes requested):
> Reviewed 4 of 9 files at r4, 1 of 2 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1440)* status: 5 of 10 files reviewed, 5 unresolved discussions (waiting on @riemanli and @tristanvuong2021)

___
*[`MODULE.bazel` line 202 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1440#-Npf_g3sDfG_sUcORiW5:-Npf_g3sDfG_sUcORiW6:b-i6cly6) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/df068a278f6620c6205d76b595341202d9ec1678/MODULE.bazel#L202)):*
> ```Starlark
> maven.install(
>     artifacts = [
>         "com.github.ben-manes.caffeine:caffeine:3.1.8",
> ```

FYI for future reviewers: this dep is recommended by the Guava authors as a replacement to Guava's caching API.
___
*[`imports/java/com/github/ben-manes/caffeine/BUILD.bazel` line 1 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1440#-Npfa1LS4uWYVcKHg8yN:-Npfa1LS4uWYVcKHg8yO:b3j6lpp) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/df068a278f6620c6205d76b595341202d9ec1678/imports/java/com/github/ben-manes/caffeine/BUILD.bazel#L1)):*
> ```Starlark
> package(default_visibility = ["//visibility:public"])
> ```

The file path should match the Java package path. Therefore it should have `benmanes` not `ben-manes`
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 269 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1440#-NpaR4qN2zUcKyjTZIIo:-Npfad_56qaoRnvlavWc:b-2v1lhb) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ea12afb26b090317ba865e60c4ea7656b382215e/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L269)):*

Done.
</blockquote></details>

My bad, this should be a Duration `certificateCacheLifetime`even if we have to convert to a different time unit underneath (avoiding the "primitive obsession" code smell).
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 278 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1440#-NpfbIcC0ebExYNOFfmD:-NpfbIcDCmJlrLZl1n8H:bakgnw8) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/df068a278f6620c6205d76b595341202d9ec1678/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L278)):*
> ```Kotlin
>         .expireAfterWrite(certificateCacheExpirationMinutes, TimeUnit.MINUTES)
>         .buildAsync { key ->
>           CoroutineScope(Dispatchers.IO)
> ```

If you need a separate scope, declare it as a property so it can properly be tracked. Allow the dispatcher (CoroutineContext) to be passed in the constructor. Note that this isn't a blocking call, and therefore should be defaulted to e.g. Dispatchers.Default not Dispatchers.IO.

That said, I don't think you need this. You should be leveraging the executor in the cache builder as your dispatcher. This may mean wrapping calls to AsyncCache.get instead of supplying a loader to the cache builder. See the discussion in https://stackoverflow.com/questions/55270732/how-do-i-use-an-async-cache-with-kotlin-coroutines



_Code quote:_
```Kotlin
CoroutineScope(Dispatchers.IO)
```
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 282 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1440#-NpfbXvz1MhtjK2m-vhj:-NpfbXvz1MhtjK2m-vhk:bqcmogo) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/df068a278f6620c6205d76b595341202d9ec1678/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L282)):*
> ```Kotlin
>               getCertificate(name = key.name, apiAuthenticationKey = key.apiAuthenticationKey)
>             }
>             .join()
> ```

Calling join here again kills the point of having this be async. It's basically using the `CacheLoader` overload of `buioldAsync` instead of the `AsyncCacheLoader` overload which has you return a future.



_Code quote:_
```Kotlin
.join()
```

tristanvuong2021 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1440)* status: 2 of 10 files reviewed, 5 unresolved discussions (waiting on @riemanli and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 269 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1440#-NpaR4qN2zUcKyjTZIIo:-Npflkr09D4kzTfRygu4:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ea12afb26b090317ba865e60c4ea7656b382215e/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L269)):*

My bad, this should be a Duration `certificateCacheLifetime`even if we have to convert to a different time unit underneath (avoiding the "primitive obsession" code smell).
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 278 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1440#-NpfbIcC0ebExYNOFfmD:-NpfjXQSCK4h9hum5BhZ:b4rc2tv) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/df068a278f6620c6205d76b595341202d9ec1678/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L278)):*

If you need a separate scope, declare it as a property so it can properly be tracked. Allow the dispatcher (CoroutineContext) to be passed in the constructor. Note that this isn't a blocking call, and therefore should be defaulted to e.g. Dispatchers.Default not Dispatchers.IO.

That said, I don't think you need this. You should be leveraging the executor in the cache builder as your dispatcher. This may mean wrapping calls to AsyncCache.get instead of supplying a loader to the cache builder. See the discussion in https://stackoverflow.com/questions/55270732/how-do-i-use-an-async-cache-with-kotlin-coroutines
</blockquote></details>

The post recommends not using the executor so I just used the existing context.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 282 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1440#-NpfbXvz1MhtjK2m-vhj:-NpfglIC3u6Yx9YusTf8:b937tnv) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/df068a278f6620c6205d76b595341202d9ec1678/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L282)):*

Calling join here again kills the point of having this be async. It's basically using the `CacheLoader` overload of `buioldAsync` instead of the `AsyncCacheLoader` overload which has you return a future.
</blockquote></details>

The cache puts the values in futures already. Without this join, retrieving the value becomes get().await().await() instead of just get().await()
___
*[`imports/java/com/github/ben-manes/caffeine/BUILD.bazel` line 1 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1440#-Npfa1LS4uWYVcKHg8yN:-NpfhL-F9Zjwk1A5s058:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/df068a278f6620c6205d76b595341202d9ec1678/imports/java/com/github/ben-manes/caffeine/BUILD.bazel#L1)):*

The file path should match the Java package path. Therefore it should have `benmanes` not `ben-manes`
</blockquote></details>

Done.

SanjayVas (changes requested):
> Reviewed 6 of 7 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1440)* status: 8 of 10 files reviewed, 2 unresolved discussions (waiting on @riemanli and @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 278 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1440#-NpfbIcC0ebExYNOFfmD:-Npfr99S9BFgyW3fENHW:bdgscwv) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/df068a278f6620c6205d76b595341202d9ec1678/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L278)):*

The post recommends not using the executor so I just used the existing context.
</blockquote></details>

I'm confused as to how there's an existing context. Can you confirm what object `this` points to here?

Note that the post has some recommendations that definitely go against accepted best practices, so take that with a grain of salt. The advice there is referring to the fact that the default executor if not specified is ForkJoinPool, and basing the use of a separate dispatcher on that. That is okay for cases where you fully control the cache building, but IMO can be a little surprising as it means that overriding the executor in the builder would have no effect.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 282 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1440#-NpfbXvz1MhtjK2m-vhj:-NpfqVGD2J3RtAhuzr58:bybm4lx) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/df068a278f6620c6205d76b595341202d9ec1678/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L282)):*

The cache puts the values in futures already. Without this join, retrieving the value becomes get().await().await() instead of just get().await()
</blockquote></details>

You might as well not use `future` if that were the case. Look at the second `buildAsync` here demonstrating the case where you already have a future: https://github.com/ben-manes/caffeine/wiki/Population#asynchronously-loading

tristanvuong2021 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1440)* status: 8 of 10 files reviewed, 2 unresolved discussions (waiting on @riemanli and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 282 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1440#-NpfbXvz1MhtjK2m-vhj:-NpfuZku4hpzE76oxnUE:b-pxhniu) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/df068a278f6620c6205d76b595341202d9ec1678/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L282)):*

You might as well not use `future` if that were the case. Look at the second `buildAsync` here demonstrating the case where you already have a future: https://github.com/ben-manes/caffeine/wiki/Population#asynchronously-loading
</blockquote></details>

I figured out the correct usage now.

tristanvuong2021 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1440)* status: 8 of 10 files reviewed, 2 unresolved discussions (waiting on @riemanli and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 278 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1440#-NpfbIcC0ebExYNOFfmD:-NpfxgcJ7ByWGSeTMJ9w:bzflv75) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/df068a278f6620c6205d76b595341202d9ec1678/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L278)):*

I'm confused as to how there's an existing context. Can you confirm what object `this` points to here?

Note that the post has some recommendations that definitely go against accepted best practices, so take that with a grain of salt. The advice there is referring to the fact that the default executor if not specified is ForkJoinPool, and basing the use of a separate dispatcher on that. That is okay for cases where you fully control the cache building, but IMO can be a little surprising as it means that overriding the executor in the builder would have no effect.
</blockquote></details>

I removed the this, but the context is a variable that is passed into the metrics service

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1440)* status: 8 of 10 files reviewed, 2 unresolved discussions (waiting on @riemanli and @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 278 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1440#-NpfbIcC0ebExYNOFfmD:-Npfzm3s1gEQCo6vlnGk:b9ayvai) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/df068a278f6620c6205d76b595341202d9ec1678/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L278)):*

I removed the this, but the context is a variable that is passed into the metrics service
</blockquote></details>

Ah, so `this` points to `MeasurementSupplier` and therefore this is accessing the property of that class. That's not the right context to use, as that's specifically annotated with `@BlockingExecutor` to indicate it's a context with a dispatcher suitable for blocking calls. This is not a blocking call.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 279 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1440#-Npg--9c3mN9-K_itwHH:-Npg--9c3mN9-K_itwHI:b-hv76u6) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/9fdb05a21802ed3e8c063b1395a16085a64de880/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L279)):*
> ```Kotlin
>         .expireAfterWrite(certificateCacheExpirationDuration)
>         .buildAsync { key: ResourceNameApiAuthenticationKey, _ ->
>           CoroutineScope(coroutineContext).future {
> ```

This has similar problems to GlobalScope as it's a scope that lives beyond the call and has no parent.



_Code quote:_
```Kotlin
CoroutineScope(coroutineContext)
```

tristanvuong2021 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1440)* status: 8 of 10 files reviewed, 2 unresolved discussions (waiting on @riemanli and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 278 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1440#-NpfbIcC0ebExYNOFfmD:-Npg9X5z9opQNe-uBhfU:b-1jmstp) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/df068a278f6620c6205d76b595341202d9ec1678/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L278)):*

Ah, so `this` points to `MeasurementSupplier` and therefore this is accessing the property of that class. That's not the right context to use, as that's specifically annotated with `@BlockingExecutor` to indicate it's a context with a dispatcher suitable for blocking calls. This is not a blocking call.
</blockquote></details>

I just switched back to Dispatchers.IO. I thought network calls are considered blocking.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 279 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1440#-Npg--9c3mN9-K_itwHH:-Npg1iFD-ynrPgrXbINN:bo6qpga) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/9fdb05a21802ed3e8c063b1395a16085a64de880/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L279)):*

This has similar problems to GlobalScope as it's a scope that lives beyond the call and has no parent.
</blockquote></details>

Switched to a single scope for the cache.

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1440)* status: 8 of 10 files reviewed, 2 unresolved discussions (waiting on @riemanli and @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 278 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1440#-NpfbIcC0ebExYNOFfmD:-Npg9gTZCYlYIU2dZfPu:b8tkgtv) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/df068a278f6620c6205d76b595341202d9ec1678/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L278)):*

I just switched back to Dispatchers.IO. I thought network calls are considered blocking.
</blockquote></details>

Blocking means it blocks but does not suspend. If it's a suspending function, it's inherently not blocking (and it should be wrapping any blocking code it calls in an appropriate dispatcher).
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 279 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1440#-Npg--9c3mN9-K_itwHH:-NpgA1IX2w9vzGDegt3Y:b-gp5hdw) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/9fdb05a21802ed3e8c063b1395a16085a64de880/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L279)):*

Switched to a single scope for the cache.
</blockquote></details>

That scope is still unparented and unmanaged.

Here, I've draft a PR for how I think this should work: https://github.com/world-federation-of-advertisers/common-jvm/pull/232. You pass in an already-built non-loading AsyncCache and your suspending load function, and your interface consists also of suspending functions.

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1440)* status: 8 of 10 files reviewed, 2 unresolved discussions (waiting on @riemanli and @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 279 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1440#-Npg--9c3mN9-K_itwHH:-NpgAVOB998uGI2rB4IN:b-ovywfg) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/9fdb05a21802ed3e8c063b1395a16085a64de880/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L279)):*

That scope is still unparented and unmanaged.

Here, I've draft a PR for how I think this should work: https://github.com/world-federation-of-advertisers/common-jvm/pull/232. You pass in an already-built non-loading AsyncCache and your suspending load function, and your interface consists also of suspending functions.
</blockquote></details>

If you want the caller of MetricsService to be able to pass in the CoroutineContext for the cache, you'll need to extract the dispatcher from it and convert it to an executor to specify in the AsyncCache builder.

e.g.

```kotlin
(context.get(ContinuationInterceptor) as CoroutineDispatcher).asExecutor()
```

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1440)* status: 8 of 10 files reviewed, 3 unresolved discussions (waiting on @riemanli and @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 267 at r9](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1440#-NpgBvUC8YmT92zxkahr:-NpgBvUC8YmT92zxkahs:bwwpww5) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4e0a8602f90f09909e8f3456394bb3f9d714d638/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L267)):*
> ```Kotlin
>     certificateCacheRefreshDuration: JavaDuration,
>     certificateCacheExpirationDuration: JavaDuration,
>     private val coroutineContext: @BlockingExecutor CoroutineContext = Dispatchers.IO,
> ```

Add separate contexts for different use cases.



_Suggestion:_
```Kotlin
private val keyReaderContext: @BlockingExecutor CoroutineContext = Dispatchers.IO,
private val certificateLoaderContext: @NonBlockingExecutor CoroutineContext = Dispatchers.Default,
```

tristanvuong2021 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1440)* status: 0 of 11 files reviewed, 3 unresolved discussions (waiting on @riemanli and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 278 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1440#-NpfbIcC0ebExYNOFfmD:-NpuxqmG82XcJfq-y4rt:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/df068a278f6620c6205d76b595341202d9ec1678/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L278)):*

Blocking means it blocks but does not suspend. If it's a suspending function, it's inherently not blocking (and it should be wrapping any blocking code it calls in an appropriate dispatcher).
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 279 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1440#-Npg--9c3mN9-K_itwHH:-NpuxtmB-lf6sSy_mk9T:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/9fdb05a21802ed3e8c063b1395a16085a64de880/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L279)):*

If you want the caller of MetricsService to be able to pass in the CoroutineContext for the cache, you'll need to extract the dispatcher from it and convert it to an executor to specify in the AsyncCache builder.

e.g.

```kotlin
(context.get(ContinuationInterceptor) as CoroutineDispatcher).asExecutor()
```
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 267 at r9](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1440#-NpgBvUC8YmT92zxkahr:-Npuy3ua9SEmhLgoFle2:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4e0a8602f90f09909e8f3456394bb3f9d714d638/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L267)):*

Add separate contexts for different use cases.
</blockquote></details>

Done.

SanjayVas (changes requested):
> Reviewed 5 of 9 files at r10, 3 of 3 files at r11, 1 of 1 files at r13, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1440)* status: 9 of 11 files reviewed, 3 unresolved discussions (waiting on @riemanli and @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 37 at r13](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1440#-Nq04SZmBCdsrxJqiQUB:-Nq04SZmBCdsrxJqiQUC:b8hw0ea) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7791ccddf24f5ee1283f834329d2b0307f695ef0/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L37)):*
> ```Kotlin
> import java.security.cert.CertPathValidatorException
> import java.security.cert.X509Certificate
> import java.time.Duration as JavaDuration
> ```

nit: prefer keeping the Java stdlib one as Duration and importing the protobuf one as ProtoDuration
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 578 at r13](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1440#-Nq07o3B6TzZgspHlukT:-Nq07o3B6TzZgspHlukU:b-5rtql5) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7791ccddf24f5ee1283f834329d2b0307f695ef0/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L578)):*
> ```Kotlin
>                     )
>                   )
>                 } catch (e: UncheckedExecutionException) {
> ```

I don't think LoadingCache throws this exception. It should just pass through any exceptions from the `load` lambda. See LoadingCacheTest.



_Code quote:_
```Kotlin
UncheckedExecutionException
```
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 1087 at r13](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1440#-Nq08tmJDLiiAbLjVKh2:-Nq08tmJDLiiAbLjVKh3:bg5oiqk) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7791ccddf24f5ee1283f834329d2b0307f695ef0/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L1087)):*
> ```Kotlin
>           .getCertificate(getCertificateRequest { this.name = name })
>       } catch (e: StatusException) {
>         throw when (e.status.code) {
> ```

This doesn't match the method docs, which state that it results in FAILED_PRECONDITION. Note that it's only safe to throw StatusRuntimeException from here if you know what context it's always called from. Regardless, I doubt this should ever result in a NOT_FOUND for any method on Metrics service.

tristanvuong2021 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1440)* status: 8 of 11 files reviewed, 2 unresolved discussions (waiting on @riemanli and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 578 at r13](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1440#-Nq07o3B6TzZgspHlukT:-Nq3KjLW0snesD_0C9T6:b2gz1e3) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7791ccddf24f5ee1283f834329d2b0307f695ef0/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L578)):*

I don't think LoadingCache throws this exception. It should just pass through any exceptions from the `load` lambda. See LoadingCacheTest.
</blockquote></details>

The common-jvm LoadingCache doesn't, but AsyncCache does when the lambda throws an exception. It wraps the exception with this exception.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 1087 at r13](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1440#-Nq08tmJDLiiAbLjVKh2:-Nq3Lvl4--GRWC5KgqV-:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7791ccddf24f5ee1283f834329d2b0307f695ef0/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L1087)):*

This doesn't match the method docs, which state that it results in FAILED_PRECONDITION. Note that it's only safe to throw StatusRuntimeException from here if you know what context it's always called from. Regardless, I doubt this should ever result in a NOT_FOUND for any method on Metrics service.
</blockquote></details>

Done.

SanjayVas (changes requested):
> Reviewed 1 of 2 files at r14.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1440)* status: 9 of 11 files reviewed, 1 unresolved discussion (waiting on @riemanli and @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 578 at r13](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1440#-Nq07o3B6TzZgspHlukT:-Nq3YoY3-kYPqD7H6ywz:bghvqm) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7791ccddf24f5ee1283f834329d2b0307f695ef0/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L578)):*

The common-jvm LoadingCache doesn't, but AsyncCache does when the lambda throws an exception. It wraps the exception with this exception.
</blockquote></details>

IIUC that's only true for AsyncLoadingCache. LoadingCacheTest has a test method that verifies that the exception is passed-through without wrapping.

tristanvuong2021 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1440)* status: 8 of 11 files reviewed, 1 unresolved discussion (waiting on @riemanli and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 578 at r13](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1440#-Nq07o3B6TzZgspHlukT:-Nq3dsn92z8LjuCZKZlq:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7791ccddf24f5ee1283f834329d2b0307f695ef0/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L578)):*

IIUC that's only true for AsyncLoadingCache. LoadingCacheTest has a test method that verifies that the exception is passed-through without wrapping.
</blockquote></details>

Done.

SanjayVas (approved):
> Reviewed 1 of 9 files at r10, 1 of 2 files at r14, 1 of 1 files at r15, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1440)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @riemanli)

stevenwarejones (approved):
> Reviewed 6 of 9 files at r10, 3 of 3 files at r11, 1 of 2 files at r14, 1 of 1 files at r16, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1440)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @riemanli)

PR #1444, opened 2024-01-26
Drop Coroutine Launcher

jonmolle:
Multiple bugfixes for panel exchange

1. Remove CoroutineLauncher
2. Cancel child threads on task failure
3. Fix some tests that were not updated properly in earlier builds
4. Attempt to improve logging.


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1444)

SanjayVas (changes requested):
> Reviewed 2 of 20 files at r1, 2 of 11 files at r2.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1444)* status: 4 of 21 files reviewed, 4 unresolved discussions (waiting on @jonmolle and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/panelmatch/client/launcher/ExchangeStepLauncher.kt` line 29 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1444#-Np6DWxtET4x2vWpyu3M:-Np6DRD0A2jVDPQsileY:bjodnqk) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8d45bd905d5cfc96f61e3686b8bca858c2ba21d9/src/main/kotlin/org/wfanet/panelmatch/client/launcher/ExchangeStepLauncher.kt#L29)):*
> ```Kotlin
>    */
>   suspend fun findAndRunExchangeStep() {
>     val result = runCatching {
> ```

Just use a regular try-catch here, using the Logger overloads that take an exception.
___
*[`src/main/kotlin/org/wfanet/panelmatch/client/launcher/ExchangeTaskExecutor.kt` line 67 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1444#-Np6Du-Z8ezu2kc1Fgu_:-Np6Du-Z8ezu2kc1Fgua:b-gizpi4) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8d45bd905d5cfc96f61e3686b8bca858c2ba21d9/src/main/kotlin/org/wfanet/panelmatch/client/launcher/ExchangeTaskExecutor.kt#L67)):*
> ```Kotlin
>   ): CoroutineExceptionHandler {
>     return CoroutineExceptionHandler { _, e ->
>       logger.severe("Uncaught Exception in child coroutine:")
> ```

Add a TODO to remove this once all exceptions are being handled properly. I personally think an exitProcess would be valid here, as I stress that if this is hit it indicates a programming error and no correctly-written application should hit this.
___
*[`src/main/kotlin/org/wfanet/panelmatch/client/launcher/ExchangeTaskExecutor.kt` line 69 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1444#-Np6DhhY55mDOOy3Jbq9:-Np6DhhY55mDOOy3JbqA:b-tequuh) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8d45bd905d5cfc96f61e3686b8bca858c2ba21d9/src/main/kotlin/org/wfanet/panelmatch/client/launcher/ExchangeTaskExecutor.kt#L69)):*
> ```Kotlin
>       logger.severe("Uncaught Exception in child coroutine:")
>       logger.severe(e.message)
>       logger.severe(e.stackTraceToString())
> ```

Use Logger overload that takes an exception and a message. That will include the stack trace and the exception message in addition to the log message.



_Code quote:_
```Kotlin
      logger.severe("Uncaught Exception in child coroutine:")
      logger.severe(e.message)
      logger.severe(e.stackTraceToString())
```
___
*[`src/main/kotlin/org/wfanet/panelmatch/client/launcher/ExchangeTaskExecutor.kt` line 83 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1444#-Np6EEAq9_mrIBSDj2QI:-Np6EEAq9_mrIBSDj2QJ:b-byy7uj) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8d45bd905d5cfc96f61e3686b8bca858c2ba21d9/src/main/kotlin/org/wfanet/panelmatch/client/launcher/ExchangeTaskExecutor.kt#L83)):*
> ```Kotlin
>   }
> 
>   suspend fun executeWithScope(
> ```

This does not follow coroutine conventions. Either suspend or launch in a separate scope, but not both. If the caller wants to launch the suspending function itself in a coroutine scope, it can do so directly.

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1444)* status: 4 of 21 files reviewed, 5 unresolved discussions (waiting on @jonmolle and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/panelmatch/client/launcher/ExchangeTaskExecutor.kt` line 83 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1444#-Np6vld6CWuFaCpyLGUY:-Np6vld798J8-jvEwikA:bc9bfng) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8d45bd905d5cfc96f61e3686b8bca858c2ba21d9/src/main/kotlin/org/wfanet/panelmatch/client/launcher/ExchangeTaskExecutor.kt#L83)):*
> ```Kotlin
>   }
> 
>   suspend fun executeWithScope(
> ```

This should be private. `execute` appears to be the public API, and all testing should be done against the public API. See https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/main/docs/code-style.md
___
*[`src/main/kotlin/org/wfanet/panelmatch/client/launcher/ExchangeTaskExecutor.kt` line 83 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1444#-Np6EEAq9_mrIBSDj2QI:-Np6xts-0NHEObD5Xeij:b-f12gn8) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8d45bd905d5cfc96f61e3686b8bca858c2ba21d9/src/main/kotlin/org/wfanet/panelmatch/client/launcher/ExchangeTaskExecutor.kt#L83)):*

This does not follow coroutine conventions. Either suspend or launch in a separate scope, but not both. If the caller wants to launch the suspending function itself in a coroutine scope, it can do so directly.
</blockquote></details>

Proposed fix: move `launch` into `execute` and just have this be a suspending function that does not take in a CoroutineScope nor return a Job.

As it is, a suspend function is basically just syntactic sugar for the following pseudocode:

```kotlin
coroutineScope {
  val job = launch { ... }
  job.join()
}
```

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1444)* status: 4 of 21 files reviewed, 5 unresolved discussions (waiting on @jonmolle and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/panelmatch/client/launcher/ExchangeTaskExecutor.kt` line 67 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1444#-Np6Du-Z8ezu2kc1Fgu_:-Np6yQN_4A8jeLcdNpys:b-87cjdk) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8d45bd905d5cfc96f61e3686b8bca858c2ba21d9/src/main/kotlin/org/wfanet/panelmatch/client/launcher/ExchangeTaskExecutor.kt#L67)):*

Add a TODO to remove this once all exceptions are being handled properly. I personally think an exitProcess would be valid here, as I stress that if this is hit it indicates a programming error and no correctly-written application should hit this.
</blockquote></details>

From offline discussion, this needs to mark a permanent failure instead of dying. Just document that this should never happen and anything that gets here should be investigated as a bug.

stevenwarejones (approved):
> Reviewed 10 of 20 files at r1, 11 of 11 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1444)* status: all files reviewed, 5 unresolved discussions (waiting on @jonmolle and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/panelmatch/client/launcher/ExchangeTaskExecutor.kt` line 67 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1444#-Np6Du-Z8ezu2kc1Fgu_:-NpL-Uae-y1dnlrX4-MP:bhd80g3) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8d45bd905d5cfc96f61e3686b8bca858c2ba21d9/src/main/kotlin/org/wfanet/panelmatch/client/launcher/ExchangeTaskExecutor.kt#L67)):*

From offline discussion, this needs to mark a permanent failure instead of dying. Just document that this should never happen and anything that gets here should be investigated as a bug.
</blockquote></details>

can that TODO exist? What if one of our dependencies doesn't handle it correctly (eg beam)

jonmolle (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1444)* status: 18 of 21 files reviewed, 5 unresolved discussions (waiting on @SanjayVas and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/panelmatch/client/launcher/ExchangeStepLauncher.kt` line 29 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1444#-Np6DWxtET4x2vWpyu3M:-Np6dCeG7EqT_MnBrpJr:bb6un1j) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8d45bd905d5cfc96f61e3686b8bca858c2ba21d9/src/main/kotlin/org/wfanet/panelmatch/client/launcher/ExchangeStepLauncher.kt#L29)):*

Just use a regular try-catch here, using the Logger overloads that take an exception.
</blockquote></details>

Done
___
*[`src/main/kotlin/org/wfanet/panelmatch/client/launcher/ExchangeTaskExecutor.kt` line 67 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1444#-Np6Du-Z8ezu2kc1Fgu_:-Np6dH2q1wQT0YggYwf8:bnoqqsn) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8d45bd905d5cfc96f61e3686b8bca858c2ba21d9/src/main/kotlin/org/wfanet/panelmatch/client/launcher/ExchangeTaskExecutor.kt#L67)):*

can that TODO exist? What if one of our dependencies doesn't handle it correctly (eg beam)
</blockquote></details>

We discussed this offline, but for context I have to stress that i don't think we can remove this in good conscience at the moment, or in the medium-term. I agree this should never be hit, but there are too many ways this can be hit later on. What I _do_ think we should do (and will begin working on) is change this to always cause a FAILED_STEP, so there is no retry. That way it is the equivalent of exiting the entire exchange until the fix can be implemented without blocking people whose exchanges work. It is too easy to create an edge case where an esoteric bad config could hit this.
___
*[`src/main/kotlin/org/wfanet/panelmatch/client/launcher/ExchangeTaskExecutor.kt` line 69 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1444#-Np6DhhY55mDOOy3Jbq9:-Np6ejCPC6-CLgiJ824C:bb6un1j) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8d45bd905d5cfc96f61e3686b8bca858c2ba21d9/src/main/kotlin/org/wfanet/panelmatch/client/launcher/ExchangeTaskExecutor.kt#L69)):*

Use Logger overload that takes an exception and a message. That will include the stack trace and the exception message in addition to the log message.
</blockquote></details>

Done

jonmolle (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1444)* status: 18 of 21 files reviewed, 5 unresolved discussions (waiting on @SanjayVas and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/panelmatch/client/launcher/ExchangeTaskExecutor.kt` line 83 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1444#-Np6EEAq9_mrIBSDj2QI:-NpLMZVT1m3mFrvvmT8_:bjlgajs) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8d45bd905d5cfc96f61e3686b8bca858c2ba21d9/src/main/kotlin/org/wfanet/panelmatch/client/launcher/ExchangeTaskExecutor.kt#L83)):*

Proposed fix: move `launch` into `execute` and just have this be a suspending function that does not take in a CoroutineScope nor return a Job.

As it is, a suspend function is basically just syntactic sugar for the following pseudocode:

```kotlin
coroutineScope {
  val job = launch { ... }
  job.join()
}
```
</blockquote></details>

We discussed this elsewhere and that was what we agreed on, done.
___
*[`src/main/kotlin/org/wfanet/panelmatch/client/launcher/ExchangeTaskExecutor.kt` line 83 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1444#-Np6vld6CWuFaCpyLGUY:-NpLMcc3BDmcSip8-FWR:b-u05au8) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8d45bd905d5cfc96f61e3686b8bca858c2ba21d9/src/main/kotlin/org/wfanet/panelmatch/client/launcher/ExchangeTaskExecutor.kt#L83)):*

This should be private. `execute` appears to be the public API, and all testing should be done against the public API. See https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/main/docs/code-style.md
</blockquote></details>

It isn't actually necessary, I've removed it. If there is a desire to keep all tasks in a shared supervisor scope this can be done another way.

stevenwarejones (changes requested):
> Reviewed 1 of 3 files at r3, 8 of 9 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1444)* status: 20 of 21 files reviewed, 8 unresolved discussions (waiting on @jonmolle and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/panelmatch/client/deploy/ExchangeWorkflowDaemon.kt` line 95 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1444#-NpLaLs84e6fu0JX6XYk:-NpLaLs84e6fu0JX6XYl:b59edyk) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e4dc1508883203e32da8e2875e7c4b7f90421a48/src/main/kotlin/org/wfanet/panelmatch/client/deploy/ExchangeWorkflowDaemon.kt#L95)):*
> ```Kotlin
>     SharedStorageSelector(certificateManager, sharedStorageFactories, sharedStorageInfo)
>   }
>   protected open val stepExecutor by lazy {
> ```

why is this open?
___
*[`src/main/kotlin/org/wfanet/panelmatch/client/launcher/ExchangeTaskExecutor.kt` line 176 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1444#-NpLb_kT9NPyN9Yzc0-f:-NpLb_kT9NPyN9Yzc0-g:bh28036) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e4dc1508883203e32da8e2875e7c4b7f90421a48/src/main/kotlin/org/wfanet/panelmatch/client/launcher/ExchangeTaskExecutor.kt#L176)):*
> ```Kotlin
> 
>   companion object {
> ```

what is the reason for this change?
___
*[`src/test/kotlin/org/wfanet/panelmatch/client/launcher/ExchangeTaskExecutorTest.kt` line 161 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1444#-NpLbtyu5CoGlOVdHuJc:-NpLbtyu5CoGlOVdHuJd:b-a86ubq) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e4dc1508883203e32da8e2875e7c4b7f90421a48/src/test/kotlin/org/wfanet/panelmatch/client/launcher/ExchangeTaskExecutorTest.kt#L161)):*
> ```Kotlin
> 
>     verify(apiClient).finishExchangeStepAttempt(eq(ATTEMPT_KEY), eq(State.FAILED_STEP), any())
>   }
> ```

can you add a test similar to what we're experiencing with Tink where a withContext is used within the task, itself, that then throws an error?

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1444)* status: 20 of 21 files reviewed, 8 unresolved discussions (waiting on @jonmolle and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/panelmatch/client/launcher/ExchangeTaskExecutor.kt` line 67 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1444#-Np6Du-Z8ezu2kc1Fgu_:-NpLdP6c2FoJZIRVD27i:bf5btq8) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8d45bd905d5cfc96f61e3686b8bca858c2ba21d9/src/main/kotlin/org/wfanet/panelmatch/client/launcher/ExchangeTaskExecutor.kt#L67)):*
<details><summary><i>Previously, jonmolle wrote…</i></summary><blockquote>

We discussed this offline, but for context I have to stress that i don't think we can remove this in good conscience at the moment, or in the medium-term. I agree this should never be hit, but there are too many ways this can be hit later on. What I _do_ think we should do (and will begin working on) is change this to always cause a FAILED_STEP, so there is no retry. That way it is the equivalent of exiting the entire exchange until the fix can be implemented without blocking people whose exchanges work. It is too easy to create an edge case where an esoteric bad config could hit this.
</blockquote></details>

I agree that this should remain as long as we're using `supervisorScope` as a just-in-case.

Can you remind me why we actually need `supervisorScope` though instead of just `coroutineScope` with proper exception handling within each launched job? Is there e.g. some other non-coroutine async library we're calling that has retry behavior (such as in [AsyncDatabaseClient](https://github.com/world-federation-of-advertisers/common-jvm/blob/3aa5f36cf937a1f82627722bfffd2240e52c2203/src/main/kotlin/org/wfanet/measurement/gcloud/spanner/AsyncDatabaseClient.kt#L303))?

jonmolle (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1444)* status: 20 of 21 files reviewed, 8 unresolved discussions (waiting on @SanjayVas and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/panelmatch/client/deploy/ExchangeWorkflowDaemon.kt` line 95 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1444#-NpLaLs84e6fu0JX6XYk:-NpLg-Cq1qul3dC98KUq:b7pnyg4) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e4dc1508883203e32da8e2875e7c4b7f90421a48/src/main/kotlin/org/wfanet/panelmatch/client/deploy/ExchangeWorkflowDaemon.kt#L95)):*

why is this open?
</blockquote></details>

It allows a subclass to add their own custom StepExecutor (and/or add their own dependencies). I actually think we ado this in the AWS implementation.

SanjayVas (changes requested):
> Reviewed 8 of 20 files at r1, 2 of 11 files at r2, 1 of 3 files at r3, 9 of 9 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1444)* status: all files reviewed, 5 unresolved discussions (waiting on @jonmolle and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/panelmatch/client/deploy/ExchangeWorkflowDaemonFromFlags.kt` line 55 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1444#-NpLfG0NBX3QL-EcRWDw:-NpLfG0NBX3QL-EcRWDx:bloubtb) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e4dc1508883203e32da8e2875e7c4b7f90421a48/src/main/kotlin/org/wfanet/panelmatch/client/deploy/ExchangeWorkflowDaemonFromFlags.kt#L55)):*
> ```Kotlin
>    */
>   override val maxParallelClaimedExchangeSteps: Int? by lazy {
>     if ((flags.maxParallelClaimedExchangeSteps ?: 1) < 1) {
> ```

This should be part of the flag validation rather than the daemon. See https://picocli.info/#_validation

Then this can just be
`get() = flags.maxParallelClaimedExchangeSteps` without the lazy delegate.

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1444)* status: all files reviewed, 5 unresolved discussions (waiting on @jonmolle)

___
*[`src/main/kotlin/org/wfanet/panelmatch/client/deploy/ExchangeWorkflowDaemon.kt` line 95 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1444#-NpLaLs84e6fu0JX6XYk:-NpLhFPn4dpcqT-yLLgb:b63m4f4) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e4dc1508883203e32da8e2875e7c4b7f90421a48/src/main/kotlin/org/wfanet/panelmatch/client/deploy/ExchangeWorkflowDaemon.kt#L95)):*
<details><summary><i>Previously, jonmolle wrote…</i></summary><blockquote>

It allows a subclass to add their own custom StepExecutor (and/or add their own dependencies). I actually think we ado this in the AWS implementation.
</blockquote></details>

the aws impl is identical. I don't see a need for this at the moment.

jonmolle (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1444)* status: all files reviewed, 5 unresolved discussions (waiting on @SanjayVas and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/panelmatch/client/launcher/ExchangeTaskExecutor.kt` line 67 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1444#-Np6Du-Z8ezu2kc1Fgu_:-NpLgQgtBeh77nJWIkOO:b-60dfy2) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8d45bd905d5cfc96f61e3686b8bca858c2ba21d9/src/main/kotlin/org/wfanet/panelmatch/client/launcher/ExchangeTaskExecutor.kt#L67)):*

I agree that this should remain as long as we're using `supervisorScope` as a just-in-case.

Can you remind me why we actually need `supervisorScope` though instead of just `coroutineScope` with proper exception handling within each launched job? Is there e.g. some other non-coroutine async library we're calling that has retry behavior (such as in [AsyncDatabaseClient](https://github.com/world-federation-of-advertisers/common-jvm/blob/3aa5f36cf937a1f82627722bfffd2240e52c2203/src/main/kotlin/org/wfanet/measurement/gcloud/spanner/AsyncDatabaseClient.kt#L303))?
</blockquote></details>

The supervisor scope is used for a common design pattern: to make sure a failed step in one recurring exchange does not crash the daemon or affect other recurring exchange tasks and is always reported up the chain to the orchestration layer (kingdom). There are a number of ways to customize dependencies and tasks (as well as add new tasks) that would happen in on exchange and not another. It is important that we identify errors and keep the exchange status as up-to-date as possible for all parties. Supervisor scopes are the standard for this from what I've seen.

I'll also add that it is pretty easy to have something like this crash the daemon and eat all the logs if improperly configured, so at least with this behaviour we can gracefully exit. It shouldn't ever get to that point, but again this is a safety net, not something we should ever accept will be hit as part of normal execution.
___
*[`src/main/kotlin/org/wfanet/panelmatch/client/launcher/ExchangeTaskExecutor.kt` line 176 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1444#-NpLb_kT9NPyN9Yzc0-f:-NpLhSrn8pEeVXkjzzU4:b-8qdm2k) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e4dc1508883203e32da8e2875e7c4b7f90421a48/src/main/kotlin/org/wfanet/panelmatch/client/launcher/ExchangeTaskExecutor.kt#L176)):*

what is the reason for this change?
</blockquote></details>

loggerFor() is a suspending function and didn't play nicely with the exception handler, which is not a suspending function. I know I added a runBlocking in there for the API call, but there isn't anything we lose by doing this vs loggerFor() in this class and I want to avoid things that feel like hacks when possible.

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1444)* status: all files reviewed, 5 unresolved discussions (waiting on @jonmolle and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/panelmatch/client/launcher/ExchangeTaskExecutor.kt` line 67 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1444#-Np6Du-Z8ezu2kc1Fgu_:-NpLthLG09B_zkoO8LBa:b-r4cxs0) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8d45bd905d5cfc96f61e3686b8bca858c2ba21d9/src/main/kotlin/org/wfanet/panelmatch/client/launcher/ExchangeTaskExecutor.kt#L67)):*
> The supervisor scope is used for a common design pattern

A common *discouraged* design pattern (at least from Google practices). Just ensuring all exceptions are caught and using `coroutineScope` is much simpler/safer. It's just a question of whether this is one of the *rare* cases where the behavior of supervisorScope is actually necessary and cannot be avoided. It's the use of supervisorScope that necessitates CoroutineExceptionHandler.

stevenwarejones (commented):
> Reviewed 1 of 9 files at r4.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1444)* status: all files reviewed, 4 unresolved discussions (waiting on @jonmolle and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/panelmatch/client/launcher/ExchangeTaskExecutor.kt` line 67 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1444#-Np6Du-Z8ezu2kc1Fgu_:-NpMNIm8CNLmwoT_Acy7:bg2ay4z) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8d45bd905d5cfc96f61e3686b8bca858c2ba21d9/src/main/kotlin/org/wfanet/panelmatch/client/launcher/ExchangeTaskExecutor.kt#L67)):*

> The supervisor scope is used for a common design pattern

A common *discouraged* design pattern (at least from Google practices). Just ensuring all exceptions are caught and using `coroutineScope` is much simpler/safer. It's just a question of whether this is one of the *rare* cases where the behavior of supervisorScope is actually necessary and cannot be avoided. It's the use of supervisorScope that necessitates CoroutineExceptionHandler.
</blockquote></details>

I believe that Eli's original intention was that no single exchange step execution should crash the launcher but instead it should be resilient. I think a better pattern would be to only catch certain types of errors and remove the Supervisor scope. Not crashing has actually turned into a bad behavior imo.

stevenwarejones (commented):
> Reviewed 2 of 5 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1444)* status: 18 of 21 files reviewed, 4 unresolved discussions (waiting on @jonmolle and @SanjayVas)

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1444)* status: 18 of 21 files reviewed, 5 unresolved discussions (waiting on @jonmolle and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/panelmatch/client/launcher/ExchangeTaskExecutor.kt` line 82 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1444#-NqZ5GJpA8kaDfOlarWt:-NqZ5GJpA8kaDfOlarWu:b-qsiugd) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5010f840d2d6a4cda593ab37de6ee532e01e79e4/src/main/kotlin/org/wfanet/panelmatch/client/launcher/ExchangeTaskExecutor.kt#L82)):*
> ```Kotlin
>           CoroutineName(attemptKey.toName()) +
>           TaskLog(attemptKey.toName()) +
>           taskExceptionHandler(attemptKey)
> ```

as discussed, lets just pull out the `taskExceptionHandler` into its own PR and close this one.

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1444)* status: 18 of 21 files reviewed, 5 unresolved discussions (waiting on @jonmolle and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/panelmatch/client/launcher/ExchangeTaskExecutor.kt` line 82 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1444#-NqZ5GJpA8kaDfOlarWt:-NqZV29s6_wbqLmvyRYV:b-ufsm92) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5010f840d2d6a4cda593ab37de6ee532e01e79e4/src/main/kotlin/org/wfanet/panelmatch/client/launcher/ExchangeTaskExecutor.kt#L82)):*

as discussed, lets just pull out the `taskExceptionHandler` into its own PR and close this one.
</blockquote></details>

The important part of this PR is cleaning up the usage of coroutine scopes to follow best practices w.r.t. structured concurrency. If we can drop supervisor scope then the exception handler isn't needed.

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1444)* status: 18 of 21 files reviewed, 5 unresolved discussions (waiting on @jonmolle and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/panelmatch/client/launcher/ExchangeTaskExecutor.kt` line 82 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1444#-NqZ5GJpA8kaDfOlarWt:-NqZcpB4E2QzXtmqwYly:b-2q36ug) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5010f840d2d6a4cda593ab37de6ee532e01e79e4/src/main/kotlin/org/wfanet/panelmatch/client/launcher/ExchangeTaskExecutor.kt#L82)):*

The important part of this PR is cleaning up the usage of coroutine scopes to follow best practices w.r.t. structured concurrency. If we can drop supervisor scope then the exception handler isn't needed.
</blockquote></details>

@jonmolle and I spoke to split this up. I think the taskExceptionHandler should be merged ASAP this week. the other could take a little longer to merge.

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1444)* status: 18 of 21 files reviewed, 5 unresolved discussions (waiting on @jonmolle and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/panelmatch/client/launcher/ExchangeTaskExecutor.kt` line 82 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1444#-NqZ5GJpA8kaDfOlarWt:-NqZsgyA5r0e_GJkKqM4:b-kaxvwa) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5010f840d2d6a4cda593ab37de6ee532e01e79e4/src/main/kotlin/org/wfanet/panelmatch/client/launcher/ExchangeTaskExecutor.kt#L82)):*

@jonmolle and I spoke to split this up. I think the taskExceptionHandler should be merged ASAP this week. the other could take a little longer to merge.
</blockquote></details>

SGTM if it's just a temporary stopgap

stevenwarejones (approved):
> Reviewed 10 of 10 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1444)* status: all files reviewed, 4 unresolved discussions (waiting on @jonmolle and @SanjayVas)

stevenwarejones (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1444)* status: all files reviewed, 4 unresolved discussions (waiting on @jonmolle and @SanjayVas)

jonmolle (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1444)* status: 20 of 21 files reviewed, 3 unresolved discussions (waiting on @SanjayVas and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/panelmatch/client/deploy/ExchangeWorkflowDaemon.kt` line 95 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1444#-NpLaLs84e6fu0JX6XYk:-NpLogqC8wizBZdeQwsv:b-5cfrq2) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e4dc1508883203e32da8e2875e7c4b7f90421a48/src/main/kotlin/org/wfanet/panelmatch/client/deploy/ExchangeWorkflowDaemon.kt#L95)):*

the aws impl is identical. I don't see a need for this at the moment.
</blockquote></details>

It pulls from the AWS overloads at the very least, I'm guessing so it grabs the right implementation of other variables? Not sure, but I'm happy to check it out and add it downstream. This is sort of out of scope for this PR.
___
*[`src/main/kotlin/org/wfanet/panelmatch/client/deploy/ExchangeWorkflowDaemonFromFlags.kt` line 55 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1444#-NpLfG0NBX3QL-EcRWDw:-NpLiXns-Dw9iJBKRM1W:b2gnix1) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e4dc1508883203e32da8e2875e7c4b7f90421a48/src/main/kotlin/org/wfanet/panelmatch/client/deploy/ExchangeWorkflowDaemonFromFlags.kt#L55)):*

This should be part of the flag validation rather than the daemon. See https://picocli.info/#_validation

Then this can just be
`get() = flags.maxParallelClaimedExchangeSteps` without the lazy delegate.
</blockquote></details>

Done

Done.
___
*[`src/main/kotlin/org/wfanet/panelmatch/client/launcher/ExchangeTaskExecutor.kt` line 67 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1444#-Np6Du-Z8ezu2kc1Fgu_:-NpNolIO5TIfAzqzCmyl:bu8a0a9) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8d45bd905d5cfc96f61e3686b8bca858c2ba21d9/src/main/kotlin/org/wfanet/panelmatch/client/launcher/ExchangeTaskExecutor.kt#L67)):*

I believe that Eli's original intention was that no single exchange step execution should crash the launcher but instead it should be resilient. I think a better pattern would be to only catch certain types of errors and remove the Supervisor scope. Not crashing has actually turned into a bad behavior imo.
</blockquote></details>

I've just removed the supervisor scope
___
*[`src/main/kotlin/org/wfanet/panelmatch/client/launcher/ExchangeTaskExecutor.kt` line 82 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1444#-NqZ5GJpA8kaDfOlarWt:-Nqe-3_B1hdH58nfubT7:b8qslrz) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5010f840d2d6a4cda593ab37de6ee532e01e79e4/src/main/kotlin/org/wfanet/panelmatch/client/launcher/ExchangeTaskExecutor.kt#L82)):*

SGTM if it's just a temporary stopgap
</blockquote></details>

Moved out of this PR
___
*[`src/test/kotlin/org/wfanet/panelmatch/client/launcher/ExchangeTaskExecutorTest.kt` line 161 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1444#-NpLbtyu5CoGlOVdHuJc:-NqhkXvW5wzcxOH0u0HG:b-iag5v2) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e4dc1508883203e32da8e2875e7c4b7f90421a48/src/test/kotlin/org/wfanet/panelmatch/client/launcher/ExchangeTaskExecutorTest.kt#L161)):*

can you add a test similar to what we're experiencing with Tink where a withContext is used within the task, itself, that then throws an error?
</blockquote></details>

This has been pulled out and is a bit moot now. I've tried to add tests that mimic the tink awskms issue, but have not been able to repro it (at least, not with the current mock structure).

stevenwarejones (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1444)* status: all files reviewed, 2 unresolved discussions (waiting on @SanjayVas)

SanjayVas (changes requested):
> Reviewed 1 of 5 files at r5, 7 of 10 files at r6, 2 of 2 files at r7, 1 of 1 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1444)* status: all files reviewed, 2 unresolved discussions (waiting on @jonmolle)

___
*[`src/main/kotlin/org/wfanet/panelmatch/client/deploy/ExchangeWorkflowFlags.kt` line 125 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1444#-Nqi6mVi--Jb1RCF45GA:-Nqi6mVi--Jb1RCF45GB:bs3jivr) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5e0477ee868caa8cebe9495e82a164f85f82766c/src/main/kotlin/org/wfanet/panelmatch/client/deploy/ExchangeWorkflowFlags.kt#L125)):*
> ```Kotlin
>     private set
> 
>   var maxParallelClaimedExchangeSteps: Int? = null
> ```

Add `private set`
___
*[`src/main/kotlin/org/wfanet/panelmatch/client/deploy/ExchangeWorkflowFlags.kt` line 132 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1444#-Nqi6cl68iwqNLZTPj3H:-Nqi6cl68iwqNLZTPj3I:b-s1s84u) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5e0477ee868caa8cebe9495e82a164f85f82766c/src/main/kotlin/org/wfanet/panelmatch/client/deploy/ExchangeWorkflowFlags.kt#L132)):*
> ```Kotlin
>     required = false,
>   )
>   private fun maxClaimedStepsIsPositive(value: Int?) {
> ```

nit: this name implies this is a function that returns a boolean



_Suggestion:_
```Kotlin
setMaxParallelClaimedSteps
```

SanjayVas (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1444)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @jonmolle)

PR #1448, opened 2024-01-29
add fixed private key for exchange

stevenwarejones:


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1448)

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1448)* status: all files reviewed, 1 unresolved discussion (waiting on @jonmolle and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/panelmatch/client/deploy/ExchangeWorkflowFlags.kt` line 134 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1448#-NpLvqrrDIDvZCdL_gI8:-NpLvqrs8HRwB-02fBwS:b9633ot) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/21dfb31cfdcb3ae8786ad5778aeb97e8c9c9d253/src/main/kotlin/org/wfanet/panelmatch/client/deploy/ExchangeWorkflowFlags.kt#L134)):*
> ```Kotlin
>     required = true,
>   )
>   lateinit var privateKeyPath: String
> ```

Is this a file system path or a blob key? If the former, use `File` instead of `String`. If the latter, call this a blob key rather than a path to avoid confusion.

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1448)* status: all files reviewed, 1 unresolved discussion (waiting on @jonmolle and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/panelmatch/client/deploy/ExchangeWorkflowFlags.kt` line 134 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1448#-NpLvqrrDIDvZCdL_gI8:-NpMN8SyBamZFis8ion0:bcsld49) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/21dfb31cfdcb3ae8786ad5778aeb97e8c9c9d253/src/main/kotlin/org/wfanet/panelmatch/client/deploy/ExchangeWorkflowFlags.kt#L134)):*

Is this a file system path or a blob key? If the former, use `File` instead of `String`. If the latter, call this a blob key rather than a path to avoid confusion.
</blockquote></details>

its a blob key path. Updating.

SanjayVas (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1448)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @jonmolle)

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1448)* status: all files reviewed, 1 unresolved discussion (waiting on @jonmolle)

___
*[`src/main/kotlin/org/wfanet/panelmatch/common/certificates/V2AlphaCertificateManager.kt` line 84 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1448#-Npfx-CUE9chv3sdaWWm:-Npfx-CUE9chv3sdaWWn:b-jnqws) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f285b874968ee6289f0cc8cb8cca10d9e515327e/src/main/kotlin/org/wfanet/panelmatch/common/certificates/V2AlphaCertificateManager.kt#L84)):*
> ```Kotlin
>       if (privateKeyBlobKey !== null && privateKeyBlobKey.isNotEmpty()) privateKeyBlobKey
>       else exchange.path
>     val signingKeys = requireNotNull(getSigningKeys(keyPath)) { "Missing keys for $keyPath" }
> ```

I was thinking about changing the behavior so that even if privateKeyBlogKey is present, it will still check for a key at the exchange.path and only if that is not present, then it will use privateKeyBlobKey. Thoughts? @SanjayVas @jonmolle ?

SanjayVas (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1448)* status: all files reviewed, 1 unresolved discussion (waiting on @jonmolle)

___
*[`src/main/kotlin/org/wfanet/panelmatch/common/certificates/V2AlphaCertificateManager.kt` line 84 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1448#-Npfx-CUE9chv3sdaWWm:-NpfzKqw5MlReW4jV-5A:b-rwxd8k) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f285b874968ee6289f0cc8cb8cca10d9e515327e/src/main/kotlin/org/wfanet/panelmatch/common/certificates/V2AlphaCertificateManager.kt#L84)):*

I was thinking about changing the behavior so that even if privateKeyBlogKey is present, it will still check for a key at the exchange.path and only if that is not present, then it will use privateKeyBlobKey. Thoughts? @SanjayVas @jonmolle ?
</blockquote></details>

No strong opinion other than to make sure the behavior is clearly documented if you do that. Maybe call it "fallbackPrivateKeyBlobKey" to be extra clear?

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1448)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @jonmolle and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/panelmatch/common/certificates/V2AlphaCertificateManager.kt` line 84 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1448#-Npfx-CUE9chv3sdaWWm:-Npu5p_gCJ8A9L0wAmNR:bb7ohn6) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f285b874968ee6289f0cc8cb8cca10d9e515327e/src/main/kotlin/org/wfanet/panelmatch/common/certificates/V2AlphaCertificateManager.kt#L84)):*

No strong opinion other than to make sure the behavior is clearly documented if you do that. Maybe call it "fallbackPrivateKeyBlobKey" to be extra clear?
</blockquote></details>

sgtm

SanjayVas (approved):
> Reviewed 4 of 4 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1448)* status: all files reviewed, 1 unresolved discussion (waiting on @jonmolle and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/panelmatch/common/certificates/V2AlphaCertificateManager.kt` line 83 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1448#-NpuO3g0-acsDgv8LPs_:-NpuO3g0-acsDgv8LPsa:b-u4cqnr) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e0c8c1b09255905e638c11a36b5fc8c42c5f7728/src/main/kotlin/org/wfanet/panelmatch/common/certificates/V2AlphaCertificateManager.kt#L83)):*
> ```Kotlin
>     val keyFromPrimaryPath = getSigningKeys(exchange.path)
>     val signingKeys = if (keyFromPrimaryPath == null) {
>       requireNotNull(getSigningKeys(fallbackPrivateKeyBlobKey!!))
> ```

nit: checkNotNull rather than non-null assertion. The latter is for cases where we *know* it's not null, but the compiler can't figure it out.



_Suggestion:_
```Kotlin
checkNotNull(fallbackPrivateKeyBlobKey)
```

SanjayVas (commented):
> Reviewed 1 of 1 files at r5.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1448)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @jonmolle)

___
*[`src/main/kotlin/org/wfanet/panelmatch/common/certificates/V2AlphaCertificateManager.kt` line 83 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1448#-NpuO3g0-acsDgv8LPs_:-Npuw3CkEWREhLzaZbfW:bmmpd1j) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e0c8c1b09255905e638c11a36b5fc8c42c5f7728/src/main/kotlin/org/wfanet/panelmatch/common/certificates/V2AlphaCertificateManager.kt#L83)):*

nit: checkNotNull rather than non-null assertion. The latter is for cases where we *know* it's not null, but the compiler can't figure it out.
</blockquote></details>

I think you misunderstood. The concern is around the non-null assertion (`!!` operator), not requireNotNull vs. checkNotNull (though that itself can be an issue depending on whether getSigningKeys here returns null based specifically on `exchange`). 

No *code* guarantees that `fallbackPrivateKeyBlobKey` is not null here, therefore a non-null assertion is incorrect. It's instead illegal state - it is an error to call this method unless `fallbackPrivateBlobKey` is correct. It is the caller's responsibility to ensure this in order to avoid the error.

PR #1453, opened 2024-01-30
add error handler to launch for task executor

stevenwarejones:


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1453)

PR #1458, opened 2024-02-01
RequisitionFulfillmentService receives combined shares for HMSS.

renjiezh:


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1458)

SanjayVas (changes requested):
> Reviewed 6 of 18 files at r1, 6 of 8 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1458)* status: 12 of 20 files reviewed, 3 unresolved discussions (waiting on @ple13 and @renjiezh)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/db/computation/ComputationsDatabase.kt` line 200 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1458#-NpakGkVA0k9lMtqoF41:-NpakGkVA0k9lMtqoF42:bffwtj4) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/aed28ba8bb45ff1e449936d504b4c69b8428bf79/src/main/kotlin/org/wfanet/measurement/duchy/db/computation/ComputationsDatabase.kt#L200)):*
> ```Kotlin
> 
>   /** Writes the requisition from a fulfillment. */
>   suspend fun writeRequisition(
> ```

I think this is fine as `writeRequisitionBlobPath` as it *always* does that, even if the seed is sometimes written as well. My concern is making it sound like this method writes requisition data as well, which it does not.

Regardless, I think the KDoc should be clear about the behavior w.r.t. the seed parameter.
___
*[`src/main/proto/wfa/measurement/internal/duchy/computations_service.proto` line 87 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1458#-NpaknpWCD6ZiWZmur87:-NpaknpWCD6ZiWZmur88:btm15jk) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/aed28ba8bb45ff1e449936d504b4c69b8428bf79/src/main/proto/wfa/measurement/internal/duchy/computations_service.proto#L87)):*
> ```proto
> 
>   // Records the requisition containing a blob path and a seed if applicable.
>   rpc RecordRequisition(RecordRequisitionRequest)
> ```

Same note re: coming up with a better name here. I don't believe this is actually recording a Requisition, as I believe the Requisition is already recorded. This is specifically recording information about the fulfillment (note that it's also not recording the fulfillment payload/data itself, as it only records the *blob key* which points to that).

Maybe `RecordRequisitionFulfillment`?
___
*[`src/main/proto/wfa/measurement/internal/duchy/computations_service.proto` line 306 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1458#-NpalYJlA5zC2NDDXy7g:-NpalYJm3bjaau3D2tEh:b-e5o3ca) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/aed28ba8bb45ff1e449936d504b4c69b8428bf79/src/main/proto/wfa/measurement/internal/duchy/computations_service.proto#L306)):*
> ```proto
>   // Required. Must be non-empty.
>   string blob_path = 3;
>   // The encrypted and signed random seed.
> ```

We use sign-then-encrypt, which means this is a ciphertext containing some structure that has the seed and the signature. What is the format of that structure? Also, how does the receiver know which key to decrypt with and which certificate to verify with?

See SignedMessage and EncryptedMessage in the [CMMS API](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/main/src/main/proto/wfa/measurement/api/v2alpha/crypto.proto).



_Suggestion:_
```proto
Ciphertext containing random seed and signature.
```

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1458)* status: 12 of 20 files reviewed, 4 unresolved discussions (waiting on @ple13 and @SanjayVas)

___
*[`src/main/proto/wfa/measurement/internal/duchy/computations_service.proto` line 306 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1458#-NpalYJlA5zC2NDDXy7g:-Npf6EGG76ZfSP91jxi-:b-7s72mq) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/aed28ba8bb45ff1e449936d504b4c69b8428bf79/src/main/proto/wfa/measurement/internal/duchy/computations_service.proto#L306)):*

We use sign-then-encrypt, which means this is a ciphertext containing some structure that has the seed and the signature. What is the format of that structure? Also, how does the receiver know which key to decrypt with and which certificate to verify with?

See SignedMessage and EncryptedMessage in the [CMMS API](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/main/src/main/proto/wfa/measurement/api/v2alpha/crypto.proto).
</blockquote></details>

The encryption key pair is created by the duchy during creating the Computation (using tink). The duchy keeps the private key in the database and sends the public key to the EDP through the Kingdom. Each Measurement has a new pair.

For signatures, new private key and certificate are created by the EDP and accessed via config. Duchies add the certificate by config as well. All Measurements are using the same pair.
___
*[`src/main/proto/wfa/measurement/internal/duchy/computations_service.proto` line 307 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1458#-Npf6Cva-n6XM45-UCli:-NpfAgE110xTrXYKAQWU:b-5enjf7) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/aed28ba8bb45ff1e449936d504b4c69b8428bf79/src/main/proto/wfa/measurement/internal/duchy/computations_service.proto#L307)):*

Is there a good reason to handle the seed differently from the data? The requisition can be considered as (data  + encrypted(seed + signature)) and will be sent to only 1 duchy. The whole thing can be written to the blob.
</blockquote></details>

Good question.
A duchy needs to forward the secret seed to the peer worker. So eventually we need to separate them. 
Having them separated internally makes it easier for following operations to get only seeds or data. Note that, the size of data is large, so it is good to save the cost of duplicated serialization and parsing.

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1458)* status: 12 of 20 files reviewed, 4 unresolved discussions (waiting on @ple13 and @renjiezh)

___
*[`src/main/proto/wfa/measurement/internal/duchy/computations_service.proto` line 306 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1458#-NpalYJlA5zC2NDDXy7g:-NpfYm1Q83NT_nsqXjrw:b-6b0qnr) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/aed28ba8bb45ff1e449936d504b4c69b8428bf79/src/main/proto/wfa/measurement/internal/duchy/computations_service.proto#L306)):*

The encryption key pair is created by the duchy during creating the Computation (using tink). The duchy keeps the private key in the database and sends the public key to the EDP through the Kingdom. Each Measurement has a new pair.

For signatures, new private key and certificate are created by the EDP and accessed via config. Duchies add the certificate by config as well. All Measurements are using the same pair.
</blockquote></details>

What about the structure for the encrypted payload which holds both the seed and the signature?

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1458)* status: 12 of 20 files reviewed, 4 unresolved discussions (waiting on @ple13 and @SanjayVas)

___
*[`src/main/proto/wfa/measurement/internal/duchy/computations_service.proto` line 306 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1458#-NpalYJlA5zC2NDDXy7g:-NpfZdrs0cdwBXkwdmlb:b-xbfg9o) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/aed28ba8bb45ff1e449936d504b4c69b8428bf79/src/main/proto/wfa/measurement/internal/duchy/computations_service.proto#L306)):*

What about the structure for the encrypted payload which holds both the seed and the signature?
</blockquote></details>

I was thinking to use serialized `EncryptedMessage` containing `SignedMessage`. But seems I can use them explicitly here and only serialize when store in database.

ple13 (changes requested):
> Reviewed 12 of 18 files at r1, 6 of 8 files at r2.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1458)* status: all files reviewed, 8 unresolved discussions (waiting on @renjiezh and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/writers/RecordRequisitionData.kt` line 48 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1458#-NpfZzwz8_6ZrA_iHftw:-NpfZzwz8_6ZrA_iHftx:b-n6bxar) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/aed28ba8bb45ff1e449936d504b4c69b8428bf79/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/writers/RecordRequisitionData.kt#L48)):*
> ```Kotlin
>     require(pathToBlob.isNotBlank()) { "Cannot insert blank path to blob. $externalRequisitionKey" }
>     if (seed != null) {
>       require(pathToBlob.isNotBlank()) {
> ```

Is this a typo? We checked for the path earlier.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/gcloud/spanner/computation/GcpSpannerComputationsDatabaseTransactor.kt` line 659 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1458#-NpfcLLxEmy_bxd5g80g:-NpfcLLyEYi_M44Wx8m8:b4oz6bf) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/aed28ba8bb45ff1e449936d504b4c69b8428bf79/src/main/kotlin/org/wfanet/measurement/duchy/deploy/gcloud/spanner/computation/GcpSpannerComputationsDatabaseTransactor.kt#L659)):*
> ```Kotlin
>     token: ComputationEditToken<ProtocolT, StageT>,
>     externalRequisitionKey: ExternalRequisitionKey,
>     pathToBlob: String,
> ```

I noticed that we checked pathToBlob != empty in writeRequisition function. However, I think it's better to check it here as well (if writeRequisitionData is called directly, no safeguard check is done). If it's not intended for this function to be called, let's add a comment to say that?
___
*[`src/test/kotlin/org/wfanet/measurement/duchy/service/api/v2alpha/RequisitionFulfillmentServiceTest.kt` line 205 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1458#-NpflUIM27ZYTmccy48i:-NpflUIM27ZYTmccy48j:b-w3xtzr) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/aed28ba8bb45ff1e449936d504b4c69b8428bf79/src/test/kotlin/org/wfanet/measurement/duchy/service/api/v2alpha/RequisitionFulfillmentServiceTest.kt#L205)):*
> ```Kotlin
>   }
> 
>   @Test
> ```

Let's add a test where the seed is empty and verify that HMSS fulfill requisition fails.
___
*[`src/test/kotlin/org/wfanet/measurement/duchy/service/internal/computationcontrol/AsyncComputationControlServiceTest.kt` line 438 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1458#-Npfm_XOEB6aba3uSEvP:-Npfm_XOEB6aba3uSEvQ:baxxe7y) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/aed28ba8bb45ff1e449936d504b4c69b8428bf79/src/test/kotlin/org/wfanet/measurement/duchy/service/internal/computationcontrol/AsyncComputationControlServiceTest.kt#L438)):*
> ```Kotlin
>         }
>       } else {
>         requisitions += requisitionMetadata {}
> ```

This doesn't cover the case where path exists but seed is empty yet.
___
*[`src/test/kotlin/org/wfanet/measurement/duchy/service/internal/computations/ComputationsServiceTest.kt` line 523 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1458#-NpfnHR5CDJyp7tY16px:-NpfnHR5CDJyp7tY16py:b5a6144) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/aed28ba8bb45ff1e449936d504b4c69b8428bf79/src/test/kotlin/org/wfanet/measurement/duchy/service/internal/computations/ComputationsServiceTest.kt#L523)):*
> ```Kotlin
>     fakeDatabase.addComputation(
>       globalId = id,
>       stage = LiquidLegionsSketchAggregationV2.Stage.EXECUTION_PHASE_ONE.toProtocolStage(),
> ```

Shouldn't this be HMSS stage?

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1458)* status: all files reviewed, 8 unresolved discussions (waiting on @renjiezh)

___
*[`src/main/proto/wfa/measurement/internal/duchy/computations_service.proto` line 306 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1458#-NpalYJlA5zC2NDDXy7g:-NpfwyFB7WVAlfB3wl8b:bolh20l) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/aed28ba8bb45ff1e449936d504b4c69b8428bf79/src/main/proto/wfa/measurement/internal/duchy/computations_service.proto#L306)):*
> seems I can use them explicitly here and only serialize when store in database.

Not sure what you mean. If using this format, it already be coming in serialized that way from the EDP calling the CMMS API. You just need to document it if you're persisting that serialized message here so we know how to deserialize/decrypt/verify it. This also means you'd need to store additional information about which version of the message this is (e.g. persisting the public API version somewhere).

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1458)* status: 7 of 21 files reviewed, 8 unresolved discussions (waiting on @ple13 and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/db/computation/ComputationsDatabase.kt` line 200 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1458#-NpakGkVA0k9lMtqoF41:-NprWZwT5PhDFAwnNmmK:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/aed28ba8bb45ff1e449936d504b4c69b8428bf79/src/main/kotlin/org/wfanet/measurement/duchy/db/computation/ComputationsDatabase.kt#L200)):*

I think this is fine as `writeRequisitionBlobPath` as it *always* does that, even if the seed is sometimes written as well. My concern is making it sound like this method writes requisition data as well, which it does not.

Regardless, I think the KDoc should be clear about the behavior w.r.t. the seed parameter.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/writers/RecordRequisitionData.kt` line 48 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1458#-NpfZzwz8_6ZrA_iHftw:-NpqwSAZ-e9mMWkcCUMT:b6rs0km) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/aed28ba8bb45ff1e449936d504b4c69b8428bf79/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/writers/RecordRequisitionData.kt#L48)):*

Is this a typo? We checked for the path earlier.
</blockquote></details>

Done.
Nice catch!
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/gcloud/spanner/computation/GcpSpannerComputationsDatabaseTransactor.kt` line 659 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1458#-NpfcLLxEmy_bxd5g80g:-NpqxBozBmVcdplREMbB:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/aed28ba8bb45ff1e449936d504b4c69b8428bf79/src/main/kotlin/org/wfanet/measurement/duchy/deploy/gcloud/spanner/computation/GcpSpannerComputationsDatabaseTransactor.kt#L659)):*

I noticed that we checked pathToBlob != empty in writeRequisition function. However, I think it's better to check it here as well (if writeRequisitionData is called directly, no safeguard check is done). If it's not intended for this function to be called, let's add a comment to say that?
</blockquote></details>

Done.
___
*[`src/main/proto/wfa/measurement/internal/duchy/computations_service.proto` line 87 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1458#-NpaknpWCD6ZiWZmur87:-Nprna5-DfTcR1QdQrrN:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/aed28ba8bb45ff1e449936d504b4c69b8428bf79/src/main/proto/wfa/measurement/internal/duchy/computations_service.proto#L87)):*

Same note re: coming up with a better name here. I don't believe this is actually recording a Requisition, as I believe the Requisition is already recorded. This is specifically recording information about the fulfillment (note that it's also not recording the fulfillment payload/data itself, as it only records the *blob key* which points to that).

Maybe `RecordRequisitionFulfillment`?
</blockquote></details>

Done.
___
*[`src/main/proto/wfa/measurement/internal/duchy/computations_service.proto` line 306 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1458#-NpalYJlA5zC2NDDXy7g:-NprbBND0taVSD9SO7kf:bgfvb8d) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/aed28ba8bb45ff1e449936d504b4c69b8428bf79/src/main/proto/wfa/measurement/internal/duchy/computations_service.proto#L306)):*

> seems I can use them explicitly here and only serialize when store in database.

Not sure what you mean. If using this format, it already be coming in serialized that way from the EDP calling the CMMS API. You just need to document it if you're persisting that serialized message here so we know how to deserialize/decrypt/verify it. This also means you'd need to store additional information about which version of the message this is (e.g. persisting the public API version somewhere).
</blockquote></details>

Done.
Updated the comments. 
___
*[`src/test/kotlin/org/wfanet/measurement/duchy/service/api/v2alpha/RequisitionFulfillmentServiceTest.kt` line 205 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1458#-NpflUIM27ZYTmccy48i:-NprnayQCqs-W-0zsP4E:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/aed28ba8bb45ff1e449936d504b4c69b8428bf79/src/test/kotlin/org/wfanet/measurement/duchy/service/api/v2alpha/RequisitionFulfillmentServiceTest.kt#L205)):*

Let's add a test where the seed is empty and verify that HMSS fulfill requisition fails.
</blockquote></details>

Done.
___
*[`src/test/kotlin/org/wfanet/measurement/duchy/service/internal/computationcontrol/AsyncComputationControlServiceTest.kt` line 438 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1458#-Npfm_XOEB6aba3uSEvP:-NprmQDKEgaWrlTKnn_o:b1y5ytk) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/aed28ba8bb45ff1e449936d504b4c69b8428bf79/src/test/kotlin/org/wfanet/measurement/duchy/service/internal/computationcontrol/AsyncComputationControlServiceTest.kt#L438)):*

This doesn't cover the case where path exists but seed is empty yet.
</blockquote></details>

Thanks for the reminder.
This is for control service that will be the topic of next PR. I will take care of it there.
___
*[`src/test/kotlin/org/wfanet/measurement/duchy/service/internal/computations/ComputationsServiceTest.kt` line 523 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1458#-NpfnHR5CDJyp7tY16px:-Nprf-jF44IIdLte6C0b:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/aed28ba8bb45ff1e449936d504b4c69b8428bf79/src/test/kotlin/org/wfanet/measurement/duchy/service/internal/computations/ComputationsServiceTest.kt#L523)):*

Shouldn't this be HMSS stage?
</blockquote></details>

Done.

SanjayVas (changes requested):
> Reviewed 12 of 15 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1458)* status: 18 of 21 files reviewed, 6 unresolved discussions (waiting on @ple13 and @renjiezh)

___
*[`src/main/proto/wfa/measurement/internal/duchy/computations_service.proto` line 308 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1458#-NpuZ7CC50pPzoxFbmUM:-NpuZ7CC50pPzoxFbmUN:b-hs9653) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/da6969d304612fdf84b8494ac6aba2cfeaaf9bcc/src/main/proto/wfa/measurement/internal/duchy/computations_service.proto#L308)):*
> ```proto
>   // The encrypted and signed random seed.
>   //
>   // In format of encrypted serialized SignedMessage.
> ```

`SignedMessage` from the CMMS public API is specifically for a signed protobuf message. The seed is not a protobuf message, so I believe you'll need to find a different structure.

The standard structure would be Cryptographic Message Syntax (CMS), the successor to PKCS #&#8203;7. Unfortunately we don't yet have a JVM implementation of that spec. The only library that I know of that handles it is part of BouncyCastle, so we'd have to be okay taking a dependency on BouncyCastle's bcpkix.



_Code quote:_
```proto
SignedMessage
```

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1458)* status: 18 of 21 files reviewed, 6 unresolved discussions (waiting on @ple13 and @SanjayVas)

___
*[`src/main/proto/wfa/measurement/internal/duchy/computations_service.proto` line 308 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1458#-NpuZ7CC50pPzoxFbmUM:-NpuebWTCRWcW1jpZJU6:b-na9d0h) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/da6969d304612fdf84b8494ac6aba2cfeaaf9bcc/src/main/proto/wfa/measurement/internal/duchy/computations_service.proto#L308)):*

`SignedMessage` from the CMMS public API is specifically for a signed protobuf message. The seed is not a protobuf message, so I believe you'll need to find a different structure.

The standard structure would be Cryptographic Message Syntax (CMS), the successor to PKCS #&#8203;7. Unfortunately we don't yet have a JVM implementation of that spec. The only library that I know of that handles it is part of BouncyCastle, so we'd have to be okay taking a dependency on BouncyCastle's bcpkix.
</blockquote></details>

This is a very important point. Thanks for it. An Easier way is to wrap the bytes:
```
// cross-media-measurement-api/random_seed.proto
Message RandomSeed {
   bytes data
}
```
Then we have `bytes secret_seed`indicating it is in encrypted and signed format. The comment is like `An encrypted serialized SignedMessage with data of RandomSeed`
How do you feel?

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1458)* status: 18 of 21 files reviewed, 6 unresolved discussions (waiting on @ple13 and @renjiezh)

___
*[`src/main/proto/wfa/measurement/internal/duchy/computations_service.proto` line 308 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1458#-NpuZ7CC50pPzoxFbmUM:-NpuiBqk0kpSghas6lS2:bey2pd5) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/da6969d304612fdf84b8494ac6aba2cfeaaf9bcc/src/main/proto/wfa/measurement/internal/duchy/computations_service.proto#L308)):*

This is a very important point. Thanks for it. An Easier way is to wrap the bytes:
```
// cross-media-measurement-api/random_seed.proto
Message RandomSeed {
   bytes data
}
```
Then we have `bytes secret_seed`indicating it is in encrypted and signed format. The comment is like `An encrypted serialized SignedMessage with data of RandomSeed`
How do you feel?
</blockquote></details>

That would work. We'd still need to

1. Document that `SignedMessage` is from the CMMS public API.
2. Make sure we're capturing the version of the `SignedMessage` type somewhere. We do this elsewhere by persisting a `public_api_version` field.

That said if we're willing to take on the bcpkix dependency I think using the CMS standard would be better to have one less place where we have to worry about cross-API message serialization.

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1458)* status: 3 of 34 files reviewed, 6 unresolved discussions (waiting on @ple13 and @SanjayVas)

___
*[`src/main/proto/wfa/measurement/internal/duchy/computations_service.proto` line 308 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1458#-NpuZ7CC50pPzoxFbmUM:-Npx8Nay8Kduu_g6X2WD:b2ru3xa) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/da6969d304612fdf84b8494ac6aba2cfeaaf9bcc/src/main/proto/wfa/measurement/internal/duchy/computations_service.proto#L308)):*

That would work. We'd still need to

1. Document that `SignedMessage` is from the CMMS public API.
2. Make sure we're capturing the version of the `SignedMessage` type somewhere. We do this elsewhere by persisting a `public_api_version` field.

That said if we're willing to take on the bcpkix dependency I think using the CMS standard would be better to have one less place where we have to worry about cross-API message serialization.
</blockquote></details>

Added `PublicApiVersion` into `RequisitionMetadata`. Renamed `seed` into `secret_seed` because it is not a real plain seed.

ple13 (approved):
> Reviewed 14 of 15 files at r3, 31 of 31 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1458)* status: all files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

SanjayVas (changes requested):
> Reviewed 13 of 31 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1458)* status: all files reviewed, 3 unresolved discussions (waiting on @renjiezh)

___
*[`MODULE.bazel` line 47 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1458#-Nq01kTp-wxPf2oWU7CD:-Nq01kTp-wxPf2oWU7CE:b-k73wmy) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/808e745c967e22ff12de1bc6aa1d6765c999268d/MODULE.bazel#L47)):*
> ```Starlark
>     name = "rules_go",
>     repo_name = "io_bazel_rules_go",
>     version = "0.43.0",
> ```

What caused these formatting changes? The existing formatting was output by the latest version of `buildifier`, which at time of writing is [6.4.0](https://github.com/bazelbuild/buildtools/releases/tag/v6.4.0).
___
*[`src/main/proto/wfa/measurement/internal/duchy/computations_service.proto` line 311 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1458#-Nq03kKl1eXsc02TmNss:-Nq03kKl1eXsc02TmNst:b-kkf7sj) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/808e745c967e22ff12de1bc6aa1d6765c999268d/src/main/proto/wfa/measurement/internal/duchy/computations_service.proto#L311)):*
> ```proto
>   // RandomSeed is defined in CMMS public api. The RandomSeed is firstly signed
>   // as a SignedMessage then encrypted into a EncryptedMessage.
>   .wfa.measurement.api.v2alpha.EncryptedMessage secret_seed = 4;
> ```

The internal Duchy API **must not** directly import message types from the CMMS public API. Instead this should just be the ciphertext bytes. The field documentation and public API version should be enough to reconstruct the type URL.



_Suggestion:_
```proto
bytes secret_seed_ciphertext = 4;
```
___
*[`src/main/resources/duchy/spanner/add-requisition-public-api-version.sql` line 24 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1458#-Nq031U7E7sgVK7rVKML:-Nq031U7E7sgVK7rVKMM:bg92dg6) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/808e745c967e22ff12de1bc6aa1d6765c999268d/src/main/resources/duchy/spanner/add-requisition-public-api-version.sql#L24)):*
> ```TSQL
> ALTER TABLE Requisitions
>     -- The public api version of the requisition fulfillment.
>     ADD COLUMN PublicApiVersion STRING(MAX);
> ```

Why do we need a new column? Can this not just go into RequisitionDetails?

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1458)* status: 10 of 36 files reviewed, 3 unresolved discussions (waiting on @ple13 and @SanjayVas)

___
*[`MODULE.bazel` line 47 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1458#-Nq01kTp-wxPf2oWU7CD:-Nq44YTJ35tmsyoH0CGf:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/808e745c967e22ff12de1bc6aa1d6765c999268d/MODULE.bazel#L47)):*

What caused these formatting changes? The existing formatting was output by the latest version of `buildifier`, which at time of writing is [6.4.0](https://github.com/bazelbuild/buildtools/releases/tag/v6.4.0).
</blockquote></details>

Done.
___
*[`src/main/proto/wfa/measurement/internal/duchy/computations_service.proto` line 311 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1458#-Nq03kKl1eXsc02TmNss:-Nq44ZS1CJo21Dym7L3D:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/808e745c967e22ff12de1bc6aa1d6765c999268d/src/main/proto/wfa/measurement/internal/duchy/computations_service.proto#L311)):*

The internal Duchy API **must not** directly import message types from the CMMS public API. Instead this should just be the ciphertext bytes. The field documentation and public API version should be enough to reconstruct the type URL.
</blockquote></details>

Done.
___
*[`src/main/resources/duchy/spanner/add-requisition-public-api-version.sql` line 24 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1458#-Nq031U7E7sgVK7rVKML:-Nq44_eC9MPyKhTGj8Hx:b9cf5bs) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/808e745c967e22ff12de1bc6aa1d6765c999268d/src/main/resources/duchy/spanner/add-requisition-public-api-version.sql#L24)):*

Why do we need a new column? Can this not just go into RequisitionDetails?
</blockquote></details>

Done.
Moved into RequisitionDetails.

SanjayVas (changes requested):
> Reviewed 1 of 8 files at r2, 6 of 31 files at r4, 25 of 26 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1458)* status: 35 of 36 files reviewed, 4 unresolved discussions (waiting on @ple13 and @renjiezh)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/service/api/v2alpha/RequisitionFulfillmentService.kt` line 103 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1458#-Nq4x656AAUQf-ZXsV9B:-Nq4x656AAUQf-ZXsV9C:b-gabzt7) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/09e1af7996e325efdff224b1fdce8d7c429ba845/src/main/kotlin/org/wfanet/measurement/duchy/service/api/v2alpha/RequisitionFulfillmentService.kt#L103)):*
> ```Kotlin
>               grpcRequire(header.honestMajorityShareShuffle.hasSecretSeed()) {
>                 "Secret seed not be specified for HMSS protocol."
>               }
> ```

Also verify that the type URL matches that of a v2alpha `SignedMessage`.
___
*[`src/main/proto/wfa/measurement/internal/duchy/computation_token.proto` line 102 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1458#-Nq4vRQ_5yJOGE0SxRVe:-Nq4vRQ_5yJOGE0SxRVf:b1n6uyp) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/09e1af7996e325efdff224b1fdce8d7c429ba845/src/main/proto/wfa/measurement/internal/duchy/computation_token.proto#L102)):*
> ```proto
>   // The requisition data in format of a secret seed. It is an serialized
>   // EncryptedMessage with data of SignedMessage.
>   bytes secret_seed = 4;
> ```

Note which API this comes from. Additionally this should just be the ciphertext from `EncryptedMessage`, not the `EncryptedMessage` itself.



_Suggestion:_
```proto
  // The requisition data in format of a secret seed. It is a ciphertext containing a serialized `SignedMessage` message from the CMMS public API.
  bytes secret_seed_ciphertext = 4;
```
___
*[`src/main/proto/wfa/measurement/internal/duchy/computations_service.proto` line 309 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1458#-Nq4w1CD2LqAdo960l_O:-Nq4w1CE8qB4Azc4Knr-:b16tiea) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/09e1af7996e325efdff224b1fdce8d7c429ba845/src/main/proto/wfa/measurement/internal/duchy/computations_service.proto#L309)):*
> ```proto
>   //
>   // RandomSeed is defined in CMMS public api. The RandomSeed is firstly signed
>   // as a SignedMessage then encrypted into a EncryptedMessage.
> ```

Same note here. This is just the ciphertext.



_Suggestion:_
```proto
  // This is a ciphertext containing a serialized `SignedMessage` from the public API.
```
___
*[`src/test/kotlin/org/wfanet/measurement/duchy/service/api/v2alpha/RequisitionFulfillmentServiceTest.kt` line 77 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1458#-Nq4xXpH9aX_iPlb5h7R:-Nq4xXpH9aX_iPlb5h7S:b7e2480) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/09e1af7996e325efdff224b1fdce8d7c429ba845/src/test/kotlin/org/wfanet/measurement/duchy/service/api/v2alpha/RequisitionFulfillmentServiceTest.kt#L77)):*
From our [Code Style](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/main/docs/code-style.md):

> -   Avoid defining extensions on common types when it may appear that they could work on all instances of that type.
    -   For example, `String.toFoo()` where the String must be a specific serialization of a `Foo` to work. This is version of the "stringly typed" anti-pattern.

Perhaps instead

```kotlin
private fun buildSecretSeed(data: String): EncryptedMessage
```

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1458)* status: 22 of 36 files reviewed, 4 unresolved discussions (waiting on @ple13 and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/service/api/v2alpha/RequisitionFulfillmentService.kt` line 103 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1458#-Nq4x656AAUQf-ZXsV9B:-Nq58NEC80ryMmSXIBv9:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/09e1af7996e325efdff224b1fdce8d7c429ba845/src/main/kotlin/org/wfanet/measurement/duchy/service/api/v2alpha/RequisitionFulfillmentService.kt#L103)):*

Also verify that the type URL matches that of a v2alpha `SignedMessage`.
</blockquote></details>

Done.
___
*[`src/main/proto/wfa/measurement/internal/duchy/computation_token.proto` line 102 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1458#-Nq4vRQ_5yJOGE0SxRVe:-Nq58P7g012246BEIuJG:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/09e1af7996e325efdff224b1fdce8d7c429ba845/src/main/proto/wfa/measurement/internal/duchy/computation_token.proto#L102)):*

Note which API this comes from. Additionally this should just be the ciphertext from `EncryptedMessage`, not the `EncryptedMessage` itself.
</blockquote></details>

Done.
___
*[`src/main/proto/wfa/measurement/internal/duchy/computations_service.proto` line 309 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1458#-Nq4w1CD2LqAdo960l_O:-Nq58Q8bAeH1lGeY6z7U:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/09e1af7996e325efdff224b1fdce8d7c429ba845/src/main/proto/wfa/measurement/internal/duchy/computations_service.proto#L309)):*

Same note here. This is just the ciphertext.
</blockquote></details>

Done.
___
*[`src/test/kotlin/org/wfanet/measurement/duchy/service/api/v2alpha/RequisitionFulfillmentServiceTest.kt` line 77 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1458#-Nq4xXpH9aX_iPlb5h7R:-Nq58R2v73FQiWpvX7tQ:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/09e1af7996e325efdff224b1fdce8d7c429ba845/src/test/kotlin/org/wfanet/measurement/duchy/service/api/v2alpha/RequisitionFulfillmentServiceTest.kt#L77)):*

From our [Code Style](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/main/docs/code-style.md):

> -   Avoid defining extensions on common types when it may appear that they could work on all instances of that type.
    -   For example, `String.toFoo()` where the String must be a specific serialization of a `Foo` to work. This is version of the "stringly typed" anti-pattern.

Perhaps instead

```kotlin
private fun buildSecretSeed(data: String): EncryptedMessage
```
</blockquote></details>

Done.

SanjayVas (approved):
> Reviewed 1 of 26 files at r5, 14 of 14 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1458)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @ple13)

stevenwarejones (changes requested):
> Reviewed 1 of 15 files at r3, 1 of 26 files at r5, 1 of 14 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1458)* status: all files reviewed, 1 unresolved discussion (waiting on @ple13 and @renjiezh)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/db/computation/testing/FakeComputationsDatabase.kt` line 324 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1458#-Nq9xAHj-j1ljkkGdAfM:-Nq9xAHj-j1ljkkGdAfN:b4p2p2v) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6840319ab9e722ba6a8b88a540f257ce532e8b14/src/main/kotlin/org/wfanet/measurement/duchy/db/computation/testing/FakeComputationsDatabase.kt#L324)):*
> ```Kotlin
>     externalRequisitionKey: ExternalRequisitionKey,
>     pathToBlob: String,
>     secretSeedCiphertext: ByteString?,
> ```

under what circumstances would this be null?

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1458)* status: all files reviewed, 1 unresolved discussion (waiting on @ple13 and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/db/computation/testing/FakeComputationsDatabase.kt` line 324 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1458#-Nq9xAHj-j1ljkkGdAfM:-Nq9zjbl25bAHm0aWQOG:b9mzy8v) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6840319ab9e722ba6a8b88a540f257ce532e8b14/src/main/kotlin/org/wfanet/measurement/duchy/db/computation/testing/FakeComputationsDatabase.kt#L324)):*

under what circumstances would this be null?
</blockquote></details>

These functions are used for both llv2 and hmss protocols. For llv2, the secret seed is null here.

stevenwarejones (changes requested):
> Reviewed 1 of 15 files at r3, 2 of 31 files at r4, 9 of 26 files at r5, 8 of 14 files at r6.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1458)* status: all files reviewed, 2 unresolved discussions (waiting on @ple13 and @renjiezh)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/PostgresComputationsService.kt` line 446 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1458#-NqD-E-lALwY7pXfZnUf:-NqD-E-lALwY7pXfZnUg:b2vjtdf) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6840319ab9e722ba6a8b88a540f257ce532e8b14/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/PostgresComputationsService.kt#L446)):*
> ```Kotlin
>           externalRequisitionKey = request.key,
>           pathToBlob = request.blobPath,
>           seed = if (!request.secretSeedCiphertext.isEmpty) request.secretSeedCiphertext else null,
> ```

should this field get renamed to reflect that it is a signed cipher text? `seed` by itself seems slightly misleading.
___
*[`src/main/proto/wfa/measurement/internal/duchy/computation_token.proto` line 102 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1458#-NqCzbes5Gh9FtHz59dz:-NqCzbes5Gh9FtHz59e-:bh33sia) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6840319ab9e722ba6a8b88a540f257ce532e8b14/src/main/proto/wfa/measurement/internal/duchy/computation_token.proto#L102)):*
> ```proto
>   // The requisition data in format of a secret seed. It is ciphertext
>   // containing a serialized `SignedMessage` message from the CMMS public API.
>   bytes secret_seed_ciphertext = 4;
> ```

is it too verbose to call this `secret_seed_signed_ciphertext`?

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1458)* status: 30 of 36 files reviewed, 2 unresolved discussions (waiting on @ple13, @SanjayVas, and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/PostgresComputationsService.kt` line 446 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1458#-NqD-E-lALwY7pXfZnUf:-NqE7tHv4mINIZrWDIDd:b-we8vcv) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6840319ab9e722ba6a8b88a540f257ce532e8b14/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/postgres/PostgresComputationsService.kt#L446)):*

should this field get renamed to reflect that it is a signed cipher text? `seed` by itself seems slightly misleading.
</blockquote></details>

Done. Renamed.
___
*[`src/main/proto/wfa/measurement/internal/duchy/computation_token.proto` line 102 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1458#-NqCzbes5Gh9FtHz59dz:-NqE7vHTBlcP2VvnAhv6:blx24c1) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6840319ab9e722ba6a8b88a540f257ce532e8b14/src/main/proto/wfa/measurement/internal/duchy/computation_token.proto#L102)):*

is it too verbose to call this `secret_seed_signed_ciphertext`?
</blockquote></details>

"secret_seed" implies the seed is signed and encrypted. `ciphertext` means this is serialized bytes. `secret_seed_ciphertext` should be accurate for the purpose.

ple13 (approved):
> Reviewed 9 of 26 files at r5, 14 of 14 files at r6, 6 of 6 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1458)* status: all files reviewed, 2 unresolved discussions (waiting on @stevenwarejones)

stevenwarejones (approved):
> Reviewed 1 of 8 files at r2, 2 of 31 files at r4, 3 of 26 files at r5, 5 of 14 files at r6, 6 of 6 files at r7, 2 of 2 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1458)* status: all files reviewed, 1 unresolved discussion (waiting on @renjiezh)

PR #1459, opened 2024-02-02
Updating Charts and fixing some Chart Bugs

bdomen-ggl:
Notes:
* Chart features:
  * Adding a new chart type (multi-bar not percentage).
  * Updating existing draw bar function (was removed from use so last PR didn't break anything).
  * Adding a legend.
  * Updating chart titles.
* Chart bugs:
  * Setting a max height on the charts so they can always be viewable in a screen.
  * Adding a chart redraw method and a window resize event. If the window size changed, the charts wouldn't resize. D3 has a responsive SVG attribute but that doesn't apply to fonts.
  * Adding a style so the charts can be scrollable on small screens.
  * Updating tick count so it makes more sense with fewer x-axes values.

### DIFF
--- a/experimental/reporting-ui/src/main/react/reporting-ui/component/chart/chart.tsx
+++ b/experimental/reporting-ui/src/main/react/reporting-ui/component/chart/chart.tsx
@@ -17,12 +17,13 @@ import Card from 'react-bootstrap/Card';
 import Row from 'react-bootstrap/Row';
 import Col from 'react-bootstrap/Col';
 import { FilterChartIcon, OptionsIcon } from '../../public/asset/icon';
-import { createMultiLineChart, createPercentBarChart, createPercentMultiLineChart } from './d3_wrapper';
+import { createMultiLineChart, createBarChart, createPercentBarChart, createPercentMultiLineChart, removeGraph } from './d3_wrapper';
 
 export enum ChartType {
   percentMultiLine,
   multiLine,
   bar,
+  barPercent,
 }
 
 type props = {
@@ -56,8 +57,10 @@ export function Chart({cardId, title, data, config, type}: props) {
       createMultiLineChart(cardId, data, dimensions, margins, config.pubColors)
     } else if (type === ChartType.percentMultiLine) {
       createPercentMultiLineChart(cardId, data, dimensions, margins, config.catColors)
+    } else if (type === ChartType.barPercent) {
+      createPercentBarChart(cardId, data, dimensions, margins, config.pubColors)
     } else if (type === ChartType.bar) {
-      createPercentBarChart(cardId, data, dimensions, margins)
+      createBarChart(cardId, data, dimensions, margins, config.pubColors)
     }
   }
 
@@ -73,9 +76,29 @@ export function Chart({cardId, title, data, config, type}: props) {
     if (refContainer.current) {
       setDimensions({
         width: refContainer.current.offsetWidth,
-        height: refContainer.current.offsetWidth * 0.6,
+        height: Math.min(refContainer.current.offsetWidth * 0.6, 300),
       });
     }
+
+    function handleResize() {
+      // Delete and re-create the whole charts.
+      // Even when using 'responsive svg', the fonts don't change.
+      removeGraph(cardId);
+      if (refContainer.current) {
+        setDimensions({
+          width: refContainer.current.offsetWidth,
+          height: Math.min(refContainer.current.offsetWidth * 0.6, 300),
+        });
+      }
+    }
+
+    // Attach the event listener to the window object
+    window.addEventListener('resize', handleResize);
+
+    // Remove the event listener when the component unmounts
+    return () => {
+      window.removeEventListener('resize', handleResize);
+    };
   }, []);
 
   return (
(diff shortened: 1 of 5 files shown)


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1459)

bdomen-ggl (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1459)* status: 2 of 5 files reviewed, 2 unresolved discussions (waiting on @SanjayVas)

___
*[`experimental/reporting-ui/src/main/react/reporting-ui/component/chart/chart.tsx` line 90 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1459#-NpulM9e8xoFWz2s0PK-:-Nq3hp0xErGAQzOQhWKt:bjdonw6) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f8282e784289c28584404b063c8387eb8f6499a7/experimental/reporting-ui/src/main/react/reporting-ui/component/chart/chart.tsx#L90)):*

Can we not use the `min-width` CSS property rather than computing this in TS?
</blockquote></details>

Had to make quite a few changes, but yes we can.

It's not trivial because I need to pass dimensions to D3 so I can do some calculations on where to draw things. D3 also creates elements on an SVG that it also creates. There's only a top level blank element that I add that D3 injects charts into. Since the canvas doesn't exist at first, the dimensions are some-width and 0 height. I updated the CSS to set an aspect ratio if the parent is less than a certain size (so the height will be based on the width), and use a max height if the parent is greater than the certain size.

I similarly moved the hardcoded legend dimensions to CSS. There were some weird issues if I moved the other styles so I'll at those later.
___
*[`experimental/reporting-ui/src/main/react/reporting-ui/component/chart/d3_wrapper.css` line 26 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1459#-NpulSyq633hc8nP9Qvc:-Nq3hVty7huunl3qT7_M:bb3xu7) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f8282e784289c28584404b063c8387eb8f6499a7/experimental/reporting-ui/src/main/react/reporting-ui/component/chart/d3_wrapper.css#L26)):*

I don't undestand this comment.
</blockquote></details>

Oh I can delete that. I copied this and the following blocks from some D3 examples. Yeah not sure what it means.

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1459)* status: all files reviewed, 1 unresolved discussion (waiting on @bdomen-ggl)

___
*[`experimental/reporting-ui/src/main/react/reporting-ui/component/chart/d3_wrapper.css` line 17 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1459#-NqEDaTc4jORGHET9G83:-NqEDaTc4jORGHET9G84:b-s77b0) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/00e54683220d1d4a885fe2db9f452c5db5dcdf4e/experimental/reporting-ui/src/main/react/reporting-ui/component/chart/d3_wrapper.css#L17)):*
> ```CSS
> .container {
>     container-type: inline-size;
>     margin: 0 !important;
> ```

`!important` is a hack that should never be relied upon. If you can't do this without `!important`, then go back to setting it in script rather than stylesheet.

bdomen-ggl (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1459)* status: 3 of 5 files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

___
*[`experimental/reporting-ui/src/main/react/reporting-ui/component/chart/d3_wrapper.css` line 17 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1459#-NqEDaTc4jORGHET9G83:-NqEJPKn1Fhtqr9j-pAv:bcddylc) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/00e54683220d1d4a885fe2db9f452c5db5dcdf4e/experimental/reporting-ui/src/main/react/reporting-ui/component/chart/d3_wrapper.css#L17)):*

`!important` is a hack that should never be relied upon. If you can't do this without `!important`, then go back to setting it in script rather than stylesheet.
</blockquote></details>

Ok will do. There's like an implicit container style that was messing up with my margins and width. Even setting the values here wasn't enough without "!important". Disabling the container style fixed it so... not sure what's up with that. Tried the other options but none of them worked completely as desired.

bdomen-ggl (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1459)* status: 3 of 5 files reviewed, 2 unresolved discussions (waiting on @SanjayVas)

*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1459#-NqhHkX97zXor_KF_6HT:-NqhHkX97zXor_KF_6HU:bb336d2) (no related file):*
@SanjayVas FYI: ready for a re-review.

___

SanjayVas (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1459)* status: all files reviewed, 1 unresolved discussion

PR #1463, opened 2024-02-06
Use hashing rather than RNG for VID sampling of synthetic events.

SanjayVas:
This allows implementations to be deterministic and language-agnostic.

### DIFF
--- a/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/BUILD.bazel
+++ b/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/BUILD.bazel
@@ -94,6 +94,7 @@ kt_jvm_library(
     deps = [
         ":labeled_event",
         "//src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata/testing:simulator_synthetic_data_spec_kt_jvm_proto",
+        "@wfa_common_jvm//imports/java/com/google/common:guava",
         "@wfa_common_jvm//src/main/kotlin/org/wfanet/measurement/common",
     ],
 )
--- a/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SyntheticDataGeneration.kt
+++ b/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SyntheticDataGeneration.kt
@@ -16,14 +16,15 @@
 
 package org.wfanet.measurement.loadtest.dataprovider
 
+import com.google.common.hash.Hashing
 import com.google.protobuf.Descriptors.FieldDescriptor
 import com.google.protobuf.Message
+import com.google.protobuf.kotlin.toByteStringUtf8
+import java.nio.ByteBuffer
+import java.nio.ByteOrder
 import java.time.ZoneOffset
-import java.util.Random
 import kotlin.math.max
 import kotlin.math.min
-import kotlin.random.Random as KotlinRandom
-import kotlin.random.asKotlinRandom
 import org.wfanet.measurement.api.v2alpha.event_group_metadata.testing.FieldValue
 import org.wfanet.measurement.api.v2alpha.event_group_metadata.testing.SimulatorSyntheticDataSpec
 import org.wfanet.measurement.api.v2alpha.event_group_metadata.testing.SyntheticEventGroupSpec
@@ -36,11 +37,14 @@ import org.wfanet.measurement.common.rangeTo
 import org.wfanet.measurement.common.toLocalDate
 
 object SyntheticDataGeneration {
+  private val VID_SAMPLING_FINGERPRINT_FUNCTION = Hashing.farmHashFingerprint64()
+  private const val FINGERPRINT_BUFFER_SIZE_BYTES = 512
+
   /**
-   * Generates a sequence of [EventQuery.LabeledEvent].
+   * Generates a sequence of [LabeledEvent].
    *
    * Consumption of [Sequence] throws
-   * * [IllegalArgumentException] when [SimulatorSyntheticDataSpec] is invalid, or incompatible
+   * * [IllegalStateException] when [SimulatorSyntheticDataSpec] is invalid, or incompatible
    * * with [T].
    *
    * @param messageInstance an instance of the event message type [T]
@@ -52,28 +56,6 @@ object SyntheticDataGeneration {
     populationSpec: SyntheticPopulationSpec,
     syntheticEventGroupSpec: SyntheticEventGroupSpec,
   ): Sequence<LabeledEvent<T>> {
-    var samplingRequired = false
-    val vidRangeSpecs =
-      syntheticEventGroupSpec.dateSpecsList
-        .flatMap { it.frequencySpecsList }
-        .flatMap { it.vidRangeSpecsList }
-
-    for (vidRangeSpec in vidRangeSpecs) {
-      val vidRangeWidth = vidRangeSpec.vidRange.endExclusive - vidRangeSpec.vidRange.start
-      check(vidRangeWidth >= vidRangeSpec.sampleSize) {
-        "all vidRange widths should be larger than sampleSizes"
-      }
-      if (vidRangeSpec.sampleSize > 0) {
-        samplingRequired = true
-      }
-    }
-
-    if (samplingRequired) {
-      check(syntheticEventGroupSpec.rngType == SyntheticEventGroupSpec.RngType.JAVA_UTIL_RANDOM) {
-        "Expecting JAVA_UTIL_RANDOM rng type, got ${syntheticEventGroupSpec.rngType}"
-      }
-    }
-
     val subPopulations = populationSpec.subPopulationsList
 
     return sequence {
@@ -84,10 +66,15 @@ object SyntheticDataGeneration {
           check(!frequencySpec.hasOverlaps()) { "The VID ranges should be non-overlapping." }
 
           for (vidRangeSpec: VidRangeSpec in frequencySpec.vidRangeSpecsList) {
-            val random = Random(vidRangeSpec.randomSeed).asKotlinRandom()
             val subPopulation: SubPopulation =
               vidRangeSpec.vidRange.findSubPopulation(subPopulations)
-                ?: throw IllegalArgumentException()
+                ?: error("Sub-population not found")
+            check(vidRangeSpec.samplingRate in 0.0..1.0) { "Invalid sampling_rate" }
+            if (vidRangeSpec.sampled) {
+              check(syntheticEventGroupSpec.samplingNonce != 0L) {
+                "sampling_nonce is required for VID sampling"
+              }
+            }
 
             val builder: Message.Builder = messageInstance.newBuilderForType()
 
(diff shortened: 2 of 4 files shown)


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1463)

uakyol (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1463)* status: all files reviewed, 2 unresolved discussions (waiting on @kungfucraig and @SanjayVas)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata/testing/simulator_synthetic_data_spec.proto` line 152 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1463#-NpxoifA904G4uJUKx1M:-NpxoifA904G4uJUKx1N:bxdwdf2) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/be4884df3d048896a228a7d140e24237b0e70722/src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata/testing/simulator_synthetic_data_spec.proto#L152)):*
> ```proto
> 
>   // Type of random number generator to sample vids.
>   enum RngType {
> ```

Can you please also get rid of this?
___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata/testing/simulator_synthetic_data_spec.proto` line 161 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1463#-Npxon3Z0F-6LEzz1BEX:-Npxon3Z0F-6LEzz1BEY:bkkrhws) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/be4884df3d048896a228a7d140e24237b0e70722/src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata/testing/simulator_synthetic_data_spec.proto#L161)):*
> ```proto
>   // Random Number Generator type for this `SyntheticEventGroupSpec`.
>   // Required if any VidRangeSpec specifies a sample_size.
>   RngType rng_type = 4;
> ```

And this

uakyol (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1463)* status: all files reviewed, 4 unresolved discussions (waiting on @kungfucraig and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SyntheticDataGeneration.kt` line 64 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1463#-NpyeLnc7I8PNKUVX1Xs:-NpyeLnc7I8PNKUVX1Xt:brsglu6) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/be4884df3d048896a228a7d140e24237b0e70722/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SyntheticDataGeneration.kt#L64)):*
> ```Kotlin
>         for (frequencySpec: SyntheticEventGroupSpec.FrequencySpec in dateSpec.frequencySpecsList) {
> 
>           check(!frequencySpec.hasOverlaps()) { "The VID ranges should be non-overlapping." }
> ```

There is actually another problem here as well from the original implementation. 

We need to not just check if an individual frequencySpec has overlaps for its vidRanges, but we also need to check across all frequencySpecs that no vidRanges overlap. Because, if they do, same vids will be sampled (or taken if no sampling is given) for different frequencies. 

example :

FrequencySpec1  = {freq =1, vid_range : [0,100] }
FrequencySpec2  = {freq =2, vid_range : [50,150] }

this means vids 50-100 will actually have freq of 3 (neither 1 or 2 but summation of them)

wdyt, hope Im not misunderstanding.

CC: @kungfucraig 
___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata/testing/simulator_synthetic_data_spec.proto` line 108 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1463#-NpyfFCg1gsU3sOHcpZH:-NpyfFCg1gsU3sOHcpZI:b-9i4spx) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/be4884df3d048896a228a7d140e24237b0e70722/src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata/testing/simulator_synthetic_data_spec.proto#L108)):*
> ```proto
>     message VidRangeSpec {
>       // A range of VIDs within a single `SubPopulation`.
>       VidRange vid_range = 1;
> ```

We should comment here that vid_ranges should not overlap across FrequencySpecs -  I explained more in the above comment.

uakyol (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1463)* status: all files reviewed, 4 unresolved discussions (waiting on @kungfucraig and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SyntheticDataGeneration.kt` line 64 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1463#-NpyeLnc7I8PNKUVX1Xs:-Npyi8wtCwTcn-cM2Sif:bmuhrh7) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/be4884df3d048896a228a7d140e24237b0e70722/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SyntheticDataGeneration.kt#L64)):*

There is actually another problem here as well from the original implementation. 

We need to not just check if an individual frequencySpec has overlaps for its vidRanges, but we also need to check across all frequencySpecs that no vidRanges overlap. Because, if they do, same vids will be sampled (or taken if no sampling is given) for different frequencies. 

example :

FrequencySpec1  = {freq =1, vid_range : [0,100] }
FrequencySpec2  = {freq =2, vid_range : [50,150] }

this means vids 50-100 will actually have freq of 3 (neither 1 or 2 but summation of them)

wdyt, hope Im not misunderstanding.

CC: @kungfucraig
</blockquote></details>

I realized this implementing the cartesian spec. This will cause a problem for that. Because for the cartesian spec, we create different 'regions' for each of the combination of frequency, population, nonpopulation dimensions. All these regions will have to be separate since we don't want a single vid to be sampled for 2 of them. E.g. a vid sampled for both age_18_25 and age_26_35 ranges.

Basically, this format of synthetic spec gives us only the opportunity to sample from a small subrange while what we want to do is to be able to sample from a the whole range of the population spec. I would vote that we abandon this approach and I will take care of sampling in the cartesian spec PR.

uakyol (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1463)* status: all files reviewed, 4 unresolved discussions (waiting on @kungfucraig and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SyntheticDataGeneration.kt` line 64 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1463#-NpyeLnc7I8PNKUVX1Xs:-NpymxoG4m2S6DuLoBY2:b8z400f) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/be4884df3d048896a228a7d140e24237b0e70722/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SyntheticDataGeneration.kt#L64)):*

I realized this implementing the cartesian spec. This will cause a problem for that. Because for the cartesian spec, we create different 'regions' for each of the combination of frequency, population, nonpopulation dimensions. All these regions will have to be separate since we don't want a single vid to be sampled for 2 of them. E.g. a vid sampled for both age_18_25 and age_26_35 ranges.

Basically, this format of synthetic spec gives us only the opportunity to sample from a small subrange while what we want to do is to be able to sample from a the whole range of the population spec. I would vote that we abandon this approach and I will take care of sampling in the cartesian spec PR.
</blockquote></details>

In any case, we still would need to implement the check of cross frequencySpec vidRanges not overlapping. That is independent of cartesian spec problem.

SanjayVas (commented):
Updated to include non-population field values in the sampling fingerprint. PTAL.

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1463#-:-Nq-vALA2DwawpTbCsQk:b-8gduk0)* status: 1 of 4 files reviewed, 5 unresolved discussions (waiting on @kungfucraig and @uakyol)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SyntheticDataGeneration.kt` line 64 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1463#-NpyeLnc7I8PNKUVX1Xs:-Nq-upVKDdYzMx_8C_J-:bvnxxsl) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/be4884df3d048896a228a7d140e24237b0e70722/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SyntheticDataGeneration.kt#L64)):*

I think this is WAI, but we're meeting later so let's discuss then.

We could certainly check the population spec for disjointness though (if we don't already).
</blockquote></details>

As per offline conversation, the ability for VID ranges within different FrequencySpecs to overlap is intentional. Two events from the same VID can have different non-population field values. For example, if the same user views an ad from two different devices.
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SyntheticDataGeneration.kt` line 143 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1463#-NpzCHlBCVicI0QGElB-:-Nq-uXF14eJVEj-gSUWO:b66taxw) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/be4884df3d048896a228a7d140e24237b0e70722/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/SyntheticDataGeneration.kt#L143)):*

Shouldn't this always be true?
</blockquote></details>

This is determining whether the VidRangeSpec should be sampled (vs. just using all values within the range). Therefore this should be true when the sampling rate is neither unspecified nor 1.
___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata/testing/simulator_synthetic_data_spec.proto` line 108 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1463#-NpyfFCg1gsU3sOHcpZH:-Nq-ulnfE1gCh2yAQyoB:b-d8pbl7) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/be4884df3d048896a228a7d140e24237b0e70722/src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata/testing/simulator_synthetic_data_spec.proto#L108)):*

We should comment here that vid_ranges should not overlap across FrequencySpecs -  I explained more in the above comment.
</blockquote></details>

This is not required. See above.
___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata/testing/simulator_synthetic_data_spec.proto` line 152 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1463#-NpxoifA904G4uJUKx1M:-Nq-v8t3Czo7lc2lzYTm:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/be4884df3d048896a228a7d140e24237b0e70722/src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata/testing/simulator_synthetic_data_spec.proto#L152)):*

Can you please also get rid of this?
</blockquote></details>

Done.
___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata/testing/simulator_synthetic_data_spec.proto` line 161 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1463#-Npxon3Z0F-6LEzz1BEX:-Nq-v9BV-h_r_KKmq9Kv:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/be4884df3d048896a228a7d140e24237b0e70722/src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata/testing/simulator_synthetic_data_spec.proto#L161)):*

And this
</blockquote></details>

Done.

stevenwarejones (approved):
> Reviewed 1 of 4 files at r1, 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1463)* status: all files reviewed, 1 unresolved discussion (waiting on @kungfucraig)

kungfucraig (approved):
> Reviewed 1 of 4 files at r1, 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1463)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @SanjayVas)

PR #1468, opened 2024-02-09
Test including time_util.h.

SanjayVas:
Test to ensure that compilation succeeds when including time_util.h.

//src/test/cc/wfa/measurement/internal/duchy/protocol/liquid_legions_v2:liquid_legions_v2_encryption_utility_test passes locally


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1468)

PR #1469, opened 2024-02-12
Adds a standalone job for running decrypt private membership queries

jonmolle:
Adds a standalone decrypt job to the panel exchange tools.

### DIFF
--- a/src/main/kotlin/org/wfanet/panelmatch/client/tools/BUILD.bazel
+++ b/src/main/kotlin/org/wfanet/panelmatch/client/tools/BUILD.bazel
@@ -72,6 +72,47 @@ java_binary(
     runtime_deps = [":serialize_exchange_workflow"],
 )
 
+
+kt_jvm_library(
+    testonly = True,
+    name = "decrypt_private_membership_query_results",
+    srcs = ["DecryptPrivateMembershipQueryResults.kt"],
+    deps = [
+        "//imports/java/org/apache/beam/runners/direct",
+        "//src/main/kotlin/org/wfanet/panelmatch/client/deploy",
+        "//src/main/kotlin/org/wfanet/panelmatch/client/deploy/example",
+        "//src/main/kotlin/org/wfanet/panelmatch/client/privatemembership",
+        "//src/main/kotlin/org/wfanet/panelmatch/client/storage",
+        "//src/main/kotlin/org/wfanet/panelmatch/client/storage/aws/s3",
+        "//src/main/kotlin/org/wfanet/panelmatch/client/storage/gcloud/gcs",
+        "//src/main/kotlin/org/wfanet/panelmatch/common",
+        "//src/main/kotlin/org/wfanet/panelmatch/common/certificates/aws",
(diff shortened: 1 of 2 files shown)


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1469)

stevenwarejones (approved):
A few comments but I'm happy to just land this and I can take care of what I said in a follow-up

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1469#-:-NqZ4j_ACkT4AbtMwMXK:bqh0f5k)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @jonmolle)

___
*[`src/main/kotlin/org/wfanet/panelmatch/client/tools/DecryptPrivateMembershipQueryResults.kt` line 95 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1469#-NqZ3D_J2EC5IFW11peD:-NqZ3D_K8djcCHt_t_Lj:bvgzu9w) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/b6258c76c75f2d771e69aba436c6e1a6cab1a1b6/src/main/kotlin/org/wfanet/panelmatch/client/tools/DecryptPrivateMembershipQueryResults.kt#L95)):*
> ```Kotlin
> 
>   @Option(
>     names = ["--exchange-step-file"],
> ```

an exchange step implies an exchange-date. I'd prefer something slightly more flexible that passes in separate flags for exchangeworkflow and exchange-date and step-index and doesn't require reserialization for each day you want to run. I can do this in a follow up if you prefer since I already have this change on my branch. 

Also, a nit but maybe just change the filename to BeamJobsMain.kt -- we can extend it to run any beam step.
___
*[`src/main/kotlin/org/wfanet/panelmatch/client/tools/DecryptPrivateMembershipQueryResults.kt` line 162 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1469#-NqZ4QkL-h2NduVhQNHa:-NqZ4QkL-h2NduVhQNHb:b9ezrq6) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/b6258c76c75f2d771e69aba436c6e1a6cab1a1b6/src/main/kotlin/org/wfanet/panelmatch/client/tools/DecryptPrivateMembershipQueryResults.kt#L162)):*
> ```Kotlin
> 
>   private fun makePipelineOptions(): PipelineOptions {
>     return PipelineOptionsFactory.`as`(BeamOptions::class.java).apply {
> ```

nit: we really shouldn't use BeamOptions here as it extends Dataflow only options, but I can take care of that in a follow-up.

stevenwarejones (changes requested):
Lets close this one out.

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1469#-:-Nqee5pU1u5fSva6-6_V:b-p40ljj)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @jonmolle)

PR #1474, opened 2024-02-14
Update UI from demo work

bdomen-ggl:
Several small changes from added features during the demo:
* Added the "demo" banner at the top of the site and made it sticky.
* Updated some card labels in the Report view and adjusted the styles so the columns aren't fixed. Didn't specifically get feedback from UX on the behavior but this was requested from users.
* Also made the summary cards more flexible for different content.
* Added an overflow to the card in case the user is on a small screen. This was noticeable in the summary table.
* Removed "on target reach" from the charts.
* Updated the title.
* Added Google Analytics.

Bug Fixes:
* Added a margin at the bottom of the site so the feedback button isn't stuck over things at the bottom like charts.
* Fixed fonts. There was an extra "./" in the font src that would work locally but screw up webpack with routes on the server. Also removed the extra font styles and just left sans serif.
* Added more cases to the number to magnitude converter including handling the 0 case to make the logic simpler.
* Updated BUILD files for some ChartGroups. These charts were removed from the current UI so they didn't fail the build.

TODO:
* Update the test data based on the latest Report model.


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1474)

PR #1476, opened 2024-02-15
Update HMSS stages. Revert Duchy ControlService into Blob-Only Pattern.

renjiezh:
After this update, for the first non-aggregator, the normal stage transition is:
 INITIALIZED -> WAIT_TO_START -> SETUP_PHASE ->
WAIT_ON_SHUFFLE_INPUT_PHASE_TWO -> SHUFFLE_PHASE
-> COMPLETE

For the second non-aggregator, the normal stage transition is:
INITIALIZED -> WAIT_ON_SHUFFLE_INPUT_PHASE_ONE -> SETUP_PHASE
-> SHUFFLE_PHASE -> COMPLETE

For the aggregator, the normal stage transition is:
WAIT_ON_AGGREGATION_INPUT -> AGGREGATION_PHASE -> COMPLETE


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1476)

SanjayVas (approved):
I'm basically trusting you on advanceComputation as I have never fully understood it.

> Reviewed 13 of 14 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1476#-:-NqnUbeMA_gRDewPCtEg:bur3ogc)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @ple13)

ple13 (changes requested):
> Reviewed 13 of 14 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1476)* status: all files reviewed, 4 unresolved discussions (waiting on @renjiezh)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/db/computation/HonestMajorityShareShuffleProtocol.kt` line 141 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1476#-NqnrZ4XErdl-eKYprMR:-NqnrZ4XErdl-eKYprMS:bhqt6lb) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e1de6b280d911c10d3301ea5ed3180217b8e5358/src/main/kotlin/org/wfanet/measurement/duchy/db/computation/HonestMajorityShareShuffleProtocol.kt#L141)):*
> ```Kotlin
>           SET_PARTICIPANT_PARAMS_PHASE,
>           WAIT_ON_REQUISITION_FULFILLMENT -> 0
>           // The output is received from the peer non-aggregator duchy:
> ```

The comment is confusing when we have an output that is received from somewhere else. Did you mean the output contains data received from another duchy?
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/service/internal/computationcontrol/ProtocolStages.kt` line 188 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1476#-NqnzdMWEYIaN_QsIg4s:-NqnzdMWEYIaN_QsIg4t:b-1apq4q) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e1de6b280d911c10d3301ea5ed3180217b8e5358/src/main/kotlin/org/wfanet/measurement/duchy/service/internal/computationcontrol/ProtocolStages.kt#L188)):*
> ```Kotlin
>     dataOrigin: String,
>   ): ComputationStageBlobMetadata {
>     return when (val protocolStage = token.computationStage.honestMajorityShareShuffle) {
> ```

I guess this works as it passes the test, but it looks dangerous.

Let's keep the own syntax:
val protocolStage = token...
return when(protocolStage)

or return when(token...) instead?
___
*[`src/main/proto/wfa/measurement/internal/duchy/crypto.proto` line 64 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1476#-NqnqIWj5K2ohGN5Xjna:-NqnqIWj5K2ohGN5Xjnb:b-sevdg1) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e1de6b280d911c10d3301ea5ed3180217b8e5358/src/main/proto/wfa/measurement/internal/duchy/crypto.proto#L64)):*
> ```proto
> 
> // Keys of the encryption cipher.
> message EncryptionKeyPair {
> ```

Will we decide the encryption scheme later?
___
*[`src/main/proto/wfa/measurement/internal/duchy/protocol/honest_majority_share_shuffle.proto` line 55 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1476#-Nqnlg8_6uefZnGxTqDY:-Nqnlg8aD-bZFON-NimC:bet5oxd) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e1de6b280d911c10d3301ea5ed3180217b8e5358/src/main/proto/wfa/measurement/internal/duchy/protocol/honest_majority_share_shuffle.proto#L55)):*
> ```proto
> 
>     // The non-aggregator sends the participant params to the kingdom.
>     SET_PARTICIPANT_PARAMS_PHASE = 2;
> ```

Lets combine this phase and the initialized phase? It seems like the workers do not need to wait between the two phases.

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1476)* status: 12 of 15 files reviewed, 4 unresolved discussions (waiting on @ple13 and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/db/computation/HonestMajorityShareShuffleProtocol.kt` line 141 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1476#-NqnrZ4XErdl-eKYprMR:-NqxOosM8pOX02EmJrOf:b-5wcok1) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e1de6b280d911c10d3301ea5ed3180217b8e5358/src/main/kotlin/org/wfanet/measurement/duchy/db/computation/HonestMajorityShareShuffleProtocol.kt#L141)):*

The comment is confusing when we have an output that is received from somewhere else. Did you mean the output contains data received from another duchy?
</blockquote></details>

Done.
Updated the comment for a bit.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/service/internal/computationcontrol/ProtocolStages.kt` line 188 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1476#-NqnzdMWEYIaN_QsIg4s:-NqxOu1uEWeqr0gdn3HN:b-itl7pj) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e1de6b280d911c10d3301ea5ed3180217b8e5358/src/main/kotlin/org/wfanet/measurement/duchy/service/internal/computationcontrol/ProtocolStages.kt#L188)):*

I guess this works as it passes the test, but it looks dangerous.

Let's keep the own syntax:
val protocolStage = token...
return when(protocolStage)

or return when(token...) instead?
</blockquote></details>

Done.
The IDEA suggest me to do this way. Let me change it back as we prefer. Btw, why is it dangerous though?
___
*[`src/main/proto/wfa/measurement/internal/duchy/crypto.proto` line 64 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1476#-NqnqIWj5K2ohGN5Xjna:-NqxPX1I0AOBBf1wOAHZ:b-vcr3yv) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e1de6b280d911c10d3301ea5ed3180217b8e5358/src/main/proto/wfa/measurement/internal/duchy/crypto.proto#L64)):*

Will we decide the encryption scheme later?
</blockquote></details>

The encryption is using tink from common-jvm. The key type will be `DHKEM_X25519_HKDF_SHA256_HKDF_SHA256_AES_256_GCM`. 
___
*[`src/main/proto/wfa/measurement/internal/duchy/protocol/honest_majority_share_shuffle.proto` line 55 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1476#-Nqnlg8_6uefZnGxTqDY:-NqxQY_b3lm6K01rHN47:b-r635av) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e1de6b280d911c10d3301ea5ed3180217b8e5358/src/main/proto/wfa/measurement/internal/duchy/protocol/honest_majority_share_shuffle.proto#L55)):*

Lets combine this phase and the initialized phase? It seems like the workers do not need to wait between the two phases.
</blockquote></details>

That was the original idea but some reason drive me to separate them:
1. SetComputationParams is a call to the Kingdom but current herald (llv2's herald) does not call kingdom except `failComputation `.
2. SetComputationParams might require retry. It is much more convenient for mills to retry things because the herald has `ContinuationToken` to manage the progress. If herald fails an task, it would claim all tasks after the failed one even the other tasks are finished. 
If we'd like to squeeze for better performance, we can still merge these two stages in the future. I am open to that. But for now, leaving them as two stages makes it consistent with LLv2 thus less error-prone.

ple13 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1476)* status: all files reviewed, 2 unresolved discussions (waiting on @renjiezh)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/service/internal/computationcontrol/ProtocolStages.kt` line 188 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1476#-NqnzdMWEYIaN_QsIg4s:-Nr6W8vL2UI2LrVUR6xd:b-8v37pr) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e1de6b280d911c10d3301ea5ed3180217b8e5358/src/main/kotlin/org/wfanet/measurement/duchy/service/internal/computationcontrol/ProtocolStages.kt#L188)):*

Done.
The IDEA suggest me to do this way. Let me change it back as we prefer. Btw, why is it dangerous though?
</blockquote></details>

My understanding is: when(protocolStage) works on the value of protocolStage, while when(val protocolStage = token..) works on the return value of the assignment expression.
___
*[`src/main/proto/wfa/measurement/internal/duchy/protocol/honest_majority_share_shuffle.proto` line 55 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1476#-Nqnlg8_6uefZnGxTqDY:-Nr6YsoHEcilu3j8P-h0:b-31b6vf) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e1de6b280d911c10d3301ea5ed3180217b8e5358/src/main/proto/wfa/measurement/internal/duchy/protocol/honest_majority_share_shuffle.proto#L55)):*

That was the original idea but some reason drive me to separate them:
1. SetComputationParams is a call to the Kingdom but current herald (llv2's herald) does not call kingdom except `failComputation `.
2. SetComputationParams might require retry. It is much more convenient for mills to retry things because the herald has `ContinuationToken` to manage the progress. If herald fails an task, it would claim all tasks after the failed one even the other tasks are finished. 
If we'd like to squeeze for better performance, we can still merge these two stages in the future. I am open to that. But for now, leaving them as two stages makes it consistent with LLv2 thus less error-prone.
</blockquote></details>

I'm out of context here as I cannot find SetComputationParams in our code. I assume that the participant params are the worker's public keys. Is it correct? In llv2, duchies sample the key pair and send their public key in the initialization phase (https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e01466201b3f42ffe8bd32fe1acb81169972e059/src/main/proto/wfa/measurement/internal/duchy/protocol/liquid_legions_sketch_aggregation_v2.proto#L36).

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1476)* status: all files reviewed, 2 unresolved discussions (waiting on @ple13)

___
*[`src/main/proto/wfa/measurement/internal/duchy/protocol/honest_majority_share_shuffle.proto` line 55 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1476#-Nqnlg8_6uefZnGxTqDY:-Nr6baivCK_J9QcSfLHB:b-oehvdk) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e1de6b280d911c10d3301ea5ed3180217b8e5358/src/main/proto/wfa/measurement/internal/duchy/protocol/honest_majority_share_shuffle.proto#L55)):*

I'm out of context here as I cannot find SetComputationParams in our code. I assume that the participant params are the worker's public keys. Is it correct? In llv2, duchies sample the key pair and send their public key in the initialization phase (https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e01466201b3f42ffe8bd32fe1acb81169972e059/src/main/proto/wfa/measurement/internal/duchy/protocol/liquid_legions_sketch_aggregation_v2.proto#L36).
</blockquote></details>

The full name of the rpc call is `SetParticipantRequisitionParams`.
I see your point. Yes, we can use INITIALIZATION_PHASE for the purpose. I will update the proto and description. Thanks.

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1476)* status: 10 of 14 files reviewed, all discussions resolved (waiting on @ple13 and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/service/internal/computationcontrol/ProtocolStages.kt` line 188 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1476#-NqnzdMWEYIaN_QsIg4s:-Nr7EBTv7stS2uDq0oIZ:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e1de6b280d911c10d3301ea5ed3180217b8e5358/src/main/kotlin/org/wfanet/measurement/duchy/service/internal/computationcontrol/ProtocolStages.kt#L188)):*

My understanding is: when(protocolStage) works on the value of protocolStage, while when(val protocolStage = token..) works on the return value of the assignment expression.
</blockquote></details>

Done.

ple13 (approved):
> Reviewed 4 of 4 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1476)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @renjiezh)

SanjayVas (approved):
> Reviewed 4 of 4 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1476)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @stevenwarejones)

stevenwarejones (approved):
> Reviewed 9 of 14 files at r1, 1 of 1 files at r2, 4 of 4 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1476)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @renjiezh)

PR #1383, opened 2024-01-03
add maxCoroutines parameter to CoroutineLauncher

stevenwarejones:
The current CoroutineLauncher has not limit on the number of coroutines it launches. This works well on a platform like DataflowRunner but performs poorly with the DirectExecutor. This PR
1. Adds a parameter to specify the max number of jobs that should be executed in parallel with the CoroutineLauncher.
2. Sets the max number of jobs to 1 for the Aws example daemon.

PR #1485, opened 2024-02-15
add maxCoroutines parameter to CoroutineLauncher (#1383)

stevenwarejones:
The current CoroutineLauncher has not limit on the number of coroutines it launches. This works well on a platform like DataflowRunner but performs poorly with the DirectExecutor. This PR
1. Adds a parameter to specify the max number of jobs that should be executed in parallel with the CoroutineLauncher.
2. Sets the max number of jobs to 1 for the Aws example daemon.


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1485)

Issue #247, opened 2021-09-02
Update grpc error codes in Kingdom Internal Services


uakyol:
Parent not found -> FAILED_PRECONDITION  and furthermore have PreconditionFailure set in the error details (see https://github.com/googleapis/api-common-protos/blob/37d5125da5c90f2124d15908a54a32ed3f470bc2/google/rpc/error_details.proto#L140)

self not found -> NOT_FOUND
some resource other than parent or self not found -> INVALID_ARGUMENT

SanjayVas:
> some resource other than parent or self not found -> INVALID_ARGUMENT

I think this would still be FAILED_PRECONDITION. As per [Status codes and their use in gRPC](https://github.com/grpc/grpc/blob/master/doc/statuscodes.md), INVALID_ARGUMENT is for when the arguments are incorrect regardless of the state of the system (e.g. some field is malformed or missing). In this case, the arguments could be correct if the entity with the given ID already existed. The fact that we don't allow user-specified IDs for creation is a separate matter.

PR #1520, opened 2024-03-12
Add detailed conversion of internal to external errors

jcorilla:


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1520)

SanjayVas (changes requested):
> Reviewed 3 of 5 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1520)* status: 3 of 5 files reviewed, 2 unresolved discussions (waiting on @jcorilla and @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/common/grpc/BUILD.bazel` line 17 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1520#-Nssa7H2ApnIwVVM1JrC:-Nssa7H2ApnIwVVM1JrD:bffiwz7) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3f9b18508df85d0b21d79a4a04dd7e12ab32fbc8/src/main/kotlin/org/wfanet/measurement/common/grpc/BUILD.bazel#L17)):*
> ```Starlark
>         "//src/main/proto/google/rpc:error_details_kt_jvm_proto",
>         "//src/main/proto/google/rpc:status_kt_jvm_proto",
>         "//src/main/proto/wfa/measurement/internal/kingdom:error_code_kt_jvm_proto",
> ```

Anything under the `common` is meant to be for things that eventually get moved to common-jvm, so nothing in there should have knowledge of or depend on the APIs. Move the parts that depend on Kingdom error code to somewhere under src/main/kotlin/org/wfanet/measurement/kingdom
___
*[`src/main/kotlin/org/wfanet/measurement/common/grpc/ErrorInfo.kt` line 51 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1520#-NssbBHo31s6sKYP-ZrO:-NssbBHp1lrZRX4Eqs3h:b-5a43un) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3f9b18508df85d0b21d79a4a04dd7e12ab32fbc8/src/main/kotlin/org/wfanet/measurement/common/grpc/ErrorInfo.kt#L51)):*
> ```Kotlin
>         when (key) {
>           "external_computation_id" ->
>             metadataMap["computation_id"] = externalIdToApiId(value.toLong())
> ```

We ideally want to convert from tuples of external ID to resource names. e.g. if an error is trying to reference a Measurement, it should have both external_measurement_consumer_id and external_measurement_id that would need to be converted into a Measurement resource name using [`MeasurementKey`](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/main/src/main/kotlin/org/wfanet/measurement/api/v2alpha/MeasurementKey.kt).

I'm not sure this can be done in a generic way and instead must be contextual based on the actual error. Note that we'd also want to build new error messages that can include these dynamic variables. e.g. "Measurement measurementConsumers/foo/measurements/bar not found". See https://google.aip.dev/193#dynamic-variables



_Code quote:_
```Kotlin
metadataMap["computation_id"] = externalIdToApiId(value.toLong())
```

tristanvuong2021 (commented):
> Reviewed 3 of 5 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1520)* status: 3 of 5 files reviewed, 2 unresolved discussions (waiting on @jcorilla)

jcorilla (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1520)* status: 1 of 7 files reviewed, 2 unresolved discussions (waiting on @SanjayVas and @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/common/grpc/BUILD.bazel` line 17 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1520#-Nssa7H2ApnIwVVM1JrC:-NszZL7zBfV7HtYLi3mm:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3f9b18508df85d0b21d79a4a04dd7e12ab32fbc8/src/main/kotlin/org/wfanet/measurement/common/grpc/BUILD.bazel#L17)):*

Anything under the `common` is meant to be for things that eventually get moved to common-jvm, so nothing in there should have knowledge of or depend on the APIs. Move the parts that depend on Kingdom error code to somewhere under src/main/kotlin/org/wfanet/measurement/kingdom
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/common/grpc/ErrorInfo.kt` line 51 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1520#-NssbBHo31s6sKYP-ZrO:-NszZQKJ4fYbhe-vK2nC:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3f9b18508df85d0b21d79a4a04dd7e12ab32fbc8/src/main/kotlin/org/wfanet/measurement/common/grpc/ErrorInfo.kt#L51)):*

We ideally want to convert from tuples of external ID to resource names. e.g. if an error is trying to reference a Measurement, it should have both external_measurement_consumer_id and external_measurement_id that would need to be converted into a Measurement resource name using [`MeasurementKey`](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/main/src/main/kotlin/org/wfanet/measurement/api/v2alpha/MeasurementKey.kt).

I'm not sure this can be done in a generic way and instead must be contextual based on the actual error. Note that we'd also want to build new error messages that can include these dynamic variables. e.g. "Measurement measurementConsumers/foo/measurements/bar not found". See https://google.aip.dev/193#dynamic-variables
</blockquote></details>

Done.

SanjayVas (changes requested):
> Reviewed 2 of 6 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1520)* status: 3 of 7 files reviewed, 4 unresolved discussions (waiting on @jcorilla and @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/common/grpc/ErrorInfo.kt` line 23 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1520#-Nt27-Yn3YtSDkmsYD85:-Nt27-Yn3YtSDkmsYD86:b-rkixkc) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5f3754b8eedc7cc7354dbcf3affb2062847d8413/src/main/kotlin/org/wfanet/measurement/common/grpc/ErrorInfo.kt#L23)):*
> ```Kotlin
> import io.grpc.protobuf.StatusProto
> 
> val StatusException.errorInfo: ErrorInfo?
> ```

Complete the move of this method from Population.kt in this same PR (remove it from there and update all references)
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/common/StatusExceptionExtension.kt` line 1 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1520#-Nt27rzpAjChTrD3cg-L:-Nt27rzpAjChTrD3cg-M:b-59ftni) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5f3754b8eedc7cc7354dbcf3affb2062847d8413/src/main/kotlin/org/wfanet/measurement/kingdom/service/common/StatusExceptionExtension.kt#L1)):*
> ```Kotlin
> /*
> ```

We don't name the files with extension methods "Extension".
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/common/StatusExceptionExtension.kt` line 17 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1520#-Nt27kHiCZK_xU3kPwwd:-Nt27kHiCZK_xU3kPwwe:ba54w) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5f3754b8eedc7cc7354dbcf3affb2062847d8413/src/main/kotlin/org/wfanet/measurement/kingdom/service/common/StatusExceptionExtension.kt#L17)):*
> ```Kotlin
>  */
> 
> package org.wfanet.measurement.kingdom.service.common
> ```

This depends on v2alpha messages, so it shouldn't be in a common package



_Suggestion:_
```Kotlin
org.wfanet.measurement.kingdom.service.api.v2alpha
```
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/common/StatusExceptionExtension.kt` line 38 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1520#-Nt29abZ2A_GW4bnTXHZ:-Nt29abZ2A_GW4bnTXH_:bpoci3s) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5f3754b8eedc7cc7354dbcf3affb2062847d8413/src/main/kotlin/org/wfanet/measurement/kingdom/service/common/StatusExceptionExtension.kt#L38)):*
> ```Kotlin
>   status: Status,
>   description: String,
> ): StatusRuntimeException {
> ```

You have two Status objects effectively passed to this method: the original one from the StatusException and the new one passed in as an argument. This is confusing. Additionally there are two error messages: the description that is passed in and the one potentially set by this method. The semantics of how this happens is unclear to the caller.

Perhaps instead an extension on Status? Note that the status already includes the description. Make sure to document the method.



_Suggestion:_
```Kotlin
/** Converts this [Status] to a [StatusRuntimeException] with details from [internalApiException]. This may replace the error info and description... */
fun Status.toRuntimeException(
  internalApiException: StatusException,
): StatusRuntimeException {
```

jcorilla (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1520)* status: 2 of 11 files reviewed, 4 unresolved discussions (waiting on @SanjayVas and @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/common/grpc/ErrorInfo.kt` line 23 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1520#-Nt27-Yn3YtSDkmsYD85:-Nt41XESE08d4Z33_aR_:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5f3754b8eedc7cc7354dbcf3affb2062847d8413/src/main/kotlin/org/wfanet/measurement/common/grpc/ErrorInfo.kt#L23)):*

Complete the move of this method from Population.kt in this same PR (remove it from there and update all references)
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/common/StatusExceptionExtension.kt` line 1 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1520#-Nt27rzpAjChTrD3cg-L:-Nt41_SR3jAaQyBAgdPh:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5f3754b8eedc7cc7354dbcf3affb2062847d8413/src/main/kotlin/org/wfanet/measurement/kingdom/service/common/StatusExceptionExtension.kt#L1)):*

We don't name the files with extension methods "Extension".
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/common/StatusExceptionExtension.kt` line 17 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1520#-Nt27kHiCZK_xU3kPwwd:-Nt41atjCeqOxeSCKifJ:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5f3754b8eedc7cc7354dbcf3affb2062847d8413/src/main/kotlin/org/wfanet/measurement/kingdom/service/common/StatusExceptionExtension.kt#L17)):*

This depends on v2alpha messages, so it shouldn't be in a common package
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/common/StatusExceptionExtension.kt` line 38 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1520#-Nt29abZ2A_GW4bnTXHZ:-Nt41c464Op3ugoFAsHp:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5f3754b8eedc7cc7354dbcf3affb2062847d8413/src/main/kotlin/org/wfanet/measurement/kingdom/service/common/StatusExceptionExtension.kt#L38)):*

You have two Status objects effectively passed to this method: the original one from the StatusException and the new one passed in as an argument. This is confusing. Additionally there are two error messages: the description that is passed in and the one potentially set by this method. The semantics of how this happens is unclear to the caller.

Perhaps instead an extension on Status? Note that the status already includes the description. Make sure to document the method.
</blockquote></details>

Done.

SanjayVas (changes requested):
> Reviewed 9 of 9 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1520)* status: all files reviewed, 5 unresolved discussions (waiting on @jcorilla and @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/common/grpc/ErrorInfo.kt` line 36 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1520#-NtHZ99V9hMfsSsRU-rT:-NtHZ99V9hMfsSsRU-rU:b-wbe9bu) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3eede55fcb09a3f5f639e43e0157db9837536daf/src/main/kotlin/org/wfanet/measurement/common/grpc/ErrorInfo.kt#L36)):*
> ```Kotlin
> val StatusRuntimeException.errorInfo: ErrorInfo?
>   get() {
>     val errorInfoFullName = ErrorInfo.getDescriptor().fullName
> ```

nit: extract a private function that takes in status and trailers. You can use this function to implement both properties.
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/MeasurementsService.kt` line 313 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1520#-NtHbksI1NOgdNr2-20i:-NtHbksI1NOgdNr2-20j:b-e0ucmi) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3eede55fcb09a3f5f639e43e0157db9837536daf/src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/MeasurementsService.kt#L313)):*
> ```Kotlin
>             throw Status.FAILED_PRECONDITION.toExternalStatusRuntimeException(ex)
>           Status.Code.NOT_FOUND ->
>             throw Status.FAILED_PRECONDITION.toExternalStatusRuntimeException(ex)
> ```

nit: I think this one is slightly out of date, as things got a little confused in terms of what error code the internal API should throw under what conditions. I believe this internal method throws NOT_FOUND when the MC isn't found, which for the public API translates to the resource we're operating on not being found as this method operates on the `measurements` resource collection of a MeasurementConsumer.

See the discussion in https://github.com/world-federation-of-advertisers/cross-media-measurement/issues/247#issuecomment-931787983

tl;dr This should be NOT_FOUND when the MC is not found
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/InternalStatusConversion.kt` line 45 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1520#-NtHZWYV-kr4Jy1RsFxZ:-NtHZWYW3057VmfScWoT:b-yfablh) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3eede55fcb09a3f5f639e43e0157db9837536daf/src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/InternalStatusConversion.kt#L45)):*
> ```Kotlin
>   var errorMessage = this.description ?: "Unknown exception."
>   val metadataMap = mutableMapOf<String, String>()
>   if (errorInfo != null)
> ```

Always use braces for `if` expressions that do not fit on one line. See https://developer.android.com/kotlin/style-guide#braces
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/InternalStatusConversion.kt` line 46 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1520#-NtH_7q14FJUq6qmudiB:-NtH_7q14FJUq6qmudiC:b7t9sbw) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3eede55fcb09a3f5f639e43e0157db9837536daf/src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/InternalStatusConversion.kt#L46)):*
> ```Kotlin
>   val metadataMap = mutableMapOf<String, String>()
>   if (errorInfo != null)
>     if (errorInfo.domain == ErrorCode.getDescriptor().fullName) {
> ```

1. Is the overall condition not just `errorInfo != null && errorInfo.domain == ErrorCode.getDescriptor().fullName` ?
2. Can we just do an early return to avoid nesting? That is to say if the above condition is false, just return the result of `toRuntimeException()`. See https://testing.googleblog.com/2017/06/code-health-reduce-nesting-reduce.html
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/InternalStatusConversion.kt` line 128 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1520#-NtH_G_RAi7V5XAe4ytn:-NtH_G_RAi7V5XAe4yto:b-byuoyl) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3eede55fcb09a3f5f639e43e0157db9837536daf/src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/InternalStatusConversion.kt#L128)):*
> ```Kotlin
>       ProtoAny.pack(
>         errorInfo {
>           reason = internalApiException.status.code.toString()
> ```

Do not pass the internal Kingdom error code, as that's part of the internal API. I think you can just leave the reason/domain out for the public one.

jcorilla (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1520)* status: 6 of 11 files reviewed, 3 unresolved discussions (waiting on @SanjayVas and @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/InternalStatusConversion.kt` line 45 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1520#-NtHZWYV-kr4Jy1RsFxZ:-NtIZ0m006seuAfN5LF0:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3eede55fcb09a3f5f639e43e0157db9837536daf/src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/InternalStatusConversion.kt#L45)):*

Always use braces for `if` expressions that do not fit on one line. See https://developer.android.com/kotlin/style-guide#braces
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/InternalStatusConversion.kt` line 46 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1520#-NtH_7q14FJUq6qmudiB:-NtIZ14x8f5d1w0bVZPq:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3eede55fcb09a3f5f639e43e0157db9837536daf/src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/InternalStatusConversion.kt#L46)):*

1. Is the overall condition not just `errorInfo != null && errorInfo.domain == ErrorCode.getDescriptor().fullName` ?
2. Can we just do an early return to avoid nesting? That is to say if the above condition is false, just return the result of `toRuntimeException()`. See https://testing.googleblog.com/2017/06/code-health-reduce-nesting-reduce.html
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/InternalStatusConversion.kt` line 128 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1520#-NtH_G_RAi7V5XAe4ytn:-NtIZ23r3OLegv7GUHj4:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3eede55fcb09a3f5f639e43e0157db9837536daf/src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/InternalStatusConversion.kt#L128)):*

Do not pass the internal Kingdom error code, as that's part of the internal API. I think you can just leave the reason/domain out for the public one.
</blockquote></details>

Done.

SanjayVas (changes requested):
> Reviewed 5 of 5 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1520)* status: all files reviewed, 7 unresolved discussions (waiting on @jcorilla and @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/InternalStatusConversion.kt` line 43 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1520#-NtI_1Os9Xitxmq7IbYo:-NtI_1Ot1r12BjOxtPaA:b-qosjgb) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/29c01818f205d55da0e2f7b541bc327db7254908/src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/InternalStatusConversion.kt#L43)):*
> ```Kotlin
>   val errorInfo = internalApiException.errorInfo
> 
>   if (errorInfo == null || errorInfo.domain == ErrorCode.getDescriptor().fullName) {
> ```

Shouldn't this be reversed?



_Suggestion:_
```Kotlin
!=
```
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/InternalStatusConversion.kt` line 47 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1520#-NtIcGIu52k_Sk9ozqZv:-NtIcGIu52k_Sk9ozqZw:b5ql5xm) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/29c01818f205d55da0e2f7b541bc327db7254908/src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/InternalStatusConversion.kt#L47)):*
> ```Kotlin
>   }
>   var errorMessage = this.description ?: "Unknown exception."
>   val metadataMap = mutableMapOf<String, String>()
> ```

use `buildMap` to reduce the scope of the mutable map
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/InternalStatusConversion.kt` line 49 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1520#-NtIaznlEKoCs0DGQj27:-NtIaznmAJ8PDWurfgDb:byxcedb) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/29c01818f205d55da0e2f7b541bc327db7254908/src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/InternalStatusConversion.kt#L49)):*
> ```Kotlin
>   val metadataMap = mutableMapOf<String, String>()
>   when (errorInfo.reason) {
>     ErrorCode.MEASUREMENT_NOT_FOUND.toString() -> {
> ```

Parse the reason to an enum instead of comparing strings.



_Code quote:_
```Kotlin
.toString(
```
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/InternalStatusConversion.kt` line 51 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1520#-NtI_SrU2Z_HmCYtX6Jp:-NtI_FxuBGUUMsva7Jab:b-2t77ie) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/29c01818f205d55da0e2f7b541bc327db7254908/src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/InternalStatusConversion.kt#L51)):*
> ```Kotlin
>     ErrorCode.MEASUREMENT_NOT_FOUND.toString() -> {
>       if (
>         errorInfo.metadataMap.containsKey("external_measurement_consumer_id") &&
> ```

This should always be the case. The only valid subclass of MeasurementNotFoundException that's used for this API is 
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/InternalStatusConversion.kt` line 70 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1520#-NtIb9mO0lUT9GLms6EU:-NtIb9mPD_Zj6mpJ2DZ2:b-r8lb2s) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/29c01818f205d55da0e2f7b541bc327db7254908/src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/InternalStatusConversion.kt#L70)):*
> ```Kotlin
>       metadataMap["measurement_consumer"] =
>         MeasurementConsumerKey(
>             externalIdToApiId(errorInfo.metadataMap["external_measurement_consumer_id"]!!.toLong())
> ```

Non-null assertion (`!!`) is just for cases where you can guarantee that it's not null but the compiler can't figure it out for a smart cast. This is a case where we're just expecting the value not to be null, so it should use `checkNotNull` instead.



_Code quote:_
```Kotlin
!!
```
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/InternalStatusConversion.kt` line 112 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1520#-NtIbyqO0qtf8qGHMAJj:-NtIbyqO0qtf8qGHMAJk:bda4xea) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/29c01818f205d55da0e2f7b541bc327db7254908/src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/InternalStatusConversion.kt#L112)):*
> ```Kotlin
>         "Measurement ${metadataMap["measurement"]} is in illegal state: ${metadataMap["state"]}"
>     }
>     else -> {
> ```

Use an exhaustive `when` rather than having an else condition. Note that this requires using enums (see above)
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/InternalStatusConversion.kt` line 113 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1520#-NtIa7l26Ku0Wbjt5FV5:-NtIa7l26Ku0Wbjt5FV6:b-mx8gz1) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/29c01818f205d55da0e2f7b541bc327db7254908/src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/InternalStatusConversion.kt#L113)):*
> ```Kotlin
>     }
>     else -> {
>       metadataMap.putAll(errorInfo.metadataMap)
> ```

Won't this potentially leak IDs from the internal API?

jcorilla (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1520)* status: 10 of 12 files reviewed, 7 unresolved discussions (waiting on @SanjayVas and @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/InternalStatusConversion.kt` line 43 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1520#-NtI_1Os9Xitxmq7IbYo:-NtNDQUW81g5r9iMNfqI:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/29c01818f205d55da0e2f7b541bc327db7254908/src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/InternalStatusConversion.kt#L43)):*

Shouldn't this be reversed?
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/InternalStatusConversion.kt` line 47 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1520#-NtIcGIu52k_Sk9ozqZv:-NtNDR1Q4o736Tu3Kshj:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/29c01818f205d55da0e2f7b541bc327db7254908/src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/InternalStatusConversion.kt#L47)):*

use `buildMap` to reduce the scope of the mutable map
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/InternalStatusConversion.kt` line 51 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1520#-NtI_SrU2Z_HmCYtX6Jp:-NtNDSUU6CE_BHXeVD0-:bc13kxn) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/29c01818f205d55da0e2f7b541bc327db7254908/src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/InternalStatusConversion.kt#L51)):*

This should always be the case. The only valid subclass of MeasurementNotFoundException that's used for this API is
</blockquote></details>

Done. Turns out the cases I thought where this would be needed happen internally only. The parent class is thrown in MeasurementReader.kt
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/InternalStatusConversion.kt` line 70 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1520#-NtIb9mO0lUT9GLms6EU:-NtNECG_2yvl2YnHpc0-:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/29c01818f205d55da0e2f7b541bc327db7254908/src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/InternalStatusConversion.kt#L70)):*

Non-null assertion (`!!`) is just for cases where you can guarantee that it's not null but the compiler can't figure it out for a smart cast. This is a case where we're just expecting the value not to be null, so it should use `checkNotNull` instead.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/InternalStatusConversion.kt` line 112 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1520#-NtIbyqO0qtf8qGHMAJj:-NtNEENI8o2eHcxbspuc:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/29c01818f205d55da0e2f7b541bc327db7254908/src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/InternalStatusConversion.kt#L112)):*

Use an exhaustive `when` rather than having an else condition. Note that this requires using enums (see above)
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/InternalStatusConversion.kt` line 113 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1520#-NtIa7l26Ku0Wbjt5FV5:-NtNEFHBE7XtDCAMPFuU:blu2eqq) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/29c01818f205d55da0e2f7b541bc327db7254908/src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/InternalStatusConversion.kt#L113)):*

Won't this potentially leak IDs from the internal API?
</blockquote></details>

Done. Yes, since the initial use was only measurement service and its exceptions I had this here, but removing it.

jcorilla (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1520)* status: 10 of 12 files reviewed, 7 unresolved discussions (waiting on @SanjayVas and @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/InternalStatusConversion.kt` line 49 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1520#-NtIaznlEKoCs0DGQj27:-NtNEj3R05Fa4pOKACbz:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/29c01818f205d55da0e2f7b541bc327db7254908/src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/InternalStatusConversion.kt#L49)):*

Parse the reason to an enum instead of comparing strings.
</blockquote></details>

Done.

SanjayVas (changes requested):
> Reviewed 2 of 2 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1520)* status: all files reviewed, 5 unresolved discussions (waiting on @jcorilla and @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/InternalStatusConversion.kt` line 75 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1520#-NtNG5BbATZlWayeEPFI:-NtNG5Bc8JP-gUiJDWYs:bj42ade) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3a63b381a9543272de65bb688501c36108c2340b/src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/InternalStatusConversion.kt#L75)):*
> ```Kotlin
>               .toName(),
>           )
>           errorMessage = "Measurement Consumer ${get("measurement_consumer")} not found."
> ```

Extract the variable before putting it in the map so you can use that rather than having to do a map lookup



_Code quote:_
```Kotlin
${get("measurement_consumer")}
```
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/InternalStatusConversion.kt` line 75 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1520#-NtNGCs98YLxh0Tgf6Ya:-NtNGCs98YLxh0Tgf6Yb:b2utm05) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3a63b381a9543272de65bb688501c36108c2340b/src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/InternalStatusConversion.kt#L75)):*
> ```Kotlin
>               .toName(),
>           )
>           errorMessage = "Measurement Consumer ${get("measurement_consumer")} not found."
> ```

nit: the resource types are UpperCamelCase without spaces



_Suggestion:_
```Kotlin
MeasurementConsumer
```
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/InternalStatusConversion.kt` line 141 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1520#-NtNGczNBjsHTGD5YXvO:-NtNGczO5apiIgJm2gO1:bqwqq5z) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3a63b381a9543272de65bb688501c36108c2340b/src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/InternalStatusConversion.kt#L141)):*
> ```Kotlin
>         }
>         ErrorCode.COMPUTATION_PARTICIPANT_STATE_ILLEGAL -> {
>           errorMessage = "ComputationParticipant not in CREATED state."
> ```

The error code alone is insufficient to determine the expected state.



_Code quote:_
```Kotlin
not in CREATED state
```

jcorilla (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1520)* status: 11 of 12 files reviewed, 4 unresolved discussions (waiting on @SanjayVas and @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/InternalStatusConversion.kt` line 75 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1520#-NtNG5BbATZlWayeEPFI:-NtO5zd00wvIR1CMLhE1:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3a63b381a9543272de65bb688501c36108c2340b/src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/InternalStatusConversion.kt#L75)):*

Extract the variable before putting it in the map so you can use that rather than having to do a map lookup
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/InternalStatusConversion.kt` line 141 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1520#-NtNGczNBjsHTGD5YXvO:-NtO60GS9QhqcLsanpjf:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3a63b381a9543272de65bb688501c36108c2340b/src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/InternalStatusConversion.kt#L141)):*

The error code alone is insufficient to determine the expected state.
</blockquote></details>

Done.

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1520)* status: all files reviewed, 2 unresolved discussions (waiting on @jcorilla and @tristanvuong2021)

*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1520#-NtOAtCmDktflTxd0Kek:-NtOAtCmDktflTxd0Kel:b-uabqa2) (no related file):*
Add at least one test somewhere that makes sure this is working. For example, add a test in MeasurementServiceTest where you stub the internal service mock to throw an exception using KingdomInternalException.toStatusRuntimeException(). Use one of the exceptions that you know needs the ID format fixed. That will properly block this PR from being merged until that issue is fixed.

___
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/InternalStatusConversion.kt` line 48 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1520#-NtO9d6A7NXokna8gs5X:-NtO9d6A7NXokna8gs5Y:bvy0cdn) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4bfa731907591927d6b68a242780f9e67b3790f3/src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/InternalStatusConversion.kt#L48)):*
> ```Kotlin
>       when (ErrorCode.valueOf(errorInfo.reason)) {
>         ErrorCode.MEASUREMENT_NOT_FOUND -> {
>           val measurementKey =
> ```

This is the name, not the key (you're calling `toName()` on the key object).



_Code quote:_
```Kotlin
measurementKey
```

jcorilla (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1520)* status: 11 of 16 files reviewed, 2 unresolved discussions (waiting on @SanjayVas and @tristanvuong2021)

*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1520#-NtOAtCmDktflTxd0Kek:-NtaBF2i0VTeV-DUz03c:b-896fix) (no related file):*

Add at least one test somewhere that makes sure this is working. For example, add a test in MeasurementServiceTest where you stub the internal service mock to throw an exception using KingdomInternalException.toStatusRuntimeException(). Use one of the exceptions that you know needs the ID format fixed. That will properly block this PR from being merged until that issue is fixed.
</blockquote></details>

Done.

___
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/InternalStatusConversion.kt` line 48 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1520#-NtO9d6A7NXokna8gs5X:-NtaBKuf-U_HZ9V3ViRT:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4bfa731907591927d6b68a242780f9e67b3790f3/src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/InternalStatusConversion.kt#L48)):*

This is the name, not the key (you're calling `toName()` on the key object).
</blockquote></details>

Done.

SanjayVas (changes requested):
> Reviewed 6 of 6 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1520)* status: all files reviewed, 5 unresolved discussions (waiting on @jcorilla and @tristanvuong2021)

___
*[`src/test/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/MeasurementsServiceTest.kt` line 109 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1520#-NtbC7SJ9wN1Qwopz5H0:-NtbC7SJ9wN1Qwopz5H1:b-4cwi0c) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/50365a302b5a54bb8f66609b9217cda861d89a13/src/test/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/MeasurementsServiceTest.kt#L109)):*
> ```Kotlin
> import org.wfanet.measurement.kingdom.deploy.common.Llv2ProtocolConfig
> import org.wfanet.measurement.kingdom.deploy.common.RoLlv2ProtocolConfig
> import org.wfanet.measurement.kingdom.deploy.gcloud.spanner.common.*
> ```

Wildcard imports are forbidden.
___
*[`src/test/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/MeasurementsServiceTest.kt` line 2397 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1520#-NtbCzctBoQxi56hVSrh:-NtbCzctBoQxi56hVSri:b-nro1d2) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/50365a302b5a54bb8f66609b9217cda861d89a13/src/test/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/MeasurementsServiceTest.kt#L2397)):*
> ```Kotlin
>             .asStatusRuntimeException(Status.Code.NOT_FOUND, "Measurement not found.")
>         )
>     }
> ```

Use `stub` and `onBlocking` for stubbing suspend functions. See other examples in this file.



_Suggestion:_
```Kotlin
    internalMeasurementsMock.stub {
      onBlocking { getMeasurement(any()) }
          .thenThrow(
            MeasurementNotFoundByMeasurementConsumerException(
                ExternalId(EXTERNAL_MEASUREMENT_CONSUMER_ID),
                ExternalId(EXTERNAL_MEASUREMENT_ID),
              )
              .asStatusRuntimeException(Status.Code.NOT_FOUND, "Measurement not found.")
          )
    }
```
___
*[`src/test/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/MeasurementsServiceTest.kt` line 2465 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1520#-NtbCGHd3_C_E84v65wO:-NtbCGHe1vQXwklCgh4P:bdke2ka) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/50365a302b5a54bb8f66609b9217cda861d89a13/src/test/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/MeasurementsServiceTest.kt#L2465)):*
> ```Kotlin
>   @Test
>   fun `createMeasurement throws FAILED_PRECONDITION with duchy name when duchy not found`() {
>     val duchyID = DuchyCertificateKey.fromName(DUCHY_CERTIFICATE_NAME)!!.duchyId
> ```

Follow the camel-casing rules in the [style guide](https://developer.android.com/kotlin/style-guide#camel_case)



_Suggestion:_
```Kotlin
duchyId
```
___
*[`src/test/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/MeasurementsServiceTest.kt` line 2488 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1520#-NtbDeJZC0koduU7Z3EG:-NtbDeJZC0koduU7Z3EH:bvqxgp7) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/50365a302b5a54bb8f66609b9217cda861d89a13/src/test/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/MeasurementsServiceTest.kt#L2488)):*
> ```Kotlin
>       }
>     assertThat(exception.status.code).isEqualTo(Status.Code.FAILED_PRECONDITION)
>     assertThat(exception.errorInfo?.metadataMap).containsEntry("duchy", DuchyKey(duchyID).toName())
> ```

nit: rather than converting from the Duchy ID, just extract a constant from DUCHY_CERTIFICATE_NAME. e.g.

```kotlin
private const val DUCHY_NAME = "duchies/AAAAAAAAAHs"
private const val DUCHY_CERTIFICATE_NAME = "$duchyName/certificates/AAAAAAAAAHs"
```



_Code quote:_
```Kotlin
DuchyKey(duchyID).toName()
```
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/InternalStatusConversion.kt` line 48 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1520#-NtO9d6A7NXokna8gs5X:-NtbBY-92sTch4lg6xLD:b-wlh57k) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4bfa731907591927d6b68a242780f9e67b3790f3/src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/InternalStatusConversion.kt#L48)):*
<details><summary><i>Previously, jcorilla wrote…</i></summary><blockquote>

Done.
</blockquote></details>

This is still not quite right. It's not the name of the key, just the name. e.g. `measurementName`

The resource key represents the tuple of IDs, and the resource name is a path containing those ID segments as defined in [AIP-122](https://google.aip.dev/122)

jcorilla (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1520)* status: 13 of 16 files reviewed, 4 unresolved discussions (waiting on @SanjayVas and @tristanvuong2021)

___
*[`src/test/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/MeasurementsServiceTest.kt` line 109 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1520#-NtbC7SJ9wN1Qwopz5H0:-Ntbhkk677Kpvcl-aAUB:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/50365a302b5a54bb8f66609b9217cda861d89a13/src/test/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/MeasurementsServiceTest.kt#L109)):*

Wildcard imports are forbidden.
</blockquote></details>

Done.
___
*[`src/test/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/MeasurementsServiceTest.kt` line 2397 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1520#-NtbCzctBoQxi56hVSrh:-Ntbhizp3Ku6_zberV4C:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/50365a302b5a54bb8f66609b9217cda861d89a13/src/test/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/MeasurementsServiceTest.kt#L2397)):*

Use `stub` and `onBlocking` for stubbing suspend functions. See other examples in this file.
</blockquote></details>

Done.
___
*[`src/test/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/MeasurementsServiceTest.kt` line 2465 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1520#-NtbCGHd3_C_E84v65wO:-NtbJGqT7fpi80sKL4-D:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/50365a302b5a54bb8f66609b9217cda861d89a13/src/test/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/MeasurementsServiceTest.kt#L2465)):*

Follow the camel-casing rules in the [style guide](https://developer.android.com/kotlin/style-guide#camel_case)
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/InternalStatusConversion.kt` line 48 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1520#-NtO9d6A7NXokna8gs5X:-NtbIgetDgxQl0oyrjnC:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4bfa731907591927d6b68a242780f9e67b3790f3/src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/InternalStatusConversion.kt#L48)):*

This is still not quite right. It's not the name of the key, just the name. e.g. `measurementName`

The resource key represents the tuple of IDs, and the resource name is a path containing those ID segments as defined in [AIP-122](https://google.aip.dev/122)
</blockquote></details>

Done.

SanjayVas (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1520)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @tristanvuong2021)

Marco-Premier (approved):
> Reviewed 1 of 5 files at r1, 1 of 6 files at r2, 4 of 9 files at r3, 4 of 5 files at r4, 3 of 6 files at r7, 3 of 3 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1520)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @tristanvuong2021)

PR #1521, opened 2024-03-13
Change all Internal Reporting Measurements Service Methods responses to empty responses

tristanvuong2021:
These methods will always be internal only and the responses are never used. For performance reasons, it would be better for the responses to be empty and the data needed for the current responses to be not generated.


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1521)

tristanvuong2021 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1521)* status: 0 of 8 files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1521#-Nst9lvL4poUuNUK8FRF:-NstPNIDAKY0pcBZLHM5:bs5r3lb) (no related file):*

> These methods will always be internal only and the responses are never used.

This is not universally true, given that the public API methods do need to return the resources. For example, if you have an idempotent request using a request ID then the public API response cannot be predicted from the request. The other example is OUTPUT_ONLY fields, where you may need to read something from the DB.

It's possible to make this change for any method that does not have either of these conditions, but it's taking on future tech debt if we end up hitting one of them. Right now we can blanket transform internal service responses to public ones without having to consider whether we have the above conditions.
</blockquote></details>

In this specific case though, these methods will never have a public equivalent. There isn't even a public measurements service.

___

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1521)* status: 0 of 8 files reviewed, 1 unresolved discussion (waiting on @tristanvuong2021)

*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1521#-Nst9lvL4poUuNUK8FRF:-Nstbn736ow_uD1OExaH:b-kmkuxi) (no related file):*

In this specific case though, these methods will never have a public equivalent. There isn't even a public measurements service.
</blockquote></details>

Use google.protobuf.Empty explicitly rather than having empty response types then.

___

tristanvuong2021 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1521)* status: 0 of 9 files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1521#-Nst9lvL4poUuNUK8FRF:-NstrYHEESyg--mdKMdI:b-896fix) (no related file):*

Use google.protobuf.Empty explicitly rather than having empty response types then.
</blockquote></details>

Done.

___

SanjayVas (approved):
> Reviewed 4 of 8 files at r1, 5 of 5 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1521)* status: all files reviewed, 1 unresolved discussion (waiting on @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt` line 879 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1521#-NstsLRZ-3i2EDsinD9z:-NstsLR_-bEf6zP5tJY0:ba2ueo3) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3fd44cc996f8461055c4ccb8e5d1f1fd57fe88a1/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsService.kt#L879)):*
> ```Kotlin
>       failedMeasurementsList: List<Measurement>,
>       cmmsMeasurementConsumerId: String,
>     ): Empty {
> ```

nit: just drop the return value rather than passing Empty objects around



_Code quote:_
```Kotlin
: Empty
```

stevenwarejones (approved):
> Reviewed 4 of 8 files at r1, 4 of 5 files at r3, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1521)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @tristanvuong2021)

PR #1563, opened 2024-04-10
Implement VidIndexMap and add PopulationSpecValidator

kungfucraig:


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1563)

renjiezh (changes requested):
> Reviewed 6 of 9 files at r1, 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1563)* status: all files reviewed, 3 unresolved discussions (waiting on @kungfucraig, @ple13, and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/api/v2alpha/PopulationSpecValidator.kt` line 56 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1563#-Nv8Au_H97yMY-y9rsEq:-Nv8Au_H97yMY-y9rsEr:b97booh) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/2cd0d9c26b491726953b4070c016f2e76db5eba2/src/main/kotlin/org/wfanet/measurement/api/v2alpha/PopulationSpecValidator.kt#L56)):*
> ```Kotlin
>  * Indicates that the [VidRange.startVid] described by indexMessage is less than or equal to zero.
>  */
> data class StartVidLessThanOrEqualToZeroError(val indexMessage: String) :
> ```

nip: what about `StartVidNotPositiveError`?
___
*[`src/main/kotlin/org/wfanet/measurement/api/v2alpha/PopulationSpecValidator.kt` line 140 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1563#-Nv8EJhD70COMKzB3q9g:-Nv8EJhD70COMKzB3q9h:b-9scb56) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/2cd0d9c26b491726953b4070c016f2e76db5eba2/src/main/kotlin/org/wfanet/measurement/api/v2alpha/PopulationSpecValidator.kt#L140)):*
> ```Kotlin
>       }
>     }
>     return errorCount == validationErrorsInternal.size
> ```

A `PopulationSpecValidator` instance contains only one `populationSpec` and no `populationSpec` could be added into it . I thought it could be `return validationErrorsInternal.size <= 0` instead. Here it looks like reusing a `PopulationSpecValidator`. I am not sure about the purpose here.
Maybe refactor `isVidRangesListValid()` as `val isValid:Boolean by lazy {}`
___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/VidIndexMap.kt` line 76 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1563#-Nv85kzx1xtdW7lGsgRQ:-Nv85kzyEQlOTX4f1gDC:b-g79dly) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/2cd0d9c26b491726953b4070c016f2e76db5eba2/src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/VidIndexMap.kt#L76)):*
> ```Kotlin
>    */
>   operator fun get(vid: Long): Int =
>     if (indexMap.containsKey(vid)) {
> ```

To avoid `!!`, it can be 
```
indexMap.getOrElse(vid) { throw VidNotFoundException(vid) }
```

SanjayVas (changes requested):
> Reviewed 1 of 9 files at r1, 1 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1563)* status: all files reviewed, 9 unresolved discussions (waiting on @kungfucraig and @ple13)

___
*[`src/main/kotlin/org/wfanet/measurement/api/v2alpha/PopulationSpecValidator.kt` line 26 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1563#-Nv8OTQsAWSyfOd8e9B4:-Nv8OTQsAWSyfOd8e9B5:blyo7cg) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/2cd0d9c26b491726953b4070c016f2e76db5eba2/src/main/kotlin/org/wfanet/measurement/api/v2alpha/PopulationSpecValidator.kt#L26)):*
> ```Kotlin
> ) : Exception(message) {
>   override fun toString(): String {
>     val message = StringBuilder()
> ```

The idiomatic Kotlin way of doing this is to use [`buildString`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.text/build-string.html).
___
*[`src/main/kotlin/org/wfanet/measurement/api/v2alpha/PopulationSpecValidator.kt` line 27 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1563#-Nv8Oc7N0jsG5SN3hDD_:-Nv8Oc7N0jsG5SN3hDDa:undefined) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/2cd0d9c26b491726953b4070c016f2e76db5eba2/src/main/kotlin/org/wfanet/measurement/api/v2alpha/PopulationSpecValidator.kt#L27)):*
> ```Kotlin
>   override fun toString(): String {
>     val message = StringBuilder()
>     message.append(super.toString() + "\n")
> ```





_Suggestion:_
```Kotlin
appendLine(super.toString())
```
___
*[`src/main/kotlin/org/wfanet/measurement/api/v2alpha/PopulationSpecValidator.kt` line 44 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1563#-Nv8QkXJB83OWAU1ZEvO:-Nv8QkXJB83OWAU1ZEvP:bk0362a) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/2cd0d9c26b491726953b4070c016f2e76db5eba2/src/main/kotlin/org/wfanet/measurement/api/v2alpha/PopulationSpecValidator.kt#L44)):*
> ```Kotlin
>  * @param [secondIndexMessage] describe the index of the second range in a [PopulationSpec]
>  */
> data class VidRangesNotDisjointError(
> ```

"Error" classes in JVM languages are ones that inherit from [`java.lang.Error`](https://docs.oracle.com/javase%2F7%2Fdocs%2Fapi%2F%2F/java/lang/Error.html), similar to how "Exception" classes are those that inherit from `java.lang.Exception`.



_Code quote:_
```Kotlin
Error
```
___
*[`src/main/kotlin/org/wfanet/measurement/api/v2alpha/PopulationSpecValidator.kt` line 74 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1563#-Nv8OuSLBoSZ1sOKO_hF:-Nv8OuSLBoSZ1sOKO_hG:b-1n58np) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/2cd0d9c26b491726953b4070c016f2e76db5eba2/src/main/kotlin/org/wfanet/measurement/api/v2alpha/PopulationSpecValidator.kt#L74)):*
> ```Kotlin
> }
> 
> class PopulationSpecValidator(val populationSpec: PopulationSpec) {
> ```

This looks like it can be private



_Suggestion:_
```Kotlin
private val populationSpec: PopulationSpec
```
___
*[`src/main/kotlin/org/wfanet/measurement/api/v2alpha/PopulationSpecValidator.kt` line 76 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1563#-Nv8RY9-7uhZw0VxxeO6:-Nv8RY9-7uhZw0VxxeO7:b1ofvio) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/2cd0d9c26b491726953b4070c016f2e76db5eba2/src/main/kotlin/org/wfanet/measurement/api/v2alpha/PopulationSpecValidator.kt#L76)):*
> ```Kotlin
> class PopulationSpecValidator(val populationSpec: PopulationSpec) {
>   /**
>    * The list of errors accumulated from making calls to [this] validator. This list is never reset.
> ```

This design has unnecessary mutable state. I would expect a validator to have `validateX` methods that throw an exception (or return a collection of error messages/objects if trying to avoid exceptions). This design results in weird behavior, such as having multiple calls to the validation methods accumulating additional errors.
___
*[`src/main/kotlin/org/wfanet/measurement/api/v2alpha/PopulationSpecValidator.kt` line 79 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1563#-Nv8PLmrABCXJXe_t2oL:-Nv8PLmsCaWlmPNAaYG6:bq1nz28) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/2cd0d9c26b491726953b4070c016f2e76db5eba2/src/main/kotlin/org/wfanet/measurement/api/v2alpha/PopulationSpecValidator.kt#L79)):*
> ```Kotlin
>    */
>   val validationErrors: List<PopulationSpecValidationError>
>     get() = Collections.unmodifiableList(validationErrorsInternal)
> ```

Unlike in Java, Kotlin has separate interfaces for mutable collections which inherit from the interfaces for unmodifiable collections. Therefore the common practice here is to just return the `MutableList` using the `List` interface (and not worry about whether the caller might try to cast it back to a MutableList).

(Google-internal) advice can be found at go/kotlin-practices/collections



_Suggestion:_
```Kotlin
 validationErrorsInternal
```

kungfucraig (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1563)* status: 6 of 9 files reviewed, 9 unresolved discussions (waiting on @ple13, @renjiezh, and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/api/v2alpha/PopulationSpecValidator.kt` line 26 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1563#-Nv8OTQsAWSyfOd8e9B4:-Nv8e2fD7-xNHg5fLq1p:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/2cd0d9c26b491726953b4070c016f2e76db5eba2/src/main/kotlin/org/wfanet/measurement/api/v2alpha/PopulationSpecValidator.kt#L26)):*

The idiomatic Kotlin way of doing this is to use [`buildString`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.text/build-string.html).
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/api/v2alpha/PopulationSpecValidator.kt` line 27 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1563#-Nv8Oc7N0jsG5SN3hDD_:-Nv8e39i2rlJ6scC5irp:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/2cd0d9c26b491726953b4070c016f2e76db5eba2/src/main/kotlin/org/wfanet/measurement/api/v2alpha/PopulationSpecValidator.kt#L27)):*
> ```Kotlin
>   override fun toString(): String {
>     val message = StringBuilder()
>     message.append(super.toString() + "\n")
> ```

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/api/v2alpha/PopulationSpecValidator.kt` line 44 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1563#-Nv8QkXJB83OWAU1ZEvO:-Nv8effo9nixtmp6N5V-:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/2cd0d9c26b491726953b4070c016f2e76db5eba2/src/main/kotlin/org/wfanet/measurement/api/v2alpha/PopulationSpecValidator.kt#L44)):*

"Error" classes in JVM languages are ones that inherit from [`java.lang.Error`](https://docs.oracle.com/javase%2F7%2Fdocs%2Fapi%2F%2F/java/lang/Error.html), similar to how "Exception" classes are those that inherit from `java.lang.Exception`.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/api/v2alpha/PopulationSpecValidator.kt` line 56 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1563#-Nv8Au_H97yMY-y9rsEq:-Nv8eguS4HiWgPveijzq:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/2cd0d9c26b491726953b4070c016f2e76db5eba2/src/main/kotlin/org/wfanet/measurement/api/v2alpha/PopulationSpecValidator.kt#L56)):*

nip: what about `StartVidNotPositiveError`?
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/api/v2alpha/PopulationSpecValidator.kt` line 74 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1563#-Nv8OuSLBoSZ1sOKO_hF:-Nv8erwd01V58snu7Tml:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/2cd0d9c26b491726953b4070c016f2e76db5eba2/src/main/kotlin/org/wfanet/measurement/api/v2alpha/PopulationSpecValidator.kt#L74)):*

This looks like it can be private
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/api/v2alpha/PopulationSpecValidator.kt` line 76 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1563#-Nv8RY9-7uhZw0VxxeO6:-Nv9GAGX4i438MGs-GzN:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/2cd0d9c26b491726953b4070c016f2e76db5eba2/src/main/kotlin/org/wfanet/measurement/api/v2alpha/PopulationSpecValidator.kt#L76)):*

This design has unnecessary mutable state. I would expect a validator to have `validateX` methods that throw an exception (or return a collection of error messages/objects if trying to avoid exceptions). This design results in weird behavior, such as having multiple calls to the validation methods accumulating additional errors.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/api/v2alpha/PopulationSpecValidator.kt` line 79 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1563#-Nv8PLmrABCXJXe_t2oL:-Nv9GB1S-VOs0S375aw8:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/2cd0d9c26b491726953b4070c016f2e76db5eba2/src/main/kotlin/org/wfanet/measurement/api/v2alpha/PopulationSpecValidator.kt#L79)):*

Unlike in Java, Kotlin has separate interfaces for mutable collections which inherit from the interfaces for unmodifiable collections. Therefore the common practice here is to just return the `MutableList` using the `List` interface (and not worry about whether the caller might try to cast it back to a MutableList).

(Google-internal) advice can be found at go/kotlin-practices/collections
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/api/v2alpha/PopulationSpecValidator.kt` line 140 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1563#-Nv8EJhD70COMKzB3q9g:-Nv8SDuPDiAkuG2wybHz:b39cgkw) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/2cd0d9c26b491726953b4070c016f2e76db5eba2/src/main/kotlin/org/wfanet/measurement/api/v2alpha/PopulationSpecValidator.kt#L140)):*

A `PopulationSpecValidator` instance contains only one `populationSpec` and no `populationSpec` could be added into it . I thought it could be `return validationErrorsInternal.size <= 0` instead. Here it looks like reusing a `PopulationSpecValidator`. I am not sure about the purpose here.
Maybe refactor `isVidRangesListValid()` as `val isValid:Boolean by lazy {}`
</blockquote></details>

Things have been substantially changed. PTAL.
___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/VidIndexMap.kt` line 76 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1563#-Nv85kzx1xtdW7lGsgRQ:-Nv9GSgzD7s8eCxVdgLK:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/2cd0d9c26b491726953b4070c016f2e76db5eba2/src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/VidIndexMap.kt#L76)):*

To avoid `!!`, it can be 
```
indexMap.getOrElse(vid) { throw VidNotFoundException(vid) }
```
</blockquote></details>

Done.

SanjayVas (changes requested):
> Reviewed 1 of 9 files at r1, 2 of 3 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1563)* status: 8 of 9 files reviewed, 7 unresolved discussions (waiting on @kungfucraig, @ple13, and @renjiezh)

___
*[`src/main/kotlin/org/wfanet/measurement/api/v2alpha/PopulationSpecValidator.kt` line 44 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1563#-Nv8QkXJB83OWAU1ZEvO:-Nv9UF5f2AxAfw4fbgem:b-fmgc15) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/2cd0d9c26b491726953b4070c016f2e76db5eba2/src/main/kotlin/org/wfanet/measurement/api/v2alpha/PopulationSpecValidator.kt#L44)):*

Done.
</blockquote></details>

Sorry, what I meant was not to call these "Error" since they are not intended to be `Throwable`s that indicate a serious issue that should not be caught and likely crash the application. Name them "Exception" instead of they are throwable or something else if not.
___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/VidIndexMap.kt` line 76 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1563#-Nv85kzx1xtdW7lGsgRQ:-Nv9Vn2cACKHM1J8JQiV:b-lk2ws4) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/2cd0d9c26b491726953b4070c016f2e76db5eba2/src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/VidIndexMap.kt#L76)):*

Done.
</blockquote></details>

In this case, the "elvis operator" would be better.

```kotlin
indexMap[vid] ?: throw VidNotFoundException(vid)
```
___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/VidIndexMap.kt` line 48 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1563#-Nv9Z6h07uTjgTXOvGp_:-Nv9Z6h07uTjgTXOvGpa:b-g4io2m) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a09809f3ca00d7286f1585e0848efe35313fc172/src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/VidIndexMap.kt#L48)):*
> ```Kotlin
> 
>   /** A map of a VID to its index in the [Frequency Vector]. */
>   private val indexMap = hashMapOf<Long, Int>()
> ```

Does the implementation here matter? i.e. do you specifically want a HashMap? Or would you instead want any MutableMap that  maintains iteration order? If the latter, then consider `mutableMapOf`



_Code quote:_
```Kotlin
hashMapOf
```
___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/VidIndexMap.kt` line 53 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1563#-Nv9ZniM5-nbPYYhjZeF:-Nv9ZniM5-nbPYYhjZeG:be4fmqp) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a09809f3ca00d7286f1585e0848efe35313fc172/src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/VidIndexMap.kt#L53)):*
> ```Kotlin
>     validateVidRangesList(populationSpec).getOrThrow()
> 
>     val hashes = mutableListOf<Pair<Long, BigInteger>>()
> ```

Just to clarify, this is a list of Pairs of VIDs and their hashes? A domain-specific data class would be preferred over Pair.
___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/VidIndexMap.kt` line 79 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1563#-Nv9WeY87zebblLIUJu_:-Nv9WeY87zebblLIUJua:b-o5siei) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a09809f3ca00d7286f1585e0848efe35313fc172/src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/VidIndexMap.kt#L79)):*
> ```Kotlin
> 
>   companion object {
>     /** Hash a VID with SHA256 */
> ```

nit: maybe be more explicit about how this serializes, i.e. that it appends `[salt]` to the big endian 64-bit representation of `[vid]`

As an aside that's likely not relevant if VIDs are guaranteed to be positive, the binary representation of a (signed) Long is in two's complement

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1563)* status: 8 of 9 files reviewed, 5 unresolved discussions (waiting on @kungfucraig and @ple13)

___
*[`src/main/kotlin/org/wfanet/measurement/api/v2alpha/PopulationSpecValidator.kt` line 140 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1563#-Nv8EJhD70COMKzB3q9g:-Nv9wUORB-FEIbOO71TY:b-ji0kpw) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/2cd0d9c26b491726953b4070c016f2e76db5eba2/src/main/kotlin/org/wfanet/measurement/api/v2alpha/PopulationSpecValidator.kt#L140)):*

Things have been substantially changed. PTAL.
</blockquote></details>

nit: might use if-else instead of when.
___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/VidIndexMap.kt` line 76 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1563#-Nv85kzx1xtdW7lGsgRQ:-Nv9x7jr10ClRs-IL4cW:b-ebclnm) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/2cd0d9c26b491726953b4070c016f2e76db5eba2/src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/VidIndexMap.kt#L76)):*

In this case, the "elvis operator" would be better.

```kotlin
indexMap[vid] ?: throw VidNotFoundException(vid)
```
</blockquote></details>

yeah, sanjay's suggestion is better.

kungfucraig (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1563)* status: 8 of 9 files reviewed, 4 unresolved discussions (waiting on @ple13, @renjiezh, and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/api/v2alpha/PopulationSpecValidator.kt` line 44 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1563#-Nv8QkXJB83OWAU1ZEvO:-NvCk1knEWXskkk-HahD:b4dktxe) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/2cd0d9c26b491726953b4070c016f2e76db5eba2/src/main/kotlin/org/wfanet/measurement/api/v2alpha/PopulationSpecValidator.kt#L44)):*

Sorry, what I meant was not to call these "Error" since they are not intended to be `Throwable`s that indicate a serious issue that should not be caught and likely crash the application. Name them "Exception" instead of they are throwable or something else if not.
</blockquote></details>

Nested them inside the exception and called them "details."
___
*[`src/main/kotlin/org/wfanet/measurement/api/v2alpha/PopulationSpecValidator.kt` line 140 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1563#-Nv8EJhD70COMKzB3q9g:-NvCiKSR5kjImmwe9I0v:b-jqda0i) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/2cd0d9c26b491726953b4070c016f2e76db5eba2/src/main/kotlin/org/wfanet/measurement/api/v2alpha/PopulationSpecValidator.kt#L140)):*

nit: might use if-else instead of when.
</blockquote></details>

Doesn't seem like a big deal one way or another.
___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/VidIndexMap.kt` line 48 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1563#-Nv9Z6h07uTjgTXOvGp_:-NvCh-TS-UMOePEC5hjK:b-c33keb) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a09809f3ca00d7286f1585e0848efe35313fc172/src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/VidIndexMap.kt#L48)):*

Does the implementation here matter? i.e. do you specifically want a HashMap? Or would you instead want any MutableMap that  maintains iteration order? If the latter, then consider `mutableMapOf`
</blockquote></details>

I distinctly do not want a TreeMap as it has higher memory overhead and higher access times. Of course, mutableMap is probably a hash map, but we ought not depend on implementation details.

Sadly, the Kotlin documentation does not explicitly talk about the performance required of the thing that is returned by "mutableMapOf."

I'm just going to leave this as is.
___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/VidIndexMap.kt` line 53 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1563#-Nv9ZniM5-nbPYYhjZeF:-Nv9bO4RAsQIn8a9zqho:b-4wjqq6) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a09809f3ca00d7286f1585e0848efe35313fc172/src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/VidIndexMap.kt#L53)):*

Just to clarify, this is a list of Pairs of VIDs and their hashes? A domain-specific data class would be preferred over Pair.
</blockquote></details>

This construct  is only used within this function. We should stick with Pair and avoid the bloat.

renjiezh (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1563)* status: all files reviewed, 2 unresolved discussions (waiting on @ple13 and @SanjayVas)

stevenwarejones (changes requested):
> Reviewed 5 of 9 files at r1, 1 of 3 files at r2, 3 of 3 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1563)* status: all files reviewed, 10 unresolved discussions (waiting on @kungfucraig and @ple13)

___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/VidIndexMap.kt` line 15 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1563#-NvIk2cx8_oZVswnpXmM:-NvIk2cx8_oZVswnpXmN:b1fqftx) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8218633d3cbe11c3b2de243f99d32dd8691f4a26/src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/VidIndexMap.kt#L15)):*
> ```Kotlin
> 
> package org.wfanet.measurement.eventdataprovider.shareshuffle
> ```

what happened to the path of this file?
`src/main/kotlin/org/wfanet/measurement/evendataprovi/shareshuffle/VidIndexMap.kt`?
___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/VidIndexMap.kt` line 57 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1563#-NvIjvMt-zIQoBg2KzWO:-NvIjvMt-zIQoBg2KzWP:b-g7kiln) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8218633d3cbe11c3b2de243f99d32dd8691f4a26/src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/VidIndexMap.kt#L57)):*
> ```Kotlin
>     for (subPop in populationSpec.subpopulationsList) {
>       for (range in subPop.vidRangesList) {
>         for (vid in range.startVid..range.endVidInclusive) {
> ```

how fast does this process in practice given a vid range of a few billion?
___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/VidIndexMap.kt` line 64 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1563#-NvIktOtC_MkfHkpn06e:-NvIktOtC_MkfHkpn06f:b-sfytzy) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8218633d3cbe11c3b2de243f99d32dd8691f4a26/src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/VidIndexMap.kt#L64)):*
> ```Kotlin
>     }
> 
>     hashes.sortWith(compareBy<Pair<Long, BigInteger>> { it.second }.thenBy { it.first })
> ```

how well does this scale up? Can you remind me why we need to have this an in-memory hash map?

kungfucraig (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1563)* status: all files reviewed, 8 unresolved discussions (waiting on @ple13, @SanjayVas, and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/api/v2alpha/PopulationSpecValidator.kt` line 22 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1563#-NvDPQ2dB2R5ZMMHAM03:-NvJuPHGBlbmQALDhZxL:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8218633d3cbe11c3b2de243f99d32dd8691f4a26/src/main/kotlin/org/wfanet/measurement/api/v2alpha/PopulationSpecValidator.kt#L22)):*

Note that callers may access the `message` property rather than calling `toString()`. You may want to instead build this as the message passed to the parent constructor.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/VidIndexMap.kt` line 15 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1563#-NvIk2cx8_oZVswnpXmM:-NvJsGql1S4iJwRfjTtn:b-avrnrg) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8218633d3cbe11c3b2de243f99d32dd8691f4a26/src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/VidIndexMap.kt#L15)):*

what happened to the path of this file?
`src/main/kotlin/org/wfanet/measurement/evendataprovi/shareshuffle/VidIndexMap.kt`?
</blockquote></details>

I'm not sure what you mean.
___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/VidIndexMap.kt` line 53 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1563#-NvDLq74B4p6jFP1R4xr:-NvJsK_51vaQrJ6we_Ig:b-4tddbh) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8218633d3cbe11c3b2de243f99d32dd8691f4a26/src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/VidIndexMap.kt#L53)):*

nit: add a comment indicating what this is a map of. I assume the conversion of the hash bytes to BigInteger is for sortability? Might be good to note that too.
</blockquote></details>

The BigInteger thing was bad. Updated.
___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/VidIndexMap.kt` line 57 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1563#-NvIjvMt-zIQoBg2KzWO:-NvJk-TS4Y3CZCp-yaeL:b-qmxfok) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8218633d3cbe11c3b2de243f99d32dd8691f4a26/src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/VidIndexMap.kt#L57)):*

how fast does this process in practice given a vid range of a few billion?
</blockquote></details>

These VidIndexMaps are meant to be built and held onto. Another option is to offload the VidIndexMap to a table and do a join. It sort of depends on what your fulfillment pipeline looks like.

10M takes about 12s under JUnit. I can do some more performance testing after this is merged. I think we have some other options for reducing space and time.

For example, we could do something like this instead. https://sourcegraph.com/github.com/world-federation-of-advertisers/any-sketch-java@3bd551646af86f7cc9ea7f57fc4941fcd023ac4e/-/blob/src/main/java/org/wfanet/sampling/VidSampler.java?L72

Happily it's all implementation detail until we release.
___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/VidIndexMap.kt` line 58 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1563#-NvDMItV-kMdqNf6XxPC:-NvJtdNJ5fQXCTP8Z0dN:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8218633d3cbe11c3b2de243f99d32dd8691f4a26/src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/VidIndexMap.kt#L58)):*

This BigInteger constructor gives you the sign-magnitude representation. Are you sure you want that rather than interpreting the bytes as a big endian two's complement integer? The latter would make it so you're using the same representation to encode the VID in the hash as you do to encode the hash as an integer.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/VidIndexMap.kt` line 64 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1563#-NvIktOtC_MkfHkpn06e:-NvJtewOD6DT7AFeskYC:bs8bwht) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8218633d3cbe11c3b2de243f99d32dd8691f4a26/src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/VidIndexMap.kt#L64)):*

how well does this scale up? Can you remind me why we need to have this an in-memory hash map?
</blockquote></details>

Let's use the above comment to have this discussion.
___
*[`src/test/kotlin/org/wfanet/measurement/api/v2alpha/PopulationSpecValidatorTest.kt` line 41 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1563#-NvDNCvgBOyhXuvkqDKz:-NvJYusL1DslGFIyjfZV:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8218633d3cbe11c3b2de243f99d32dd8691f4a26/src/test/kotlin/org/wfanet/measurement/api/v2alpha/PopulationSpecValidatorTest.kt#L41)):*

We generally don't want to do equality assertions on the entire message as we want to avoid them being considered part of the API contract. Tests against the entire message become change-detector tests for the message. Instead we only do the most narrow assertion possible if we need to disambiguate exceptions.

e.g. assertThat(exception.details.first().toString()).contains("endVidExclusive")

That said, given that the `Detail` objects are a public part of the interface you can check which Detail objects you got back in order to disambiguate.
</blockquote></details>

Done.
___
*[`src/test/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/VidIndexMapTest.kt` line 98 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1563#-NvDOYMoCpzjniJGPw61:-NvJ_VhZ0y4nCaBa1MW8:b-cz97gc) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8218633d3cbe11c3b2de243f99d32dd8691f4a26/src/test/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/VidIndexMapTest.kt#L98)):*

Doing an assertion inside of a passed function object means that the assertion would get skipped if the function is never called. Instead, capture the arguments and assert on the captured value instead.
</blockquote></details>

It's inconceivable that the last test would pass without this function being called.

I took a slightly different approach.

kungfucraig (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1563)* status: 4 of 9 files reviewed, 8 unresolved discussions (waiting on @ple13, @renjiezh, @SanjayVas, and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/VidIndexMap.kt` line 57 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1563#-NvIjvMt-zIQoBg2KzWO:-NvNHbmU2qa2jHG3w8uR:b-8qb6wo) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8218633d3cbe11c3b2de243f99d32dd8691f4a26/src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/VidIndexMap.kt#L57)):*

These VidIndexMaps are meant to be built and held onto. Another option is to offload the VidIndexMap to a table and do a join. It sort of depends on what your fulfillment pipeline looks like.

10M takes about 12s under JUnit. I can do some more performance testing after this is merged. I think we have some other options for reducing space and time.

For example, we could do something like this instead. https://sourcegraph.com/github.com/world-federation-of-advertisers/any-sketch-java@3bd551646af86f7cc9ea7f57fc4941fcd023ac4e/-/blob/src/main/java/org/wfanet/sampling/VidSampler.java?L72

Happily it's all implementation detail until we release.
</blockquote></details>

As of now, I think we'll update this class to have a function to write the map to a file and to read it from a file too. That should provide some additional flexibility for those who need it.

SanjayVas (changes requested):
> Reviewed 1 of 3 files at r2, 5 of 5 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1563)* status: all files reviewed, 4 unresolved discussions (waiting on @kungfucraig, @ple13, and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/VidIndexMap.kt` line 94 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1563#-NvXJHPZEc4xDfbzim5e:-NvXIjpkBQdDWM1cAO52:b-ps4l21) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/2f8e7a4b0355b2984f53037fa9605802d2f3f891/src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/VidIndexMap.kt#L94)):*
> ```Kotlin
>         .asReadOnlyByteBuffer()
>         .order(ByteOrder.BIG_ENDIAN)
>         .getLong(0)
> ```

This is taking the first 64 bits of a 256-bit hash output. I would expect that we'd either want to use the full 256 bits or use a 64-bit hash. If this is intentional, document it as such.

kungfucraig (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1563)* status: 8 of 9 files reviewed, 4 unresolved discussions (waiting on @ple13, @SanjayVas, and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/VidIndexMap.kt` line 94 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1563#-NvXJHPZEc4xDfbzim5e:-NvXTkRO8srloEOLCs0V:b-3jcon5) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/2f8e7a4b0355b2984f53037fa9605802d2f3f891/src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/VidIndexMap.kt#L94)):*

This is taking the first 64 bits of a 256-bit hash output. I would expect that we'd either want to use the full 256 bits or use a 64-bit hash. If this is intentional, document it as such.
</blockquote></details>

It's intentional. Docs are updated.

ple13 (approved):
> Reviewed 4 of 9 files at r1, 1 of 3 files at r2, 4 of 5 files at r5, 1 of 1 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1563)* status: all files reviewed, 4 unresolved discussions (waiting on @SanjayVas and @stevenwarejones)

stevenwarejones (changes requested):
> Reviewed 1 of 5 files at r5, 1 of 1 files at r6.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1563)* status: all files reviewed, 4 unresolved discussions (waiting on @kungfucraig and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/VidIndexMap.kt` line 15 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1563#-NvIk2cx8_oZVswnpXmM:-NvXqIYsCTGltr_7bxM8:bl2kz2u) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8218633d3cbe11c3b2de243f99d32dd8691f4a26/src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/VidIndexMap.kt#L15)):*

I'm not sure what you mean.
</blockquote></details>

this looks fixed now. Not sure what was happening earlier - maybe a reviewable bug.
___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/VidIndexMap.kt` line 57 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1563#-NvIjvMt-zIQoBg2KzWO:-NvXqZDyCSDntdJpSsCv:bvvbdwh) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8218633d3cbe11c3b2de243f99d32dd8691f4a26/src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/VidIndexMap.kt#L57)):*

As of now, I think we'll update this class to have a function to write the map to a file and to read it from a file too. That should provide some additional flexibility for those who need it.
</blockquote></details>

So, I'd actually call this file as it currently is `InMemoryVidIndexMap` and it should be marked test only. 

Going back to my earlier comment, it may be helpful to make an interface for `VidIndexMap` that doesn't assume any particular underlying back-end or storage method.

We can make another implementation in the future that assumes a postgres backend or a PCollection or something like that.
___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/VidIndexMap.kt` line 94 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1563#-NvXJHPZEc4xDfbzim5e:-NvXrP2d861Oky4GhhRk:b48tbuu) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/2f8e7a4b0355b2984f53037fa9605802d2f3f891/src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/VidIndexMap.kt#L94)):*

It's intentional. Docs are updated.
</blockquote></details>

a 64 bit space will definitely have collisions. Is the reduced hash size intended to help with performance? Have you looked at other hash space sizes as well?
___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/VidIndexMap.kt` line 90 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1563#-NvXzA-30lb3YHMLWJtM:-NvXzA-30lb3YHMLWJtN:b-7jlnh0) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/b8f8f85cf32e8b67b84abc75d345d31bbf62f68a/src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/VidIndexMap.kt#L90)):*
> ```Kotlin
>      * @returns The first 64-bits of the hash as a [Long]
>      */
>     fun hashVidSha256(vid: Long, salt: ByteString = ByteString.EMPTY): Long {
> ```

to Sanjay's point, this is a poor name.  hashVid64 or hashVidToLong. I feel like we had this exact same function in panel exchange at some point but I can't find it. Maybe it was factored out.

SanjayVas (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1563)* status: all files reviewed, 3 unresolved discussions (waiting on @kungfucraig and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/VidIndexMap.kt` line 94 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1563#-NvXJHPZEc4xDfbzim5e:-NvY1lcNDsx_SLV62UWd:b-z7d99z) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/2f8e7a4b0355b2984f53037fa9605802d2f3f891/src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/VidIndexMap.kt#L94)):*

a 64 bit space will definitely have collisions. Is the reduced hash size intended to help with performance? Have you looked at other hash space sizes as well?
</blockquote></details>

Approving as this is stated to be intentional, but I am somewhat surprised as I'd expect to use all of the hash bits. Otherwise why use SHA-256 over e.g. Farmhash fingerprint64

kungfucraig (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1563)* status: 7 of 9 files reviewed, 2 unresolved discussions (waiting on @ple13, @SanjayVas, and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/VidIndexMap.kt` line 57 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1563#-NvIjvMt-zIQoBg2KzWO:-NvYUv5g8vNWcE8hfMqp:b-ilqk14) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8218633d3cbe11c3b2de243f99d32dd8691f4a26/src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/VidIndexMap.kt#L57)):*

So, I'd actually call this file as it currently is `InMemoryVidIndexMap` and it should be marked test only. 

Going back to my earlier comment, it may be helpful to make an interface for `VidIndexMap` that doesn't assume any particular underlying back-end or storage method.

We can make another implementation in the future that assumes a postgres backend or a PCollection or something like that.
</blockquote></details>

Using InMemoryVidIndexMap for non-test scenarios is perfectly reasonable depending upon how you're architected. Under JUnit building a map for 55M takes about 50 seconds. Obviously this is not something you want to do for each fulfillment though. For Farmhash it took about 30s. Reading from a file should be faster as we can preclude the need to sort. I wouldn't worry about the memory.

I added the interface so we have the insulation in case we need it. We can consider moving things to other files if/when we end up with another implementation. 
___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/VidIndexMap.kt` line 94 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1563#-NvXJHPZEc4xDfbzim5e:-NvYT9MI5ezeVV13IujV:b-qtamq7) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/2f8e7a4b0355b2984f53037fa9605802d2f3f891/src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/VidIndexMap.kt#L94)):*

Approving as this is stated to be intentional, but I am somewhat surprised as I'd expect to use all of the hash bits. Otherwise why use SHA-256 over e.g. Farmhash fingerprint64
</blockquote></details>

Reduced size is intended to help with performance, which I did some ad hoc testing on. I think I'd like to come back to this after we get the initial implementation of the library in place, and before we officially release it.

That said, having a collision is not a huge problem as we have the secondary sorting criteria of the VID itself. At least as long as there are not too many collisions, which with a 64-bit space should be acceptable. I can do the math if folks are interested, but the existing VID Sampler uses a 32-bit space.

I had written, "I'm considering a move to Farm," in my previous comment but deleted it. :)
___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/VidIndexMap.kt` line 90 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1563#-NvXzA-30lb3YHMLWJtM:-NvYVvUD6U3fipgi9X0e:bq6jxs4) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/b8f8f85cf32e8b67b84abc75d345d31bbf62f68a/src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/VidIndexMap.kt#L90)):*

to Sanjay's point, this is a poor name.  hashVid64 or hashVidToLong. I feel like we had this exact same function in panel exchange at some point but I can't find it. Maybe it was factored out.
</blockquote></details>

I renamed it, but we're not building a general purpose library for hashing. It's an implementation detail of this class that is only visible because I wanted a way to test this class using a fake hash function.

stevenwarejones (changes requested):
> Reviewed 2 of 5 files at r5, 3 of 3 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1563)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @kungfucraig)

___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/VidIndexMap.kt` line 94 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1563#-NvXJHPZEc4xDfbzim5e:-NvcdHMTAMoE3oOCP7Ta:b-aszohw) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/2f8e7a4b0355b2984f53037fa9605802d2f3f891/src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/VidIndexMap.kt#L94)):*

Reduced size is intended to help with performance, which I did some ad hoc testing on. I think I'd like to come back to this after we get the initial implementation of the library in place, and before we officially release it.

That said, having a collision is not a huge problem as we have the secondary sorting criteria of the VID itself. At least as long as there are not too many collisions, which with a 64-bit space should be acceptable. I can do the math if folks are interested, but the existing VID Sampler uses a 32-bit space.

I had written, "I'm considering a move to Farm," in my previous comment but deleted it. :)
</blockquote></details>

why not just use Farmhash fingerprint64? what's the negative of doing that?

kungfucraig (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1563)* status: 7 of 10 files reviewed, all discussions resolved (waiting on @ple13, @renjiezh, @SanjayVas, and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/VidIndexMap.kt` line 94 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1563#-NvXJHPZEc4xDfbzim5e:-Nvie89VED0V7M_86eNW:b-w2bx8g) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/2f8e7a4b0355b2984f53037fa9605802d2f3f891/src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/VidIndexMap.kt#L94)):*

why not just use Farmhash fingerprint64? what's the negative of doing that?
</blockquote></details>

Okay, why not.

stevenwarejones (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1563)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @kungfucraig)

___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/VidIndexMap.kt` line 103 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1563#-NvlP6Pi-yo2j3zVeECs:-NvlP6Pi-yo2j3zVeECt:b-lm2eea) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/2b21fb6cd848986c9f796808f9cc96867785a0cf/src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/VidIndexMap.kt#L103)):*
> ```Kotlin
>      */
>     fun hashVidToLongWithFarmHash(vid: Long, salt: Long): Long {
>       val hasher = Hashing.farmHashFingerprint64().newHasher()
> ```

slightly shorter
```
return Hashing.farmHashFingerprint64().hashLong(vid+salt)
```

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1563)* status: all files reviewed, 1 unresolved discussion (waiting on @kungfucraig)

___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/VidIndexMap.kt` line 100 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1563#-NvmDKuv2jnvDptPH3k-:-NvmDKuv2jnvDptPH3k0:b-yie14n) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/2b21fb6cd848986c9f796808f9cc96867785a0cf/src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/VidIndexMap.kt#L100)):*
> ```Kotlin
>      *
>      * @param [vid] the vid to hash
>      * @returns The first 64-bits of the hash as a [Long]
> ```

Document that it's now a little endian representation of the hash as a `Long` ([`HashCode.asLong`](https://guava.dev/releases/snapshot/api/docs/com/google/common/hash/HashCode.html#asLong()) uses little endian byte order).

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1563)* status: all files reviewed, 2 unresolved discussions (waiting on @kungfucraig)

___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/VidIndexMap.kt` line 104 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1563#-NvmEDHk3cDRUC3zbF2z:-NvmEDHk3cDRUC3zbF3-:b-lbeli4) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/2b21fb6cd848986c9f796808f9cc96867785a0cf/src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/VidIndexMap.kt#L104)):*
> ```Kotlin
>     fun hashVidToLongWithFarmHash(vid: Long, salt: Long): Long {
>       val hasher = Hashing.farmHashFingerprint64().newHasher()
>       return hasher.putLong(vid + salt).hash().asLong()
> ```

Note that this is losing bits on the hash input now. You're adding two 64-bit ints into a 64-bit result, which may overflow. It's more common when salting to concatenate rather than add.

I'd recommend the ByteBuffer overload of `putBytes` instead. The function documentation should also be clear on the details (e.g. which byte order any integer types are encoded in and that the salt is concatenated).



_Code quote:_
```Kotlin
putLong(vid + salt)
```

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1563)* status: all files reviewed, 2 unresolved discussions (waiting on @kungfucraig and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/VidIndexMap.kt` line 104 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1563#-NvmEDHk3cDRUC3zbF2z:-NvrHOA62GHbf8g8Uuzu:byp8yt0) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/2b21fb6cd848986c9f796808f9cc96867785a0cf/src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/VidIndexMap.kt#L104)):*

Note that this is losing bits on the hash input now. You're adding two 64-bit ints into a 64-bit result, which may overflow. It's more common when salting to concatenate rather than add.

I'd recommend the ByteBuffer overload of `putBytes` instead. The function documentation should also be clear on the details (e.g. which byte order any integer types are encoded in and that the salt is concatenated).
</blockquote></details>

i agree with Sanjay here. concat is clearer.

kungfucraig (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1563)* status: 7 of 10 files reviewed, 2 unresolved discussions (waiting on @SanjayVas and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/VidIndexMap.kt` line 100 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1563#-NvmDKuv2jnvDptPH3k-:-NvrVf8U7Ab58IOpJUXp:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/2b21fb6cd848986c9f796808f9cc96867785a0cf/src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/VidIndexMap.kt#L100)):*

Document that it's now a little endian representation of the hash as a `Long` ([`HashCode.asLong`](https://guava.dev/releases/snapshot/api/docs/com/google/common/hash/HashCode.html#asLong()) uses little endian byte order).
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/VidIndexMap.kt` line 104 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1563#-NvmEDHk3cDRUC3zbF2z:-Nvr39ft180g-coqlitz:b4ma2o1) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/2b21fb6cd848986c9f796808f9cc96867785a0cf/src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/VidIndexMap.kt#L104)):*

i agree with Sanjay here. concat is clearer.
</blockquote></details>

For what it's worth, this is how the current sampler does it, and it doesn't really matter if it overflows as we still get the effect we want.

I did however change it.

SanjayVas (approved):
> Reviewed 1 of 3 files at r7, 3 of 3 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1563)* status: all files reviewed, 1 unresolved discussion (waiting on @kungfucraig)

___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/VidIndexMap.kt` line 111 at r10](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1563#-Nvrj9uv5edNsbMaWRMz:-Nvrj9uv5edNsbMaWRN-:bk0cqaw) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/208da2689d4071c80a2f571632a2dd94f5a897ef/src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/VidIndexMap.kt#L111)):*
> ```Kotlin
>      * @param [vid] the vid to hash
>      * @param [salt] the value concatenated to the [vid] prior to hashing
>      * @returns the hash of the vid
> ```

nit: This is fine, but you don't need to go quite into code-level detail



_Suggestion:_
```Kotlin
     *
     * @param vid VID to hash
     * @param salt value concatenated to the big endian representation of [vid] before hashing
     * @returns the hash of the concatenation of the big endian representation of [vid] and [salt], as a [Long] with little endian byte ordering
```

stevenwarejones (approved):
Approving this since I'll be out. Its already 99% there and Sanjay can cover the rest.

> Reviewed 1 of 3 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1563#-:-NvrqTcf-YCaDMBYXz5P:b-2xrrjo)* status: all files reviewed, 1 unresolved discussion (waiting on @kungfucraig)

kungfucraig (commented):
> Reviewed 2 of 9 files at r1, 1 of 3 files at r2, 3 of 5 files at r5, 1 of 3 files at r7, 3 of 3 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1563)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @kungfucraig)

PR #1593, opened 2024-04-29
Fix setting of Metric state to SUCCEEDED

tristanvuong2021:
There is an issue with a query when deciding whether to set the Metric state to SUCCEEDED. This PR aims to fix it.


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1593)

SanjayVas (changes requested):
> Reviewed 1 of 2 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1593)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/postgres/readers/MetricReader.kt` line 238 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1593#-NxEpeEp-pH2jWUGZYiv:-NxEpeEqEWnBrxvBMePr:bmu6irc) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/bb1652d98ce1e4a5cf5a037d64f8202263bc7dfc/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/postgres/readers/MetricReader.kt#L238)):*
> ```Kotlin
>             $baseSqlJoins
>           WHERE Metrics.MeasurementConsumerId = $1
>             AND Metrics.MetricId IN (
> ```

The top-level query is already joining MeasurementConsumers and Metrics, and already has `Metrics.MeasurementConsumerId = $1` in the WHERE clause. I'm not certain how this change is actually helping.

tristanvuong2021 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1593)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/postgres/readers/MetricReader.kt` line 238 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1593#-NxEpeEp-pH2jWUGZYiv:-NxExtro5GrLyiL96Q4G:bnhppek) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/bb1652d98ce1e4a5cf5a037d64f8202263bc7dfc/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/postgres/readers/MetricReader.kt#L238)):*

The top-level query is already joining MeasurementConsumers and Metrics, and already has `Metrics.MeasurementConsumerId = $1` in the WHERE clause. I'm not certain how this change is actually helping.
</blockquote></details>

The original query only selects the rows with the cmms measurement IDs in the list. If not every measurement ID for a metric is in the list, then it only selects the measurements that have been updated in the current transaction and not the ones that haven't been updated yet. The state check will then always be SUCCEEDED. The best solution I could come up with was to grab the metric IDs associated with the updated measurements first, then grab the metrics.

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1593)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/postgres/readers/MetricReader.kt` line 238 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1593#-NxEpeEp-pH2jWUGZYiv:-NxF0JEz95kHecCdvFRH:bmznkc6) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/bb1652d98ce1e4a5cf5a037d64f8202263bc7dfc/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/postgres/readers/MetricReader.kt#L238)):*
> then it only selects the measurements that have been updated in the current transaction and not the ones that haven't been updated yet

Huh? It should select all Metrics with that CmmsMeasurementId, regardless of which transaction inserted them, as long as those previous transactions are committed.

Stripping down to the bare joins needed for the WHERE clause, I'm not seeing why this wouldn't work when expanded to add the additional columns/joins you need:

```sql
SELECT
  <columns>
FROM
  Metrics
  JOIN MetricMeasurements USING(MeasurementConsumerId, MetricId)
  JOIN Measurements USING(MeasurementConsumerId, MetricId)
WHERE
  Metric.MeasurementConsumerId = $1
  AND CmmsMeasurementId IN (VALUES ${ValuesListBoundStatement.VALUES_LIST_PLACEHOLDER})
```

tristanvuong2021 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1593)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/postgres/readers/MetricReader.kt` line 238 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1593#-NxEpeEp-pH2jWUGZYiv:-NxIhjF8CKjE_v83UAT6:b1x1aet) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/bb1652d98ce1e4a5cf5a037d64f8202263bc7dfc/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/postgres/readers/MetricReader.kt#L238)):*

> then it only selects the measurements that have been updated in the current transaction and not the ones that haven't been updated yet

Huh? It should select all Metrics with that CmmsMeasurementId, regardless of which transaction inserted them, as long as those previous transactions are committed.

Stripping down to the bare joins needed for the WHERE clause, I'm not seeing why this wouldn't work when expanded to add the additional columns/joins you need:

```sql
SELECT
  <columns>
FROM
  Metrics
  JOIN MetricMeasurements USING(MeasurementConsumerId, MetricId)
  JOIN Measurements USING(MeasurementConsumerId, MetricId)
WHERE
  Metric.MeasurementConsumerId = $1
  AND CmmsMeasurementId IN (VALUES ${ValuesListBoundStatement.VALUES_LIST_PLACEHOLDER})
```
</blockquote></details>

The original query does selects the right metrics, but doesn't select all the measurements for each metric. There is no struct in Postgres so if a measurement doesn't match the where condition, then the measurement will not be included when checking the state of every measurement for a metric. I added a test for this scenario in this PR after realizing this issue.

tristanvuong2021 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1593)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/postgres/readers/MetricReader.kt` line 238 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1593#-NxEpeEp-pH2jWUGZYiv:-NxJVERK5c4cSPPaInmX:b-e4lh2d) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/bb1652d98ce1e4a5cf5a037d64f8202263bc7dfc/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/postgres/readers/MetricReader.kt#L238)):*

I think what you're explaining is a cardinality issue: Metric has a 1:many relationship with Measurement. Therefore without using array structures you need one result row for each Measurement instead of one row per Metric. Since the relation is in a separate table (MetricMeasurements), this means one result row per MetricMeasurement row. Then you can collect the many part into the appropriate data structure when post-processing.
</blockquote></details>

That is it. So the issue is that all the measurements aren't available for post-processing and this PR fixes that.

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1593)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/postgres/readers/MetricReader.kt` line 238 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1593#-NxEpeEp-pH2jWUGZYiv:-NxJga7yDkRZLLu3CYBB:b-hs1q8h) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a350c2bd42f2c6f3b5ecfcb6953d1fca886ce62f/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/postgres/readers/MetricReader.kt#L238)):*

That is it. So the issue is that all the measurements aren't available for post-processing and this PR fixes that.
</blockquote></details>

[INNER] JOIN should already handle the 1:many though. If you inner join a one to a many, you should get back the number or rows on the *many* side.

SanjayVas (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1593)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/postgres/readers/MetricReader.kt` line 238 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1593#-NxEpeEp-pH2jWUGZYiv:-NxPGAj-7eY6p14-nyh0:b4b3c8d) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/61524dcb2f770096f04b0e5016825d767ed427c7/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/postgres/readers/MetricReader.kt#L238)):*

[INNER] JOIN should already handle the 1:many though. If you inner join a one to a many, you should get back the number or rows on the *many* side.
</blockquote></details>

Clarified from offline discussion and function docs.

SanjayVas:
Is this fixing a bug that made it into the latest release (v0.5.4)?

tristanvuong2021 (commented):
It is.

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1593#-:-NxU_A742W2s3vRMJI9D:bumi984)* status: 1 of 2 files reviewed, all discussions resolved (waiting on @Marco-Premier and @SanjayVas)

SanjayVas (commented):
Can you please file an issue then? And we'll need to have a Slack post to flag that we found a bug in v0.5.4.

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1593#-:-NxUlyZD0hzT3gm2qLKN:b-8kw70r)* status: 1 of 2 files reviewed, all discussions resolved (waiting on @Marco-Premier)

stevenwarejones (approved):
> Reviewed 1 of 1 files at r2, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1593)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Marco-Premier)

PR #1602, opened 2024-05-02
Add support to run Panel exchange task on EMR

sandip80:
## Changes
* Add EMR Serverless execution support for Decrypt step of Panel exchange
* Move to BouncyCastle for csr generation instead of openssl (CSR is used by AWS private CA IssueCertificate API)
* Update panel exchange image and terraform deployment to incorporate EMR serverless execution permission and resources.

## Testing
* Unit test
* E2E Cloud test


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1602)

sandip80:
*[`MODULE.bazel` line 129 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1602#-NzdJXNgEempZfkUCzGW:-NzdJXNgEempZfkUCzGX:b-b6s05r) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/19bceba64ec2a3a78f5860da0ab88f10905a2fd9/MODULE.bazel#L129)):*
> ```Starlark
> bazel_dep(
>     name = "common-jvm",
>     version = "0.82.2",
> ```

Version upgrade required after common-jvm PR is merged

sandip80:
*[`MODULE.bazel` line 136 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1602#-NzdK6GVEs1g2YqeJdAp:-NzdK6GWCS4pUf2W5dF3:b4nz137) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/19bceba64ec2a3a78f5860da0ab88f10905a2fd9/MODULE.bazel#L136)):*
> ```Starlark
>     commit = "1cdb3696bce54b0906396657029cdbea4b76a793",
>     remote = "https://github.com/world-federation-of-advertisers/common-jvm",
> )
> ```

To be removed



_Code quote:_
```Starlark
git_override(
    module_name = "common-jvm",
    commit = "1cdb3696bce54b0906396657029cdbea4b76a793",
    remote = "https://github.com/world-federation-of-advertisers/common-jvm",
)
```

stevenwarejones (changes requested):
> Reviewed 4 of 17 files at r1, 36 of 38 files at r2, 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1602)* status: all files reviewed, 16 unresolved discussions (waiting on @sandip80 and @SanjayVas)

___
*[`imports/java/org/apache/beam/runners/spark/BUILD.bazel` line 5 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1602#-Nzi4q_n4HJEZhvPNyWR:-Nzi4q_n4HJEZhvPNyWS:bua22un) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/651063c96d1d8a60ab136c3efda5992041ddeb45/imports/java/org/apache/beam/runners/spark/BUILD.bazel#L5)):*
> ```Starlark
> alias(
>     name = "spark",
>     actual = "@maven//:org_apache_beam_beam_runners_spark_3",
> ```

does anyone from the Google side want to test this and make sure there's not issues on the google cloud? @SanjayVas  @robinsons @jonmolle
___
*[`src/main/kotlin/org/wfanet/panelmatch/client/deploy/ExchangeWorkflowDaemonFromFlags.kt` line 100 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1602#-Nzi6Ev62oqQayP9aH8l:-Nzi6Ev62oqQayP9aH8m:be5gqts) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/651063c96d1d8a60ab136c3efda5992041ddeb45/src/main/kotlin/org/wfanet/panelmatch/client/deploy/ExchangeWorkflowDaemonFromFlags.kt#L100)):*
> ```Kotlin
>   override val throttler: Throttler by lazy { createThrottler() }
> 
>   protected val taskContext: TaskParameters by lazy {
> ```

I think this can stay private. see my other comment.
___
*[`src/main/kotlin/org/wfanet/panelmatch/client/deploy/ExchangeWorkflowDaemonFromFlags.kt` line 139 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1602#-Nzi6H4x-Gelqscwh40M:-Nzi6H4x-Gelqscwh40N:b-qwnu7s) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/651063c96d1d8a60ab136c3efda5992041ddeb45/src/main/kotlin/org/wfanet/panelmatch/client/deploy/ExchangeWorkflowDaemonFromFlags.kt#L139)):*
> ```Kotlin
>   override val identity: Identity by lazy { Identity(flags.id, flags.partyType) }
> 
>   protected fun createThrottler(): Throttler = MinimumIntervalThrottler(clock, flags.pollingInterval)
> ```

ditto on staying private.
___
*[`src/main/kotlin/org/wfanet/panelmatch/client/deploy/ProductionExchangeTaskMapper.kt` line 185 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1602#-NzhuWbK7CSHDcLCAz_Q:-NzhuWbK7CSHDcLCAz_R:b2ht4lt) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/651063c96d1d8a60ab136c3efda5992041ddeb45/src/main/kotlin/org/wfanet/panelmatch/client/deploy/ProductionExchangeTaskMapper.kt#L185)):*
> ```Kotlin
> 
>     return if (emrBeamTaskExecutorOnDaemon) {
>       requireNotNull(emrService)
> ```

is there a use case where this would be set but emrBeamTaskExeuctor would be false? If not, you could just drop this variable and rely on if emrService is set or not.
___
*[`src/main/kotlin/org/wfanet/panelmatch/client/deploy/ProductionExchangeTaskMapper.kt` line 193 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1602#-NzhvLmA2JfWPI_wVthS:-NzhvLmA2JfWPI_wVthT:b-acbrdo) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/651063c96d1d8a60ab136c3efda5992041ddeb45/src/main/kotlin/org/wfanet/panelmatch/client/deploy/ProductionExchangeTaskMapper.kt#L193)):*
> ```Kotlin
>         "${ExchangeWorkflow.Step.StepCase.DECRYPT_PRIVATE_MEMBERSHIP_QUERY_RESULTS_STEP.name.lowercase()}-$exchangeId-${date.format(DateTimeFormatter.ISO_LOCAL_DATE)}",
>         workflow.stepsList.indexOfFirst {
>           it.stepCase == ExchangeWorkflow.Step.StepCase.DECRYPT_PRIVATE_MEMBERSHIP_QUERY_RESULTS_STEP
> ```

it.stepCase == step.StepCase
___
*[`src/main/kotlin/org/wfanet/panelmatch/client/deploy/example/aws/AwsExampleDaemon.kt` line 95 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1602#-Nzi6Lro1zFLcDcqhrG-:-Nzi6Lro1zFLcDcqhrG0:bq97cg2) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/651063c96d1d8a60ab136c3efda5992041ddeb45/src/main/kotlin/org/wfanet/panelmatch/client/deploy/example/aws/AwsExampleDaemon.kt#L95)):*
> ```Kotlin
>   private var s3FromBeam by Delegates.notNull<Boolean>()
> 
>   override val exchangeTaskMapper: ExchangeTaskMapper by lazy {
> ```

I don't think you need to override the exchangeTaskMapper. Instead, what I'd suggest is adding 
```
protected abstract emrService...
```
to ExchangeWorkflowDaemonFromFlags.kt and passing it in there. Similar to how makePipelineOptions is declared and consumed.
___
*[`src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/EmrExchangeTask.kt` line 27 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1602#-Nzhwl0XE8u-sUICiYIu:-Nzhwl0XE8u-sUICiYIv:b-aksaxp) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/651063c96d1d8a60ab136c3efda5992041ddeb45/src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/EmrExchangeTask.kt#L27)):*
> ```Kotlin
>   private val emrExchangeTaskService: EmrExchangeTaskService,
>   private val exchangeId: String,
>   private val exchangeStepName: String,
> ```

see my other comment. I suggest just passing in the stepCase here instead of the exchangeStepName (which is not really an exchangeStepName but a jobRunId)
___
*[`src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/EmrExchangeTask.kt` line 36 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1602#-NzhwWBb-btc9Qj6dqaq:-NzhwWBb-btc9Qj6dqar:bkugqgc) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/651063c96d1d8a60ab136c3efda5992041ddeb45/src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/EmrExchangeTask.kt#L36)):*
> ```Kotlin
>       emrExchangeTaskService.runPanelExchangeStepOnEmrApp(
>         exchangeId,
>         exchangeStepName,
> ```

this is a jobRunId not an exchangeStepName, right? Can you just calculate it within this task
```
val jobRunId = "${stepCase.name.lowercase()}-$exchangeId-${exchangeDate.format(DateTimeFormatter.ISO_LOCAL_DATE)}"
```

___
*[`src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/emr/EmrExchangeTaskService.kt` line 29 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1602#-Nzi3Ou55hU2umm9E5oA:-Nzi3Ou55hU2umm9E5oB:bidcex6) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/651063c96d1d8a60ab136c3efda5992041ddeb45/src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/emr/EmrExchangeTaskService.kt#L29)):*
> ```Kotlin
> private const val EMR_EXCHANGE_TASK_APP_NAME = "panel-exchange-beam-task"
> 
> open class EmrExchangeTaskService(
> ```

see my other commend about kdoc and making this not open
___
*[`src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/emr/EmrExchangeTaskService.kt` line 39 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1602#-Nzhx0_9DqgaAVUEZrul:-Nzhx0_9DqgaAVUEZrum:b-eflvyw) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/651063c96d1d8a60ab136c3efda5992041ddeb45/src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/emr/EmrExchangeTaskService.kt#L39)):*
> ```Kotlin
> ) {
> 
>   private var appId: String
> ```

this could just be a lazy variable
```
private val appId: String by lazy  {
    val exchangeTaskAppIdBlob = runBlocking { storageClient.getBlob(exchangeTaskAppIdPath) }
    if (exchangeTaskAppIdBlob != null) {
      runBlocking { exchangeTaskAppIdBlob.toStringUtf8() }
    } else {
      val applicationId = emrServerlessClientService.createApplication(EMR_EXCHANGE_TASK_APP_NAME)
      runBlocking { storageClient.writeBlob(exchangeTaskAppIdPath, applicationId.toByteStringUtf8()) }
      applicationId
    }

}
```

___
*[`src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/emr/EmrServerlessClientService.kt` line 43 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1602#-Nzi1i-42up_qwoQFhZI:-Nzi1i-42up_qwoQFhZJ:bmj30js) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/651063c96d1d8a60ab136c3efda5992041ddeb45/src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/emr/EmrServerlessClientService.kt#L43)):*
> ```Kotlin
> private val STATE_CHECK_SLEEP_INTERVAL: Duration = 10.toDuration(SECONDS)
> 
> open class EmrServerlessClientService(
> ```

can you add a kdoc? Is this class open for mocking? If so, I'd prefer you create an interface instead.
___
*[`src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/emr/EmrServerlessClientService.kt` line 113 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1602#-Nzi30RM11KYiphbybQP:-Nzi30RM11KYiphbybQQ:b-gthszy) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/651063c96d1d8a60ab136c3efda5992041ddeb45/src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/emr/EmrServerlessClientService.kt#L113)):*
> ```Kotlin
>               "spark.driver.memory" to "20g",
>               "spark.emr-serverless.driverEnv.JAVA_HOME" to "/usr/lib/jvm/java-17-amazon-corretto.x86_64/",
>               "spark.executorEnv.JAVA_HOME" to "/usr/lib/jvm/java-17-amazon-corretto.x86_64/",
> ```

do these always need to be set? what happens if these last two aren't set?
___
*[`src/main/kotlin/org/wfanet/panelmatch/common/certificates/BUILD.bazel` line 12 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1602#-Nzi3qFBCRP5or7VN5io:-Nzi3qFBCRP5or7VN5ip:bhwaz1s) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/651063c96d1d8a60ab136c3efda5992041ddeb45/src/main/kotlin/org/wfanet/panelmatch/common/certificates/BUILD.bazel#L12)):*
> ```Starlark
>     srcs = glob(["*.kt"]),
>     deps = [
>         "//imports/java/org/bouncycastle/bcpkix",
> ```

I think we should have a separate target for GenerateCsr and have a more narrow exposure. Its really only the AWS CA that needs to consume it.
___
*[`imports/java/org/bouncycastle/bcpkix/BUILD.bazel` line 1 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1602#-NzhbZHmEHbVgm5H6Y4b:-NzhbZHmEHbVgm5H6Y4c:bkin15w) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/651063c96d1d8a60ab136c3efda5992041ddeb45/imports/java/org/bouncycastle/bcpkix/BUILD.bazel#L1)):*
> ```Starlark
> package(default_visibility = ["//visibility:public"])
> ```

I suggest restricting visibility

SanjayVas (changes requested):
> Reviewed 2 of 17 files at r1, 10 of 38 files at r2, 1 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1602)* status: all files reviewed, 26 unresolved discussions (waiting on @sandip80 and @stevenwarejones)

___
*[`src/main/k8s/panelmatch/dev/example_edp_daemon_aws.cue` line 21 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1602#-Nzk0HfIDd_nc6TmXnmF:-Nzk0HfIDd_nc6TmXnmG:b4lz82w) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/651063c96d1d8a60ab136c3efda5992041ddeb45/src/main/k8s/panelmatch/dev/example_edp_daemon_aws.cue#L21)):*
> ```CUE
> 	partyName:          string @tag("party_name")
> 	partyType:          "DATA_PROVIDER"
> 	cloudStorageBucket: "edp-blob-storage"
> ```

The storage bucket should come from a Make var, since they must be globally unique. K8s configs for any environment must be generatable from Bazel without any code changes.
___
*[`src/main/k8s/panelmatch/dev/example_edp_daemon_aws.cue` line 30 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1602#-Nzk0NC7C5ZS1FuJEB_r:-Nzk0NC7C5ZS1FuJEB_s:b-snn6d5) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/651063c96d1d8a60ab136c3efda5992041ddeb45/src/main/k8s/panelmatch/dev/example_edp_daemon_aws.cue#L30)):*
> ```CUE
> }
> 
> // Blank strings are populated by the Terraform-backed deployment supporting
> ```

Where does this sequence relative to @Marco-Premier's changes to refactor the PX Terraform? (Terraform should not be used to manage K8s object configs).
___
*[`src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/emr/EmrExchangeTaskService.kt` line 62 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1602#-Nzk16cD4GiZmICIczzv:-Nzk16cD4GiZmICIczzw:b-ijanq4) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/651063c96d1d8a60ab136c3efda5992041ddeb45/src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/emr/EmrExchangeTaskService.kt#L62)):*
> ```Kotlin
> 
>     if (!started) {
>       throw RuntimeException("Panel exchange app was not started successfully")
> ```

nit: If you're going to use RuntimeException, just use Exception instead. Kotlin doesn't use caught exceptions, so there's not a useful distinction between Exception and RuntimeException.
___
*[`src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/emr/EmrServerlessClientService.kt` line 43 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1602#-Nzk1_oi3df5RRRhtf7B:-Nzk1_oi3df5RRRhtf7C:bw22g9h) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/651063c96d1d8a60ab136c3efda5992041ddeb45/src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/emr/EmrServerlessClientService.kt#L43)):*
> ```Kotlin
> private val STATE_CHECK_SLEEP_INTERVAL: Duration = 10.toDuration(SECONDS)
> 
> open class EmrServerlessClientService(
> ```

"Service" for an application that depends on gRPC generally means a gRPC service impl. This looks like it's just a client (which happens to be implemented as a wrapper around `EmrServerlessClient`).



_Suggestion:_
```Kotlin
EmrServerlessClient
```
___
*[`src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/emr/EmrServerlessClientService.kt` line 43 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1602#-Nzi1i-42up_qwoQFhZI:-Nzk1kKf8AAFhk6I8Tlw:b-949wsb) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/651063c96d1d8a60ab136c3efda5992041ddeb45/src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/emr/EmrServerlessClientService.kt#L43)):*

can you add a kdoc? Is this class open for mocking? If so, I'd prefer you create an interface instead.
</blockquote></details>

+1 to interface rather than open. A class should only be open if it's intended to be subclassed (and not just for mocking).
___
*[`src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/emr/EmrServerlessClientService.kt` line 61 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1602#-Nzk2fPn5hO3SflAQv6C:-Nzk2fPn5hO3SflAQv6D:bc93yom) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/651063c96d1d8a60ab136c3efda5992041ddeb45/src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/emr/EmrServerlessClientService.kt#L61)):*
> ```Kotlin
> 
>   open suspend fun startOrStopApplication(applicationId: String, start: Boolean): Boolean {
>     var checks = 0
> ```

Having a "startOrStop" method with a "start" bool param is very odd design. Separate "start" and "stop" methods that throw IllegalStateException if it's in the wrong starting state make more sense.
___
*[`src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/emr/EmrServerlessClientService.kt` line 65 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1602#-Nzk3J7UB1gVpYELSJbe:-Nzk3J7UB1gVpYELSJbf:b-nji1po) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/651063c96d1d8a60ab136c3efda5992041ddeb45/src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/emr/EmrServerlessClientService.kt#L65)):*
> ```Kotlin
>     while (checks < maxStateChecks) {
>       val request = GetApplicationRequest.builder().applicationId(applicationId).build()
>       val response = emrServerlessClient.getApplication(request)
> ```

Are these client methods blocking? If so, they should be wrapped in `withContext` when called from coroutine code. Pass in a CoroutineContext to the constructor annotated with `@BlockingExecutor` for this purpose.
___
*[`src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/emr/EmrServerlessClientService.kt` line 79 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1602#-Nzk4LwM-Dpo5MmGwTxw:-Nzk4LwM-Dpo5MmGwTxx:b4itc1) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/651063c96d1d8a60ab136c3efda5992041ddeb45/src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/emr/EmrServerlessClientService.kt#L79)):*
> ```Kotlin
>       }
> 
>       if (checks == 0) { delay(sleepTime * 3) } else { delay(sleepTime) }
> ```

It'd be better class design IMO to separate out the polling wait. e.g. have a non-suspending `start` method annotated with `@Blocking` (assuming the client calls are blocking), and a separate suspending `waitForApplicationState(applicationId: String, state: ApplicationState, pollingInterval: Duration = DEFAULT_POLLING_INTERVAL)`. Also rather than a number of attempts, the caller can just wrap in `withTimeout` if desired. Obviously if you really want, you could still wrap those two calls into `startAndWait` like you have now.
___
*[`src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/emr/EmrServerlessClientService.kt` line 108 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1602#-Nzk5Z726Nzbrd3PPijW:-Nzk5Z726Nzbrd3PPijX:bghjmpy) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/651063c96d1d8a60ab136c3efda5992041ddeb45/src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/emr/EmrServerlessClientService.kt#L108)):*
> ```Kotlin
>             .classification("spark-defaults")
>             .properties(mapOf(
>               "spark.executor.cores" to "4",
> ```

nit: these should be configurable via command-line options
___
*[`src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/emr/EmrServerlessClientService.kt` line 154 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1602#-Nzk5xcN0buIzLKzf3gl:-Nzk5xcO8XeBGcFdK2v_:b-ms1s7h) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/651063c96d1d8a60ab136c3efda5992041ddeb45/src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/emr/EmrServerlessClientService.kt#L154)):*
> ```Kotlin
>   }
> 
>   private val logger = Logger.getLogger(EmrServerlessClientService::class.java.name)
> ```

This should live in the companion object instead.



_Suggestion:_
```Kotlin
private val logger: Logger = Logger.getLogger(this::class.java.name)
```
___
*[`src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/emr/EmrServerlessClientService.kt` line 156 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1602#-Nzk5ji-3scqknKf1Nnd:-Nzk5ji-3scqknKf1Nne:b-nf1ny7) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/651063c96d1d8a60ab136c3efda5992041ddeb45/src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/emr/EmrServerlessClientService.kt#L156)):*
> ```Kotlin
>   private val logger = Logger.getLogger(EmrServerlessClientService::class.java.name)
> 
>   private val terminalStates = setOf(
> ```

CONST_CASE and move to companion object



_Suggestion:_
```Kotlin
TERMINAL_STATES
```

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1602)* status: all files reviewed, 26 unresolved discussions (waiting on @sandip80 and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/emr/EmrServerlessClientService.kt` line 61 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1602#-Nzk2fPn5hO3SflAQv6C:-NzmnSRAAf-xIuHRNBYZ:b-yi1qx1) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/651063c96d1d8a60ab136c3efda5992041ddeb45/src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/emr/EmrServerlessClientService.kt#L61)):*

Having a "startOrStop" method with a "start" bool param is very odd design. Separate "start" and "stop" methods that throw IllegalStateException if it's in the wrong starting state make more sense.
</blockquote></details>

I also thought this. I'd prefer separate start and stop. The example AWS code I saw I all had separate start and stops

sandip80:
*[`imports/java/org/bouncycastle/bcpkix/BUILD.bazel` line 1 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1602#-NzhbZHmEHbVgm5H6Y4b:-O62yaau6j_z15Sxo6n-:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/651063c96d1d8a60ab136c3efda5992041ddeb45/imports/java/org/bouncycastle/bcpkix/BUILD.bazel#L1)):*

I suggest restricting visibility
</blockquote></details>

Done.

sandip80:
*[`src/main/k8s/panelmatch/dev/example_edp_daemon_aws.cue` line 21 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1602#-Nzk0HfIDd_nc6TmXnmF:-O62yxvA6fcNEOrPrtPJ:b-xfs7jh) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/651063c96d1d8a60ab136c3efda5992041ddeb45/src/main/k8s/panelmatch/dev/example_edp_daemon_aws.cue#L21)):*

The storage bucket should come from a Make var, since they must be globally unique. K8s configs for any environment must be generatable from Bazel without any code changes.
</blockquote></details>

Removed the _defaultAwsConfig and moved the dynamic variables to Make var. Storage bucket name isn't dynamic so we should be fine to have it here (same as Google cloud).

sandip80:
*[`src/main/k8s/panelmatch/dev/example_edp_daemon_aws.cue` line 30 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1602#-Nzk0NC7C5ZS1FuJEB_r:-O62zCiP9GXn-RMCsfR5:b-h77kd0) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/651063c96d1d8a60ab136c3efda5992041ddeb45/src/main/k8s/panelmatch/dev/example_edp_daemon_aws.cue#L30)):*

Where does this sequence relative to @Marco-Premier's changes to refactor the PX Terraform? (Terraform should not be used to manage K8s object configs).
</blockquote></details>

Removed the default config in favor of Make var

sandip80:
*[`src/main/kotlin/org/wfanet/panelmatch/client/deploy/ExchangeWorkflowDaemonFromFlags.kt` line 100 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1602#-Nzi6Ev62oqQayP9aH8l:-O62zH1y5gDjlhdiQBRV:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/651063c96d1d8a60ab136c3efda5992041ddeb45/src/main/kotlin/org/wfanet/panelmatch/client/deploy/ExchangeWorkflowDaemonFromFlags.kt#L100)):*

I think this can stay private. see my other comment.
</blockquote></details>

Done.

sandip80:
*[`src/main/kotlin/org/wfanet/panelmatch/client/deploy/ExchangeWorkflowDaemonFromFlags.kt` line 139 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1602#-Nzi6H4x-Gelqscwh40M:-O62zcZx4CoCaOUvG32g:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/651063c96d1d8a60ab136c3efda5992041ddeb45/src/main/kotlin/org/wfanet/panelmatch/client/deploy/ExchangeWorkflowDaemonFromFlags.kt#L139)):*

ditto on staying private.
</blockquote></details>

Done.

sandip80:
*[`src/main/kotlin/org/wfanet/panelmatch/client/deploy/ProductionExchangeTaskMapper.kt` line 185 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1602#-NzhuWbK7CSHDcLCAz_Q:-O62zyng1xUSE_9kbB-X:b-o3jct8) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/651063c96d1d8a60ab136c3efda5992041ddeb45/src/main/kotlin/org/wfanet/panelmatch/client/deploy/ProductionExchangeTaskMapper.kt#L185)):*

is there a use case where this would be set but emrBeamTaskExeuctor would be false? If not, you could just drop this variable and rely on if emrService is set or not.
</blockquote></details>

Removed the extra boolean

sandip80:
*[`src/main/kotlin/org/wfanet/panelmatch/client/deploy/example/aws/AwsExampleDaemon.kt` line 95 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1602#-Nzi6Lro1zFLcDcqhrG-:-O63-5kT6VPTsb2sqK47:b-2ofgx4) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/651063c96d1d8a60ab136c3efda5992041ddeb45/src/main/kotlin/org/wfanet/panelmatch/client/deploy/example/aws/AwsExampleDaemon.kt#L95)):*

I don't think you need to override the exchangeTaskMapper. Instead, what I'd suggest is adding 
```
protected abstract emrService...
```
to ExchangeWorkflowDaemonFromFlags.kt and passing it in there. Similar to how makePipelineOptions is declared and consumed.
</blockquote></details>

Removed the override from here and added makeRemoteTaskOrchestrator similar to makePipelineOptions

sandip80:
*[`src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/EmrExchangeTask.kt` line 27 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1602#-Nzhwl0XE8u-sUICiYIu:-O63-ERl-fSAYj53L16-:bfqlgf2) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/651063c96d1d8a60ab136c3efda5992041ddeb45/src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/EmrExchangeTask.kt#L27)):*

see my other comment. I suggest just passing in the stepCase here instead of the exchangeStepName (which is not really an exchangeStepName but a jobRunId)
</blockquote></details>

Changed the var name to orchestratorName

sandip80:
*[`src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/EmrExchangeTask.kt` line 36 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1602#-NzhwWBb-btc9Qj6dqaq:-O6362tD3Ih6uyg0a-fy:bagpujs) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/651063c96d1d8a60ab136c3efda5992041ddeb45/src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/EmrExchangeTask.kt#L36)):*

this is a jobRunId not an exchangeStepName, right? Can you just calculate it within this task
```
val jobRunId = "${stepCase.name.lowercase()}-$exchangeId-${exchangeDate.format(DateTimeFormatter.ISO_LOCAL_DATE)}"
```

</blockquote></details>

Moved into the EMR orchestrator task

sandip80:
*[`src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/emr/EmrExchangeTaskService.kt` line 29 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1602#-Nzi3Ou55hU2umm9E5oA:-O6367ZBBSPMfTq16mzH:b-bu91xm) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/651063c96d1d8a60ab136c3efda5992041ddeb45/src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/emr/EmrExchangeTaskService.kt#L29)):*

see my other commend about kdoc and making this not open
</blockquote></details>

Updated

sandip80:
*[`src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/emr/EmrExchangeTaskService.kt` line 39 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1602#-Nzhx0_9DqgaAVUEZrul:-O636A0PCse_YcC95W9e:bb6un1j) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/651063c96d1d8a60ab136c3efda5992041ddeb45/src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/emr/EmrExchangeTaskService.kt#L39)):*

this could just be a lazy variable
```
private val appId: String by lazy  {
    val exchangeTaskAppIdBlob = runBlocking { storageClient.getBlob(exchangeTaskAppIdPath) }
    if (exchangeTaskAppIdBlob != null) {
      runBlocking { exchangeTaskAppIdBlob.toStringUtf8() }
    } else {
      val applicationId = emrServerlessClientService.createApplication(EMR_EXCHANGE_TASK_APP_NAME)
      runBlocking { storageClient.writeBlob(exchangeTaskAppIdPath, applicationId.toByteStringUtf8()) }
      applicationId
    }

}
```

</blockquote></details>

Done

sandip80:
*[`src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/emr/EmrExchangeTaskService.kt` line 62 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1602#-Nzk16cD4GiZmICIczzv:-O636Z7Q6UsNiN7rYzs_:bb6un1j) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/651063c96d1d8a60ab136c3efda5992041ddeb45/src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/emr/EmrExchangeTaskService.kt#L62)):*

nit: If you're going to use RuntimeException, just use Exception instead. Kotlin doesn't use caught exceptions, so there's not a useful distinction between Exception and RuntimeException.
</blockquote></details>

Done

sandip80:
*[`src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/emr/EmrServerlessClientService.kt` line 113 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1602#-Nzi30RM11KYiphbybQP:-O6372RBEPLjkP3VXWg0:b1lljt3) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/651063c96d1d8a60ab136c3efda5992041ddeb45/src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/emr/EmrServerlessClientService.kt#L113)):*

do these always need to be set? what happens if these last two aren't set?
</blockquote></details>

This was added since default java runtime on older EMR image was set to 1.8 but I see that this has changed to 17 as per this doc: https://docs.aws.amazon.com/emr/latest/ReleaseGuide/configuring-java8.html . Will remove these 2 settings.

sandip80:
*[`src/main/kotlin/org/wfanet/panelmatch/client/deploy/ProductionExchangeTaskMapper.kt` line 193 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1602#-NzhvLmA2JfWPI_wVthS:-O63JNzQDg8y_ilJuIwv:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/651063c96d1d8a60ab136c3efda5992041ddeb45/src/main/kotlin/org/wfanet/panelmatch/client/deploy/ProductionExchangeTaskMapper.kt#L193)):*

it.stepCase == step.StepCase
</blockquote></details>

Done.

sandip80:
*[`src/main/kotlin/org/wfanet/panelmatch/common/certificates/BUILD.bazel` line 12 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1602#-Nzi3qFBCRP5or7VN5io:-O63O0iM0m8IOf_LFizJ:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/651063c96d1d8a60ab136c3efda5992041ddeb45/src/main/kotlin/org/wfanet/panelmatch/common/certificates/BUILD.bazel#L12)):*

I think we should have a separate target for GenerateCsr and have a more narrow exposure. Its really only the AWS CA that needs to consume it.
</blockquote></details>

Done.

sandip80:
*[`src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/emr/EmrServerlessClientService.kt` line 43 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1602#-Nzi1i-42up_qwoQFhZI:-O63OQX64HczYMXEvCLD:b-upns0h) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/651063c96d1d8a60ab136c3efda5992041ddeb45/src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/emr/EmrServerlessClientService.kt#L43)):*

+1 to interface rather than open. A class should only be open if it's intended to be subclassed (and not just for mocking).
</blockquote></details>

Created a separate interface for unit test

sandip80:
*[`src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/emr/EmrServerlessClientService.kt` line 43 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1602#-Nzk1_oi3df5RRRhtf7B:-O63OVjB8Rr4rv8G4tEm:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/651063c96d1d8a60ab136c3efda5992041ddeb45/src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/emr/EmrServerlessClientService.kt#L43)):*

"Service" for an application that depends on gRPC generally means a gRPC service impl. This looks like it's just a client (which happens to be implemented as a wrapper around `EmrServerlessClient`).
</blockquote></details>

Done.

sandip80:
*[`src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/emr/EmrServerlessClientService.kt` line 61 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1602#-Nzk2fPn5hO3SflAQv6C:-O63OXk0E4Qzyizljo8k:btuakxw) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/651063c96d1d8a60ab136c3efda5992041ddeb45/src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/emr/EmrServerlessClientService.kt#L61)):*

I also thought this. I'd prefer separate start and stop. The example AWS code I saw I all had separate start and stops
</blockquote></details>

Created separate start and stop functions

sandip80:
*[`src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/emr/EmrServerlessClientService.kt` line 154 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1602#-Nzk5xcN0buIzLKzf3gl:-O63P7mi2nMMbctPWNzq:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/651063c96d1d8a60ab136c3efda5992041ddeb45/src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/emr/EmrServerlessClientService.kt#L154)):*

This should live in the companion object instead.
</blockquote></details>

Done.

SanjayVas (changes requested):
> Reviewed 2 of 46 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1602)* status: 19 of 63 files reviewed, 27 unresolved discussions (waiting on @Marco-Premier, @sandip80, and @stevenwarejones)

*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1602#-O67LPmb4oPgt-U6S6_T:-O67LPmb4oPgt-U6S6_U:bncdc3z) (no related file):*
Can you rebase this on head? There's relevant base changes that make this harder to review.

___
___
*[`MODULE.bazel` line 129 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1602#-NzdJXNgEempZfkUCzGW:-O67Lq6P6TP76ALzFGPS:bhig3a8) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/19bceba64ec2a3a78f5860da0ab88f10905a2fd9/MODULE.bazel#L129)):*
<details><summary><i>Previously, sandip80 (Sandip Samantaray) wrote…</i></summary><blockquote>

Version upgrade required after common-jvm PR is merged
</blockquote></details>

I went ahead and cut a v0.90.0 release for you: https://github.com/world-federation-of-advertisers/common-jvm/releases/tag/v0.90.0

You just need a PR to https://github.com/world-federation-of-advertisers/bazel-registry so you can use it here.
___
*[`MODULE.bazel` line 231 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1602#-O67Nkl0CM4jO8MfeIqB:-O67Nkl0CM4jO8MfeIqC:b-6g02sn) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/48d5b2e06d0837d92effb0c79009ffa342410011/MODULE.bazel#L231)):*
> ```Starlark
>         "org.slf4j:slf4j-simple:1.7.32",
>         "software.amazon.awssdk:acmpca:2.23.18",
>         "software.amazon.awssdk:emrserverless:2.23.18",
> ```

nit: Split out an AWS_SDK_VERSION constant here like we do in common-jvm, with a comment indicating that it matches the common-jvm version.

#1790 will simplify this, but I don't know which PR will land first



_Code quote:_
```Starlark
2.23.18
```

sandip80:
*[`src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/emr/EmrServerlessClientService.kt` line 79 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1602#-Nzk4LwM-Dpo5MmGwTxw:-O67ooXiDq2lxRxdkkbf:bdld8k2) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/651063c96d1d8a60ab136c3efda5992041ddeb45/src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/emr/EmrServerlessClientService.kt#L79)):*

It'd be better class design IMO to separate out the polling wait. e.g. have a non-suspending `start` method annotated with `@Blocking` (assuming the client calls are blocking), and a separate suspending `waitForApplicationState(applicationId: String, state: ApplicationState, pollingInterval: Duration = DEFAULT_POLLING_INTERVAL)`. Also rather than a number of attempts, the caller can just wrap in `withTimeout` if desired. Obviously if you really want, you could still wrap those two calls into `startAndWait` like you have now.
</blockquote></details>

Changed logic to use timeout based polling instead of max count.

sandip80:
*[`src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/emr/EmrServerlessClientService.kt` line 156 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1602#-Nzk5ji-3scqknKf1Nnd:-O67omuUDUwy-f1zGoYl:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/651063c96d1d8a60ab136c3efda5992041ddeb45/src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/emr/EmrServerlessClientService.kt#L156)):*

CONST_CASE and move to companion object
</blockquote></details>

Done.

sandip80:
*[`src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/emr/EmrServerlessClientService.kt` line 65 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1602#-Nzk3J7UB1gVpYELSJbe:-O63ObWxDkgEEyHTKLcg:bbws0c) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/651063c96d1d8a60ab136c3efda5992041ddeb45/src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/emr/EmrServerlessClientService.kt#L65)):*

Are these client methods blocking? If so, they should be wrapped in `withContext` when called from coroutine code. Pass in a CoroutineContext to the constructor annotated with `@BlockingExecutor` for this purpose.
</blockquote></details>

Added ``withContext(Dispatchers.IO``

stevenwarejones (approved):
> Reviewed 34 of 46 files at r4, 11 of 11 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1602)* status: 62 of 63 files reviewed, 17 unresolved discussions (waiting on @Marco-Premier, @sandip80, and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/remote/aws/EmrServerlessClientImpl.kt` line 122 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1602#-O67xVzA7BgUyPaNwOOG:-O67xVzA7BgUyPaNwOOH:bn5f7zf) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/74ce57d26f434f3097e764254589e69f43140262/src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/remote/aws/EmrServerlessClientImpl.kt#L122)):*
> ```Kotlin
>                 .properties(
>                   mapOf(
>                     "spark.executor.cores" to "4",
> ```

can you add a TODO to move this to a config?
___
*[`src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/emr/EmrExchangeTaskService.kt` line 29 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1602#-Nzi3Ou55hU2umm9E5oA:-O67xH_n7NxkA8xa91b0:b5q4d58) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/651063c96d1d8a60ab136c3efda5992041ddeb45/src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/emr/EmrExchangeTaskService.kt#L29)):*
<details><summary><i>Previously, sandip80 (Sandip Samantaray) wrote…</i></summary><blockquote>

Updated
</blockquote></details>

Still needs a kdoc

sandip80:
*[`MODULE.bazel` line 129 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1602#-NzdJXNgEempZfkUCzGW:-O68iikv9kixdNuNLZhR:b-bu91xm) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/19bceba64ec2a3a78f5860da0ab88f10905a2fd9/MODULE.bazel#L129)):*

I went ahead and cut a v0.90.0 release for you: https://github.com/world-federation-of-advertisers/common-jvm/releases/tag/v0.90.0

You just need a PR to https://github.com/world-federation-of-advertisers/bazel-registry so you can use it here.
</blockquote></details>

Updated

sandip80:
*[`MODULE.bazel` line 231 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1602#-O67Nkl0CM4jO8MfeIqB:-O68ilr6Ebx6yuWXc7V1:b29pyj0) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/48d5b2e06d0837d92effb0c79009ffa342410011/MODULE.bazel#L231)):*

nit: Split out an AWS_SDK_VERSION constant here like we do in common-jvm, with a comment indicating that it matches the common-jvm version.

#1790 will simplify this, but I don't know which PR will land first
</blockquote></details>

Added AWS_SDK_VERSION

sandip80:
*[`src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/remote/aws/EmrServerlessClientImpl.kt` line 122 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1602#-O67xVzA7BgUyPaNwOOG:-O68j0OO-xbqF9kDAym7:b-jjz34w) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/74ce57d26f434f3097e764254589e69f43140262/src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/remote/aws/EmrServerlessClientImpl.kt#L122)):*

can you add a TODO to move this to a config?
</blockquote></details>

Let me move them to AWS Daemon flags

sandip80:
*[`src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/emr/EmrServerlessClientService.kt` line 108 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1602#-Nzk5Z726Nzbrd3PPijW:-O68jM5y0Jo5TytMGUNs:bebrt1w) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/651063c96d1d8a60ab136c3efda5992041ddeb45/src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/emr/EmrServerlessClientService.kt#L108)):*

nit: these should be configurable via command-line options
</blockquote></details>

Will do

sandip80 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1602)* status: 50 of 64 files reviewed, 16 unresolved discussions (waiting on @Marco-Premier, @SanjayVas, and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/emr/EmrExchangeTaskService.kt` line 29 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1602#-Nzi3Ou55hU2umm9E5oA:-O68jB1v6hTsRBh4ZtN2:b-4yalvr) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/651063c96d1d8a60ab136c3efda5992041ddeb45/src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/emr/EmrExchangeTaskService.kt#L29)):*

Still needs a kdoc
</blockquote></details>

Added 1 to the interface

sandip80:

Can you rebase this on head? There's relevant base changes that make this harder to review.
</blockquote></details>

Done.

sandip80:
*[`src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/emr/EmrExchangeTaskService.kt` line 29 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1602#-Nzi3Ou55hU2umm9E5oA:-O68l4wtCTjFwsayfelv:b2hw3yd) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/651063c96d1d8a60ab136c3efda5992041ddeb45/src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/emr/EmrExchangeTaskService.kt#L29)):*
<details><summary><i>Previously, sandip80 (Sandip Samantaray) wrote…</i></summary><blockquote>

Added 1 to the interface
</blockquote></details>

Oh wait, wrong class, will add one here.

stevenwarejones (approved):
> Reviewed 15 of 20 files at r6, 8 of 8 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1602)* status: all files reviewed, 14 unresolved discussions (waiting on @Marco-Premier, @sandip80, and @SanjayVas)

SanjayVas (changes requested):
> Reviewed 1 of 17 files at r1, 1 of 38 files at r2, 14 of 46 files at r4, 6 of 11 files at r5, 11 of 20 files at r6, 6 of 8 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1602)* status: all files reviewed, 29 unresolved discussions (waiting on @Marco-Premier, @sandip80, and @stevenwarejones)

___
*[`build/variables.bzl` line 161 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1602#-O6MwL8Q08NxqY8vSGtB:-O6MwL8Q08NxqY8vSGtC:bi5wwzk) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/bb0c366f400b614021703341726adacd1e83f591/build/variables.bzl#L161)):*
> ```Starlark
>     account_id = "$(aws_account_id)",
>     region = "$(aws_region)",
>     tink_key_uri = "$(aws_kms_key_arn)",
> ```

An ARN is not a KMS key URI. The key URI is something like `gcp-kms://projects/halo-cmm-dev/locations/us-central1/keyRings/mp-test-key-ring/cryptoKeys/mp-test-key` (though for AWS it would start with gcp-aws and use a slightly different pattern).
___
*[`src/main/k8s/panelmatch/dev/example_daemon_aws.cue` line 38 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1602#-O6MxK-z9gKQ1sYBYx6s:-O6MxK-z9gKQ1sYBYx6t:bnrlkm) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/bb0c366f400b614021703341726adacd1e83f591/src/main/k8s/panelmatch/dev/example_daemon_aws.cue#L38)):*
> ```CUE
> #ExchangeDaemonConfig: {
> 	accountId:          string
> 	region:             *"us-west-2" | string
> ```

nit: the default generally goes at the end



_Suggestion:_
```CUE
string | *"us-west-2"
```
___
*[`src/main/k8s/panelmatch/dev/example_daemon_aws.cue` line 94 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1602#-O6N0NoQ-mhDNB9AyOXO:-O6N0NoQ-mhDNB9AyOXP:b-j05cm3) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/bb0c366f400b614021703341726adacd1e83f591/src/main/k8s/panelmatch/dev/example_daemon_aws.cue#L94)):*
> ```CUE
> 		}
> 		_podSpec: _container: {
> 			image:           "\(_exchangeDaemonConfig.accountId).dkr.ecr.\(_exchangeDaemonConfig.region).amazonaws.companel-exchange/aws-example-daemon"
> ```

This should support any registry using the vars from IMAGE_REPOSITORY_SETTINGS. We strongly encourage using our published images in our GitHub container registry, where the image tag is the release version.
___
*[`src/main/k8s/panelmatch/dev/example_daemon_aws.cue` line 109 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1602#-O6MxcHM83Pk8S7HFIIU:-O6MxcHM83Pk8S7HFIIV:b-cuipsi) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/bb0c366f400b614021703341726adacd1e83f591/src/main/k8s/panelmatch/dev/example_daemon_aws.cue#L109)):*
> ```CUE
> 						"--x509-common-name=SomeCommonName",
> 						"--x509-organization=SomeOrganization",
> 						"--x509-dns-name=example.com",
> ```

Like anything else that differs per environment, this should come from make vars.



_Code quote:_
```CUE
						"--x509-common-name=SomeCommonName",
						"--x509-organization=SomeOrganization",
						"--x509-dns-name=example.com",
```
___
*[`src/main/k8s/panelmatch/dev/example_edp_daemon_aws.cue` line 21 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1602#-Nzk0HfIDd_nc6TmXnmF:-O6N1JIP0LkBezKZvrcD:b-zbxepz) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/651063c96d1d8a60ab136c3efda5992041ddeb45/src/main/k8s/panelmatch/dev/example_edp_daemon_aws.cue#L21)):*
> Storage bucket name isn't dynamic so we should be fine to have it here (same as Google cloud).

Storage bucket name will absolutely differ for each environment. On GCS at least, I know that bucket names are globally unique which means they *cannot* be the same for more than one environment. e.g. Halo QA will have a different value than Halo head or Origin staging.
___
*[`src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/remote/aws/EmrServerlessClient.kt` line 21 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1602#-O6N7sJx745s_CbjaEN_:-O6N7sJx745s_CbjaENa:b-xvdi6j) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/bb0c366f400b614021703341726adacd1e83f591/src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/remote/aws/EmrServerlessClient.kt#L21)):*
> ```Kotlin
>  * service.
>  */
> interface EmrServerlessClient {
> ```

KDoc for all public symbols at minimum.
___
*[`src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/remote/aws/EmrServerlessClientImpl.kt` line 45 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1602#-O6N8Bu7EHUsaPm6eUpP:-O6N8Bu7EHUsaPm6eUpQ:b-9tojve) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/bb0c366f400b614021703341726adacd1e83f591/src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/remote/aws/EmrServerlessClientImpl.kt#L45)):*
> ```Kotlin
> private val DEFAULT_POLLING_INTERVAL: Duration = Duration.ofSeconds(10L)
> 
> class EmrServerlessClientImpl(
> ```

Document params
___
*[`src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/remote/aws/EmrServerlessClientImpl.kt` line 51 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1602#-O6N8AF45fsq8bOIZy-L:-O6N8AF45fsq8bOIZy-M:undefined) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/bb0c366f400b614021703341726adacd1e83f591/src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/remote/aws/EmrServerlessClientImpl.kt#L51)):*
> ```Kotlin
>   private val emrServerlessClient: EmrServerlessAsyncClient,
>   private val appStateCheckTimeout: Duration = DEFAULT_APP_STATE_CHECK_TIMEOUT,
>   private val jobExexutionTimeout: Duration = DEFAULT_JOB_EXECUTION_TIMEOUT,
> ```





_Suggestion:_
```Kotlin
jobExecutionTimeout
```
___
*[`src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/remote/aws/EmrServerlessClientImpl.kt` line 61 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1602#-O6N5d4w-IVp1uG4buk1:-O6N5d4w-IVp1uG4buk2:bjugu6j) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/bb0c366f400b614021703341726adacd1e83f591/src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/remote/aws/EmrServerlessClientImpl.kt#L61)):*
> ```Kotlin
>         .type("SPARK")
>         .build()
>     val response = emrServerlessClient.createApplication(request)
> ```

Specify types explicitly where not obvious. See https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/main/docs/code-style.md#kotlin
___
*[`src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/remote/aws/EmrServerlessClientImpl.kt` line 85 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1602#-O6N8OQz7vrkwZL5a4QH:-O6N8OQz7vrkwZL5a4QI:b379gpv) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/bb0c366f400b614021703341726adacd1e83f591/src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/remote/aws/EmrServerlessClientImpl.kt#L85)):*
> ```Kotlin
>     val state = emrServerlessClient.getApplication(request).await().application().state()
> 
>     if (state == ApplicationState.STARTED) {
> ```

This is confusing behavior. The function *sometimes* waits for the application to stop, but not always? Under what condition is the caller expected to get `false` back vs. waiting until it would be `true`?

Perhaps instead make the call and await the response, and optionally expose `waitForApplicationState` if we think the caller may want to choose to wait for a state change. Note that looking at the [SDK example](https://github.com/aws-samples/emr-serverless-samples/blob/main/examples/java-api/myapp/src/main/java/com/example/myapp/EMRServerlessService.java), it doesn't look like the stop operation needs this.
___
*[`src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/remote/aws/EmrServerlessClientImpl.kt` line 100 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1602#-O6N6Fje3O84kolLAyNI:-O6N6Fje3O84kolLAyNJ:b-8twhze) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/bb0c366f400b614021703341726adacd1e83f591/src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/remote/aws/EmrServerlessClientImpl.kt#L100)):*
> ```Kotlin
>   ): Boolean {
> 
>     val startJobReq =
> ```

nit: avoid shortenings or words. Either `startJobRequest`, or even just `request` if it's the only request.



_Suggestion:_
```Kotlin
request
```
___
*[`src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/remote/aws/EmrServerlessClientImpl.kt` line 158 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1602#-O6N6hr95VrFDsLPRLWy:-O6N6hr95VrFDsLPRLWz:b-32u0tm) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/bb0c366f400b614021703341726adacd1e83f591/src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/remote/aws/EmrServerlessClientImpl.kt#L158)):*
> ```Kotlin
>         var state: ApplicationState? = null
>         while (state != finalState) {
>           delay(pollingInterval.toKotlinDuration())
> ```

There's an overload of `delay` that takes in a java.time.Duration. Import and use that.



_Suggestion:_
```Kotlin
delay(pollingInterval)
```
___
*[`src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/remote/aws/EmrServerlessClientImpl.kt` line 164 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1602#-O6N7YDn8cSn2tOhzJNc:-O6N7YDn8cSn2tOhzJNd:b-m9p4sk) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/bb0c366f400b614021703341726adacd1e83f591/src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/remote/aws/EmrServerlessClientImpl.kt#L164)):*
> ```Kotlin
>           state = response.application().state()
> 
>           if (APP_UNEXPECTED_STATES.contains(state))
> ```

Use an exhaustive `when` instead. This way we're forced to update the code if there are new states added.
___
*[`src/main/kotlin/org/wfanet/panelmatch/common/certificates/GenerateCsr.kt` line 26 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1602#-O6NDkWS2bypZqxz2yqU:-O6NDkWS2bypZqxz2yqV:bix26cl) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/bb0c366f400b614021703341726adacd1e83f591/src/main/kotlin/org/wfanet/panelmatch/common/certificates/GenerateCsr.kt#L26)):*
> ```Kotlin
> 
> /**
>  * Generates a PEM format CSR through BouncyCastle API for a given private key, organization and common name.
> ```

This is an implementation detail



_Code quote:_
```Kotlin
through BouncyCastle API
```
___
*[`src/main/kotlin/org/wfanet/panelmatch/common/certificates/GenerateCsr.kt` line 28 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1602#-O6NEGZG8hQMkXHuGWry:-O6NEGZHCELWXWioFIyR:b-cp9k9q) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/bb0c366f400b614021703341726adacd1e83f591/src/main/kotlin/org/wfanet/panelmatch/common/certificates/GenerateCsr.kt#L28)):*
> ```Kotlin
>  * Generates a PEM format CSR through BouncyCastle API for a given private key, organization and common name.
>  */
> fun generateCsrFromPrivateKey(keyPair: KeyPair, organization: String, commonName: String, algorithm: String): String {
> ```

Use `SignatureAlgorithm` enum from common-jvm. You may need another function or extension property to convert to whatever algorithm string that BouncyCastle expects if it's not the Java security API name.



_Code quote:_
```Kotlin
String
```
___
*[`src/main/kotlin/org/wfanet/panelmatch/common/certificates/GenerateCsr.kt` line 35 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1602#-O6NEv6YBHhJzZV9bH29:-O6NEv6YBHhJzZV9bH2A:brcy1vw) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/bb0c366f400b614021703341726adacd1e83f591/src/main/kotlin/org/wfanet/panelmatch/common/certificates/GenerateCsr.kt#L35)):*
> ```Kotlin
>   val pemWriter = JcaPEMWriter(strWriter)
>   pemWriter.writeObject(pkcs10CsrBuilder.build(signer))
>   pemWriter.close()
> ```

The writers implement AutoCloseable, so you should be using the try-with-resources pattern. In Kotlin, this is done with the `use` function.
___
*[`src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/remote/aws/EmrRemoteTaskOrchestrator.kt` line 27 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1602#-O6N3etB-ve0_bOJsaEx:-O6N3etB-ve0_bOJsaEy:b-pgtf6c) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/bb0c366f400b614021703341726adacd1e83f591/src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/remote/aws/EmrRemoteTaskOrchestrator.kt#L27)):*
> ```Kotlin
> import org.wfanet.panelmatch.common.storage.toStringUtf8
> 
> private const val EMR_EXCHANGE_TASK_APP_NAME = "panel-exchange-beam-task"
> ```

New consts should go in objects. In this case, the companion object.

___
*[`src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/remote/aws/EmrRemoteTaskOrchestrator.kt` line 39 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1602#-O6N49q34__2BpPgd9t5:-O6N49q34__2BpPgd9t6:btussea) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/bb0c366f400b614021703341726adacd1e83f591/src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/remote/aws/EmrRemoteTaskOrchestrator.kt#L39)):*
> ```Kotlin
> ) : RemoteTaskOrchestrator {
>   private val appId: String by lazy {
>     runBlocking {
> ```

runBlocking should only be used in one of two cases:

1. Top-level interfacing to coroutine code, e.g. from `main()`.
2. Tests

This is called from coroutine-aware code (i.e. the suspend function `orchestrateTask`), meaning it definitely should not switch back to blocking code and then coroutine code again. You really want this run in the same coroutine scope as the calling function.

`lazy` is not really suitable for this case. If you want coroutine-aware memoization, you'll have to implement it yourself. e.g. having this be a `lateinit var` that you conditionally initialize in all calls to `orchestrateTask`.

PR #1606, opened 2024-05-03
Use service config default when setting timeouts on integration tests

jcorilla:


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1606)

Issue #1530, opened 2024-03-20
Switch to AutoBazelRepository for runfiles when rules_kotlin is fixed


SanjayVas:
There are many call sites for the deprecated `org.wfanet.measurement.common.getRuntimePath` function. The suggested fix depends on a fix for https://github.com/bazelbuild/rules_kotlin/issues/1095 being released.

PR #1610, opened 2024-05-08
Add Support for different Privacy Params for Single and Multiple Data Providers

tristanvuong2021:


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1610)

kungfucraig (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1610)* status: 0 of 31 files reviewed, 1 unresolved discussion (waiting on @tristanvuong2021)

___
*[`src/main/proto/wfa/measurement/reporting/v2alpha/metric.proto` line 54 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1610#-NxOlVp28LaNPL_fw5Dx:-NxOlVp28LaNPL_fw5Dy:b1pxxlg) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d08978ab3113b23cd4935f692748ba5817cc1179/src/main/proto/wfa/measurement/reporting/v2alpha/metric.proto#L54)):*
> ```proto
>   }
> 
>   message Params {
> ```

Maybe just have a separate Params structure for R/F?

The stray field is weird.

tristanvuong2021 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1610)* status: 0 of 31 files reviewed, 1 unresolved discussion (waiting on @kungfucraig)

___
*[`src/main/proto/wfa/measurement/reporting/v2alpha/metric.proto` line 54 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1610#-NxOlVp28LaNPL_fw5Dx:-NxPMQ677FUA3xml9GQr:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d08978ab3113b23cd4935f692748ba5817cc1179/src/main/proto/wfa/measurement/reporting/v2alpha/metric.proto#L54)):*

Maybe just have a separate Params structure for R/F?

The stray field is weird.
</blockquote></details>

Done.

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1610)* status: 0 of 31 files reviewed, 4 unresolved discussions (waiting on @kungfucraig, @riemanli, and @tristanvuong2021)

___
*[`src/main/proto/wfa/measurement/reporting/v2alpha/metric.proto` line 54 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1610#-NxUFv_WCYPSe8C2DuWn:-NxUFv_WCYPSe8C2DuWo:b-j6g7u8) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c5e73aa3d289830478ab0646036d8bc919a779f2/src/main/proto/wfa/measurement/reporting/v2alpha/metric.proto#L54)):*
> ```proto
>   }
> 
>   message Params {
> ```

Mark the body fields of the new Params types as REQUIRED.
___
*[`src/main/proto/wfa/measurement/reporting/v2alpha/metric.proto` line 59 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1610#-NxUGSt98J3Zb-R6rku2:-NxUGSt98J3Zb-R6rku3:b-g22wpt) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c5e73aa3d289830478ab0646036d8bc919a779f2/src/main/proto/wfa/measurement/reporting/v2alpha/metric.proto#L59)):*
> ```proto
>   }
> 
>   message FrequencyParams {
> ```

nit: not the biggest fan of this name, but can't come up with anything that's objectively better

Maybe call the other one `ScalarParams` and this one `ScalarAndHistogramParams`, since the first is used for all metrics with a single scalar result?



_Code quote:_
```proto
FrequencyParams
```
___
*[`src/main/proto/wfa/measurement/reporting/v2alpha/metric.proto` line 60 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1610#-NxUGh6_4buVinczZ_Uo:-NxUGh6_4buVinczZ_Up:b-yonuxv) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c5e73aa3d289830478ab0646036d8bc919a779f2/src/main/proto/wfa/measurement/reporting/v2alpha/metric.proto#L60)):*
> ```proto
> 
>   message FrequencyParams {
>     DifferentialPrivacyParams privacy_params = 1;
> ```

Might as well be explicit here.



_Suggestion:_
```proto
reach_privacy_params
```

kungfucraig (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1610)* status: 0 of 31 files reviewed, 4 unresolved discussions (waiting on @riemanli, @SanjayVas, and @tristanvuong2021)

___
*[`src/main/proto/wfa/measurement/reporting/v2alpha/metric.proto` line 59 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1610#-NxUGSt98J3Zb-R6rku2:-NxUKEix4yANlLHjdW2y:ba90p0b) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c5e73aa3d289830478ab0646036d8bc919a779f2/src/main/proto/wfa/measurement/reporting/v2alpha/metric.proto#L59)):*

nit: not the biggest fan of this name, but can't come up with anything that's objectively better

Maybe call the other one `ScalarParams` and this one `ScalarAndHistogramParams`, since the first is used for all metrics with a single scalar result?
</blockquote></details>

I really don't like "FrequencyParams."

Maybe: SamplingAndPrivacyParams and SamplingAndPrivacyParamsForReachAndFrequency?

Won't quite win a camel case award, but getting there.

tristanvuong2021 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1610)* status: 0 of 31 files reviewed, 4 unresolved discussions (waiting on @kungfucraig, @riemanli, and @SanjayVas)

___
*[`src/main/proto/wfa/measurement/reporting/v2alpha/metric.proto` line 59 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1610#-NxUGSt98J3Zb-R6rku2:-NxY-vGABQKanP0TBpsC:b9iv0hl) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/b47b775f69cb6cf2134e9489c5dbc3c656ef3f87/src/main/proto/wfa/measurement/reporting/v2alpha/metric.proto#L59)):*

I really don't like "FrequencyParams."

Maybe: SamplingAndPrivacyParams and SamplingAndPrivacyParamsForReachAndFrequency?

Won't quite win a camel case award, but getting there.
</blockquote></details>

I went with your naming suggestion
___
*[`src/main/proto/wfa/measurement/reporting/v2alpha/metric.proto` line 60 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1610#-NxUGh6_4buVinczZ_Uo:-NxY-v2o0LSwHyCqWlhz:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/b47b775f69cb6cf2134e9489c5dbc3c656ef3f87/src/main/proto/wfa/measurement/reporting/v2alpha/metric.proto#L60)):*

Might as well be explicit here.
</blockquote></details>

Done.

SanjayVas (changes requested):
Taking this in passes since it's a big PR.

> Reviewed 3 of 31 files at r1, 4 of 21 files at r4.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1610#-:-NxZ-3v5CYEHKs04pNQV:b1ev6nv)* status: 7 of 31 files reviewed, 5 unresolved discussions (waiting on @riemanli and @tristanvuong2021)

___
*[`src/main/proto/wfa/measurement/reporting/v2alpha/metric.proto` line 59 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1610#-NxUGSt98J3Zb-R6rku2:-NxUHcsu4FTzzlqYiux_:b-88ee8n) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/b47b775f69cb6cf2134e9489c5dbc3c656ef3f87/src/main/proto/wfa/measurement/reporting/v2alpha/metric.proto#L59)):*

lgtm

I will defer to Sanjay on the rest of the PR. If you end up needing a final approval from me at the end let me know.
</blockquote></details>

Just one nit on the new names: avoid "For" and use a prefix instead. `ReachAndFrequencySamplingAndPrivacyParams`.
___
*[`src/main/resources/reporting/postgres/add-additional-metric-spec-columns-to-metrics.sql` line 40 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1610#-NxYxHig6-SaRsudI9Eb:-NxYxHig6-SaRsudI9Ec:b-hv84m3) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d5b5977649890e39bdd0d2d64bfecc53fe68755f/src/main/resources/reporting/postgres/add-additional-metric-spec-columns-to-metrics.sql#L40)):*
> ```TSQL
> 
> 
>   DifferentialPrivacyEpsilon DOUBLE PRECISION NOT NULL,
> ```

Err, why does this changeset just start with a list of columns?
___
*[`src/test/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricCalculationSpecsServiceTest.kt` line 1440 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1610#-NxYxkgC-HuPXeTyH_54:-NxYxkgC-HuPXeTyH_55:b-7m0iqm) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d5b5977649890e39bdd0d2d64bfecc53fe68755f/src/test/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricCalculationSpecsServiceTest.kt#L1440)):*
> ```Kotlin
> 
>     private val SECRET_FILES_PATH: Path = run {
>       val runfiles =
> ```

Continue to use the (deprecated) common-jvm functions for Runfiles until #1530 is fixed.
___
*[`src/test/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricSpecDefaultsTest.kt` line 215 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1610#-NxYzj0s7lGi7sVCmKRC:-NxYzj0s7lGi7sVCmKRD:b4o383s) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d5b5977649890e39bdd0d2d64bfecc53fe68755f/src/test/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricSpecDefaultsTest.kt#L215)):*
> ```Kotlin
> 
>   @Test
>   fun `buildMetricSpec builds reach spec when fields filled in multiple and single edp fields`() {
> ```

nit: the tests like this which are basically checking that `withDefaults` does not override specified fields should have test method names that more clearly indicate that.
___
*[`src/test/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricSpecDefaultsTest.kt` line 1500 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1610#-NxYyc7ZAgzANy-mfXdm:-NxYyc7_3qeUszqBgbpt:b-o0rglq) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d5b5977649890e39bdd0d2d64bfecc53fe68755f/src/test/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricSpecDefaultsTest.kt#L1500)):*
> ```Kotlin
>     // Metric Specs
> 
>     private val OLD_EMPTY_REACH_METRIC_SPEC: MetricSpec = metricSpec {
> ```

nit: LEGACY for the old messages that still need to be handled, and no prefix for the new ones



_Suggestion:_
```Kotlin
LEGACY_EMPTY
```

tristanvuong2021 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1610)* status: 4 of 32 files reviewed, 3 unresolved discussions (waiting on @riemanli and @SanjayVas)

___
*[`src/main/resources/reporting/postgres/add-additional-metric-spec-columns-to-metrics.sql` line 40 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1610#-NxYxHig6-SaRsudI9Eb:-NxZAJnE6oNXMYr7Erha:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d5b5977649890e39bdd0d2d64bfecc53fe68755f/src/main/resources/reporting/postgres/add-additional-metric-spec-columns-to-metrics.sql#L40)):*

Err, why does this changeset just start with a list of columns?
</blockquote></details>

Done.
___
*[`src/test/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricCalculationSpecsServiceTest.kt` line 1440 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1610#-NxYxkgC-HuPXeTyH_54:-NxZDs3X3ZLdZlVYBe-C:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d5b5977649890e39bdd0d2d64bfecc53fe68755f/src/test/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricCalculationSpecsServiceTest.kt#L1440)):*

Continue to use the (deprecated) common-jvm functions for Runfiles until #1530 is fixed.
</blockquote></details>

Done.

SanjayVas (changes requested):
> Reviewed 1 of 14 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1610)* status: 5 of 32 files reviewed, 4 unresolved discussions (waiting on @riemanli and @tristanvuong2021)

___
*[`src/main/resources/reporting/postgres/add-is-single-data-provider-column-to-measurements.sql` line 40 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1610#-NxZKqgj5ct5mnCf0IfW:-NxZKqgj5ct5mnCf0IfX:bjz0lsg) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/9125a155411fdad6b963e022bbb6f04417106fcf/src/main/resources/reporting/postgres/add-is-single-data-provider-column-to-measurements.sql#L40)):*
> ```TSQL
> 
> -- changeset tristanvuong2021:add-is-single-data-provider-col-measurements-table dbms:postgresql
> ALTER TABLE Measurements ADD COLUMN IsSingleDataProvider boolean NOT NULL DEFAULT FALSE;
> ```

1. Does this need to be a separate column rather than in the serialized Details? e.g. is it used in any queries?
2. What do you think about having it be DataProviderCount instead?

tristanvuong2021 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1610)* status: 4 of 32 files reviewed, 4 unresolved discussions (waiting on @riemanli and @SanjayVas)

___
*[`src/main/proto/wfa/measurement/reporting/v2alpha/metric.proto` line 54 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1610#-NxUFv_WCYPSe8C2DuWn:-Nxn_1SmDpeJXRQERyCp:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/b47b775f69cb6cf2134e9489c5dbc3c656ef3f87/src/main/proto/wfa/measurement/reporting/v2alpha/metric.proto#L54)):*

Mark the body fields of the new Params types as REQUIRED.
</blockquote></details>

Done.
___
*[`src/main/resources/reporting/postgres/add-is-single-data-provider-column-to-measurements.sql` line 40 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1610#-NxZKqgj5ct5mnCf0IfW:-Nxn_mOo16z-Ip-zwRwZ:b6dwo1i) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/9125a155411fdad6b963e022bbb6f04417106fcf/src/main/resources/reporting/postgres/add-is-single-data-provider-column-to-measurements.sql#L40)):*

1. Does this need to be a separate column rather than in the serialized Details? e.g. is it used in any queries?
2. What do you think about having it be DataProviderCount instead?
</blockquote></details>

1. It is not used in any queries, but because I set details later when measurements complete, it is easier to not put it details.
2. Is it because we might have other cases besides single and multiple Data Providers in the future.

tristanvuong2021 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1610)* status: 2 of 35 files reviewed, 4 unresolved discussions (waiting on @riemanli and @SanjayVas)

___
*[`src/main/resources/reporting/postgres/add-is-single-data-provider-column-to-measurements.sql` line 40 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1610#-NxZKqgj5ct5mnCf0IfW:-Ny6Nvz9Dum0QpaMf5mM:b-xw3fbm) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/9125a155411fdad6b963e022bbb6f04417106fcf/src/main/resources/reporting/postgres/add-is-single-data-provider-column-to-measurements.sql#L40)):*

1. It is not used in any queries, but because I set details later when measurements complete, it is easier to not put it details.
2. Is it because we might have other cases besides single and multiple Data Providers in the future.
</blockquote></details>

Used your suggestions.

tristanvuong2021 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1610)* status: 6 of 35 files reviewed, 4 unresolved discussions (waiting on @riemanli and @SanjayVas)

___
*[`src/main/proto/wfa/measurement/reporting/v2alpha/metric.proto` line 47 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1610#-NyMVseH5FzSBRNW40Vw:-NyMwCOhDzFaZmNK8hUf:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/9a33c0269d9e11ec56547a84951cd260df34c952/src/main/proto/wfa/measurement/reporting/v2alpha/metric.proto#L47)):*

Is there a behavioral difference between `0` and unset here? If so, it should be documented here. If not, then do not use the `optional` keyword.
</blockquote></details>

Done.

SanjayVas (changes requested):
> Reviewed 2 of 20 files at r7, 1 of 2 files at r8.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1610)* status: 9 of 35 files reviewed, 7 unresolved discussions (waiting on @riemanli and @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricSpecDefaults.kt` line 677 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1610#-NyNPPkj2A2qHNR2xTFb:-NyNPPkj2A2qHNR2xTFc:b7r59sm) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4a96fbd1b5dc2583687c1378626186db8f43eccc/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricSpecDefaults.kt#L677)):*
> ```Kotlin
>     val maxStart = NUM_BUCKETS - (source.randomStart.width * NUM_BUCKETS).toInt()
>     // The `- 1` is in case the rounding from source.randomStart.width * NUM_BUCKETS rounds down.
>     // This
> ```

nit: wrapping
___
*[`src/main/proto/wfa/measurement/config/reporting/metric_spec_config.proto` line 34 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1610#-NyNQ7CH79bv0AjHkOnt:-NyNQ7CH79bv0AjHkOnu:b-7ia43g) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4a96fbd1b5dc2583687c1378626186db8f43eccc/src/main/proto/wfa/measurement/config/reporting/metric_spec_config.proto#L34)):*
> ```proto
>   // Computer Science, 9(3-4), pp.211-407."
>   message DifferentialPrivacyParams {
>     optional double epsilon = 1;
> ```

Nothing should be optional in the config. All fields must be set, meaning unset is equivalent to setting to `0`.

We don't need backwards compatibility in the config like we do for the API message. The incompatible change just needs to be made clear in the release notes (and therefore in the PR description).
___
*[`src/main/proto/wfa/measurement/reporting/v2alpha/metric.proto` line 47 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1610#-NyNUuv5Dj-o1iwDaVym:-NyNUuv5Dj-o1iwDaVyn:b-3i3pir) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4a96fbd1b5dc2583687c1378626186db8f43eccc/src/main/proto/wfa/measurement/reporting/v2alpha/metric.proto#L47)):*
> ```proto
>   //
>   // If any of the fields are not specified in a request message, the service
>   // implementation will choose values.
> ```

Just to confirm, we want to allow you to e.g. specify width but not start to get the default start but the specified width? Does that actually make sense?



_Code quote:_
```proto
  // If any of the fields are not specified in a request message, the service
  // implementation will choose values.
```
___
*[`src/main/proto/wfa/measurement/reporting/v2alpha/metric.proto` line 54 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1610#-NyNVPJVASTjOYxCCZRf:-NyNVPJW6VXsnP2RdV3M:b-unopvi) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4a96fbd1b5dc2583687c1378626186db8f43eccc/src/main/proto/wfa/measurement/reporting/v2alpha/metric.proto#L54)):*
> ```proto
>     //
>     // start + width cannot be larger than 1
>     optional float width = 2;
> ```

To clarify, this means that 0 is a valid width. Is that correct?



_Code quote:_
```proto
optional 
```

SanjayVas (changes requested):
> Reviewed 2 of 4 files at r9.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1610)* status: 9 of 35 files reviewed, 3 unresolved discussions (waiting on @riemanli and @tristanvuong2021)

___
*[`src/main/proto/wfa/measurement/reporting/v2alpha/metric.proto` line 47 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1610#-NyNUuv5Dj-o1iwDaVym:-NyRUTCREf_lR63VWPbQ:b9y9nl7) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4a96fbd1b5dc2583687c1378626186db8f43eccc/src/main/proto/wfa/measurement/reporting/v2alpha/metric.proto#L47)):*

Changed it so it uses defaults for both or none.
</blockquote></details>

The clearer way to do that is to mark width as required and have the `VidSamplingInterval` fields be optional. This way you can just check the presence of the `VidSamplingInterval` field.
___
*[`src/main/proto/wfa/measurement/reporting/v2alpha/metric.proto` line 54 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1610#-NyNVPJVASTjOYxCCZRf:-NyRUj8v4cCaD_fRIJcj:bo91l9v) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4a96fbd1b5dc2583687c1378626186db8f43eccc/src/main/proto/wfa/measurement/reporting/v2alpha/metric.proto#L54)):*

I added the comment to clarify it, but I realized I am not sure about the valid interval for epsilon and delta. We do have a check in the Kingdom Measurements service, but we don't document it in the API proto.
</blockquote></details>

To reiterate my statement: if this field has the `optional` keyword, it is implying that `0` is a valid value. The way to indicate that `0` is not valid is to specify that the field behavior is `REQUIRED`.

tristanvuong2021 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1610)* status: 8 of 35 files reviewed, 1 unresolved discussion (waiting on @riemanli and @SanjayVas)

___
*[`src/main/proto/wfa/measurement/reporting/v2alpha/metric.proto` line 47 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1610#-NyNUuv5Dj-o1iwDaVym:-NyRk2eLCqL9-FxrV5gF:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4a96fbd1b5dc2583687c1378626186db8f43eccc/src/main/proto/wfa/measurement/reporting/v2alpha/metric.proto#L47)):*

The clearer way to do that is to mark width as required and have the `VidSamplingInterval` fields be optional. This way you can just check the presence of the `VidSamplingInterval` field.
</blockquote></details>

Done.
___
*[`src/main/proto/wfa/measurement/reporting/v2alpha/metric.proto` line 54 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1610#-NyNVPJVASTjOYxCCZRf:-NyRkRg-EVzsrXwpjB89:b-96law2) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4a96fbd1b5dc2583687c1378626186db8f43eccc/src/main/proto/wfa/measurement/reporting/v2alpha/metric.proto#L54)):*

To reiterate my statement: if this field has the `optional` keyword, it is implying that `0` is a valid value. The way to indicate that `0` is not valid is to specify that the field behavior is `REQUIRED`.
</blockquote></details>

That is not what i wanted to confirm here. I wanted to confirm if it is still the case that delta can be 0, but epsilon can't. That is what we have in code, but we don't document it.

SanjayVas (changes requested):
> Reviewed 1 of 3 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1610)* status: 9 of 35 files reviewed, 5 unresolved discussions (waiting on @riemanli and @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricSpecDefaults.kt` line 25 at r10](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1610#-NyRw0n-AuC0jRX6EZop:-NyRw0n-AuC0jRX6EZoq:b-ubgeby) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/60a8e70fde0898a3033a5d3a5fbf9abf49f229f4/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricSpecDefaults.kt#L25)):*
> ```Kotlin
> import org.wfanet.measurement.reporting.v2alpha.copy
> 
> private const val NUM_BUCKETS = 10000
> ```

Where does this come from and what does it mean? Document it.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricSpecDefaults.kt` line 293 at r10](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1610#-NyS4bSX5-MQu5AcrT42:-NyS4bSY5iCZQWQh8zqH:bfi923x) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/60a8e70fde0898a3033a5d3a5fbf9abf49f229f4/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricSpecDefaults.kt#L293)):*
> ```Kotlin
>   secureRandom: Random,
> ): MetricSpec.ReachParams {
>   if (this.hasMultipleDataProviderParams() && this.hasSingleDataProviderParams()) {
> ```

nit: personally I think this would be clearer as

```kotlin
if (hasMultipleDataProviderParams() || hasSingleDataProviderParams()) {
  if (hasMultipleDataProviderParams() != hasSingleDataProviderParams()) {
    // Throw exception for not having both multi and single set ...
  }
  // Validate the subfields of single and multi ...
} else {
  if (!hasPrivacyParams()) {
    // Throw exception for not having any privacy params set ...
  }
  // Validate the subfields of privacy params.
}
```
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricSpecDefaults.kt` line 355 at r10](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1610#-NyS5bfB9chR2ZLyArXz:-NyS5bfB9chR2ZLyArY-:b9qjg5g) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/60a8e70fde0898a3033a5d3a5fbf9abf49f229f4/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricSpecDefaults.kt#L355)):*
> ```Kotlin
>         }
>     } else {
>       val source = this
> ```

I believe you need this line outside of the `copy` lambda.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricSpecDefaults.kt` line 371 at r10](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1610#-NyS5kGq0cIIIBc3ZcVq:-NyS5kGq0cIIIBc3ZcVr:bekmga7) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/60a8e70fde0898a3033a5d3a5fbf9abf49f229f4/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricSpecDefaults.kt#L371)):*
> ```Kotlin
>           vidSamplingInterval.validate()
>         }
>       clearPrivacyParams()
> ```

Don't you also need to set singleDataProviderParams to be the same as multipleDataProviderParams in this case?
___
*[`src/main/proto/wfa/measurement/reporting/v2alpha/metric.proto` line 54 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1610#-NyNVPJVASTjOYxCCZRf:-NyRuyOHCigZacaTsP14:b-xopf69) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4a96fbd1b5dc2583687c1378626186db8f43eccc/src/main/proto/wfa/measurement/reporting/v2alpha/metric.proto#L54)):*

That is not what i wanted to confirm here. I wanted to confirm if it is still the case that delta can be 0, but epsilon can't. That is what we have in code, but we don't document it.
</blockquote></details>

Privacy params are a separate discussion. Marking this resolvedw.r.t. sampling interval, as `width` now has field_behavior=REQUIRED and no `optional` keyword.
___
*[`src/main/proto/wfa/measurement/reporting/v2alpha/metric.proto` line 47 at r10](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1610#-NyRuVRN94WptQy0Xmbc:-NyRuVRN94WptQy0Xmbd:b-cvh2gl) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/60a8e70fde0898a3033a5d3a5fbf9abf49f229f4/src/main/proto/wfa/measurement/reporting/v2alpha/metric.proto#L47)):*
> ```proto
>   //
>   // If both start and width are not specified, the service implementation will
>   // choose values for both start and width.
> ```

This comment is inaccurate. Any comment about the defaults should go on the fields that use this message type, and are based on whether the field itself is set.



_Code quote:_
```proto
  //
  // If both start and width are not specified, the service implementation will
  // choose values for both start and width.
```

tristanvuong2021 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1610)* status: 8 of 35 files reviewed, 4 unresolved discussions (waiting on @riemanli and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricSpecDefaults.kt` line 25 at r10](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1610#-NyRw0n-AuC0jRX6EZop:-NySJWx11Om3AIupca0k:bpxq4dj) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/60a8e70fde0898a3033a5d3a5fbf9abf49f229f4/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricSpecDefaults.kt#L25)):*

Where does this come from and what does it mean? Document it.
</blockquote></details>

This is to help with the math. I just removed it since it is only ever used in one function.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricSpecDefaults.kt` line 355 at r10](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1610#-NyS5bfB9chR2ZLyArXz:-NySLwU31c66GAZf1qYr:bct368) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/60a8e70fde0898a3033a5d3a5fbf9abf49f229f4/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricSpecDefaults.kt#L355)):*

I believe you need this line outside of the `copy` lambda.
</blockquote></details>

It doesn't change how it works, but I moved it for clarity.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricSpecDefaults.kt` line 371 at r10](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1610#-NyS5kGq0cIIIBc3ZcVq:-NySM6e82QYPWSujeerr:b-jcrqql) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/60a8e70fde0898a3033a5d3a5fbf9abf49f229f4/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricSpecDefaults.kt#L371)):*

Don't you also need to set singleDataProviderParams to be the same as multipleDataProviderParams in this case?
</blockquote></details>

Right now, I have it set up to only set multiple and when creating the measurements, if only multiple is available, it uses that for single edp as well. And when retrieving the metric spec, if only multiple is set, it sets the legacy privacy_params.
___
*[`src/main/proto/wfa/measurement/reporting/v2alpha/metric.proto` line 47 at r10](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1610#-NyRuVRN94WptQy0Xmbc:-NySM6255mvQt_ooO7mP:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/60a8e70fde0898a3033a5d3a5fbf9abf49f229f4/src/main/proto/wfa/measurement/reporting/v2alpha/metric.proto#L47)):*

This comment is inaccurate. Any comment about the defaults should go on the fields that use this message type, and are based on whether the field itself is set.
</blockquote></details>

Done.

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1610)* status: 8 of 35 files reviewed, 2 unresolved discussions (waiting on @riemanli and @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricSpecDefaults.kt` line 371 at r10](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1610#-NyS5kGq0cIIIBc3ZcVq:-NySS056Eu8dWg2cCQjp:bb7zdl4) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/60a8e70fde0898a3033a5d3a5fbf9abf49f229f4/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricSpecDefaults.kt#L371)):*

Right now, I have it set up to only set multiple and when creating the measurements, if only multiple is available, it uses that for single edp as well. And when retrieving the metric spec, if only multiple is set, it sets the legacy privacy_params.
</blockquote></details>

So this is to persist the fact that a legacy request set the legacy params? I'm not entirely certain that's necessary, but I guess it's fine.

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1610)* status: 8 of 35 files reviewed, 2 unresolved discussions (waiting on @riemanli and @tristanvuong2021)

___
*[`src/main/proto/wfa/measurement/reporting/v2alpha/metric.proto` line 58 at r11](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1610#-NySkULT9AAWRgMG9Pbw:-NySkULT9AAWRgMG9Pbx:b-krod4w) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a6d7dee24704208245a7718786beb79571b7ffd8/src/main/proto/wfa/measurement/reporting/v2alpha/metric.proto#L58)):*
> ```proto
> 
>     // If both start and width are not specified, the service implementation
>     // will choose values for both start and width.
> ```

This is not quite accurate. Again, it's based upon the field presence. If the field is present, then all of the subfield field behavior annotations apply. Therefore it's sufficient to say "If not specified, the service implementation will choose a default value."

Also every field comment should start with a summary fragment describing the field.



_Code quote:_
```proto
    // If both start and width are not specified, the service implementation
    // will choose values for both start and width.
```

tristanvuong2021 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1610)* status: 8 of 35 files reviewed, 1 unresolved discussion (waiting on @riemanli and @SanjayVas)

___
*[`src/main/proto/wfa/measurement/reporting/v2alpha/metric.proto` line 58 at r11](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1610#-NySkULT9AAWRgMG9Pbw:-NyTAKXdCWzwEky9Fm26:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a6d7dee24704208245a7718786beb79571b7ffd8/src/main/proto/wfa/measurement/reporting/v2alpha/metric.proto#L58)):*

This is not quite accurate. Again, it's based upon the field presence. If the field is present, then all of the subfield field behavior annotations apply. Therefore it's sufficient to say "If not specified, the service implementation will choose a default value."

Also every field comment should start with a summary fragment describing the field.
</blockquote></details>

Done.

SanjayVas (changes requested):
> Reviewed 1 of 31 files at r1, 2 of 9 files at r2, 1 of 17 files at r3, 4 of 21 files at r4, 2 of 14 files at r6, 8 of 20 files at r7, 1 of 2 files at r11, 1 of 1 files at r12, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1610)* status: 28 of 35 files reviewed, 2 unresolved discussions (waiting on @riemanli and @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/measurementconsumer/stats/MeasurementStatistics.kt` line 93 at r12](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1610#-NyW7AvV46B8GcW_pF_w:-NyW7AvV46B8GcW_pF_x:b3mdk53) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e103761db8c000402e0f47e012ac6b6e6dda3d0b/src/main/kotlin/org/wfanet/measurement/measurementconsumer/stats/MeasurementStatistics.kt#L93)):*
> ```Kotlin
>  * reach_result +/- 1.96 * reach_std.
>  */
> private const val REACH_THRESHLEGACY_CONSTANT_FOR_RELATIVE_FREQUENCY_VARIANCE = 1.96
> ```

I think your find/replace was a little aggressive here.
___
*[`src/main/proto/wfa/measurement/internal/reporting/v2/metric.proto` line 47 at r12](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1610#-NyW-HxABcpfYykHhjSO:-NyW-HxABcpfYykHhjSP:b-dx5d4o) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e103761db8c000402e0f47e012ac6b6e6dda3d0b/src/main/proto/wfa/measurement/internal/reporting/v2/metric.proto#L47)):*
> ```proto
> 
>   message ReachParams {
>     SamplingAndPrivacyParams multiple_data_provider_params = 2;
> ```

Mark the previously used field numbers as `reserved`. See https://protobuf.dev/programming-guides/proto3/#fieldreserved

tristanvuong2021 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1610)* status: 26 of 35 files reviewed, 2 unresolved discussions (waiting on @riemanli and @SanjayVas)

___
*[`src/main/proto/wfa/measurement/internal/reporting/v2/metric.proto` line 47 at r12](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1610#-NyW-HxABcpfYykHhjSO:-NyaPrRs3CLdC3J8vCV6:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e103761db8c000402e0f47e012ac6b6e6dda3d0b/src/main/proto/wfa/measurement/internal/reporting/v2/metric.proto#L47)):*

Mark the previously used field numbers as `reserved`. See https://protobuf.dev/programming-guides/proto3/#fieldreserved
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/measurementconsumer/stats/MeasurementStatistics.kt` line 93 at r12](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1610#-NyW7AvV46B8GcW_pF_w:-NyaPs1rEPUe_xe934wB:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e103761db8c000402e0f47e012ac6b6e6dda3d0b/src/main/kotlin/org/wfanet/measurement/measurementconsumer/stats/MeasurementStatistics.kt#L93)):*

I think your find/replace was a little aggressive here.
</blockquote></details>

Done.

SanjayVas (approved):
> Reviewed 2 of 14 files at r6, 4 of 20 files at r7, 1 of 3 files at r10, 2 of 2 files at r13, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1610)* status: all files reviewed, 1 unresolved discussion (waiting on @riemanli and @tristanvuong2021)

___
*[`src/main/proto/wfa/measurement/internal/reporting/v2/metric.proto` line 52 at r13](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1610#-NyafxtrCtfd2ROp1-3q:-NyafxtsDnAzUqHFAiu4:b-837go4) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f0cdb593eb4c0b87c6b121ee5013ec4a22ab4832/src/main/proto/wfa/measurement/internal/reporting/v2/metric.proto#L52)):*
> ```proto
>   }
>   message ReachAndFrequencyParams {
>     reserved = 1;
> ```

nit: you can technicaly do this as `reserved 1,2;`

stevenwarejones (changes requested):
> Reviewed 1 of 31 files at r1, 2 of 9 files at r2, 1 of 17 files at r3, 5 of 21 files at r4, 3 of 14 files at r6, 16 of 20 files at r7, 1 of 3 files at r10, 1 of 2 files at r11, 1 of 1 files at r12, 1 of 2 files at r13, 2 of 2 files at r14, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1610)* status: all files reviewed, 1 unresolved discussion (waiting on @riemanli and @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricSpecDefaults.kt` line 703 at r14](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1610#-NzA0kHL8VtlrbDBWB_g:-NzA0kHL8VtlrbDBWB_h:b2w4iu4) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/b6658882dcd3a6115eecf25d59c91e4e84d731dd/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricSpecDefaults.kt#L703)):*
> ```Kotlin
>     // The `- 1` is in case the rounding from source.randomStart.width * NUM_BUCKETS rounds down.
>     // This prevents the random start from being too big if rounding down does occur.
>     val randomStart = secureRandom.nextInt(maxStart - 1)
> ```

assuming we support wraparound (which we will with HMShuffle), why not just use a random float 0 < x < 1?

kungfucraig (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1610)* status: all files reviewed, 1 unresolved discussion (waiting on @riemanli and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricSpecDefaults.kt` line 703 at r14](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1610#-NzA0kHL8VtlrbDBWB_g:-NzAIwhe5IRNGaw7xdau:b32v57) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/b6658882dcd3a6115eecf25d59c91e4e84d731dd/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricSpecDefaults.kt#L703)):*

assuming we support wraparound (which we will with HMShuffle), why not just use a random float 0 < x < 1?
</blockquote></details>

+1

kungfucraig (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1610)* status: all files reviewed, 1 unresolved discussion (waiting on @riemanli and @stevenwarejones)

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1610)* status: all files reviewed, 2 unresolved discussions (waiting on @riemanli and @tristanvuong2021)

___
*[`src/main/proto/wfa/measurement/config/reporting/metric_spec_config.proto` line 76 at r14](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1610#-NzTyryV-KFeV8eUWCSN:-NzTyryV-KFeV8eUWCSO:b-7ox0v1) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7c82d515fc59d34857a88f8853a0a63205f9f702/src/main/proto/wfa/measurement/config/reporting/metric_spec_config.proto#L76)):*
> ```proto
> 
>     // Parameters for a single DataProvider.
>     SamplingAndPrivacyParams single_data_provider_params = 2;
> ```

is there a deck on why an MC would want different single vs multiple dp params?

kungfucraig (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1610)* status: all files reviewed, 2 unresolved discussions (waiting on @riemanli and @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/config/reporting/metric_spec_config.proto` line 76 at r14](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1610#-NzTyryV-KFeV8eUWCSN:-NzYiJrb2g0jZlIxzq0F:bjc33on) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7c82d515fc59d34857a88f8853a0a63205f9f702/src/main/proto/wfa/measurement/config/reporting/metric_spec_config.proto#L76)):*

is there a deck on why an MC would want different single vs multiple dp params?
</blockquote></details>

This basically comes down to have different privacy params for MPC vs. direct measurements. It was supported in Report v1 and didn't get translated into Report v2 correct. This PR is correcting a design flaw that was discovered by Accenture a few weeks back. There's no document or deck per se outside of the slack thread I tagged you on.

stevenwarejones (approved):
> Reviewed 1 of 4 files at r9, 1 of 1 files at r15, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1610)* status: all files reviewed, 1 unresolved discussion (waiting on @riemanli)

PR #1632, opened 2024-05-28
Add executable binary classes and docker images for HMSS.

renjiezh:


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1632)

ple13 (approved):
> Reviewed 46 of 46 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1632)* status: all files reviewed, 3 unresolved discussions (waiting on @renjiezh and @SanjayVas)

___
*[`src/main/k8s/kingdom.cue` line 65 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1632#-NzEv1IBE3eO7OmSP_Jd:-NzEv1IBE3eO7OmSP_Je:bnl4sv0) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/eb9e4e230d34925e1584d6dbeae997a4a0dd8f77/src/main/k8s/kingdom.cue#L65)):*
> ```CUE
> 	_hmssRfEnableFlag:                       "--enable-hmss-for-rf"
> 	_hmssReachEnableFlag:                    "--enable-hmss-for-reach"
> 	_hmssProtocolConfigConfig:               "--hmss-protocol-config-config=/var/run/secrets/files/hmss_protocol_config_config.textproto"
> ```

nit: can we use the same format as in other lines?



_Code quote:_
```CUE
	_hmssRfEnableFlag:                       "--enable-hmss-for-rf"
	_hmssReachEnableFlag:                    "--enable-hmss-for-reach"
	_hmssProtocolConfigConfig:               "--hmss-protocol-config-config=/var/run/secrets/files/hmss_protocol_config_config.textproto"
```
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/daemon/herald/HeraldDaemon.kt` line 90 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1632#-NzUQYag5UHfOuRIQn5f:-NzUQYag5UHfOuRIQn5g:b-7qfnis) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/eb9e4e230d34925e1584d6dbeae997a4a0dd8f77/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/daemon/herald/HeraldDaemon.kt#L90)):*
> ```Kotlin
>     description = ["The key encryption key file (binary format) used for private key store."],
>   )
>   var keyEncryptionKeyTinkFile: File? = null
> ```

nit: Do we need `private set` to make it read-only?
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/daemon/mill/shareshuffle/HonestMajorityShareShuffleMillDaemon.kt` line 140 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1632#-NzUXa4Y25mvm8OLflYg:-NzUXa4Z7RhMYzUb6lrz:b-pb76uj) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/eb9e4e230d34925e1584d6dbeae997a4a0dd8f77/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/daemon/mill/shareshuffle/HonestMajorityShareShuffleMillDaemon.kt#L140)):*
> ```Kotlin
>       flags.csCertificateDerFile.inputStream().use { input -> readCertificate(input) }
>     val csCertificate = Certificate(flags.csCertificateName, csX509Certificate)
>     // TODO: Read from a KMS-encrypted store instead.
> ```

nit: use the correct format for TODO (See https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/bb94176d3b3a925e054eaee272fa151e587721c8/docs/code-style.md?plain=1#L167).

SanjayVas (changes requested):
> Reviewed 11 of 46 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1632)* status: all files reviewed, 6 unresolved discussions (waiting on @renjiezh)

___
*[`src/main/k8s/duchy.cue` line 133 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1632#-Nz_bX5-0GFmfuYW9CJc:-Nz_bX5-0GFmfuYW9CJd:bsmip6e) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/eb9e4e230d34925e1584d6dbeae997a4a0dd8f77/src/main/k8s/duchy.cue#L133)):*
> ```CUE
> 						_kingdom_system_api_target_flag,
> 						_kingdom_system_api_cert_host_flag,
> 						if (_duchyKeyEncryptionKeyFile != _|_) {_duchyKeyEncryptionKeyFileFlag},
> ```

This conditional doesn't make sense, as this field appears to be required.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/daemon/herald/HeraldDaemon.kt` line 146 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1632#-Nz_e52tApgTcaS-n-S_:-Nz_e52tApgTcaS-n-Sa:byi2lr0) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a4ca481f014409ff414d446008c5037a601a2e7f/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/daemon/herald/HeraldDaemon.kt#L146)):*
> ```Kotlin
>   val heraldId = System.getenv("HOSTNAME")
> 
>   // TODO(@renjiez): Use real PrivateKeyStore when enabling HMSS.
> ```

Why was the TODO dropped? At what point pre-launch are we going to swap this over to a real KMS client? IMO at this point, FakeKmsClient should only be used from ForwardedStorageHeraldDaemon.
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/MeasurementsService.kt` line 106 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1632#-Nz_cUZA2yoMBGU7RRDw:-Nz_cUZA2yoMBGU7RRDx:br2erxp) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/eb9e4e230d34925e1584d6dbeae997a4a0dd8f77/src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/MeasurementsService.kt#L106)):*
> ```Kotlin
>   private val noiseMechanisms: List<NoiseMechanism>,
>   private val reachOnlyLlV2Enabled: Boolean = false,
>   private val reachAndFrequencyHmssEnabled: Boolean = false,
> ```

Is this just being extra cautious, e.g. giving us the flexibility to disable HMSS for the REACH Measurement type? I would assume that if HMSS works at all, it works for both measurement types. This is because it's a single protocol.

Note that this is different than `reachOnlyLlv2Enabled`, which controls whether the separate RO LLv2 protocol is used. That is to say it does not control whether LLv2 can be used for the REACH measurement type. As a result, we may want to use different naming and some extra documentation here.

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1632)* status: 43 of 46 files reviewed, 4 unresolved discussions (waiting on @ple13 and @SanjayVas)

___
*[`src/main/k8s/duchy.cue` line 133 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1632#-Nz_bX5-0GFmfuYW9CJc:-NzeJP4GC7ayGZXgzG-X:b-3vsz0q) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/eb9e4e230d34925e1584d6dbeae997a4a0dd8f77/src/main/k8s/duchy.cue#L133)):*

This conditional doesn't make sense, as this field appears to be required.
</blockquote></details>

It is required only for worker1 and worker2. Aggregator does not need it.
___
*[`src/main/k8s/kingdom.cue` line 65 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1632#-NzEv1IBE3eO7OmSP_Jd:-NzeNfnG48Uvc8LsTSrp:bjtwlc7) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/eb9e4e230d34925e1584d6dbeae997a4a0dd8f77/src/main/k8s/kingdom.cue#L65)):*

nit: can we use the same format as in other lines?
</blockquote></details>

Do you mean camelCase vs all_lowercase? We will stick to camelCase forward.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/daemon/herald/HeraldDaemon.kt` line 146 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1632#-Nz_e52tApgTcaS-n-S_:-NzeMkV-09FJfUnbfpgd:btvmn70) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a4ca481f014409ff414d446008c5037a601a2e7f/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/daemon/herald/HeraldDaemon.kt#L146)):*

Why was the TODO dropped? At what point pre-launch are we going to swap this over to a real KMS client? IMO at this point, FakeKmsClient should only be used from ForwardedStorageHeraldDaemon.
</blockquote></details>

Added it back. Will turn back to kms after reach-only HMSS.
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/MeasurementsService.kt` line 106 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1632#-Nz_cUZA2yoMBGU7RRDw:-NzeLkFE1Eszo8k-4V0T:brpcy4k) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/eb9e4e230d34925e1584d6dbeae997a4a0dd8f77/src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/MeasurementsService.kt#L106)):*

Is this just being extra cautious, e.g. giving us the flexibility to disable HMSS for the REACH Measurement type? I would assume that if HMSS works at all, it works for both measurement types. This is because it's a single protocol.

Note that this is different than `reachOnlyLlv2Enabled`, which controls whether the separate RO LLv2 protocol is used. That is to say it does not control whether LLv2 can be used for the REACH measurement type. As a result, we may want to use different naming and some extra documentation here.
</blockquote></details>

We are implementing reach-only HMSS now. The current HMSS is primarily targeting R/F though it works for reach-only (with too much noise). I leave this flexibility for now. Once reach-only implemented, I will merge these two options.
A TODO added.

SanjayVas (changes requested):
> Reviewed 7 of 46 files at r1, 1 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1632)* status: 44 of 46 files reviewed, 6 unresolved discussions (waiting on @ple13 and @renjiezh)

___
*[`src/main/k8s/duchy.cue` line 133 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1632#-Nz_bX5-0GFmfuYW9CJc:-NzeYI1BCogM-pLzMLxN:b-j4p61f) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/eb9e4e230d34925e1584d6dbeae997a4a0dd8f77/src/main/k8s/duchy.cue#L133)):*

It is required only for worker1 and worker2. Aggregator does not need it.
</blockquote></details>

What I mean is that the field isn't listed as optional, and is also used in a another non-optional field.

You'll want to declare it with a question mark to indicate it's optional, and then do the same for the derived flag field (wrapping it in a conditional).
___
*[`src/main/k8s/dev/duchy_eks.cue` line 42 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1632#-Nze_1Xt1xlFuRWhynsw:-Nze_1Xt1xlFuRWhynsx:bhhqqjj) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6d0cde50c2018ff3ef60e0b62c8b3eadb45e7a18/src/main/k8s/dev/duchy_eks.cue#L42)):*
> ```CUE
> 	requests: {
> 		cpu:    "25m"
> 		memory: "1024Mi"
> ```

Why are these values different on EKS vs. GKE? Is it that S3 client takes different memory than GCS client?
___
*[`src/main/k8s/dev/duchy_gke.cue` line 49 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1632#-Nze_TY-164QAqlSzfsp:-Nze_TY-164QAqlSzfsq:b-msozxb) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6d0cde50c2018ff3ef60e0b62c8b3eadb45e7a18/src/main/k8s/dev/duchy_gke.cue#L49)):*
> ```CUE
> 	}
> }
> #HeraldMaxHeapSize:        "400M"
> ```

I assume this memory increase is due to having a storage client?
___
*[`src/main/k8s/dev/duchy_gke.cue` line 59 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1632#-Nze_fX28OpkM8U87w5U:-Nze_fX28OpkM8U87w5V:b-i8nsa3) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6d0cde50c2018ff3ef60e0b62c8b3eadb45e7a18/src/main/k8s/dev/duchy_gke.cue#L59)):*
> ```CUE
> 	}
> }
> #MillMaxHeapSize:                 "5G"
> ```

This seems quite significant. Is this because of the HMSS mill needing the VID index map? Can we instead set the HMSS and LLv2 mills independently?
___
*[`src/main/k8s/dev/duchy_gke.cue` line 70 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1632#-Nzea0Gc2N5BkVgmqgCD:-Nzea0Gc2N5BkVgmqgCE:b-74ai9a) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6d0cde50c2018ff3ef60e0b62c8b3eadb45e7a18/src/main/k8s/dev/duchy_gke.cue#L70)):*
> ```CUE
> 	}
> }
> #FulfillmentMaxHeapSize: "350M"
> ```

Why did this get so much bigger? Is it the VID index map thing again?

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1632)* status: 38 of 50 files reviewed, 5 unresolved discussions (waiting on @ple13 and @SanjayVas)

___
*[`src/main/k8s/duchy.cue` line 133 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1632#-Nz_bX5-0GFmfuYW9CJc:-NzoQBX_Dn16Sq-NF-Ex:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/eb9e4e230d34925e1584d6dbeae997a4a0dd8f77/src/main/k8s/duchy.cue#L133)):*

What I mean is that the field isn't listed as optional, and is also used in a another non-optional field.

You'll want to declare it with a question mark to indicate it's optional, and then do the same for the derived flag field (wrapping it in a conditional).
</blockquote></details>

Done.
___
*[`src/main/k8s/dev/duchy_eks.cue` line 42 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1632#-Nze_1Xt1xlFuRWhynsw:-NziyWPl0-tWxwZ58C6J:bexgzxq) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6d0cde50c2018ff3ef60e0b62c8b3eadb45e7a18/src/main/k8s/dev/duchy_eks.cue#L42)):*

Why are these values different on EKS vs. GKE? Is it that S3 client takes different memory than GCS client?
</blockquote></details>

Updated to keep them consistent. Now heralds use 0.5 GB.
___
*[`src/main/k8s/dev/duchy_gke.cue` line 49 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1632#-Nze_TY-164QAqlSzfsp:-NzixZpDEPmJQ3ofpGu_:b-4p689m) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6d0cde50c2018ff3ef60e0b62c8b3eadb45e7a18/src/main/k8s/dev/duchy_gke.cue#L49)):*

I assume this memory increase is due to having a storage client?
</blockquote></details>

Yes. And due to the computation details of HMSS,  the memory usage is lifted for a bit.
___
*[`src/main/k8s/dev/duchy_gke.cue` line 59 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1632#-Nze_fX28OpkM8U87w5U:-Nziyb0K8-MOQSOij8CF:b-pb6nro) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6d0cde50c2018ff3ef60e0b62c8b3eadb45e7a18/src/main/k8s/dev/duchy_gke.cue#L59)):*

This seems quite significant. Is this because of the HMSS mill needing the VID index map? Can we instead set the HMSS and LLv2 mills independently?
</blockquote></details>

Good idea. HMSS requires more memory but less CPU. They should've have different setting.
___
*[`src/main/k8s/dev/duchy_gke.cue` line 70 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1632#-Nzea0Gc2N5BkVgmqgCD:-Nziy2AY-uZ4K7M9CV_Y:b-jrchjn) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6d0cde50c2018ff3ef60e0b62c8b3eadb45e7a18/src/main/k8s/dev/duchy_gke.cue#L70)):*

Why did this get so much bigger? Is it the VID index map thing again?
</blockquote></details>

The fulfillment(sketch1) size of HMSS is fixed by population and is larger. It is not about vid index map.

SanjayVas (changes requested):
> Reviewed 16 of 46 files at r1, 1 of 3 files at r2, 9 of 11 files at r3, 2 of 2 files at r4, 2 of 2 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1632)* status: 49 of 50 files reviewed, 3 unresolved discussions (waiting on @ple13 and @renjiezh)

___
*[`src/main/k8s/duchy.cue` line 42 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1632#-NzoYsHz-X1xfGEHb2tN:-NzoYsI-5mtGxuIqhEOI:b-vs10ic) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7ecb073ab0b53d6054b21c7ea7d338daeec971e6/src/main/k8s/duchy.cue#L42)):*
> ```CUE
> 	_cs_cert_resource_name:      _duchy.cs_cert_resource_name
> 	_duchyKeyEncryptionKeyFile?: string
> 	if (_duchy.duchyKeyEncryptionKeyFile != _|_) {
> ```

You shouldn't need the conditional here since it's setting an optional field from an optional field.
___
*[`src/main/k8s/dev/duchy_eks.cue` line 42 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1632#-Nzo_KJv462gZ1S8SeVV:-Nzo_KJv462gZ1S8SeVW:b-5frsm4) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7ecb073ab0b53d6054b21c7ea7d338daeec971e6/src/main/k8s/dev/duchy_eks.cue#L42)):*
> ```CUE
> 	requests: {
> 		cpu:    "25m"
> 		memory: "512M"
> ```

Incorrect units. K8s uses "Mi" for mebibyte, whereas JVM flags use "M".



_Suggestion:_
```CUE
memory: "512Mi"
```
___
*[`src/main/k8s/dev/duchy_gke.cue` line 70 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1632#-Nzea0Gc2N5BkVgmqgCD:-Nzo_tyREZzGHA5Zqy0A:b-i9wh4o) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6d0cde50c2018ff3ef60e0b62c8b3eadb45e7a18/src/main/k8s/dev/duchy_gke.cue#L70)):*

The fulfillment(sketch1) size of HMSS is fixed by population and is larger. It is not about vid index map.
</blockquote></details>

Ah, I mistakenly assumed that HMSS sketches would actually be smaller than LLv2 for UK-sized population.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/daemon/mill/shareshuffle/HonestMajorityShareShuffleMillFlags.kt` line 28 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1632#-NzoayQY9G7GJF8Uj8BF:-NzoayQZ8vdTehNL6GCY:byf4800) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7ecb073ab0b53d6054b21c7ea7d338daeec971e6/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/daemon/mill/shareshuffle/HonestMajorityShareShuffleMillFlags.kt#L28)):*
> ```Kotlin
> import picocli.CommandLine
> 
> class HonestMajorityShareShuffleMillFlags {
> ```

nit: is it worth factoring out the common Mill flags?

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1632)* status: 46 of 50 files reviewed, 3 unresolved discussions (waiting on @ple13 and @SanjayVas)

___
*[`src/main/k8s/duchy.cue` line 42 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1632#-NzoYsHz-X1xfGEHb2tN:-NzolPsz156qyuScuIS1:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7ecb073ab0b53d6054b21c7ea7d338daeec971e6/src/main/k8s/duchy.cue#L42)):*

You shouldn't need the conditional here since it's setting an optional field from an optional field.
</blockquote></details>

Done.
___
*[`src/main/k8s/dev/duchy_eks.cue` line 42 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1632#-Nzo_KJv462gZ1S8SeVV:-NzohvVsCLwkK5JwPIwS:b-8vntgq) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7ecb073ab0b53d6054b21c7ea7d338daeec971e6/src/main/k8s/dev/duchy_eks.cue#L42)):*

Incorrect units. K8s uses "Mi" for mebibyte, whereas JVM flags use "M".
</blockquote></details>

Done.
Good to know. Thanks!
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/daemon/mill/shareshuffle/HonestMajorityShareShuffleMillFlags.kt` line 28 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1632#-NzoayQY9G7GJF8Uj8BF:-Nzoi_EJC8VJA3kkTBBG:b-429ofa) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7ecb073ab0b53d6054b21c7ea7d338daeec971e6/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/daemon/mill/shareshuffle/HonestMajorityShareShuffleMillFlags.kt#L28)):*

nit: is it worth factoring out the common Mill flags?
</blockquote></details>

Added a TODO for that.

ple13 (approved):
> Reviewed 2 of 3 files at r2, 10 of 11 files at r3, 2 of 2 files at r4, 2 of 2 files at r5, 4 of 4 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1632)* status: all files reviewed, 3 unresolved discussions (waiting on @SanjayVas)

SanjayVas (approved):
> Reviewed 5 of 46 files at r1, 4 of 4 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1632)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @stevenwarejones)

stevenwarejones (changes requested):
> Reviewed 34 of 46 files at r1, 2 of 3 files at r2, 7 of 11 files at r3, 1 of 2 files at r4, 2 of 2 files at r5, 4 of 4 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1632)* status: all files reviewed, 1 unresolved discussion (waiting on @renjiezh)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/daemon/mill/shareshuffle/HonestMajorityShareShuffleMillDaemon.kt` line 139 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1632#-O-6Vbb4ErbAOtw31QtE:-O-6Vbb4ErbAOtw31QtF:b-spml12) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/83428bd8214b76c3795ef9e06f19b88096d3edb8/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/daemon/mill/shareshuffle/HonestMajorityShareShuffleMillDaemon.kt#L139)):*
> ```Kotlin
>     // This will be the name of the pod when deployed to Kubernetes. Note that the millId is
>     // included in mill logs to help debugging.
>     val millId = System.getenv("HOSTNAME")
> ```

Are there any scenarios where someone would want to override this? eg make this a flag with a default value you have here?

SanjayVas (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1632)* status: all files reviewed, 1 unresolved discussion (waiting on @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/daemon/mill/shareshuffle/HonestMajorityShareShuffleMillDaemon.kt` line 139 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1632#-O-6Vbb4ErbAOtw31QtE:-O-7YaA8EIAnfH-Rot9N:b-nzvmr2) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/83428bd8214b76c3795ef9e06f19b88096d3edb8/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/daemon/mill/shareshuffle/HonestMajorityShareShuffleMillDaemon.kt#L139)):*

Are there any scenarios where someone would want to override this? eg make this a flag with a default value you have here?
</blockquote></details>

This is copied from the existing Mill code. I left this same comment on one of Matthew's PRs requesting a TODO to convert this to a command line option.

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1632)* status: 49 of 50 files reviewed, 1 unresolved discussion (waiting on @ple13, @SanjayVas, and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/daemon/mill/shareshuffle/HonestMajorityShareShuffleMillDaemon.kt` line 139 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1632#-O-6Vbb4ErbAOtw31QtE:-O-8DubxEL6ReSpcZP2O:bwpue5f) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/83428bd8214b76c3795ef9e06f19b88096d3edb8/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/daemon/mill/shareshuffle/HonestMajorityShareShuffleMillDaemon.kt#L139)):*

This is copied from the existing Mill code. I left this same comment on one of Matthew's PRs requesting a TODO to convert this to a command line option.
</blockquote></details>

Added a TODO for it.
FYI, while using kubernetes, by default HOSTNAME is the pod name.

stevenwarejones (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1632)* status: all files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

renjiezh (commented):
> Reviewed 34 of 46 files at r1, 2 of 3 files at r2, 6 of 11 files at r3, 1 of 2 files at r4, 2 of 2 files at r5, 4 of 4 files at r6, 1 of 1 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1632)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @renjiezh)

stevenwarejones (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1632)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @renjiezh)

PR #1634, opened 2024-05-29
Add logs for duchy requisition fulfillment service.

renjiezh:


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1634)

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1634)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @renjiezh)

*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1634#-Nz_MPut-Juoae780R_h:-Nz_MPut-Juoae780R_i:b-vcou3a) (no related file):*
I'd rather use our interceptors so we get consistent logging, including trace IDs. My understanding is that we don't tend to have the interceptor installed for this service due to the fact that logging all the payloads isn't terribly useful.

Can we instead modify LoggingServerInterceptor to allow us to skip payload logging? I'm thinking something like checking the ServerCall object to see if the method type is CLIENT_STREAMING and then only logging the first request message.

___

PR #1635, opened 2024-05-31
Split reach and frequency privacy budget for HMSS protocol.

ple13:


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1635)

kungfucraig (changes requested):
> Reviewed 10 of 13 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1635)* status: 12 of 13 files reviewed, 2 unresolved discussions (waiting on @ple13, @renjiezh, and @SanjayVas)

___
*[`src/main/cc/wfa/measurement/internal/duchy/protocol/share_shuffle/honest_majority_share_shuffle_utility.cc` line 388 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1635#-NzFNdNpCaxymiWRa71H:-NzFNGw9ENS54q1P8jRW:b-a35ajx) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5bcb502359ff90700294485ace231fc3b8f177ac/src/main/cc/wfa/measurement/internal/duchy/protocol/share_shuffle/honest_majority_share_shuffle_utility.cc#L388)):*
> ```cpp
>   // approach will have a lot more noise.
>   int64_t non_empty_register_count =
>       (frequency_histogram.find(0) == frequency_histogram.end())
> ```

This seems like a bug in the original code more than a change required by the splitting of the R/F noise types. Do you agree? Just want to make sure I'm not missing something.
___
*[`src/test/cc/wfa/measurement/internal/duchy/protocol/share_shuffle/honest_majority_share_shuffle_utility_test.cc` line 1275 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1635#-NzTXv5P5ZRQdybM-EGW:-NzTXv5P5ZRQdybM-EGX:bgnwotl) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5bcb502359ff90700294485ace231fc3b8f177ac/src/test/cc/wfa/measurement/internal/duchy/protocol/share_shuffle/honest_majority_share_shuffle_utility_test.cc#L1275)):*
> ```cpp
>                        test_data.RunReachAndFrequencyAggregationPhase());
>   ASSERT_OK_AND_ASSIGN(int64_t expected_reach,
>                        EstimateReach(8.0, kVidSamplingIntervalWidth));
> ```

I don't understand why this would need to change, can you help me understand?

ple13 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1635)* status: 12 of 13 files reviewed, 2 unresolved discussions (waiting on @kungfucraig, @renjiezh, and @SanjayVas)

___
*[`src/main/cc/wfa/measurement/internal/duchy/protocol/share_shuffle/honest_majority_share_shuffle_utility.cc` line 388 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1635#-NzFNdNpCaxymiWRa71H:-NzTi9A5Dp2KJjr8l-ge:b-7digh3) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5bcb502359ff90700294485ace231fc3b8f177ac/src/main/cc/wfa/measurement/internal/duchy/protocol/share_shuffle/honest_majority_share_shuffle_utility.cc#L388)):*

This seems like a bug in the original code more than a change required by the splitting of the R/F noise types. Do you agree? Just want to make sure I'm not missing something.
</blockquote></details>

Yes, this is not related to the splitting of the R/F noise. According to c++ standard (https://en.cppreference.com/w/cpp/container/map/operator_at) when the key 0 doesn't exist, 0 is inserted into the map and the value for f[0] is assigned with the default initialized value for the declared value type, which is 0 for int. In my opinion, this works but it could potentially be an issue as the implementation depends on the behavior of the standard library. That's why I decided to change the code to remove that dependency.
___
*[`src/test/cc/wfa/measurement/internal/duchy/protocol/share_shuffle/honest_majority_share_shuffle_utility_test.cc` line 1275 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1635#-NzTXv5P5ZRQdybM-EGW:-NzThHe55EfP0xji1kPD:bvlvyjm) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5bcb502359ff90700294485ace231fc3b8f177ac/src/test/cc/wfa/measurement/internal/duchy/protocol/share_shuffle/honest_majority_share_shuffle_utility_test.cc#L1275)):*

I don't understand why this would need to change, can you help me understand?
</blockquote></details>

In the original test, we use the same (eps, delta) for both reach and frequency, and the offset for all f[i] is 2. As a result, the adjusted f[0] after removing the offset is 6 - 2 = 4. In the new test, we use (eps/10, delta) for reach, and (eps, delta) for frequency. The offset for reach is now 6 instead, making the adjusted f[0] to be 6 - 6 = 0. The non empty register count becomes (8 - 0) = 8 (it was 8 - 4 = 4 before).

kungfucraig (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1635)* status: 12 of 13 files reviewed, 1 unresolved discussion (waiting on @ple13, @renjiezh, and @SanjayVas)

___
*[`src/main/cc/wfa/measurement/internal/duchy/protocol/share_shuffle/honest_majority_share_shuffle_utility.cc` line 388 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1635#-NzFNdNpCaxymiWRa71H:-NzTmlZSBye18cVM6sSJ:bsftkw5) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5bcb502359ff90700294485ace231fc3b8f177ac/src/main/cc/wfa/measurement/internal/duchy/protocol/share_shuffle/honest_majority_share_shuffle_utility.cc#L388)):*

Yes, this is not related to the splitting of the R/F noise. According to c++ standard (https://en.cppreference.com/w/cpp/container/map/operator_at) when the key 0 doesn't exist, 0 is inserted into the map and the value for f[0] is assigned with the default initialized value for the declared value type, which is 0 for int. In my opinion, this works but it could potentially be an issue as the implementation depends on the behavior of the standard library. That's why I decided to change the code to remove that dependency.
</blockquote></details>

This is a good change, just wanted to confirm my understanding.
___
*[`src/test/cc/wfa/measurement/internal/duchy/protocol/share_shuffle/honest_majority_share_shuffle_utility_test.cc` line 1275 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1635#-NzTXv5P5ZRQdybM-EGW:-NzTnUKJ6nhKUzJ55mFt:b-jy1y67) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5bcb502359ff90700294485ace231fc3b8f177ac/src/test/cc/wfa/measurement/internal/duchy/protocol/share_shuffle/honest_majority_share_shuffle_utility_test.cc#L1275)):*

In the original test, we use the same (eps, delta) for both reach and frequency, and the offset for all f[i] is 2. As a result, the adjusted f[0] after removing the offset is 6 - 2 = 4. In the new test, we use (eps/10, delta) for reach, and (eps, delta) for frequency. The offset for reach is now 6 instead, making the adjusted f[0] to be 6 - 6 = 0. The non empty register count becomes (8 - 0) = 8 (it was 8 - 4 = 4 before).
</blockquote></details>

I feel like the expected_reach should not be computed, but should be explicitly coded into the test. Mainly because it allows us to spot check that the expected value makes sense when reading the code. Wdyt?

ple13 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1635)* status: 12 of 13 files reviewed, 1 unresolved discussion (waiting on @kungfucraig, @renjiezh, and @SanjayVas)

___
*[`src/test/cc/wfa/measurement/internal/duchy/protocol/share_shuffle/honest_majority_share_shuffle_utility_test.cc` line 1275 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1635#-NzTXv5P5ZRQdybM-EGW:-Nzcufs76cJ3KDTUEHTQ:bqi2x6n) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5bcb502359ff90700294485ace231fc3b8f177ac/src/test/cc/wfa/measurement/internal/duchy/protocol/share_shuffle/honest_majority_share_shuffle_utility_test.cc#L1275)):*

I feel like the expected_reach should not be computed, but should be explicitly coded into the test. Mainly because it allows us to spot check that the expected value makes sense when reading the code. Wdyt?
</blockquote></details>

Yeah, I agree. I've made the computation of the expected reach explicit so that it's easier to verify the tests.

kungfucraig (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1635)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @renjiezh and @SanjayVas)

SanjayVas (approved):
> Reviewed 10 of 13 files at r1, 2 of 2 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1635)* status: all files reviewed, 2 unresolved discussions (waiting on @ple13, @renjiezh, and @stevenwarejones)

___
*[`src/main/cc/wfa/measurement/internal/duchy/protocol/share_shuffle/honest_majority_share_shuffle_utility.cc` line 268 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1635#-Nzdf8du9OZBhHohCMsX:-Nzdf8dv-Pjy-bds572L:b-s61xlv) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/dfe9d212e25caf99413149acaba4ed57aa0b1858/src/main/cc/wfa/measurement/internal/duchy/protocol/share_shuffle/honest_majority_share_shuffle_utility.cc#L268)):*
nit: Only use auto when it assists readability, not just to avoid writing an explicit type.

From https://google.github.io/styleguide/cppguide.html#Type_deduction: 

> The fundamental rule is: use type deduction only to make the code clearer or safer, and do not use it merely to avoid the inconvenience of writing an explicit type. When judging whether the code is clearer, keep in mind that your readers are not necessarily on your team, or familiar with your project, so types that you and your reviewer experience as unnecessary clutter will very often provide useful information to others. For example, you can assume that the return type of `make_unique<Foo>()` is obvious, but the return type of `MyWidgetFactory()` probably isn't.



_Code quote:_
```cpp
auto
```
___
*[`src/main/proto/wfa/measurement/internal/duchy/protocol/honest_majority_share_shuffle_methods.proto` line 32 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1635#-Nzdh8Ik08t26cQoEw4I:-Nzdh8Ik08t26cQoEw4J:b-mahsas) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/dfe9d212e25caf99413149acaba4ed57aa0b1858/src/main/proto/wfa/measurement/internal/duchy/protocol/honest_majority_share_shuffle_methods.proto#L32)):*
> ```proto
> // called via kotlin or JNI in the Mill.
> 
> message HmssDifferentialPrivacyParams {
> ```

nit: I don't think this is necessarily specific to HMSS. That is to say that other (future) protocols may want the same params.

I'm also not entirely certain of the value of splitting this out into a separate message type, given that it's only used in two request messages which themselves are protocol-specific.

stevenwarejones (approved):
> Reviewed 10 of 13 files at r1, 2 of 2 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1635)* status: all files reviewed, 2 unresolved discussions (waiting on @ple13 and @renjiezh)

renjiezh (commented):
> Reviewed 5 of 13 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1635)* status: all files reviewed, 2 unresolved discussions (waiting on @ple13 and @SanjayVas)

___
*[`src/main/proto/wfa/measurement/internal/duchy/protocol/honest_majority_share_shuffle_methods.proto` line 32 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1635#-Nzdh8Ik08t26cQoEw4I:-Nzejldd47fqpaB5L9oo:b-4ti5sn) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/dfe9d212e25caf99413149acaba4ed57aa0b1858/src/main/proto/wfa/measurement/internal/duchy/protocol/honest_majority_share_shuffle_methods.proto#L32)):*

nit: I don't think this is necessarily specific to HMSS. That is to say that other (future) protocols may want the same params.

I'm also not entirely certain of the value of splitting this out into a separate message type, given that it's only used in two request messages which themselves are protocol-specific.
</blockquote></details>

For LLv2, we have similar dp params in a separated file. That one contains 4 dp_params and the noise mechanism. I think it is fine to have reach_dp_params and frequency_dp_params directly in request.

ple13 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1635)* status: 8 of 13 files reviewed, 1 unresolved discussion (waiting on @kungfucraig, @renjiezh, @SanjayVas, and @stevenwarejones)

___
*[`src/main/cc/wfa/measurement/internal/duchy/protocol/share_shuffle/honest_majority_share_shuffle_utility.cc` line 268 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1635#-Nzdf8du9OZBhHohCMsX:-NzdpPbR2Bc42sL6Gh9w:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/dfe9d212e25caf99413149acaba4ed57aa0b1858/src/main/cc/wfa/measurement/internal/duchy/protocol/share_shuffle/honest_majority_share_shuffle_utility.cc#L268)):*

nit: Only use auto when it assists readability, not just to avoid writing an explicit type.

From https://google.github.io/styleguide/cppguide.html#Type_deduction: 

> The fundamental rule is: use type deduction only to make the code clearer or safer, and do not use it merely to avoid the inconvenience of writing an explicit type. When judging whether the code is clearer, keep in mind that your readers are not necessarily on your team, or familiar with your project, so types that you and your reviewer experience as unnecessary clutter will very often provide useful information to others. For example, you can assume that the return type of `make_unique<Foo>()` is obvious, but the return type of `MyWidgetFactory()` probably isn't.
</blockquote></details>

Done.
___
*[`src/main/proto/wfa/measurement/internal/duchy/protocol/honest_majority_share_shuffle_methods.proto` line 32 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1635#-Nzdh8Ik08t26cQoEw4I:-Nzi1ydl52VRvN-vtVwJ:bj6a67q) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/dfe9d212e25caf99413149acaba4ed57aa0b1858/src/main/proto/wfa/measurement/internal/duchy/protocol/honest_majority_share_shuffle_methods.proto#L32)):*

For LLv2, we have similar dp params in a separated file. That one contains 4 dp_params and the noise mechanism. I think it is fine to have reach_dp_params and frequency_dp_params directly in request.
</blockquote></details>

It's true that this is not specific to HMSS. It's probably better to have a separate proto for it. However, as @SanjayVas pointed out, we only have 2 references, so it's not necessary to have a separate message. The change needed to remove the HmssDifferentialPrivacyParams message is minimal, so I just did it.

renjiezh (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1635)* status: 8 of 13 files reviewed, all discussions resolved (waiting on @kungfucraig, @SanjayVas, and @stevenwarejones)

kungfucraig (approved):
> Reviewed 4 of 5 files at r4, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1635)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @SanjayVas and @stevenwarejones)

SanjayVas (approved):
> Reviewed 4 of 5 files at r4, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1635)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @stevenwarejones)

stevenwarejones (commented):
> Reviewed 4 of 5 files at r4, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1635)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #1638, opened 2024-06-04
Add FulfillRequisitionRequestBuilder

kungfucraig:


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1638)

SanjayVas (changes requested):
> Reviewed 2 of 9 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1638)* status: 2 of 9 files reviewed, 1 unresolved discussion (waiting on @kungfucraig)

___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/v2alpha/FulfillRequisitionRequestBuilder.kt` line 55 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1638#-O-2MIUQ7COpFJVZKARv:-O-2MIUQ7COpFJVZKARw:b-2bcwvy) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8189aee883da31f110e7873314c159bf85de216c/src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/v2alpha/FulfillRequisitionRequestBuilder.kt#L55)):*
> ```Kotlin
>  * @throws IllegalArgumentException if the requisition is malformed or the frequency vector is empty
>  */
> class FulfillRequisitionRequestBuilder(
> ```

Is this meant to be part of an EDP library? If that's the case, don't we want that library to be callable from Java? If so, then it can't use Flows as that's only usable from Kotlin w/grpc-kotlin.

kungfucraig (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1638)* status: 2 of 9 files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/v2alpha/FulfillRequisitionRequestBuilder.kt` line 55 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1638#-O-2MIUQ7COpFJVZKARv:-O-2N5ig884mApwZT8dY:bcrdaq8) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8189aee883da31f110e7873314c159bf85de216c/src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/v2alpha/FulfillRequisitionRequestBuilder.kt#L55)):*

Is this meant to be part of an EDP library? If that's the case, don't we want that library to be callable from Java? If so, then it can't use Flows as that's only usable from Kotlin w/grpc-kotlin.
</blockquote></details>

All of our existing clients are using Kotlin, so I'd prefer we use a Flow. We can always add a wrapper method in the future if we need it. If you feel strongly about it, I can implement the wrapper too.

Whatever we decide here will also apply to the iterator that is already merged, and which uses a Flow.

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1638)* status: 2 of 9 files reviewed, 1 unresolved discussion (waiting on @kungfucraig)

___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/v2alpha/FulfillRequisitionRequestBuilder.kt` line 55 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1638#-O-2MIUQ7COpFJVZKARv:-O-2Q_Bl5Iq8-w4h748C:b-54w2u) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8189aee883da31f110e7873314c159bf85de216c/src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/v2alpha/FulfillRequisitionRequestBuilder.kt#L55)):*
> Whatever we decide here will also apply to the iterator that is already merged, and which uses a Flow.

Can you point to this? If you're talking about the VidIndexMapEntry iterator, that does not use a Flow.

From what I can tell, the entire share vector is in memory at once. Therefore there's not really any difference in this case between outputting a Flow vs. outputting a Kotlin Sequence or Java Stream. It would be different if the share vector could actually be generated in a streaming fashion.

A Flow cannot be easily converted to a grpc-java StreamObserver, but a Sequence can be trivially converted to either a Flow or a Java Stream. Therefore I'd recommend using Sequence for Kotlin callers and perhaps adding a method that calls the `asStream()` Kotlin extension for Java callers.

kungfucraig (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1638)* status: 2 of 9 files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/v2alpha/FulfillRequisitionRequestBuilder.kt` line 55 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1638#-O-2MIUQ7COpFJVZKARv:-O-2h41H8JXQwLKldBI0:b-47btzy) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8189aee883da31f110e7873314c159bf85de216c/src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/v2alpha/FulfillRequisitionRequestBuilder.kt#L55)):*

> Whatever we decide here will also apply to the iterator that is already merged, and which uses a Flow.

Can you point to this? If you're talking about the VidIndexMapEntry iterator, that does not use a Flow.

From what I can tell, the entire share vector is in memory at once. Therefore there's not really any difference in this case between outputting a Flow vs. outputting a Kotlin Sequence or Java Stream. It would be different if the share vector could actually be generated in a streaming fashion.

A Flow cannot be easily converted to a grpc-java StreamObserver, but a Sequence can be trivially converted to either a Flow or a Java Stream. Therefore I'd recommend using Sequence for Kotlin callers and perhaps adding a method that calls the `asStream()` Kotlin extension for Java callers.
</blockquote></details>

See: https://sourcegraph.com/github.com/world-federation-of-advertisers/cross-media-measurement/-/blob/src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/v2alpha/InMemoryVidIndexMap.kt?L139

My main point is that everyone who is using this uses Kotlin, and that we can add other functions if/when we have Java callers. YAGNI and all of that. :)

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1638)* status: 2 of 9 files reviewed, 1 unresolved discussion (waiting on @kungfucraig)

___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/v2alpha/FulfillRequisitionRequestBuilder.kt` line 55 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1638#-O-2MIUQ7COpFJVZKARv:-O-39PUIBShxrJocd6Uc:bhq84g1) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8189aee883da31f110e7873314c159bf85de216c/src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/v2alpha/FulfillRequisitionRequestBuilder.kt#L55)):*
> See: [https://sourcegraph.com/github.com/world-federation-of-advertisers/cross-media-measurement/-/blob/src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/v2alpha/InMemoryVidIndexMap.kt?L139](https://sourcegraph.com/github.com/world-federation-of-advertisers/cross-media-measurement/-/blob/src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/v2alpha/InMemoryVidIndexMap.kt?L139)

This doesn't use coroutine Flows.

> My main point is that everyone who is using this uses Kotlin, and that we can add other functions if/when we have Java callers. YAGNI and all of that. :)

Yes, my suggestion matches this: use a Kotlin Sequence initially and add Java Streams if/when we have Java callers. Starting with Flow precludes the ability to do the second without having any benefit. Sequence is trivially convertible to Flow, but not the other way around.

kungfucraig (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1638)* status: 2 of 9 files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/v2alpha/FulfillRequisitionRequestBuilder.kt` line 55 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1638#-O-2MIUQ7COpFJVZKARv:-O-DfZxhAVIOWZkkZsmG:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8189aee883da31f110e7873314c159bf85de216c/src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/v2alpha/FulfillRequisitionRequestBuilder.kt#L55)):*

> See: [https://sourcegraph.com/github.com/world-federation-of-advertisers/cross-media-measurement/-/blob/src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/v2alpha/InMemoryVidIndexMap.kt?L139](https://sourcegraph.com/github.com/world-federation-of-advertisers/cross-media-measurement/-/blob/src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/v2alpha/InMemoryVidIndexMap.kt?L139)

This doesn't use coroutine Flows.

> My main point is that everyone who is using this uses Kotlin, and that we can add other functions if/when we have Java callers. YAGNI and all of that. :)

Yes, my suggestion matches this: use a Kotlin Sequence initially and add Java Streams if/when we have Java callers. Starting with Flow precludes the ability to do the second without having any benefit. Sequence is trivially convertible to Flow, but not the other way around.
</blockquote></details>

Done.

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1638)* status: 2 of 9 files reviewed, 2 unresolved discussions (waiting on @kungfucraig)

___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/v2alpha/FulfillRequisitionRequestBuilder.kt` line 123 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1638#-O-HfiUY55P8szynT8Wn:-O-HfiUY55P8szynT8Wo:befnb5w) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1e71f77a248645add1276f3468ba26692a0f5f0b/src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/v2alpha/FulfillRequisitionRequestBuilder.kt#L123)):*
> ```Kotlin
>   /** Builds the Sequence of requests. */
>   fun build(): Sequence<FulfillRequisitionRequest> =
>     buildList() {
> ```

Using buildList first removes the benefit of using a Sequence. Use the `sequence` DSL builder instead.

kungfucraig (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1638)* status: 2 of 9 files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/v2alpha/FulfillRequisitionRequestBuilder.kt` line 123 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1638#-O-HfiUY55P8szynT8Wn:-O-MqA5q2gbb_dXiXpd0:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1e71f77a248645add1276f3468ba26692a0f5f0b/src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/v2alpha/FulfillRequisitionRequestBuilder.kt#L123)):*

Using buildList first removes the benefit of using a Sequence. Use the `sequence` DSL builder instead.
</blockquote></details>

Done.

SanjayVas (approved):
> Reviewed 3 of 9 files at r1, 1 of 1 files at r2, 2 of 4 files at r3, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1638)* status: all files reviewed, 1 unresolved discussion (waiting on @kungfucraig)

___
*[`src/test/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulatorTest.kt` line 3111 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1638#-O-NxGGA-reD2Smvy7ct:-O-NxGGA-reD2Smvy7cu:b-fjlhnt) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/291f67c8857cc71094c8a8a047a4799d9c7a4707/src/test/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulatorTest.kt#L3111)):*
> ```Kotlin
>     }
> 
>     // TODO(@kungfucraig): Replace this with the object in HmmsRequisitions.kt
> ```

nit: use KDoc syntax for properties, putting the TODO in the KDoc block

stevenwarejones (approved):
> Reviewed 4 of 9 files at r1, 1 of 1 files at r2, 2 of 4 files at r3, 1 of 1 files at r4, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1638)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @kungfucraig)

PR #1641, opened 2024-06-05
Use index in existing Measurements check when creating Measurements

tristanvuong2021:
The request ID index isn't actually being used right now. Instead, a table scan is being done. This refactors the query to actually use the index.

### DIFF
--- a/src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/readers/MeasurementReader.kt
+++ b/src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/readers/MeasurementReader.kt
@@ -46,7 +46,7 @@ import org.wfanet.measurement.kingdom.deploy.common.DuchyIds
 import org.wfanet.measurement.kingdom.deploy.gcloud.spanner.common.MeasurementNotFoundException
 import org.wfanet.measurement.kingdom.deploy.gcloud.spanner.readers.MeasurementReader.Companion.getEtag
 
-class MeasurementReader(private val view: Measurement.View) :
+class MeasurementReader(private val view: Measurement.View, measurementsIndex: Index = Index.NONE) :
   SpannerReader<MeasurementReader.Result>() {
 
   data class Result(
@@ -56,8 +56,24 @@ class MeasurementReader(private val view: Measurement.View) :
     val measurement: Measurement,
   )
 
+  enum class Index(internal val sql: String) {
+    NONE(""),
+    CREATE_REQUEST_ID("@{FORCE_INDEX=MeasurementsByCreateRequestId}"),
+  }
+
   override val baseSql: String
     get() = BASE_SQL
+
+  private val BASE_SQL =
+    """
+      @{spanner_emulator.disable_query_null_filtered_index_check=true}
+      WITH FilteredMeasurements AS (
+        SELECT *
+        FROM
+          Measurements${measurementsIndex.sql}
+          JOIN MeasurementConsumers USING (MeasurementConsumerId)
+      """
+      .trimIndent()
 
   private var filled = false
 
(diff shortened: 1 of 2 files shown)


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1641)

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1641)* status: all files reviewed, 1 unresolved discussion (waiting on @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/writers/CreateMeasurements.kt` line 429 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1641#-Nzdq3feAl_wIxZ_n_tm:-Nzdq3ff2SLsyPCxUdtc:b8r0i4h) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1f65d08f6c8540357de85a5cb61b5acd34e50027/src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/writers/CreateMeasurements.kt#L429)):*
> ```Kotlin
>     val fromClause =
>       """
>       FROM
> ```

MeasurementReader shouldn't have a caller-specified FROM clause. More specifically, we shouldn't need to parameterize the JOINs - it should pick those values up from the WHERE clause. If you *really* think this is needed, then factor it out to a separate path in MeasurementReader rather than having the FROM clause passed by the caller.

If you just want the caller to be able to specify the index, take a look at how RecurringExchangeReader does it as an example.

tristanvuong2021 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1641)* status: all files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/writers/CreateMeasurements.kt` line 429 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1641#-Nzdq3feAl_wIxZ_n_tm:-Nze5DBN1DkxsZOJEF-m:b-adazvb) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/2402db9b76f6481d61546e5fe2aed78fe2412d0f/src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/writers/CreateMeasurements.kt#L429)):*

MeasurementReader shouldn't have a caller-specified FROM clause. More specifically, we shouldn't need to parameterize the JOINs - it should pick those values up from the WHERE clause. If you *really* think this is needed, then factor it out to a separate path in MeasurementReader rather than having the FROM clause passed by the caller.

If you just want the caller to be able to specify the index, take a look at how RecurringExchangeReader does it as an example.
</blockquote></details>

The performance is much worse without this change.

tristanvuong2021 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1641)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/writers/CreateMeasurements.kt` line 429 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1641#-Nzdq3feAl_wIxZ_n_tm:-NzeLUqX2wgMKAN6B8F1:b-r36x0w) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a1f06d428df3b1a45b7005b0486c1526dea0531e/src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/writers/CreateMeasurements.kt#L429)):*

The performance is much worse without this change.
</blockquote></details>

I modeled it after RecurringExchangeReader. I just needed to make sure the null check is in the where clause.

SanjayVas (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1641)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @tristanvuong2021)

stevenwarejones (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1641)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @tristanvuong2021)

PR #1662, opened 2024-06-12
feat: Define Access API

SanjayVas:


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1662)

kungfucraig (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1662)* status: 0 of 14 files reviewed, 5 unresolved discussions (waiting on @SanjayVas)

___
*[`src/main/proto/wfa/measurement/access/v1alpha/permissions_service.proto` line 29 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1662#-OB7iian4YI_4RyrZ3iQ:-OB7iiao-cQEdzavOgid:b4nzxsn) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a90eeaecfc72e2f4c4525244530954a855275b0b/src/main/proto/wfa/measurement/access/v1alpha/permissions_service.proto#L29)):*
> ```proto
> service Permissions {
>   // Checks what permissions a `Principal` has on a given resource.
>   rpc CheckPermissions(CheckPermissionsRequest)
> ```

How do you intend this to be called? I can think of at least a couple of ways...

1. The caller collects the entire set of permissions that may be needed for any incoming request of the given type and calls this method
2. The caller collects the exact set of permissions that are needed for the request and calls this method

In practice, the difference between these two may be small, but I can imagine as we have finer-grained permissions they could diverge quite a bit.

Anyway, the latter makes more sense to me, and if this is the case then having an additional boolean on the response that indicates all permissions in the request are granted could be useful.
___
*[`src/main/proto/wfa/measurement/access/v1alpha/permissions_service.proto` line 36 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1662#-OB7jxoV-Im0GWl0kwhK:-OB7jxoV-Im0GWl0kwhL:borsbz3) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a90eeaecfc72e2f4c4525244530954a855275b0b/src/main/proto/wfa/measurement/access/v1alpha/permissions_service.proto#L36)):*
> ```proto
> message CheckPermissionsRequest {
>   // Name of the resource on which to check permissions. If not specified, this
>   // means the root of the protected API.
> ```

What's the value of the default behavior?
___
*[`src/main/proto/wfa/measurement/access/v1alpha/policy.proto` line 51 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1662#-OB7mCaD1dDY5awkxzMo:-OB7mCaD1dDY5awkxzMp:bkhpou) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a90eeaecfc72e2f4c4525244530954a855275b0b/src/main/proto/wfa/measurement/access/v1alpha/policy.proto#L51)):*
> ```proto
>   message Binding {
>     // Resource name of the role.
>     string role = 1 [
> ```

This is unique across all bindings for this resource?
___
*[`src/main/proto/wfa/measurement/access/v1alpha/policy.proto` line 66 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1662#-OB7nCfQ-zk8H1N-9l4D:-OB7nCfQ-zk8H1N-9l4E:bl3atwa) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a90eeaecfc72e2f4c4525244530954a855275b0b/src/main/proto/wfa/measurement/access/v1alpha/policy.proto#L66)):*
> ```proto
>   }
>   // Policy bindings.
>   repeated Binding bindings = 3 [
> ```

I have the sense that this could get largish.

Did you consider having a separate resource to manage role membership instead of managing role membership and the role/resource relationship in the same resource? 
___
*[`src/main/proto/wfa/measurement/access/v1alpha/principal.proto` line 53 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1662#-OB7nc9sEUiBwVtWwUK3:-OB7nc9sEUiBwVtWwUK4:bdd8yc8) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a90eeaecfc72e2f4c4525244530954a855275b0b/src/main/proto/wfa/measurement/access/v1alpha/principal.proto#L53)):*
> ```proto
>   }
>   // A TLS client.
>   message TlsClient {
> ```

Do we have bring TLS clients into this, or could we continue to manage the TLS clients the way we do today until we can delete them?

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1662)* status: 0 of 14 files reviewed, 5 unresolved discussions (waiting on @kungfucraig)

___
*[`src/main/proto/wfa/measurement/access/v1alpha/permissions_service.proto` line 29 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1662#-OB7iian4YI_4RyrZ3iQ:-OBC-SBI4Gd6zZB3UI34:bjyvz3y) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a90eeaecfc72e2f4c4525244530954a855275b0b/src/main/proto/wfa/measurement/access/v1alpha/permissions_service.proto#L29)):*

How do you intend this to be called? I can think of at least a couple of ways...

1. The caller collects the entire set of permissions that may be needed for any incoming request of the given type and calls this method
2. The caller collects the exact set of permissions that are needed for the request and calls this method

In practice, the difference between these two may be small, but I can imagine as we have finer-grained permissions they could diverge quite a bit.

Anyway, the latter makes more sense to me, and if this is the case then having an additional boolean on the response that indicates all permissions in the request are granted could be useful.
</blockquote></details>

The intent is that the caller passes the exact list of permissions needed for the operation. There should only be one call per resource per frontend RPC (meaning potentially multiple calls if checking ancestor resources).

The reason this returns a subset is so that the caller can choose to show an error with which permissions the caller may have or be missing. A boolean wouldn't be useful since you can just check whether the response permissions has the same size as the requested permissions.
___
*[`src/main/proto/wfa/measurement/access/v1alpha/permissions_service.proto` line 36 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1662#-OB7jxoV-Im0GWl0kwhK:-OBC03GaBgpEW5aYUYGn:b-199rcf) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a90eeaecfc72e2f4c4525244530954a855275b0b/src/main/proto/wfa/measurement/access/v1alpha/permissions_service.proto#L36)):*

What's the value of the default behavior?
</blockquote></details>

Not sure what you mean. Not specified in proto3 is equivalent to being set to the default value for that type. The default value for string is empty string.
___
*[`src/main/proto/wfa/measurement/access/v1alpha/policy.proto` line 51 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1662#-OB7mCaD1dDY5awkxzMo:-OBC0JBn5AtkoHDdBqQA:b-g72lb2) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a90eeaecfc72e2f4c4525244530954a855275b0b/src/main/proto/wfa/measurement/access/v1alpha/policy.proto#L51)):*

This is unique across all bindings for this resource?
</blockquote></details>

It need not be in requests, but naturally will be in responses.
___
*[`src/main/proto/wfa/measurement/access/v1alpha/policy.proto` line 66 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1662#-OB7nCfQ-zk8H1N-9l4D:-OBC0Pnt37g684paywOX:b-4x7ols) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a90eeaecfc72e2f4c4525244530954a855275b0b/src/main/proto/wfa/measurement/access/v1alpha/policy.proto#L66)):*

I have the sense that this could get largish.

Did you consider having a separate resource to manage role membership instead of managing role membership and the role/resource relationship in the same resource?
</blockquote></details>

I expect the members list to get large, but not the number of bindings per resource unless there are a very large number of roles.

Regardless, another resource wouldn't work. A policy binding is effectively a (protected_resource, role, principal) tuple - i.e. whether a principal has a specific role on a specific resource. e.g. Alice having role reportReader on measurementConsumers/foo

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1662)* status: 0 of 14 files reviewed, 5 unresolved discussions (waiting on @kungfucraig)

___
*[`src/main/proto/wfa/measurement/access/v1alpha/principal.proto` line 53 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1662#-OB7nc9sEUiBwVtWwUK3:-OBC0xym5CmRcWjFQLms:b-zxip0) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a90eeaecfc72e2f4c4525244530954a855275b0b/src/main/proto/wfa/measurement/access/v1alpha/principal.proto#L53)):*

Do we have bring TLS clients into this, or could we continue to manage the TLS clients the way we do today until we can delete them?
</blockquote></details>

I want all of the calling code to shift to using this API, regardless of whether it's TLS or not. Note that some paths will always support mTLS, such as from EDP callers. If we change the Kingdom public API impl to use the access API, then this path actually wouldn't be deleted.

The part that can be deleted eventually is the code that reads from the current TLS AKID mapping config rather than storing this in the DB like other principal types.

kungfucraig (commented):
> Reviewed 1 of 13 files at r1, 5 of 8 files at r2, 7 of 8 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1662)* status: 13 of 14 files reviewed, 4 unresolved discussions (waiting on @SanjayVas)

___
*[`src/main/proto/wfa/measurement/access/v1alpha/permissions_service.proto` line 29 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1662#-OB7iian4YI_4RyrZ3iQ:-OBCrjAdEi3ON4lH1jM6:bfk0iql) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a90eeaecfc72e2f4c4525244530954a855275b0b/src/main/proto/wfa/measurement/access/v1alpha/permissions_service.proto#L29)):*

The intent is that the caller passes the exact list of permissions needed for the operation. There should only be one call per resource per frontend RPC (meaning potentially multiple calls if checking ancestor resources).

The reason this returns a subset is so that the caller can choose to show an error with which permissions the caller may have or be missing. A boolean wouldn't be useful since you can just check whether the response permissions has the same size as the requested permissions.
</blockquote></details>

Can add some comments that summarize this?
___
*[`src/main/proto/wfa/measurement/access/v1alpha/permissions_service.proto` line 36 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1662#-OB7jxoV-Im0GWl0kwhK:-OBCsC_0A1C47VvlvZww:b-xd1e58) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a90eeaecfc72e2f4c4525244530954a855275b0b/src/main/proto/wfa/measurement/access/v1alpha/permissions_service.proto#L36)):*

Not sure what you mean. Not specified in proto3 is equivalent to being set to the default value for that type. The default value for string is empty string.
</blockquote></details>

Ah, sorry - "value" was ambiguous. What purpose does the default behavior serve? We actually discussed this yesterday though.
___
*[`src/main/proto/wfa/measurement/access/v1alpha/policy.proto` line 51 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1662#-OB7mCaD1dDY5awkxzMo:-OBCsjGl2P4g8m4Itl9M:b69vxl8) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a90eeaecfc72e2f4c4525244530954a855275b0b/src/main/proto/wfa/measurement/access/v1alpha/policy.proto#L51)):*

It need not be in requests, but naturally will be in responses.
</blockquote></details>

Can you add this to the documentation?
___
*[`src/main/proto/wfa/measurement/access/v1alpha/roles_service.proto` line 37 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1662#-OBCuh3b8D9R1GJxTTqI:-OBCuh3b8D9R1GJxTTqJ:bstvk8m) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a90eeaecfc72e2f4c4525244530954a855275b0b/src/main/proto/wfa/measurement/access/v1alpha/roles_service.proto#L37)):*
> ```proto
> 
>   // Lists `Role` resources.
>   rpc ListRoles(ListRolesRequest) returns (ListRolesResponse) {
> ```

What use case is "List" supporting, and why is the request referencing a parent. Role doesn't appear to have a parent.
___
*[`src/main/proto/wfa/measurement/access/v1alpha/policies_service.proto` line 29 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1662#-OBCtd_k65EWIufpWOLu:-OBCtd_k65EWIufpWOLv:b-4w5e9c) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a90eeaecfc72e2f4c4525244530954a855275b0b/src/main/proto/wfa/measurement/access/v1alpha/policies_service.proto#L29)):*
> ```proto
> 
> // Service for interacting with `Policy` resources.
> service Policies {
> ```

Need a delete?

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1662)* status: 7 of 15 files reviewed, 4 unresolved discussions (waiting on @kungfucraig and @SanjayVas)

___
*[`src/main/proto/wfa/measurement/access/v1alpha/permissions_service.proto` line 29 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1662#-OB7iian4YI_4RyrZ3iQ:-OBRcYSgEmvx9Qos9aA_:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a90eeaecfc72e2f4c4525244530954a855275b0b/src/main/proto/wfa/measurement/access/v1alpha/permissions_service.proto#L29)):*

Can add some comments that summarize this?
</blockquote></details>

Done.
___
*[`src/main/proto/wfa/measurement/access/v1alpha/permissions_service.proto` line 36 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1662#-OB7jxoV-Im0GWl0kwhK:-OBRc7XqElVLKZxK4rHZ:b-7ulprb) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a90eeaecfc72e2f4c4525244530954a855275b0b/src/main/proto/wfa/measurement/access/v1alpha/permissions_service.proto#L36)):*

Ah, sorry - "value" was ambiguous. What purpose does the default behavior serve? We actually discussed this yesterday though.
</blockquote></details>

For posterity: operations like Create inherently operate on the resource that owns the collection, i.e. the parent resource. Some resource collections are top-level and don't have a parent.
___
*[`src/main/proto/wfa/measurement/access/v1alpha/policies_service.proto` line 29 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1662#-OBCtd_k65EWIufpWOLu:-OBRYFd-AW9CllK9Vnzh:b-19rncx) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a90eeaecfc72e2f4c4525244530954a855275b0b/src/main/proto/wfa/measurement/access/v1alpha/policies_service.proto#L29)):*

Need a delete?
</blockquote></details>

We shouldn't need one. A policy with no bindings is the same as one that does not exist.

If not for the need to specify a resource ID, I'd actually not even have a Create - meaning a Policy would implicitly exist for every protected resource. In fact, I may still do this - drop the Create method and generate a resource ID as-needed on Lookup.
___
*[`src/main/proto/wfa/measurement/access/v1alpha/policy.proto` line 51 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1662#-OB7mCaD1dDY5awkxzMo:-OBRaoDR8BiJRaMhwzH1:b-fop9f3) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a90eeaecfc72e2f4c4525244530954a855275b0b/src/main/proto/wfa/measurement/access/v1alpha/policy.proto#L51)):*

Can you add this to the documentation?
</blockquote></details>

Done by just describing that this should form a map
___
*[`src/main/proto/wfa/measurement/access/v1alpha/roles_service.proto` line 37 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1662#-OBCuh3b8D9R1GJxTTqI:-OBRb6us2Gr6BU4Ibv1e:bmqr2ww) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a90eeaecfc72e2f4c4525244530954a855275b0b/src/main/proto/wfa/measurement/access/v1alpha/roles_service.proto#L37)):*
List is useful for when you're trying to figure out which existing role you may want to use, whether the role you want exists, what it's named, etc. Note that the AIPs actually state that every non-singleton resource type must provide a List method - we only avoid this for cases where we *know* it wouldn't be used.

> why is the request referencing a parent. Role doesn't appear to have a parent.

Copy-paste error. Fixed.

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1662)* status: 7 of 15 files reviewed, 4 unresolved discussions (waiting on @kungfucraig)

___
*[`src/main/proto/wfa/measurement/access/v1alpha/policies_service.proto` line 29 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1662#-OBCtd_k65EWIufpWOLu:-OBRdBhG7FTWK70HJctI:b-6rswfj) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a90eeaecfc72e2f4c4525244530954a855275b0b/src/main/proto/wfa/measurement/access/v1alpha/policies_service.proto#L29)):*
> In fact, I may still do this - drop the Create method and generate a resource ID as-needed on Lookup.

I changed my mind on this, as it would result in anyone who can call Lookup having the ability to generate garbage DB rows for resources that don't exist. If we wanted to do this, it would be best to have a deterministic algorithm for generating a unique valid resource ID from a protected resource name. Such an algorithm would need to be collision-free.

kungfucraig (approved):
> Reviewed 1 of 8 files at r3, 7 of 7 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1662)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @SanjayVas)

stevenwarejones (approved):
> Reviewed 1 of 13 files at r1, 5 of 8 files at r2, 1 of 8 files at r3, 4 of 7 files at r4, 3 of 3 files at r5, 1 of 1 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1662)* status: all files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

___
*[`src/main/proto/wfa/measurement/access/v1alpha/principal.proto` line 39 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1662#-OCd_CE0-LImUNFZ488K:-OCd_CE0-LImUNFZ488L:bueyf4t) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ed51a262065819c55f69d3582d08e503d11e90ad/src/main/proto/wfa/measurement/access/v1alpha/principal.proto#L39)):*
> ```proto
> 
>   // A single user.
>   message User {
> ```

see the other comment about renaming to OAuthUser

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1662)* status: 13 of 15 files reviewed, 1 unresolved discussion (waiting on @kungfucraig and @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/access/v1alpha/principal.proto` line 39 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1662#-OCd_CE0-LImUNFZ488K:-OCdlQWxDby1qfhIZMSY:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ed51a262065819c55f69d3582d08e503d11e90ad/src/main/proto/wfa/measurement/access/v1alpha/principal.proto#L39)):*

see the other comment about renaming to OAuthUser
</blockquote></details>

Done.

SanjayVas (commented):
> Reviewed 2 of 2 files at r7.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1662)* status: all files reviewed, 1 unresolved discussion (waiting on @kungfucraig and @stevenwarejones)

SanjayVas (commented):
> Dismissed @stevenwarejones from a discussion.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1662)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @kungfucraig and @stevenwarejones)

PR #1665, opened 2024-06-14
Retry updating ComputationParticipant to avoid Measurement failure.

SanjayVas:

PR #1671, opened 2024-06-21
Allow Herald to skip stage to catch up with Measurement state.

renjiezh:


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1671)

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1671)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @renjiezh)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/HonestMajorityShareShuffleStarter.kt` line 144 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1671#-O0FeLcD1q2HjpoQx0O6:-O0FeLcE1sYWobHjsKUL:b-wkbaax) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/9d4b10016e00b3c86d345bb1fa51d221c8155064/src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/HonestMajorityShareShuffleStarter.kt#L144)):*
> ```Kotlin
>       // key has been set, skip the stage to catch up to the Measurement state.
>       Stage.INITIALIZED -> {
>         if (!isInitialized(token)) {
> ```

Both pieces of information we're checking here (stage and encryption_key_pair) come from the computation details in the token. Shouldn't these always be consistent?

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1671)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/HonestMajorityShareShuffleStarter.kt` line 144 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1671#-O0FeLcD1q2HjpoQx0O6:-O0GK1704ZWsPrqDskcD:b-voopkt) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/9d4b10016e00b3c86d345bb1fa51d221c8155064/src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/HonestMajorityShareShuffleStarter.kt#L144)):*

Both pieces of information we're checking here (stage and encryption_key_pair) come from the computation details in the token. Shouldn't these always be consistent?
</blockquote></details>

Nope. They are not guaranteed consistent.
https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/00a8897353e28513952b16fec873ea1b18297c7b/src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/liquidlegionsv2/ReachFrequencyLiquidLegionsV2Mill.kt#L251-L263
The process is:
1. Update the token with algamal keys if not created, otherwise skip.
2. Update the Participant at the kingdom (with the algamal keys)
3. Update the own stage.
This was to avoid duplication of creating algamal keys.

The check here is to avoid the case that Kingdom's participant state is abnormal (not happening now, but it is a check).

SanjayVas (commented):
> Reviewed 1 of 3 files at r1.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1671)* status: 1 of 3 files reviewed, all discussions resolved

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/HonestMajorityShareShuffleStarter.kt` line 144 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1671#-O0FeLcD1q2HjpoQx0O6:-O0GQYoi5qZKXJtNOMWe:b-aj2p54) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/9d4b10016e00b3c86d345bb1fa51d221c8155064/src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/HonestMajorityShareShuffleStarter.kt#L144)):*

Nope. They are not guaranteed consistent.
https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/00a8897353e28513952b16fec873ea1b18297c7b/src/main/kotlin/org/wfanet/measurement/duchy/daemon/mill/liquidlegionsv2/ReachFrequencyLiquidLegionsV2Mill.kt#L251-L263
The process is:
1. Update the token with algamal keys if not created, otherwise skip.
2. Update the Participant at the kingdom (with the algamal keys)
3. Update the own stage.
This was to avoid duplication of creating algamal keys.

The check here is to avoid the case that Kingdom's participant state is abnormal (not happening now, but it is a check).
</blockquote></details>

Is this mentioning the existing process, or the new one? It sounds like we need to handle two cases where the encryption key has already been written to the local Duchy DB:

1. Participant not yet updated at Kingdom (crash after step 1 above)
2. Participant updated at Kingdom, but local state transition not yet recorded (crash after step 2 above)

Doesn't this PR only handle (2)?

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1671)* status: 1 of 3 files reviewed, all discussions resolved (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/HonestMajorityShareShuffleStarter.kt` line 144 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1671#-O0FeLcD1q2HjpoQx0O6:-O0GTL9K6LOkoK0tznHI:b7kv1ed) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/9d4b10016e00b3c86d345bb1fa51d221c8155064/src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/HonestMajorityShareShuffleStarter.kt#L144)):*

Is this mentioning the existing process, or the new one? It sounds like we need to handle two cases where the encryption key has already been written to the local Duchy DB:

1. Participant not yet updated at Kingdom (crash after step 1 above)
2. Participant updated at Kingdom, but local state transition not yet recorded (crash after step 2 above)

Doesn't this PR only handle (2)?
</blockquote></details>

The process is the existing one. This PR is to handle (2)

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1671)* status: 1 of 3 files reviewed, all discussions resolved (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/HonestMajorityShareShuffleStarter.kt` line 144 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1671#-O0FeLcD1q2HjpoQx0O6:-O0GUoOGBYyXHZ8bodCb:b-ewdnh5) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/9d4b10016e00b3c86d345bb1fa51d221c8155064/src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/HonestMajorityShareShuffleStarter.kt#L144)):*

The process is the existing one. This PR is to handle (2)
</blockquote></details>

(1) should have been addressed by your PR 
https://github.com/world-federation-of-advertisers/cross-media-measurement/pull/1665/files#diff-1eebd5edc220318daad624461a19b99a772fa8238c8d9c9128ca2416de6655c4R315

SanjayVas (approved):
> Reviewed 2 of 3 files at r1.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1671)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @renjiezh)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/HonestMajorityShareShuffleStarter.kt` line 144 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1671#-O0FeLcD1q2HjpoQx0O6:-O0GVCLO3IK_XR82kqWy:byl8nv1) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/9d4b10016e00b3c86d345bb1fa51d221c8155064/src/main/kotlin/org/wfanet/measurement/duchy/daemon/herald/HonestMajorityShareShuffleStarter.kt#L144)):*

(1) should have been addressed by your PR 
https://github.com/world-federation-of-advertisers/cross-media-measurement/pull/1665/files#diff-1eebd5edc220318daad624461a19b99a772fa8238c8d9c9128ca2416de6655c4R315
</blockquote></details>

Ah, I missed the fact that only (2) could hit this logic as this is called from the Herald based on computation state.

stevenwarejones (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1671)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @renjiezh)

PR #1678, opened 2024-07-03
Added method to batch search privacy entries to the interface

RibeiroAna:
At Meta we were facing some performance issues when trying to find privacy bucket groups one by one. We created internally some methods to batch the select queries and that helped. We think this should be added to the interface so other EDP implementing the PBM can also create methods to batch their entries.

### DIFF
--- a/src/main/kotlin/org/wfanet/measurement/eventdataprovider/privacybudgetmanagement/PrivacyBudgetLedger.kt
+++ b/src/main/kotlin/org/wfanet/measurement/eventdataprovider/privacybudgetmanagement/PrivacyBudgetLedger.kt
@@ -114,12 +114,13 @@ class PrivacyBudgetLedger(
     context: PrivacyBudgetLedgerTransactionContext,
     privacyBucketGroups: Set<PrivacyBucketGroup>,
     acdpCharges: Set<AcdpCharge>,
-  ): List<PrivacyBucketGroup> =
-    privacyBucketGroups.filter { privacyBucketGroup ->
-      val acdpBalanceEntries = setOf(context.findAcdpBalanceEntry(privacyBucketGroup))
-      val balanceAcdpCharges = acdpBalanceEntries.map { it.acdpCharge }
-      exceedsUnderAcdpComposition(balanceAcdpCharges + acdpCharges)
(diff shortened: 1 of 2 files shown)


stevenwarejones (changes requested):
@SanjayVas - I noticed the tests didn't run on this PR. I'm assuming its since she opened it from her fork? If that's the case, then can you add her to the org so she can close and re-open.

> Reviewed 1 of 2 files at r1.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1678#-:-O0tl49C7XTiGoJdg0Yn:b-2n0nu7)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on @RibeiroAna)

___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/privacybudgetmanagement/PrivacyBudgetLedgerBackingStore.kt` line 103 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1678#-O0tkr0sEvogRw-TjWZ3:-O0tkr0sEvogRw-TjWZ4:b-t5dw19) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/19354e0225f467b33adeeef7cff1e6a8b09bc0a6/src/main/kotlin/org/wfanet/measurement/eventdataprovider/privacybudgetmanagement/PrivacyBudgetLedgerBackingStore.kt#L103)):*
> ```Kotlin
>    * The default implementation calls the findAcdpBalanceEntry method individually.
>    */
>   suspend fun findAcdpBalanceEntries(
> ```

this should be open, right, so it can be overridden?

This function also needs a test

SanjayVas (commented):
Yes. You'll need Tej to add to the WFA GitHub org, as we do access management at the org level.

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1678#-:-O0tv9vyDNy7Z7FVmE29:b-dpecru)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on @iverson52000, @RibeiroAna, and @uakyol)

PR #1681, opened 2024-07-08
Implement CertificateManager for Kingdom-less panel match protocol.

robinsons:


stevenwarejones (changes requested):
> Reviewed 6 of 8 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1681)* status: 6 of 8 files reviewed, 5 unresolved discussions (waiting on @robinsons)

___
*[`src/main/kotlin/org/wfanet/panelmatch/common/certificates/KingomlessCertificateManager.kt` line 64 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1681#-O1J2oyK4f1aSHB628UM:-O1J2oyK4f1aSHB628UN:b2fwtx) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f0d8a9efe8b3b32c2ba4851203e529520334e42f/src/main/kotlin/org/wfanet/panelmatch/common/certificates/KingomlessCertificateManager.kt#L64)):*
> ```Kotlin
>         }
>       val x509 = readCertificate(certificate.x509Der)
>       verifyCertificate(x509, certificate.ownerId)
> ```

the ownerId shouldn't be derived from the certificate. The ownerId should already be known from the ExchangeDateKey ie a map should already exist, right? And, instead of an owner Id, I think using the AKID is preferable.
___
*[`src/main/kotlin/org/wfanet/panelmatch/common/certificates/KingomlessCertificateManager.kt` line 124 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1681#-O1J3R1_87j_9QoYVe98:-O1J3R1a2Lf9ZJCELfWv:bgszgmd) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f0d8a9efe8b3b32c2ba4851203e529520334e42f/src/main/kotlin/org/wfanet/panelmatch/common/certificates/KingomlessCertificateManager.kt#L124)):*
> ```Kotlin
>     ownerId: String,
>   ): X509Certificate {
>     val rootCert = getRootCertificate(ownerId)
> ```

see my other comment... you should just look up the root by the AKID on the leaf certificate imo.
___
*[`src/main/kotlin/org/wfanet/panelmatch/common/certificates/KingomlessCertificateManager.kt` line 162 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1681#-O1J1rEA91cUKhy9JIuG:-O1J1rEA91cUKhy9JIuH:bntvlmb) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f0d8a9efe8b3b32c2ba4851203e529520334e42f/src/main/kotlin/org/wfanet/panelmatch/common/certificates/KingomlessCertificateManager.kt#L162)):*
> ```Kotlin
>     private const val BLOB_KEY_PREFIX = "certificates"
>     private const val DATA_PROVIDER_CERT_PREFIX = "edp-certificate"
>     private const val MODEL_PROVIDER_CERT_PREFIX = "mp-certificate"
> ```

in the init cannot you verify that the model provider and data provider prefixes differ?
___
*[`src/main/proto/wfa/panelmatch/client/internal/certificate.proto` line 27 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1681#-O1J3-zA6RlBT3jzn1Qv:-O1J3-zA6RlBT3jzn1Qw:bikmf1k) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f0d8a9efe8b3b32c2ba4851203e529520334e42f/src/main/proto/wfa/panelmatch/client/internal/certificate.proto#L27)):*
> ```proto
>   // ID of the entity that created this certificate. This is the ID of either a
>   // data provider or model provider participating in an exchange.
>   string owner_id = 1;
> ```

see my other comment, I don't think we need this field.
___
*[`src/main/proto/wfa/panelmatch/common/certificates/signing_keys.proto` line 25 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1681#-O1J3jZuE3Dzs27DFiZB:-O1J3jZuE3Dzs27DFiZC:b-8oms8) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f0d8a9efe8b3b32c2ba4851203e529520334e42f/src/main/proto/wfa/panelmatch/common/certificates/signing_keys.proto#L25)):*
> ```proto
>   // TODO(@robinsons): Rename field to cert_name since not all panel match
>   // protocols use resources.
>   string cert_resource_name = 1;
> ```

why not go ahead and rename it?

stevenwarejones (changes requested):
> Reviewed 2 of 6 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1681)* status: 6 of 10 files reviewed, 3 unresolved discussions (waiting on @robinsons)

___
*[`src/main/kotlin/org/wfanet/panelmatch/common/certificates/KingomlessCertificateManager.kt` line 64 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1681#-O1J2oyK4f1aSHB628UM:-O1Nseoy8-zn94p4wDsO:b-wj5mtb) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f0d8a9efe8b3b32c2ba4851203e529520334e42f/src/main/kotlin/org/wfanet/panelmatch/common/certificates/KingomlessCertificateManager.kt#L64)):*

One of my goals was to keep this close to the existing V2AlphaCertificateManager for the sake of simplicity. With that in mind, I've updated it to remove ownerId from the certificate and instead derive it from certName.

---

As for switching to AKID, the main reason I hesitate is that the daemon is currently set up such that it doesn't use AKID anywhere. For example, the rootCerts map is just a StorageClientSecretMap whose blob keys are the names of the entity to whom the cert belongs, i.e. the ownerId. So the system already seems designed to work using owner IDs.

IIUC, switching to AKID would entail passing an additional map from AKID to owner ID, so that getCertificate() could look up the owner ID for the cert, and then the remaining flow would remain as is. It seems the main win from this would be that we don't have to rely on a particular certificate naming scheme in order to derive ownerId - is that right?

I guess my main question is how strongly you feel about this. If it's something you definitely want, would you be OK if I filed an issue for this and we tackled it for both certificate manager impls at once sometime after the Kingdomless work is done?
</blockquote></details>

my main concern was removing the field from the certificate proto. The rest is lower priority.
___
*[`src/main/kotlin/org/wfanet/panelmatch/common/certificates/KingomlessCertificateManager.kt` line 58 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1681#-O1NtmvX2vFgsn5FPtKg:-O1NtmvX2vFgsn5FPtKh:bt2022x) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/42bd0101a4e2f6aa42c85c7e339f9823f0f6a775/src/main/kotlin/org/wfanet/panelmatch/common/certificates/KingomlessCertificateManager.kt#L58)):*
> ```Kotlin
>   ): X509Certificate {
>     val certKey =
>       requireNotNull(CertificateKey.fromCertName(certName)) { "Invalid certName: $certName" }
> ```

can you also add something verifying that the ownerId maps to this exchange. we should have that mapping, right?
___
*[`src/main/proto/wfa/panelmatch/client/internal/certificate.proto` line 27 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1681#-O1J3-zA6RlBT3jzn1Qv:-O1NsVYZ5VYgzb4IfQ82:b8qkdvq) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f0d8a9efe8b3b32c2ba4851203e529520334e42f/src/main/proto/wfa/panelmatch/client/internal/certificate.proto#L27)):*

Done.
</blockquote></details>

my main concern was removing this field.

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1681)* status: 6 of 10 files reviewed, 3 unresolved discussions (waiting on @robinsons)

___
*[`src/main/kotlin/org/wfanet/panelmatch/common/certificates/KingomlessCertificateManager.kt` line 165 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1681#-O1NuHZkCkbKKYzvaofh:-O1NuHZkCkbKKYzvaofi:b-3jhsx) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/42bd0101a4e2f6aa42c85c7e339f9823f0f6a775/src/main/kotlin/org/wfanet/panelmatch/common/certificates/KingomlessCertificateManager.kt#L165)):*
> ```Kotlin
>         val match = CERT_NAME_PATTERN.find(certName) ?: return null
>         val ownerId = match.groupValues[OWNER_ID_INDEX]
>         val createTime = Instant.ofEpochSecond(match.groupValues[CREATE_TIME_INDEX].toLong())
> ```

you should probably incorporate a UUID in case this is done in parallel with multiple clients.

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1681)* status: 6 of 10 files reviewed, 4 unresolved discussions (waiting on @robinsons)

___
*[`src/main/kotlin/org/wfanet/panelmatch/common/certificates/KingomlessCertificateManager.kt` line 162 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1681#-O1NuQjX0FswuZjjitcz:-O1NuQjX0FswuZjjitd-:b2vw0s6) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/42bd0101a4e2f6aa42c85c7e339f9823f0f6a775/src/main/kotlin/org/wfanet/panelmatch/common/certificates/KingomlessCertificateManager.kt#L162)):*
> ```Kotlin
>       private val CERT_NAME_PATTERN = Regex("(.+)$SEPARATOR([0-9]+)")
> 
>       fun fromCertName(certName: String): CertificateKey? {
> ```

can you add a test for this function?

stevenwarejones (commented):
> Reviewed 2 of 6 files at r2, 1 of 2 files at r3.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1681)* status: 7 of 10 files reviewed, 1 unresolved discussion (waiting on @robinsons)

robinsons (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1681)* status: 7 of 10 files reviewed, 1 unresolved discussion (waiting on @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/panelmatch/common/certificates/KingomlessCertificateManager.kt` line 64 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1681#-O1J2oyK4f1aSHB628UM:-O1O2eBI9Sy-Dpp2Yb64:b-gz7whk) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f0d8a9efe8b3b32c2ba4851203e529520334e42f/src/main/kotlin/org/wfanet/panelmatch/common/certificates/KingomlessCertificateManager.kt#L64)):*

my main concern was removing the field from the certificate proto. The rest is lower priority.
</blockquote></details>

Got it, thank you for clarifying.
___
*[`src/main/kotlin/org/wfanet/panelmatch/common/certificates/KingomlessCertificateManager.kt` line 58 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1681#-O1NtmvX2vFgsn5FPtKg:-O1OL8BX60nFs4KM6dvQ:b-37b4ts) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/42bd0101a4e2f6aa42c85c7e339f9823f0f6a775/src/main/kotlin/org/wfanet/panelmatch/common/certificates/KingomlessCertificateManager.kt#L58)):*

can you also add something verifying that the ownerId maps to this exchange. we should have that mapping, right?
</blockquote></details>

We can get it from the exchange workflow, yes. Added this.
___
*[`src/main/kotlin/org/wfanet/panelmatch/common/certificates/KingomlessCertificateManager.kt` line 162 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1681#-O1NuQjX0FswuZjjitcz:-O1OIEQJ90t-Yby7a-Vd:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/42bd0101a4e2f6aa42c85c7e339f9823f0f6a775/src/main/kotlin/org/wfanet/panelmatch/common/certificates/KingomlessCertificateManager.kt#L162)):*

can you add a test for this function?
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/panelmatch/common/certificates/KingomlessCertificateManager.kt` line 165 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1681#-O1NuHZkCkbKKYzvaofh:-O1O9HA_8M2fd1TXZSsG:b-1y5b2m) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/42bd0101a4e2f6aa42c85c7e339f9823f0f6a775/src/main/kotlin/org/wfanet/panelmatch/common/certificates/KingomlessCertificateManager.kt#L165)):*

you should probably incorporate a UUID in case this is done in parallel with multiple clients.
</blockquote></details>

Replaced the timestamp with a UUID. The timestamp was supposed to accomplish this, but UUID is a better bet for uniqueness.

stevenwarejones (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1681)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @robinsons)

PR #1685, opened 2024-07-10
Implement ApiClient for Kingdom-less panel match protocol.

robinsons:


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1685)

stevenwarejones (changes requested):
> Reviewed 1 of 11 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1685)* status: 1 of 11 files reviewed, 1 unresolved discussion (waiting on @robinsons)

___
*[`src/main/kotlin/org/wfanet/panelmatch/client/launcher/KingdomlessApiClient.kt` line 506 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1685#-O1rzF771sgKs6Dh_Got:-O1rzF78Bj8WoVzhst1Z:b-tuy5z7) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/868e554b42901dff143c400938744351c196939c/src/main/kotlin/org/wfanet/panelmatch/client/launcher/KingdomlessApiClient.kt#L506)):*
> ```Kotlin
>               while (date >= minDate) {
>                 yield(ExchangeDateKey(recurringExchangeId, date))
>                 date = date.minusDays(1L)
> ```

so, the iterator starts with the newest exchange dates to process first? It should be the opposite. Oldest should be processed first. In fact, the best is if the oldest across all exchanges is processed first and then progressively to newer dates since older data expires.

robinsons (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1685)* status: 1 of 11 files reviewed, 1 unresolved discussion (waiting on @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/panelmatch/client/launcher/KingdomlessApiClient.kt` line 506 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1685#-O1rzF771sgKs6Dh_Got:-O1vtarJ7JUj3U8NyncL:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/868e554b42901dff143c400938744351c196939c/src/main/kotlin/org/wfanet/panelmatch/client/launcher/KingdomlessApiClient.kt#L506)):*

so, the iterator starts with the newest exchange dates to process first? It should be the opposite. Oldest should be processed first. In fact, the best is if the oldest across all exchanges is processed first and then progressively to newer dates since older data expires.
</blockquote></details>

Done.

PR #1698, opened 2024-07-22
Prepare example Kingdom-less exchange daemon for GKE.

robinsons:
Summary:
* Creates `.cue` files for deploying Kingdom-less daemon to GKE.
* Adds Kingdom-less daemon flags.
    * Uses `@ArgGroup` to ensure mutual exclusitivity between Kingdom-based flags and Kingdom-less flags.
    * Updates the example daemon main class to build the correct type of `ApiClient` and `CertificateManager` based on which group of flags was set.
* Several other tweaks to make the deployment work:
    * Factory function for building a shared storage client needed to be `suspend`ing.
    * Explicitly call `withDefaultCredentails()` when building `GcpKmsClient`. This used to happen by default, but seems to have been (inadvertently?) removed in a previous PR.
    * In `KingdomlessApiClient.kt`:
        * Make sure that `ExchangeWorkflowDependencyGraph` is built in step order. The dependency graph will throw if e.g. you try to mark a step as completed before its prerequisite steps.
        * Add retries when loading checkpoints from shared storage, as a temporary measure to mitigate race conditions. A longer term solution will be to add a direct upload option to `StorageClient`, but need to investigate this further.

### DIFF
--- a/build/variables.bzl
+++ b/build/variables.bzl
@@ -140,6 +140,13 @@ EXAMPLE_PANEL_EXCHANGE_CLIENT_DAEMON_CONFIG = struct(
     mp_name = "$(mp_name)",
 )
 
+# Config for Kingdom-less Panel Exchange Client Example Daemon.
+EXAMPLE_KINGDOMLESS_PANEL_EXCHANGE_CLIENT_DAEMON_CONFIG = struct(
+    edp_id = "$(edp_id)",
+    mp_id = "$(mp_id)",
+    recurring_exchange_ids = "$(recurring_exchange_ids)",
+)
+
 # Settings for deploying tests to Google Cloud.
 PANEL_EXCHANGE_CLIENT_TEST_GOOGLE_CLOUD_SETTINGS = struct(
     secret_name = "$(k8s_secret_name)",
--- a/src/main/k8s/panelmatch/base.cue
+++ b/src/main/k8s/panelmatch/base.cue
@@ -103,18 +103,20 @@ objects: [ for objectSet in objectSets for object in objectSet {object}]
 }
 
 #Deployment: Deployment={
-	_name:       string
-	_secretName: string
-	_ports:      [{containerPort: #GrpcServicePort}] | *[]
-	_component:  string
-	_jvmFlags:   string | *""
+	_name:        string
+	_secretName?: string
+	_ports:       [{containerPort: #GrpcServicePort}] | *[]
+	_component:   string
+	_jvmFlags:    string | *""
 	_dependencies: [...string]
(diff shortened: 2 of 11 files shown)


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1698)

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1698)* status: all files reviewed, 1 unresolved discussion (waiting on @robinsons)

___
*[`src/main/kotlin/org/wfanet/panelmatch/client/launcher/KingdomlessApiClient.kt` line 639 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1698#-O2kvLwMC4aXd_sBfT1T:-O2kvLwMC4aXd_sBfT1U:ba44q4x) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/40a3ae8d6808005287d97e7c7dec413a4f630440/src/main/kotlin/org/wfanet/panelmatch/client/launcher/KingdomlessApiClient.kt#L639)):*
> ```Kotlin
>       try {
>         return block()
>       } catch (e: Exception) {
> ```

this should only retry nonfatal exceptions

robinsons (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1698)* status: 9 of 11 files reviewed, 1 unresolved discussion (waiting on @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/panelmatch/client/launcher/KingdomlessApiClient.kt` line 639 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1698#-O2kvLwMC4aXd_sBfT1T:-O2l5h8m5TWjFu88D2BV:b-g4q1dk) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/40a3ae8d6808005287d97e7c7dec413a4f630440/src/main/kotlin/org/wfanet/panelmatch/client/launcher/KingdomlessApiClient.kt#L639)):*

this should only retry nonfatal exceptions
</blockquote></details>

Updated to allow the caller to do some very basic exception handling and suppress specific exception types.

stevenwarejones (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1698)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @robinsons)

PR #1703, opened 2024-07-26
Add flag to allow VidSamplingInterval to wrap around 1

renjiezh:
Kingdom: only allow HMSS protocol measurement to wrap interval. Otherwise exception will be raised during creation.
Duchy: No change as it only care about the width.
Reporting: Add a flag to enable wrapping. When disabled, all wrapping will cause exception. When enabled, only multi-dataprovider measurement allow wrapping.
PBM: Allow wrapping. Updated the code of filter .


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703)

SanjayVas (changes requested):
> Reviewed 7 of 13 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703)* status: 7 of 13 files reviewed, 2 unresolved discussions (waiting on @renjiezh, @tristanvuong2021, and @uakyol)

___
*[`src/main/kotlin/org/wfanet/measurement/measurementconsumer/stats/Covariances.kt` line 196 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703#-O2kw3Oc88boWLl0qGIQ:-O2kw3Oc88boWLl0qGIR:b-qstkcf) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e6c683c724f680fd96b942d21d59b08b114b93d7/src/main/kotlin/org/wfanet/measurement/measurementconsumer/stats/Covariances.kt#L196)):*
> ```Kotlin
> 
>   /** Returns a list of non-wrapping interval represented by pairs of <start, end> */
>   private fun VidSamplingInterval.toNonWrappingIntervals(): List<Pair<Double, Double>> {
> ```

Use specific types rather than Pair. In this case, `ClosedRange<Double>`



_Code quote:_
```Kotlin
Pair<Double, Double>
```
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/common/ReportingApiServerFlags.kt` line 49 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703#-O2kwm5BBVxLUw-dz1g1:-O2kwm5BBVxLUw-dz1g2:bf323lj) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e6c683c724f680fd96b942d21d59b08b114b93d7/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/common/ReportingApiServerFlags.kt#L49)):*
> ```Kotlin
> 
>   @set:CommandLine.Option(
>     names = ["--allow-sampling-interval-wrapping"],
> ```

The API documentation in this PR is written as if this flag is always true. If there's a flag, the API documentation should take that into account.

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703)* status: 6 of 13 files reviewed, 2 unresolved discussions (waiting on @SanjayVas, @tristanvuong2021, and @uakyol)

___
*[`src/main/kotlin/org/wfanet/measurement/measurementconsumer/stats/Covariances.kt` line 196 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703#-O2kw3Oc88boWLl0qGIQ:-O2l2LL77ihhdFES_0MY:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e6c683c724f680fd96b942d21d59b08b114b93d7/src/main/kotlin/org/wfanet/measurement/measurementconsumer/stats/Covariances.kt#L196)):*

Use specific types rather than Pair. In this case, `ClosedRange<Double>`
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/common/ReportingApiServerFlags.kt` line 49 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703#-O2kwm5BBVxLUw-dz1g1:-O2l2OE08dLQ6Y_-Z0vT:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e6c683c724f680fd96b942d21d59b08b114b93d7/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/common/ReportingApiServerFlags.kt#L49)):*

The API documentation in this PR is written as if this flag is always true. If there's a flag, the API documentation should take that into account.
</blockquote></details>

Done.

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703)* status: 7 of 13 files reviewed, 1 unresolved discussion (waiting on @renjiezh, @robinsons, @tristanvuong2021, and @uakyol)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/common/ReportingApiServerFlags.kt` line 49 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703#-O2kwm5BBVxLUw-dz1g1:-O2lUIfQ-WjnEU4RfPba:b-htqbm8) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e6c683c724f680fd96b942d21d59b08b114b93d7/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/common/ReportingApiServerFlags.kt#L49)):*

Done.
</blockquote></details>

I'm not seeing any change reflecting this in metric.proto

robinsons (changes requested):
> Reviewed 12 of 13 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703)* status: all files reviewed, 2 unresolved discussions (waiting on @renjiezh, @tristanvuong2021, and @uakyol)

___
*[`src/main/kotlin/org/wfanet/measurement/measurementconsumer/stats/Covariances.kt` line 201 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703#-O2z4fxN7wL-TvgObMnN:-O2z4fxN7wL-TvgObMnO:bcuurx8) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/2ee6921821616588b05a8513263d1e8387d1b100/src/main/kotlin/org/wfanet/measurement/measurementconsumer/stats/Covariances.kt#L201)):*
> ```Kotlin
> 
>     return if (end > 1.0) {
>       listOf(0.0..start, end..1.0)
> ```

If `end > 1.0`, then `end..1.0` would be empty, right?

Is this supposed to be `listOf(0.0..(end-1.0), start..1.0)`?

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703)* status: 11 of 13 files reviewed, 2 unresolved discussions (waiting on @robinsons, @SanjayVas, @tristanvuong2021, and @uakyol)

___
*[`src/main/kotlin/org/wfanet/measurement/measurementconsumer/stats/Covariances.kt` line 201 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703#-O2z4fxN7wL-TvgObMnN:-O2zia5Z3WetypWevBJd:b6rs0kz) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/2ee6921821616588b05a8513263d1e8387d1b100/src/main/kotlin/org/wfanet/measurement/measurementconsumer/stats/Covariances.kt#L201)):*

If `end > 1.0`, then `end..1.0` would be empty, right?

Is this supposed to be `listOf(0.0..(end-1.0), start..1.0)`?
</blockquote></details>

Done.
Nice catch.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/common/ReportingApiServerFlags.kt` line 49 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703#-O2kwm5BBVxLUw-dz1g1:-O2ziU3GCUW2OX9a0qYN:brbanft) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e6c683c724f680fd96b942d21d59b08b114b93d7/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/common/ReportingApiServerFlags.kt#L49)):*

I'm not seeing any change reflecting this in metric.proto
</blockquote></details>

Done.
Sorry I misunderstood.

SanjayVas (approved):
> Reviewed 6 of 13 files at r1, 1 of 2 files at r3, 1 of 1 files at r4, 2 of 2 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703)* status: all files reviewed, 3 unresolved discussions (waiting on @ple13, @renjiezh, @robinsons, @tristanvuong2021, and @uakyol)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricSpecDefaults.kt` line 714 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703#-O2zqca_4yUIa-SNXVY9:-O2zqca_4yUIa-SNXVYA:b-vy54q8) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4be33d8c8da7a43c6499c1b15d764d38523e02e2/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricSpecDefaults.kt#L714)):*
> ```Kotlin
>   } else {
>     // The 10000 is to help turn the float into an int without losing too much data.
>     val maxStart =
> ```

nit: add "scaled" to the variable names where the value is scaled



_Code quote:_
```Kotlin
maxStart
```
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricSpecDefaults.kt` line 716 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703#-O2zpwjY7zIEcRvWMkbB:-O2zpwjY7zIEcRvWMkbC:b-1y8gnb) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4be33d8c8da7a43c6499c1b15d764d38523e02e2/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricSpecDefaults.kt#L716)):*
> ```Kotlin
>     val maxStart =
>       if (allowSamplingIntervalWrapping) {
>         10000
> ```

nit: extract this scaling factor to a constant



_Code quote:_
```Kotlin
10000
```

robinsons (approved):
> Reviewed 2 of 2 files at r3, 2 of 2 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703)* status: all files reviewed, 2 unresolved discussions (waiting on @ple13, @renjiezh, @tristanvuong2021, and @uakyol)

uakyol (commented):
Taking myself out of review since @robinsons is going to be the leading pbm effort going forward

ple13 (approved):
> Reviewed 11 of 13 files at r1, 2 of 2 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703)* status: all files reviewed, 5 unresolved discussions (waiting on @renjiezh and @tristanvuong2021)

___
*[`src/main/proto/wfa/measurement/config/reporting/measurement_spec_config.proto` line 39 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703#-O348B1f3OCN8u5XcCI8:-O348B1f3OCN8u5XcCI9:b210xw5) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a93cf5c38d775855d112994760fe05d859348112/src/main/proto/wfa/measurement/config/reporting/measurement_spec_config.proto#L39)):*
> ```proto
>   // Specifies a range of VIDs to be sampled.
>   //
>   // The interval is within range of [0.0, 1.0]. If start + width > 0, it will
> ```

nit: start + width > 1.0?
___
*[`src/main/proto/wfa/measurement/config/reporting/metric_spec_config.proto` line 40 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703#-O348J9SCvDb6UoVURGV:-O348J9SCvDb6UoVURGW:b-5v3gaq) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a93cf5c38d775855d112994760fe05d859348112/src/main/proto/wfa/measurement/config/reporting/metric_spec_config.proto#L40)):*
> ```proto
>   // Specifies a range of VIDs to be sampled.
>   //
>   // The interval is within range of [0.0, 1.0]. If start + width > 0, it will
> ```

nit: ditto.
___
*[`src/main/proto/wfa/measurement/reporting/v2alpha/metric.proto` line 45 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703#-O348Yv6A5b9230HzeCX:-O348Yv6A5b9230HzeCY:b-5v3gaq) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a93cf5c38d775855d112994760fe05d859348112/src/main/proto/wfa/measurement/reporting/v2alpha/metric.proto#L45)):*
> ```proto
>   // The interval is within range of [0.0, 1.0].
>   // When --allow-sampling-interval-wrapping feature flag is enabled, if start +
>   // width > 0, it will wrap around 1. E.g., start=0.8, width=0.5 stands for
> ```

nit: ditto.

tristanvuong2021 (changes requested):
> Reviewed 7 of 13 files at r1, 1 of 2 files at r5, 4 of 4 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703)* status: all files reviewed, 1 unresolved discussion (waiting on @renjiezh and @stevenwarejones)

___
*[`src/test/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricSpecDefaultsTest.kt` line 1865 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703#-O39MjuP6l-rH8yx-ijL:-O39MjuQ-m_E_oZfhMDC:b-45zy35) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d22668f68543e84883cbe1b80339449f52c53b6d/src/test/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricSpecDefaultsTest.kt#L1865)):*
> ```Kotlin
> 
>   @Test
>   fun `buildMetricSpec throws MetricSpecBuildingException when allowSamplingIntervalWrapping disabled`() {
> ```

Missing test for width > 1

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703)* status: 12 of 13 files reviewed, 1 unresolved discussion (waiting on @ple13, @robinsons, @SanjayVas, @stevenwarejones, and @tristanvuong2021)

___
*[`src/test/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricSpecDefaultsTest.kt` line 1865 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703#-O39MjuP6l-rH8yx-ijL:-O39i4Gn7G99WewTXUIY:biz55rt) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d22668f68543e84883cbe1b80339449f52c53b6d/src/test/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricSpecDefaultsTest.kt#L1865)):*

Missing test for width > 1
</blockquote></details>

Done.
Thanks for catching that.

tristanvuong2021 (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @stevenwarejones)

stevenwarejones (changes requested):
> Reviewed 6 of 13 files at r1, 1 of 2 files at r5, 3 of 4 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703)* status: all files reviewed, 1 unresolved discussion (waiting on @renjiezh)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/common/ReportingApiServerFlags.kt` line 48 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703#-O3D_6lL0ILUQmGqDgaw:-O3D_6lL0ILUQmGqDgax:b-3r1d2f) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/591556264b768928feca93f1dbb9887a1604a1a6/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/common/ReportingApiServerFlags.kt#L48)):*
> ```Kotlin
>     private set
> 
>   @set:CommandLine.Option(
> ```

i don't understand why a flag is desirable. Shouldn't `allowSamplingIntervalWrapping` be part of the protocol setup config for each protocol?

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703)* status: all files reviewed, 1 unresolved discussion (waiting on @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/common/ReportingApiServerFlags.kt` line 48 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703#-O3D_6lL0ILUQmGqDgaw:-O3D_m3W-u15OFLdUepN:b-9du1in) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/591556264b768928feca93f1dbb9887a1604a1a6/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/common/ReportingApiServerFlags.kt#L48)):*

i don't understand why a flag is desirable. Shouldn't `allowSamplingIntervalWrapping` be part of the protocol setup config for each protocol?
</blockquote></details>

Good question.
Sampling interval wrapping is only for HMSS protocol. We don't want to allow it for LLv2. However, when reporting server creating measurements, it isn't aware of the protocol. It only knows the measurement type (reach or r/f, etc). So we need the flag to enable it only when HMSS is widely used.

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703)* status: all files reviewed, 1 unresolved discussion (waiting on @renjiezh)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/common/ReportingApiServerFlags.kt` line 48 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703#-O3D_6lL0ILUQmGqDgaw:-O3Yr5oI2yS8Lu3OLSrA:b-qizk9m) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/591556264b768928feca93f1dbb9887a1604a1a6/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/common/ReportingApiServerFlags.kt#L48)):*

Good question.
Sampling interval wrapping is only for HMSS protocol. We don't want to allow it for LLv2. However, when reporting server creating measurements, it isn't aware of the protocol. It only knows the measurement type (reach or r/f, etc). So we need the flag to enable it only when HMSS is widely used.
</blockquote></details>

Then I would argue this logic (including vid interval selection logic) should be in the same place where the protocol is selected. How it is now is confusing. Since the reporting API never knows what protocol will be selected, how can it know whether that protocol supports interval wrapping?

SanjayVas (approved):
> Reviewed 4 of 4 files at r6, 1 of 1 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703)* status: all files reviewed, 1 unresolved discussion (waiting on @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/common/ReportingApiServerFlags.kt` line 48 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703#-O3D_6lL0ILUQmGqDgaw:-O3Z2WSm1tugQYOph_9e:b-5kv8ac) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/591556264b768928feca93f1dbb9887a1604a1a6/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/common/ReportingApiServerFlags.kt#L48)):*
> Since the reporting API never knows what protocol will be selected, how can it know whether that protocol supports interval wrapping?

I believe that all protocols support interval wrapping. @renjiezh what's the reason that we don't want to flip the flag until HMSS is predominant? Meaning why do we prefer that it's not used with LLv2?

kungfucraig (commented):
> Reviewed 1 of 13 files at r1.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703)* status: all files reviewed, 1 unresolved discussion (waiting on @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/common/ReportingApiServerFlags.kt` line 48 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703#-O3D_6lL0ILUQmGqDgaw:-O3Z4mym0RpRazun_L82:bsfycze) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/591556264b768928feca93f1dbb9887a1604a1a6/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/common/ReportingApiServerFlags.kt#L48)):*

> Since the reporting API never knows what protocol will be selected, how can it know whether that protocol supports interval wrapping?

I believe that all protocols support interval wrapping. @renjiezh what's the reason that we don't want to flip the flag until HMSS is predominant? Meaning why do we prefer that it's not used with LLv2?
</blockquote></details>

This is an EDP contract change. They are not required to support wrapping until HMSS is launched.

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703)* status: all files reviewed, 1 unresolved discussion (waiting on @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/common/ReportingApiServerFlags.kt` line 48 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703#-O3D_6lL0ILUQmGqDgaw:-O3Z7G4J4f1UMYZzphbH:b-2amj7k) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/591556264b768928feca93f1dbb9887a1604a1a6/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/common/ReportingApiServerFlags.kt#L48)):*

This is an EDP contract change. They are not required to support wrapping until HMSS is launched.
</blockquote></details>

I assume EDPs are not fully support interval wrapping,  so by default wrapping is not allowed for direct or MPC measurement. However, with HMSS EDP library, EDPs will support wrapping for reach-only and r/f MPC measurements. The flag in this PR is only aiming for HMSS.

kungfucraig (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703)* status: all files reviewed, 1 unresolved discussion (waiting on @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/common/ReportingApiServerFlags.kt` line 48 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703#-O3D_6lL0ILUQmGqDgaw:-O3ZDbPC70w8a22UXr5f:b-przq4s) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/591556264b768928feca93f1dbb9887a1604a1a6/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/common/ReportingApiServerFlags.kt#L48)):*

I assume EDPs are not fully support interval wrapping,  so by default wrapping is not allowed for direct or MPC measurement. However, with HMSS EDP library, EDPs will support wrapping for reach-only and r/f MPC measurements. The flag in this PR is only aiming for HMSS.
</blockquote></details>

Exactly. We will only enable this flag once HMSS is fully launched as HMSS requires that EDPs support wrapping.

I also think EDPs will need to support wrapping more generally (i.e. for direct), but since we use the full interval for direct it shouldn't matter all that much.

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703)* status: all files reviewed, 1 unresolved discussion (waiting on @kungfucraig, @renjiezh, and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/common/ReportingApiServerFlags.kt` line 48 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703#-O3D_6lL0ILUQmGqDgaw:-O3c2wyeELbRLrzHvlax:b40jkr7) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/591556264b768928feca93f1dbb9887a1604a1a6/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/common/ReportingApiServerFlags.kt#L48)):*

Exactly. We will only enable this flag once HMSS is fully launched as HMSS requires that EDPs support wrapping.

I also think EDPs will need to support wrapping more generally (i.e. for direct), but since we use the full interval for direct it shouldn't matter all that much.
</blockquote></details>

should vid interval be a capability then? If that's the case, we don't need this flag.

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703)* status: all files reviewed, 1 unresolved discussion (waiting on @kungfucraig, @renjiezh, and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/common/ReportingApiServerFlags.kt` line 48 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703#-O3D_6lL0ILUQmGqDgaw:-O3c3JaB2uVM6EeKWBIV:bo9y5wx) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/591556264b768928feca93f1dbb9887a1604a1a6/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/common/ReportingApiServerFlags.kt#L48)):*

should vid interval be a capability then? If that's the case, we don't need this flag.
</blockquote></details>

^meaning vid interval wrapping as a capability

kungfucraig (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703)* status: all files reviewed, 1 unresolved discussion (waiting on @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/common/ReportingApiServerFlags.kt` line 48 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703#-O3D_6lL0ILUQmGqDgaw:-O3cSR2L7u5rFXprpbZk:bezrbkn) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/591556264b768928feca93f1dbb9887a1604a1a6/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/common/ReportingApiServerFlags.kt#L48)):*

^meaning vid interval wrapping as a capability
</blockquote></details>

The reporting server doesn't have a capabilities service. Moreover, it's implied by HMSS. For HMSS rollout we expect the following list of things to happen:

1. EDPs support HMSS and set their capability
2. Kingdom HMSS MC allow list has one MC added to it for testing
3. Kingdom HMSS is enabled fully
4. All EDPs support HMSS
5. Reporting server flag for wrapping intervals is enabled

SanjayVas (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703)* status: all files reviewed, 1 unresolved discussion (waiting on @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/common/ReportingApiServerFlags.kt` line 48 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703#-O3D_6lL0ILUQmGqDgaw:-O3cbNeO23J2D-TdaX2Z:btpf3cn) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/591556264b768928feca93f1dbb9887a1604a1a6/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/common/ReportingApiServerFlags.kt#L48)):*

The reporting server doesn't have a capabilities service. Moreover, it's implied by HMSS. For HMSS rollout we expect the following list of things to happen:

1. EDPs support HMSS and set their capability
2. Kingdom HMSS MC allow list has one MC added to it for testing
3. Kingdom HMSS is enabled fully
4. All EDPs support HMSS
5. Reporting server flag for wrapping intervals is enabled
</blockquote></details>

From the EDP point-of-view, it's already controlled by the HMSS capability. The Kingdom will throw an error if the value is >1 and the HMSS protocol is not selected, and the HMSS protocol can only be selected if all involved EDPs have the HMSS capability.

As you pointed out, the Reporting system is removed from capabilities and protocol decisions. Despite this, the caller of the Reporting API still needs to be able to specify a valid interval in the request. Furthermore, the Reporting API documentation must be clear on the behavior for the field.

FWIW, I'm not the happiest about how this works either. I'm just not sure we can do better given the design constraints:

1. The Reporting server does not (and should not) know about what protocol may be selected
2. The CreateReport caller must be allowed to specify the sampling interval

SanjayVas (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703)* status: all files reviewed, 1 unresolved discussion (waiting on @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/common/ReportingApiServerFlags.kt` line 48 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703#-O3D_6lL0ILUQmGqDgaw:-O3ceUxHCEbYGNwPIAqk:b-6ms4tc) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/591556264b768928feca93f1dbb9887a1604a1a6/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/common/ReportingApiServerFlags.kt#L48)):*

From the EDP point-of-view, it's already controlled by the HMSS capability. The Kingdom will throw an error if the value is >1 and the HMSS protocol is not selected, and the HMSS protocol can only be selected if all involved EDPs have the HMSS capability.

As you pointed out, the Reporting system is removed from capabilities and protocol decisions. Despite this, the caller of the Reporting API still needs to be able to specify a valid interval in the request. Furthermore, the Reporting API documentation must be clear on the behavior for the field.

FWIW, I'm not the happiest about how this works either. I'm just not sure we can do better given the design constraints:

1. The Reporting server does not (and should not) know about what protocol may be selected
2. The CreateReport caller must be allowed to specify the sampling interval
</blockquote></details>

@kungfucraig for the period of time when both HMSS and LLv2 are enabled, does that mean that interval wrapping will just not be supported? Is that okay?

kungfucraig (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703)* status: all files reviewed, 1 unresolved discussion (waiting on @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/common/ReportingApiServerFlags.kt` line 48 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703#-O3D_6lL0ILUQmGqDgaw:-O3cpZGQ9zJ2DMS3OcDk:b4f7tvy) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/591556264b768928feca93f1dbb9887a1604a1a6/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/common/ReportingApiServerFlags.kt#L48)):*

@kungfucraig for the period of time when both HMSS and LLv2 are enabled, does that mean that interval wrapping will just not be supported? Is that okay?
</blockquote></details>

@SanjayVas Yeah, I think that's the most reasonable way of dealing with this as the period of time they will both be supported should be rather short, and the wrapping interval is just an optimization not something that is strictly required.

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703)* status: all files reviewed, 1 unresolved discussion (waiting on @kungfucraig, @renjiezh, and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/common/ReportingApiServerFlags.kt` line 48 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703#-O3D_6lL0ILUQmGqDgaw:-O3cfKOxD7qOY2A5_wRk:btdghuo) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/591556264b768928feca93f1dbb9887a1604a1a6/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/common/ReportingApiServerFlags.kt#L48)):*

@SanjayVas Yeah, I think that's the most reasonable way of dealing with this as the period of time they will both be supported should be rather short, and the wrapping interval is just an optimization not something that is strictly required.
</blockquote></details>

1. As I understand this flag, as long as any EDP doesn't support HMSS, then this flag should not be enabled right?
2. Also, why must the CreateReport caller be allowed to specify the sampling interval?
3. What will be the behavior if this flag is turned on for a measurement with an EDP that does not support HMSS?

kungfucraig (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703)* status: all files reviewed, 1 unresolved discussion (waiting on @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/common/ReportingApiServerFlags.kt` line 48 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703#-O3D_6lL0ILUQmGqDgaw:-O3dCzUVCPD9xAYjHqyN:b-281v8b) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/591556264b768928feca93f1dbb9887a1604a1a6/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/common/ReportingApiServerFlags.kt#L48)):*

1. As I understand this flag, as long as any EDP doesn't support HMSS, then this flag should not be enabled right?
2. Also, why must the CreateReport caller be allowed to specify the sampling interval?
3. What will be the behavior if this flag is turned on for a measurement with an EDP that does not support HMSS?
</blockquote></details>

@stevenwarejones 

1. Yes. It can only be enabled after all EDPs support HMSS.
2. To allow the caller to optimize the accuracy/privacy budget trade-off. We have defaults for these. 
3. It depends on the inputs. For example start=0, width=0.5 should be fine regardless. However, when the interval truly wraps (e.g. start=0.8 width=0.5), I'd expect an error because the EDP should calculate end=0.8+0.5=1.3, which is greater than one. 

Basically, the flag should not be activated until all EDPs support HMSS. Also note that this behavior is built into the client libraries for frequency vector creation.

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703)* status: all files reviewed, 1 unresolved discussion (waiting on @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/common/ReportingApiServerFlags.kt` line 48 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703#-O3D_6lL0ILUQmGqDgaw:-O3dCuNGChKNIHCcIdKP:bmjczh6) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/591556264b768928feca93f1dbb9887a1604a1a6/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/common/ReportingApiServerFlags.kt#L48)):*

@stevenwarejones 

1. Yes. It can only be enabled after all EDPs support HMSS.
2. To allow the caller to optimize the accuracy/privacy budget trade-off. We have defaults for these. 
3. It depends on the inputs. For example start=0, width=0.5 should be fine regardless. However, when the interval truly wraps (e.g. start=0.8 width=0.5), I'd expect an error because the EDP should calculate end=0.8+0.5=1.3, which is greater than one. 

Basically, the flag should not be activated until all EDPs support HMSS. Also note that this behavior is built into the client libraries for frequency vector creation.
</blockquote></details>

1. Yes.
2. CreateReport caller is the MeasurementConsumer, they are allowed to specify the interval just like they create Measurements directly.
3. When the flag is on, the wrapping interval will be passes to the kingdom to create Measurements. The kingdom checks the capabilities of EDPs and chooses LLv2 instead of HMSS. Then it sees wrapping interval is not allowed for LLv2, thus error raised from the Kingdom.

kungfucraig (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703)* status: all files reviewed, 1 unresolved discussion (waiting on @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/common/ReportingApiServerFlags.kt` line 48 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703#-O3D_6lL0ILUQmGqDgaw:-O3dE6D2Bswbc_S7bzwV:b2yb2h6) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/591556264b768928feca93f1dbb9887a1604a1a6/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/common/ReportingApiServerFlags.kt#L48)):*

1. Yes.
2. CreateReport caller is the MeasurementConsumer, they are allowed to specify the interval just like they create Measurements directly.
3. When the flag is on, the wrapping interval will be passes to the kingdom to create Measurements. The kingdom checks the capabilities of EDPs and chooses LLv2 instead of HMSS. Then it sees wrapping interval is not allowed for LLv2, thus error raised from the Kingdom.
</blockquote></details>

I defer to @renjiezh on #3, but the EDP should also throw an error if it ever gets to that point.

kungfucraig (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703)* status: all files reviewed, 1 unresolved discussion (waiting on @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/common/ReportingApiServerFlags.kt` line 48 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703#-O3D_6lL0ILUQmGqDgaw:-O3dEDMz3_cUwHn2eZbQ:bgpeoko) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/591556264b768928feca93f1dbb9887a1604a1a6/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/common/ReportingApiServerFlags.kt#L48)):*

I defer to @renjiezh on #3, but the EDP should also throw an error if it ever gets to that point.
</blockquote></details>

I defer to @renjiezh on number three.

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703)* status: all files reviewed, 1 unresolved discussion (waiting on @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/common/ReportingApiServerFlags.kt` line 48 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703#-O3D_6lL0ILUQmGqDgaw:-O3dEONODenHGYAyMIe7:bjonj3u) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/591556264b768928feca93f1dbb9887a1604a1a6/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/common/ReportingApiServerFlags.kt#L48)):*

I defer to @renjiezh on number three.
</blockquote></details>

EDP has its own implementation of checking interval. But from Kingdom's point of view, it will never successfully create a llv2 with wrapping interval.

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703)* status: all files reviewed, 1 unresolved discussion (waiting on @kungfucraig, @renjiezh, and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/common/ReportingApiServerFlags.kt` line 48 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703#-O3D_6lL0ILUQmGqDgaw:-O3dFSG21GdZvXGMuiCs:b-6l2ig5) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/591556264b768928feca93f1dbb9887a1604a1a6/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/common/ReportingApiServerFlags.kt#L48)):*

EDP has its own implementation of checking interval. But from Kingdom's point of view, it will never successfully create a llv2 with wrapping interval.
</blockquote></details>

where's the kingdom logic that would throw an error if llv2 gets selected but the vid interval wraps? because it should never make it to EDPs and have them waste cycles as some EDPs might fulfill a wasted requisition assuming others will reject.

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703)* status: all files reviewed, 1 unresolved discussion (waiting on @kungfucraig, @renjiezh, and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/common/ReportingApiServerFlags.kt` line 48 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703#-O3D_6lL0ILUQmGqDgaw:-O3dG-Sm7eFQbe2Kw7lQ:bklc8gu) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/591556264b768928feca93f1dbb9887a1604a1a6/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/common/ReportingApiServerFlags.kt#L48)):*

where's the kingdom logic that would throw an error if llv2 gets selected but the vid interval wraps? because it should never make it to EDPs and have them waste cycles as some EDPs might fulfill a wasted requisition assuming others will reject.
</blockquote></details>

why can't we roll out this flag after llv2 is fully deprecated? Why can't we start running hmshuffle without interval wrapping and revisit this flag once llv2 is completely gone. That simplifies things.

ple13 (approved):
> Reviewed 4 of 4 files at r6, 1 of 1 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703)* status: all files reviewed, 1 unresolved discussion (waiting on @kungfucraig, @renjiezh, and @SanjayVas)

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703)* status: all files reviewed, 1 unresolved discussion (waiting on @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/common/ReportingApiServerFlags.kt` line 48 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703#-O3D_6lL0ILUQmGqDgaw:-O3dOAOk4YFXG4AjhmtZ:b-mlbjuj) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/591556264b768928feca93f1dbb9887a1604a1a6/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/common/ReportingApiServerFlags.kt#L48)):*

why can't we roll out this flag after llv2 is fully deprecated? Why can't we start running hmshuffle without interval wrapping and revisit this flag once llv2 is completely gone. That simplifies things.
</blockquote></details>

[Here](https://github.com/world-federation-of-advertisers/cross-media-measurement/pull/1703/files#diff-b43880636c43431262a156112df3ad853a72045d67d236d28cc14b1f36fea390R850) is the Kingdom code checking the interval.
Yes, this flag will only be enabled once llv2 is fully deprecated. The PR is to merge the code in thus we enable it in the future.

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703)* status: all files reviewed, 1 unresolved discussion (waiting on @kungfucraig, @renjiezh, and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/common/ReportingApiServerFlags.kt` line 48 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703#-O3D_6lL0ILUQmGqDgaw:-O3dUzcLCqBvmKjXBSyy:bdkflmi) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/591556264b768928feca93f1dbb9887a1604a1a6/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/common/ReportingApiServerFlags.kt#L48)):*

[Here](https://github.com/world-federation-of-advertisers/cross-media-measurement/pull/1703/files#diff-b43880636c43431262a156112df3ad853a72045d67d236d28cc14b1f36fea390R850) is the Kingdom code checking the interval.
Yes, this flag will only be enabled once llv2 is fully deprecated. The PR is to merge the code in thus we enable it in the future.
</blockquote></details>

i'm saying to add the code in but not this flag until llv2 is fully deprecated so its not accidentally enabled.

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703)* status: all files reviewed, 1 unresolved discussion (waiting on @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/common/ReportingApiServerFlags.kt` line 48 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703#-O3D_6lL0ILUQmGqDgaw:-O3dVLG086XnRxXXgoaV:b8aiux5) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/591556264b768928feca93f1dbb9887a1604a1a6/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/common/ReportingApiServerFlags.kt#L48)):*

i'm saying to add the code in but not this flag until llv2 is fully deprecated so its not accidentally enabled.
</blockquote></details>

Sorry I might not fully catch it. Without the flag, how do we add the code without running it?

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703)* status: all files reviewed, 1 unresolved discussion (waiting on @kungfucraig, @renjiezh, and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/common/ReportingApiServerFlags.kt` line 48 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703#-O3D_6lL0ILUQmGqDgaw:-O3eN-GbDHOtI4LqwDUU:b-x2mp3z) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/591556264b768928feca93f1dbb9887a1604a1a6/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/common/ReportingApiServerFlags.kt#L48)):*

Sorry I might not fully catch it. Without the flag, how do we add the code without running it?
</blockquote></details>

I don't really agree with giving a flag that should not be used but will make it into a release unless that flag will fail on startup. ..... maybe one option is to get the list of available protocols from the kingdom at startup and if this flag is enabled then throw an error if llv2 is enabled.

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703)* status: all files reviewed, 1 unresolved discussion (waiting on @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/common/ReportingApiServerFlags.kt` line 48 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703#-O3D_6lL0ILUQmGqDgaw:-O3hdvscC6wHrqPPOGe5:b-vpx1vi) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/591556264b768928feca93f1dbb9887a1604a1a6/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/common/ReportingApiServerFlags.kt#L48)):*

I don't really agree with giving a flag that should not be used but will make it into a release unless that flag will fail on startup. ..... maybe one option is to get the list of available protocols from the kingdom at startup and if this flag is enabled then throw an error if llv2 is enabled.
</blockquote></details>

My main concern with the API design. The caller has no way to prevent an invalid argument error since they cannot (and should not) know the selected protocol ahead of time. Therefore the condition should be based on the EDP capabilities, not the selected protocol.

As discussed offline, add a new separate capability for VID interval wrapping (for all protocols). There can still be an additional flag to control whether the new behavior is enabled, but it must be in addition to the capability check.

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703)* status: all files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/common/ReportingApiServerFlags.kt` line 48 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703#-O3D_6lL0ILUQmGqDgaw:-O46E8YP52cH2cdby18u:bi6n0b5) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/591556264b768928feca93f1dbb9887a1604a1a6/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/common/ReportingApiServerFlags.kt#L48)):*

My main concern with the API design. The caller has no way to prevent an invalid argument error since they cannot (and should not) know the selected protocol ahead of time. Therefore the condition should be based on the EDP capabilities, not the selected protocol.

As discussed offline, add a new separate capability for VID interval wrapping (for all protocols). There can still be an additional flag to control whether the new behavior is enabled, but it must be in addition to the capability check.
</blockquote></details>

agreed.

stevenwarejones (changes requested):
> Reviewed 1 of 4 files at r6, 1 of 1 files at r7, 13 of 13 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703)* status: all files reviewed, 4 unresolved discussions (waiting on @renjiezh and @SanjayVas)

___
*[`src/main/proto/wfa/measurement/internal/kingdom/data_provider_capabilities.proto` line 27 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703#-OCFfQr8Ct53WI7ffsms:-OCFfQr8Ct53WI7ffsmt:bodiiv) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/10540e4f3d5483e9ec75d9b7327a001ab158b4da/src/main/proto/wfa/measurement/internal/kingdom/data_provider_capabilities.proto#L27)):*
> ```proto
>   // Whether the VID sampling interval wrapping around 1 is supported for LLv2
>   // protocol.
>   bool llv2_vid_sampling_interval_wrapping_supported = 2;
> ```

i say this is never supported. 
___
*[`src/main/proto/wfa/measurement/internal/kingdom/data_provider_capabilities.proto` line 30 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703#-OCFfTDQElr3aW3iI9R2:-OCFfTDQElr3aW3iI9R3:buo4gjs) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/10540e4f3d5483e9ec75d9b7327a001ab158b4da/src/main/proto/wfa/measurement/internal/kingdom/data_provider_capabilities.proto#L30)):*
> ```proto
>   // Whether the VID sampling interval wrapping around 1 is supported for HMSS
>   // protocol.
>   bool hmss_vid_sampling_interval_wrapping_supported = 3;
> ```

this capability should be implicit with `honest_majority_share_shuffle_supported`
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt` line 218 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703#-OCFexcJA9MmuCjrMTcs:-OCFexcJA9MmuCjrMTct:b-mkta55) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/10540e4f3d5483e9ec75d9b7327a001ab158b4da/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt#L218)):*
> ```Kotlin
>           DataProviderKt.capabilities {
>             honestMajorityShareShuffleSupported = (hmssVidIndexMap != null)
>             hmssVidSamplingIntervalWrappingSupported = (hmssVidIndexMap != null)
> ```

does this really need to be a new capability? All hmms edps should inherently support this.

SanjayVas (commented):
> Reviewed 1 of 13 files at r8.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703)* status: all files reviewed, 4 unresolved discussions (waiting on @renjiezh and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt` line 218 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703#-OCFexcJA9MmuCjrMTcs:-OCFoR8S9AHcDm3dQjy7:b-1ucqpt) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/10540e4f3d5483e9ec75d9b7327a001ab158b4da/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt#L218)):*
> All hmms edps should inherently support this.

Is this true? I thought that the API itself was documented to not accept a wrapped value before this, and therefore EDPs were free and correct to implement HMSS already without this.

We're trying to very cautious about gating features/capabilities going forward as part of enabling frequent automated deployments.
___
*[`src/main/proto/wfa/measurement/internal/kingdom/data_provider_capabilities.proto` line 30 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703#-OCFfTDQElr3aW3iI9R2:-OCFoNvo8dFzooX9_otL:bxnjuu4) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/10540e4f3d5483e9ec75d9b7327a001ab158b4da/src/main/proto/wfa/measurement/internal/kingdom/data_provider_capabilities.proto#L30)):*

this capability should be implicit with `honest_majority_share_shuffle_supported`
</blockquote></details>

See above comment re: whether this is true

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703)* status: all files reviewed, 4 unresolved discussions (waiting on @SanjayVas and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt` line 218 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703#-OCFexcJA9MmuCjrMTcs:-OCFqQfp6p2lVw6w0jRM:b-xlw417) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/10540e4f3d5483e9ec75d9b7327a001ab158b4da/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt#L218)):*

> All hmms edps should inherently support this.

Is this true? I thought that the API itself was documented to not accept a wrapped value before this, and therefore EDPs were free and correct to implement HMSS already without this.

We're trying to very cautious about gating features/capabilities going forward as part of enabling frequent automated deployments.
</blockquote></details>

If EDPs are using our HMSS library, yes, they support vid interval wrapping. But theoretically, HMSS itself doesn't require the support.

As steven mentioned below, LLv2 is not supporting interval wrapping, and HMSS implicitly means supporting interval wrapping. It means we can get rid of the extra capabilities?
@SanjayVas What do you think?

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703)* status: all files reviewed, 4 unresolved discussions (waiting on @kungfucraig, @renjiezh, and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt` line 218 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703#-OCFexcJA9MmuCjrMTcs:-OCFti2n9kCbqM3tPoTs:b-k77lct) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/10540e4f3d5483e9ec75d9b7327a001ab158b4da/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt#L218)):*
> HMSS implicitly means supporting interval wrapping

Since EDPs have already been free to implement HMSS without interval wrapping, it means that the above statement is not true. Therefore this has to have a new capability.

I think the other thing Steven is suggesting is never supporting wrapping on LLv2. I have no opinion of this, and would defer to @kungfucraig

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703)* status: all files reviewed, 4 unresolved discussions (waiting on @kungfucraig, @renjiezh, and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt` line 218 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703#-OCFexcJA9MmuCjrMTcs:-OCJke1T0R3qoFCfgCtP:b-ve8tr4) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/10540e4f3d5483e9ec75d9b7327a001ab158b4da/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt#L218)):*

> HMSS implicitly means supporting interval wrapping

Since EDPs have already been free to implement HMSS without interval wrapping, it means that the above statement is not true. Therefore this has to have a new capability.

I think the other thing Steven is suggesting is never supporting wrapping on LLv2. I have no opinion of this, and would defer to @kungfucraig
</blockquote></details>

No one has adopted HM Shuffle. I think we should assume it only supports wrapping. In fact, we would _never_ want an EDP to enable only the HM Shuffle capability without also supporting interval wrapping.

kungfucraig (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703)* status: all files reviewed, 4 unresolved discussions (waiting on @renjiezh, @SanjayVas, and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt` line 218 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703#-OCFexcJA9MmuCjrMTcs:-OC_PthqCFGeFsEns74q:b-axdoqg) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/10540e4f3d5483e9ec75d9b7327a001ab158b4da/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt#L218)):*

No one has adopted HM Shuffle. I think we should assume it only supports wrapping. In fact, we would _never_ want an EDP to enable only the HM Shuffle capability without also supporting interval wrapping.
</blockquote></details>

LLv2 should not have wrapping.

Anyone who uses the HM Shuffle support libraries gets the wrapping capability by default. This should include TechEdge. I've confirmed Google implemented wrapping.

Seems like we don't need this capability and can just piggyback on the general HMSS one.

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703)* status: 15 of 21 files reviewed, 4 unresolved discussions (waiting on @SanjayVas and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/common/ReportingApiServerFlags.kt` line 48 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703#-O3D_6lL0ILUQmGqDgaw:-ODNzDD44bHj64-azErx:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/591556264b768928feca93f1dbb9887a1604a1a6/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/common/ReportingApiServerFlags.kt#L48)):*

agreed.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt` line 218 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703#-OCFexcJA9MmuCjrMTcs:-ODNzEFoBYnwGndP0HE0:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/10540e4f3d5483e9ec75d9b7327a001ab158b4da/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt#L218)):*

LLv2 should not have wrapping.

Anyone who uses the HM Shuffle support libraries gets the wrapping capability by default. This should include TechEdge. I've confirmed Google implemented wrapping.

Seems like we don't need this capability and can just piggyback on the general HMSS one.
</blockquote></details>

Done.
___
*[`src/main/proto/wfa/measurement/internal/kingdom/data_provider_capabilities.proto` line 27 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703#-OCFfQr8Ct53WI7ffsms:-ODNzEkPDXoQyXZkwtj8:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/10540e4f3d5483e9ec75d9b7327a001ab158b4da/src/main/proto/wfa/measurement/internal/kingdom/data_provider_capabilities.proto#L27)):*

i say this is never supported.
</blockquote></details>

Done.
___
*[`src/main/proto/wfa/measurement/internal/kingdom/data_provider_capabilities.proto` line 30 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703#-OCFfTDQElr3aW3iI9R2:-ODNzFCJ3HGTZbajQysR:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/10540e4f3d5483e9ec75d9b7327a001ab158b4da/src/main/proto/wfa/measurement/internal/kingdom/data_provider_capabilities.proto#L30)):*

See above comment re: whether this is true
</blockquote></details>

Done.

stevenwarejones (approved):
> Reviewed 6 of 6 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1703)* status: all files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

PR #1729, opened 2024-07-31
fix: avoid voluntary eviction of Mill Job Pods

SanjayVas:
While the Mill should be tolerant of disruption, evicting Mill Jobs unnecessarily can result in delays in processing and duplicate work.

### DIFF
--- a/src/main/k8s/base.cue
+++ b/src/main/k8s/base.cue
@@ -486,6 +486,19 @@ objects: [ for objectSet in objectSets for object in objectSet {object}]
 	spec: #PodSpec
 }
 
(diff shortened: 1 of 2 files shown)


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1729)

SanjayVas:
It turns out that maxUnavailable cannot be used with Job Pods. See https://kubernetes.io/docs/tasks/run-application/configure-pdb/#arbitrary-controllers-and-selectors

PR #1731, opened 2024-08-01
fix: Throw StorageException as TransientError from the mill

renjiezh:
According to the practice of stress testing, given that[ common-jvm](https://github.com/world-federation-of-advertisers/common-jvm/blob/main/src/main/kotlin/org/wfanet/measurement/gcloud/gcs/GcsStorageClient.kt#L68-L70) already have retry mechanism for retryable exceptions, gcs writing still pops up unretryable exceptions which actually looks retryable. (Unexpected Error code 504 trying to get security access token from Compute Engine metadata for the default service account: Gateway Timeout)
This PR makes the error from storage transient so that the mill will retry it after the period of time.


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1731)

SanjayVas (commented):
I'm confused at this statement. If the StorageClient impl is handling retries, it should only be bubbling up non-retryable exceptions by definition.

A reminder that the philosophy is that errors are permanent and non-retryable by default, and we only retry at any level under the narrow circumstances where we can guarantee that it's safe to so do.

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1731#-:-O3EF-HDCkhwB0q0GLnC:bvpemvm)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @kungfucraig)

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1731)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @kungfucraig and @renjiezh)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/db/computation/ComputationDataClients.kt` line 137 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1731#-O3EFU6VAqhJByL0Sydu:-O3EEtGP48a5fdxyEc-d:b-qyv1u1) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/9ce4a0504ffa6b160bfe16c94c70aacc8d5b0e50/src/main/kotlin/org/wfanet/measurement/duchy/db/computation/ComputationDataClients.kt#L137)):*
> ```Kotlin
>         )
>       } catch (e: StorageException) {
>         throw TransientErrorException("Error writing blob to storage.", e)
> ```

This must inspect the StorageException to determine whether it's the equivalent of an ABORTED status, i.e. known to be safe to retry at a higher level. Only then can it be considered a transient error.

renjiezh (commented):
There are two reasons that I change the behavior for cloud storage:
1. Potential bugs in the storage client library. The storage client misclassifies the error as non-retryable so we follow it to fail the Computation. (I have filed the bug to the storage client dev team).
2. Temporary unavailability from cloud storage. When the cloud storage is unavailable for couple of minutes (I guess it is caused by the high volume of traffic), the common-jvm will use up attempts and fail the computation, However, we can rely on the mill's retry mechanism to do that in 5 minutes. Thus we can avoid to fail unnecessarily.

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1731#-:-O3EI8irB-6vuwOSyUWf:b92v5x1)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @kungfucraig)

kungfucraig (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1731)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/db/computation/ComputationDataClients.kt` line 137 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1731#-O3EFU6VAqhJByL0Sydu:-O3EZvqK2kfsdyX_8veq:b2hokwt) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/9ce4a0504ffa6b160bfe16c94c70aacc8d5b0e50/src/main/kotlin/org/wfanet/measurement/duchy/db/computation/ComputationDataClients.kt#L137)):*

This must inspect the StorageException to determine whether it's the equivalent of an ABORTED status, i.e. known to be safe to retry at a higher level. Only then can it be considered a transient error.
</blockquote></details>

I think we should handle this in the Mill and allow a retry of the stage even when there is a permanent error. Given that this is a storage client, it seems the worst thing that can happen is that we overwrite an otherwise successfully written blob.

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1731)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @kungfucraig and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/db/computation/ComputationDataClients.kt` line 137 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1731#-O3EFU6VAqhJByL0Sydu:-O3J7ULJ6KkXojl_n9iH:bs3mhdg) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/9ce4a0504ffa6b160bfe16c94c70aacc8d5b0e50/src/main/kotlin/org/wfanet/measurement/duchy/db/computation/ComputationDataClients.kt#L137)):*

I think we should handle this in the Mill and allow a retry of the stage even when there is a permanent error. Given that this is a storage client, it seems the worst thing that can happen is that we overwrite an otherwise successfully written blob.
</blockquote></details>

Comment added.

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1731)* status: all files reviewed, 1 unresolved discussion (waiting on @renjiezh)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/db/computation/ComputationDataClients.kt` line 143 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1731#-O3JKzpc-5ey4fCCV6V4:-O3JKzpdDzFHCUmDgGyt:bjh4fjg) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/acffd53874871f6b912a5e37aef86c0e2eea8bba/src/main/kotlin/org/wfanet/measurement/duchy/db/computation/ComputationDataClients.kt#L143)):*
> ```Kotlin
>         // A Google internal bug ticket has been created. Ticket id: 356682439
>         //
>         // TODO(@renjiez): Handle the exception case by case after the storage client fixes the
> ```

Since this project was initially a Google prototype we have a TODO format for Google's issue tracker. See https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/main/docs/code-style.md#format

In the future, please file all issues using external trackers so that non-Google contributors see the context.



_Suggestion:_
```Kotlin
TODO(b/356682439)
```

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1731)* status: all files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/db/computation/ComputationDataClients.kt` line 143 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1731#-O3JKzpc-5ey4fCCV6V4:-O3JYPcs6Bp3lTKJ0iSV:bp2hcda) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/acffd53874871f6b912a5e37aef86c0e2eea8bba/src/main/kotlin/org/wfanet/measurement/duchy/db/computation/ComputationDataClients.kt#L143)):*

Since this project was initially a Google prototype we have a TODO format for Google's issue tracker. See https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/main/docs/code-style.md#format

In the future, please file all issues using external trackers so that non-Google contributors see the context.
</blockquote></details>

Done.
Updated with a link to github issue. I noted there was a google internal issue created.

SanjayVas (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1731)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @renjiezh)

stevenwarejones (approved):
> Reviewed 2 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1731)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @renjiezh)

PR #1753, opened 2024-08-13
feat: Add new filter for checking externalComputationId, and checking externalComputationId if view is COMPUTATION or COMPUTATION_STATS

tristanvuong2021:
feat: Add new filter for checking externalComputationId, and checking externalComputationId if view is COMPUTATION or COMPUTATION_STATS

### DIFF
--- a/src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/queries/StreamMeasurements.kt
+++ b/src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/queries/StreamMeasurements.kt
@@ -62,7 +62,7 @@ class StreamMeasurements(
       return MeasurementReader(view).apply {
         this.orderByClause = orderByClause
         fillStatementBuilder {
-          appendWhereClause(requestFilter)
+          appendWhereClause(view, requestFilter)
           appendClause(orderByClause)
           if (limit > 0) {
             appendClause("LIMIT @$LIMIT_PARAM")
@@ -84,8 +84,13 @@ class StreamMeasurements(
       }
     }
 
-    private fun Statement.Builder.appendWhereClause(filter: StreamMeasurementsRequest.Filter) {
+    private fun Statement.Builder.appendWhereClause(      view: Measurement.View,
+                                                          filter: StreamMeasurementsRequest.Filter) {
       val conjuncts = mutableListOf<String>()
+
+      if (filter.hasExternalComputationId || view == Measurement.View.COMPUTATION || view == Measurement.View.COMPUTATION_STATS) {
+        conjuncts.add("ExternalComputationId IS NOT NULL")
+      }
 
       if (filter.externalMeasurementConsumerId != 0L) {
         conjuncts.add("ExternalMeasurementConsumerId = @$EXTERNAL_MEASUREMENT_CONSUMER_ID_PARAM")
--- a/src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/queries/StreamMeasurements.kt
+++ b/src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/queries/StreamMeasurements.kt
@@ -84,11 +84,17 @@ class StreamMeasurements(
       }
     }
 
-    private fun Statement.Builder.appendWhereClause(      view: Measurement.View,
-                                                          filter: StreamMeasurementsRequest.Filter) {
+    private fun Statement.Builder.appendWhereClause(
+      view: Measurement.View,
+      filter: StreamMeasurementsRequest.Filter,
+    ) {
       val conjuncts = mutableListOf<String>()
 
-      if (filter.hasExternalComputationId || view == Measurement.View.COMPUTATION || view == Measurement.View.COMPUTATION_STATS) {
+      if (
+        filter.hasExternalComputationId ||
+          view == Measurement.View.COMPUTATION ||
+          view == Measurement.View.COMPUTATION_STATS
+      ) {
         conjuncts.add("ExternalComputationId IS NOT NULL")
       }
 
--- a/src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/readers/MeasurementReader.kt
+++ b/src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/readers/MeasurementReader.kt
@@ -555,14 +555,6 @@ private fun MeasurementKt.Dsl.fillComputationView(struct: Struct) {
   val requisitionsStructs = struct.getStructList("Requisitions")
   val dataProvidersCount = requisitionsStructs.size
 
-  if (struct.isNull("ExternalComputationId")) {
-    for (requisitionStruct in requisitionsStructs) {
-      requisitions +=
-        RequisitionReader.buildRequisition(struct, requisitionStruct, mapOf(), dataProvidersCount)
-    }
-    return
-  }
-
   val externalMeasurementId = ExternalId(struct.getLong("ExternalMeasurementId"))
   val externalMeasurementConsumerId = ExternalId(struct.getLong("ExternalMeasurementConsumerId"))
   val externalComputationId = ExternalId(struct.getLong("ExternalComputationId"))
--- a/src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/readers/MeasurementReader.kt
+++ b/src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/readers/MeasurementReader.kt
@@ -592,10 +592,6 @@ private fun MeasurementKt.Dsl.fillComputationView(struct: Struct) {
 private fun MeasurementKt.Dsl.fillComputationStatsView(struct: Struct) {
   fillMeasurementCommon(struct)
 
-  if (struct.isNull("ExternalComputationId")) {
-    return
-  }
-
   val externalMeasurementId = ExternalId(struct.getLong("ExternalMeasurementId"))
   val externalMeasurementConsumerId = ExternalId(struct.getLong("ExternalMeasurementConsumerId"))
   val externalComputationId = ExternalId(struct.getLong("ExternalComputationId"))
--- a/src/main/kotlin/org/wfanet/measurement/kingdom/service/internal/testing/MeasurementsServiceTest.kt
+++ b/src/main/kotlin/org/wfanet/measurement/kingdom/service/internal/testing/MeasurementsServiceTest.kt
@@ -1608,7 +1608,7 @@ abstract class MeasurementsServiceTest<T : MeasurementsCoroutineImplBase> {
         )
 
       val streamMeasurementsRequest = streamMeasurementsRequest {
-        limit = 2
+        limit = 3
         filter = filter {
           externalMeasurementConsumerId = measurementConsumer.externalMeasurementConsumerId
         }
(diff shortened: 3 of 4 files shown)


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1753)

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1753)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/queries/StreamMeasurements.kt` line 91 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1753#-O4CWZaZ6OR9b3a8fi2c:-O4CWZaZ6OR9b3a8fi2d:b4pddvj) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/acc35e0bce9fd2976ffd8146989f3f39693e8337/src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/queries/StreamMeasurements.kt#L91)):*
> ```Kotlin
>       val conjuncts = mutableListOf<String>()
> 
>       if (view == Measurement.View.COMPUTATION) {
> ```

The view shouldn't filter, but rather should error. Filtering should *only* happen based on the `filter` field.

tristanvuong2021 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1753)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/queries/StreamMeasurements.kt` line 91 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1753#-O4CWZaZ6OR9b3a8fi2c:-O4Calzp4RyXe05RviRU:b5m6yb2) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/acc35e0bce9fd2976ffd8146989f3f39693e8337/src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/queries/StreamMeasurements.kt#L91)):*

The view shouldn't filter, but rather should error. Filtering should *only* happen based on the `filter` field.
</blockquote></details>

If I filter then that would imply that including direct measurements is possible.

tristanvuong2021 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1753)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/queries/StreamMeasurements.kt` line 91 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1753#-O4CWZaZ6OR9b3a8fi2c:-O4Cb4Jl9AAnhfAPtabE:b2l87s) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/acc35e0bce9fd2976ffd8146989f3f39693e8337/src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/queries/StreamMeasurements.kt#L91)):*

If I filter then that would imply that including direct measurements is possible.
</blockquote></details>

*If I filter based on a filter field*

tristanvuong2021 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1753)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/queries/StreamMeasurements.kt` line 91 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1753#-O4CWZaZ6OR9b3a8fi2c:-O4Cc9Lw-oah-zL1gQs_:b-9mrglw) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/acc35e0bce9fd2976ffd8146989f3f39693e8337/src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/queries/StreamMeasurements.kt#L91)):*

*If I filter based on a filter field*
</blockquote></details>

I can leave a comment mentioning that it will throw an error if the wrong value is set and the view is COMPUTATION.

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1753)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/queries/StreamMeasurements.kt` line 91 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1753#-O4CWZaZ6OR9b3a8fi2c:-O4CbEEI0cRtgqNczQ6C:b-hbqhnf) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/acc35e0bce9fd2976ffd8146989f3f39693e8337/src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/queries/StreamMeasurements.kt#L91)):*

I can leave a comment mentioning that it will throw an error if the wrong value is set and the view is COMPUTATION.
</blockquote></details>

I'm saying that using the COMPUTATION view when the filter does not cause the result to only contain Measurements having computations IDs should be an error. Our existing use cases avoid this as they always filter by computation participant (Duchy) ID.

What this PR should do is:
1. Add a way to filter based on the presence of computation ID (e.g. `having_computation_id`).
2. (Optional) error if we'd return any rows without computation IDs when the view is COMPUTATION, documenting this behavior.

tristanvuong2021 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1753)* status: 0 of 8 files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/queries/StreamMeasurements.kt` line 91 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1753#-O4CWZaZ6OR9b3a8fi2c:-O4CuSRi9vjXbb6x5L4J:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/acc35e0bce9fd2976ffd8146989f3f39693e8337/src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/queries/StreamMeasurements.kt#L91)):*

I'm saying that using the COMPUTATION view when the filter does not cause the result to only contain Measurements having computations IDs should be an error. Our existing use cases avoid this as they always filter by computation participant (Duchy) ID.

What this PR should do is:
1. Add a way to filter based on the presence of computation ID (e.g. `having_computation_id`).
2. (Optional) error if we'd return any rows without computation IDs when the view is COMPUTATION, documenting this behavior.
</blockquote></details>

Done.

SanjayVas (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1753)* status: all files reviewed, 2 unresolved discussions (waiting on @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/SpannerMeasurementsService.kt` line 148 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1753#-O4Curuh42gSkgyLSziH:-O4Curuh42gSkgyLSziI:b14lvpf) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7e3dff2970fb75172d4fd2ba499c238de61a774e/src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/SpannerMeasurementsService.kt#L148)):*
> ```Kotlin
>       .catch {
>         if (it is MeasurementNotComputationException) {
>           throw it.asStatusRuntimeException(Status.Code.INVALID_ARGUMENT, "Measurement that is not Computation found")
> ```

nit: this is not INVALID_ARGUMENT, as whether the error condition depends on the system state (i.e. whether there are any filtered rows without a computation ID). For example, the error is not thrown if the filter would allow direct Measurements, but there don't happen to be any.

If instead the error condition was based on the filters themselves (e.g. whether hasExtenrnalComputationId is true or whether the filter contains an external computation ID), then this would make sense as INVALID_ARGUMENT.
___
*[`src/main/proto/wfa/measurement/internal/kingdom/measurements_service.proto` line 44 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1753#-O4CwNjh4Zuv3K4AZFsh:-O4CwNjh4Zuv3K4AZFsi:b-w35uvh) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7e3dff2970fb75172d4fd2ba499c238de61a774e/src/main/proto/wfa/measurement/internal/kingdom/measurements_service.proto#L44)):*
> ```proto
>   //
>   // If the view is `Measurement.View.COMPUTATION` and the response contains
>   // Measurements that aren't Computations, then an error is returned.
> ```

nit: this should probably be about whether the filter ensures that no values without computation IDs would be returned (see above comment)



_Code quote:_
```proto
  // If the view is `Measurement.View.COMPUTATION` and the response contains
  // Measurements that aren't Computations, then an error is returned.
```

SanjayVas (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1753)* status: all files reviewed, 3 unresolved discussions (waiting on @tristanvuong2021)

___
*[`src/main/proto/wfa/measurement/internal/kingdom/measurements_service.proto` line 94 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1753#-O4Cwfc0-WrcRr5Pxbh_:-O4Cwfc0-WrcRr5Pxbha:bix81nh) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7e3dff2970fb75172d4fd2ba499c238de61a774e/src/main/proto/wfa/measurement/internal/kingdom/measurements_service.proto#L94)):*
> ```proto
>     repeated Measurement.State states = 3;
>     google.protobuf.Timestamp updated_after = 4;
>     string externalDuchyId = 7;
> ```

nit: maybe fix this while in here? It shouldn't really change the generated code



_Suggestion:_
```proto
external_duchy_id
```

stevenwarejones (changes requested):
Can you elaborate under what scenarios this exception would be expected to be thrown?

> Reviewed 2 of 8 files at r3, 2 of 3 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1753#-:-O4DkV4g5wphYH1H_3iy:b-7zjdd4)* status: 7 of 8 files reviewed, all discussions resolved (waiting on @SanjayVas)

stevenwarejones (approved):
> Reviewed 8 of 8 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1753)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @SanjayVas)

SanjayVas (approved):
> Reviewed 8 of 8 files at r5, 1 of 1 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1753)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @tristanvuong2021)

tristanvuong2021 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1753)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @tristanvuong2021)

PR #1754, opened 2024-08-13
fix: Add index for improving StreamMeasurements performance

tristanvuong2021:
fix: Add index for improving StreamMeasurements performance

### DIFF
--- a/src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/queries/StreamMeasurements.kt
+++ b/src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/queries/StreamMeasurements.kt
@@ -58,7 +58,13 @@ class StreamMeasurements(
       limit: Int,
     ): MeasurementReader {
       val orderByClause = getOrderByClause(view)
-      return MeasurementReader(view).apply {
+      val index =
+        if (view == Measurement.View.DEFAULT) {
+          MeasurementReader.Index.UPDATE_TIME
+        } else {
+          MeasurementReader.Index.NONE
+        }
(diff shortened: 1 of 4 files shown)


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1754)

SanjayVas (changes requested):
> Reviewed 1 of 4 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1754)* status: 1 of 4 files reviewed, 1 unresolved discussion (waiting on @tristanvuong2021)

___
*[`src/main/resources/kingdom/spanner/create-measurements-by-update-time-index.sql` line 20 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1754#-O4C_k-nBgnD25bGrgcv:-O4C_k-nBgnD25bGrgcw:bt6qr5b) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/55bce4c187349a1b87f21d41434f78216a2a8b80/src/main/resources/kingdom/spanner/create-measurements-by-update-time-index.sql#L20)):*
> ```TSQL
> -- comment: Create index for ordering Measurements by UpdateTime.
> 
> CREATE INDEX MeasurementsByUpdateTime ON Measurements(UpdateTime);
> ```

I tried out the StreamMeasurements query on Measurements in states (1, 2, 3, 4, 6, 7) updated after 2024-08-01 and no forced indexes. It uses the MeasurementsByContinuationToken index by default, which already has UpdateTime as its first column.

PR #1767, opened 2024-08-20
fix: Add missing pod selector in opentelemetry-collector NetworkPolicy

SanjayVas:
The NetworkPolicy was missing its pod selector, and was therefore applying to all pods.

### DIFF
--- a/src/main/k8s/dev/open_telemetry_eks.cue
+++ b/src/main/k8s/dev/open_telemetry_eks.cue
@@ -14,11 +14,7 @@
 
 package k8s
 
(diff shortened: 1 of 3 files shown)


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1767)

SanjayVas:
Combining this with https://github.com/world-federation-of-advertisers/cross-media-measurement/pull/1768 as the changes are tightly coupled.

PR #1768, opened 2024-08-20
fix!: Ensure default network policies are created

SanjayVas:
The previous behavior was to add some default ingresses and egresses to every NetworkPolicy other than default-deny. The flaw with this behavior is that if no NetworkPolicy matches the pod, the pod will not have these default ingresses and egresses. This instead creates separate NetworkPolicy objects for defaults.

This also fixes an issue where the opentelemetry-collector NetworkPolicy was missing its pod selector and was therefore applying to all pods.

### DIFF
--- a/docs/eks/duchy-deployment.md
+++ b/docs/eks/duchy-deployment.md
@@ -52,7 +52,8 @@ For a Duchy named `worker2`, the cluster will be populated with the following:
 -   CronJob
     -   `worker2-computations-cleaner-cronjob`
 -   NetworkPolicy
-    -   `default-deny-ingress-and-egress`
+    -   `default-deny-network-policy`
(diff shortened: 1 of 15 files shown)


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1768)

SanjayVas:
Depends on #1767

Marco-Premier (approved):
> Reviewed 15 of 15 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1768)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @SanjayVas)

PR #1796, opened 2024-09-09
feat: Implement prober's new measurement creation

roaminggypsy:
Implemented:
- (Before new measurement creation) check if a new measurement should be created by examining previous measurements
- create a new measurement

TODO (documented in the code): 
- use OpenTelemetry to report measurement result
- use QA event group for QA env


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1796)

roaminggypsy (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1796)* status: 0 of 2 files reviewed, 10 unresolved discussions (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/batch/MeasurementSystemProber.kt` line 30 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1796#-O6RudJJDGjuXuz1eZiu:-O6SjFVx3tw0LielgD7p:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/645a5db64718eae0395c78933aaba21dac84da9e/src/main/kotlin/org/wfanet/measurement/kingdom/batch/MeasurementSystemProber.kt#L30)):*

Wildcard imports are forbidden
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/batch/MeasurementSystemProber.kt` line 47 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1796#-O6RuTRT53V8MyEUGmuK:-O6SjprT3RKHgX-Dp9sY:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/645a5db64718eae0395c78933aaba21dac84da9e/src/main/kotlin/org/wfanet/measurement/kingdom/batch/MeasurementSystemProber.kt#L47)):*

Move these to companion object. See https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/main/docs/code-style.md#kotlin
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/batch/MeasurementSystemProber.kt` line 63 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1796#-O6Rwi6J7RcYDoMzGAca:-O6SkiGH1WXslH4iCdJ9:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/645a5db64718eae0395c78933aaba21dac84da9e/src/main/kotlin/org/wfanet/measurement/kingdom/batch/MeasurementSystemProber.kt#L63)):*

This isn't a service and therefore shouldn't be named as such. The codebase uses either "stub" or "client" for these.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/batch/MeasurementSystemProber.kt` line 67 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1796#-O6RxIyVCCHK6wVV-eji:-O6Xpu6o7xwyitNVWHet:btwkngk) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/645a5db64718eae0395c78933aaba21dac84da9e/src/main/kotlin/org/wfanet/measurement/kingdom/batch/MeasurementSystemProber.kt#L67)):*

The prober shouldn't have access to any internal services.
</blockquote></details>

I will add the ListMeasurements to enable filtering update time, and then update this.

(so I did not add exception throwing and handling here)
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/batch/MeasurementSystemProber.kt` line 72 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1796#-O6Rup136WsCAehgVjqp:-O6SlM0RDiuG9QJFgoID:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/645a5db64718eae0395c78933aaba21dac84da9e/src/main/kotlin/org/wfanet/measurement/kingdom/batch/MeasurementSystemProber.kt#L72)):*

This doesn't implement Runnable, so you can make this a suspend function rather than using runBlocking everywhere. The caller can wrap in runBlocking at the top-level.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/batch/MeasurementSystemProber.kt` line 75 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1796#-O6Rv63YEH-sl5GWm6nB:-O6XrAExCNDFtmDiHNuV:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/645a5db64718eae0395c78933aaba21dac84da9e/src/main/kotlin/org/wfanet/measurement/kingdom/batch/MeasurementSystemProber.kt#L75)):*

This won't work once the Measurement validation and metric recording are implemented. You may want to extract everything below into another function for creating the Measurement and wrap it in the shouldCreate check.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/batch/MeasurementSystemProber.kt` line 79 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1796#-O6RvYtX-k8YWrkrdCRl:-O6TIM2V7wT1Amtyqjux:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/645a5db64718eae0395c78933aaba21dac84da9e/src/main/kotlin/org/wfanet/measurement/kingdom/batch/MeasurementSystemProber.kt#L79)):*

Extract a function for building this map.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/batch/MeasurementSystemProber.kt` line 82 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1796#-O6Rw0Q50UyESVH-2TKZ:-O6TCsCpBVgRkvxtUElz:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/645a5db64718eae0395c78933aaba21dac84da9e/src/main/kotlin/org/wfanet/measurement/kingdom/batch/MeasurementSystemProber.kt#L82)):*

Specify types explicitly where not obvious. See https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/main/docs/code-style.md#kotlin
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/batch/MeasurementSystemProber.kt` line 96 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1796#-O6RwJof8ttUrgTmhRrZ:-O6TDJCO-L1HGYI7DYDt:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/645a5db64718eae0395c78933aaba21dac84da9e/src/main/kotlin/org/wfanet/measurement/kingdom/batch/MeasurementSystemProber.kt#L96)):*

This is not a gRPC service implementation and therefore should not throw StatusRuntimeException. Throw another exception type. If you won't be catching it at a higher level, you can just use `Exception`.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/batch/MeasurementSystemProber.kt` line 112 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1796#-O6RwZer4Ymq_qlbZGSV:-O6TIL4T8zZ3kkrb9h3o:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/645a5db64718eae0395c78933aaba21dac84da9e/src/main/kotlin/org/wfanet/measurement/kingdom/batch/MeasurementSystemProber.kt#L112)):*

All gRPC stub calls should catch StatusException at the lowest level
</blockquote></details>

Done.

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1796)* status: all files reviewed, 4 unresolved discussions (waiting on @roaminggypsy)

___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/batch/MeasurementSystemProber.kt` line 30 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1796#-O6RudJJDGjuXuz1eZiu:-O6gJg63DqXgyyoGNtZk:b-ii9oyz) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/645a5db64718eae0395c78933aaba21dac84da9e/src/main/kotlin/org/wfanet/measurement/kingdom/batch/MeasurementSystemProber.kt#L30)):*

Done.
</blockquote></details>

I still see a wildcard import: `import org.wfanet.measurement.api.v2alpha.*`
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/batch/MeasurementSystemProber.kt` line 67 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1796#-O6RxIyVCCHK6wVV-eji:-O6gJmIwC2dZr4o-dNg0:b-ylp3ms) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/645a5db64718eae0395c78933aaba21dac84da9e/src/main/kotlin/org/wfanet/measurement/kingdom/batch/MeasurementSystemProber.kt#L67)):*
> I will add the ListMeasurements to enable filtering update time, and then update this.

To clarify, this PR cannot be merged until that is done.
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/batch/MeasurementSystemProber.kt` line 96 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1796#-O6RwJof8ttUrgTmhRrZ:-O6gK7pK3AzUZc1gE5LU:b-uqvkrv) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/645a5db64718eae0395c78933aaba21dac84da9e/src/main/kotlin/org/wfanet/measurement/kingdom/batch/MeasurementSystemProber.kt#L96)):*

Done.
</blockquote></details>

This is now throwing a gRPC `StatusException`, which is the same problem. What you want is something like 

```kotlin
catch (e: StatusException) {
  throw Exception("Unable to get measurement consumer $measurementConsumerName", e)
}
```

Though you'll need to inspect `e.status.code` if you need to do something different depending on the status code.
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/deploy/common/job/MeasurementSystemProberJob.kt` line 118 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1796#-O6gKyEC8CiNBpxjdDCP:-O6gKyEDApR9P1mhJEet:bwslcb2) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d68fbe0baa776ecd928173f25a28adca9822fcb7/src/main/kotlin/org/wfanet/measurement/kingdom/deploy/common/job/MeasurementSystemProberJob.kt#L118)):*
> ```Kotlin
>   showDefaultValues = true,
> )
> private suspend fun run(@Mixin flags: MeasurementSystemProberFlags) {
> ```

This one can't be suspending since it's a `@Command`.

roaminggypsy (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1796)* status: 0 of 2 files reviewed, 4 unresolved discussions (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/batch/MeasurementSystemProber.kt` line 30 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1796#-O6RudJJDGjuXuz1eZiu:-O6hSGSq9uVd9SQ5oG6C:b-8rryk4) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/645a5db64718eae0395c78933aaba21dac84da9e/src/main/kotlin/org/wfanet/measurement/kingdom/batch/MeasurementSystemProber.kt#L30)):*

I still see a wildcard import: `import org.wfanet.measurement.api.v2alpha.*`
</blockquote></details>

ugh, Intellij changed it back for me. Had to uncheck the wildcard import in the language setting.
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/batch/MeasurementSystemProber.kt` line 96 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1796#-O6RwJof8ttUrgTmhRrZ:-O6hQ12QCgPZqS6Cquql:bze845u) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/645a5db64718eae0395c78933aaba21dac84da9e/src/main/kotlin/org/wfanet/measurement/kingdom/batch/MeasurementSystemProber.kt#L96)):*

This is now throwing a gRPC `StatusException`, which is the same problem. What you want is something like 

```kotlin
catch (e: StatusException) {
  throw Exception("Unable to get measurement consumer $measurementConsumerName", e)
}
```

Though you'll need to inspect `e.status.code` if you need to do something different depending on the status code.
</blockquote></details>

ah, got it! I don't think I want to do something different for now.
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/deploy/common/job/MeasurementSystemProberJob.kt` line 118 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1796#-O6gKyEC8CiNBpxjdDCP:-O6hKIM06iG92mBYC6-C:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d68fbe0baa776ecd928173f25a28adca9822fcb7/src/main/kotlin/org/wfanet/measurement/kingdom/deploy/common/job/MeasurementSystemProberJob.kt#L118)):*

This one can't be suspending since it's a `@Command`.
</blockquote></details>

Done.

roaminggypsy (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1796)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/batch/MeasurementSystemProber.kt` line 67 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1796#-O6RxIyVCCHK6wVV-eji:-O7Bfhv50S_NV5_OVQgX:bs7wumb) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/645a5db64718eae0395c78933aaba21dac84da9e/src/main/kotlin/org/wfanet/measurement/kingdom/batch/MeasurementSystemProber.kt#L67)):*

> I will add the ListMeasurements to enable filtering update time, and then update this.

To clarify, this PR cannot be merged until that is done.
</blockquote></details>

Done.

SanjayVas (changes requested):
> Reviewed 2 of 2 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1796)* status: all files reviewed, 1 unresolved discussion (waiting on @roaminggypsy)

___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/batch/MeasurementSystemProber.kt` line 230 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1796#-O7Bhs66Dqr8lKJTgarc:-O7Bhs66Dqr8lKJTgard:b-gvo8u2) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7ebc698413ab1545ad161d92c682a2c885fad4b3/src/main/kotlin/org/wfanet/measurement/kingdom/batch/MeasurementSystemProber.kt#L230)):*
> ```Kotlin
>     }
> 
>     val finishedMeasurements: List<Measurement> =
> ```

You need to handle pagination here. ListMeasurements could return fewer results than the specified page size, so you need to look at the page token and potentially make more requests to build up the whole list.

roaminggypsy (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1796)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/batch/MeasurementSystemProber.kt` line 230 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1796#-O7Bhs66Dqr8lKJTgarc:-O7BtWm5B0DX1280zX-X:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7ebc698413ab1545ad161d92c682a2c885fad4b3/src/main/kotlin/org/wfanet/measurement/kingdom/batch/MeasurementSystemProber.kt#L230)):*

You need to handle pagination here. ListMeasurements could return fewer results than the specified page size, so you need to look at the page token and potentially make more requests to build up the whole list.
</blockquote></details>

Done.

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1796)* status: 1 of 2 files reviewed, 7 unresolved discussions (waiting on @roaminggypsy)

___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/batch/MeasurementSystemProber.kt` line 66 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1796#-O7Ur5Gn7EkSz41tUjd6:-O7Ur5Go4VEjcVH8MYOc:b-3njfn4) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/af88991660c1b6647c7278793db6d8f03c1f2d03/src/main/kotlin/org/wfanet/measurement/kingdom/batch/MeasurementSystemProber.kt#L66)):*
> ```Kotlin
> import org.wfanet.measurement.consent.client.measurementconsumer.signRequisitionSpec
> 
> private const val MAX_PAGE_SIZE = 1000
> ```

nit: this isn't a maximum here, but rather the actual page size you're requesting for Measurements



_Suggestion:_
```Kotlin
LIST_MEASUREMENTS_PAGE_SIZE
```
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/batch/MeasurementSystemProber.kt` line 66 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1796#-O7UrEJ3AqZ73wXUpBE7:-O7UrEJ3AqZ73wXUpBE8:b-trf6r5) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/af88991660c1b6647c7278793db6d8f03c1f2d03/src/main/kotlin/org/wfanet/measurement/kingdom/batch/MeasurementSystemProber.kt#L66)):*
> ```Kotlin
> import org.wfanet.measurement.consent.client.measurementconsumer.signRequisitionSpec
> 
> private const val MAX_PAGE_SIZE = 1000
> ```

Move this to the companion object
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/batch/MeasurementSystemProber.kt` line 227 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1796#-O7UsYXJ6bsS6aaR3QuG:-O7Us7DY2ZMKq0o083g6:b-bgxz5j) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/af88991660c1b6647c7278793db6d8f03c1f2d03/src/main/kotlin/org/wfanet/measurement/kingdom/batch/MeasurementSystemProber.kt#L227)):*
> ```Kotlin
>    */
>   private suspend fun shouldCreateNewMeasurement(): Boolean {
>     val previousMeasurements: List<Measurement> =
> ```

I'm not sure what "previous" means here. It just looks like unfiltered Measurements from all time. Shouldn't there be a timestamp filter?
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/batch/MeasurementSystemProber.kt` line 228 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1796#-O7UsCU8AD9KAT0MEpjP:-O7UsCU8AD9KAT0MEpjQ:b-ei7qc7) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/af88991660c1b6647c7278793db6d8f03c1f2d03/src/main/kotlin/org/wfanet/measurement/kingdom/batch/MeasurementSystemProber.kt#L228)):*
> ```Kotlin
>   private suspend fun shouldCreateNewMeasurement(): Boolean {
>     val previousMeasurements: List<Measurement> =
>       measurementsStub.listMeasurements(listMeasurementsRequest {}).measurementsList
> ```

This also needs to call your function which handles pagination
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/batch/MeasurementSystemProber.kt` line 263 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1796#-O7UsK4IBvN2bfk2aPGx:-O7UsK4IBvN2bfk2aPGy:bu021wg) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/af88991660c1b6647c7278793db6d8f03c1f2d03/src/main/kotlin/org/wfanet/measurement/kingdom/batch/MeasurementSystemProber.kt#L263)):*
> ```Kotlin
>     while (isFirstRequest || nextPageToken.isNotBlank()) {
>       val response: ListMeasurementsResponse =
>         measurementsStub.listMeasurements(
> ```

All gRPC stub calls must catch StatusException
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/batch/MeasurementSystemProber.kt` line 265 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1796#-O7UsMXG7g_vWchmLhqi:-O7UsMXG7g_vWchmLhqj:b-ijpg2e) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/af88991660c1b6647c7278793db6d8f03c1f2d03/src/main/kotlin/org/wfanet/measurement/kingdom/batch/MeasurementSystemProber.kt#L265)):*
> ```Kotlin
>         measurementsStub.listMeasurements(
>           listMeasurementsRequest {
>             this.filter = filter
> ```

`parent` field is required

roaminggypsy (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1796)* status: 1 of 2 files reviewed, 6 unresolved discussions (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/batch/MeasurementSystemProber.kt` line 66 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1796#-O7UrEJ3AqZ73wXUpBE7:-O7VlA3GDITPousvjLkn:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/af88991660c1b6647c7278793db6d8f03c1f2d03/src/main/kotlin/org/wfanet/measurement/kingdom/batch/MeasurementSystemProber.kt#L66)):*

Move this to the companion object
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/batch/MeasurementSystemProber.kt` line 227 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1796#-O7UsYXJ6bsS6aaR3QuG:-O7VlHGE07QY9-PqHdKC:b6q2j3p) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/af88991660c1b6647c7278793db6d8f03c1f2d03/src/main/kotlin/org/wfanet/measurement/kingdom/batch/MeasurementSystemProber.kt#L227)):*

I'm not sure what "previous" means here. It just looks like unfiltered Measurements from all time. Shouldn't there be a timestamp filter?
</blockquote></details>

It means 'previously (created)' measurements. I
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/batch/MeasurementSystemProber.kt` line 228 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1796#-O7UsCU8AD9KAT0MEpjP:-O7Vm1jS33Qpv-GiAXNv:b85tune) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/af88991660c1b6647c7278793db6d8f03c1f2d03/src/main/kotlin/org/wfanet/measurement/kingdom/batch/MeasurementSystemProber.kt#L228)):*

This also needs to call your function which handles pagination
</blockquote></details>

Switched to the wrapper method to avoid duplicate pagination handling logic
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/batch/MeasurementSystemProber.kt` line 263 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1796#-O7UsK4IBvN2bfk2aPGx:-O7Vo_DbEHerKzA-At1P:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/af88991660c1b6647c7278793db6d8f03c1f2d03/src/main/kotlin/org/wfanet/measurement/kingdom/batch/MeasurementSystemProber.kt#L263)):*

All gRPC stub calls must catch StatusException
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/batch/MeasurementSystemProber.kt` line 265 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1796#-O7UsMXG7g_vWchmLhqi:-O7Vn8lz7-MV3hYGmneN:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/af88991660c1b6647c7278793db6d8f03c1f2d03/src/main/kotlin/org/wfanet/measurement/kingdom/batch/MeasurementSystemProber.kt#L265)):*

`parent` field is required
</blockquote></details>

Done.

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1796)* status: all files reviewed, 2 unresolved discussions (waiting on @roaminggypsy)

___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/batch/MeasurementSystemProber.kt` line 227 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1796#-O7UsYXJ6bsS6aaR3QuG:-O7VqeSlCqjZnUsj4VBm:b5mo6bj) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/af88991660c1b6647c7278793db6d8f03c1f2d03/src/main/kotlin/org/wfanet/measurement/kingdom/batch/MeasurementSystemProber.kt#L227)):*

It means 'previously (created)' measurements. I
</blockquote></details>

Getting all measurements with no filter is a bad idea, as the number of measurements will grow unbounded. What you instead likely want to do is see if there are any non-completed measurements created since `now - durationBetweenMeasurements`
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/batch/MeasurementSystemProber.kt` line 241 at r9](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1796#-O7VrSMqBIQ10VI-2avm:-O7VrSMqBIQ10VI-2avn:b-x0048j) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/30273c294cf71aaadb36cd140119ff5d615d2cb2/src/main/kotlin/org/wfanet/measurement/kingdom/batch/MeasurementSystemProber.kt#L241)):*
> ```Kotlin
> 
>     val oldFinishedMeasurements =
>       listAllMeasurements(
> ```

This is a strict subset of `finishedMeasurements` so you don't need to make another RPC. You can just filter the results locally.

roaminggypsy (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1796)* status: 1 of 9 files reviewed, 2 unresolved discussions (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/batch/MeasurementSystemProber.kt` line 227 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1796#-O7UsYXJ6bsS6aaR3QuG:-O7VzxCbEI2Q2b_aHqff:b-ht1zp) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/af88991660c1b6647c7278793db6d8f03c1f2d03/src/main/kotlin/org/wfanet/measurement/kingdom/batch/MeasurementSystemProber.kt#L227)):*

Getting all measurements with no filter is a bad idea, as the number of measurements will grow unbounded. What you instead likely want to do is see if there are any non-completed measurements created since `now - durationBetweenMeasurements`
</blockquote></details>

Rewrote the whole function.
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/batch/MeasurementSystemProber.kt` line 241 at r9](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1796#-O7VrSMqBIQ10VI-2avm:-O7W-qHF3GQShNI6pnHI:bkiv4w4) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/30273c294cf71aaadb36cd140119ff5d615d2cb2/src/main/kotlin/org/wfanet/measurement/kingdom/batch/MeasurementSystemProber.kt#L241)):*

This is a strict subset of `finishedMeasurements` so you don't need to make another RPC. You can just filter the results locally.
</blockquote></details>

Not applicable for the new implementation? (Since I'm only getting 1 result from the request)

SanjayVas (changes requested):
> Reviewed 3 of 8 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1796)* status: 4 of 9 files reviewed, 4 unresolved discussions (waiting on @roaminggypsy)

___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/batch/MeasurementSystemProber.kt` line 217 at r10](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1796#-O7ebNAr9bJLgjtwuZHX:-O7ebNAr9bJLgjtwuZHY:bfz5t4y) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d1e0c8da69b5794295ddfd82f98d6f56768f9089/src/main/kotlin/org/wfanet/measurement/kingdom/batch/MeasurementSystemProber.kt#L217)):*
You technically still need to check if `next_page_token` is specified in the event that you get no results. From [AIP-158](https://google.aip.dev/158):

> The API **may** return fewer results than the number requested (including zero results), even if not at the end of the collection.

The current implementation of this specific method doesn't do this, but callers should not rely on undocumented implementation-specific behavior.
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/batch/MeasurementSystemProber.kt` line 241 at r10](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1796#-O7ecX4dDs9q0Xtf_A6I:-O7ecX4dDs9q0Xtf_A6J:b-t6s1rv) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d1e0c8da69b5794295ddfd82f98d6f56768f9089/src/main/kotlin/org/wfanet/measurement/kingdom/batch/MeasurementSystemProber.kt#L241)):*
> ```Kotlin
>     }
> 
>     val lastFinishedMeasurementWithinWindow: Measurement? =
> ```

You already have the last completed measurement. You can check locally whether it's within the window without having to make another RPC.
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/queries/StreamMeasurements.kt` line 134 at r10](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1796#-O7ecfaCEvQZLTHPKHfD:-O7ecfaCEvQZLTHPKHfE:bs2aijs) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d1e0c8da69b5794295ddfd82f98d6f56768f9089/src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/queries/StreamMeasurements.kt#L134)):*
> ```Kotlin
>       }
> 
>       if (filter.hasCreatedAfter()) {
> ```

These changes don't belong in this PR. Have you synced the PR branch?

roaminggypsy (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1796)* status: 1 of 9 files reviewed, 4 unresolved discussions (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/batch/MeasurementSystemProber.kt` line 241 at r9](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1796#-O7VrSMqBIQ10VI-2avm:-O7mFOy596gzLigMjUdZ:bb6tyiv) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/30273c294cf71aaadb36cd140119ff5d615d2cb2/src/main/kotlin/org/wfanet/measurement/kingdom/batch/MeasurementSystemProber.kt#L241)):*

Not applicable for the new implementation? (Since I'm only getting 1 result from the request)
</blockquote></details>

DONE
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/batch/MeasurementSystemProber.kt` line 217 at r10](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1796#-O7ebNAr9bJLgjtwuZHX:-O7mFMtt0dzhumksko8p:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d1e0c8da69b5794295ddfd82f98d6f56768f9089/src/main/kotlin/org/wfanet/measurement/kingdom/batch/MeasurementSystemProber.kt#L217)):*

You technically still need to check if `next_page_token` is specified in the event that you get no results. From [AIP-158](https://google.aip.dev/158):

> The API **may** return fewer results than the number requested (including zero results), even if not at the end of the collection.

The current implementation of this specific method doesn't do this, but callers should not rely on undocumented implementation-specific behavior.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/batch/MeasurementSystemProber.kt` line 241 at r10](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1796#-O7ecX4dDs9q0Xtf_A6I:-O7mFHfFDaGPQAxUDZ-n:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d1e0c8da69b5794295ddfd82f98d6f56768f9089/src/main/kotlin/org/wfanet/measurement/kingdom/batch/MeasurementSystemProber.kt#L241)):*

You already have the last completed measurement. You can check locally whether it's within the window without having to make another RPC.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/queries/StreamMeasurements.kt` line 134 at r10](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1796#-O7ecfaCEvQZLTHPKHfD:-O7mFFNm8glFp0HYj0cK:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d1e0c8da69b5794295ddfd82f98d6f56768f9089/src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/queries/StreamMeasurements.kt#L134)):*

These changes don't belong in this PR. Have you synced the PR branch?
</blockquote></details>

Done.

SanjayVas (changes requested):
> Reviewed 8 of 8 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1796)* status: all files reviewed, 2 unresolved discussions (waiting on @roaminggypsy)

___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/batch/MeasurementSystemProber.kt` line 215 at r11](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1796#-O7oTlbN0k7PwHP28pGL:-O7oTlbN0k7PwHP28pGM:b-s5qr5e) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f36113c6bf6567c7ba703f55395e35226a742c20/src/main/kotlin/org/wfanet/measurement/kingdom/batch/MeasurementSystemProber.kt#L215)):*
> ```Kotlin
> 
>   private suspend fun shouldCreateNewMeasurement(): Boolean {
>     var lastMeasurement: Measurement? = null
> ```

nit: extract this out to a `getLastUpdatedMeasurement` function
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/batch/MeasurementSystemProber.kt` line 249 at r11](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1796#-O7oU6_o0SDcytIRa1LS:-O7oU6_o0SDcytIRa1LT:bl0j2dk) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f36113c6bf6567c7ba703f55395e35226a742c20/src/main/kotlin/org/wfanet/measurement/kingdom/batch/MeasurementSystemProber.kt#L249)):*
> ```Kotlin
>         lastMeasurement.updateTime.seconds,
>         lastMeasurement.updateTime.nanos.toLong(),
>       )
> ```

Use the `toInstant` extension from `org.wfanet.measurement.common`



_Suggestion:_
```Kotlin
      lastMeasurement.updateTime.toInstant()
```

roaminggypsy (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1796)* status: 8 of 9 files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/batch/MeasurementSystemProber.kt` line 215 at r11](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1796#-O7oTlbN0k7PwHP28pGL:-O7odbO41m03JCng_8gr:b-bm6yv9) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f36113c6bf6567c7ba703f55395e35226a742c20/src/main/kotlin/org/wfanet/measurement/kingdom/batch/MeasurementSystemProber.kt#L215)):*

nit: extract this out to a `getLastUpdatedMeasurement` function
</blockquote></details>

I named it getLastCreatedMeasurement. For prober measurements, last updated should always be last created because 1)  new measurement is only created when previous measurement is finished or there are no measurements. 2) (I assume) Finished measurements' update time won't be updated again 3) When measurement is created, both [creation time and update time is set to the same value](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/main/src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/writers/CreateMeasurements.kt#L333-L334)
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/batch/MeasurementSystemProber.kt` line 249 at r11](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1796#-O7oU6_o0SDcytIRa1LS:-O7ob14M0v-yrzBw-Xee:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f36113c6bf6567c7ba703f55395e35226a742c20/src/main/kotlin/org/wfanet/measurement/kingdom/batch/MeasurementSystemProber.kt#L249)):*

Use the `toInstant` extension from `org.wfanet.measurement.common`
</blockquote></details>

Done.

SanjayVas (changes requested):
> Reviewed 1 of 1 files at r12.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1796)* status: all files reviewed, 1 unresolved discussion (waiting on @roaminggypsy)

___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/batch/MeasurementSystemProber.kt` line 215 at r11](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1796#-O7oTlbN0k7PwHP28pGL:-O7pJLKa8tAbNECvTNQt:bgbwm0l) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f36113c6bf6567c7ba703f55395e35226a742c20/src/main/kotlin/org/wfanet/measurement/kingdom/batch/MeasurementSystemProber.kt#L215)):*

I named it getLastCreatedMeasurement. For prober measurements, last updated should always be last created because 1)  new measurement is only created when previous measurement is finished or there are no measurements. 2) (I assume) Finished measurements' update time won't be updated again 3) When measurement is created, both [creation time and update time is set to the same value](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/main/src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/writers/CreateMeasurements.kt#L333-L334)
</blockquote></details>

I think the function name should still be lastUpdated, as that's technically what's being returned from the API. Using that name will be more clear, especially since it's depending on the fact that ListMeasurements is ordered by update time.

The fact that it *should* also be the last created is specific to current prober behavior, which may change.
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/batch/MeasurementSystemProber.kt` line 243 at r12](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1796#-O7pK4xSBCbOCLi1IaK0:-O7pK4xSBCbOCLi1IaK1:b-zhdl2h) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f47813be03fa413f149a1b1bce3dc87af34c1658/src/main/kotlin/org/wfanet/measurement/kingdom/batch/MeasurementSystemProber.kt#L243)):*
> ```Kotlin
>             }
>           )
>         lastCreatedMeasurement = response.measurementsList[0]
> ```

You could just make the while condition `nextPageToken.isNotEmpty()` and return as soon as `response.measurementsList.isNotEmpty()`. That eliminates the need for an extra var.

e.g.

```kotlin
var nextPageToken = ""
while (nextPageToken.isNotEmpty()) {
  val response: ListMeasurementsResponse = try {
     // call stub
    // throw
  }
  if (response.measurementsList.isNotEmpty()) {
    return response.measurementsList.single()
  }
  nextPageToken = response.nextPageToken
}
```

roaminggypsy (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1796)* status: 8 of 9 files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/batch/MeasurementSystemProber.kt` line 215 at r11](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1796#-O7oTlbN0k7PwHP28pGL:-O7pgOq11YA32Tih3dck:b-woqx8e) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f36113c6bf6567c7ba703f55395e35226a742c20/src/main/kotlin/org/wfanet/measurement/kingdom/batch/MeasurementSystemProber.kt#L215)):*

I think the function name should still be lastUpdated, as that's technically what's being returned from the API. Using that name will be more clear, especially since it's depending on the fact that ListMeasurements is ordered by update time.

The fact that it *should* also be the last created is specific to current prober behavior, which may change.
</blockquote></details>

agree! Done
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/batch/MeasurementSystemProber.kt` line 243 at r12](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1796#-O7pK4xSBCbOCLi1IaK0:-O7pgSOL-Xi6Uva-uvwB:bkc5q22) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f47813be03fa413f149a1b1bce3dc87af34c1658/src/main/kotlin/org/wfanet/measurement/kingdom/batch/MeasurementSystemProber.kt#L243)):*

You could just make the while condition `nextPageToken.isNotEmpty()` and return as soon as `response.measurementsList.isNotEmpty()`. That eliminates the need for an extra var.

e.g.

```kotlin
var nextPageToken = ""
while (nextPageToken.isNotEmpty()) {
  val response: ListMeasurementsResponse = try {
     // call stub
    // throw
  }
  if (response.measurementsList.isNotEmpty()) {
    return response.measurementsList.single()
  }
  nextPageToken = response.nextPageToken
}
```
</blockquote></details>

I think isNotEmpty() will return false for empty string "". 
I changed it to checking null. It seems like [nextPageToken is only set to a value when the results size is larger than the page size](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/main/src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/MeasurementsService.kt#L244). If not set, it should have the default value null.

roaminggypsy (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1796)* status: 8 of 9 files reviewed, 1 unresolved discussion (waiting on @SanjayVas and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/batch/MeasurementSystemProber.kt` line 243 at r12](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1796#-O7pK4xSBCbOCLi1IaK0:-O83UJq568Yjp0GX5sMA:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f47813be03fa413f149a1b1bce3dc87af34c1658/src/main/kotlin/org/wfanet/measurement/kingdom/batch/MeasurementSystemProber.kt#L243)):*

> I think isNotEmpty() will return false for empty string "".

isEmpty() returns true for empty string, so isNotEmpty() returns the opposite

> If not set, it should have the default value null.

The protobuf default value for a string field is empty string. Protobuf field accessors in Java/Kotlin **never** return null.
</blockquote></details>

Done.

SanjayVas (approved):
> Reviewed 3 of 3 files at r14, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1796)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @stevenwarejones)

stevenwarejones (approved):
> Reviewed 1 of 2 files at r6, 5 of 8 files at r11, 3 of 3 files at r14.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1796)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @roaminggypsy)

PR #1801, opened 2024-09-11
Update Duchy control service to tolerate certain stage mismathes.

renjiezh:


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1801)

SanjayVas (approved):
> Reviewed 8 of 8 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1801)* status: all files reviewed, 1 unresolved discussion (waiting on @renjiezh)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/service/internal/computations/Protos.kt` line 80 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1801#-O6gMLTOBZJioKy7MxCP:-O6gMLTOBZJioKy7MxCQ:b3lw808) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/856db69a247efa637c5a461c36c53de2d9ee11d4/src/main/kotlin/org/wfanet/measurement/duchy/service/internal/computations/Protos.kt#L80)):*
> ```Kotlin
>   this.blobsList.filter { it.dependencyType == INPUT }.map { it.path }
> 
> /** Extract */
> ```

nit: unfinished KDoc

stevenwarejones (commented):
> Reviewed 1 of 8 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1801)* status: all files reviewed, 1 unresolved discussion (waiting on @renjiezh)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/service/internal/computationcontrol/AsyncComputationControlService.kt` line 106 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1801#-O6wRBby4yKKVzOUcOOb:-O6wRBby4yKKVzOUcOOc:b-74yqcq) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/2135fa49c3be9cbcfc55014520601dc1e5324390/src/main/kotlin/org/wfanet/measurement/duchy/service/internal/computationcontrol/AsyncComputationControlService.kt#L106)):*
> ```Kotlin
>         return AdvanceComputationResponse.getDefaultInstance()
>       }
>       // Catch up if stage in the request is one step ahead.
> ```

why does this happen? why do these stage mismatches occur?

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1801)* status: all files reviewed, 1 unresolved discussion (waiting on @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/service/internal/computationcontrol/AsyncComputationControlService.kt` line 106 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1801#-O6wRBby4yKKVzOUcOOb:-O7-z07fAIGmYO6XEm__:b-dn9jye) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/2135fa49c3be9cbcfc55014520601dc1e5324390/src/main/kotlin/org/wfanet/measurement/duchy/service/internal/computationcontrol/AsyncComputationControlService.kt#L106)):*

why does this happen? why do these stage mismatches occur?
</blockquote></details>

The root cause is that we cannot update both remote state and own state at the same time and guarantee the consistency. 

The worker has two steps to update the computation stage. They are:
1. update the peer's stage by the RPC call;
2. update its own stage by calling the internal server.
These two steps are done in sequence and it is possible to be interrupted between (1) and (2). 
If the interruption happens, the peer worker's stage has been updated and it continue to next stage. On the other hand, the worker itself will rerun the old stage thus it fell behind the peer worker.
In this scenario,  a control service could receive stale stage or future stage. It should try to tolerate it instead of failing the computation.

stevenwarejones (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1801)* status: all files reviewed, 1 unresolved discussion (waiting on @renjiezh)

SanjayVas:
This introduces a new compiler warning:

```
src/main/kotlin/org/wfanet/measurement/duchy/service/internal/computations/Protos.kt:82:16: warning: enum argument can be null in Java, but exhaustive when contains no null branch
  return when (computationDetails.protocolCase) {
```

PR #1802, opened 2024-09-12
feat: Implement CreateIncident method in Internal Incidents Service

tristanvuong2021:
feat: Implement CreateIncident method in Internal Incidents Service


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1802)

PR #1834, opened 2024-09-30
build: Use bazel-remote for Bazel remote caching

SanjayVas:
This should be more efficient as it uses gRPC over HTTP/2. It has the additional benefit that it does not depend on Google Cloud credentials, so it can be used in cases where different credentials are already in use (e.g. applying K8s configs).


SanjayVas:
Closing in favor of #1840

PR #1835, opened 2024-09-30
feat: Update logging message for participant failure and computation claiming.

renjiezh:


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1835)

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1835)* status: all files reviewed, 4 unresolved discussions (waiting on @renjiezh and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/daemon/mill/MillJobScheduler.kt` line 141 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1835#-O88DsoP7GmTUI7sOVGa:-O88DsoP7GmTUI7sOVGb:bkz26r3) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7280adbd9dd7e08d73feb0812319df9e2712038e/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/daemon/mill/MillJobScheduler.kt#L141)):*
> ```Kotlin
> 
>     val claimedComputationId: String = claimedToken.globalComputationId
>     logger.info { "MillScheduler claimed work item for Computation $claimedComputationId" }
> ```

nit: Is this necessary? The log source will be the scheduler, so it should be obvious that it's the scheduler that claimed the work.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/mill/MillBase.kt` line 232 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1835#-O88DdUiE9YxsC4LAyoV:-O88DdUj-y20mFgQOvR3:beh9cb5) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7280adbd9dd7e08d73feb0812319df9e2712038e/src/main/kotlin/org/wfanet/measurement/duchy/mill/MillBase.kt#L232)):*
> ```Kotlin
>       val globalComputationId = token.globalComputationId
>       val stage = token.computationStage
>       "@Mill $millId: Claimed work item for Computation $globalComputationId at stage $stage"
> ```

nit: Is this necessary? The mill ID will correspond to the job name, which is known as the log source.
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/writers/FailComputationParticipant.kt` line 123 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1835#-O88FU9NByNahXk8S-TC:-O88FU9NByNahXk8S-TD:b9oh48i) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7280adbd9dd7e08d73feb0812319df9e2712038e/src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/writers/FailComputationParticipant.kt#L123)):*
> ```Kotlin
>           reason = MeasurementFailure.Reason.COMPUTATION_PARTICIPANT_FAILED
>           message =
>             "Computation Participant failed. ${request.externalDuchyId}: ${request.logMessage}"
> ```

This log message is copied directly to the public API in Measurement.Failure, so it should not contain IDs specific to the internal API. It's effectively by coincidence that the external Duchy ID is also the pubic API Duchy ID.

One option is to add a oneof field in the internal API for the failure source, which could be the certificate, requisition, or duchy key. This could then be surfaced in the public API as a resource name.
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/writers/FailComputationParticipant.kt` line 129 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1835#-O88EV1nAsTBtF1a9dV0:-O88EV1nAsTBtF1a9dV1:bz6w2cn) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7280adbd9dd7e08d73feb0812319df9e2712038e/src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/writers/FailComputationParticipant.kt#L129)):*
> ```Kotlin
>     val measurementLogEntryDetails = measurementLogEntryDetails {
>       logMessage =
>         "Computation Participant failed. ${request.externalDuchyId}: ${request.logMessage}"
> ```

Note that this information is already part of the log entry. The log messages should also be surfacable in the public API, so should not contain IDs specific to the internal API.

PR #1856, opened 2024-10-14
ci: Read Spanner processing units from GitHub environment

SanjayVas:
Different Halo test environments have different Spanner resource needs. For example, the `head` environment tends to have much lower traffic compared to the `qa` environment. Spanner autoscaling only supports whole nodes (multiples of 1000 processing units), whereas some test environments are often idle and do not need a whole node. Spanner nodes are relatively expensive (currently $657/month for the us-central1 region).

This also removes the specification of 100 processing nodes for the example Terraform configuration, as that is far below Spanner recommendations.


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1856)

PR #1857, opened 2024-10-14
feat: Throw Transient exceptions in HMSS GetCertificate call.

renjiezh:


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1857)

SanjayVas (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1857)* status: all files reviewed, 1 unresolved discussion (waiting on @renjiezh and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/mill/shareshuffle/HonestMajorityShareShuffleMill.kt` line 339 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1857#-O9GMIT95QPmweBZ1bHX:-O9GMIT95QPmweBZ1bHY:by7vwvx) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/48e1324a78f81301aaacb59af29ade0d530bad19/src/main/kotlin/org/wfanet/measurement/duchy/mill/shareshuffle/HonestMajorityShareShuffleMill.kt#L339)):*
> ```Kotlin
>         val message = "Fail to get certificate for $dataProviderCertificateName"
>         when (e.status.code) {
>           Status.Code.UNAVAILABLE,
> ```

This is *okay*, but UNAVAILABLE and DEADLINE_EXCEEDED could also be immediately retried instead. Though I understand if we'd ideally handle that in service configs instead of code. Can you add a TODO for doing immediate retries for those via service config?

stevenwarejones (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1857)* status: all files reviewed, 1 unresolved discussion (waiting on @renjiezh)

PR #1861, opened 2024-10-16
Classes for the TEE SDK

Marco-Premier:
- Added Crypto class
- Added GcsStorageClient class


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1861)

kungfucraig (changes requested):
Please drop the underscores from the package and directory names to match the convention elsewhere in the repo.

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1861#-:-O9KywiA1uBrwCthxleb:b-8mt8yn)* status: 0 of 12 files reviewed, 5 unresolved discussions (waiting on @Marco-Premier, @SanjayVas, and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/secure_computation/tee_sdk/cloud_storage/v1alpha/CloudStorage.kt` line 20 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1861#-O9Kui5g37aSpwZEl9aT:-O9Kui5g37aSpwZEl9aU:b-flhc81) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/35575ea9b89bcbe3c4bf0a17613e942a53920285/src/main/kotlin/org/wfanet/measurement/secure_computation/tee_sdk/cloud_storage/v1alpha/CloudStorage.kt#L20)):*
> ```Kotlin
> import java.io.InputStream
> 
> abstract class CloudStorage {
> ```

You should use the classes here instead of implementing new ones: https://github.com/world-federation-of-advertisers/common-jvm/tree/main/src/main/kotlin/org/wfanet/measurement/storage
___
*[`src/main/kotlin/org/wfanet/measurement/secure_computation/tee_sdk/crypto/v1alpha/Crypto.kt` line 25 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1861#-O9KumVO3KBZ5veF3deM:-O9KumVO3KBZ5veF3deN:b-ufjuz7) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/35575ea9b89bcbe3c4bf0a17613e942a53920285/src/main/kotlin/org/wfanet/measurement/secure_computation/tee_sdk/crypto/v1alpha/Crypto.kt#L25)):*
> ```Kotlin
> import com.google.protobuf.ByteString
> 
> class Crypto {
> ```

This should be a cloud agnostic interface with a Google Cloud specific implementation. A fake would be useful for testing.

Take a look at how the storage client packages are structured in common-jv. Same idea.
___
*[`src/main/kotlin/org/wfanet/measurement/secure_computation/tee_sdk/crypto/v1alpha/Crypto.kt` line 26 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1861#-O9Kuk6R13HBbTnDfxK5:-O9Kuk6R13HBbTnDfxK6:b-xwlm7u) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/35575ea9b89bcbe3c4bf0a17613e942a53920285/src/main/kotlin/org/wfanet/measurement/secure_computation/tee_sdk/crypto/v1alpha/Crypto.kt#L26)):*
> ```Kotlin
> 
> class Crypto {
>     private val transformation: String 
> ```

Please add a doc string.
___
*[`src/main/kotlin/org/wfanet/measurement/secure_computation/tee_sdk/crypto/v1alpha/Crypto.kt` line 33 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1861#-O9KurpLBp-QkkrLrQZj:-O9KurpLBp-QkkrLrQZk:bstgcqa) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/35575ea9b89bcbe3c4bf0a17613e942a53920285/src/main/kotlin/org/wfanet/measurement/secure_computation/tee_sdk/crypto/v1alpha/Crypto.kt#L33)):*
> ```Kotlin
> 
>     fun getKeyManagementServiceClient(): KeyManagementServiceClient {
>         return KeyManagementServiceClient.create()
> ```

Any reason not to cache this vs. creating it each time this function is called?
___
*[`src/main/kotlin/org/wfanet/measurement/secure_computation/tee_sdk/crypto/v1alpha/Crypto.kt` line 36 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1861#-O9KxVJ0967C8BbBDdw8:-O9KxVJ0967C8BbBDdw9:b-84s735) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/35575ea9b89bcbe3c4bf0a17613e942a53920285/src/main/kotlin/org/wfanet/measurement/secure_computation/tee_sdk/crypto/v1alpha/Crypto.kt#L36)):*
> ```Kotlin
>     }
> 
>     fun encryptData(data: ByteArray, symmetricKey: String): ByteArray {
> ```

As per comments in the design doc, you're going to need to operate on streams of data to make this scalable. This as a primitive works, but it's  what you want to build the applications on.

Think about the application operating over a row oriented data structure, where each row is independently decrypted/encrypted and you can stream rows through whatever application you are implementing. 

I think you should assume you can't fit the files you are operating on in memory.

stevenwarejones (commented):
> Reviewed 4 of 12 files at r1.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1861)* status: 4 of 12 files reviewed, 5 unresolved discussions (waiting on @kungfucraig, @Marco-Premier, and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/secure_computation/tee_sdk/crypto/v1alpha/Crypto.kt` line 25 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1861#-O9KumVO3KBZ5veF3deM:-O9LbAgR16NMOCkPActg:b-i6oa87) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/35575ea9b89bcbe3c4bf0a17613e942a53920285/src/main/kotlin/org/wfanet/measurement/secure_computation/tee_sdk/crypto/v1alpha/Crypto.kt#L25)):*

This should be a cloud agnostic interface with a Google Cloud specific implementation. A fake would be useful for testing.

Take a look at how the storage client packages are structured in common-jv. Same idea.
</blockquote></details>

i agree on creating an interface or abstract class which will help with testing. I'm agnostic if you just to create a FakeImpl or just mock it in tests.

kungfucraig (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1861)* status: 4 of 12 files reviewed, 5 unresolved discussions (waiting on @Marco-Premier, @SanjayVas, and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/secure_computation/tee_sdk/crypto/v1alpha/Crypto.kt` line 25 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1861#-O9KumVO3KBZ5veF3deM:-O9LuGyW971Sa3RSDaY1:b-icst2e) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/35575ea9b89bcbe3c4bf0a17613e942a53920285/src/main/kotlin/org/wfanet/measurement/secure_computation/tee_sdk/crypto/v1alpha/Crypto.kt#L25)):*

i agree on creating an interface or abstract class which will help with testing. I'm agnostic if you just to create a FakeImpl or just mock it in tests.
</blockquote></details>

In favor of the fake: It's highly likely that we'll want to bottom out all of application tests out on this functionality, and having a Fake will mean that each application's test can avoid doing its own mocks.

Marco-Premier (commented):
@kungfucraig @stevenwarejones I committed a `RecordioStorageClient` which implements `StorageClient` from `common-jvm`.

To addressed the line-by-line encryption / decryption I created a new class that handles Record IO file format that can be found here: `https://mesos.apache.org/documentation/latest/recordio/`.

This will work for files row-encrypted that uses RecordIO file format. 

Please let me know what you think about this approach. If you think this is the right path, we will need similar classes to handle other file formats (eg. Apache Sketch / Parquet ... )

N.B I haven't wrote any test or deleted other classes yet.

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1861#-:-O9PzehH4JnqGEb7h7Nt:b-30ucf4)* status: 3 of 13 files reviewed, 5 unresolved discussions (waiting on @SanjayVas and @stevenwarejones)

SanjayVas (changes requested):
> Reviewed 3 of 12 files at r1.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1861)* status: 3 of 13 files reviewed, 8 unresolved discussions (waiting on @kungfucraig, @Marco-Premier, and @stevenwarejones)

___
*[`MODULE.bazel` line 222 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1861#-O9QOtpm3F-1lmNgGB7W:-O9QOtpm3F-1lmNgGB7X:b-tndtnf) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1b28e7f4c37e746e68b65039374fe36daa9ce414/MODULE.bazel#L222)):*
> ```Starlark
>         # TODO(googleapis/java-cloud-bom#5279): Remove when managed by BOM.
>         "com.google.apis:google-api-services-storage:v1-rev20240706-2.0.0",
>         "com.google.cloud:google-cloud-kms:2.55.0",
> ```

Leave out the version for anything covered by a BOM. This should be covered by `com.google.cloud:libraries-bom` which is included by common-jvm.



_Code quote:_
```Starlark
:2.55.0
```
___
*[`MODULE.bazel.lock` line 0 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1861#-O9QPdyBD-cBLDgDV2D9:-O9QPdyBD-cBLDgDV2DA:b-7smni4) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1b28e7f4c37e746e68b65039374fe36daa9ce414/MODULE.bazel.lock#L0)):*
Why is this deleted? It should be updated by running `bazel mod deps --lockfile_mode=update`
___
*[`maven_install.json` line 0 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1861#-O9QPvMa4UogiFSPubbf:-O9QPvMa4UogiFSPubbg:b-4fmifm) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1b28e7f4c37e746e68b65039374fe36daa9ce414/maven_install.json#L0)):*
Why is this deleted? It should be updated by running `REPIN=1 bazel run @maven//:pin`
___
*[`src/main/kotlin/org/wfanet/measurement/secure_computation/tee_sdk/cloud_storage/v1alpha/CloudStorage.kt` line 20 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1861#-O9Kui5g37aSpwZEl9aT:-O9QQ48v4TpIjjTg1PNS:b32v57) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/35575ea9b89bcbe3c4bf0a17613e942a53920285/src/main/kotlin/org/wfanet/measurement/secure_computation/tee_sdk/cloud_storage/v1alpha/CloudStorage.kt#L20)):*

You should use the classes here instead of implementing new ones: https://github.com/world-federation-of-advertisers/common-jvm/tree/main/src/main/kotlin/org/wfanet/measurement/storage
</blockquote></details>

+1

kungfucraig (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1861)* status: 3 of 13 files reviewed, 11 unresolved discussions (waiting on @Marco-Premier and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/cloudstorage/v1alpha/RecordioStorageClient.kt` line 42 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1861#-O9VX2cE0iQM5TiEHv7b:-O9VX2cF6Yq4K7r9_RZL:b-q0dajp) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1b28e7f4c37e746e68b65039374fe36daa9ce414/src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/cloudstorage/v1alpha/RecordioStorageClient.kt#L42)):*
> ```Kotlin
>  * @param dataKey a base64-encoded symmetric data key
>  */
> class RecordioStorageClient(
> ```

Overall approach makes sense.
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/cloudstorage/v1alpha/RecordioStorageClient.kt` line 47 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1861#-O9VWr-l5rKURZXjJf1c:-O9VWr-l5rKURZXjJf1d:b3ur93p) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1b28e7f4c37e746e68b65039374fe36daa9ce414/src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/cloudstorage/v1alpha/RecordioStorageClient.kt#L47)):*
> ```Kotlin
> ) : StorageClient {
> 
>   private fun createKeysetHandleFromBase64(base64Key: String): KeysetHandle {
> ```

Maybe create a member that is KeysetHandle and put this in an init block for that.
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/cloudstorage/v1alpha/RecordioStorageClient.kt` line 101 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1861#-O9VXSizBZvyb6D0T1d5:-O9VXSizBZvyb6D0T1d6:b-kdu9lc) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1b28e7f4c37e746e68b65039374fe36daa9ce414/src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/cloudstorage/v1alpha/RecordioStorageClient.kt#L101)):*
> ```Kotlin
>    */
>   override suspend fun getBlob(blobKey: String): StorageClient.Blob? {
>     val blob = storageClient.getBlob(blobKey)
> ```

Does this read the whole blob into memory? Do we have an option to stream the blob instead? No need to do anything, but I'd like to discuss.

Marco-Premier (commented):
This PR now focus only on the RecordioStorageClient. Classes for accessing gcp kms will be added in a following PR

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1861#-:-O9k9ZDU1CDBxT6inEEy:b-7etlqc)* status: 0 of 14 files reviewed, 11 unresolved discussions (waiting on @kungfucraig, @SanjayVas, and @stevenwarejones)

___
*[`MODULE.bazel.lock` line  at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1861#-O9QPdyBD-cBLDgDV2D9:-O9k9Pp73EOnBHI-BXnw:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1b28e7f4c37e746e68b65039374fe36daa9ce414/MODULE.bazel.lock#L)):*

Why is this deleted? It should be updated by running `bazel mod deps --lockfile_mode=update`
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/cloudstorage/v1alpha/RecordioStorageClient.kt` line 42 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1861#-O9VX2cE0iQM5TiEHv7b:-O9j-ytfEVlvyOcm9Sbm:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1b28e7f4c37e746e68b65039374fe36daa9ce414/src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/cloudstorage/v1alpha/RecordioStorageClient.kt#L42)):*

Overall approach makes sense.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/cloudstorage/v1alpha/RecordioStorageClient.kt` line 47 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1861#-O9VWr-l5rKURZXjJf1c:-O9k9SWj-HjnoXsLtIVM:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1b28e7f4c37e746e68b65039374fe36daa9ce414/src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/cloudstorage/v1alpha/RecordioStorageClient.kt#L47)):*

Maybe create a member that is KeysetHandle and put this in an init block for that.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/cloudstorage/v1alpha/RecordioStorageClient.kt` line 101 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1861#-O9VXSizBZvyb6D0T1d5:-O9k9iro0PG6jGtzglSe:b-ps3unm) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1b28e7f4c37e746e68b65039374fe36daa9ce414/src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/cloudstorage/v1alpha/RecordioStorageClient.kt#L101)):*

Does this read the whole blob into memory? Do we have an option to stream the blob instead? No need to do anything, but I'd like to discuss.
</blockquote></details>

Actually not because the `read` method of `RecordioBlob` processes one row at the time. Process means read a row, decrypts it and emit the decrypted row.
So the client will decide the pace through which data flows from the google storage into the TEE app
___
*[`MODULE.bazel` line 222 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1861#-O9QOtpm3F-1lmNgGB7W:-O9j-x0t14KTQ5sD4YwK:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1b28e7f4c37e746e68b65039374fe36daa9ce414/MODULE.bazel#L222)):*

Leave out the version for anything covered by a BOM. This should be covered by `com.google.cloud:libraries-bom` which is included by common-jvm.
</blockquote></details>

Done.
___
*[`maven_install.json` line  at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1861#-O9QPvMa4UogiFSPubbf:-O9k9UuH42lLcZlQvK8G:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1b28e7f4c37e746e68b65039374fe36daa9ce414/maven_install.json#L)):*

Why is this deleted? It should be updated by running `REPIN=1 bazel run @maven//:pin`
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/secure_computation/tee_sdk/cloud_storage/v1alpha/CloudStorage.kt` line 20 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1861#-O9Kui5g37aSpwZEl9aT:-O9j-y6V-loL2P-4jrXK:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/35575ea9b89bcbe3c4bf0a17613e942a53920285/src/main/kotlin/org/wfanet/measurement/secure_computation/tee_sdk/cloud_storage/v1alpha/CloudStorage.kt#L20)):*

+1
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/secure_computation/tee_sdk/crypto/v1alpha/Crypto.kt` line 25 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1861#-O9KumVO3KBZ5veF3deM:-O9k9XIB1RdENCmDr2dP:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/35575ea9b89bcbe3c4bf0a17613e942a53920285/src/main/kotlin/org/wfanet/measurement/secure_computation/tee_sdk/crypto/v1alpha/Crypto.kt#L25)):*

In favor of the fake: It's highly likely that we'll want to bottom out all of application tests out on this functionality, and having a Fake will mean that each application's test can avoid doing its own mocks.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/secure_computation/tee_sdk/crypto/v1alpha/Crypto.kt` line 26 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1861#-O9Kuk6R13HBbTnDfxK5:-O9k9XwsAEENuXDnTJNP:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/35575ea9b89bcbe3c4bf0a17613e942a53920285/src/main/kotlin/org/wfanet/measurement/secure_computation/tee_sdk/crypto/v1alpha/Crypto.kt#L26)):*

Please add a doc string.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/secure_computation/tee_sdk/crypto/v1alpha/Crypto.kt` line 33 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1861#-O9KurpLBp-QkkrLrQZj:-O9k9X69CXHH5KYzeNM2:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/35575ea9b89bcbe3c4bf0a17613e942a53920285/src/main/kotlin/org/wfanet/measurement/secure_computation/tee_sdk/crypto/v1alpha/Crypto.kt#L33)):*

Any reason not to cache this vs. creating it each time this function is called?
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/secure_computation/tee_sdk/crypto/v1alpha/Crypto.kt` line 36 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1861#-O9KxVJ0967C8BbBDdw8:-O9k9Wk225bWsBDqe7Nz:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/35575ea9b89bcbe3c4bf0a17613e942a53920285/src/main/kotlin/org/wfanet/measurement/secure_computation/tee_sdk/crypto/v1alpha/Crypto.kt#L36)):*

As per comments in the design doc, you're going to need to operate on streams of data to make this scalable. This as a primitive works, but it's  what you want to build the applications on.

Think about the application operating over a row oriented data structure, where each row is independently decrypted/encrypted and you can stream rows through whatever application you are implementing. 

I think you should assume you can't fit the files you are operating on in memory.
</blockquote></details>

Done.

SanjayVas (changes requested):
> Reviewed 12 of 14 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1861)* status: 12 of 14 files reviewed, 15 unresolved discussions (waiting on @kungfucraig, @Marco-Premier, and @stevenwarejones)

___
*[`MODULE.bazel.lock` line 2 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1861#-O9kwNy69UKpWfP5obZ4:-O9kwNy69UKpWfP5obZ5:b-eg0c00) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a9a6846a932251fa16d0e171c9ae132cb9f2e722/MODULE.bazel.lock#L2)):*
> ```JSON
> {
>   "lockFileVersion": 6,
> ```

This is showing an older lockfile version. Are you using `bazelisk` to run Bazel so it respects the `.bazelversion` file for the module?
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/cloudstorage/v1alpha/RecordioStorageClient.kt` line 31 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1861#-O9kykMKBf27D4-VRL8F:-O9kykMKBf27D4-VRL8G:b-fwivw1) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a9a6846a932251fa16d0e171c9ae132cb9f2e722/src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/cloudstorage/v1alpha/RecordioStorageClient.kt#L31)):*
> ```Kotlin
> 
> /**
>  * A wrapper class for [StorageClient] interface that uses Tink AEAD encryption/decryption for
> ```

Document why this doesn't use `org.wfanet.measurement.common.crypto.tink.KmsStorageClient`. I assume the reason is because the encryption is happening on the individual records inside the file rather than on the whole blob?



_Code quote:_
```Kotlin
uses Tink AEAD encryption/decryption
```
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/cloudstorage/v1alpha/RecordioStorageClient.kt` line 32 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1861#-O9kz06x2Mou8N90Y9wj:-O9kz06x2Mou8N90Y9wk:b6jfmkn) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a9a6846a932251fa16d0e171c9ae132cb9f2e722/src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/cloudstorage/v1alpha/RecordioStorageClient.kt#L32)):*
> ```Kotlin
> /**
>  * A wrapper class for [StorageClient] interface that uses Tink AEAD encryption/decryption for
>  * blob/object storage operations on RecordIO-based files.
> ```

What does this mean? If this is specifically supposed to be to be Apache Mesos RecordIO, then state that. If it's something else, make it clear exactly what it is.



_Code quote:_
```Kotlin
RecordIO-based
```
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/cloudstorage/v1alpha/RecordioStorageClient.kt` line 37 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1861#-O9kzNv88pxDXuvoRyJr:-O9kzNv88pxDXuvoRyJs:bpmlnw8) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a9a6846a932251fa16d0e171c9ae132cb9f2e722/src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/cloudstorage/v1alpha/RecordioStorageClient.kt#L37)):*
> ```Kotlin
>  * @param dataKey a base64-encoded symmetric data key
>  */
> class RecordioStorageClient(
> ```

See https://developer.android.com/kotlin/style-guide#camel_case



_Suggestion:_
```Kotlin
RecordIo
```
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/cloudstorage/v1alpha/RecordioStorageClient.kt` line 39 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1861#-O9kxGTS3S_6vZYkCjwl:-O9kxGTT90rgANmYquEi:b-k95r4a) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a9a6846a932251fa16d0e171c9ae132cb9f2e722/src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/cloudstorage/v1alpha/RecordioStorageClient.kt#L39)):*
> ```Kotlin
> class RecordioStorageClient(
>   private val storageClient: StorageClient,
>   private val base64Dek: String,
> ```

This should take in a `StreamingAead` object, not key material. Keys should be read into KeySetHandle objects and then wrapped in crypto primitives at the edges.



_Code quote:_
```Kotlin
private val base64Dek: String,
```
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/cloudstorage/v1alpha/RecordioStorageClient.kt` line 48 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1861#-O9kzcD1Bdt-evUiQCzb:-O9kzcD2CfbhhqVdwFQR:bpkmco8) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a9a6846a932251fa16d0e171c9ae132cb9f2e722/src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/cloudstorage/v1alpha/RecordioStorageClient.kt#L48)):*
> ```Kotlin
>   }
> 
>   private fun createAeadFromBase64(base64Key: String): AesGcmJce {
> ```

The Tink `subtle` subpackages are explicitly not part of the public API. Use `StreamingAead` instead.



_Code quote:_
```Kotlin
AesGcmJce
```
___
*[`src/test/kotlin/org/wfanet/measurement/securecomputation/teesdk/cloudstorage/v1alpha/RecordioStorageClientTest.kt` line 53 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1861#-O9l-Uf08NpiCxQ9hWVQ:-O9l-Uf08NpiCxQ9hWVR:b-wtrdng) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a9a6846a932251fa16d0e171c9ae132cb9f2e722/src/test/kotlin/org/wfanet/measurement/securecomputation/teesdk/cloudstorage/v1alpha/RecordioStorageClientTest.kt#L53)):*
> ```Kotlin
>   @Before
>   fun setUp() {
>     storageClient = mock()
> ```

Use a real StorageClient for the test. e.g. InMemoryStorageClient (which is specifically for tests like this) or FileSystemStorageClient (if you want to deal with bigger blobs).

kungfucraig (changes requested):
Thanks for the updates. I'm generally going to defer to Sanjay on the crypto stuff, and will wait for those comments to be resolved before reviewing again.

> Reviewed 4 of 14 files at r3.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1861#-:-O9lHXWLAecqluEc8mtt:b-jxt97m)* status: all files reviewed, 15 unresolved discussions (waiting on @Marco-Premier, @SanjayVas, and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/cloudstorage/v1alpha/RecordioStorageClient.kt` line 37 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1861#-O9lFst57T21wIr61b31:-O9lFst57T21wIr61b32:b-arould) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a9a6846a932251fa16d0e171c9ae132cb9f2e722/src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/cloudstorage/v1alpha/RecordioStorageClient.kt#L37)):*
> ```Kotlin
>  * @param dataKey a base64-encoded symmetric data key
>  */
> class RecordioStorageClient(
> ```

Capitalize I and O as it's short for RecordInputOutput. In the file name too.
___
*[`src/test/kotlin/org/wfanet/measurement/securecomputation/teesdk/cloudstorage/v1alpha/RecordioStorageClientTest.kt` line 53 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1861#-O9l-Uf08NpiCxQ9hWVQ:-O9lHfYz8qhXB4lUN_SW:bu8nlrt) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a9a6846a932251fa16d0e171c9ae132cb9f2e722/src/test/kotlin/org/wfanet/measurement/securecomputation/teesdk/cloudstorage/v1alpha/RecordioStorageClientTest.kt#L53)):*

Use a real StorageClient for the test. e.g. InMemoryStorageClient (which is specifically for tests like this) or FileSystemStorageClient (if you want to deal with bigger blobs).
</blockquote></details>

+1 

In general I'd suggest a test that runs a round trip write/reead as follows:
1. Write an encrypted file row by row to the blob client
2. Read it back and decrypt it.
3. Make sure you have what you started with.

Ideally no mocks would be required at all for this test.

stevenwarejones (changes requested):
> Reviewed 1 of 6 files at r2, 1 of 14 files at r3, 1 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1861)* status: 13 of 14 files reviewed, 17 unresolved discussions (waiting on @dawnwang22, @kungfucraig, @Marco-Premier, and @SanjayVas)

___
*[`src/test/kotlin/org/wfanet/measurement/securecomputation/teesdk/cloudstorage/v1alpha/RecordioStorageClientTest.kt` line 53 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1861#-O9l-Uf08NpiCxQ9hWVQ:-O9ohj_iAR1ALUAm1HSX:b32v57) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a9a6846a932251fa16d0e171c9ae132cb9f2e722/src/test/kotlin/org/wfanet/measurement/securecomputation/teesdk/cloudstorage/v1alpha/RecordioStorageClientTest.kt#L53)):*

+1 

In general I'd suggest a test that runs a round trip write/reead as follows:
1. Write an encrypted file row by row to the blob client
2. Read it back and decrypt it.
3. Make sure you have what you started with.

Ideally no mocks would be required at all for this test.
</blockquote></details>

+1
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/cloudstorage/v1alpha/BUILD.bazel` line 1 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1861#-O9ogRkQ179-P5RD5LAD:-O9ogRkQ179-P5RD5LAE:b-jpuerb) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f0f617bc9c0b812cffb1dddec32db549adf47043/src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/cloudstorage/v1alpha/BUILD.bazel#L1)):*
> ```Starlark
> load("@wfa_rules_kotlin_jvm//kotlin:defs.bzl", "kt_jvm_library")
> ```

why is the path under v1alpha? Aren't we only doing that for the api?
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/cloudstorage/v1alpha/RecordIOStorageClient.kt` line 121 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1861#-O9oh_wOEZUeEVv9KNkA:-O9oh_wOEZUeEVv9KNkB:bht8inh) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f0f617bc9c0b812cffb1dddec32db549adf47043/src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/cloudstorage/v1alpha/RecordIOStorageClient.kt#L121)):*
> ```Kotlin
> 
>         while (true) {
>           val lengthEndIndex = buffer.indexOf('\n'.code.toByte())
> ```

can you make `\n' a separator constant
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/cloudstorage/v1alpha/RecordioStorageClient.kt` line 39 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1861#-O9kxGTS3S_6vZYkCjwl:-O9ogtNg0FhwhlbkJXWb:b-ijm0a6) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a9a6846a932251fa16d0e171c9ae132cb9f2e722/src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/cloudstorage/v1alpha/RecordioStorageClient.kt#L39)):*

@SanjayVas I'm happy to use `StreamingAead` but I just want to clarify the use case here. We use hybrid encryption so the dek will be decrypted using the kek (both stored on kms) using attestation.
So in the code we will have a symmetric key that will be used to encrypt/decrypt data.
Also, following @kungfucraig 's suggestion, this class supports row-based encryption / decryption to be able to process large file avoiding memory constraints.

Imagine VID labeling: a potentially 1TB file that won't fit in memory, the idea is to read 1 row at the time, label it and then use writeBlob to write the labelled row back to google storage, maintaining a low memory usage through the entire process.

So I'm not sure I can take full advantage of `StreamingAead` since I'll be encrypting / decrypting a single row at the time.
Finally given that the DEK will be available on the code, I'm not sure how else to build the `KeyMaterial` without using some deprecated class such as `CleartextKeysetHandle` ?
</blockquote></details>

I'd do something like this @Marco-Premier - https://github.com/tink-crypto/tink/blob/master/java_src/examples/streamingaead/StreamingAeadExample.java#L116

SanjayVas (changes requested):
> Reviewed 1 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1861)* status: 13 of 14 files reviewed, 16 unresolved discussions (waiting on @dawnwang22, @kungfucraig, and @Marco-Premier)

___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/cloudstorage/v1alpha/RecordioStorageClient.kt` line 37 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1861#-O9lFst57T21wIr61b31:-O9q-y-MCC0URYY1re-P:bqgmi5l) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a9a6846a932251fa16d0e171c9ae132cb9f2e722/src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/cloudstorage/v1alpha/RecordioStorageClient.kt#L37)):*
> Capitalize I and O as it's short for RecordInputOutput. In the file name too.

This is incorrect. Only the `I` should be capitalized. See https://developer.android.com/kotlin/style-guide#camel_case
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/cloudstorage/v1alpha/RecordioStorageClient.kt` line 39 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1861#-O9kxGTS3S_6vZYkCjwl:-O9q0OPN4qMjvx2_xqFA:bhs9xet) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a9a6846a932251fa16d0e171c9ae132cb9f2e722/src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/cloudstorage/v1alpha/RecordioStorageClient.kt#L39)):*
> So I'm not sure I can take full advantage of `StreamingAead` since I'll be encrypting / decrypting a single row at the time.

Entirely fine to just use regular `Aead` if we're okay having one record in memory at a time (as opposed to streaming sub-record chunks). We just need to be clear in documentation that each item in the resulting flow from Blob.read is a record.

> We use hybrid encryption so the dek will be decrypted using the kek (both stored on kms) using attestation.

I believe this is envelope encryption, not hybrid. Meaning I assume both the KEK and DEK are symmetric keys, whereas hybrid encryption refers to a scheme using both asymmetric and symmetric encryption.

> Finally given that the DEK will be available on the code, I'm not sure how else to build the `KeyMaterial` without using some deprecated class such as `CleartextKeysetHandle` ?

The point is to avoid ever having the DEK be available in cleartext to application code. If this is coming from KMS, then let's not reinvent the wheel here. Just follow Tink examples on exactly how to do envelope encryption with a KMS key. e.g. https://cloud.google.com/kms/docs/client-side-encryption#java

Note that I'd personally prefer to follow our existing pattern of keeping Tink usage abstracted under `org.wfanet.measurement.common.crypto.tink` using interfaces defined in `org.wfanet.measurement.common.crypto`. In this case, I think it would be a matter of adding envelope encryption support where we currently only support asymmetric keys for hybrid encryption (PublicKeyHandle/PrivateKeyHandle). That said, you can hold off on doing that until you're done prototyping.

kungfucraig (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1861)* status: 13 of 14 files reviewed, 14 unresolved discussions (waiting on @dawnwang22, @Marco-Premier, and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/cloudstorage/v1alpha/RecordioStorageClient.kt` line 37 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1861#-O9lFst57T21wIr61b31:-O9qHOJW8yB7LqvBsG82:b-bb7sn9) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a9a6846a932251fa16d0e171c9ae132cb9f2e722/src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/cloudstorage/v1alpha/RecordioStorageClient.kt#L37)):*

> Capitalize I and O as it's short for RecordInputOutput. In the file name too.

This is incorrect. Only the `I` should be capitalized. See https://developer.android.com/kotlin/style-guide#camel_case
</blockquote></details>

@SanjayVas is right

kungfucraig (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1861)* status: 13 of 14 files reviewed, 14 unresolved discussions (waiting on @dawnwang22, @Marco-Premier, and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/cloudstorage/v1alpha/RecordioStorageClient.kt` line 39 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1861#-O9kxGTS3S_6vZYkCjwl:-O9wGQpZ-BQ_D_tjU44X:b8rqm8t) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a9a6846a932251fa16d0e171c9ae132cb9f2e722/src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/cloudstorage/v1alpha/RecordioStorageClient.kt#L39)):*

We didn't really discuss this but I was thinking we should use whole file streaming encryption vs record-level encryption. @kungfucraig - but I'm open to regular aead if we're doing record-level. Wouldn't streaming whole file be more efficient?
</blockquote></details>

Streaming the whole can work, but we should emit records.

The two goals I have in mind are 1) to be able to process records in parallel, and 2) to avoid requiring the entire file or all of its records be in memory at once.

Labeling should basically be a stream of read/label/write operations, where the labeling part could be multi-threaded and the write part could be a sharded.

Marco-Premier:
Closing in favor of: https://github.com/world-federation-of-advertisers/common-jvm/pull/285

Marco-Premier:
Closing --> https://github.com/world-federation-of-advertisers/common-jvm/pull/285

PR #1870, opened 2024-10-21
perf: Use CASE expression in StreamModelRollouts query

SanjayVas:
This allows short-circuiting.


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1870)

Issue #1368, opened 2023-12-14
SpannerModelRolloutsServiceTest is flaky (times out)


SanjayVas:
`//src/test/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner:SpannerModelRolloutsServiceTest` frequently times out on in the GitHub Actions workflow. Note that other Spanner-based service tests do not appear to have this issue.

[test.log](https://github.com/world-federation-of-advertisers/cross-media-measurement/files/13679021/test.log)

SanjayVas:
This now appears to be consistently failing in the build-and-test GitHub Actions workflow.

SanjayVas:
The best theory I have as to why it's just _this_ service and no others that's having this issue is on the database schema. There's more layers of interleaved tables for model rollouts relative to its usage pattern. This at minimum may be difficult for the Spanner emulator used in tests to handle. It's also possible this may result in performance issues for the real Cloud Spanner, but no callers to this service yet exist.

The interleaved table structure is as follows, with `->` representing "interleaved in":
ModelRollouts -> ModelLines -> ModelSuites -> ModelProviders

Note that the above theory is *very* loose and does not have much to back it up.

SanjayVas:
A contributing factor may be the lack of a LIMIT on the "readLatestModelRolloutData" query. Instead of applying a limit, the code is just taking the first result. I discovered this while working on #2353.

PR #1871, opened 2024-10-21
test: Add per-test timeout to SpannerModelRolloutsServiceTest

SanjayVas:
This test is flaky in such a way that the failures cause the test to hang indefinitely. This change at least allows it to fail faster.

See #1368


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1871)

SanjayVas:
This doesn't appear to be helping :(

PR #1873, opened 2024-10-22
rabbit mq handler

Marco-Premier:
RabbitMQ handler that receives messages from the Rabbit MQ and dispatch them to another process that runs the TEE APP.

This will ensure that each `Work` will be processed with fresh memory and no data is leaked between `Work` executions.

Note: 
- Tests will be added after discussing the class
- Is authentication using username and password sufficient considering that rabbit mq server will run in the same VPC as MIG VMs and will only accept connection from them?


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1873)

stevenwarejones (changes requested):
wdyt about adding a unit test with a local broker? https://github.com/rabbitmq/rabbitmq-java-client?tab=readme-ov-file#launching-tests-with-a-local-broker

> Reviewed 4 of 5 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1873#-:-O9okOOiBNBl4scnn66i:b4wxfi7)* status: 4 of 5 files reviewed, 1 unresolved discussion (waiting on @kungfucraig, @Marco-Premier, and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/client/v1alpha/BUILD.bazel` line 3 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1873#-O9oi9B68CEW-_aZuUqq:-O9oi9B68CEW-_aZuUqr:bg91qmn) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/971beea91aae0ca71fdfc5757818dcb63d39b70a/src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/client/v1alpha/BUILD.bazel#L3)):*
> ```Starlark
> load("@wfa_rules_kotlin_jvm//kotlin:defs.bzl", "kt_jvm_library")
> 
> package(default_visibility = [
> ```

ditto on the v1alpha comment. I don't think we need `v1alpha` in the path for kotlin files

kungfucraig (changes requested):
> Reviewed 2 of 5 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1873)* status: 4 of 5 files reviewed, 4 unresolved discussions (waiting on @Marco-Premier and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/client/v1alpha/RabbitMQHandler.kt` line 32 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1873#-O9qMpKjD64RxtHCob4e:-O9qMpKjD64RxtHCob4f:b-7969wq) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/971beea91aae0ca71fdfc5757818dcb63d39b70a/src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/client/v1alpha/RabbitMQHandler.kt#L32)):*
> ```Kotlin
>  * - Unsubscribe from a RabbitMQ queue by safely closing the channel and connection.
>  */
> class RabbitMQHandler {
> ```

Is this code supposed to be running inside of the TEE VM?

I left a question on your design doc about auto scaling.
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/client/v1alpha/RabbitMQHandler.kt` line 81 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1873#-O9qLj7Z-iR09iuAVIan:-O9qLj7Z-iR09iuAVIao:b-s53of3) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/971beea91aae0ca71fdfc5757818dcb63d39b70a/src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/client/v1alpha/RabbitMQHandler.kt#L81)):*
> ```Kotlin
>     channel.basicConsume(rabbit_queue_name, false, { _: String, delivery: Delivery ->
>       val body = delivery.body
>       val process = ProcessBuilder("java", "-jar", "/app/tee_app.jar").start()
> ```

What's going on here? 
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/client/v1alpha/RabbitMQHandler.kt` line 131 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1873#-O9qM5ei6-RxxJKqy3Tc:-O9qM5ei6-RxxJKqy3Td:b6e48v7) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/971beea91aae0ca71fdfc5757818dcb63d39b70a/src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/client/v1alpha/RabbitMQHandler.kt#L131)):*
> ```Kotlin
> 
> fun main() {
>   val rabbitMQHandler = RabbitMQHandler()
> ```

We typically use flags not environment variables for this kind of thing.

SanjayVas (changes requested):
> Reviewed 2 of 5 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1873)* status: 4 of 5 files reviewed, 7 unresolved discussions (waiting on @kungfucraig, @Marco-Premier, and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/client/v1alpha/RabbitMQHandler.kt` line 32 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1873#-O9qH0c59b2W8VaYSD0D:-O9qH0c59b2W8VaYSD0E:bf8098) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/971beea91aae0ca71fdfc5757818dcb63d39b70a/src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/client/v1alpha/RabbitMQHandler.kt#L32)):*
> ```Kotlin
>  * - Unsubscribe from a RabbitMQ queue by safely closing the channel and connection.
>  */
> class RabbitMQHandler {
> ```

nit: "Handler" to me sounds like an event handler or callback. Perhaps "RabbitMqClient"?



_Suggestion:_
```Kotlin
RabbitMqClient
```
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/client/v1alpha/RabbitMQHandler.kt` line 71 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1873#-O9qGXtmEH_ezAvl_bYa:-O9qGXtn7KTIPpsfsEy3:b-tt5qt0) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/971beea91aae0ca71fdfc5757818dcb63d39b70a/src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/client/v1alpha/RabbitMQHandler.kt#L71)):*
> ```Kotlin
>    * - Acknowledges the message if the process completes successfully, or negatively acknowledges it if the process fails.
>    */
>   fun subscribeToQueue(rabbit_host: String, rabbit_port: Int, rabbit_username: String, rabbit_password: String, rabbit_queue_name: String) {
> ```

See https://developer.android.com/kotlin/style-guide#non-constant_names



_Suggestion:_
```Kotlin
rabbitHost
```
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/client/v1alpha/RabbitMQHandler.kt` line 71 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1873#-O9qHq7P0Zf9iSGHEFyd:-O9qHq7P0Zf9iSGHEFye:bfdcezz) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/971beea91aae0ca71fdfc5757818dcb63d39b70a/src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/client/v1alpha/RabbitMQHandler.kt#L71)):*
> ```Kotlin
>    * - Acknowledges the message if the process completes successfully, or negatively acknowledges it if the process fails.
>    */
>   fun subscribeToQueue(rabbit_host: String, rabbit_port: Int, rabbit_username: String, rabbit_password: String, rabbit_queue_name: String) {
> ```

1. Use idiomatic Kotlin, including using coroutines
2. Decouple the RabbitMQ client code from the consumer.

I'd imagine the client API to be something like

```kotlin
/**
 * Subscribes to the specified queue.
 *
 * @return channel to receive messages on, which cancels the subscription when closed
 */
fun subscribe( /* params */): ReceiveChannel<QueueMessage>
```

Where I assume QueueMessage would expose the ack/nack methods.
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/client/v1alpha/RabbitMQHandler.kt` line 131 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1873#-O9qM5ei6-RxxJKqy3Tc:-O9qN5Fh0E48Gp532X8H:b-1ng7j4) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/971beea91aae0ca71fdfc5757818dcb63d39b70a/src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/client/v1alpha/RabbitMQHandler.kt#L131)):*

We typically use flags not environment variables for this kind of thing.
</blockquote></details>

+1. See commandLineMain.

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1873)* status: 4 of 5 files reviewed, 7 unresolved discussions (waiting on @kungfucraig and @Marco-Premier)

___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/client/v1alpha/RabbitMQHandler.kt` line 81 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1873#-O9qLj7Z-iR09iuAVIan:-O9tTPnGADoPd1qZ5Ox0:b-dpdat1) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/971beea91aae0ca71fdfc5757818dcb63d39b70a/src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/client/v1alpha/RabbitMQHandler.kt#L81)):*

What's going on here?
</blockquote></details>

I believe the goal is to launch the app each time a new task is received. then shut down the app afterwards.

There's two things we want to balance:
1. Make each task as ephemeral as possible
   i. no two tasks running simultaneously on the same machine
   ii. ideally, so the possibility of communication across tasks is minimized
2. Not have to re-provision the vm each time (although I'm open to this if the latency is sufficiently low)

Marco-Premier (commented):
Thanks @stevenwarejones for this.
I've updated the PR, still no test. Will add them as soon as the @kungfucraig and @SanjayVas are happy with the code / approach?

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1873#-:-O9u4xizEzH3sY6qHLLF:b-38ag00)* status: 4 of 5 files reviewed, 6 unresolved discussions (waiting on @kungfucraig, @SanjayVas, and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/client/v1alpha/BUILD.bazel` line 3 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1873#-O9oi9B68CEW-_aZuUqq:-O9tP6Xy3AMpK-cUlIZB:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/971beea91aae0ca71fdfc5757818dcb63d39b70a/src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/client/v1alpha/BUILD.bazel#L3)):*

ditto on the v1alpha comment. I don't think we need `v1alpha` in the path for kotlin files
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/client/v1alpha/RabbitMQHandler.kt` line 32 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1873#-O9qMpKjD64RxtHCob4e:-O9tNoVv2wkSOv24Pqvp:b-vreixd) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/971beea91aae0ca71fdfc5757818dcb63d39b70a/src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/client/v1alpha/RabbitMQHandler.kt#L32)):*

Is this code supposed to be running inside of the TEE VM?

I left a question on your design doc about auto scaling.
</blockquote></details>

Yes, both this code and the TEE APP will run in the same VM deployed using MIG.
The idea is to run each `Work` on a separate process to ensure that no memory is shared between `Works` for privacy reasons.
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/client/v1alpha/RabbitMQHandler.kt` line 71 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1873#-O9qGXtmEH_ezAvl_bYa:-O9tPablE7wV3sAKDmM5:bkxtmha) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/971beea91aae0ca71fdfc5757818dcb63d39b70a/src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/client/v1alpha/RabbitMQHandler.kt#L71)):*

See https://developer.android.com/kotlin/style-guide#non-constant_names
</blockquote></details>

auch, sorry. This is a python habit.
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/client/v1alpha/RabbitMQHandler.kt` line 71 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1873#-O9qHq7P0Zf9iSGHEFyd:-O9tow8i8muQlTUw0EnP:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/971beea91aae0ca71fdfc5757818dcb63d39b70a/src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/client/v1alpha/RabbitMQHandler.kt#L71)):*

1. Use idiomatic Kotlin, including using coroutines
2. Decouple the RabbitMQ client code from the consumer.

I'd imagine the client API to be something like

```kotlin
/**
 * Subscribes to the specified queue.
 *
 * @return channel to receive messages on, which cancels the subscription when closed
 */
fun subscribe( /* params */): ReceiveChannel<QueueMessage>
```

Where I assume QueueMessage would expose the ack/nack methods.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/client/v1alpha/RabbitMQHandler.kt` line 81 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1873#-O9qLj7Z-iR09iuAVIan:-O9towoeEJwpOQW1roZ7:b-kcc0kg) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/971beea91aae0ca71fdfc5757818dcb63d39b70a/src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/client/v1alpha/RabbitMQHandler.kt#L81)):*

I believe the goal is to launch the app each time a new task is received. then shut down the app afterwards.

There's two things we want to balance:
1. Make each task as ephemeral as possible
   i. no two tasks running simultaneously on the same machine
   ii. ideally, so the possibility of communication across tasks is minimized
2. Not have to re-provision the vm each time (although I'm open to this if the latency is sufficiently low)
</blockquote></details>

+1

both apps live in the same VM and are deployed together
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/client/v1alpha/RabbitMQHandler.kt` line 131 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1873#-O9qM5ei6-RxxJKqy3Tc:-O9u4ue40lO-rlMkGvPW:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/971beea91aae0ca71fdfc5757818dcb63d39b70a/src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/client/v1alpha/RabbitMQHandler.kt#L131)):*

+1. See commandLineMain.
</blockquote></details>

Done.

SanjayVas (changes requested):
> Reviewed 1 of 5 files at r1, 2 of 4 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1873)* status: 4 of 6 files reviewed, 8 unresolved discussions (waiting on @kungfucraig, @Marco-Premier, and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/client/RabbitMQClient.kt` line 102 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1873#-O9vaX4h5fnPcOLKTDi2:-O9vaX4h5fnPcOLKTDi3:ba1o5ie) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3b302d55a4c2d7938f9d5e4230f08842d40eef96/src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/client/RabbitMQClient.kt#L102)):*
> ```Kotlin
>    */
>   @OptIn(ExperimentalCoroutinesApi::class)
>   suspend fun subscribe(): ReceiveChannel<QueueMessage> = coroutineScope {
> ```

It looks like a single connection can be used for multiple subscriptions. This means that the client should handle the single connection and itself be Autocloseable where close() cancels the connection. `subscribe` should take in the queue name. Closing the returned coroutine Channel is unsubscribing from the queue, which I assume should result in closing the RabbitMQ channel.
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/client/RabbitMQClient.kt` line 132 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1873#-O9v_JyH0adbc-w_6Mx8:-O9v_JyH0adbc-w_6Mx9:b-oxn2h8) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3b302d55a4c2d7938f9d5e4230f08842d40eef96/src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/client/RabbitMQClient.kt#L132)):*
> ```Kotlin
>   suspend fun processQueueMessages(queue: ReceiveChannel<QueueMessage>) {
>     for (message in queue) {
>       val process = ProcessBuilder("java", "-jar", "/app/tee_app.jar").start()
> ```

I'll leave @kungfucraig to comment on the other thread if launching a new process is actually what we want to do. Assuming it is, I think this should just be an arbitrary executable and not necessarily a JAR. In practice it's likely to be a shell script calling a JAR just like this, but we don't need the Java piece hard-coded here.
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/client/v1alpha/RabbitMQHandler.kt` line 71 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1873#-O9qHq7P0Zf9iSGHEFyd:-O9vXkLe116gQrE-ED76:b-mg0tot) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/971beea91aae0ca71fdfc5757818dcb63d39b70a/src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/client/v1alpha/RabbitMQHandler.kt#L71)):*

Done.
</blockquote></details>

I think you only did part 1 here. What I'm looking for is a RabbitMqClient class/library which just provides a Kotlin coroutine API around RabbitMQ. This should eventually be in common-jvm under its own `common.rabbitmq` package as it's useful for more in Halo than just the TEE control plane.

Separately, have code that uses that client for the control plane behavior.
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/client/BUILD.bazel` line 12 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1873#-O9vXS0S49oIKtJgcmAF:-O9vXS0S49oIKtJgcmAG:b-ezd10o) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3b302d55a4c2d7938f9d5e4230f08842d40eef96/src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/client/BUILD.bazel#L12)):*
> ```Starlark
>     srcs = ["RabbitMQClient.kt"],
>     deps = [
>         "@maven//:com_rabbitmq_amqp_client",
> ```

Create an alias under //imports and use that for any maven dep.

kungfucraig (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1873)* status: 4 of 6 files reviewed, 7 unresolved discussions (waiting on @Marco-Premier, @SanjayVas, and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/client/RabbitMQClient.kt` line 132 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1873#-O9v_JyH0adbc-w_6Mx8:-O9wE2fl3Xv3vaFUcWx4:b-dj4z4o) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3b302d55a4c2d7938f9d5e4230f08842d40eef96/src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/client/RabbitMQClient.kt#L132)):*

I'll leave @kungfucraig to comment on the other thread if launching a new process is actually what we want to do. Assuming it is, I think this should just be an arbitrary executable and not necessarily a JAR. In practice it's likely to be a shell script calling a JAR just like this, but we don't need the Java piece hard-coded here.
</blockquote></details>

Updated the other thread. 

+1 on using an arbitrary executable and not assuming java.
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/client/RabbitMQClient.kt` line 137 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1873#-O9wESFpDkPx2oyQGxta:-O9wESFpDkPx2oyQGxtb:b-8xtrz1) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3b302d55a4c2d7938f9d5e4230f08842d40eef96/src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/client/RabbitMQClient.kt#L137)):*
> ```Kotlin
>         outputStream.flush()
>       }
>       process.inputStream.bufferedReader().use { it.readText() }
> ```

Is the intent to just drop all of the output?
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/client/v1alpha/RabbitMQHandler.kt` line 32 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1873#-O9qMpKjD64RxtHCob4e:-O9w8jEs6taIcOcXnS5_:bzcqd5y) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/971beea91aae0ca71fdfc5757818dcb63d39b70a/src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/client/v1alpha/RabbitMQHandler.kt#L32)):*

Yes, both this code and the TEE APP will run in the same VM deployed using MIG.
The idea is to run each `Work` on a separate process to ensure that no memory is shared between `Works` for privacy reasons.
</blockquote></details>

Let's continue the TEE/security conversation on the thread below, and the auto scaling conversation on the design doc. I still don't understand how autoscaling will work in practice.
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/client/v1alpha/RabbitMQHandler.kt` line 81 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1873#-O9qLj7Z-iR09iuAVIan:-O9w9USZ5B0gWDRWsdoL:b-72fftx) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/971beea91aae0ca71fdfc5757818dcb63d39b70a/src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/client/v1alpha/RabbitMQHandler.kt#L81)):*

+1

both apps live in the same VM and are deployed together
</blockquote></details>

I understand. This needs to be documented in the code, especially the reasons for why such a thing is necessary.

Would it make sense to re-provisioned the VM any time it would do work for more than a single EDP?

stevenwarejones (commented):
> Reviewed 2 of 4 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1873)* status: 5 of 6 files reviewed, 6 unresolved discussions (waiting on @kungfucraig, @Marco-Premier, and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/client/RabbitMQClient.kt` line 137 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1873#-O9wESFpDkPx2oyQGxta:-OA2rF89DWZhRwI5SBvM:bwioq3g) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3b302d55a4c2d7938f9d5e4230f08842d40eef96/src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/client/RabbitMQClient.kt#L137)):*

Is the intent to just drop all of the output?
</blockquote></details>

that makes sense to me
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/client/v1alpha/RabbitMQHandler.kt` line 81 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1873#-O9qLj7Z-iR09iuAVIan:-OA2rPp12sB6IYA95klS:b-yvv2b0) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/971beea91aae0ca71fdfc5757818dcb63d39b70a/src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/client/v1alpha/RabbitMQHandler.kt#L81)):*

I understand. This needs to be documented in the code, especially the reasons for why such a thing is necessary.

Would it make sense to re-provisioned the VM any time it would do work for more than a single EDP?
</blockquote></details>

I'm okay with re-provisioning if we don't think it will cause performance/cost issues

kungfucraig (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1873)* status: 5 of 6 files reviewed, 6 unresolved discussions (waiting on @Marco-Premier, @SanjayVas, and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/client/RabbitMQClient.kt` line 137 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1873#-O9wESFpDkPx2oyQGxta:-OA38e0JDXE_a9gUsdXy:b-fl8tmk) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3b302d55a4c2d7938f9d5e4230f08842d40eef96/src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/client/RabbitMQClient.kt#L137)):*

that makes sense to me
</blockquote></details>

Really, shouldn't we log the output so that debugging is possible? I think it's reasonable for these TEE apps to log as long as the logs are not leaking sensitive information.

kungfucraig (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1873)* status: 5 of 6 files reviewed, 6 unresolved discussions (waiting on @Marco-Premier, @SanjayVas, and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/client/v1alpha/RabbitMQHandler.kt` line 81 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1873#-O9qLj7Z-iR09iuAVIan:-OA38oBpE5UAIxIJORwe:b-a7lrea) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/971beea91aae0ca71fdfc5757818dcb63d39b70a/src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/client/v1alpha/RabbitMQHandler.kt#L81)):*

I'm okay with re-provisioning if we don't think it will cause performance/cost issues
</blockquote></details>

Let's just go with reuse for now and revisit. It'll keep things simpler.

Marco-Premier:
Move code inside common-jvm

PR #1886, opened 2024-10-29
feat(edp-simulator)!: Specify EventGroup metadata via command-line option

### LINKED ISSUE #1880: Allow use of custom metadata descriptors in EDP simulator
# Background

The primary intent of the EDP simulator is to facilitate Halo integration tests. It is however also useful for operators of Halo systems to use it for their own testing, so some functionality has been added to support this. The EDP simulator currently uses specific EventGroup metadata descriptors to control requisition fulfillment behavior w.r.t. how events are generated. The intent was that real EDPs could also be made to implement the same behavior to support conformance testing. With the advent of efforts such as HMSS, EDP libraries, and the EDP aggregator, the need for this kind of performance testing has diminished. Separately, Halo operators have indicated a desire to use the EDP simulator for end-to-end testing of other parts of their systems such as EventGroup search.

# Proposal

Allow operators to specify EventGroup metadata using any EventGroup metadata descriptor. This means that the metadata descriptor cannot be used to control event generation behavior. Instead, this behavior is solely controlled via configuration of the simulator itself.

SanjayVas:
EventGroup metadata is now supplied by the `--event-group-metadata` option instead of using the specs passed to the `--event-group-spec` option. The `--event-group-metadata-type-url` option is used to indicate the message type of the EventGroup metadata, which must be a known metadata type.

BREAKING CHANGE: The EDP simulator has a new required `--event-group-metadata` option.
BREAKING CHANGE: The EDP simulator now takes the `--known-event-group-metadata-type` option.

Closes #1880


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1886)

stevenwarejones (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1886)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @SanjayVas)

PR #1843, opened 2024-10-09
ControlPlane initial proto definitions.

Marco-Premier:

PR #286, opened 2021-09-22
Update consent-signaling-client version.

wangyaopw:
https://rally1.rallydev.com/#/604612930531d/userstories?detail=%2Ftask%2F609465361881

---
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/286)

PR #1890, opened 2024-10-31
Controlplane api

Marco-Premier:
Add WorkItemsService to handle the creation of work items through RabbitMQ queues. 
The service provides queue validation, persistent message publishing, and proper 
channel recovery after failures.


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1890)

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1890)* status: 6 of 7 files reviewed, 3 unresolved discussions (waiting on @kungfucraig, @Marco-Premier, and @SanjayVas)

___
*[`MODULE.bazel` line 236 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1890#-OAY3rL2AYMK1eq-pzHH:-OAY3rL2AYMK1eq-pzHI:b-f5pp0k) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d04d88ea6f0c0b4bf2537e0d19886a377e138239/MODULE.bazel#L236)):*
> ```Starlark
> 
>         # RabbitMQ
>         # TODO(@marcopremier): Remove this once common-jvm PR #286 is merged
> ```

can you instead reference your common-jvm commit in this pr temporarily?
___
*[`imports/java/com/rabbitmq/client/BUILD.bazel` line 1 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1890#-OAY3z7P1CF6W5YVlpTq:-OAY3z7P1CF6W5YVlpTr:b-7ropln) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d04d88ea6f0c0b4bf2537e0d19886a377e138239/imports/java/com/rabbitmq/client/BUILD.bazel#L1)):*
> ```Starlark
> # TODO(@marcopremier): Remove this once common-jvm PR #286 is merged
> ```

ditto
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/WorkItemsService.kt` line 84 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1890#-OAY4GPr3ZCBFwuZzsKO:-OAY4GPr3ZCBFwuZzsKP:b-pf9rde) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d04d88ea6f0c0b4bf2537e0d19886a377e138239/src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/WorkItemsService.kt#L84)):*
> ```Kotlin
>       throw StatusException(
>         Status.PERMISSION_DENIED.withDescription("Queue '$queueName' does not exist")
> ```

is this a NOT_FOUND error? It feels like there could be more reasons than just the queue name not existing that would cause an error.

kungfucraig (changes requested):
Please follow the conventional commit guidelines.

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1890#-:-OAYBew54449l-1ZBJL2:bx2s3p5)* status: 6 of 7 files reviewed, 9 unresolved discussions (waiting on @Marco-Premier and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/WorkItemsService.kt` line 42 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1890#-OAYCJVu2r35ophXvFKd:-OAYCJVu2r35ophXvFKe:bk6jm3) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d04d88ea6f0c0b4bf2537e0d19886a377e138239/src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/WorkItemsService.kt#L42)):*
> ```Kotlin
>     }
> 
>   private lateinit var connection: Connection
> ```

Do these need to be lateinit? They are initialized in the init block.
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/WorkItemsService.kt` line 55 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1890#-OAYEzqP03b9lfO2sioY:-OAYEzqP03b9lfO2sioZ:b45xue5) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d04d88ea6f0c0b4bf2537e0d19886a377e138239/src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/WorkItemsService.kt#L55)):*
> ```Kotlin
>       throw StatusException(
>         Status.UNAVAILABLE.withDescription("Failed to connect to RabbitMQ: ${e.message}")
> ```

Write the host, port, and user name to the message?
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/WorkItemsService.kt` line 76 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1890#-OAYFXim4dYdqZyJJPOf:-OAYFXim4dYdqZyJJPOg:b-fsqd1m) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d04d88ea6f0c0b4bf2537e0d19886a377e138239/src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/WorkItemsService.kt#L76)):*
> ```Kotlin
> 
>     val workItem = request.workItem
>     val queueName = workItem.queue
> ```

Check that this is a valid queue before using it. Or is this what the call to the channel below is doing? Seems like relying on rabbit config for this may not be ideal.
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/WorkItemsService.kt` line 90 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1890#-OAYFCccC8pV04-it4P5:-OAYFCcd4jed4kXZMldv:bz19bi2) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d04d88ea6f0c0b4bf2537e0d19886a377e138239/src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/WorkItemsService.kt#L90)):*
> ```Kotlin
>     try {
>       // Makes the message persistent.
>       val props = AMQP.BasicProperties.Builder().deliveryMode(2).build()
> ```

Please use a constant instead of "2"
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/WorkItemsService.kt` line 95 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1890#-OAYGfNO0oEWiltWhFsi:-OAYGfNO0oEWiltWhFsj:b-fazxiu) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d04d88ea6f0c0b4bf2537e0d19886a377e138239/src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/WorkItemsService.kt#L95)):*
> ```Kotlin
>       throw StatusException(
>         Status.INTERNAL.withDescription("Failed to enqueue work item: ${e.message}")
> ```

Write the queue name to the message.
___
*[`src/test/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/WorkItemsServiceTest.kt` line 47 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1890#-OAYGmow0JuE2AwLbnCV:-OAYGmow0JuE2AwLbnCW:b-khy9co) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d04d88ea6f0c0b4bf2537e0d19886a377e138239/src/test/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/WorkItemsServiceTest.kt#L47)):*
> ```Kotlin
>   private lateinit var workItemsService: WorkItemsService
> 
>   @Before
> ```

Add a testing utility to common JVM to do all of these things. I think this is the third time I've seen this block of code, and would expect to see it many more times. :)

stevenwarejones (commented):
> Reviewed 9 of 14 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1890)* status: 9 of 14 files reviewed, 9 unresolved discussions (waiting on @Marco-Premier and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/GooglePubSubWorkItemsService.kt` line 31 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1890#-OCeYLLFAlJ97XM0QZWp:-OCeYLLFAlJ97XM0QZWq:b-7u8hwt) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1c57ac5d7f08ae6e9493c9907e82fba4147adefb/src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/GooglePubSubWorkItemsService.kt#L31)):*
> ```Kotlin
> class GooglePubSubWorkItemsService(
>   private val googlePubSubClient: GooglePubSubClient = DefaultGooglePubSubClient()
> //  private val rabbitMqHost: String,
> ```

delete these

stevenwarejones (changes requested):
> Reviewed 6 of 9 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1890)* status: 8 of 14 files reviewed, 9 unresolved discussions (waiting on @kungfucraig, @Marco-Premier, and @SanjayVas)

___
*[`src/main/proto/wfa/measurement/securecomputation/controlplane/v1alpha/work_item.proto` line 43 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1890#-OD6ibxx7gt8sBEzdTjW:-OD6ibxx7gt8sBEzdTjX:bsjhgnt) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/916824a91528ef9729e9fe0e6fe7650224bbf079/src/main/proto/wfa/measurement/securecomputation/controlplane/v1alpha/work_item.proto#L43)):*
> ```proto
>   // The queue into which this `WorkItem` must be enqueued.
>   // Available queues are specified in a configuration file.
>   string queue = 2 [
> ```

why is this a string now? - do you mean queue_resource_name?
___
*[`src/test/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/GooglePubSubWorkItemsServiceTest.kt` line 129 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1890#-OD6jVU5Dj1g-m7Y6Qb1:-OD6jVU5Dj1g-m7Y6Qb2:b-8dlkgv) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/916824a91528ef9729e9fe0e6fe7650224bbf079/src/test/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/GooglePubSubWorkItemsServiceTest.kt#L129)):*
> ```Kotlin
>   }
> 
>   @Test
> ```

can you add a test with multiple queues.... one subscriber per queue?

stevenwarejones (approved):
> Reviewed 2 of 14 files at r2, 2 of 9 files at r3, 2 of 3 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1890)* status: 12 of 14 files reviewed, 8 unresolved discussions (waiting on @kungfucraig, @Marco-Premier, and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/GooglePubSubWorkItemsService.kt` line 33 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1890#-OD9AYaiD51LBXhJ_grN:-OD9AYaiD51LBXhJ_grO:b-mwomxl) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e20e6185d5b0ac47c1d7749b76157c07606807cb/src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/GooglePubSubWorkItemsService.kt#L33)):*
> ```Kotlin
>   private val publisher: Publisher = Publisher(projectId, googlePubSubClient)
> 
>   override suspend fun createWorkItem(
> ```

could this function just be part of the abstract class?

stevenwarejones (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1890)* status: 12 of 14 files reviewed, 9 unresolved discussions (waiting on @kungfucraig, @Marco-Premier, and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/GooglePubSubWorkItemsService.kt` line 31 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1890#-ODIOWRk0zJPf9JHwhr5:-ODIOWRk0zJPf9JHwhr6:b-jvqkc4) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e20e6185d5b0ac47c1d7749b76157c07606807cb/src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/GooglePubSubWorkItemsService.kt#L31)):*
> ```Kotlin
> ) : WorkItemsCoroutineImplBase(), WorkItemsService {
> 
>   private val publisher: Publisher = Publisher(projectId, googlePubSubClient)
> ```

publisher<Message> based on the update to common-jvm

Marco-Premier (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1890)* status: 8 of 14 files reviewed, 8 unresolved discussions (waiting on @kungfucraig, @SanjayVas, and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/GooglePubSubWorkItemsService.kt` line 33 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1890#-OD9AYaiD51LBXhJ_grN:-ODgLwdNC-_f0R0FJIl6:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e20e6185d5b0ac47c1d7749b76157c07606807cb/src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/GooglePubSubWorkItemsService.kt#L33)):*

could this function just be part of the abstract class?
</blockquote></details>

Done.
___
*[`src/main/proto/wfa/measurement/securecomputation/controlplane/v1alpha/work_item.proto` line 43 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1890#-OD6ibxx7gt8sBEzdTjW:-OD79rbDEr14p42eb_Ru:bmwykjd) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/916824a91528ef9729e9fe0e6fe7650224bbf079/src/main/proto/wfa/measurement/securecomputation/controlplane/v1alpha/work_item.proto#L43)):*

why is this a string now? - do you mean queue_resource_name?
</blockquote></details>

I changes it to add new fields to the `Queue` message but those are actually not needed so I rollback to the initial implementation: https://github.com/world-federation-of-advertisers/cross-media-measurement/pull/1843

AIP: https://google.aip.dev/122
___
*[`src/test/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/GooglePubSubWorkItemsServiceTest.kt` line 129 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1890#-OD6jVU5Dj1g-m7Y6Qb1:-OD7PpTo8EIR9egTxQe7:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/916824a91528ef9729e9fe0e6fe7650224bbf079/src/test/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/GooglePubSubWorkItemsServiceTest.kt#L129)):*

can you add a test with multiple queues.... one subscriber per queue?
</blockquote></details>

Done.

Marco-Premier (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1890)* status: 7 of 14 files reviewed, 8 unresolved discussions (waiting on @kungfucraig, @SanjayVas, and @stevenwarejones)

___
*[`src/test/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/WorkItemsServiceTest.kt` line 47 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1890#-OAYGmow0JuE2AwLbnCV:-ODgPvJm61MTS6RonfjZ:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d04d88ea6f0c0b4bf2537e0d19886a377e138239/src/test/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/WorkItemsServiceTest.kt#L47)):*

Add a testing utility to common JVM to do all of these things. I think this is the third time I've seen this block of code, and would expect to see it many more times. :)
</blockquote></details>

Done.

kungfucraig (approved):
:lgtm:

> Reviewed 2 of 14 files at r2, 2 of 9 files at r3, 1 of 3 files at r4, 5 of 5 files at r5, 1 of 1 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1890#-:-OEUlpCn9MKGVW4iQGkF:bnfp4nl)* status: 13 of 14 files reviewed, 2 unresolved discussions (waiting on @SanjayVas and @stevenwarejones)

SanjayVas (changes requested):
> Reviewed 2 of 9 files at r3, 5 of 6 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1890)* status: 13 of 14 files reviewed, 6 unresolved discussions (waiting on @kungfucraig, @Marco-Premier, and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/WorkItemsService.kt` line 28 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1890#-OG1gYH0DFep9Sdtp_6V:-OG1gYH0DFep9Sdtp_6W:b-vtnn26) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/64ee53d3004751ba17a7ce782ac71aa9bd07d8e9/src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/WorkItemsService.kt#L28)):*
If this just publishes a message, how would you go about implementing GetWorkItem? If you don't intend to ever support Get, perhaps WorkItem shouldn't be a resource type.

From [AIP-131](https://google.aip.dev/131):

> APIs **must** provide a get method for resources. The purpose of the get method is to return data from a single resource.
___
*[`src/main/proto/wfa/measurement/securecomputation/controlplane/v1alpha/work_item.proto` line 22 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1890#-OG1g89I4rHWtHKQAp8C:-OG1g89I4rHWtHKQAp8D:b-fe2y5h) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/64ee53d3004751ba17a7ce782ac71aa9bd07d8e9/src/main/proto/wfa/measurement/securecomputation/controlplane/v1alpha/work_item.proto#L22)):*
> ```proto
> import "google/api/field_behavior.proto";
> import "google/api/resource.proto";
> import "wfa/measurement/securecomputation/controlplane/v1alpha/queue.proto";
> ```

This appears to be unused, since there's just a resource reference but not a Queue message instance.
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/WorkItemsService.kt` line 26 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1890#-OG1gMGf7R6R51cdzBXI:-OG1gMGg4PJ6j8s4qhO4:b-5dqnqs) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/64ee53d3004751ba17a7ce782ac71aa9bd07d8e9/src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/WorkItemsService.kt#L26)):*
> ```Kotlin
> abstract class WorkItemsService: WorkItemsCoroutineImplBase() {
> 
>   abstract fun publishMessage(queueName: String, message: Message)
> ```

Shouldn't this be a suspend function?
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/WorkItemsService.kt` line 39 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1890#-OG1hQTy4_RqOnEZjgIi:-OG1hQTy4_RqOnEZjgIj:b-1gt6ov) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/64ee53d3004751ba17a7ce782ac71aa9bd07d8e9/src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/WorkItemsService.kt#L39)):*
> ```Kotlin
>       throw when {
>         e.message?.contains("Topic id: $topicId does not exist") == true -> {
> ```

You should never programmatically inspect an exception message. Use specific exception types instead.

Marco-Premier (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1890)* status: 11 of 14 files reviewed, 4 unresolved discussions (waiting on @kungfucraig, @SanjayVas, and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/WorkItemsService.kt` line 26 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1890#-OG1gMGf7R6R51cdzBXI:-OG5YX0M2aysq3y1iZbH:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/64ee53d3004751ba17a7ce782ac71aa9bd07d8e9/src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/WorkItemsService.kt#L26)):*

Shouldn't this be a suspend function?
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/WorkItemsService.kt` line 28 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1890#-OG1gYH0DFep9Sdtp_6V:-OG5mE81DDPswSFfPEeq:bz6eoz0) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/64ee53d3004751ba17a7ce782ac71aa9bd07d8e9/src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/WorkItemsService.kt#L28)):*

If this just publishes a message, how would you go about implementing GetWorkItem? If you don't intend to ever support Get, perhaps WorkItem shouldn't be a resource type.

From [AIP-131](https://google.aip.dev/131):

> APIs **must** provide a get method for resources. The purpose of the get method is to return data from a single resource.
</blockquote></details>

Agreed. we may have persistence in the future but for the moment `WorkItem` is not a resource type anymore.
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/WorkItemsService.kt` line 39 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1890#-OG1hQTy4_RqOnEZjgIi:-OG5XzT98BBSalSaJwt5:bs7wumx) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/64ee53d3004751ba17a7ce782ac71aa9bd07d8e9/src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/WorkItemsService.kt#L39)):*

You should never programmatically inspect an exception message. Use specific exception types instead.
</blockquote></details>

Done. 
___
*[`src/main/proto/wfa/measurement/securecomputation/controlplane/v1alpha/work_item.proto` line 22 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1890#-OG1g89I4rHWtHKQAp8C:-OG5O9de9qctidbDjX40:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/64ee53d3004751ba17a7ce782ac71aa9bd07d8e9/src/main/proto/wfa/measurement/securecomputation/controlplane/v1alpha/work_item.proto#L22)):*

This appears to be unused, since there's just a resource reference but not a Queue message instance.
</blockquote></details>

Done.

stevenwarejones (approved):
> Reviewed 1 of 5 files at r5, 3 of 3 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1890)* status: all files reviewed, 4 unresolved discussions (waiting on @kungfucraig and @SanjayVas)

kungfucraig (approved):
> Reviewed 1 of 9 files at r3, 6 of 6 files at r7, 3 of 3 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1890)* status: all files reviewed, 4 unresolved discussions (waiting on @SanjayVas)

SanjayVas (changes requested):
> Reviewed 1 of 9 files at r3, 1 of 6 files at r7, 2 of 3 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1890)* status: all files reviewed, 1 unresolved discussion (waiting on @Marco-Premier)

___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/WorkItemsService.kt` line 28 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1890#-OG1gYH0DFep9Sdtp_6V:-OGX9e9H5O_UuPmvVjtN:b-f63y5v) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/64ee53d3004751ba17a7ce782ac71aa9bd07d8e9/src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/WorkItemsService.kt#L28)):*

Agreed. we may have persistence in the future but for the moment `WorkItem` is not a resource type anymore.
</blockquote></details>

Not having WorkItem be a resource type has broader implications, given that we're building a resource-oriented API. e.g. having a Create method for WorkItem no longer makes sense if there's no WorkItem resource.

I'm guessing the resource we're operating on is Queue instead then. If that's the case, maybe a custom method on the Queue resource e.g. EnqueueWork on a Queues service.

Marco-Premier (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1890)* status: 7 of 15 files reviewed, 1 unresolved discussion (waiting on @kungfucraig, @SanjayVas, and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/WorkItemsService.kt` line 28 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1890#-OG1gYH0DFep9Sdtp_6V:-OG_q5zv9DjWxthFzUoI:b-8vnzdi) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/64ee53d3004751ba17a7ce782ac71aa9bd07d8e9/src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/WorkItemsService.kt#L28)):*

Hrmmm so maybe WorkItem should be a resource? 

Otherwise, these things are just going to get pushed into a pubsub and lost in the ether. It will be pretty much impossible to debug.

At the risk of designing on the fly:
1. Put the work item in spanner
2. Hook up a spanner change listener for the pubsub part https://cloud.google.com/dataflow/docs/guides/templates/provided/cloud-spanner-change-streams-to-pubsub
</blockquote></details>

Just read your comment Craig and published my updates already.
I agreed that messages should be persistent, also for letting  the system the ability to retrieve failures or request statuses.

Logging published messages may not be enough (what if they fails, what if they end up in the dead letter queue etc...). I thought I had shared that persistence would have come in phase 2. For debugging purposes may be faster and more insightful to use Google Cloud Logging instead?

what you think?

kungfucraig (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1890)* status: 7 of 15 files reviewed, 1 unresolved discussion (waiting on @SanjayVas and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/WorkItemsService.kt` line 28 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1890#-OG1gYH0DFep9Sdtp_6V:-OGa5ObB7qdFb6T3niLo:bqmq76m) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/64ee53d3004751ba17a7ce782ac71aa9bd07d8e9/src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/WorkItemsService.kt#L28)):*

Just read your comment Craig and published my updates already.
I agreed that messages should be persistent, also for letting  the system the ability to retrieve failures or request statuses.

Logging published messages may not be enough (what if they fails, what if they end up in the dead letter queue etc...). I thought I had shared that persistence would have come in phase 2. For debugging purposes may be faster and more insightful to use Google Cloud Logging instead?

what you think?
</blockquote></details>

I think what's in the PR now can work, but it's probably not as good as the other approach.

The advantage to having the WorkItems service is that the workers can actually go back to the Workitem and add additional metadata about the task execution. I suspect we will find lots of reasons to want this kind of thing.

But it seems like this could be added later without too much fuss. The Enqueue method could be redone as pass through to WorkItem.Create and you could then provide WorkItem.Get and WorkItem.Update.

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1890)* status: 7 of 15 files reviewed, 1 unresolved discussion (waiting on @kungfucraig, @Marco-Premier, and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/WorkItemsService.kt` line 28 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1890#-OG1gYH0DFep9Sdtp_6V:-OGaAm4JCqWiuV2zLHxT:bnnxsd7) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/64ee53d3004751ba17a7ce782ac71aa9bd07d8e9/src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/WorkItemsService.kt#L28)):*
If we plan on having a WorkItem resource type at any point, then just have it now but have a plan on how to implement Get. If you want the ability to support operations like getting status or cancelling a WorkItem, then you'll need this.

> I thought I had shared that persistence would have come in phase 2.

I had not assumed this at all, and I had expected persistence to happen up front as soon as it was originally proposed that WorkItem would be a resource.

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1890)* status: 7 of 15 files reviewed, 1 unresolved discussion (waiting on @kungfucraig, @Marco-Premier, and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/WorkItemsService.kt` line 28 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1890#-OG1gYH0DFep9Sdtp_6V:-OGaDxaN3mnxosI-aSry:b-67nwh4) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/64ee53d3004751ba17a7ce782ac71aa9bd07d8e9/src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/WorkItemsService.kt#L28)):*

If we plan on having a WorkItem resource type at any point, then just have it now but have a plan on how to implement Get. If you want the ability to support operations like getting status or cancelling a WorkItem, then you'll need this.

> I thought I had shared that persistence would have come in phase 2.

I had not assumed this at all, and I had expected persistence to happen up front as soon as it was originally proposed that WorkItem would be a resource.
</blockquote></details>

Sorry, sent early.

The reasoning for this is that this is *not* simply a wrapper around a message queue. This is a *work queue*, where part of the implementation happens to be a message queue. You still need to build all the standard CRUD as with any resource-oriented API.

Marco-Premier (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1890)* status: 7 of 15 files reviewed, 1 unresolved discussion (waiting on @kungfucraig, @SanjayVas, and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/WorkItemsService.kt` line 28 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1890#-OG1gYH0DFep9Sdtp_6V:-OGdNXnp5CFpfmAyNRA0:bmio63r) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/64ee53d3004751ba17a7ce782ac71aa9bd07d8e9/src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/WorkItemsService.kt#L28)):*
> I had not assumed this at all, and I had expected persistence to happen up front as soon as it was originally proposed that WorkItem would be a resource.

I'll make sure to be more explicit next time.

I'll reset the branch. `WorkItem` is indeed a resource and will implement proper CRUD operations on following PR.
I'll experiment to understand what's the best way monitor message's status and keep track of them.

SanjayVas (changes requested):
> Reviewed 2 of 8 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1890)* status: 10 of 15 files reviewed, 2 unresolved discussions (waiting on @kungfucraig, @Marco-Premier, and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/WorkItemsService.kt` line 28 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1890#-OG1gYH0DFep9Sdtp_6V:-OGfk0p60pCmyTHEPuzh:b-jq1g7j) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/64ee53d3004751ba17a7ce782ac71aa9bd07d8e9/src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/WorkItemsService.kt#L28)):*

> I had not assumed this at all, and I had expected persistence to happen up front as soon as it was originally proposed that WorkItem would be a resource.

I'll make sure to be more explicit next time.

I'll reset the branch. `WorkItem` is indeed a resource and will implement proper CRUD operations on following PR.
I'll experiment to understand what's the best way monitor message's status and keep track of them.
</blockquote></details>

There's some design questions here. e.g. what's the lifetime of a WorkItem? I assume it's different than that of the resulting Pub/Sub message.

For the moment, I think you can consider the WorkItem to be its own concept. Creating a WorkItem *results* in a message being published as part of the implementation of how the item is processed, but the message isn't the WorkItem.
___
*[`src/main/proto/wfa/measurement/securecomputation/controlplane/v1alpha/work_item.proto` line 22 at r10](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1890#-OGfjfeZCxNRmd2kWjT1:-OGfjfeZCxNRmd2kWjT2:b-yrumkl) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/64ee53d3004751ba17a7ce782ac71aa9bd07d8e9/src/main/proto/wfa/measurement/securecomputation/controlplane/v1alpha/work_item.proto#L22)):*
> ```proto
> import "google/api/field_behavior.proto";
> import "google/api/resource.proto";
> import "wfa/measurement/securecomputation/controlplane/v1alpha/queue.proto";
> ```

Unused

Marco-Premier (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1890)* status: 10 of 15 files reviewed, 2 unresolved discussions (waiting on @kungfucraig, @SanjayVas, and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/WorkItemsService.kt` line 28 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1890#-OG1gYH0DFep9Sdtp_6V:-OGkQAtL9uVqj388vrtP:bk59u2h) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/64ee53d3004751ba17a7ce782ac71aa9bd07d8e9/src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/WorkItemsService.kt#L28)):*

There's some design questions here. e.g. what's the lifetime of a WorkItem? I assume it's different than that of the resulting Pub/Sub message.

For the moment, I think you can consider the WorkItem to be its own concept. Creating a WorkItem *results* in a message being published as part of the implementation of how the item is processed, but the message isn't the WorkItem.
</blockquote></details>

I'm not sure I follow @SanjayVas . 

The goal is to have `WorkItem` as a persistent resource with Create / Get operations. A `WorkItem` naturally translate into a pub/sub message or better pub/sub is the "transport layer" of a `WorkItem` from which we can get information about its status. Pub/Sub infra also defines the following:
1. for how long a message can wait once published before being dispatched (default 7 days)
2. the same message cannot be delivered twice until the first has been nacked or ack_deadline expires

I'd like to leave the `WorkItem` as a resource as it is in this PR to move things forward and add the `Get` operation on a following PR.
Does it make sense. I would discourage to create a design now that is going to be soon replaced.

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1890)* status: 10 of 15 files reviewed, 2 unresolved discussions (waiting on @kungfucraig, @Marco-Premier, and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/WorkItemsService.kt` line 28 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1890#-OG1gYH0DFep9Sdtp_6V:-OGkYhqh6b54IMR0XLoV:b-dyc295) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/64ee53d3004751ba17a7ce782ac71aa9bd07d8e9/src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/WorkItemsService.kt#L28)):*
> A `WorkItem` naturally translate into a pub/sub message or better pub/sub is the "transport layer" of a `WorkItem` from which we can get information about its status.

This only works if a `WorkItem` is guaranteed to have the same lifetime as a message and that you actually have a way to get the message state. Otherwise you should treat it as a separate concept as I mentioned above. This means e.g. persisting the `WorkItem` state in a database and keeping it in sync with that of the message processing.

What I envisioned is something like the following:

* Creating a `WorkItem` results in a PubSub message being sent with the `WorkItem` ID in it.
* When ACKing the message, also update the `WorkItem` state (e.g. call `UpdateWorkItem`)
  * This means needing to handle the case where the message gets ACKed but the WorkItem update fails (or vice-versa if you update the WorkItem before ACKing).

Marco-Premier (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1890)* status: 3 of 15 files reviewed, 2 unresolved discussions (waiting on @kungfucraig, @SanjayVas, and @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/securecomputation/controlplane/v1alpha/work_item.proto` line 22 at r10](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1890#-OGfjfeZCxNRmd2kWjT1:-OHDGvgH8O-NheoPhNpE:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/64ee53d3004751ba17a7ce782ac71aa9bd07d8e9/src/main/proto/wfa/measurement/securecomputation/controlplane/v1alpha/work_item.proto#L22)):*

Unused
</blockquote></details>

Done.

SanjayVas (changes requested):
> Reviewed 4 of 8 files at r10, 7 of 7 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1890)* status: 14 of 15 files reviewed, 3 unresolved discussions (waiting on @kungfucraig, @Marco-Premier, and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/GooglePubSubWorkItemsService.kt` line 24 at r11](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1890#-OHJTOT5AF4fkB3f-aVr:-OHJTOT5AF4fkB3f-aVs:b-ext1l4) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/fa5e3323e0d28941438344681b022506e102d341/src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/GooglePubSubWorkItemsService.kt#L24)):*
> ```Kotlin
> import kotlinx.coroutines.runBlocking
> 
> class GooglePubSubWorkItemsService(
> ```

nit The convention is for cloud-specific impls to go under a deploy package for the cloud provider. I'm leaving this as a nit since I know this will be replaced where the pubsub message is sent in response to a Spanner change stream event.
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/GooglePubSubWorkItemsService.kt` line 32 at r11](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1890#-OHJT3xgEGtR5l7xFJ_C:-OHJT3xgEGtR5l7xFJ_D:b-9r4jvm) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/fa5e3323e0d28941438344681b022506e102d341/src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/GooglePubSubWorkItemsService.kt#L32)):*
> ```Kotlin
> 
>   override fun publishMessage(queueName: String, message: Message) {
>     runBlocking {
> ```

Don't use runBlocking - make the function suspending instead.

Marco-Premier (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1890)* status: 11 of 19 files reviewed, 3 unresolved discussions (waiting on @kungfucraig, @SanjayVas, and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/GooglePubSubWorkItemsService.kt` line 24 at r11](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1890#-OHJTOT5AF4fkB3f-aVr:-OHNYC-U2jnlBqkCxaqO:bzbr002) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/fa5e3323e0d28941438344681b022506e102d341/src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/GooglePubSubWorkItemsService.kt#L24)):*

nit The convention is for cloud-specific impls to go under a deploy package for the cloud provider. I'm leaving this as a nit since I know this will be replaced where the pubsub message is sent in response to a Spanner change stream event.
</blockquote></details>

will remind this, thanks. I leave it here for the moment as this class will most likely replaced by another that stores data into the spanner DB
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/GooglePubSubWorkItemsService.kt` line 32 at r11](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1890#-OHJT3xgEGtR5l7xFJ_C:-OHNZVeH6gggZSZoHf8u:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/fa5e3323e0d28941438344681b022506e102d341/src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/GooglePubSubWorkItemsService.kt#L32)):*

Don't use runBlocking - make the function suspending instead.
</blockquote></details>

Done.

SanjayVas (changes requested):
> Reviewed 1 of 12 files at r10, 7 of 7 files at r12, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1890)* status: all files reviewed, 4 unresolved discussions (waiting on @kungfucraig, @Marco-Premier, and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/WorkItemsService.kt` line 28 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1890#-OG1gYH0DFep9Sdtp_6V:-OHO22bz7EoTbp2QF4KJ:b-2wwkn6) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/64ee53d3004751ba17a7ce782ac71aa9bd07d8e9/src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/WorkItemsService.kt#L28)):*

> A `WorkItem` naturally translate into a pub/sub message or better pub/sub is the "transport layer" of a `WorkItem` from which we can get information about its status.

This only works if a `WorkItem` is guaranteed to have the same lifetime as a message and that you actually have a way to get the message state. Otherwise you should treat it as a separate concept as I mentioned above. This means e.g. persisting the `WorkItem` state in a database and keeping it in sync with that of the message processing.

What I envisioned is something like the following:

* Creating a `WorkItem` results in a PubSub message being sent with the `WorkItem` ID in it.
* When ACKing the message, also update the `WorkItem` state (e.g. call `UpdateWorkItem`)
  * This means needing to handle the case where the message gets ACKed but the WorkItem update fails (or vice-versa if you update the WorkItem before ACKing).
</blockquote></details>

Per the updated design doc, the intent is to fast follow this with a refactor that persists WorkItems in a DB which then triggers message publishing.
___
*[`src/test/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/GooglePubSubWorkItemsServiceTest.kt` line 24 at r12](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1890#-OHO2aOu-dqsPHk3Te4t:-OHO2aOu-dqsPHk3Te4u:b-pcf9eu) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/42766114e2924955300f3ae0a6c0ad442bd0cd45/src/test/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/GooglePubSubWorkItemsServiceTest.kt#L24)):*
> ```Kotlin
> import io.grpc.Status
> import io.grpc.StatusRuntimeException
> import java.util.*
> ```

Wildcard imports are forbidden
___
*[`src/test/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/GooglePubSubWorkItemsServiceTest.kt` line 76 at r12](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1890#-OHO31JcCOXs4D8jkXNw:-OHO31JcCOXs4D8jkXNx:b-7f1fhq) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/42766114e2924955300f3ae0a6c0ad442bd0cd45/src/test/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/GooglePubSubWorkItemsServiceTest.kt#L76)):*
> ```Kotlin
>     val deferred = CompletableDeferred<String>()
> 
>     withTimeout(1000) {
> ```

Use JUnit for test timeouts (e.g. specifying the timeout to the `@Test` annotation)
___
*[`src/test/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/GooglePubSubWorkItemsServiceTest.kt` line 115 at r12](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1890#-OHO3r0UAtgR4AeMZf5L:-OHO3r0UAtgR4AeMZf5M:b-hb3nca) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/42766114e2924955300f3ae0a6c0ad442bd0cd45/src/test/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/GooglePubSubWorkItemsServiceTest.kt#L115)):*
> ```Kotlin
> 
>   @Test
>   fun `test sending multiple messages in sequence`() {
> ```

nit: if the entire body is a runBlocking call, the convention is to use a function expression like you're doing in the test method above this one
___
*[`src/test/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/GooglePubSubWorkItemsServiceTest.kt` line 243 at r12](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1890#-OHO3fAZ-R9JKEGlqtoR:-OHO3fAZ-R9JKEGlqtoS:b-yey1xb) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/42766114e2924955300f3ae0a6c0ad442bd0cd45/src/test/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/GooglePubSubWorkItemsServiceTest.kt#L243)):*
> ```Kotlin
>   }
>   private fun createTestRequest(workItemId: String, workItemParams: Any, topicId: String): CreateWorkItemRequest {
>     val workItem = WorkItem.newBuilder()
> ```

Use Kotlin DSL builders for protos rather than Java builders.

SanjayVas (approved):
> Reviewed 8 of 8 files at r13, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1890)* status: all files reviewed, 2 unresolved discussions (waiting on @kungfucraig, @Marco-Premier, and @stevenwarejones)

___
*[`src/test/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/GooglePubSubWorkItemsServiceTest.kt` line 243 at r12](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1890#-OHO3fAZ-R9JKEGlqtoR:-OHhzZJw7isM9q07m9HE:b-6pqueq) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/42766114e2924955300f3ae0a6c0ad442bd0cd45/src/test/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/GooglePubSubWorkItemsServiceTest.kt#L243)):*

Done. Won't work the same with createWorkItemRequest as work_item_params is immutable and would trigger an error, eg. `error: val cannot be reassigned`
</blockquote></details>

It does work. You're probably just specifying the wrong `workItem` since there are two such symbols in scope. You need to disambiguate between the two. You're in fact already doing the same when building the `WorkItem` message.

```kotlin
    return createWorkItemRequest {
      this.workItemId = workItemId
      this.workItem = workItem
    }
```
___
*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1890#-OHhz4GxCePYdHRO_-mn:-OHhz4GxCePYdHRO_-mo:b1m2jg8) (no related file):*
Calling out again to follow the [dev standards](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/main/docs/dev-standards.md#pull-request-description) for PR descriptions, including conventional commits. The description is also currently out-of-date as it references RabbitMQ.

I'd personally suggest a title like "feat: Add PubSub implementation of WorkItems service"

stevenwarejones (approved):
> Reviewed 2 of 12 files at r10, 6 of 8 files at r13, 10 of 10 files at r14, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1890)* status: all files reviewed, 2 unresolved discussions (waiting on @kungfucraig and @Marco-Premier)

Marco-Premier (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1890)* status: 12 of 19 files reviewed, 2 unresolved discussions (waiting on @kungfucraig, @SanjayVas, and @stevenwarejones)

___
*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1890#-OHhz4GxCePYdHRO_-mn:-OHruMeXDKE7nu6ww3OA:b-896fix) (no related file):*

Calling out again to follow the [dev standards](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/main/docs/dev-standards.md#pull-request-description) for PR descriptions, including conventional commits. The description is also currently out-of-date as it references RabbitMQ.

I'd personally suggest a title like "feat: Add PubSub implementation of WorkItems service"
</blockquote></details>

Done.
___
*[`src/test/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/GooglePubSubWorkItemsServiceTest.kt` line 243 at r12](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1890#-OHO3fAZ-R9JKEGlqtoR:-OHruVN49I8H7pid9RvD:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/42766114e2924955300f3ae0a6c0ad442bd0cd45/src/test/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/GooglePubSubWorkItemsServiceTest.kt#L243)):*

It does work. You're probably just specifying the wrong `workItem` since there are two such symbols in scope. You need to disambiguate between the two. You're in fact already doing the same when building the `WorkItem` message.

```kotlin
    return createWorkItemRequest {
      this.workItemId = workItemId
      this.workItem = workItem
    }
```
</blockquote></details>

Done.

Marco-Premier (commented):
> Dismissed @SanjayVas from 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1890)* status: 12 of 19 files reviewed, all discussions resolved (waiting on @kungfucraig, @SanjayVas, and @stevenwarejones)

Marco-Premier (commented):
> Reviewed 2 of 12 files at r10, 10 of 10 files at r14, 7 of 7 files at r15, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1890)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @kungfucraig)

PR #1904, opened 2024-11-07
feat: Add BaseTeeApplication 

Marco-Premier:


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1904)

SanjayVas (changes requested):
> Reviewed 4 of 9 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1904)* status: 4 of 9 files reviewed, 6 unresolved discussions (waiting on @Marco-Premier and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt` line 48 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1904#-OB7-xYD9Gz2se_1WBQu:-OB7-xYD9Gz2se_1WBQv:b9vbdnt) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/92f8b9b71b10b71d1256094ceec279e942908e74/src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt#L48)):*
> ```Kotlin
>   private val queueName: String,
>   private val queueClient: QueueClient,
>   private val parser: (ByteArray) -> T,
> ```

Use the existing Parser interface, which offers more flexibility in how to deserialize.



_Suggestion:_
```Kotlin
Parser<T>
```
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt` line 49 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1904#-OB6yw3Q8kspDc2M0CoC:-OB6yw3Q8kspDc2M0CoD:bqlnd29) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/92f8b9b71b10b71d1256094ceec279e942908e74/src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt#L49)):*
> ```Kotlin
>   private val queueClient: QueueClient,
>   private val parser: (ByteArray) -> T,
>   private val blockingContext: CoroutineContext = Dispatchers.IO,
> ```

Is this actually used for blocking code? I thought we wrap all of that in the queue client.

If it is, annotate it with `@BlockingExecutor`. If not, annotate it with `@NonBlockingExecutor`, default it to Dispatchers.Default. In either case, you probably want to rename it and make it clear what it's used for.

That said, this may be irrelevant if you give this a primary run method.



_Code quote:_
```Kotlin
blockingContext
```
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt` line 52 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1904#-OB6zBx3Ca1ZVkuQ45nJ:-OB6zBx4DM511q7DkGAA:bs4nbou) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/92f8b9b71b10b71d1256094ceec279e942908e74/src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt#L52)):*
> ```Kotlin
> ) : AutoCloseable {
> 
>   private val scope = CoroutineScope(blockingContext + SupervisorJob())
> ```

If you really need supervisor behavior (rare), document exactly why you need it.



_Code quote:_
```Kotlin
SupervisorJob()
```
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt` line 54 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1904#-OB71cwT5DWRR_AjrocO:-OB71cwU2KKVlsRxgHak:bkk9cm1) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/92f8b9b71b10b71d1256094ceec279e942908e74/src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt#L54)):*
> ```Kotlin
>   private val scope = CoroutineScope(blockingContext + SupervisorJob())
> 
>   init {
> ```

Usually for the top-level "application" class called from `main`, you'd have a `run` method. In the case of coroutine applications, this is most likely to be a suspending function that suspends until the application ends. For a server or daemon, this is forever (until the application crashes or is terminated externally). There are multiple examples of this in the code base.

I believe if you do this, you won't need a separate coroutine scope or context. `main` will most likely just wrap this the call to `run` with `runBlocking`.
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt` line 59 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1904#-OB7-DZT1xpWwhHBGX5h:-OB7-DZT1xpWwhHBGX5i:b-y5qb3n) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/92f8b9b71b10b71d1256094ceec279e942908e74/src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt#L59)):*
> ```Kotlin
>         startListening()
>       } catch (e: Exception) {
>         logger.severe("Error starting to listen to queue: ${e.message}")
> ```

Log-and-throw is an anti-pattern. Do one or the other.
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt` line 71 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1904#-OB70Ihn1HNYFlAX-VFW:-OB70Iho6WQXfI_CSnP7:b17w088) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/92f8b9b71b10b71d1256094ceec279e942908e74/src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt#L71)):*
> ```Kotlin
>   private suspend fun startListening() {
>     try {
>       val messageChannel: ReceiveChannel<QueueClient.QueueMessage<ByteArray>> =
> ```

Using ByteArray defeats the purpose of having QueueMessage be generic. You'll want to push the parsing code into the queue client. The underlying message type will always be a protobuf, and therefore extend Message.



_Suggestion:_
```Kotlin
T
```

Marco-Premier (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1904)* status: 2 of 9 files reviewed, 6 unresolved discussions (waiting on @SanjayVas and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt` line 48 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1904#-OB7-xYD9Gz2se_1WBQu:-OBALGxO22ZEzV6k6Yni:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/92f8b9b71b10b71d1256094ceec279e942908e74/src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt#L48)):*

Use the existing Parser interface, which offers more flexibility in how to deserialize.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt` line 49 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1904#-OB6yw3Q8kspDc2M0CoC:-OBAdZzs5_Kdw5-Fv97S:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/92f8b9b71b10b71d1256094ceec279e942908e74/src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt#L49)):*

Is this actually used for blocking code? I thought we wrap all of that in the queue client.

If it is, annotate it with `@BlockingExecutor`. If not, annotate it with `@NonBlockingExecutor`, default it to Dispatchers.Default. In either case, you probably want to rename it and make it clear what it's used for.

That said, this may be irrelevant if you give this a primary run method.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt` line 52 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1904#-OB6zBx3Ca1ZVkuQ45nJ:-OBAdZTsCnU3qYQhU6He:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/92f8b9b71b10b71d1256094ceec279e942908e74/src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt#L52)):*

If you really need supervisor behavior (rare), document exactly why you need it.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt` line 54 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1904#-OB71cwT5DWRR_AjrocO:-OBAd_ZsDC0dAZHEU_Lh:b-9gny58) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/92f8b9b71b10b71d1256094ceec279e942908e74/src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt#L54)):*

Usually for the top-level "application" class called from `main`, you'd have a `run` method. In the case of coroutine applications, this is most likely to be a suspending function that suspends until the application ends. For a server or daemon, this is forever (until the application crashes or is terminated externally). There are multiple examples of this in the code base.

I believe if you do this, you won't need a separate coroutine scope or context. `main` will most likely just wrap this the call to `run` with `runBlocking`.
</blockquote></details>

I connected the dots, thanks. No need for blocking executor here then. Left it inside `QueueClient` classes only.
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt` line 59 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1904#-OB7-DZT1xpWwhHBGX5h:-OBAdl_42k7-hH91tlWw:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/92f8b9b71b10b71d1256094ceec279e942908e74/src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt#L59)):*

Log-and-throw is an anti-pattern. Do one or the other.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt` line 71 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1904#-OB70Ihn1HNYFlAX-VFW:-OBAxVBB-AY4B03kip4C:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/92f8b9b71b10b71d1256094ceec279e942908e74/src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt#L71)):*

Using ByteArray defeats the purpose of having QueueMessage be generic. You'll want to push the parsing code into the queue client. The underlying message type will always be a protobuf, and therefore extend Message.
</blockquote></details>

Done.

SanjayVas (changes requested):
> Reviewed 4 of 9 files at r1, 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1904)* status: all files reviewed, 3 unresolved discussions (waiting on @Marco-Premier and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt` line 36 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1904#-OBCGGK4CCx9VaVRoIcf:-OBCGGK4CCx9VaVRoIcg:b-1hvqqm) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/08932f76822c55b37563a9413a2a51b73be35d06/src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt#L36)):*
> ```Kotlin
>  * @param queueClient A client that manages connections and interactions with the queue.
>  * @param parser A `Parser` from `com.google.protobuf` used to parse raw `ByteArray` data from the
>  * queue into the desired message type [T], typically a Protobuf message.
> ```

nit: this is enforced here to always be a protobuf message



_Suggestion:_
```Kotlin
 * @param parser [Parser] used to parse serialized queue messages into [T] instances
```
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt` line 49 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1904#-OBCGtHL-ho0EBguMavu:-OBCGtHL-ho0EBguMavv:b-frxned) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/08932f76822c55b37563a9413a2a51b73be35d06/src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt#L49)):*
> ```Kotlin
>    * If an error occurs during the message flow, it is logged and handling continues.
>    */
>   suspend open fun startListening() {
> ```

Why would this be overridable and public? It should be a private method called from the single entry point to the application, i.e. a `suspend fun run()`
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt` line 59 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1904#-OBCH7-yETwHfFiAfTBt:-OBCH7-yETwHfFiAfTBu:b-s5c0ue) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/08932f76822c55b37563a9413a2a51b73be35d06/src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt#L59)):*
> ```Kotlin
>         .collect {
>           queueMessage -> processMessage(queueMessage)
>           println("~~~~~~~~~~~~~~~~~~~~~~~~ collecting message")
> ```

Clean up debugging prints. Use logger if needed.

mikkokotila:
@Marco-Premier as a small housekeeping matter, have a look through the PR titles in [here](https://github.com/world-federation-of-advertisers/cross-media-measurement/pulls?q=is%3Apr+is%3Aclosed). Good to always name the PRs following the same scheme. For additional context have a look through [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/#specification).

Marco-Premier (commented):
thanks

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1904#-:-OByfpUe5DZ7rBEKwvBY:b-k73z76)* status: 6 of 9 files reviewed, 2 unresolved discussions (waiting on @SanjayVas and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt` line 49 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1904#-OBCGtHL-ho0EBguMavu:-OByfPta3bRGrjaKTIXj:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/08932f76822c55b37563a9413a2a51b73be35d06/src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt#L49)):*

Why would this be overridable and public? It should be a private method called from the single entry point to the application, i.e. a `suspend fun run()`
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt` line 59 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1904#-OBCH7-yETwHfFiAfTBt:-OByPtonEtLgOZIqB56w:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/08932f76822c55b37563a9413a2a51b73be35d06/src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt#L59)):*

Clean up debugging prints. Use logger if needed.
</blockquote></details>

Done.

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1904)* status: all files reviewed, 3 unresolved discussions (waiting on @Marco-Premier and @stevenwarejones)

___
*[`src/test/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplicationTest.kt` line 72 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1904#-OC-ebl87ZQ3rK9u7BHH:-OC-ebl87ZQ3rK9u7BHI:bnmtmb0) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4f82708fb53e2a12f7c32fe77be87bb47a48000b/src/test/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplicationTest.kt#L72)):*
> ```Kotlin
>           .build()
> 
>       launch {
> ```

nit: I don't think this code doesn't need to be launched in a separate coroutine. You should be able to launch the sender and then have this code be top-level inside `runBlocking`.
___
*[`src/test/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplicationTest.kt` line 75 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1904#-OC-fajUDjCLBnSlOb_u:-OC-fajUDjCLBnSlOb_v:b-aca1ii) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4f82708fb53e2a12f7c32fe77be87bb47a48000b/src/test/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplicationTest.kt#L75)):*
> ```Kotlin
>         app.messageProcessed.await()
>         assertThat(app.processedMessages.contains(testWork)).isTrue()
>         inMemoryQueueClient.close()
> ```

nit: Your BaseTeeApplication implementation closes the queue client, meaning it "owns" the client. Therefore you don't need to close the queue client separately.
___
*[`src/test/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplicationTest.kt` line 80 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1904#-OC-fCdE5atqUwhLE0BC:-OC-fCdE5atqUwhLE0BD:b-h5u34i) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4f82708fb53e2a12f7c32fe77be87bb47a48000b/src/test/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplicationTest.kt#L80)):*
> ```Kotlin
> 
>       launch {
>         inMemoryQueueClient.sendMessage(testWork.toByteArray())
> ```

sendMessage should be taking a TestWork instance as the client should be handling the binary serialization/deserialization.

Marco-Premier (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1904)* status: 1 of 9 files reviewed, 1 unresolved discussion (waiting on @SanjayVas and @stevenwarejones)

___
*[`src/test/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplicationTest.kt` line 80 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1904#-OC-fCdE5atqUwhLE0BC:-OCdWEP1EGRuNVSsFFrv:bc32j6c) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4f82708fb53e2a12f7c32fe77be87bb47a48000b/src/test/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplicationTest.kt#L80)):*

sendMessage should be taking a TestWork instance as the client should be handling the binary serialization/deserialization.
</blockquote></details>

With the client you refer to the common-jvm `GooglePubSubClient`? I guess my question is: do we want to limit the PubSub client to only work with proto messages?

stevenwarejones (approved):
> Reviewed 1 of 9 files at r1, 3 of 8 files at r4, 5 of 5 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1904)* status: all files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

___
*[`src/test/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplicationTest.kt` line 80 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1904#-OC-fCdE5atqUwhLE0BC:-OCeZRJS7FWcjG7kf-7k:b-6pfki1) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4f82708fb53e2a12f7c32fe77be87bb47a48000b/src/test/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplicationTest.kt#L80)):*

With the client you refer to the common-jvm `GooglePubSubClient`? I guess my question is: do we want to limit the PubSub client to only work with proto messages?
</blockquote></details>

I think I'd prefer ByteString but I'm okay with limiting it to Message.

SanjayVas (changes requested):
> Reviewed 1 of 8 files at r4, 3 of 5 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1904)* status: all files reviewed, 1 unresolved discussion (waiting on @Marco-Premier and @stevenwarejones)

___
*[`src/test/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplicationTest.kt` line 80 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1904#-OC-fCdE5atqUwhLE0BC:-OCepCeSC-saR3Qci5yi:b-svpmj6) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4f82708fb53e2a12f7c32fe77be87bb47a48000b/src/test/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplicationTest.kt#L80)):*

I think I'd prefer ByteString but I'm okay with limiting it to Message.
</blockquote></details>

Yes, I'm referring to `GooglePubSubClient` (or `QueueSubscriber`). There is no reason to allow for anything other than protobuf messages for pubsub or message queues.

Marco-Premier (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1904)* status: all files reviewed, 1 unresolved discussion (waiting on @SanjayVas and @stevenwarejones)

___
*[`src/test/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplicationTest.kt` line 80 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1904#-OC-fCdE5atqUwhLE0BC:-OCiciAEARhKCBnVIfR0:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4f82708fb53e2a12f7c32fe77be87bb47a48000b/src/test/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplicationTest.kt#L80)):*

Yes, I'm referring to `GooglePubSubClient` (or `QueueSubscriber`). There is no reason to allow for anything other than protobuf messages for pubsub or message queues.
</blockquote></details>

Done.

stevenwarejones (changes requested):
> Reviewed 3 of 4 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1904)* status: 8 of 9 files reviewed, 2 unresolved discussions (waiting on @Marco-Premier and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt` line 79 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1904#-OD6jwoaD5TwxV5xjtr_:-OD6jwoaD5TwxV5xjtra:b4hndv8) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a555153d377dbdd8ad4c16915703aa77c424c237/src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt#L79)):*
> ```Kotlin
>       logger.severe("Failed to parse protobuf message: ${e.message}")
>       queueMessage.nack()
> ```

aren't there other exceptions we don't want to retry? I'd prefer to have a list of retryable excpetions and only retry for those.

Marco-Premier (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1904)* status: 8 of 9 files reviewed, 2 unresolved discussions (waiting on @SanjayVas and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt` line 79 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1904#-OD6jwoaD5TwxV5xjtr_:-OD6yordBfsqO4GBqb49:b-tcbqke) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a555153d377dbdd8ad4c16915703aa77c424c237/src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt#L79)):*

aren't there other exceptions we don't want to retry? I'd prefer to have a list of retryable excpetions and only retry for those.
</blockquote></details>

Can we use a dead-letter-queue policy with a max-retry attempts set to (eg.) 5 for the moment?
I'd like to investigate on a separate PR what are the exception we want to retry for ans which ones we don't want

SanjayVas (changes requested):
> Reviewed 2 of 8 files at r4, 2 of 4 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1904)* status: 8 of 9 files reviewed, 6 unresolved discussions (waiting on @Marco-Premier and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt` line 52 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1904#-ODSlsX9CPkJs4uTxSsz:-ODSlsX9CPkJs4uTxSt-:bpnpax1) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a555153d377dbdd8ad4c16915703aa77c424c237/src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt#L52)):*
> ```Kotlin
>    * If an error occurs during the message flow, it is logged and handling continues.
>    */
>   private suspend fun startListening() {
> ```

nit: this isn't actually "starting" as it suspends until cancelled or the channel is closed. That is to say it's actually receiving and processing.



_Code quote:_
```Kotlin
startListening
```
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt` line 57 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1904#-ODSkyMG-38DAk1f7udf:-ODSkyMHAORyBprVKNzm:b-zardgt) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a555153d377dbdd8ad4c16915703aa77c424c237/src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt#L57)):*
> ```Kotlin
>         queueSubscriber.subscribe(queueName, parser)
>       messageChannel
>         .consumeAsFlow()
> ```

Why? You're not passing this to any other code that uses Flows rather than Channels. You can just use a for loop using the channel iterator

```kotlin
for (message: QueueSubscriber.QueueMessage<T> in messageChannel) {
  processMessage(message)
}
```

Or  a while loop if you want to call receive yourself e.g. if you want to do more specific error handling

```kotlin
while (true) {
  val message: QueueSubscriber.QueueMessage<T> = messageChannel.receive()
  processMessage(message)
}
```



_Code quote:_
```Kotlin
consumeAsFlow
```
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt` line 78 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1904#-ODSo6NK-b--fOaIuJpP:-ODSo6NK-b--fOaIuJpQ:bbo9kup) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a555153d377dbdd8ad4c16915703aa77c424c237/src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt#L78)):*
> ```Kotlin
>     } catch (e: InvalidProtocolBufferException) {
>       logger.severe("Failed to parse protobuf message: ${e.message}")
>       queueMessage.nack()
> ```

Do we actually want BaseTeeApplication to handle this, or let each application implementation handle acking/nacking? If we're just going to do this, I don't see a huge benefit in exposing the ack/nack from QueueSubscriber.
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt` line 91 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1904#-ODSoPhkCgmdgqyiXx6v:-ODSoPhkCgmdgqyiXx6w:b-vrnigi) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a555153d377dbdd8ad4c16915703aa77c424c237/src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt#L91)):*
> ```Kotlin
>       queueSubscriber.close()
>       logger.severe("Error during close: ${e.message}")
> ```

I don't think you need to log here, since I assume this will be called from `main` of the application. Therefore it can just bubble up.



_Code quote:_
```Kotlin
logger.severe("Error during close: ${e.message}")
```

Marco-Premier (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1904)* status: 4 of 9 files reviewed, 5 unresolved discussions (waiting on @SanjayVas and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt` line 57 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1904#-ODSkyMG-38DAk1f7udf:-ODfhPKi-G4F59fpOg72:b-jhceso) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a555153d377dbdd8ad4c16915703aa77c424c237/src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt#L57)):*

Why? You're not passing this to any other code that uses Flows rather than Channels. You can just use a for loop using the channel iterator

```kotlin
for (message: QueueSubscriber.QueueMessage<T> in messageChannel) {
  processMessage(message)
}
```

Or  a while loop if you want to call receive yourself e.g. if you want to do more specific error handling

```kotlin
while (true) {
  val message: QueueSubscriber.QueueMessage<T> = messageChannel.receive()
  processMessage(message)
}
```
</blockquote></details>

I'm happy with the `for` loop. `processMessage` already handle TEE application-level exceptions. For other (eg. channel close, connection issue with pubsub) I'm happy if the app breaks and then eventually rely on  `auto-healing`  capabilities of `MIG`.

I'm afraid that handle all the possible exceptions there might be quite complex.
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt` line 78 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1904#-ODSo6NK-b--fOaIuJpP:-ODfiLi36bAEAqXUlJLx:b-h6dmml) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a555153d377dbdd8ad4c16915703aa77c424c237/src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt#L78)):*

Do we actually want BaseTeeApplication to handle this, or let each application implementation handle acking/nacking? If we're just going to do this, I don't see a huge benefit in exposing the ack/nack from QueueSubscriber.
</blockquote></details>



I see the value of leaving `ack` / `nack` here as this class is just a wrapper around the `Subscriber` and abstracts that logic away from developers. I think that letting each application handling `ack` / `nack` is prone to errors (one may forget them ecc...), especially with `Tee Apps` where we want to enforce the processing of one message at the time.

However it doesn’t prevent other classes in future from consuming the `Subscriber` with different acknowledgment policies. For example, one might choose to acknowledge a message immediately upon receipt to enable parallel message processing.
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt` line 91 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1904#-ODSoPhkCgmdgqyiXx6v:-ODfkLWM5Sj2ecENztOp:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a555153d377dbdd8ad4c16915703aa77c424c237/src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt#L91)):*

I don't think you need to log here, since I assume this will be called from `main` of the application. Therefore it can just bubble up.
</blockquote></details>

Done.

SanjayVas (changes requested):
> Reviewed 5 of 5 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1904)* status: all files reviewed, 6 unresolved discussions (waiting on @Marco-Premier and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt` line 72 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1904#-ODi8vKj1mjECKZslZbf:-ODi8vKj1mjECKZslZbg:b-kv1tbw) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6dc21f6e0e00d0ec02f40e7fb1f12899042e3c3a/src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt#L72)):*
> ```Kotlin
>       queueMessage.ack()
>     } catch (e: InvalidProtocolBufferException) {
>       logger.severe("Failed to parse protobuf message: ${e.message}")
> ```

If you're going to log using an exception, use the method that takes in the exception itself rather than including the message.



_Suggestion:_
```Kotlin
logger.log(Level.SEVERE, e) { "Failed to parse protobuf message" }
```
___
*[`src/test/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplicationTest.kt` line 44 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1904#-ODiAmO-1hZ4-GlcA0WP:-ODiAmO-1hZ4-GlcA0WQ:b7qr3d1) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6dc21f6e0e00d0ec02f40e7fb1f12899042e3c3a/src/test/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplicationTest.kt#L44)):*
> ```Kotlin
>     parser = parser,
>   ) {
>   val processedMessages: MutableList<TestWork> = mutableListOf()
> ```

nit: since this implementation only has a single Deferred, you can simplify this by dropping the list and having the Deferred contain the message.
___
*[`src/test/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplicationTest.kt` line 55 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1904#-ODi9lW30z4BRhA3fLEK:-ODi9lW47IP0R2eMfqe6:b-cm08d0) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6dc21f6e0e00d0ec02f40e7fb1f12899042e3c3a/src/test/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplicationTest.kt#L55)):*
> ```Kotlin
> class BaseTeeApplicationTest {
> 
>   @Rule
> ```

This should go in the companion object, assuming it's supposed to be one emulator shared across all test methods.
___
*[`src/test/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplicationTest.kt` line 58 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1904#-ODiANvpCnFq6bUDGMuG:-ODiANvpCnFq6bUDGMuH:b9shtkq) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6dc21f6e0e00d0ec02f40e7fb1f12899042e3c3a/src/test/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplicationTest.kt#L58)):*
> ```Kotlin
>   @JvmField val pubSubEmulatorProvider = GooglePubSubEmulatorProvider()
> 
>   private val projectId = "test-project"
> ```

These should be constants in the companion object
___
*[`src/test/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplicationTest.kt` line 65 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1904#-ODi9uv_AnB9k_lXBlSl:-ODi9uv_AnB9k_lXBlSm:bfavjzl) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6dc21f6e0e00d0ec02f40e7fb1f12899042e3c3a/src/test/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplicationTest.kt#L65)):*
> ```Kotlin
> 
>   @Before
>   fun setup() {
> ```

nit: same comment as before on using meaningful names



_Code quote:_
```Kotlin
setup
```

Marco-Premier (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1904)* status: all files reviewed, 3 unresolved discussions (waiting on @SanjayVas and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt` line 72 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1904#-ODi8vKj1mjECKZslZbf:-ODsDYZV7fetXMJFH01v:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6dc21f6e0e00d0ec02f40e7fb1f12899042e3c3a/src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt#L72)):*

If you're going to log using an exception, use the method that takes in the exception itself rather than including the message.
</blockquote></details>

Done.
___
*[`src/test/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplicationTest.kt` line 55 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1904#-ODi9lW30z4BRhA3fLEK:-ODsGzuP1-oZDUkLbTFV:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6dc21f6e0e00d0ec02f40e7fb1f12899042e3c3a/src/test/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplicationTest.kt#L55)):*

This should go in the companion object, assuming it's supposed to be one emulator shared across all test methods.
</blockquote></details>

Done.
___
*[`src/test/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplicationTest.kt` line 58 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1904#-ODiANvpCnFq6bUDGMuG:-ODsH5EEDAsp7xNeKFss:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6dc21f6e0e00d0ec02f40e7fb1f12899042e3c3a/src/test/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplicationTest.kt#L58)):*

These should be constants in the companion object
</blockquote></details>

Done.

SanjayVas (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1904)* status: all files reviewed, 1 unresolved discussion (waiting on @Marco-Premier and @stevenwarejones)

___
*[`src/test/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplicationTest.kt` line 104 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1904#-ODsO3Qu4PiCSIsyziF1:-ODsO3Qv86lBNOrFVwuo:b8tis06) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5c2d1757f2bc398f9d5bb41127511ae8987cc36b/src/test/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplicationTest.kt#L104)):*
> ```Kotlin
>   companion object {
> 
>     private const val projectId = "test-project"
> ```

nit: CONST_CASE



_Code quote:_
```Kotlin
projectId
```

Marco-Premier (commented):
> Reviewed 2 of 8 files at r4, 1 of 5 files at r5, 1 of 5 files at r7, 5 of 5 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1904)* status: all files reviewed, 1 unresolved discussion (waiting on @stevenwarejones)

Marco-Premier (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1904)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @stevenwarejones)

stevenwarejones (approved):
> Reviewed 1 of 5 files at r7, 4 of 5 files at r9, 1 of 1 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1904)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Marco-Premier)

stevenwarejones (approved):
> Reviewed 1 of 5 files at r7, 4 of 5 files at r9, 1 of 1 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1904)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Marco-Premier)

Marco-Premier (commented):
> Reviewed 1 of 1 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1904)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Marco-Premier)

PR #1916, opened 2024-11-13
fix!: Update maximum and default page_size for ListRequisitions and ListEventGroups

SanjayVas:
Closes #1915

PR #1923, opened 2024-11-15
refactor: Extract listResources utility function for handling pagination

SanjayVas:
This also fixes an issue where EventGroupsServiceTest was waiting for a real 30s deadline to pass.
Also fixes the simulator issue of PR https://github.com/world-federation-of-advertisers/cross-media-measurement/pull/1927:
The MC and EDP simulators were only reading the first page of results from ListEventGroups. This means that in environments with more EventGroups, not all test EventGroups would be read. This is exacerbated by https://github.com/world-federation-of-advertisers/cross-media-measurement/pull/1916, as the default page size is now smaller.


SanjayVas:
CC @roaminggypsy

wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1923)

tristanvuong2021 (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1923)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @SanjayVas)

stevenwarejones (changes requested):
> Reviewed 7 of 11 files at r1, 4 of 4 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1923)* status: all files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/common/api/grpc/ListResources.kt` line 66 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1923#-OCFdCeo08llhNQzC5fP:-OCFdCep3jgofBeM_o8A:bxx66il) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/929c89a643a03479bb04d01bf54f31260984dab6/src/main/kotlin/org/wfanet/measurement/common/api/grpc/ListResources.kt#L66)):*
> ```Kotlin
>       coroutineContext.ensureActive()
> 
>       val resourceList: ResourceList<T> = list(nextPageToken, remaining)
> ```

why not just do
```
val resourceList = list(nextPagetoken, remaining).take(remaining-resourceList.size)
```

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1923)* status: all files reviewed, 1 unresolved discussion (waiting on @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/common/api/grpc/ListResources.kt` line 66 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1923#-OCFdCeo08llhNQzC5fP:-OCFiPHFAZumkhrgmOHe:byrwlnb) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/929c89a643a03479bb04d01bf54f31260984dab6/src/main/kotlin/org/wfanet/measurement/common/api/grpc/ListResources.kt#L66)):*

why not just do
```
val resourceList = list(nextPagetoken, remaining).take(remaining-resourceList.size)
```

</blockquote></details>

The items that are returned must match the page token that is returned so that the end caller can store the page token and use it for subsequent requests. Therefore it is an error to return fewer items than where the next page token would start from.

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1923)* status: all files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/common/api/grpc/ListResources.kt` line 66 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1923#-OCFdCeo08llhNQzC5fP:-OCFjDN24hINi1kxdd5c:bpfsao6) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/929c89a643a03479bb04d01bf54f31260984dab6/src/main/kotlin/org/wfanet/measurement/common/api/grpc/ListResources.kt#L66)):*

The items that are returned must match the page token that is returned so that the end caller can store the page token and use it for subsequent requests. Therefore it is an error to return fewer items than where the next page token would start from.
</blockquote></details>

I guess that's a difference in perceived usage. I think the function should cease to return a pageToken once the limit is reached.

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1923)* status: all files reviewed, 1 unresolved discussion (waiting on @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/common/api/grpc/ListResources.kt` line 66 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1923#-OCFdCeo08llhNQzC5fP:-OCFksnXD4E15mYQuWt3:b76bpe) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/929c89a643a03479bb04d01bf54f31260984dab6/src/main/kotlin/org/wfanet/measurement/common/api/grpc/ListResources.kt#L66)):*

> I think the function should cease to return a pageToken once the limit is reached.

That is strictly incorrect, as it gives no signal that there are items beyond the limit. The implementation **must** return a next page token in this case.
</blockquote></details>

To clarify, this function is intended to be used to *implement* pagination, not just to call a paginated method. e.g. one List method impl calling another, or a UI that wants to display full pages of a specific size.

renjiezh:
I accidentally merged https://github.com/world-federation-of-advertisers/cross-media-measurement/pull/1927 into this branch instead of main branch.
The PR description has been updated to reflect #1927.

renjiezh (approved):
> Reviewed 6 of 11 files at r1, 2 of 4 files at r2, 3 of 3 files at r3, 3 of 3 files at r4, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1923)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @SanjayVas)

PR #1926, opened 2024-11-18
ci: Add conventional commit PR check

### LINKED ISSUE #1908: Add Conventional Commit In Pull Requests GitHub Action
Once this is done, we'll have [Conventional Commit In Pull Requests GitHub Action](https://github.com/marketplace/actions/conventional-commit-in-pull-requests) as part of the automated PR checks.

mikkokotila:
Closes #1908

### DIFF
--- a/.github/workflows/conventional-commit-pr.yml
+++ b/.github/workflows/conventional-commit-pr.yml
@@ -1,0 +1,15 @@
+name: PR Conventional Commit Validation
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened, edited]
+
+jobs:
+  validate-pr-title:
+    runs-on: ubuntu-latest
+    steps:
+      - name: PR Conventional Commit Validation
+        uses:  ytanikin/pr-conventional-commits@1.4.0
+        with:
+          task_types: '["feat", "fix", "docs", "test", "ci", "refactor", "perf", "chore", "revert"]'
+          add_label: 'false'
--- a/.github/workflows/conventional-commit-pr.yml
+++ b/.github/workflows/conventional-commit-pr.yml
@@ -1,3 +1,17 @@
+# Copyright 2024 The Cross-Media Measurement Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
 name: PR Conventional Commit Validation
 
 on:


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1926)

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1926)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @mikkokotila)

___
*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1926#-OC0vvfV0MlLZ4Dngu4H:-OC0vvfV0MlLZ4Dngu4I:bqt36bp) (no related file):*
Did you evaluate different actions? What signals did you use to determine that of the available actions, that this is of good quality and is well-maintained? Did you consider the alternatives such as wrapping CommitLint with our own action and weigh that against the off-the-shelf options?

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1926)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @mikkokotila)

___
*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1926#-OC0vvfV0MlLZ4Dngu4H:-OC0w0VSBTtcQFAXtzF7:bg3vyjr) (no related file):*

Did you evaluate different actions? What signals did you use to determine that of the available actions, that this is of good quality and is well-maintained? Did you consider the alternatives such as wrapping CommitLint with our own action and weigh that against the off-the-shelf options?
</blockquote></details>

Is the action you chose as configurable as CommitLint? We probably want to start with the Angular configuration as a base, but use slightly different rules such as allowing the message itself to start with an uppercase letter, etc.

mikkokotila:
> Did you evaluate different actions? What signals did you use to determine that of the available actions, that this is of good quality and is well-maintained? Did you consider the alternatives such as wrapping CommitLint with our own action and weigh that against the off-the-shelf options?

No, none of that. I simply tested it in a another repo, and ran through few common scenarios to test it. It is a very simple action, which is very easy to replace if need to. In other words, risk of negative consequences arising from adoption are very low and easy to mitigate in case of arising.

SanjayVas (commented):
That is not something you can just simply state. It is below the standards of rigor for this project. For example, from our [Code Style](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/main/docs/code-style.md):

> Limit usage of unnecessary third party dependencies. Due to the privacy requirements of this project, we prefer avoiding unaudited dependencies.

While this is a process dependency and not a library dependency and therefore less susceptible to risk, we still always evaluate third-party dependencies. More broadly, every change on this project is expected to have gone through the appropriate level of rigor in terms of considering alternatives. For example, this is the purpose of the Caveats section in the [design doc template](https://docs.google.com/document/d/1Tx00oeoQB5dB8Un-Lm3WJOQr6V5edfwLJ8c78sxbHQI/template/preview?tab=t.0&resourcekey=0-KCD_-FERFR7Kysr-_1dTSA).

As an aside, it appears that you replied to the comment directly on GitHub rather than using Reviewable as mentioned in the [Dev Standards doc](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/main/docs/dev-standards.md#use-reviewable)

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1926#-:-OC4dDNr7kfBiG7gkuLR:b-xkurkd)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @mikkokotila)

mikkokotila (commented):
> we still always evaluate third-party dependencies

Do you mean that this level of step requires completing the design doc template? So, using a more typical open source way of saying it, if one wants to introduce a new workflow action, one must make an RFC for it first? 

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1926#-:-OCXpNONA6jTnH1wwC9O:bxje4qo)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @kungfucraig and @SanjayVas)

___
*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1926#-OC0vvfV0MlLZ4Dngu4H:-OCXonebDRA-Xb71Tn9k:b-2m4nq8) (no related file):*

Is the action you chose as configurable as CommitLint? We probably want to start with the Angular configuration as a base, but use slightly different rules such as allowing the message itself to start with an uppercase letter, etc.
</blockquote></details>

Sure, if we want to take an alternative route, there is no need to put more time into this here. But I would strongly advise against making it more flexible. There is great value in adopting a common way of working, like Conventional Commits, and adhering to it. Especially things like allowing flexible capitalization should be avoided, as it does not offer any value but has the important downside of lack of consistency.

kungfucraig:
Hey @mikkokotila,

I think the way to go about this would be to write the "RFC", aka one-pager, that outlines one or more options for this. Then circulate that on slack and we can discuss at our eng review meeting in a week and decide on something. I think that will end up being more efficient than designing within the scope of a PR.

mikkokotila (commented):
Okay, I think I get it. Our approach makes proposing even minimal changes to the current way of working "expensive." I understand it is a way to reduce disturbance to the current work window.

For example, this particular case would involve first investigating what we are already doing that is relevant and could be extended, then evaluating the available ready-made solutions as an alternative to that. Then, craft the proposal, get somebody to review it, and publish it. I'd imagine that doing this well would take a couple of hours. I can imagine many things I can do in a few hours that have exponentially more value to the project. If a few engineers end up looking at it, maybe a junior one makes a context switch for it, there is some discussion, and again, it might end up being a couple of man-hours, at least when the productivity disturbance is considered. 

So, it makes no sense for Halo to proceed with such proposals. It makes sense for the engineers, at least until we get more hands on deck, to focus on what they already focus on. For janitorial staff such as myself, it makes sense to focus on things that do not take time from engineers but make their and others' lives better.

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1926#-:-OCbfdI38o74cPSQGWP8:b71saj7)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @kungfucraig and @SanjayVas)

kungfucraig (commented):
We operate on a consensus model.  It's less about disrupting the flow and more about ensuring the set of stakeholders are on board with the changes.

More specifically, and Sanjay made this point above, we like to be very conscious about the dependencies we take.

As for understanding what we are already doing, this is extremely important for any change that is being proposed. Having those short conversations would have saved us all some time on this, the issue labeling, and the org rename proposals. You should consider joining our engineering discussion meetings on Monday so that you can get to know the team and how it operates a bit better.

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1926#-:-OCdMifRDG2PUAXc2kDD:bwhhdv)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

mikkokotila (commented):
Thanks for clarifying.  Being conscious about dependencies and other things that affect our ergonomics and economics is mission-critical for a project like Halo.

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1926#-:-OD6EDBH9jtiHH5ZNLod:b-bjbavv)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

Issue #1913, opened 2024-11-13
Require EventGroup reference IDs to be unique per (DataProvider, MeasurementConsumer) pair


Saarahx:
**Describe the bug**
Meta flagged that they were able to create an event group multiple times with the same event_group_reference_id

**Steps to reproduce**
Event groups have since been deleted so unable to replicate issue

**Component(s) affected**
Kingdom

**Version**
Halo v5.11

**Environment**
Production

**Additional context**

SanjayVas:
Note that this is technically working as intended. We had a feature request discussed at the workshop in LA to require EventGroup reference IDs to be unique per (DataProvider, MeasurementConsumer) pair.

PR #1927, opened 2024-11-18
fix: Read all EventGroups from simulators rather than stopping at first page

SanjayVas:
The MC and EDP simulators were only reading the first page of results from ListEventGroups. This means that in environments with more EventGroups, not all test EventGroups would be read. This is exacerbated by #1916, as the default page size is now smaller.

### DIFF
--- a/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt
+++ b/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt
@@ -36,9 +36,11 @@ import kotlin.math.roundToInt
 import kotlin.random.Random
 import kotlin.random.asJavaRandom
 import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.ExperimentalCoroutinesApi
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.asFlow
 import kotlinx.coroutines.flow.emitAll
+import kotlinx.coroutines.flow.firstOrNull
 import kotlinx.coroutines.flow.flow
 import kotlinx.coroutines.flow.map
 import kotlinx.coroutines.withContext
@@ -109,6 +111,9 @@ import org.wfanet.measurement.api.v2alpha.updateEventGroupRequest
 import org.wfanet.measurement.common.Health
 import org.wfanet.measurement.common.ProtoReflection
 import org.wfanet.measurement.common.SettableHealth
+import org.wfanet.measurement.common.api.grpc.ResourceList
+import org.wfanet.measurement.common.api.grpc.flattenConcat
+import org.wfanet.measurement.common.api.grpc.listResources
 import org.wfanet.measurement.common.asBufferedFlow
 import org.wfanet.measurement.common.crypto.authorityKeyIdentifier
 import org.wfanet.measurement.common.crypto.readCertificate
@@ -339,27 +344,29 @@ class EdpSimulator(
    * Returns the first [EventGroup] for this `DataProvider` and [MeasurementConsumer] with
    * [eventGroupReferenceId], or `null` if not found.
    */
+  @OptIn(ExperimentalCoroutinesApi::class) // For `flattenConcat`.
   private suspend fun getEventGroupByReferenceId(eventGroupReferenceId: String): EventGroup? {
-    val response =
-      try {
-        eventGroupsStub.listEventGroups(
-          listEventGroupsRequest {
-            parent = edpData.name
-            filter =
-              ListEventGroupsRequestKt.filter { measurementConsumers += measurementConsumerName }
-            pageSize = Int.MAX_VALUE
+    return eventGroupsStub
+      .listResources { pageToken ->
+        val response =
+          try {
+            listEventGroups(
+              listEventGroupsRequest {
+                parent = edpData.name
+                filter =
+                  ListEventGroupsRequestKt.filter {
+                    measurementConsumers += measurementConsumerName
+                  }
+                this.pageToken = pageToken
+              }
+            )
+          } catch (e: StatusException) {
+            throw Exception("Error listing EventGroups", e)
           }
-        )
-      } catch (e: StatusException) {
-        throw Exception("Error listing EventGroups", e)
+        ResourceList(response.eventGroupsList, response.nextPageToken)
       }
-
-    // TODO(@SanjayVas): Support filtering by reference ID so we don't need to handle multiple pages
-    // of EventGroups.
-    check(response.nextPageToken.isEmpty()) {
-      "Too many EventGroups for ${edpData.name} and $measurementConsumerName"
-    }
-    return response.eventGroupsList.find { it.eventGroupReferenceId == eventGroupReferenceId }
+      .flattenConcat()
+      .firstOrNull { it.eventGroupReferenceId == eventGroupReferenceId }
   }
 
   private suspend fun ensureMetadataDescriptor(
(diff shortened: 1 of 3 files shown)


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1927)

SanjayVas:
Depends on #1923

stevenwarejones (changes requested):
> Reviewed 2 of 6 files at r1, 2 of 14 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1927)* status: 4 of 20 files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/common/api/grpc/ListResources.kt` line 67 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1927#-OC3pBLUDUOEi6JDTKqQ:-OC3pBLUDUOEi6JDTKqR:b-nwr0p5) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c21917ab186cc0ce408a120a77099f7985a89b5b/src/main/kotlin/org/wfanet/measurement/common/api/grpc/ListResources.kt#L67)):*
> ```Kotlin
> 
>       val resourceList: ResourceList<T> = list(nextPageToken, remaining)
>       require(resourceList.size <= remaining) {
> ```

should this really be required here and throw an error? this is a problem with the list that is passed in..... meaning an API that returned too many items. I'd prefer it took only that many elements if there are too many.

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1927)* status: 4 of 20 files reviewed, 1 unresolved discussion (waiting on @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/common/api/grpc/ListResources.kt` line 67 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1927#-OC3pBLUDUOEi6JDTKqQ:-OC4_O9_EgKabkuugWB9:b-2k5n61) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c21917ab186cc0ce408a120a77099f7985a89b5b/src/main/kotlin/org/wfanet/measurement/common/api/grpc/ListResources.kt#L67)):*
> I'd prefer it took only that many elements if there are too many.

That wouldn't work, as the returned next page token wouldn't have the right value to get subsequent items.

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1927)* status: all files reviewed, 1 unresolved discussion (waiting on @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt` line 369 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1927#-OCFcJza0hDKXt55msMK:-OCFiwSjC2EiB5wNNF4q:b-v2b7vv) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/673ba6f900a0762cd007f46e34efc83e9e1d6efa/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt#L369)):*

I'd prefer you prefilter instead of unncessarily fetching all event groups.... something like 

```
eturn eventGroupsStub
      .listResources(
          limit = 1
) { pageToken ->
        val response =
          try {
            listEventGroups(
              listEventGroupsRequest {
                parent = edpData.name
                filter =
                  ListEventGroupsRequestKt.filter {
                    measurementConsumers += measurementConsumerName
                  }
                this.pageToken = pageToken
              }
            )
            throw Exception("Error listing EventGroups", e)
          }
        ResourceList(
response.eventGroupsList.filter { it.eventGroupReferenceId == eventGroupReferenceId }, 
response.nextPageToken
)
      }
      .flattenConcat()
```

</blockquote></details>

That would be less efficient as it means making a call for one EventGroup at a time rather than batching by the default page size. This will still stop making calls once we get a response that contains the event group.

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1927)* status: all files reviewed, 1 unresolved discussion (waiting on @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt` line 369 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1927#-OCFcJza0hDKXt55msMK:-OCG4dhH1ghMh7w9fp92:b8lj1jv) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e026654bcd7035f642145036ef15edd67b76505b/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt#L369)):*

That would be less efficient as it means making a call for one EventGroup at a time rather than batching by the default page size. This will still stop making calls once we get a response that contains the event group.
</blockquote></details>

To clarify, what I mean is that the implementation in this PR will already stop making requests once it gets a response that has an EventGroup that has the specified reference ID. This is because `Flow.flattenConcat` is an *intermediate* Flow operator analogous to [`Sequence.flatten`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.sequences/flatten.html).

Your suggestion does not add any benefit and furthermore does not meet the constraint of never returning more items than `limit` since EventGroup reference IDs are not yet guaranteed to be unique (see #1913). In order to meet that constraint, it would have to pass the `remaining` value as the `pageSize` in the request resulting in the non-batching behavior I mentioned.

stevenwarejones (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1927)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt` line 369 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1927#-OCFcJza0hDKXt55msMK:-OCJjuy77oB7YcQaIJRb:bk5wxjg) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7b4e9220c88e41a373478eeca1cbbfae6e80008a/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt#L369)):*

To clarify, what I mean is that the implementation in this PR will already stop making requests once it gets a response that has an EventGroup that has the specified reference ID. This is because `Flow.flattenConcat` is an *intermediate* Flow operator analogous to [`Sequence.flatten`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.sequences/flatten.html).

Your suggestion does not add any benefit and furthermore does not meet the constraint of never returning more items than `limit` since EventGroup reference IDs are not yet guaranteed to be unique (see #1913). In order to meet that constraint, it would have to pass the `remaining` value as the `pageSize` in the request resulting in the non-batching behavior I mentioned.
</blockquote></details>

I was not aware that Flow.flattenConcat would end up causing the requests to stop once an event group is found.

However, I think you're misunderstanding what what I suggested... I'm suggesting that the listPredicate itself filter out non-matching resource names. That means that it will always return either 1 or 0 items. If it finds no matches, it returns 0 items. Otherwise, it returns 1. Once once is returned, there is no need to continue. I think this is a little more straightforward. The reader may read your code and make the same assumption I did.

Issue #1934, opened 2024-11-26
Simulator event groups created without event templates from --event-message-descriptor-set


SanjayVas:
Reported by @nitamanna18

**Describe the bug**
When the EDP simulator creates event groups using `--event-message-descriptor-set` to specify the event message descriptor, the `event_templates` field on those event groups is empty.

**Steps to reproduce**
1. Configure the EDP simulator using `--event-message-descriptor-set` set to a FileDescriptorSet containing the descriptor for the [UK Pilot Event message type](https://github.com/world-federation-of-advertisers/uk-pilot-event-templates/blob/main/src/main/proto/halo_cmm/uk/pilot/event.proto) and that type referenced by the population spec provided to `--population-spec`
2. Run the simulator

**Component(s) affected**
EDP simulator

**Version**
v0.5.13

**Environment**
Origin dev

**Additional context**

ListEventGroups response as JSON:

```json
{
   "eventGroups": [
       {
           "name": "measurementConsumers/a1fyyNkb-TQ/eventGroups/KHrh5NWF6Fc",
           "cmmsEventGroup": "dataProviders/LkJlb3UZkY8/eventGroups/KHrh5NWF6Fc",
           "cmmsDataProvider": "dataProviders/LkJlb3UZkY8",
           "eventGroupReferenceId": "sim-eg-edp3ff",
           "eventTemplates": [],
           "metadata": {
               "fastflixId": "1234567890123456",
               "campaignEndDate": "07-12-2024",
               "displayName": "Fastflix Alpha Campaign",
               "brand": "SuperBrand",
               "campaignStartDate": "07-11-2024"
           }
       },
       {
           "name": "measurementConsumers/a1fyyNkb-TQ/eventGroups/N_DhhtWE2nw",
           "cmmsEventGroup": "dataProviders/TbsgjIrmbj0/eventGroups/N_DhhtWE2nw",
           "cmmsDataProvider": "dataProviders/TbsgjIrmbj0",
           "eventGroupReferenceId": "sim-eg-edp2tv",
           "eventTemplates": [],
           "metadata": {
               "campaignEndDate": "07-12-2024",
               "displayName": "Origin Alpha Campaign",
               "brand": "OriginBrand",
               "campaignStartDate": "07-11-2024"
           }
       },
       {
           "name": "measurementConsumers/a1fyyNkb-TQ/eventGroups/BFkdZ9WFOps",
[...]

PR #1945, opened 2024-12-03
fix: Ensure event annotations is included in compile-time protobuf descriptors

SanjayVas:
A FileDescriptor that is a compile-time dependency is not considered the same as the FileDescriptor loaded at run-time from a FileDescriptorSet for the same file. This ensures that the FileDescriptor for event_annotations.proto known at compile-time is used instead of one that's possibly loaded from the `--event-message-descriptor-set` option.


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1945)

SanjayVas:
This is intended to resolve #1934. For reasons discussed offline, there is no good way to test this in the Halo codebase without significant refactoring to ensure that the event message type used for the test is _not_ a compile-time dependency. Therefore testing/verification is reliant on operators that use this feature, e.g. Origin.

stevenwarejones (approved):
> Reviewed 4 of 4 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1945)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @SanjayVas)

PR #1978, opened 2024-12-19
refactor: Add logging for report processor.

ple13:


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1978)

kungfucraig (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1978)* status: 0 of 9 files reviewed, 14 unresolved discussions (waiting on @ple13 and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/postprocessing/v2alpha/ReportProcessor.kt` line 93 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1978#-OEZ_J3n-_GH8Zl84Qqd:-OEZ_J3n-_GH8Zl84Qqe:b80p7qn) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e58e65749787c70df156e9a555c84716565caad6/src/main/kotlin/org/wfanet/measurement/reporting/postprocessing/v2alpha/ReportProcessor.kt#L93)):*
> ```Kotlin
> 
>       // TODO(bazelbuild/bazel#17629): Execute the Python zip directly once this bug is fixed.
>       val processBuilder = ProcessBuilder("python3", tempFile.toPath().toString())
> ```

We need a way to set debug level logging in the Python program from here. A flag would make sense on the python side, but I suspect there could be a more idiomatic way to do it. 

In addition, you'll want provide the client of the Kotlin class a way to set the debug log level.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/postprocessing/v2alpha/ReportProcessor.kt` line 107 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1978#-OEZYFuuEfleNznMewOR:-OEZYFuuEfleNznMewOS:bmromlx) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e58e65749787c70df156e9a555c84716565caad6/src/main/kotlin/org/wfanet/measurement/reporting/postprocessing/v2alpha/ReportProcessor.kt#L107)):*
> ```Kotlin
>         BufferedReader(InputStreamReader(process.inputStream)).use { it.readText() }
> 
>       // Logs from python program, which are written to stderr, are read and re-logged. When
> ```

You should use the exit code of the process to determine whether to throw an exception. In the case you detect a non-zero exit then you should also read the entire stderr and package that into the exception message.

You should create a custom exception as well. Sorry I didn't mention this the other day.

In the case you have a zero exit code, you could read the stderr from the python process and log it back to Info here. I have mixed feelings about this though. @SanjayVas wdyt?
___
*[`src/main/python/wfa/measurement/reporting/postprocessing/noiseninja/solver.py` line 50 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1978#-OEZXHxI-awhmTMR4sRr:-OEZXHxI-awhmTMR4sRs:bcc3x2t) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e58e65749787c70df156e9a555c84716565caad6/src/main/python/wfa/measurement/reporting/postprocessing/noiseninja/solver.py#L50)):*
> ```Python
>   def __init__(self, set_measurement_spec: SetMeasurementsSpec):
>     logger.info(
>         "Initialize the solver with constraints obtained from the set "
> ```

You should say "Initializing" here. Similarly with the verb tense in other messages.

Also consider logging the number of variables here. Similarly with messages below.
___
*[`src/main/python/wfa/measurement/reporting/postprocessing/noiseninja/solver.py` line 79 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1978#-OEZWywL2BK0moifMxKu:-OEZWywL2BK0moifMxKv:b-uvxzeg) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e58e65749787c70df156e9a555c84716565caad6/src/main/python/wfa/measurement/reporting/postprocessing/noiseninja/solver.py#L79)):*
> ```Python
>         (mean_measurement_by_variable[i]
>          for i in range(0, self.num_variables))))
>     logger.info(f"The base values are {self.base_value}.")
> ```

nit: Consider using the debug level for this. 
___
*[`src/main/python/wfa/measurement/reporting/postprocessing/noiseninja/solver.py` line 194 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1978#-OEZZhpi4JhbeJIbZBD4:-OEZZhpj1etQzTgv_veC:b-a9evn0) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e58e65749787c70df156e9a555c84716565caad6/src/main/python/wfa/measurement/reporting/postprocessing/noiseninja/solver.py#L194)):*
> ```Python
>   def solve(self) -> Solution:
>     logger.info(
>         "Solves the quadratic program with the constraints extracted from set "
> ```

You can be a little more concise: "Solving the quadratic program."
___
*[`src/main/python/wfa/measurement/reporting/postprocessing/noiseninja/solver.py` line 200 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1978#-OEZZm0I3VmeWk-YIhm-:-OEZZm0I3VmeWk-YIhm0:bl23oqs) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e58e65749787c70df156e9a555c84716565caad6/src/main/python/wfa/measurement/reporting/postprocessing/noiseninja/solver.py#L200)):*
> ```Python
>     if self._is_feasible(self.base_value):
>       logger.info(
>           "The set measurement spec meets all the constraints, no action needed."
> ```

Same here. "The set measurement spec is feasible."
___
*[`src/main/python/wfa/measurement/reporting/postprocessing/noiseninja/solver.py` line 211 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1978#-OEZ_3Z3BWB58anEOVMy:-OEZ_3Z3BWB58anEOVMz:b-zcp5t5) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e58e65749787c70df156e9a555c84716565caad6/src/main/python/wfa/measurement/reporting/postprocessing/noiseninja/solver.py#L211)):*
> ```Python
>       )
>       while attempt_count < MAX_ATTEMPTS:
>         logger.info(f"Attempt {attempt_count}.")
> ```

nit: attempt_count+1 
___
*[`src/main/python/wfa/measurement/reporting/postprocessing/report/report.py` line 14 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1978#-OEZajES-Pz2sbfiZoyR:-OEZajES-Pz2sbfiZoyS:biwal14) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e58e65749787c70df156e9a555c84716565caad6/src/main/python/wfa/measurement/reporting/postprocessing/report/report.py#L14)):*
> ```Python
> # distributed under the License is distributed on an "AS IS" BASIS,
> # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
> # See the License for the specific language governing permissions and
> ```

I don't think it makes sense to log the entry into every function. Maybe focusing on the higher level operations would be as useful but less verbose? Happy to be convinced otherwise though.

Also in general, log lines within a for loop, if necessary, should be debug.
___
*[`src/main/python/wfa/measurement/reporting/postprocessing/report/report.py` line 53 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1978#-OEZ_B4J9tu4hQFpwsMY:-OEZ_B4J9tu4hQFpwsMZ:bgwbc8a) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e58e65749787c70df156e9a555c84716565caad6/src/main/python/wfa/measurement/reporting/postprocessing/report/report.py#L53)):*
> ```Python
>     elif comb2.issubset(comb1):
>       subset_relationships.append((comb1, comb2))
>   logger.info(
> ```

How long will this log line be? If it's very long, consider making it a debug line.
___
*[`src/main/python/wfa/measurement/reporting/postprocessing/report/report.py` line 95 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1978#-OEZa1aS0XIfLYf04Czu:-OEZa1aS0XIfLYf04Czv:b-dtcyzp) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e58e65749787c70df156e9a555c84716565caad6/src/main/python/wfa/measurement/reporting/postprocessing/report/report.py#L95)):*
> ```Python
>     is a tuple containing the sets from `other_sets` that cover it.
>   """
>   logger.info(
> ```

Same possible issue with size of log line.
___
*[`src/main/python/wfa/measurement/reporting/postprocessing/report/report.py` line 125 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1978#-OEZaGfC-kqC-WiofdCq:-OEZaGfC-kqC-WiofdCr:b-816h6l) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e58e65749787c70df156e9a555c84716565caad6/src/main/python/wfa/measurement/reporting/postprocessing/report/report.py#L125)):*
> ```Python
>       cover_relationship.append((target_set, possible_cover))
>   logger.info(
>       f"The cover relationship is {cover_relationship}."
> ```

Same.
___
*[`src/main/python/wfa/measurement/reporting/postprocessing/report/report.py` line 138 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1978#-OEZaHq602YdXXkIknkm:-OEZaHq602YdXXkIknkn:b-816h6l) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e58e65749787c70df156e9a555c84716565caad6/src/main/python/wfa/measurement/reporting/postprocessing/report/report.py#L138)):*
> ```Python
>   """
>   logger.info(
>       "Get all cover relationships from a list of EDP combinations "
> ```

Same.
___
*[`src/main/python/wfa/measurement/reporting/postprocessing/report/report.py` line 425 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1978#-OEZaX9P9yKWqT0YSEds:-OEZaX9P9yKWqT0YSEdt:b-rtgj8v) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e58e65749787c70df156e9a555c84716565caad6/src/main/python/wfa/measurement/reporting/postprocessing/report/report.py#L425)):*
> ```Python
>       for cover_relationship in self._metric_reports[
>         metric].get_cumulative_cover_relationships():
>         logger.info(f"{metric} cover relations for cumulative measurements.")
> ```

debug?
___
*[`src/main/python/wfa/measurement/reporting/postprocessing/report/report.py` line 439 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1978#-OEZaY-c6CyTueltDcUe:-OEZaY-c6CyTueltDcUf:b-rtgj8v) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e58e65749787c70df156e9a555c84716565caad6/src/main/python/wfa/measurement/reporting/postprocessing/report/report.py#L439)):*
> ```Python
>         metric].get_whole_campaign_cover_relationships():
>         logger.info(
>           f"{metric} cover relations for total campaign measurements.")
> ```

debug?

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1978)* status: 0 of 9 files reviewed, 15 unresolved discussions (waiting on @kungfucraig and @ple13)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/postprocessing/v2alpha/ReportProcessor.kt` line 93 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1978#-OEZ_J3n-_GH8Zl84Qqd:-OE_Ippp2SKn5H-lZe-z:blvcjze) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e58e65749787c70df156e9a555c84716565caad6/src/main/kotlin/org/wfanet/measurement/reporting/postprocessing/v2alpha/ReportProcessor.kt#L93)):*

We need a way to set debug level logging in the Python program from here. A flag would make sense on the python side, but I suspect there could be a more idiomatic way to do it. 

In addition, you'll want provide the client of the Kotlin class a way to set the debug log level.
</blockquote></details>

See my comment about using the logging library from abseil-py, which offers this functionality
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/postprocessing/v2alpha/ReportProcessor.kt` line 107 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1978#-OEZYFuuEfleNznMewOR:-OE_F1LZ9OkrxxLlVyv9:blbxvnk) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e58e65749787c70df156e9a555c84716565caad6/src/main/kotlin/org/wfanet/measurement/reporting/postprocessing/v2alpha/ReportProcessor.kt#L107)):*
+1 to throwing a custom exception with the output from stderr on a non-zero exit code, as that's exactly what I was thinking. No need to log on the Kotlin side in that case.

> In the case you have a zero exit code, you could read the stderr from the python process and log it back to Info here. I have mixed feelings about this though

Not sure about this one either.

Basically, you should not attempt to parse the Python logs in the Kotlin side unless you're using some kind of structured logging. Since we're not, then the log levels on the Python side are only relevant when running the Python code directly (i.e. not as a CLI from Kotlin). From the Kotlin side, we should just treat stderr as a monolith.
___
*[`src/main/python/wfa/measurement/reporting/postprocessing/noiseninja/solver.py` line 15 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1978#-OE_HkImEM0o8SMGBjy8:-OE_HkIn7Z-3kLSynQWV:b-v8s7r9) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e58e65749787c70df156e9a555c84716565caad6/src/main/python/wfa/measurement/reporting/postprocessing/noiseninja/solver.py#L15)):*
> ```Python
> 
> import logging
> ```

If we're going to use a logging library in Python, it should be Abseil logging from [abseil-py](https://github.com/abseil/abseil-py).

Among other reasons, this allows the CLI caller can control logging via command-line options.



_Suggestion:_
```Python
from absl import logging
```

ple13 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1978)* status: 0 of 12 files reviewed, 13 unresolved discussions (waiting on @kungfucraig and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/postprocessing/v2alpha/ReportProcessor.kt` line 93 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1978#-OEZ_J3n-_GH8Zl84Qqd:-OEa4V6gE0Kt1jJJroCF:b-eczif8) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e58e65749787c70df156e9a555c84716565caad6/src/main/kotlin/org/wfanet/measurement/reporting/postprocessing/v2alpha/ReportProcessor.kt#L93)):*

See my comment about using the logging library from abseil-py, which offers this functionality
</blockquote></details>

Thanks. I've added a flag to the process builder to indicate the verbosity.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/postprocessing/v2alpha/ReportProcessor.kt` line 107 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1978#-OEZYFuuEfleNznMewOR:-OEa4hi_5o0SQTVoDJ56:b-6h4v76) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e58e65749787c70df156e9a555c84716565caad6/src/main/kotlin/org/wfanet/measurement/reporting/postprocessing/v2alpha/ReportProcessor.kt#L107)):*

+1 to throwing a custom exception with the output from stderr on a non-zero exit code, as that's exactly what I was thinking. No need to log on the Kotlin side in that case.

> In the case you have a zero exit code, you could read the stderr from the python process and log it back to Info here. I have mixed feelings about this though

Not sure about this one either.

Basically, you should not attempt to parse the Python logs in the Kotlin side unless you're using some kind of structured logging. Since we're not, then the log levels on the Python side are only relevant when running the Python code directly (i.e. not as a CLI from Kotlin). From the Kotlin side, we should just treat stderr as a monolith.
</blockquote></details>

Thanks. I've updated the code with a custom exception. When exitcode is zero, I just log everything from stderr.
___
*[`src/main/python/wfa/measurement/reporting/postprocessing/noiseninja/solver.py` line 15 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1978#-OE_HkImEM0o8SMGBjy8:-OEa4nOPDr3wddkhV74r:boyexbg) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e58e65749787c70df156e9a555c84716565caad6/src/main/python/wfa/measurement/reporting/postprocessing/noiseninja/solver.py#L15)):*

If we're going to use a logging library in Python, it should be Abseil logging from [abseil-py](https://github.com/abseil/abseil-py).

Among other reasons, this allows the CLI caller can control logging via command-line options.
</blockquote></details>

Done. Thanks.
___
*[`src/main/python/wfa/measurement/reporting/postprocessing/noiseninja/solver.py` line 50 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1978#-OEZXHxI-awhmTMR4sRr:-OEZl1J93F6y6SBFRT_j:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e58e65749787c70df156e9a555c84716565caad6/src/main/python/wfa/measurement/reporting/postprocessing/noiseninja/solver.py#L50)):*

You should say "Initializing" here. Similarly with the verb tense in other messages.

Also consider logging the number of variables here. Similarly with messages below.
</blockquote></details>

Done.
___
*[`src/main/python/wfa/measurement/reporting/postprocessing/noiseninja/solver.py` line 79 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1978#-OEZWywL2BK0moifMxKu:-OE_Osj7EaMQLCIaYiBN:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e58e65749787c70df156e9a555c84716565caad6/src/main/python/wfa/measurement/reporting/postprocessing/noiseninja/solver.py#L79)):*

nit: Consider using the debug level for this.
</blockquote></details>

Done.
___
*[`src/main/python/wfa/measurement/reporting/postprocessing/noiseninja/solver.py` line 194 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1978#-OEZZhpi4JhbeJIbZBD4:-OEZg55bDYiW1RXrih_T:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e58e65749787c70df156e9a555c84716565caad6/src/main/python/wfa/measurement/reporting/postprocessing/noiseninja/solver.py#L194)):*

You can be a little more concise: "Solving the quadratic program."
</blockquote></details>

Done.
___
*[`src/main/python/wfa/measurement/reporting/postprocessing/noiseninja/solver.py` line 200 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1978#-OEZZm0I3VmeWk-YIhm-:-OEZfvDu6_g36szoLAZW:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e58e65749787c70df156e9a555c84716565caad6/src/main/python/wfa/measurement/reporting/postprocessing/noiseninja/solver.py#L200)):*

Same here. "The set measurement spec is feasible."
</blockquote></details>

Done.
___
*[`src/main/python/wfa/measurement/reporting/postprocessing/noiseninja/solver.py` line 211 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1978#-OEZ_3Z3BWB58anEOVMy:-OEZh_COESN41dCAx2yl:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e58e65749787c70df156e9a555c84716565caad6/src/main/python/wfa/measurement/reporting/postprocessing/noiseninja/solver.py#L211)):*

nit: attempt_count+1
</blockquote></details>

Done.
___
*[`src/main/python/wfa/measurement/reporting/postprocessing/report/report.py` line 14 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1978#-OEZajES-Pz2sbfiZoyR:-OEZoY0f6gTkPHx2psvE:bz4mdpz) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e58e65749787c70df156e9a555c84716565caad6/src/main/python/wfa/measurement/reporting/postprocessing/report/report.py#L14)):*

I don't think it makes sense to log the entry into every function. Maybe focusing on the higher level operations would be as useful but less verbose? Happy to be convinced otherwise though.

Also in general, log lines within a for loop, if necessary, should be debug.
</blockquote></details>

I've removed the log entries that seem redundant.
___
*[`src/main/python/wfa/measurement/reporting/postprocessing/report/report.py` line 53 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1978#-OEZ_B4J9tu4hQFpwsMY:-OEZj6IY6fLAn4Tt7XuF:bhh96cq) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e58e65749787c70df156e9a555c84716565caad6/src/main/python/wfa/measurement/reporting/postprocessing/report/report.py#L53)):*

How long will this log line be? If it's very long, consider making it a debug line.
</blockquote></details>

It could be long if there are many EDPs. I've changed it to debug.
___
*[`src/main/python/wfa/measurement/reporting/postprocessing/report/report.py` line 95 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1978#-OEZa1aS0XIfLYf04Czu:-OEZjq7yDB3cFXWtYAMt:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e58e65749787c70df156e9a555c84716565caad6/src/main/python/wfa/measurement/reporting/postprocessing/report/report.py#L95)):*

Same possible issue with size of log line.
</blockquote></details>

Done.
___
*[`src/main/python/wfa/measurement/reporting/postprocessing/report/report.py` line 125 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1978#-OEZaGfC-kqC-WiofdCq:-OEZjp-R5iTuKvJqLnYK:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e58e65749787c70df156e9a555c84716565caad6/src/main/python/wfa/measurement/reporting/postprocessing/report/report.py#L125)):*

Same.
</blockquote></details>

Done.
___
*[`src/main/python/wfa/measurement/reporting/postprocessing/report/report.py` line 138 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1978#-OEZaHq602YdXXkIknkm:-OEZjnCgC-2QOl01O1u1:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e58e65749787c70df156e9a555c84716565caad6/src/main/python/wfa/measurement/reporting/postprocessing/report/report.py#L138)):*

Same.
</blockquote></details>

Done.
___
*[`src/main/python/wfa/measurement/reporting/postprocessing/report/report.py` line 425 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1978#-OEZaX9P9yKWqT0YSEds:-OEZjlvE4tkTUeg_oe_K:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e58e65749787c70df156e9a555c84716565caad6/src/main/python/wfa/measurement/reporting/postprocessing/report/report.py#L425)):*

debug?
</blockquote></details>

Done.
___
*[`src/main/python/wfa/measurement/reporting/postprocessing/report/report.py` line 439 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1978#-OEZaY-c6CyTueltDcUe:-OEZjbwP86bkUntFJvuN:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e58e65749787c70df156e9a555c84716565caad6/src/main/python/wfa/measurement/reporting/postprocessing/report/report.py#L439)):*

debug?
</blockquote></details>

Done.

SanjayVas (changes requested):
> Reviewed 3 of 10 files at r3, 1 of 2 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1978)* status: 4 of 12 files reviewed, 14 unresolved discussions (waiting on @kungfucraig and @ple13)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/postprocessing/v2alpha/ReportProcessor.kt` line 126 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1978#-OEaILn-6_B33D1_lk5F:-OEaILn-6_B33D1_lk5G:badymof) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7d386c10e2b7bea216242ceb72101bc8c73908ac/src/main/kotlin/org/wfanet/measurement/reporting/postprocessing/v2alpha/ReportProcessor.kt#L126)):*
> ```Kotlin
> 
>       if (exitCode == 0) {
>         logger.info(processError)
> ```

@kungfucraig Do we want to always log this at this level? Maybe log at debug level here, so it'll only get logged if the JVM options for this Kotlin class specify to do so?

Basically, `verbose` specifies the log level threshold for Python but JVM flags are used to specify the log level threshold for Kotlin.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/postprocessing/v2alpha/ReportProcessor.kt` line 128 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1978#-OEaJKCnD0BvzcXNNsNC:-OEaJKCnD0BvzcXNNsND:b8owqa4) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7d386c10e2b7bea216242ceb72101bc8c73908ac/src/main/kotlin/org/wfanet/measurement/reporting/postprocessing/v2alpha/ReportProcessor.kt#L128)):*
> ```Kotlin
>         logger.info(processError)
>       } else {
>         logger.info("Failed to process the report with exitCode $exitCode.")
> ```

Log *or* throw. If you're throwing an exception, you generally don't also log the condition. It's up to the caller to choose how to handle the exception and log it if desired.

kungfucraig (changes requested):
> Reviewed 6 of 10 files at r3, 2 of 3 files at r4.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1978)* status: 8 of 12 files reviewed, 9 unresolved discussions (waiting on @ple13 and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/postprocessing/v2alpha/ReportProcessor.kt` line 126 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1978#-OEaILn-6_B33D1_lk5F:-OEaJsH92rYRCevZHurg:b36z6o2) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7d386c10e2b7bea216242ceb72101bc8c73908ac/src/main/kotlin/org/wfanet/measurement/reporting/postprocessing/v2alpha/ReportProcessor.kt#L126)):*

@kungfucraig Do we want to always log this at this level? Maybe log at debug level here, so it'll only get logged if the JVM options for this Kotlin class specify to do so?

Basically, `verbose` specifies the log level threshold for Python but JVM flags are used to specify the log level threshold for Kotlin.
</blockquote></details>

That makes sense. Use Level=fine?
___
*[`src/main/python/wfa/measurement/reporting/postprocessing/tools/post_process_origin_report.py` line 187 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1978#-OEaL4XQ3qJkdMWdYxlo:-OEaL4XQ3qJkdMWdYxlp:b-u7ufa) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7aef66c26becb19d5f74ccc717731a536dc19058/src/main/python/wfa/measurement/reporting/postprocessing/tools/post_process_origin_report.py#L187)):*
> ```Python
>           frozenset(entry.data_providers)] = measurements[0]
> 
>     # Gets the number of cumulative measurements per EDP combination.
> ```

What's going on here? This seems to be adding more than just logging.
___
*[`src/main/python/wfa/measurement/reporting/postprocessing/tools/post_process_origin_report.py` line 209 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1978#-OEaLjnn3E2R6eBcE4wb:-OEaLjno35mtgLavBilH:bi2tkur) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7aef66c26becb19d5f74ccc717731a536dc19058/src/main/python/wfa/measurement/reporting/postprocessing/tools/post_process_origin_report.py#L209)):*
> ```Python
>         "measurements by EDP combination are:" \
>         f"{cumulative_measurements_length_by_edp_combination}"
>       logging.fatal(message)
> ```

Use critical. See: https://peps.python.org/pep-0282/#levels

If it makes sense you can throw an exception here instead.
___
*[`src/main/python/wfa/measurement/reporting/postprocessing/tools/post_process_origin_report.py` line 259 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1978#-OEaLw0kBEFZyXcsdF4q:-OEaLw0kBEFZyXcsdF4r:b-34vmyd) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7aef66c26becb19d5f74ccc717731a536dc19058/src/main/python/wfa/measurement/reporting/postprocessing/tools/post_process_origin_report.py#L259)):*
> ```Python
>         if len(subset) == 0:
>           message = "The right hand side EDP combination must not be empty."
>           logging.fatal(message)
> ```

See above
___
*[`src/main/python/wfa/measurement/reporting/postprocessing/tools/post_process_origin_report.py` line 316 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1978#-OEaLyGiDIkykmWnT7b8:-OEaLyGiDIkykmWnT7b9:bqujbyh) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7aef66c26becb19d5f74ccc717731a536dc19058/src/main/python/wfa/measurement/reporting/postprocessing/tools/post_process_origin_report.py#L316)):*
> ```Python
>             f" of the left hand side measurement {superset_measurement.name}, "\
>             f"which is {superset_measurement.sigma}"
>           logging.fatal(message)
> ```

Here too
___
*[`src/main/python/wfa/measurement/reporting/postprocessing/tools/post_process_origin_report.py` line 337 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1978#-OEaMMNG1t6CvAgbX0kK:-OEaMMNG1t6CvAgbX0kL:b-3z98un) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7aef66c26becb19d5f74ccc717731a536dc19058/src/main/python/wfa/measurement/reporting/postprocessing/tools/post_process_origin_report.py#L337)):*
> ```Python
> def main():
>   # Sets the log level.
>   if '-v' in sys.argv:
> ```

You want to define your verbose flag using the flags library. See: https://abseil.io/docs/python/guides/flags

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1978)* status: 8 of 12 files reviewed, 9 unresolved discussions (waiting on @kungfucraig and @ple13)

___
*[`src/main/python/wfa/measurement/reporting/postprocessing/tools/post_process_origin_report.py` line 337 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1978#-OEaMMNG1t6CvAgbX0kK:-OEaOXlpA0RwSLx3ih4q:bb371io) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7aef66c26becb19d5f74ccc717731a536dc19058/src/main/python/wfa/measurement/reporting/postprocessing/tools/post_process_origin_report.py#L337)):*

You want to define your verbose flag using the flags library. See: https://abseil.io/docs/python/guides/flags
</blockquote></details>

The flags are already defined by the abseil logging library. You don't want to *set* them here.

kungfucraig (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1978)* status: 8 of 12 files reviewed, 9 unresolved discussions (waiting on @ple13 and @SanjayVas)

___
*[`src/main/python/wfa/measurement/reporting/postprocessing/tools/post_process_origin_report.py` line 337 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1978#-OEaMMNG1t6CvAgbX0kK:-OEaZdqH4P3kA-7VF16o:b-sttvvt) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7aef66c26becb19d5f74ccc717731a536dc19058/src/main/python/wfa/measurement/reporting/postprocessing/tools/post_process_origin_report.py#L337)):*

The flags are already defined by the abseil logging library. You don't want to *set* them here.
</blockquote></details>

Abseil defines "-v." That's bold. 

I don't think we want to have the caller pass the flag to log to stderr since not having it makes the program unusable.

We should just set that here no matter what. That means we have to write:  FLAGS.logtostderr = True



For verbosity: I suppose we can have the caller pass "-v" and let abseil consume it.

An alternative is to provide a "--debug" flag or similar in this program and set the verbosity based on its presence. That would make the caller's life easier.

ple13 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1978)* status: 6 of 12 files reviewed, 9 unresolved discussions (waiting on @kungfucraig and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/postprocessing/v2alpha/ReportProcessor.kt` line 126 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1978#-OEaILn-6_B33D1_lk5F:-OEo6dbp00hESj4Dl-Zs:b-w8k4yl) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7d386c10e2b7bea216242ceb72101bc8c73908ac/src/main/kotlin/org/wfanet/measurement/reporting/postprocessing/v2alpha/ReportProcessor.kt#L126)):*

That makes sense. Use Level=fine?
</blockquote></details>

Thanks, I've updated the log level to fine when exitcode is zero. 
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/postprocessing/v2alpha/ReportProcessor.kt` line 128 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1978#-OEaJKCnD0BvzcXNNsNC:-OEo9vFmCFyeRgfZRzgH:b-7wjzyx) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7d386c10e2b7bea216242ceb72101bc8c73908ac/src/main/kotlin/org/wfanet/measurement/reporting/postprocessing/v2alpha/ReportProcessor.kt#L128)):*

Log *or* throw. If you're throwing an exception, you generally don't also log the condition. It's up to the caller to choose how to handle the exception and log it if desired.
</blockquote></details>

I've removed the log.
___
*[`src/main/python/wfa/measurement/reporting/postprocessing/tools/post_process_origin_report.py` line 187 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1978#-OEaL4XQ3qJkdMWdYxlo:-OEoA3sNB8hrbnaGcB76:b-pm9oql) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7aef66c26becb19d5f74ccc717731a536dc19058/src/main/python/wfa/measurement/reporting/postprocessing/tools/post_process_origin_report.py#L187)):*

What's going on here? This seems to be adding more than just logging.
</blockquote></details>

I wanted to log the length of the cumulative measurements per EDP combination right when the program parses the primitive measurements. All cumulative measurements should have the same length, however, a buggy report may not meet that constraint, and needs a check.

However, I think I should do the checking in another PR.
___
*[`src/main/python/wfa/measurement/reporting/postprocessing/tools/post_process_origin_report.py` line 209 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1978#-OEaLjnn3E2R6eBcE4wb:-OEoUTuL2jUOmBaqGqS4:b-ek33zl) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7aef66c26becb19d5f74ccc717731a536dc19058/src/main/python/wfa/measurement/reporting/postprocessing/tools/post_process_origin_report.py#L209)):*

Use critical. See: https://peps.python.org/pep-0282/#levels

If it makes sense you can throw an exception here instead.
</blockquote></details>

Thanks. I'll move this to a separate PR where we check for the report issues.
___
*[`src/main/python/wfa/measurement/reporting/postprocessing/tools/post_process_origin_report.py` line 259 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1978#-OEaLw0kBEFZyXcsdF4q:-OEoFV_Q9VFMrO_AwhB3:bzgtuw3) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7aef66c26becb19d5f74ccc717731a536dc19058/src/main/python/wfa/measurement/reporting/postprocessing/tools/post_process_origin_report.py#L259)):*

See above
</blockquote></details>

I've seen a buggy report where a right hand side EDP combination is missing for a difference measurement, that's why I added this check here. However, I'll have a separate PR to do this. For now, I'll just do the logging.
___
*[`src/main/python/wfa/measurement/reporting/postprocessing/tools/post_process_origin_report.py` line 316 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1978#-OEaLyGiDIkykmWnT7b8:-OEoH4iaCrdPwOpunmMf:bmrbst1) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7aef66c26becb19d5f74ccc717731a536dc19058/src/main/python/wfa/measurement/reporting/postprocessing/tools/post_process_origin_report.py#L316)):*

Here too
</blockquote></details>

I was thinking about a buggy report when I added this check too. I'll move it to a separate PR as well.
___
*[`src/main/python/wfa/measurement/reporting/postprocessing/tools/post_process_origin_report.py` line 337 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1978#-OEaMMNG1t6CvAgbX0kK:-OEoUDH6AtTwShjzDZzm:b-pu5jbu) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7aef66c26becb19d5f74ccc717731a536dc19058/src/main/python/wfa/measurement/reporting/postprocessing/tools/post_process_origin_report.py#L337)):*

Abseil defines "-v." That's bold. 

I don't think we want to have the caller pass the flag to log to stderr since not having it makes the program unusable.

We should just set that here no matter what. That means we have to write:  FLAGS.logtostderr = True



For verbosity: I suppose we can have the caller pass "-v" and let abseil consume it.

An alternative is to provide a "--debug" flag or similar in this program and set the verbosity based on its presence. That would make the caller's life easier.
</blockquote></details>

Thanks. I've updated the code to use the --debug flag instead, and set FLAGS.logtostderr=True to always send the log to stderr.

SanjayVas (approved):
Kotlin approved

> Reviewed 5 of 6 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1978#-:-OEoj0_Z7t1Ka0uYdNcn:b-ebpbdb)* status: 11 of 12 files reviewed, 7 unresolved discussions (waiting on @kungfucraig)

kungfucraig (approved):
> Reviewed 6 of 6 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1978)* status: all files reviewed, 1 unresolved discussion (waiting on @ple13)

___
*[`src/main/python/wfa/measurement/reporting/postprocessing/tools/post_process_origin_report.py` line 34 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1978#-OEooQUoEbbT0Gs5DFE4:-OEooQUoEbbT0Gs5DFE5:b-2jis1g) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f88491f9860a4ead72aea764c56b4273c4519ab4/src/main/python/wfa/measurement/reporting/postprocessing/tools/post_process_origin_report.py#L34)):*
> ```Python
> FLAGS = flags.FLAGS
> 
> flags.DEFINE_boolean('debug', False, 'Enable debug mode.')
> ```

nit: Use double quotes for your strings to be consistent with the rest of the file.

stevenwarejones (approved):
> Reviewed 5 of 10 files at r3, 1 of 3 files at r4, 5 of 6 files at r6, 1 of 1 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1978)* status: all files reviewed, 1 unresolved discussion (waiting on @ple13)

ple13 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1978)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @ple13)

___
*[`src/main/python/wfa/measurement/reporting/postprocessing/tools/post_process_origin_report.py` line 34 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/1978#-OEooQUoEbbT0Gs5DFE4:-OEouB741Xrd24yYVrGY:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f88491f9860a4ead72aea764c56b4273c4519ab4/src/main/python/wfa/measurement/reporting/postprocessing/tools/post_process_origin_report.py#L34)):*

nit: Use double quotes for your strings to be consistent with the rest of the file.
</blockquote></details>

Done.

PR #2003, opened 2025-01-14
feat: Add libraries for Access API clients

SanjayVas:


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2003)

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2003)* status: 0 of 23 files reviewed, all discussions resolved

___
*[`src/main/kotlin/org/wfanet/measurement/access/client/v1alpha/Authorization.kt` line 77 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2003#-OGa_KX-27Hhp0uPEKbu:-OGa_KX-27Hhp0uPEKbv:b-k7iyb5) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/9e8bff715d60e644051fb64e33736155585baadd/src/main/kotlin/org/wfanet/measurement/access/client/v1alpha/Authorization.kt#L77)):*
> ```Kotlin
>     }
> 
>     checkPermissionsParallel(protectedResourceNames, principal, requiredPermissionIds).getOrThrow()
> ```

FYI @kungfucraig This is the part we discussed (and is mentioned in the design doc) about making parallel requests for checking multiple protected resources. I use it in a following PR for doing authorization checks for Get methods.

SanjayVas:
Depends on #2005

tristanvuong2021 (changes requested):
> Reviewed 15 of 23 files at r1, 3 of 5 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2003)* status: 18 of 26 files reviewed, 2 unresolved discussions (waiting on @SanjayVas)

___
*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2003#-OHOlMMZCkm9ed4kLahY:-OHOlMMZCkm9ed4kLahZ:bq9ceye) (no related file):*
The files look like they were committed this year not 2024. I see a mix of 2024 and 2025.
___
*[`src/main/kotlin/org/wfanet/measurement/access/client/ValueInScope.kt` line 43 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2003#-OHOlYPRBaG-9d0LPEGk:-OHOkRmN4OMI600bBEOI:b1nr0zs) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/71d9b55c94f6f366bb21b1ff193c8670d48a705b/src/main/kotlin/org/wfanet/measurement/access/client/ValueInScope.kt#L43)):*
> ```Kotlin
>       }
>       parts.removeLast()
>     } while (parts.isNotEmpty())
> ```

nit: This could be parts.size > 1 since a size of 1 will just be a wildcard check, which was already done earlier.

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2003)* status: 18 of 26 files reviewed, 2 unresolved discussions (waiting on @tristanvuong2021)

___
*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2003#-OHOlMMZCkm9ed4kLahY:-OHOn4r9ELz9poIvhRWL:b-7hl34e) (no related file):*

The files look like they were committed this year not 2024. I see a mix of 2024 and 2025.
</blockquote></details>

They were actually written in 2024 and were published in a previous branch that has since been deleted.

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2003)* status: 18 of 26 files reviewed, 2 unresolved discussions (waiting on @tristanvuong2021)

___
*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2003#-OHOlMMZCkm9ed4kLahY:-OHOnXQgD8PLqexOARGa:b-qett39) (no related file):*

They were actually written in 2024 and were published in a previous branch that has since been deleted.
</blockquote></details>

To clarify, *some* of the files were written/published in 2024. The dates on the files in this PR should be accurate.

tristanvuong2021 (commented):
> Reviewed 1 of 5 files at r2.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2003)* status: all files reviewed, 4 unresolved discussions (waiting on @SanjayVas)

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2003)* status: all files reviewed, 2 unresolved discussions (waiting on @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/access/client/v1alpha/Authorization.kt` line 129 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2003#-OHOyOG6-K0f1tVi3GRD:-OHPN9RU9dId5mGoc6MZ:b-sy1xmd) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/71d9b55c94f6f366bb21b1ff193c8670d48a705b/src/main/kotlin/org/wfanet/measurement/access/client/v1alpha/Authorization.kt#L129)):*
This line make result contain the exception:

> ```kotlin
>                 result =
>                   Result.failure(
>                     buildPermissionDeniedException(protectedResourceName, missingPermissionId)
>                   )
> ```

And then the `.getOrThrow()` call on the return value of this function throws it
___
*[`src/test/kotlin/org/wfanet/measurement/access/client/v1alpha/PrincipalAuthInterceptorTest.kt` line 103 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2003#-OHOqxvM2X_JjYjkL0Gm:-OHPOAu-5GkPjS5rJywF:bzc3dx1) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/71d9b55c94f6f366bb21b1ff193c8670d48a705b/src/test/kotlin/org/wfanet/measurement/access/client/v1alpha/PrincipalAuthInterceptorTest.kt#L103)):*

nit: Does using this capturedContext variable always work, or only because there is just one test using it right now.
</blockquote></details>

It will work for any test method in this class because the instance gets reinitialized for every test method invocation.

tristanvuong2021 (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2003)* status: all files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

kungfucraig (changes requested):
> Reviewed 19 of 23 files at r1, 1 of 5 files at r2.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2003)* status: all files reviewed, 5 unresolved discussions (waiting on @SanjayVas and @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/access/client/ValueInScope.kt` line 43 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2003#-OHOlYPRBaG-9d0LPEGk:-OHZVz0w9K4pet8FIIp6:b-4d5xgx) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/71d9b55c94f6f366bb21b1ff193c8670d48a705b/src/main/kotlin/org/wfanet/measurement/access/client/ValueInScope.kt#L43)):*

nit: This could be parts.size > 1 since a size of 1 will just be a wildcard check, which was already done earlier.
</blockquote></details>

Arguably it'd be easier to read to drop the "WILDCARD in scopes" at the top.
___
*[`src/main/kotlin/org/wfanet/measurement/access/client/v1alpha/Authorization.kt` line 42 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2003#-OHZYeCtC4CX_bV2XsV5:-OHZW_1P7seorOKYeYWs:bcsic8c) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/71d9b55c94f6f366bb21b1ff193c8670d48a705b/src/main/kotlin/org/wfanet/measurement/access/client/v1alpha/Authorization.kt#L42)):*
> ```Kotlin
>    * protected resources.
>    *
>    * @param protectedResourceNames names of the protected resources, in order of most to least
> ```

This bothers me a bit. For one, you don't check the precondition, nor should you because it's expensive to do so. 

Do we always (only typically perhaps) expect the first item to be the resource itself?

nit: a Collection may be unordered, so I'm not exactly sure what the "first" method is supposed to do. Can you use List?
___
*[`src/main/kotlin/org/wfanet/measurement/access/client/v1alpha/Authorization.kt` line 48 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2003#-OHZX6Ho3lgUKWxI4rfI:-OHZX6Ho3lgUKWxI4rfJ:b-s7m8y) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/71d9b55c94f6f366bb21b1ff193c8670d48a705b/src/main/kotlin/org/wfanet/measurement/access/client/v1alpha/Authorization.kt#L48)):*
> ```Kotlin
>    * @throws io.grpc.StatusRuntimeException
>    */
>   suspend fun check(
> ```

We need to monitor the latency of this function.
___
*[`src/main/kotlin/org/wfanet/measurement/access/client/v1alpha/Authorization.kt` line 71 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2003#-OHZauCA3g2JUkODaMUX:-OHZauCA3g2JUkODaMUY:b-jekg4l) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/71d9b55c94f6f366bb21b1ff193c8670d48a705b/src/main/kotlin/org/wfanet/measurement/access/client/v1alpha/Authorization.kt#L71)):*
> ```Kotlin
>         checkPermissions(principal, primaryProtectedResourceName, requiredPermissionIds)
>       if (response.permissionsCount != requiredPermissionIds.size) {
>         val missingPermissionId: String = getMissingPermissionId(response, requiredPermissionIds)
> ```

There may be multiple. We should find them all since it makes it easier to triage the error if you know everything you're missing.
___
*[`src/main/kotlin/org/wfanet/measurement/access/client/v1alpha/Authorization.kt` line 163 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2003#-OHZaZVC8Z7yHBAkHASx:-OHZaZVC8Z7yHBAkHASy:byu4v8i) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/71d9b55c94f6f366bb21b1ff193c8670d48a705b/src/main/kotlin/org/wfanet/measurement/access/client/v1alpha/Authorization.kt#L163)):*
> ```Kotlin
>     val notInScope = ValueInScope(scopes).negate()::test
> 
>     val missingPermissionId = requiredPermissionIds.find(notInScope)
> ```

It would be helpful to find all of the permissions that are missing so that the client doesn't have to play whack-a-mole with permissions.

kungfucraig (commented):
> Reviewed 2 of 23 files at r1, 4 of 5 files at r2.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2003)* status: all files reviewed, 2 unresolved discussions (waiting on @SanjayVas and @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/access/client/v1alpha/Authorization.kt` line 42 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2003#-OHZYeCtC4CX_bV2XsV5:-OHde7lnA8lsuBULCbdx:bx8h65e) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/71d9b55c94f6f366bb21b1ff193c8670d48a705b/src/main/kotlin/org/wfanet/measurement/access/client/v1alpha/Authorization.kt#L42)):*

> This bothers me a bit. For one, you don't check the precondition, nor should you because it's expensive to do so.

Not just expensive, but impossible. The Access API itself doesn't know about parent-child relationships. That's something that is determined by the caller.

Really this is just so that we know which resource to put in the error message, as we always want it to be the most-specific one.

> Do we always (only typically perhaps) expect the first item to be the resource itself?

Often, but not always. It depends on what resource types a permission may be defined. For example, a Create permission never makes sense at the specific resource level. So this would at the very least be at the parent level, but you could for example only allow it to be granted at an even higher level. This is true for DataProviders for example, which is granted at the root level.

Again, the Access API doesn't actually know anything about these parent-child relationships. It furthermore doesn't understand the semantics of any permission. For example, instead of "create" you could have a coarser "writer" permission at some higher level that just lets you do any write mutations under that resource.

> nit: a Collection may be unordered

This is not true. `Collection` extends `Iterable`, which means it has a defined iteration order. The underlying implementation may not guarantee that e.g. the order matches the insertion order, but there's still always an order.
</blockquote></details>

Collection has a defined iteration order, but you are asking the client to provide a container whose iteration order matches the order they specified during construction.

But it looks like the ordering isn't that important. Perhaps update the documentation to say the first item in the input is used for error messages?
___
*[`src/main/kotlin/org/wfanet/measurement/access/client/v1alpha/Authorization.kt` line 71 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2003#-OHZauCA3g2JUkODaMUX:-OHdeteoBziq5KqgHTQ7:b32wwp) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/71d9b55c94f6f366bb21b1ff193c8670d48a705b/src/main/kotlin/org/wfanet/measurement/access/client/v1alpha/Authorization.kt#L71)):*

I can add a TODO to enhance this, since I intentionally give the ability to find all of them from the CheckPermissions response. For now, I'm just doing the minimum to satisfy the guidance in [AIP-211](https://google.aip.dev/211)

> the corresponding error message **should** look like: "Permission `{p}` denied on resource `{r}` (or it might not exist)."
</blockquote></details>

sg
___
*[`src/main/kotlin/org/wfanet/measurement/access/client/v1alpha/Authorization.kt` line 163 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2003#-OHZaZVC8Z7yHBAkHASx:-OHdeytc0I-jQt5h24zI:bn14wlo) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/71d9b55c94f6f366bb21b1ff193c8670d48a705b/src/main/kotlin/org/wfanet/measurement/access/client/v1alpha/Authorization.kt#L163)):*

Marking this resolved here since it's duplicated in the above comment.
</blockquote></details>

drop a TODO here too?

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2003)* status: all files reviewed, 2 unresolved discussions (waiting on @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/access/client/v1alpha/Authorization.kt` line 42 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2003#-OHZYeCtC4CX_bV2XsV5:-OHdpdq34eINrkm1UVj6:b-rs6km3) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/71d9b55c94f6f366bb21b1ff193c8670d48a705b/src/main/kotlin/org/wfanet/measurement/access/client/v1alpha/Authorization.kt#L42)):*
> Collection has a defined iteration order, but you are asking the client to provide a container whose iteration order matches the order they specified during construction.

No, just passing a container where the iteration order is that of specificity descending. Whether they implement that with a collection that uses the construction order or whether they are sorting by some specificity comparator is not relevant, just that the precondition is that they ensure this order before calling.

> But it looks like the ordering isn't that important. Perhaps update the documentation to say the first item in the input is used for error messages?

I'd rather just keep it as ordered by specificity and just have the fact that the most specific is used for error messages as an implementation detail. Defining an order is better to me than arbitrarily saying a specific item has to be first and the order doesn't matter after that.

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2003)* status: 24 of 26 files reviewed, 1 unresolved discussion (waiting on @kungfucraig and @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/access/client/ValueInScope.kt` line 43 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2003#-OHOlYPRBaG-9d0LPEGk:-OHeReWQ60E2sVsK4z1e:boy7eai) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/71d9b55c94f6f366bb21b1ff193c8670d48a705b/src/main/kotlin/org/wfanet/measurement/access/client/ValueInScope.kt#L43)):*

Arguably it'd be easier to read to drop the "WILDCARD in scopes" at the top.
</blockquote></details>

Documented the `WILDCARD in scopes` check as an optimization, implying it can be safely removed without changing behavior.
___
*[`src/main/kotlin/org/wfanet/measurement/access/client/v1alpha/Authorization.kt` line 48 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2003#-OHZX6Ho3lgUKWxI4rfI:-OHeRHSa4YIRZ7YelhMB:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/71d9b55c94f6f366bb21b1ff193c8670d48a705b/src/main/kotlin/org/wfanet/measurement/access/client/v1alpha/Authorization.kt#L48)):*

We need to monitor the latency of this function.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/access/client/v1alpha/Authorization.kt` line 71 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2003#-OHZauCA3g2JUkODaMUX:-OHe8zh96tdsppXmD0Kc:brnnllu) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/71d9b55c94f6f366bb21b1ff193c8670d48a705b/src/main/kotlin/org/wfanet/measurement/access/client/v1alpha/Authorization.kt#L71)):*

sg
</blockquote></details>

Done (put the single TODO on buildPermissionDeniedException)
___
*[`src/main/kotlin/org/wfanet/measurement/access/client/v1alpha/Authorization.kt` line 163 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2003#-OHZaZVC8Z7yHBAkHASx:-OHe93kq7LH8v48Ab0YI:brnnllu) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/71d9b55c94f6f366bb21b1ff193c8670d48a705b/src/main/kotlin/org/wfanet/measurement/access/client/v1alpha/Authorization.kt#L163)):*

drop a TODO here too?
</blockquote></details>

Done (put the single TODO on buildPermissionDeniedException)

kungfucraig (approved):
> Reviewed 10 of 10 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2003)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @tristanvuong2021)

stevenwarejones (approved):
> Reviewed 16 of 23 files at r1, 1 of 5 files at r2, 6 of 10 files at r3, 9 of 9 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2003)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @tristanvuong2021)

PR #2024, opened 2025-01-24
fix: Ensure TlsClient principals grant access to root resource

SanjayVas:
This ensures that callers using mTLS auth can still call the proxied DataProviders and EventGroupMetadataDescriptors services.

PR #2035, opened 2025-01-28
fix: Ensure TlsClient principals grant access to root resource 

SanjayVas:
This ensures that callers using mTLS auth can still call the proxied DataProviders and EventGroupMetadataDescriptors services.

Replaces #2024


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2035)

PR #2054, opened 2025-02-05
feat: Create resources for Population Requisition Fulfiller k8 test

jojijac0b:
Create resources for k8 test. This includes population spec and eventMessageDescriptor. It also adds in flags that will be used in later PRs


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2054)

PR #2077, opened 2025-02-13
fix: Assign named IP address to Access public API server

SanjayVas:


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2077)

PR #2085, opened 2025-02-24
feat: setup github workflows to deploy control plane infra

Marco-Premier:
Add github workflows to deploy a new gke cluster for control plane API with access to spanner DB.


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2085)

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2085)* status: 0 of 19 files reviewed, 2 unresolved discussions (waiting on @jojijac0b and @stevenwarejones)

___
*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2085#-OJu81aND7ax71gXRej9:-OJu81aO8zjQoxJD3Bvl:b-9jtof0) (no related file):*
Just "control plane" or "control-plane" is too generic for the references here that are not in the "securecomputation" package/namespace. Many APIs are control plane APIs.
___
*[`src/main/docker/images.bzl` line 304 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2085#-OJu8Qzg-NvWcWIDnbLZ:-OJu8Qzh6iA2Sjih_j1K:bnt7iz8) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/9e5d99799ea247d18804429ed61c88a948a1d3a1/src/main/docker/images.bzl#L304)):*
> ```Starlark
>         name = "gcloud_control_plane_server_image",
>         image = "//src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha:google_pub_sub_work_items_service_image",
>         repository = _PREFIX + "/control-plane/v1/v1alpha-api",
> ```

Put this under "secure-computation". Note that the extra version in the path is an anomaly for Reporting due to some backend conflicts while we had both versions running.



_Suggestion:_
```Starlark
/secure-computation/control-plane-api-v1alpha
```

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2085)* status: 0 of 19 files reviewed, 3 unresolved discussions (waiting on @jojijac0b, @Marco-Premier, and @stevenwarejones)

___
*[`src/main/terraform/gcloud/modules/control-plane/main.tf` line 15 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2085#-OJuD0zc7JidWZB_Gmzk:-OJuD0zc7JidWZB_Gmzl:btfhju7) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/9e5d99799ea247d18804429ed61c88a948a1d3a1/src/main/terraform/gcloud/modules/control-plane/main.tf#L15)):*
> ```HCL
> 
> module "control_plane_internal" {
> ```

What's the justification for this not following our existing pattern/principle of having separate public and internal API servers? So far we've done that so that the server that has a public IP is not the one that has access to the DB.

Note that there's two potentially independent pieces here:

1. Having separate public and internal APIs. This is a best practice that we follow to insulate the persistence layer and ensure that both APIs can evolve independently. The TEE control plane API is currently a public API in that it's called from another system. I haven't yet seen an internal API definition.
2. Having separate servers for the public and internal APIs (as opposed to translating within the same process). This isn't *strictly* necessary, but it's what we've done for all APIs so far as it gives some benefits. From the security perspective, only the internal API server is allowed to access the DB. The tradeoff is having to maintain another image, deployment, etc.

Public APIs are versioned and follow the AIPs. Public API messages must never be persisted, while internal API messages can. Finally, the public API server tends to do end user auth checks while the internal API does not, relying on trusted mTLS credentials.

PR #2091, opened 2025-02-26
feat: add proto definition for control plane persistence

Marco-Premier:
- Add proto definitions for control plane internal api.
- Update proto definition for public control plane api to support state


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2091)

SanjayVas (changes requested):
> Reviewed 9 of 10 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2091)* status: 9 of 10 files reviewed, 9 unresolved discussions (waiting on @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/securecomputation/controlplane/v1alpha/work_item_attempt.proto` line 61 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2091#-OK2X43u9H20JTJU8ZQM:-OK2X43v6SMVeBE84Wb1:bpvci8y) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4fc02c25c9e9382db143cef8a60d4c309d8d76d7/src/main/proto/wfa/measurement/securecomputation/controlplane/v1alpha/work_item_attempt.proto#L61)):*
> ```proto
>   int32 attempt_number = 3 [(google.api.field_behavior) = OUTPUT_ONLY];
> 
>   // Optional. Log messages recorded by the subscriber for this `WorkItemAttempt`.
> ```

nit: worker? There is no subscriber role in the API. That's part of the implementation detail of Pub/Sub.



_Code quote:_
```proto
subscriber
```
___
*[`src/main/proto/wfa/measurement/securecomputation/controlplane/v1alpha/work_item_attempt.proto` line 61 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2091#-OK2XMfn7Zys8zPPtyKQ:-OK2XMfn7Zys8zPPtyKR:b-i5ac3p) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4fc02c25c9e9382db143cef8a60d4c309d8d76d7/src/main/proto/wfa/measurement/securecomputation/controlplane/v1alpha/work_item_attempt.proto#L61)):*
> ```proto
>   int32 attempt_number = 3 [(google.api.field_behavior) = OUTPUT_ONLY];
> 
>   // Optional. Log messages recorded by the subscriber for this `WorkItemAttempt`.
> ```

Use the field behavior annotation to indicate optional.

Our older APIs use the convention of any field not being marked REQUIRED as being optional, but [AIP-203](https://google.aip.dev/203) was updated in 2023 to always specify field behavior and use OPTIONAL for the relevant fields.



_Code quote:_
```proto
Optional
```
___
*[`src/main/proto/wfa/measurement/securecomputation/controlplane/v1alpha/work_item_attempt.proto` line 62 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2091#-OK2WgT77FLZHUrhkFGC:-OK2WgT77FLZHUrhkFGD:bucjg5q) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4fc02c25c9e9382db143cef8a60d4c309d8d76d7/src/main/proto/wfa/measurement/securecomputation/controlplane/v1alpha/work_item_attempt.proto#L62)):*
> ```proto
> 
>   // Optional. Log messages recorded by the subscriber for this `WorkItemAttempt`.
>   string logs = 4 [(google.api.field_behavior) = OUTPUT_ONLY];
> ```

If this is OUTPUT_ONLY, where does it get written?

I'd personally only expect this to be set on failure, meaning it can be OUTPUT_ONLY but specifically included in `FailWorkItemAttemptRequest`. In that case, `error_message` may be clearer. This would be a similar pattern to what we have for Computation attempts.
___
*[`src/main/proto/wfa/measurement/internal/securecomputation/controlplane/v1alpha/work_item.proto` line 17 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2091#-OK2QLHZ0g6wLyuQDLPb:-OK2QLHZ0g6wLyuQDLPc:brdyuqi) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4fc02c25c9e9382db143cef8a60d4c309d8d76d7/src/main/proto/wfa/measurement/internal/securecomputation/controlplane/v1alpha/work_item.proto#L17)):*
> ```proto
> syntax = "proto3";
> 
> package wfa.measurement.internal.securecomputation.controlplane.v1alpha;
> ```

Internal APIs are not versioned. (Reporting is an exception as it was in an odd state when the public v2 was being added, and cleaning it up hasn't been a big priority).
___
*[`src/main/proto/wfa/measurement/internal/securecomputation/controlplane/v1alpha/work_item.proto` line 30 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2091#-OK2QSe3-LHkB9eaR43C:-OK2QSe3-LHkB9eaR43D:b-wa0uvc) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4fc02c25c9e9382db143cef8a60d4c309d8d76d7/src/main/proto/wfa/measurement/internal/securecomputation/controlplane/v1alpha/work_item.proto#L30)):*
> ```proto
> message WorkItem {
> 
>   fixed64 external_work_item_id = 1;
> ```

For new APIs, we've switched to the term "resource ID" over "external ID" for clarity. See the [internal Access API](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/main/src/main/proto/wfa/measurement/internal/access/) as the most recent example of an internal API.



_Suggestion:_
```proto
work_item_resource_id
```
___
*[`src/main/proto/wfa/measurement/internal/securecomputation/controlplane/v1alpha/work_item.proto` line 33 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2091#-OK2SJ_2CGF6OiIY2GaS:-OK2SJ_2CGF6OiIY2GaT:bwokukn) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4fc02c25c9e9382db143cef8a60d4c309d8d76d7/src/main/proto/wfa/measurement/internal/securecomputation/controlplane/v1alpha/work_item.proto#L33)):*
> ```proto
> 
>   // The queue into which this `WorkItem` must be enqueued.
>   string queue = 2;
> ```

Queue is a resource in the public API, which means it should always be referenced by ID in the internal API.



_Suggestion:_
```proto
queue_resource_id
```
___
*[`src/main/proto/wfa/measurement/internal/securecomputation/controlplane/v1alpha/BUILD.bazel` line 17 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2091#-OK2RKZK6BQw9fHWtvFn:-OK2RKZK6BQw9fHWtvFo:by8kvga) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4fc02c25c9e9382db143cef8a60d4c309d8d76d7/src/main/proto/wfa/measurement/internal/securecomputation/controlplane/v1alpha/BUILD.bazel#L17)):*
> ```Starlark
> #Resources and shared message types.
> 
> proto_and_kt_jvm_proto_library(
> ```

nit: Consider following the newer pattern where we define the proto_library targets and then use list comprehensions for the language-specific targets. See the recently refactored [BUILD file for the Reporting API](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/main/src/main/proto/wfa/measurement/reporting/v2alpha/BUILD.bazel) as an example.
___
*[`src/main/proto/wfa/measurement/securecomputation/controlplane/v1alpha/work_item.proto` line 72 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2091#-OK2V3u02_SOQZXKXh6c:-OK2V3u02_SOQZXKXh6d:b-azce4j) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4fc02c25c9e9382db143cef8a60d4c309d8d76d7/src/main/proto/wfa/measurement/securecomputation/controlplane/v1alpha/work_item.proto#L72)):*
> ```proto
> 
>   // State of this `WorkItem`.
>   State state = 4 [(google.api.field_behavior) = OUTPUT_ONLY];
> ```

How does this relate to the attempt state?

More broadly, document the relationship between WorkItems and WorkItemAttempts somewhere in this API definition. e.g. what does an attempt represent and what's the lifecycle? The results of all the design discussion needs to end up here.
___
*[`src/main/proto/wfa/measurement/securecomputation/controlplane/v1alpha/work_item.proto` line 75 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2091#-OK2UL828vzBYXViyjfg:-OK2UL828vzBYXViyjfh:b-w1g7sy) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4fc02c25c9e9382db143cef8a60d4c309d8d76d7/src/main/proto/wfa/measurement/securecomputation/controlplane/v1alpha/work_item.proto#L75)):*
> ```proto
> 
>   // The attempts to process this `WorkItem`.
>   repeated WorkItemAttempt work_item_attempts = 5;
> ```

Attempt is another resource type. Child attempts should be listed separately.

Marco-Premier (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2091)* status: 1 of 11 files reviewed, 7 unresolved discussions (waiting on @SanjayVas and @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/securecomputation/controlplane/v1alpha/work_item.proto` line 72 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2091#-OK2V3u02_SOQZXKXh6c:-OK2hg2T97d7kz0aYYwo:b-ou4cox) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4fc02c25c9e9382db143cef8a60d4c309d8d76d7/src/main/proto/wfa/measurement/securecomputation/controlplane/v1alpha/work_item.proto#L72)):*

How does this relate to the attempt state?

More broadly, document the relationship between WorkItems and WorkItemAttempts somewhere in this API definition. e.g. what does an attempt represent and what's the lifecycle? The results of all the design discussion needs to end up here.
</blockquote></details>

Added description on `work_item_attempt.proto`
___
*[`src/main/proto/wfa/measurement/securecomputation/controlplane/v1alpha/work_item.proto` line 75 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2091#-OK2UL828vzBYXViyjfg:-OK2deryAHLMX2wnT9L6:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4fc02c25c9e9382db143cef8a60d4c309d8d76d7/src/main/proto/wfa/measurement/securecomputation/controlplane/v1alpha/work_item.proto#L75)):*

Attempt is another resource type. Child attempts should be listed separately.
</blockquote></details>

Done.
___
*[`src/main/proto/wfa/measurement/securecomputation/controlplane/v1alpha/work_item_attempt.proto` line 61 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2091#-OK2XMfn7Zys8zPPtyKQ:-OK2bOYF-TWv-6_qgLkM:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4fc02c25c9e9382db143cef8a60d4c309d8d76d7/src/main/proto/wfa/measurement/securecomputation/controlplane/v1alpha/work_item_attempt.proto#L61)):*

Use the field behavior annotation to indicate optional.

Our older APIs use the convention of any field not being marked REQUIRED as being optional, but [AIP-203](https://google.aip.dev/203) was updated in 2023 to always specify field behavior and use OPTIONAL for the relevant fields.
</blockquote></details>

Done.
___
*[`src/main/proto/wfa/measurement/securecomputation/controlplane/v1alpha/work_item_attempt.proto` line 62 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2091#-OK2WgT77FLZHUrhkFGC:-OK2dTfACCIuHBkvR7fN:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4fc02c25c9e9382db143cef8a60d4c309d8d76d7/src/main/proto/wfa/measurement/securecomputation/controlplane/v1alpha/work_item_attempt.proto#L62)):*

If this is OUTPUT_ONLY, where does it get written?

I'd personally only expect this to be set on failure, meaning it can be OUTPUT_ONLY but specifically included in `FailWorkItemAttemptRequest`. In that case, `error_message` may be clearer. This would be a similar pattern to what we have for Computation attempts.
</blockquote></details>

Done.
___
*[`src/main/proto/wfa/measurement/internal/securecomputation/controlplane/v1alpha/work_item.proto` line 17 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2091#-OK2QLHZ0g6wLyuQDLPb:-OK5_yKz8VebwRzayKp1:b-1g62sm) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4fc02c25c9e9382db143cef8a60d4c309d8d76d7/src/main/proto/wfa/measurement/internal/securecomputation/controlplane/v1alpha/work_item.proto#L17)):*

Internal APIs are not versioned. (Reporting is an exception as it was in an odd state when the public v2 was being added, and cleaning it up hasn't been a big priority).
</blockquote></details>

Done. But I'm curious, why don't we version them? Shouldn't we allow them for "breaking changes" or support backward compatibility?
___
*[`src/main/proto/wfa/measurement/internal/securecomputation/controlplane/v1alpha/work_item.proto` line 30 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2091#-OK2QSe3-LHkB9eaR43C:-OK5b_HW5IbWGrT3vHw6:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4fc02c25c9e9382db143cef8a60d4c309d8d76d7/src/main/proto/wfa/measurement/internal/securecomputation/controlplane/v1alpha/work_item.proto#L30)):*

For new APIs, we've switched to the term "resource ID" over "external ID" for clarity. See the [internal Access API](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/main/src/main/proto/wfa/measurement/internal/access/) as the most recent example of an internal API.
</blockquote></details>

Done.
___
*[`src/main/proto/wfa/measurement/internal/securecomputation/controlplane/v1alpha/work_item.proto` line 33 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2091#-OK2SJ_2CGF6OiIY2GaS:-OK2eB-7-GRzLPlk2l_Z:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4fc02c25c9e9382db143cef8a60d4c309d8d76d7/src/main/proto/wfa/measurement/internal/securecomputation/controlplane/v1alpha/work_item.proto#L33)):*

Queue is a resource in the public API, which means it should always be referenced by ID in the internal API.
</blockquote></details>

Done.

SanjayVas (changes requested):
> Reviewed 1 of 10 files at r1, 9 of 9 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2091)* status: all files reviewed, 8 unresolved discussions (waiting on @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/internal/securecomputation/controlplane/v1alpha/work_item.proto` line 17 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2091#-OK2QLHZ0g6wLyuQDLPb:-OK7hhRw9jZSUg8WyV-z:b-g4r3oq) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4fc02c25c9e9382db143cef8a60d4c309d8d76d7/src/main/proto/wfa/measurement/internal/securecomputation/controlplane/v1alpha/work_item.proto#L17)):*

Done. But I'm curious, why don't we version them? Shouldn't we allow them for "breaking changes" or support backward compatibility?
</blockquote></details>

The idea is that the same internal API should be able to support multiple public API versions. For example, you generally don't want to create a new DB for a new API version - the new version *should* be able to read existing data written from the old version.

Again, Reporting was an exception as we did a major overhaul for v2 where the two versions were actually entirely different systems that didn't interact.
___
*[`src/main/proto/wfa/measurement/securecomputation/controlplane/v1alpha/work_item.proto` line 58 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2091#-OK7qqANDfCfHVlZ6HTp:-OK7qqANDfCfHVlZ6HTq:b-5e5weh) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/bee50217392019303293c7b124ec76eb73de337d/src/main/proto/wfa/measurement/securecomputation/controlplane/v1alpha/work_item.proto#L58)):*
> ```proto
>     STATE_UNSPECIFIED = 0;
> 
>     // The `WorkItem` has been queued.
> ```

I think you can add a note here that this means that there are no `WorkItemAttempt`s for this `WorkItem`.
___
*[`src/main/proto/wfa/measurement/securecomputation/controlplane/v1alpha/work_item.proto` line 61 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2091#-OK7r8QG6ExNUgeBUoHo:-OK7r8QG6ExNUgeBUoHp:bkwvmcb) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/bee50217392019303293c7b124ec76eb73de337d/src/main/proto/wfa/measurement/securecomputation/controlplane/v1alpha/work_item.proto#L61)):*
> ```proto
>     QUEUED = 1;
> 
>     // The `WorkItem` has been delivered to the subscriber.
> ```

I think this can be redefined as meaning that there is at least one `WorkItemAttempt` for this `WorkItem`.
___
*[`src/main/proto/wfa/measurement/securecomputation/controlplane/v1alpha/work_item_attempts_service.proto` line 34 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2091#-OK7s_We0pPe0AwsgqQt:-OK7s_Wf61xhgmxgQTef:bit3dus) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/bee50217392019303293c7b124ec76eb73de337d/src/main/proto/wfa/measurement/securecomputation/controlplane/v1alpha/work_item_attempts_service.proto#L34)):*
> ```proto
>   // Creates a new `WorkItemAttempt` with state `CLAIMED`.
>   // Set the state in the parent `WorkItem` as `CLAIMED`.
>   rpc CreateWorkItemAttempt(CreateWorkItemAttemptRequest) returns (WorkItemAttempt) {
> ```

Maybe note here that this fails any existing WorkItemAttempt?
___
*[`src/main/proto/wfa/measurement/securecomputation/controlplane/v1alpha/work_item_attempts_service.proto` line 67 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2091#-OK7pE2d9V9dn2ouY_PJ:-OK7pE2d9V9dn2ouY_PK:bl4yj40) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/bee50217392019303293c7b124ec76eb73de337d/src/main/proto/wfa/measurement/securecomputation/controlplane/v1alpha/work_item_attempts_service.proto#L67)):*
> ```proto
>   // Required. The parent resource where this `WorkItemAttempt` will be created.
>   // Format: "workItems/{work_item}"
>   string parent = 1 [(google.api.field_behavior) = REQUIRED];
> ```

This also needs a `work_item_attempt` field for the resource being created.
___
*[`src/main/proto/wfa/measurement/securecomputation/controlplane/v1alpha/work_item_attempts_service.proto` line 113 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2091#-OK7jA8q2RM788wUr4Sd:-OK7jA8q2RM788wUr4Se:bp1zbua) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/bee50217392019303293c7b124ec76eb73de337d/src/main/proto/wfa/measurement/securecomputation/controlplane/v1alpha/work_item_attempts_service.proto#L113)):*
> ```proto
> 
>   // Human-readable error message. This should not include any sensitive info.
>   string error_message = 2;
> ```

I assume you want this to be OPTIONAL here?
___
*[`src/main/proto/wfa/measurement/securecomputation/controlplane/v1alpha/work_item_attempt.proto` line 54 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2091#-OK7o4_-7s73jD2ayj3t:-OK7o4_-7s73jD2ayj3u:bhqpfrr) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/bee50217392019303293c7b124ec76eb73de337d/src/main/proto/wfa/measurement/securecomputation/controlplane/v1alpha/work_item_attempt.proto#L54)):*
> ```proto
> 
>     // The `WorkItem` has been delivered to the subscriber.
>     CLAIMED = 1;
> ```

I had made a similar comment to this on the design doc, but I think `CLAIMED` may not be accurate for what is being represented here. When I first suggested the term, I was thinking along the lines of how we track Computations in the Duchy but this is a bit different.

Rather, I think perhaps `ACTIVE` makes more sense, especially as `ClaimWorkItem` in the original design has been replaced with `CreateWorkItemAttempt`. This also fits with the idea that there will be at most one `ACTIVE` attempt at a time.

Finally, we shouldn't be noting anything about delivery or subscribers here as the fact that we're using a message system is an implementation detail. e.g. you could theoretically use polling over this API and still get similar functionality with less efficiency.



_Code quote:_
```proto
    // The `WorkItem` has been delivered to the subscriber.
    CLAIMED = 1;
```
___
*[`src/main/proto/wfa/measurement/securecomputation/controlplane/v1alpha/work_items_service.proto` line 47 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2091#-OK7prQdBsbmJD0Zy_VJ:-OK7prQdBsbmJD0Zy_VK:bxv72lg) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/bee50217392019303293c7b124ec76eb73de337d/src/main/proto/wfa/measurement/securecomputation/controlplane/v1alpha/work_items_service.proto#L47)):*
> ```proto
>   rpc ListWorkItems(ListWorkItemsRequest) returns (ListWorkItemsResponse) {}
> 
>   // Set the `WorkItem` status to FAILED, in case of a non-retriable error.
> ```

Maybe note that this will also fail the active WorkItemAttempt?
___
*[`src/main/proto/wfa/measurement/internal/securecomputation/controlplane/work_item_attempt.proto` line 58 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2091#-OK7ixFMAHzOgYPU5M0g:-OK7ixFN6oKE21mLVRNz:b-10nw90) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/bee50217392019303293c7b124ec76eb73de337d/src/main/proto/wfa/measurement/internal/securecomputation/controlplane/work_item_attempt.proto#L58)):*
> ```proto
>     (google.api.field_behavior) = OUTPUT_ONLY,
>     (google.api.field_behavior) = OPTIONAL
>   ];;
> ```

extra semicolon



_Code quote:_
```proto
;
```

Marco-Premier (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2091)* status: 3 of 11 files reviewed, 8 unresolved discussions (waiting on @SanjayVas and @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/securecomputation/controlplane/v1alpha/work_item.proto` line 58 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2091#-OK7qqANDfCfHVlZ6HTp:-OKAUfd26XORAeVNahqy:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/bee50217392019303293c7b124ec76eb73de337d/src/main/proto/wfa/measurement/securecomputation/controlplane/v1alpha/work_item.proto#L58)):*

I think you can add a note here that this means that there are no `WorkItemAttempt`s for this `WorkItem`.
</blockquote></details>

Done.
___
*[`src/main/proto/wfa/measurement/securecomputation/controlplane/v1alpha/work_item.proto` line 61 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2091#-OK7r8QG6ExNUgeBUoHo:-OKAV3f-9A_juBts5403:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/bee50217392019303293c7b124ec76eb73de337d/src/main/proto/wfa/measurement/securecomputation/controlplane/v1alpha/work_item.proto#L61)):*

I think this can be redefined as meaning that there is at least one `WorkItemAttempt` for this `WorkItem`.
</blockquote></details>

Done.
___
*[`src/main/proto/wfa/measurement/securecomputation/controlplane/v1alpha/work_item_attempt.proto` line 54 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2091#-OK7o4_-7s73jD2ayj3t:-OKAWBso5d6ec7QXYACL:b7tqdfp) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/bee50217392019303293c7b124ec76eb73de337d/src/main/proto/wfa/measurement/securecomputation/controlplane/v1alpha/work_item_attempt.proto#L54)):*

I had made a similar comment to this on the design doc, but I think `CLAIMED` may not be accurate for what is being represented here. When I first suggested the term, I was thinking along the lines of how we track Computations in the Duchy but this is a bit different.

Rather, I think perhaps `ACTIVE` makes more sense, especially as `ClaimWorkItem` in the original design has been replaced with `CreateWorkItemAttempt`. This also fits with the idea that there will be at most one `ACTIVE` attempt at a time.

Finally, we shouldn't be noting anything about delivery or subscribers here as the fact that we're using a message system is an implementation detail. e.g. you could theoretically use polling over this API and still get similar functionality with less efficiency.
</blockquote></details>

Done. While the `work_item.proto` can remain `CLAIMED` right? as what's `active` is the attempt and not the `WorkItem`
___
*[`src/main/proto/wfa/measurement/securecomputation/controlplane/v1alpha/work_item_attempts_service.proto` line 34 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2091#-OK7s_We0pPe0AwsgqQt:-OKAWtFmAkeS20ORMez4:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/bee50217392019303293c7b124ec76eb73de337d/src/main/proto/wfa/measurement/securecomputation/controlplane/v1alpha/work_item_attempts_service.proto#L34)):*

Maybe note here that this fails any existing WorkItemAttempt?
</blockquote></details>

Done.
___
*[`src/main/proto/wfa/measurement/securecomputation/controlplane/v1alpha/work_item_attempts_service.proto` line 67 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2091#-OK7pE2d9V9dn2ouY_PJ:-OKAXCjw2sH2T8uc-m_P:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/bee50217392019303293c7b124ec76eb73de337d/src/main/proto/wfa/measurement/securecomputation/controlplane/v1alpha/work_item_attempts_service.proto#L67)):*

This also needs a `work_item_attempt` field for the resource being created.
</blockquote></details>

Done.
___
*[`src/main/proto/wfa/measurement/securecomputation/controlplane/v1alpha/work_item_attempts_service.proto` line 113 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2091#-OK7jA8q2RM788wUr4Sd:-OKAXrHxBzoMEE7wx4v1:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/bee50217392019303293c7b124ec76eb73de337d/src/main/proto/wfa/measurement/securecomputation/controlplane/v1alpha/work_item_attempts_service.proto#L113)):*

I assume you want this to be OPTIONAL here?
</blockquote></details>

Done.
___
*[`src/main/proto/wfa/measurement/securecomputation/controlplane/v1alpha/work_items_service.proto` line 47 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2091#-OK7prQdBsbmJD0Zy_VJ:-OKAYFJX0H0zQtITEKI9:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/bee50217392019303293c7b124ec76eb73de337d/src/main/proto/wfa/measurement/securecomputation/controlplane/v1alpha/work_items_service.proto#L47)):*

Maybe note that this will also fail the active WorkItemAttempt?
</blockquote></details>

Done.
___
*[`src/main/proto/wfa/measurement/internal/securecomputation/controlplane/work_item_attempt.proto` line 58 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2091#-OK7ixFMAHzOgYPU5M0g:-OKAUGcX0-qs-XsNLrC2:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/bee50217392019303293c7b124ec76eb73de337d/src/main/proto/wfa/measurement/internal/securecomputation/controlplane/work_item_attempt.proto#L58)):*

extra semicolon
</blockquote></details>

Done.

SanjayVas (changes requested):
> Reviewed 9 of 9 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2091)* status: all files reviewed, 1 unresolved discussion (waiting on @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/securecomputation/controlplane/v1alpha/work_item_attempt.proto` line 54 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2091#-OK7o4_-7s73jD2ayj3t:-OKCkogAAJaKcVLj-iK9:b-fgx3wo) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/bee50217392019303293c7b124ec76eb73de337d/src/main/proto/wfa/measurement/securecomputation/controlplane/v1alpha/work_item_attempt.proto#L54)):*

Done. While the `work_item.proto` can remain `CLAIMED` right? as what's `active` is the attempt and not the `WorkItem`
</blockquote></details>

I don't think `CLAIMED` make sense either since there's not really a concept of "claiming" (e.g. having an exclusive lock on) a work item. Looking at the example list in [AIP-216](https://google.aip.dev/216), maybe `RUNNING`?
___
*[`src/main/proto/wfa/measurement/internal/securecomputation/controlplane/work_item_attempts_service.proto` line 69 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2091#-OKCjcBm65qRdB6CJU-o:-OKCjcBn-f31IItOjAS0:b-zfpsf9) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/70d8697b01222b52a20a13d175af4f976b661924/src/main/proto/wfa/measurement/internal/securecomputation/controlplane/work_item_attempts_service.proto#L69)):*
> ```proto
> message ListWorkItemAttemptsPageToken {
>   message After {
>     fixed64 work_item_resource_id = 1;
> ```

We probably want to order by create time (effectively item number) first. Generally the IDs are just the tie-breakers to give a total ordering.

Document the ordering on the method.

Marco-Premier (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2091)* status: 8 of 11 files reviewed, 1 unresolved discussion (waiting on @SanjayVas and @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/internal/securecomputation/controlplane/work_item_attempts_service.proto` line 69 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2091#-OKCjcBm65qRdB6CJU-o:-OKCtGBn1nGWVj_Lq5Sv:b9mbwrp) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/70d8697b01222b52a20a13d175af4f976b661924/src/main/proto/wfa/measurement/internal/securecomputation/controlplane/work_item_attempts_service.proto#L69)):*

We probably want to order by create time (effectively item number) first. Generally the IDs are just the tie-breakers to give a total ordering.

Document the ordering on the method.
</blockquote></details>

Done. documented at the rpc service declaration

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2091)* status: all files reviewed, 1 unresolved discussion (waiting on @Marco-Premier and @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/internal/securecomputation/controlplane/work_item_attempts_service.proto` line 69 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2091#-OKCjcBm65qRdB6CJU-o:-OKCzWKJ0V_TLSF4f4pc:bu2d5vb) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/70d8697b01222b52a20a13d175af4f976b661924/src/main/proto/wfa/measurement/internal/securecomputation/controlplane/work_item_attempts_service.proto#L69)):*

Done. documented at the rpc service declaration
</blockquote></details>

You still need the tie-breaker fields in the page token.

SanjayVas (changes requested):
> Reviewed 5 of 5 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2091)* status: all files reviewed, 3 unresolved discussions (waiting on @Marco-Premier and @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/internal/securecomputation/controlplane/work_item_attempts_service.proto` line 69 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2091#-OKCjcBm65qRdB6CJU-o:-OKSdm454ascwoNjLepm:b1v5v9n) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/70d8697b01222b52a20a13d175af4f976b661924/src/main/proto/wfa/measurement/internal/securecomputation/controlplane/work_item_attempts_service.proto#L69)):*

You still need the tie-breaker fields in the page token.
</blockquote></details>

This also needs to include the work item resource ID in the tie-breaker, as any resource ID is only guaranteed to be unique within the context of its parent. To make things clearer, you probably want to order these fields in the order they are used.
___
*[`src/main/proto/wfa/measurement/internal/securecomputation/controlplane/work_item_attempts_service.proto` line 52 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2091#-OKSdcSF-vfQfmvZKcYE:-OKSdcSF-vfQfmvZKcYF:bvcyqwb) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7087e5ad2363712e3eb5896bbdb5f73a1dbcba00/src/main/proto/wfa/measurement/internal/securecomputation/controlplane/work_item_attempts_service.proto#L52)):*
> ```proto
> 
> message ListWorkItemAttemptsRequest {
>   fixed64 work_item_resource_id = 1;
> ```

The resource ID will be a string. Internal IDs are ints.



_Code quote:_
```proto
fixed64
```
___
*[`src/main/proto/wfa/measurement/internal/securecomputation/controlplane/work_item_attempts_service.proto` line 74 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2091#-OKSe6yhAPCO3x6RNDrb:-OKSe6yi5X48gYfan25g:b-nzgn8d) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7087e5ad2363712e3eb5896bbdb5f73a1dbcba00/src/main/proto/wfa/measurement/internal/securecomputation/controlplane/work_item_attempts_service.proto#L74)):*
> ```proto
>   message After {
>     fixed64 work_item_attempt_resource_id = 1;
>     google.protobuf.Timestamp create_after = 2;
> ```

nit: `created_after` for natural English or `create_time_after` if you'd rather reference the field name.



_Suggestion:_
```proto
created_after
```

Marco-Premier (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2091)* status: 9 of 11 files reviewed, 2 unresolved discussions (waiting on @SanjayVas and @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/internal/securecomputation/controlplane/work_item_attempts_service.proto` line 69 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2091#-OKCjcBm65qRdB6CJU-o:-OKT-MSs6wax6xXmtOKh:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/70d8697b01222b52a20a13d175af4f976b661924/src/main/proto/wfa/measurement/internal/securecomputation/controlplane/work_item_attempts_service.proto#L69)):*

This also needs to include the work item resource ID in the tie-breaker, as any resource ID is only guaranteed to be unique within the context of its parent. To make things clearer, you probably want to order these fields in the order they are used.
</blockquote></details>

Done.
___
*[`src/main/proto/wfa/measurement/internal/securecomputation/controlplane/work_item_attempts_service.proto` line 52 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2091#-OKSdcSF-vfQfmvZKcYE:-OKSzDaH-R97G5oXsENV:b-mrksgs) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7087e5ad2363712e3eb5896bbdb5f73a1dbcba00/src/main/proto/wfa/measurement/internal/securecomputation/controlplane/work_item_attempts_service.proto#L52)):*

The resource ID will be a string. Internal IDs are ints.
</blockquote></details>

Cool, now all resource ids are string also for `WorkItem`.
Help me clarify something here, `external ids` used to be ints and generated by the system.
`resource ids` are now string, I see we want the client to provide them when creating resources (eg. `CreateUserPrincipalRequest`). Does the same apply for `WorkItem` and `WorkItemAttempt`? Is the client that provides the resources ids for those?
How do we prevent collision?

Keboletsetse (approved):

SanjayVas (approved):
> Reviewed 5 of 5 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2091)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/internal/securecomputation/controlplane/work_item_attempts_service.proto` line 52 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2091#-OKSdcSF-vfQfmvZKcYE:-OKY5aTf5tkRMjRe6QzW:b-2v03hq) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7087e5ad2363712e3eb5896bbdb5f73a1dbcba00/src/main/proto/wfa/measurement/internal/securecomputation/controlplane/work_item_attempts_service.proto#L52)):*

Cool, now all resource ids are string also for `WorkItem`.
Help me clarify something here, `external ids` used to be ints and generated by the system.
`resource ids` are now string, I see we want the client to provide them when creating resources (eg. `CreateUserPrincipalRequest`). Does the same apply for `WorkItem` and `WorkItemAttempt`? Is the client that provides the resources ids for those?
How do we prevent collision?
</blockquote></details>

Yes, the client provides the resource IDs now. We just check whether the ID is already in use and throw an ALREADY_EXISTS error. See the usage of https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/main/src/main/kotlin/org/wfanet/measurement/common/IdGenerator.kt

Marco-Premier (commented):
> Reviewed 8 of 8 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2091)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @stevenwarejones)

PR #2102, opened 2025-02-28
build: add code coverage report for kotlin packages

laureanobrs:
This adds the generation of the code coverage report for Kotlin packages in the form of a .dat file.


laureanobrs (commented):
This PR adds Kotlin code coverage. The next steps are:

-   Review coverage generation across different PRs to check for indirect changes (see [https://docs.codecov.com/docs/unexpected-coverage-changes#reasons-for-indirect-changes](https://docs.codecov.com/docs/unexpected-coverage-changes#reasons-for-indirect-changes)) and determine if any action is needed.
    
-   Create a PR to add documentation for Kotlin coverage, including how to run it locally, the ongoing issue with inline functions, and how to configure exclusions.
    
-   Evaluate re-enabling the Codecov summary comment on PRs.

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2102#-:-O_wit0fDYTI805yMHLg:bt753u7)* status: 0 of 5 files reviewed, all discussions resolved

___
*[`.github/workflows/build-test.yml` line 116 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2102#-O_wkObP0aKY80kYBd_x:-O_wkObP0aKY80kYBd_y:bj3iu5p) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5387150da1e4435a2209f2ccda7b78694fd9e438/.github/workflows/build-test.yml#L116)):*
> ```YAML
>           set -euo pipefail
> 
>           readarray -t kt_files_with_inline_fun < <(grep -rl --include '*.kt' "inline fun" src/main/kotlin/*)
> ```

This isn't a robust method for detecting which files use inline functions, but it serves as an aid so we don't have to manually configure the exclusion filter for every single case. If any aren't detected, the intention is to find and add them manually (or improve this code). We hope to remove this as soon as the problem is fixed in rules\_kotlin.
___
*[`src/main/python/wfa/measurement/reporting/postprocessing/report/BUILD.bazel` line 13 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2102#-O_wiQSN15OSxwYfkxMM:-O_wiQSN15OSxwYfkxMN:b-9jr3w7) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5387150da1e4435a2209f2ccda7b78694fd9e438/src/main/python/wfa/measurement/reporting/postprocessing/report/BUILD.bazel#L13)):*
> ```Starlark
>         "//src/main/python/wfa/measurement/reporting/postprocessing/noiseninja",
>         requirement("absl-py"),
>         requirement("protobuf"),
> ```

To fix the exception that happens only when running the test with bazel coverage:



_Code snippet:_
```Starlark
        from google.protobuf import descriptor as _descriptor
    ModuleNotFoundError: No module named 'google'
```

SanjayVas (approved):
> @SanjayVas reviewed 5 of 5 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2102)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @laureanobrs)

___
*[`.github/workflows/build-test.yml` line 116 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2102#-O_wkObP0aKY80kYBd_x:-O_xtSrR1Vd88wNGuy5v:b-k6bpu9) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5387150da1e4435a2209f2ccda7b78694fd9e438/.github/workflows/build-test.yml#L116)):*

This isn't a robust method for detecting which files use inline functions, but it serves as an aid so we don't have to manually configure the exclusion filter for every single case. If any aren't detected, the intention is to find and add them manually (or improve this code). We hope to remove this as soon as the problem is fixed in rules\_kotlin.
</blockquote></details>

Nice work

PR #2115, opened 2025-03-18
build: Update to non-yanked version of rules_cc

SanjayVas:
rules_cc@0.1.0 has been yanked


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2115)

PR #2104, opened 2025-03-03
feat: add internal services for controlplane API persistence

Marco-Premier:
Depends on #2091

PR #2126, opened 2025-03-25
feat: add control plane public api

Marco-Premier:
Depends on #2104


SanjayVas (changes requested):
> Reviewed 13 of 17 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2126)* status: 13 of 17 files reviewed, 9 unresolved discussions (waiting on @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/WorkItemsService.kt` line 173 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2126#-OMDIGTl440r4i4v1RSz:-OMDIGTl440r4i4v1RT-:b-xbxqfs) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/25f2ace8ad91210cf5bbeae0245d187db1522e84/src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/WorkItemsService.kt#L173)):*
> ```Kotlin
> 
>     val internalResponse: InternalListWorkItemsResponse =
>       internalWorkItemsStub.listWorkItems(
> ```

You may want to use our listResources utility function to handle pagination since you have a higher max page size documented for the public method vs. the internal one. Even if you fix them to be the same now, using it would still give you the benefit of not needing to change the impl to handle independent changes in sizes.
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/WorkItemsService.kt` line 175 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2126#-OMDHhra4ubIT25L6ZjS:-OMDHhra4ubIT25L6ZjT:b-9lee3g) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/25f2ace8ad91210cf5bbeae0245d187db1522e84/src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/WorkItemsService.kt#L175)):*
> ```Kotlin
>       internalWorkItemsStub.listWorkItems(
>         internalListWorkItemsRequest {
>           pageSize = request.pageSize
> ```

If the default/max page sizes differ from the internal method, you need to account for that here.
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/GooglePubSubWorkItemsService.kt` line 24 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2126#-OMDU97K7cSGx-1FI0bl:-OMDU97K7cSGx-1FI0bm:b-oqd51s) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/25f2ace8ad91210cf5bbeae0245d187db1522e84/src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/GooglePubSubWorkItemsService.kt#L24)):*
> ```Kotlin
> import org.wfanet.measurement.internal.securecomputation.controlplane.WorkItemsGrpcKt.WorkItemsCoroutineStub as InternalWorkItemsCoroutineStub
> 
> class GooglePubSubWorkItemsService(
> ```

Normally the public service is not the one that potentially has multiple implementations, but rather the internal service. The public service is instead fully cloud-agnostic. You can see that in our conventions for internal services:

1. Cloud-specific bits go in a subpackage under `deploy` for the relevant cloud.
2. There's an abstract test class for each service which the implementation-specific test inherits from

The tradeoff for having the message delivery be handled in the public service impl is that it's more clearly decoupled from the choice of DB impl. Given that both choices are cloud-specific[^1] I think it's best to handle this in the internal service. This way only the internal service impl needs to have dependencies on all the gcloud client libraries.

[^1]: Even if we were using Postgres for the DB, there's some specifics for connecting to a Postgres instance hosted in Google Cloud SQL vs. AWS's hosted option
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/testing/FakeWorkItemsService.kt` line 23 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2126#-OMDMbtM0OIFU1rkk196:-OMDMbtM0OIFU1rkk197:b-i975m) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/25f2ace8ad91210cf5bbeae0245d187db1522e84/src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/testing/FakeWorkItemsService.kt#L23)):*
> ```Kotlin
> import org.wfanet.measurement.securecomputation.controlplane.v1alpha.WorkItemsService
> 
> class TestWorkItemsService(
> ```

Match the class name and file name.



_Code quote:_
```Kotlin
TestWorkItemsService
```
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/service/Errors.kt` line 28 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2126#-OMDNm5Y9KLFTxIW83K0:-OMDNm5Y9KLFTxIW83K1:b4dvqeu) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/25f2ace8ad91210cf5bbeae0245d187db1522e84/src/main/kotlin/org/wfanet/measurement/securecomputation/service/Errors.kt#L28)):*
> ```Kotlin
> 
> object Errors {
>   const val DOMAIN = "securecomputation.halo-cmm.org"
> ```

Match the domain used in the resource type.



_Suggestion:_
```Kotlin
control-plane.secure-computation.halo-cmm.org
```
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/service/Errors.kt` line 113 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2126#-OMDSdZp438vd0X6OoIx:-OMDSdZqE-MkJBx_81TS:byy1xig) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/25f2ace8ad91210cf5bbeae0245d187db1522e84/src/main/kotlin/org/wfanet/measurement/securecomputation/service/Errors.kt#L113)):*
> ```Kotlin
>     cause,
>   ) {
>     companion object : Factory<WorkItemNotFoundException>() {
> ```

nit: you only need to implement the factory interface for exceptions where you need to parse information from the internal method's error info. You can drop all of the unused factories.
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/WorkItemAttemptsService.kt` line 63 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2126#-OMDKntB8oevUhnFF8M-:-OMDKntB8oevUhnFF8M0:b-tnb74u) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/25f2ace8ad91210cf5bbeae0245d187db1522e84/src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/WorkItemAttemptsService.kt#L63)):*
> ```Kotlin
>     }
> 
>     val parentKey = WorkItemKey.fromName(request.parent)
> ```

Throw an InvalidFieldValueException if this returns null so that parentKey can be non-nullable. This will also avoid the need for the non-null assertion later.



_Code quote:_
```Kotlin
val parentKey = WorkItemKey.fromName(request.parent)
```
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/WorkItemAttemptsService.kt` line 216 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2126#-OMDJkJXEDUgFFBHHFKg:-OMDJkJXEDUgFFBHHFKh:bpppkfo) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/25f2ace8ad91210cf5bbeae0245d187db1522e84/src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/WorkItemAttemptsService.kt#L216)):*
> ```Kotlin
>   }
> 
>   override suspend fun listWorkItemAttempts(request: ListWorkItemAttemptsRequest): org.wfanet.measurement.securecomputation.controlplane.v1alpha.ListWorkItemAttemptsResponse {
> ```

You should be able to import this symbol



_Code quote:_
```Kotlin
org.wfanet.measurement.securecomputation.controlplane.v1alpha.ListWorkItemAttemptsResponse
```
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/service/WorkItemAttemptKey.kt` line 22 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2126#-OMDT4yo1hLyL__ks8mA:-OMDT4yo1hLyL__ks8mB:undefined) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/25f2ace8ad91210cf5bbeae0245d187db1522e84/src/main/kotlin/org/wfanet/measurement/securecomputation/service/WorkItemAttemptKey.kt#L22)):*
> ```Kotlin
> import org.wfanet.measurement.common.api.ResourceKey
> 
> class WorkItemAttemptKey(val workItemId: String, val workItemAttemptId: String) : ResourceKey {
> ```





_Suggestion:_
```Kotlin
ChildResourceKey
```

Marco-Premier (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2126)* status: 4 of 21 files reviewed, 8 unresolved discussions (waiting on @SanjayVas and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/WorkItemAttemptsService.kt` line 63 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2126#-OMDKntB8oevUhnFF8M-:-OMI6Ih55hJ5rnugzn3Z:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/25f2ace8ad91210cf5bbeae0245d187db1522e84/src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/WorkItemAttemptsService.kt#L63)):*

Throw an InvalidFieldValueException if this returns null so that parentKey can be non-nullable. This will also avoid the need for the non-null assertion later.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/WorkItemAttemptsService.kt` line 216 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2126#-OMDJkJXEDUgFFBHHFKg:-OMHe5E56XjD1D7WaP0r:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/25f2ace8ad91210cf5bbeae0245d187db1522e84/src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/WorkItemAttemptsService.kt#L216)):*

You should be able to import this symbol
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/WorkItemsService.kt` line 173 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2126#-OMDIGTl440r4i4v1RSz:-OMIRJeH-Ihnrrqf7zOP:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/25f2ace8ad91210cf5bbeae0245d187db1522e84/src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/WorkItemsService.kt#L173)):*

You may want to use our listResources utility function to handle pagination since you have a higher max page size documented for the public method vs. the internal one. Even if you fix them to be the same now, using it would still give you the benefit of not needing to change the impl to handle independent changes in sizes.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/WorkItemsService.kt` line 175 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2126#-OMDHhra4ubIT25L6ZjS:-OMIRKDL3MWFqPl6Bazo:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/25f2ace8ad91210cf5bbeae0245d187db1522e84/src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/WorkItemsService.kt#L175)):*

If the default/max page sizes differ from the internal method, you need to account for that here.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/service/Errors.kt` line 28 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2126#-OMDNm5Y9KLFTxIW83K0:-OMHfFxr8uFIJaj064lA:b-ds45sd) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/25f2ace8ad91210cf5bbeae0245d187db1522e84/src/main/kotlin/org/wfanet/measurement/securecomputation/service/Errors.kt#L28)):*

Match the domain used in the resource type.
</blockquote></details>

Done. So should the domain for internal errors become:

`"internal.control-plane.secure-computation.halo-cmm.org"` then?
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/GooglePubSubWorkItemsService.kt` line 24 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2126#-OMDU97K7cSGx-1FI0bl:-OMHnnOcELVDptCECKfo:bqbnbmc) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/25f2ace8ad91210cf5bbeae0245d187db1522e84/src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/GooglePubSubWorkItemsService.kt#L24)):*

Normally the public service is not the one that potentially has multiple implementations, but rather the internal service. The public service is instead fully cloud-agnostic. You can see that in our conventions for internal services:

1. Cloud-specific bits go in a subpackage under `deploy` for the relevant cloud.
2. There's an abstract test class for each service which the implementation-specific test inherits from

The tradeoff for having the message delivery be handled in the public service impl is that it's more clearly decoupled from the choice of DB impl. Given that both choices are cloud-specific[^1] I think it's best to handle this in the internal service. This way only the internal service impl needs to have dependencies on all the gcloud client libraries.

[^1]: Even if we were using Postgres for the DB, there's some specifics for connecting to a Postgres instance hosted in Google Cloud SQL vs. AWS's hosted option
</blockquote></details>

I don't like having this tightly integrated with the Spanner. What I did was to remove the `GooglePubSubWorkItemsService` in favour of a `WorkItemsPublisher` which is responsible for publishing the message and the `publishMessage` method is now triggered directly within the `createWorkItem` method.
Now public api are cloud agnostic.
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/testing/FakeWorkItemsService.kt` line 23 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2126#-OMDMbtM0OIFU1rkk196:-OMHeitNBlCFwQ3-2qHn:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/25f2ace8ad91210cf5bbeae0245d187db1522e84/src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/testing/FakeWorkItemsService.kt#L23)):*

Match the class name and file name.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/service/WorkItemAttemptKey.kt` line 22 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2126#-OMDT4yo1hLyL__ks8mA:-OMI7eVACp45pzkpfhMi:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/25f2ace8ad91210cf5bbeae0245d187db1522e84/src/main/kotlin/org/wfanet/measurement/securecomputation/service/WorkItemAttemptKey.kt#L22)):*
> ```Kotlin
> import org.wfanet.measurement.common.api.ResourceKey
> 
> class WorkItemAttemptKey(val workItemId: String, val workItemAttemptId: String) : ResourceKey {
> ```

Done.

stevenwarejones (approved):
> Reviewed 8 of 17 files at r1, 14 of 14 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2126)* status: all files reviewed, 8 unresolved discussions (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/GooglePubSubWorkItemsService.kt` line 24 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2126#-OMDU97K7cSGx-1FI0bl:-OMICGD30nhuvd1019Fw:b-8n6f1y) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/25f2ace8ad91210cf5bbeae0245d187db1522e84/src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/GooglePubSubWorkItemsService.kt#L24)):*

I don't like having this tightly integrated with the Spanner. What I did was to remove the `GooglePubSubWorkItemsService` in favour of a `WorkItemsPublisher` which is responsible for publishing the message and the `publishMessage` method is now triggered directly within the `createWorkItem` method.
Now public api are cloud agnostic.
</blockquote></details>

I think its fine to have a public service that can have multiple implementations. However, it looks like he refactored this so its that the rpc layer is cloud agnostic and accepts a WorkItemsPublisher in the constructor. that works for me for now and I"d prefer to adjust any changes in the future.
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/deploy/WorkItemsPublisher.kt` line 17 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2126#-OMIs4BU6it8k3m3tYFB:-OMIs4BV473fl8_enYqR:b-7amhfx) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a059f423b0d91023c1333a4fd9bc20a773661683/src/main/kotlin/org/wfanet/measurement/securecomputation/deploy/WorkItemsPublisher.kt#L17)):*
> ```Kotlin
>  */
> 
> package org.wfanet.measurement.securecomputation.deploy
> ```

nit: Since this is cloud agnostic, it doesn't need to be under deploy

SanjayVas (changes requested):
> Reviewed 1 of 14 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2126)* status: all files reviewed, 8 unresolved discussions (waiting on @Marco-Premier and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/GooglePubSubWorkItemsService.kt` line 24 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2126#-OMDU97K7cSGx-1FI0bl:-OMJqYnMEWFW6OgJtf6s:b-qvv1do) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/25f2ace8ad91210cf5bbeae0245d187db1522e84/src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/GooglePubSubWorkItemsService.kt#L24)):*
The issue here is that you end up with gcloud client dependencies in the public server, which we've avoided thus far.

> I don't like having this tightly integrated with the Spanner.

You could do the same pattern with WorkItemPublisher, just in the internal server. The gcloud version of the internal server would initialize `GoogleWorkItemsPublisher`.

Marco-Premier (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2126)* status: 9 of 46 files reviewed, 8 unresolved discussions (waiting on @SanjayVas and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/GooglePubSubWorkItemsService.kt` line 24 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2126#-OMDU97K7cSGx-1FI0bl:-OMM7yNMAr0GVKNW7qp6:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/25f2ace8ad91210cf5bbeae0245d187db1522e84/src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/GooglePubSubWorkItemsService.kt#L24)):*

The issue here is that you end up with gcloud client dependencies in the public server, which we've avoided thus far.

> I don't like having this tightly integrated with the Spanner.

You could do the same pattern with WorkItemPublisher, just in the internal server. The gcloud version of the internal server would initialize `GoogleWorkItemsPublisher`.
</blockquote></details>

Done.

SanjayVas (changes requested):
> Reviewed 1 of 14 files at r2, 6 of 37 files at r3.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2126)* status: 15 of 46 files reviewed, 6 unresolved discussions (waiting on @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/WorkItemsService.kt` line 173 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2126#-OMDIGTl440r4i4v1RSz:-OMOhgLH9BzNzwutmSsl:b-7ygy37) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/25f2ace8ad91210cf5bbeae0245d187db1522e84/src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/WorkItemsService.kt#L173)):*

Done.
</blockquote></details>

nit: Oof, this is my bad. I didn't realize listResources only handled string page tokens for public APIs. If you keep the default/max page sizes the same between the public/internal services, you can undo this and just call the internal service with the request page size. I can fix listResources later to make this cleaner.

You're also welcome to leave it as-is.
___
*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2126#-OMOlDGi4U_UV6vMjNTF:-OMOlDGi4U_UV6vMjNTG:bn9zfj4) (no related file):*
Same note as the other PR: If one PR depends on another in the same repo, you need to actually have one PR branch based on the other and then update the base branch in the GitHub UI.
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/WorkItemAttemptsService.kt` line 50 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2126#-OMNyIQj7Gn_zXRlxtcY:-OMNyIQj7Gn_zXRlxtcZ:bu0hwrg) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/9491fdfbef863caebc03700d90cece04fc0115b0/src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/WorkItemAttemptsService.kt#L50)):*
> ```Kotlin
> import org.wfanet.measurement.securecomputation.service.WorkItemInvalidStateException
> 
> private const val DEFAULT_PAGE_SIZE = 50
> ```

nit: new constants should go inside the companion object

___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/WorkItemAttemptsService.kt` line 86 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2126#-OMNyf6P0AM582bYzkWF:-OMNyf6P0AM582bYzkWG:bhify1) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/9491fdfbef863caebc03700d90cece04fc0115b0/src/main/kotlin/org/wfanet/measurement/securecomputation/controlplane/v1alpha/WorkItemAttemptsService.kt#L86)):*
> ```Kotlin
>         throw when (InternalErrors.getReason(e)) {
>           InternalErrors.Reason.WORK_ITEM_NOT_FOUND ->
>             WorkItemNotFoundException(parentKey.toName(), e).asStatusRuntimeException(e.status.code)
> ```

nit: you technically already have the string name in the request



_Suggestion:_
```Kotlin
request.parent
```

Marco-Premier (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2126)* status: 15 of 46 files reviewed, 4 unresolved discussions (waiting on @SanjayVas and @stevenwarejones)

___
*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2126#-OMOlDGi4U_UV6vMjNTF:-OMRrhON4kz7sSTw_HP8:b-5m2x0) (no related file):*

Same note as the other PR: If one PR depends on another in the same repo, you need to actually have one PR branch based on the other and then update the base branch in the GitHub UI.
</blockquote></details>

I'm pretty much sure I did that, definitely I checkout this branch on top of the internal services implementation otherwise they wouldn't be here. Can this happen after merging the branch this branch depends on?
If not, I missed it then.

Marco-Premier (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2126)* status: 13 of 46 files reviewed, 4 unresolved discussions (waiting on @SanjayVas and @stevenwarejones)

___
*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2126#-OMOlDGi4U_UV6vMjNTF:-OMbROmz2PW7Djf6A5iI:b-896fix) (no related file):*

After you merge the branch it depends on, GitHub will update the base of the PR. You still need to update the PR *branch* when this happens, e.g. rebase your PR branch on the default branch and force push. You can tell if the state is correct by whether the "commits" shows **only** the commits for this PR. Until then, the PR is unreviewable as it's showing the incorrect diffs.
</blockquote></details>

Done.

SanjayVas (changes requested):
> Reviewed 1 of 17 files at r1, 3 of 14 files at r2, 10 of 37 files at r3, 3 of 3 files at r4, 20 of 20 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2126)* status: all files reviewed, 11 unresolved discussions (waiting on @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/service/Errors.kt` line 28 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2126#-OMDNm5Y9KLFTxIW83K0:-OMh44WU-v82_AGDeQuI:b2nhswy) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/25f2ace8ad91210cf5bbeae0245d187db1522e84/src/main/kotlin/org/wfanet/measurement/securecomputation/service/Errors.kt#L28)):*

Done. So should the domain for internal errors become:

`"internal.control-plane.secure-computation.halo-cmm.org"` then?
</blockquote></details>

Yes
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/service/internal/testing/WorkItemsServiceTest.kt` line 92 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2126#-OMh9QHU0tRE_zwaFHX4:-OMh9QHU0tRE_zwaFHX5:b2ni8l0) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/07df30a75d92d1e68b40393b973227026cc2ffba/src/main/kotlin/org/wfanet/measurement/securecomputation/service/internal/testing/WorkItemsServiceTest.kt#L92)):*
> ```Kotlin
>   }
> 
>   private suspend fun deleteSubscriptoinAndTopic() {
> ```

nit



_Suggestion:_
```Kotlin
Subscription
```
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/service/internal/testing/WorkItemsServiceTest.kt` line 110 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2126#-OMhKrfE5fWdYulRmj4e:-OMhKrfFAcwjXAzVa7R9:b-6aueab) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/07df30a75d92d1e68b40393b973227026cc2ffba/src/main/kotlin/org/wfanet/measurement/securecomputation/service/internal/testing/WorkItemsServiceTest.kt#L110)):*
> ```Kotlin
> 
>   @Test
>   fun `createWorkItem returns created WorkItem`() = runBlocking {
> ```

nit: technically you're not testing this case unless the test also covers that it matches what's returned by getWorkItem.

Typically for these internal service tests we don't have a separate test for `get` since it inherently depends on `create`. We instead just include the fact that the `create` and `get` responses are the same in the `create` test and just verify the error cases for the `get` tests.



_Code quote:_
```Kotlin
returns created WorkItem
```
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/service/internal/testing/WorkItemsServiceTest.kt` line 138 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2126#-OMhLBgz6YWH4SltcTcC:-OMhLBgz6YWH4SltcTcD:b-71kiuy) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/07df30a75d92d1e68b40393b973227026cc2ffba/src/main/kotlin/org/wfanet/measurement/securecomputation/service/internal/testing/WorkItemsServiceTest.kt#L138)):*
> ```Kotlin
>         WorkItem.UPDATE_TIME_FIELD_NUMBER,
>         WorkItem.WORK_ITEM_RESOURCE_ID_FIELD_NUMBER,
>         WorkItem.WORK_ITEM_PARAMS_FIELD_NUMBER
> ```

Don't ignore this. It should be verified. You may need to create a TypeRegistry and pass that to [`unpackAnyUsing`](https://truth.dev/api/latest/com/google/common/truth/extensions/proto/ProtoFluentAssertion.html#unpackingAnyUsing(com.google.protobuf.TypeRegistry,com.google.protobuf.ExtensionRegistry)) so that Truth knows how to compare the Any.
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/service/internal/WorkItemsPublisher.kt` line 5 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2126#-OMh4DzIA8ZgqmeOiafs:-OMh4DzIA8ZgqmeOiaft:b-ij9v04) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/07df30a75d92d1e68b40393b973227026cc2ffba/src/main/kotlin/org/wfanet/measurement/securecomputation/service/internal/WorkItemsPublisher.kt#L5)):*
> ```Kotlin
> import com.google.protobuf.Message
> 
> interface WorkItemsPublisher {
> ```

nit: this only publishes one item at a time. Even if it allowed multiple, the singular form is more common for cases like this.



_Suggestion:_
```Kotlin
WorkItemPublisher
```
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/deploy/testing/FakeWorkItemsPublisher.kt` line 22 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2126#-OMh8-UeChvGTjagaZnm:-OMh8-UeChvGTjagaZnn:bprpkz4) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/07df30a75d92d1e68b40393b973227026cc2ffba/src/main/kotlin/org/wfanet/measurement/securecomputation/deploy/testing/FakeWorkItemsPublisher.kt#L22)):*
> ```Kotlin
> import org.wfanet.measurement.securecomputation.service.internal.WorkItemsPublisher
> 
> class FakeWorkItemsPublisher(): WorkItemsPublisher {
> ```

I'm still not sure I see a benefit of having a no-op impl that's only used by one test here. This is trivial to define in the test that uses it.
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/deploy/testing/FakeWorkItemsPublisher.kt` line 24 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2126#-OMh834HCGKAIu6j0oGh:-OMh834HCGKAIu6j0oGi:bby9e8n) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/07df30a75d92d1e68b40393b973227026cc2ffba/src/main/kotlin/org/wfanet/measurement/securecomputation/deploy/testing/FakeWorkItemsPublisher.kt#L24)):*
> ```Kotlin
> class FakeWorkItemsPublisher(): WorkItemsPublisher {
>   override suspend fun publishMessage(queueName: String, message: Message) {
>     println("message published")
> ```

Use a logger rather than printing
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/deploy/gcloud/spanner/SpannerWorkItemService.kt` line 78 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2126#-OMh0-gP22TnAtWDHVx0:-OMh0-gP22TnAtWDHVx1:b6wodro) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/07df30a75d92d1e68b40393b973227026cc2ffba/src/main/kotlin/org/wfanet/measurement/securecomputation/deploy/gcloud/spanner/SpannerWorkItemService.kt#L78)):*
> ```Kotlin
>     }
> 
>     if (request.workItem.workItemParams == Any.getDefaultInstance()) {
> ```

All message fields have explicit presence[^1] and therefore have a "has" presence accessor[^2].

[^1]: https://protobuf.dev/programming-guides/field_presence/
[^2]: https://protobuf.dev/reference/java/java-generated/#singular-proto3



_Suggestion:_
```Kotlin
request.workItem.hasWorkItemParams()
```
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/deploy/gcloud/spanner/SpannerWorkItemService.kt` line 93 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2126#-OMh0zS95beMDCorUpZ9:-OMh0zS95beMDCorUpZA:b-lrep54) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/07df30a75d92d1e68b40393b973227026cc2ffba/src/main/kotlin/org/wfanet/measurement/securecomputation/deploy/gcloud/spanner/SpannerWorkItemService.kt#L93)):*
> ```Kotlin
>       throw when {
>         e.message?.contains("Topic id: ${request.workItem.queueResourceId} does not exist") == true -> {
> ```

Never inspect the message of an exception. Use more specific exception types if you need to carry more information. Note that the exception type effectively becomes part of the WorkItemPublisher API.

Note that this also shouldn't be a NOT_FOUND error from the API as this error doesn't indicate anything about the existence of the WorkItem resource. This is instead a FAILED_PRECONDITION that the PubSub topic doesn't exist.

Furthermore, this kills the point of abstracting out WorkItemPublisher since it's checking for something from a specific impl.
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/deploy/gcloud/spanner/SpannerWorkItemService.kt` line 94 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2126#-OMh2A9XEj0YU77H7Wcy:-OMh2A9XEj0YU77H7Wcz:bmdi0kd) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/07df30a75d92d1e68b40393b973227026cc2ffba/src/main/kotlin/org/wfanet/measurement/securecomputation/deploy/gcloud/spanner/SpannerWorkItemService.kt#L94)):*
> ```Kotlin
>       throw when {
>         e.message?.contains("Topic id: ${request.workItem.queueResourceId} does not exist") == true -> {
>           Status.NOT_FOUND.withDescription(e.message).asRuntimeException()
> ```

Include the cause for debugging. `withCause(e)`
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/deploy/gcloud/spanner/SpannerWorkItemService.kt` line 98 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2126#-OMh276d6DgV6pOvGySw:-OMh1tfr4bMipdXEtT-3:b-5r6n74) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/07df30a75d92d1e68b40393b973227026cc2ffba/src/main/kotlin/org/wfanet/measurement/securecomputation/deploy/gcloud/spanner/SpannerWorkItemService.kt#L98)):*
> ```Kotlin
> 
>         else -> {
>           Status.UNKNOWN.withDescription("An unknown error occurred: ${e.message}")
> ```

Use INTERNAL if you're catching an exception that there's no other info for.

We've been using UNKNOWN incorrectly and should be using INTERNAL for backend errors. UNKNOWN should just be thrown automatically by gRPC infra when an exception is uncaught.



_Code quote:_
```Kotlin
UNKNOWN
```
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/deploy/gcloud/spanner/SpannerWorkItemService.kt` line 104 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2126#-OMh2wBvD7L3yKT7Sw5H:-OMh2wBvD7L3yKT7Sw5I:b-rci092) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/07df30a75d92d1e68b40393b973227026cc2ffba/src/main/kotlin/org/wfanet/measurement/securecomputation/deploy/gcloud/spanner/SpannerWorkItemService.kt#L104)):*
> ```Kotlin
>     }
> 
>     val transactionRunner = databaseClient.readWriteTransaction(Options.tag("action=createWorkItem"))
> ```

Don't we need to insert in the DB first and *then* publish the message? Otherwise you could have a message received with no WorkItem to fail or create an attempt against.

That's what the [flow diagrams](https://docs.google.com/document/d/11Dzgt0Pqb-fcXdk0xJ0wrPgDR6Mgdo1SvcTgKUSXZ-s/edit?pli=1&tab=t.0#heading=h.sa83ulmrgw63) in the design doc show.

Marco-Premier (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2126)* status: all files reviewed, 8 unresolved discussions (waiting on @SanjayVas and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/deploy/gcloud/spanner/SpannerWorkItemService.kt` line 78 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2126#-OMh0-gP22TnAtWDHVx0:-OMhR05N0nafiG25NCUv:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/07df30a75d92d1e68b40393b973227026cc2ffba/src/main/kotlin/org/wfanet/measurement/securecomputation/deploy/gcloud/spanner/SpannerWorkItemService.kt#L78)):*

All message fields have explicit presence[^1] and therefore have a "has" presence accessor[^2].

[^1]: https://protobuf.dev/programming-guides/field_presence/
[^2]: https://protobuf.dev/reference/java/java-generated/#singular-proto3
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/deploy/gcloud/spanner/SpannerWorkItemService.kt` line 93 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2126#-OMh0zS95beMDCorUpZ9:-OMkG-ay6p5nEAakxgXq:b-53eh44) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/07df30a75d92d1e68b40393b973227026cc2ffba/src/main/kotlin/org/wfanet/measurement/securecomputation/deploy/gcloud/spanner/SpannerWorkItemService.kt#L93)):*

Never inspect the message of an exception. Use more specific exception types if you need to carry more information. Note that the exception type effectively becomes part of the WorkItemPublisher API.

Note that this also shouldn't be a NOT_FOUND error from the API as this error doesn't indicate anything about the existence of the WorkItem resource. This is instead a FAILED_PRECONDITION that the PubSub topic doesn't exist.

Furthermore, this kills the point of abstracting out WorkItemPublisher since it's checking for something from a specific impl.
</blockquote></details>

yeah that's true, don't like either coupling WorkItemPublisher here.
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/deploy/gcloud/spanner/SpannerWorkItemService.kt` line 94 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2126#-OMh2A9XEj0YU77H7Wcy:-OMhhn0-2PnL05_sc1eU:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/07df30a75d92d1e68b40393b973227026cc2ffba/src/main/kotlin/org/wfanet/measurement/securecomputation/deploy/gcloud/spanner/SpannerWorkItemService.kt#L94)):*

Include the cause for debugging. `withCause(e)`
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/deploy/gcloud/spanner/SpannerWorkItemService.kt` line 98 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2126#-OMh276d6DgV6pOvGySw:-OMhheVd3C_M04FqvZZL:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/07df30a75d92d1e68b40393b973227026cc2ffba/src/main/kotlin/org/wfanet/measurement/securecomputation/deploy/gcloud/spanner/SpannerWorkItemService.kt#L98)):*

Use INTERNAL if you're catching an exception that there's no other info for.

We've been using UNKNOWN incorrectly and should be using INTERNAL for backend errors. UNKNOWN should just be thrown automatically by gRPC infra when an exception is uncaught.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/deploy/gcloud/spanner/SpannerWorkItemService.kt` line 104 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2126#-OMh2wBvD7L3yKT7Sw5H:-OMhhTA-0Cd-7Xji7XOl:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/07df30a75d92d1e68b40393b973227026cc2ffba/src/main/kotlin/org/wfanet/measurement/securecomputation/deploy/gcloud/spanner/SpannerWorkItemService.kt#L104)):*

Don't we need to insert in the DB first and *then* publish the message? Otherwise you could have a message received with no WorkItem to fail or create an attempt against.

That's what the [flow diagrams](https://docs.google.com/document/d/11Dzgt0Pqb-fcXdk0xJ0wrPgDR6Mgdo1SvcTgKUSXZ-s/edit?pli=1&tab=t.0#heading=h.sa83ulmrgw63) in the design doc show.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/service/internal/testing/WorkItemsServiceTest.kt` line 138 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2126#-OMhLBgz6YWH4SltcTcC:-OMhdnlS5KY0MQoqpQ3A:bhjs5d1) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/07df30a75d92d1e68b40393b973227026cc2ffba/src/main/kotlin/org/wfanet/measurement/securecomputation/service/internal/testing/WorkItemsServiceTest.kt#L138)):*

Don't ignore this. It should be verified. You may need to create a TypeRegistry and pass that to [`unpackAnyUsing`](https://truth.dev/api/latest/com/google/common/truth/extensions/proto/ProtoFluentAssertion.html#unpackingAnyUsing(com.google.protobuf.TypeRegistry,com.google.protobuf.ExtensionRegistry)) so that Truth knows how to compare the Any.
</blockquote></details>

but that's actually not returned from `createWorkItem` api as it's not persisted and needed only to publish the message. If I include it in the response than I should remove it from all other APIs tests while comparing responses as it can't be returned from other API.
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/deploy/testing/FakeWorkItemsPublisher.kt` line 22 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2126#-OMh8-UeChvGTjagaZnm:-OMhhEqJ0iIVohx-GdCS:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/07df30a75d92d1e68b40393b973227026cc2ffba/src/main/kotlin/org/wfanet/measurement/securecomputation/deploy/testing/FakeWorkItemsPublisher.kt#L22)):*

I'm still not sure I see a benefit of having a no-op impl that's only used by one test here. This is trivial to define in the test that uses it.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/deploy/testing/FakeWorkItemsPublisher.kt` line 24 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2126#-OMh834HCGKAIu6j0oGh:-OMhhEbWDpwg31e6V7eB:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/07df30a75d92d1e68b40393b973227026cc2ffba/src/main/kotlin/org/wfanet/measurement/securecomputation/deploy/testing/FakeWorkItemsPublisher.kt#L24)):*

Use a logger rather than printing
</blockquote></details>

Done.

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2126)* status: 32 of 46 files reviewed, 8 unresolved discussions (waiting on @Marco-Premier and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/service/internal/testing/WorkItemsServiceTest.kt` line 138 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2126#-OMhLBgz6YWH4SltcTcC:-OMnhY738bcL97qJ8Olf:bjql64m) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/07df30a75d92d1e68b40393b973227026cc2ffba/src/main/kotlin/org/wfanet/measurement/securecomputation/service/internal/testing/WorkItemsServiceTest.kt#L138)):*

It should be persisted. Again, the message delivery is an implementation detail. The service should still work even if you were to swap that out and switch to polling.
</blockquote></details>

To put it another way: it should technically be sufficient to pass just the WorkItem resource name in the message and have the receiver call `GetWorkItem`. We pass the whole WorkItem resource for efficiency as it saves an RPC at the cost of a slightly bigger payload for the PubSub message.

Marco-Premier (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2126)* status: 30 of 46 files reviewed, 8 unresolved discussions (waiting on @SanjayVas and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/service/internal/testing/WorkItemsServiceTest.kt` line 138 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2126#-OMhLBgz6YWH4SltcTcC:-OMpIkbMAA1RSg-I8aMQ:byxteuf) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/07df30a75d92d1e68b40393b973227026cc2ffba/src/main/kotlin/org/wfanet/measurement/securecomputation/service/internal/testing/WorkItemsServiceTest.kt#L138)):*

To put it another way: it should technically be sufficient to pass just the WorkItem resource name in the message and have the receiver call `GetWorkItem`. We pass the whole WorkItem resource for efficiency as it saves an RPC at the cost of a slightly bigger payload for the PubSub message.
</blockquote></details>

I added persistence to workItemParams field which is now returns from all grpc api

SanjayVas (changes requested):
> Reviewed 1 of 17 files at r1, 12 of 14 files at r6, 4 of 4 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2126)* status: all files reviewed, 4 unresolved discussions (waiting on @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/deploy/gcloud/spanner/SpannerWorkItemService.kt` line 95 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2126#-OMsrfpV6FXZUMRmvG5o:-OMsrfpV6FXZUMRmvG5p:b941c13) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f37d9488dfcbe671eda6b8c31310f3c645ca123a/src/main/kotlin/org/wfanet/measurement/securecomputation/deploy/gcloud/spanner/SpannerWorkItemService.kt#L95)):*
> ```Kotlin
>           val workItemId: Long = idGenerator.generateNewId { id -> txn.workItemIdExists(id) }
> 
>           val state =
> ```

nit: explicit types
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/deploy/gcloud/spanner/SpannerWorkItemService.kt` line 119 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2126#-OMsro202FUZTb4W7t-3:-OMsro202FUZTb4W7t-4:b1okj20) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f37d9488dfcbe671eda6b8c31310f3c645ca123a/src/main/kotlin/org/wfanet/measurement/securecomputation/deploy/gcloud/spanner/SpannerWorkItemService.kt#L119)):*
> ```Kotlin
>       throw Status.INTERNAL
>         .withDescription(e.message)
> ```

nit: leave this out or specify your own description/message



_Code quote:_
```Kotlin
.withDescription(e.message)
```
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/deploy/gcloud/spanner/db/WorkItems.kt` line 81 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2126#-OMssrPu8iegMyVG6-PG:-OMssrPu8iegMyVG6-PH:bpogldx) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f37d9488dfcbe671eda6b8c31310f3c645ca123a/src/main/kotlin/org/wfanet/measurement/securecomputation/deploy/gcloud/spanner/db/WorkItems.kt#L81)):*
> ```Kotlin
>     set("QueueId").to(queueId)
>     set("State").to(state)
>     set("WorkItemParams").to(value)
> ```

The client knows how to serialize protobuf messages



_Suggestion:_
```Kotlin
.to(workItemParams)
```
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/deploy/gcloud/spanner/db/WorkItems.kt` line 182 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2126#-OMssxyU2DQpU959VkXP:-OMssxyU2DQpU959VkXQ:undefined) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f37d9488dfcbe671eda6b8c31310f3c645ca123a/src/main/kotlin/org/wfanet/measurement/securecomputation/deploy/gcloud/spanner/db/WorkItems.kt#L182)):*
> ```Kotlin
>         queueResourceId = queue.queueResourceId
>         state = row.getProtoEnum("State", WorkItem.State::forNumber)
>         this.workItemParams = workItemParams
> ```





_Suggestion:_
```Kotlin
row.getProtoMessage("WorkItemParams", Any.getDefaultInstance())
```

Marco-Premier (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2126)* status: 44 of 46 files reviewed, 2 unresolved discussions (waiting on @SanjayVas and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/deploy/gcloud/spanner/db/WorkItems.kt` line 81 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2126#-OMssrPu8iegMyVG6-PG:-OMv5AwfEqmm0ZJSRTGL:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f37d9488dfcbe671eda6b8c31310f3c645ca123a/src/main/kotlin/org/wfanet/measurement/securecomputation/deploy/gcloud/spanner/db/WorkItems.kt#L81)):*

The client knows how to serialize protobuf messages
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/deploy/gcloud/spanner/db/WorkItems.kt` line 182 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2126#-OMssxyU2DQpU959VkXP:-OMv5LJY8vkWYRNCE-XS:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f37d9488dfcbe671eda6b8c31310f3c645ca123a/src/main/kotlin/org/wfanet/measurement/securecomputation/deploy/gcloud/spanner/db/WorkItems.kt#L182)):*
> ```Kotlin
>         queueResourceId = queue.queueResourceId
>         state = row.getProtoEnum("State", WorkItem.State::forNumber)
>         this.workItemParams = workItemParams
> ```

Done.

SanjayVas (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2126)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @stevenwarejones)

Marco-Premier (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2126#-:-ON-FBtA4dGcGjx4pOiO:bnfp4nl)* status: 28 of 46 files reviewed, all discussions resolved (waiting on @SanjayVas and @stevenwarejones)

Marco-Premier (commented):
> Reviewed 18 of 18 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2126)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @stevenwarejones)

Marco-Premier (commented):
> Reviewed 1 of 17 files at r1, 1 of 14 files at r2, 5 of 37 files at r3, 1 of 3 files at r4, 11 of 20 files at r5, 7 of 14 files at r6, 1 of 4 files at r7, 2 of 2 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2126)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @stevenwarejones)

PR #2129, opened 2025-03-27
feat: update tee sdk to consume new ControlPlane API

Marco-Premier:
Depends on #2126 

BaseTeeApplication now only accepts messages of type WorkItem and forward to TEE Application workItemParams of type Any.


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2129)

SanjayVas (changes requested):
> Reviewed 1 of 5 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2129)* status: 1 of 5 files reviewed, 1 unresolved discussion (waiting on @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt` line 49 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2129#-OMOmCBA0RTAj_wHlyFB:-OMOmCBB4MDRGIC2jL2t:bgz1gzf) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/dfb0bd31d8cf8eb2d58b5828c80c014e43bc03b4/src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt#L49)):*
> ```Kotlin
>   private val queueSubscriber: QueueSubscriber,
>   private val parser: Parser<WorkItem>,
>   private val workItemsService: WorkItemsService,
> ```

This needs to be a stub pointing to the API server, not a service impl. The service impl should only be instantiated in the API server itself.



_Code quote:_
```Kotlin
WorkItemsService
```

Marco-Premier (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2129)* status: 1 of 5 files reviewed, 1 unresolved discussion (waiting on @SanjayVas and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt` line 49 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2129#-OMOmCBA0RTAj_wHlyFB:-OMRnBiH3vrJsqgwkd9B:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/dfb0bd31d8cf8eb2d58b5828c80c014e43bc03b4/src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt#L49)):*

This needs to be a stub pointing to the API server, not a service impl. The service impl should only be instantiated in the API server itself.
</blockquote></details>

Done.

SanjayVas (changes requested):
> Reviewed 1 of 5 files at r1, 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2129)* status: all files reviewed, 7 unresolved discussions (waiting on @stevenwarejones)

___
*[`src/test/kotlin/org/wfanet/measurement/securecomputation/teesdk/BUILD.bazel` line 12 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2129#-OMT3rdx0Q-_8ile0PBo:-OMT3rdx0Q-_8ile0PBp:bns6vwc) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/2f701bbeeb3f05ea4babd84ae127800911a959ae/src/test/kotlin/org/wfanet/measurement/securecomputation/teesdk/BUILD.bazel#L12)):*
> ```Starlark
>     deps = [
>         "//src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk:base_tee_application",
>         "//src/main/proto/wfa/measurement/internal/securecomputation/controlplane:work_item_attempts_service_kt_jvm_grpc_proto",
> ```

This should not depend on the internal API definition. In fact, the package on the internal API definition should have its visibility restricted such that this **cannot** depend on it.
___
*[`src/test/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplicationTest.kt` line 94 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2129#-OMT2Hyx61YIoLQ8ab7q:-OMT2Hyx61YIoLQ8ab7r:b-f33jko) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/2f701bbeeb3f05ea4babd84ae127800911a959ae/src/test/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplicationTest.kt#L94)):*
> ```Kotlin
>     val pubSubClient = Subscriber(projectId = PROJECT_ID, googlePubSubClient = emulatorClient)
>     val publisher = Publisher<WorkItem>(PROJECT_ID, emulatorClient)
>     val workItemsService: WorkItemsCoroutineStub = mock {}
> ```

Never mock gRPC client stubs. You instead mock the service impl and use a test gRPC server, such as GrpcTestServerRule.

Until mockito-inline started shipping by default with mockito-kotlin, gRPC stubs couldn't even be mocked due to being final. This is an intentional restriction.
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt` line 49 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2129#-OMT230VEEdHcXwunVif:-OMT230W9_RV6oDIAlLb:b-gwxau5) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/2f701bbeeb3f05ea4babd84ae127800911a959ae/src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt#L49)):*
> ```Kotlin
>   private val queueSubscriber: QueueSubscriber,
>   private val parser: Parser<WorkItem>,
>   private val workItemsService: WorkItemsCoroutineStub,
> ```

nit: name this "stub" or "client" since it's a client stub rather than a service impl



_Suggestion:_
```Kotlin
workItemsStub
```
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt` line 78 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2129#-OMT0615-GXgq8pKZZwZ:-OMT0615-GXgq8pKZZw_:bbxudbb) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/2f701bbeeb3f05ea4babd84ae127800911a959ae/src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt#L78)):*
> ```Kotlin
>    */
>   private suspend fun processMessage(queueMessage: QueueSubscriber.QueueMessage<WorkItem>) {
>     val body = queueMessage.body
> ```

Explicit types where not obvious. I had to look up QueueSubcriber.QueueMessage to determine this is of type `WorkItem`.
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt` line 80 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2129#-OMT0IYo8SEVKKpVqm62:-OMT0IYo8SEVKKpVqm63:bv00wsc) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/2f701bbeeb3f05ea4babd84ae127800911a959ae/src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt#L80)):*
> ```Kotlin
>     val body = queueMessage.body
> 
>     if (body.name.isNullOrEmpty()) {
> ```

nit: WorkItem is a protobuf Message, so its accessors can never return null



_Suggestion:_
```Kotlin
body.name.isEmpty()
```
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt` line 90 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2129#-OMT0rFk8Ta3tr8Twq3c:-OMT0rFk8Ta3tr8Twq3d:b-2ba79u) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/2f701bbeeb3f05ea4babd84ae127800911a959ae/src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt#L90)):*
> ```Kotlin
> 
>       val workItemAttemptId = UUID.randomUUID().toString()
>       workItemAttempt = createWorkItemAttempt(workItemName, workItemAttemptId)
> ```

Avoid unnecessary mutability. Separate this out into another try block so that `workItemAttempt` can be a non-null val.

```kotlin
val workItemAttempt: WorkItemAttempt = try {
  val workItemAttemptId = UUID.randomUUID().toString()
  createWorkItemAttempt(workItemName, workItemAttemptId)
} catch (e: WhateverCreateThrows) {
  // Handle the exception
}

```
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt` line 107 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2129#-OMT1bbjCpbFJQg6ilBB:-OMT1bbjCpbFJQg6ilBC:b-tzrdlg) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/2f701bbeeb3f05ea4babd84ae127800911a959ae/src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt#L107)):*
> ```Kotlin
> 
>   private suspend fun createWorkItemAttempt(parent: String, workItemAttemptId: String) : WorkItemAttempt {
>     return workItemAttemptsService.createWorkItemAttempt(
> ```

Handle StatusException at the lowest level for any gRPC stub call. If you need to bubble up an exception to be handled by the caller of this function, wrap it in another exception type.

This is covered in the [style guide](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/main/docs/code-style.md#kotlin)

Marco-Premier (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2129)* status: 1 of 8 files reviewed, 5 unresolved discussions (waiting on @SanjayVas and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt` line 78 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2129#-OMT0615-GXgq8pKZZwZ:-OMbSCSS1nLIa1jWWIHO:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/2f701bbeeb3f05ea4babd84ae127800911a959ae/src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt#L78)):*

Explicit types where not obvious. I had to look up QueueSubcriber.QueueMessage to determine this is of type `WorkItem`.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt` line 90 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2129#-OMT0rFk8Ta3tr8Twq3c:-OMbVJsy5F7XxV2T1RH3:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/2f701bbeeb3f05ea4babd84ae127800911a959ae/src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt#L90)):*

Avoid unnecessary mutability. Separate this out into another try block so that `workItemAttempt` can be a non-null val.

```kotlin
val workItemAttempt: WorkItemAttempt = try {
  val workItemAttemptId = UUID.randomUUID().toString()
  createWorkItemAttempt(workItemName, workItemAttemptId)
} catch (e: WhateverCreateThrows) {
  // Handle the exception
}

```
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt` line 107 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2129#-OMT1bbjCpbFJQg6ilBB:-OMbfj2o9LHA_PzdGZOi:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/2f701bbeeb3f05ea4babd84ae127800911a959ae/src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt#L107)):*

Handle StatusException at the lowest level for any gRPC stub call. If you need to bubble up an exception to be handled by the caller of this function, wrap it in another exception type.

This is covered in the [style guide](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/main/docs/code-style.md#kotlin)
</blockquote></details>

Done.
___
*[`src/test/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplicationTest.kt` line 94 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2129#-OMT2Hyx61YIoLQ8ab7q:-OMc8LlV6lijVKa4gwXB:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/2f701bbeeb3f05ea4babd84ae127800911a959ae/src/test/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplicationTest.kt#L94)):*

Never mock gRPC client stubs. You instead mock the service impl and use a test gRPC server, such as GrpcTestServerRule.

Until mockito-inline started shipping by default with mockito-kotlin, gRPC stubs couldn't even be mocked due to being final. This is an intentional restriction.
</blockquote></details>

Done.
___
*[`src/test/kotlin/org/wfanet/measurement/securecomputation/teesdk/BUILD.bazel` line 12 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2129#-OMT3rdx0Q-_8ile0PBo:-OMcGfim63vZAPCs0RyP:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/2f701bbeeb3f05ea4babd84ae127800911a959ae/src/test/kotlin/org/wfanet/measurement/securecomputation/teesdk/BUILD.bazel#L12)):*

This should not depend on the internal API definition. In fact, the package on the internal API definition should have its visibility restricted such that this **cannot** depend on it.
</blockquote></details>

Done.

SanjayVas (changes requested):
> Reviewed 5 of 8 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2129)* status: 5 of 8 files reviewed, 8 unresolved discussions (waiting on @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt` line 91 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2129#-OMibsFH8aMScFbgz4H0:-OMibsFI0imywxfOFkvb:bxfiyam) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/41d68dd2d85983a783caccbb82ecd26081926cec/src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt#L91)):*
> ```Kotlin
>       createWorkItemAttempt(workItemName, workItemAttemptId)
>     } catch (e: ControlPlaneApiException) {
>       return
> ```

We just do nothing in this case? Not even log a warning?
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt` line 92 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2129#-OMicFsM2JYZErFKv1l9:-OMicFsM2JYZErFKv1lA:b6qlfqn) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/41d68dd2d85983a783caccbb82ecd26081926cec/src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt#L92)):*
> ```Kotlin
>     } catch (e: ControlPlaneApiException) {
>       return
> ```

Don't bother with this. You've already caught any exceptions we expected to be thrown. This represents a harder unexpected error, where we should just let the application crash by leaving it uncaught.
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt` line 93 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2129#-OMiZVXm-vfKs7WlC4cn:-OMiZVXm-vfKs7WlC4co:bpxge0g) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/41d68dd2d85983a783caccbb82ecd26081926cec/src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt#L93)):*
> ```Kotlin
>       return
>       logger.severe("Unknown error creating a WorkItemAttempt: $e")
> ```

Don't include exception in message. Instead, use logging method that takes an exception.



_Suggestion:_
```Kotlin
logger.log(Level.SEVERE, e) { "Unknown error creating a WorkItemAttempt" }
```
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt` line 96 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2129#-OMig4fF1ez4beNhArph:-OMig4fF1ez4beNhArpi:b6po5hw) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/41d68dd2d85983a783caccbb82ecd26081926cec/src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt#L96)):*
> ```Kotlin
>       return
>     }
>     try {
> ```

I'm not sure this is matching the behavior from the design doc. There's supposed to be a distinction between retryable and non-retryable errors. For the latter, the behavior should be fail but ACK. I believe being unable to parse the protobuf message fits into that category.

Note that this means you probably want to inspect the status and errorInfo from control plane RPCs to make smarter decisions. e.g. if you go to complete and you get a FAILED_PRECONDITION where the errorinfo says the state is already complete, you can probably just ACK and stop as it means another process took care of things (e.g. duplicate message delivery)
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt` line 124 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2129#-OMi_hKuCkPDiTv8jJxh:-OMi_hKuCkPDiTv8jJxi:b-divyqa) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/41d68dd2d85983a783caccbb82ecd26081926cec/src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt#L124)):*
> ```Kotlin
>     } catch (e: StatusRuntimeException) {
>       val errorMessage = "Error creating WorkItemAttempt"
>       logger.severe("$errorMessage: $e")
> ```

Don't log here since you're throwing an exception.
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt` line 125 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2129#-OMi_xiY1zjVzd9El-Zf:-OMi_xiY1zjVzd9El-Zg:bh23d7u) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/41d68dd2d85983a783caccbb82ecd26081926cec/src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt#L125)):*
> ```Kotlin
>       val errorMessage = "Error creating WorkItemAttempt"
>       logger.severe("$errorMessage: $e")
>       throw ControlPlaneApiException("$errorMessage: ${e.message}")
> ```

Don't blindly pass the API error message since you have no idea if it contains sensitive information or information that's at the wrong level of granularity for whomever might be catching this exception. Instead put whatever message makes sense here with what you do know. The only time it's safe to use it is when it's an error from a known domain, i.e. when you've parsed out the errorInfo and know what to do with it.

Include the caught exception as the cause. This is what I meant by "wrapping" in another exception. Otherwise you lose debug information.

Marco-Premier (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2129)* status: 3 of 8 files reviewed, 8 unresolved discussions (waiting on @SanjayVas and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt` line 91 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2129#-OMibsFH8aMScFbgz4H0:-OMkIdc8DFqpWMKXnKxW:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/41d68dd2d85983a783caccbb82ecd26081926cec/src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt#L91)):*

We just do nothing in this case? Not even log a warning?
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt` line 92 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2129#-OMicFsM2JYZErFKv1l9:-OMkIe_pBDfPVkel7fyh:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/41d68dd2d85983a783caccbb82ecd26081926cec/src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt#L92)):*

Don't bother with this. You've already caught any exceptions we expected to be thrown. This represents a harder unexpected error, where we should just let the application crash by leaving it uncaught.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt` line 93 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2129#-OMiZVXm-vfKs7WlC4cn:-OMkIKlpDxOLfldEzQSK:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/41d68dd2d85983a783caccbb82ecd26081926cec/src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt#L93)):*

Don't include exception in message. Instead, use logging method that takes an exception.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt` line 96 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2129#-OMig4fF1ez4beNhArph:-OMkJ9xz2vjlQDhNhBQ2:b-71vp99) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/41d68dd2d85983a783caccbb82ecd26081926cec/src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt#L96)):*

I'm not sure this is matching the behavior from the design doc. There's supposed to be a distinction between retryable and non-retryable errors. For the latter, the behavior should be fail but ACK. I believe being unable to parse the protobuf message fits into that category.

Note that this means you probably want to inspect the status and errorInfo from control plane RPCs to make smarter decisions. e.g. if you go to complete and you get a FAILED_PRECONDITION where the errorinfo says the state is already complete, you can probably just ACK and stop as it means another process took care of things (e.g. duplicate message delivery)
</blockquote></details>

nice catch! thanks
proto buffer exception definitely acks the message.
For the latter I'm not sure since I'm acking the message before calling `completeWorkItemAttempt`. Overall I don't think grpc exceptions should interfere with `ack` and `nack` that's why I wrapper them inside `runCatching`. 
For example:
1) I'm acking the message before calling `completeWorkItemAttempt`, so if the grpc fails because of an internal problem (which is not state is alredy `SUCCEEDED`), should the work item be processed again?
2) if `failWorkItem` fails, should should I `nack` instead of `ack` the message?

I don't think so. Maybe those are just edge cases where we may end up having inconsistency at DB level for `WorkItemAttempts.State` even though the state of `WorkItem` should always reflect the correct one. Furthermore I'd expect duplicate message delivery to not happen as the system is design to handle 1 message at the time.

WYT?
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt` line 124 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2129#-OMi_hKuCkPDiTv8jJxh:-OMkp2j43tMINaZZFYxF:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/41d68dd2d85983a783caccbb82ecd26081926cec/src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt#L124)):*

Don't log here since you're throwing an exception.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt` line 125 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2129#-OMi_xiY1zjVzd9El-Zf:-OMkq_lG1R6YLQ2JJuqn:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/41d68dd2d85983a783caccbb82ecd26081926cec/src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt#L125)):*

Don't blindly pass the API error message since you have no idea if it contains sensitive information or information that's at the wrong level of granularity for whomever might be catching this exception. Instead put whatever message makes sense here with what you do know. The only time it's safe to use it is when it's an error from a known domain, i.e. when you've parsed out the errorInfo and know what to do with it.

Include the caught exception as the cause. This is what I meant by "wrapping" in another exception. Otherwise you lose debug information.
</blockquote></details>

Done.

SanjayVas (changes requested):
> Reviewed 2 of 3 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2129)* status: 4 of 8 files reviewed, 3 unresolved discussions (waiting on @Marco-Premier and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt` line 96 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2129#-OMig4fF1ez4beNhArph:-OMnHdLx9oeLjvBeAC8_:bd9gcpu) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/41d68dd2d85983a783caccbb82ecd26081926cec/src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt#L96)):*
> I'd expect duplicate message delivery to not happen as the system is design to handle 1 message at the time.

This is not true. The system should handle at-least-once delivery. Consider the simple case where a message gets re-delivered due to missing the ACK deadline. It's possible that the original receiver ends up eventually processing the message before the new receiver.

> Maybe those are just edge cases where we may end up having inconsistency at DB level for `WorkItemAttempts.State` even though the state of `WorkItem` should always reflect the correct one.

The DB must always be self-consistent, and I believe the service implementation guarantees this. What may not be consistent is that the message state may not match the API resource state. Consider the case where the server crashes after writing the WorkItem but before publishing the corresponding message.

> I'm acking the message before calling `completeWorkItemAttempt`

I don't think that's correct, nor does it match what is in the design doc. The diagram there shows calling completeWorkItem first and *then* ACKing the message. The same is true of the failWorkItemAttempt case  where it should be called and *then* NACKed.

The basic idea is that you want to get the resource state correct first and then worry about eventually getting the message state right. Since it's at-least-once delivery, you want to gracefully handle the case where a message may have already been processed - hence things like ACKing the message if the WorkItem is already in a terminal state.

> if `failWorkItem` fails, should should I `nack` instead of `ack` the message?

NACK really just means "retry at a higher level", where the number of retries is determined by the configured maximum number of delivery attempts. In the case that you're calling failWorkItem, you already know the error is unrecoverable - i.e. that retrying at a higher level won't do anything. Basically, there's a known failure case where we stop trying to deliver the message and give up regardless of the WorkItem state. The dead-letter processing should hopefully eventually resolve the WorkItem state.

Marco-Premier (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2129)* status: 3 of 8 files reviewed, 3 unresolved discussions (waiting on @SanjayVas and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt` line 96 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2129#-OMig4fF1ez4beNhArph:-OMpMcPd6Vspe0uzJ6h-:b-90zvg7) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/41d68dd2d85983a783caccbb82ecd26081926cec/src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt#L96)):*
> This is not true. The system should handle at-least-once delivery. Consider the simple case where a message gets re-delivered due to missing the ACK deadline. It's possible that the original receiver ends up eventually processing the message before the new receiver.

Ideally this shouldn't happen since we extend the ack deadline up to 6 hours but in general I agree that there might be re-delivery due to google pubsub errors.

> The DB must always be self-consistent, and I believe the service implementation guarantees this. What may not be consistent is that the message state may not match the API resource state. Consider the case where the server crashes after writing the WorkItem but before publishing the corresponding message.

I fixed the flow by follow the diagram in the doc

> The basic idea is that you want to get the resource state correct first and then worry about eventually getting the message state right. Since it's at-least-once delivery, you want to gracefully handle the case where a message may have already been processed - hence things like ACKing the message if the WorkItem is already in a terminal state.

I added this check now

> The dead-letter processing should hopefully eventually resolve the WorkItem state.

Yes but I was planning to add that in a following PR as this is already long-living


Overall I think now the flow conform to the desing doc and better handle the case where duplicated message are delivered or other exceptions arise while updating the message state against spanner db.

SanjayVas (approved):
> Reviewed 3 of 8 files at r3, 1 of 3 files at r4, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2129)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @stevenwarejones)

Marco-Premier (commented):
:lgtm:

> Reviewed 6 of 6 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2129#-:-ON-i55-2oKcI4gyGx7V:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @stevenwarejones)

PR #2145, opened 2025-04-03
ci: Add test coverage report for cc tests

laureanobrs:


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2145)

SanjayVas (changes requested):
> Reviewed 8 of 8 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2145)* status: all files reviewed, 3 unresolved discussions (waiting on @laureanobrs)

___
*[`.bazelrc` line 9 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2145#-OMxWx5L4_O0T0x_ht9l:-OMxWx5M4ZOl16Ak9B-R:bdzzvyf) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/05e1a35558ef6849b7d1ff86b22a6815caeba275/.bazelrc#L9)):*
> ```
> build --repo_env='CC=clang'
> 
> # Configuration for running CC coverage locally
> ```

Put these in an explicit config. e.g. `common:llvm-cov` so that these are only used when passing `--config=llvm-cov`
___
*[`.bazelrc` line 12 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2145#-OMxWVq66lvTneN0sjac:-OMxWVq66lvTneN0sjad:b-nm6kra) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/05e1a35558ef6849b7d1ff86b22a6815caeba275/.bazelrc#L12)):*
> ```
> # See https://github.com/bazelbuild/bazel/blob/106903d38f6d41b5eb1f100698f257184bea47a3/src/test/shell/bazel/bazel_coverage_cc_test_llvm.sh#L59
> # and https://github.com/bazelbuild/bazel/commit/21b5eb627d78c09d47c4de957e9e6b56a9ae6fad
> common --repo_env=BAZEL_LLVM_COV=/usr/bin/llvm-cov
> ```

We don't want to hard-code local paths as we can't assume every developer has these installed at these locations. We can only do that under the `container` config since we know which image is being used.

Do you know if this searches the PATH, e.g. if we can leave off the `/usr/bin`?



_Code quote:_
```
/usr/bin/llvm-cov
```
___
*[`.github/workflows/build-test.yml` line 102 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2145#-OMxXX_P0yL33MHd7ZHz:-OMxXX_P0yL33MHd7ZI-:b-1kw72p) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/05e1a35558ef6849b7d1ff86b22a6815caeba275/.github/workflows/build-test.yml#L102)):*
> ```YAML
>       - name: Run tests
>         id: run-tests
>         run: bazel test -- //... -//src/test/python/... -//src/test/cc/...
> ```

Why exclude these from the regular test? We still want to make sure these pass in this configuration even if they also pass with the coverage run. Meaning passing only when using `bazel coverage` is not sufficient, since that might introduce slightly different behavior.



_Code quote:_
```YAML
-//src/test/python/... -//src/test/cc/...
```

laureanobrs (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2145)* status: 6 of 8 files reviewed, 3 unresolved discussions (waiting on @SanjayVas)

___
*[`.bazelrc` line 9 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2145#-OMxWx5L4_O0T0x_ht9l:-ON1KrJK72nyPDh30Lfi:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/05e1a35558ef6849b7d1ff86b22a6815caeba275/.bazelrc#L9)):*

Put these in an explicit config. e.g. `common:llvm-cov` so that these are only used when passing `--config=llvm-cov`
</blockquote></details>

Done.
___
*[`.bazelrc` line 12 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2145#-OMxWVq66lvTneN0sjac:-ON1KNbDBHH5CuuT0CGY:b3isujo) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/05e1a35558ef6849b7d1ff86b22a6815caeba275/.bazelrc#L12)):*

We don't want to hard-code local paths as we can't assume every developer has these installed at these locations. We can only do that under the `container` config since we know which image is being used.

Do you know if this searches the PATH, e.g. if we can leave off the `/usr/bin`?
</blockquote></details>

Good point. It works without hard-coding the paths
___
*[`.github/workflows/build-test.yml` line 102 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2145#-OMxXX_P0yL33MHd7ZHz:-ON1L0sa35AualmTAR1t:bh3pv5z) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/05e1a35558ef6849b7d1ff86b22a6815caeba275/.github/workflows/build-test.yml#L102)):*

Why exclude these from the regular test? We still want to make sure these pass in this configuration even if they also pass with the coverage run. Meaning passing only when using `bazel coverage` is not sufficient, since that might introduce slightly different behavior.
</blockquote></details>

I did the exclusion to avoid having the same test take twice as long (running all tests takes 38 minutes in GHA).
AFAIK If we run the tests only with bazel coverage and a test fails, it might not be clear whether the failure comes from an instrumentation/coverage issue or from a change in the code. Is there another difference?
I'm OK with running both test and coverage, and if we start having problems because of how long it takes, we can revisit this issue.
I updated the PR to run both test and coverage.

SanjayVas (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2145)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @laureanobrs)

___
*[`.github/workflows/build-test.yml` line 102 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2145#-OMxXX_P0yL33MHd7ZHz:-ON1dNW2ERAVQyLdOuG6:b-mxnjiw) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/05e1a35558ef6849b7d1ff86b22a6815caeba275/.github/workflows/build-test.yml#L102)):*
> if we start having problems because of how long it takes, we can revisit this issue

If this becomes an issue, we can split coverage out of this workflow so it doesn't block merging.

PR #2161, opened 2025-04-08
add fake in process kingdom

stevenwarejones:
This is meant to be used in scenarios where the kingdom isn't directly controlled in the test.

In the case of the EDP Aggregator End To End Test, the following will occur.
1. The requisition fetcher will fetch requisitions
2. The event group sync will sync event groups
3. The TEE fulfiller app will fulfill to the kingdom

In the EDP Aggregator End to End Test, the fake kingdom is needed for event group registration, fetching and fulfilling requisitions.


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2161)

kungfucraig (commented):
A couple questions here, which I also asked offline. 1) Is there any reason you cannot use the real Kingdom. 2) If not I think we should have a design doc for this. In the limit a fake Kingdom will become very complicated.

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2161#-:-ONMKLWt75eo_Yr-kDC4:b5czvyn)* status: 0 of 4 files reviewed, all discussions resolved (waiting on @Marco-Premier)

stevenwarejones:
Talked with @kungfucraig more. I think we'll close this in lieu of using a real kingdom.

cc: @Marco-Premier

PR #2218, opened 2025-04-24
feat: add requisition fetcher deployment

Marco-Premier:

PR #2232, opened 2025-04-28
feat: add requisition fetcher deployment

Marco-Premier:


Marco-Premier (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2232)* status: 0 of 11 files reviewed, 1 unresolved discussion (waiting on @kungfucraig, @SanjayVas, and @stevenwarejones)

___
*[`.github/workflows/build_and_deploy_cloud_functions.yml` line 95 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2232#-OOxGQJg3CCEXQXprjYV:-OOxGQJg3CCEXQXprjYW:b-hjd44r) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/066974b30b650b83313159293b1cb7a05d38c2c1/.github/workflows/build_and_deploy_cloud_functions.yml#L95)):*
> ```YAML
>           --set-env-vars="CERT_JAR_RESOURCE_PATH=src/main/k8s/testing/secretfiles/data_watcher_tls.pem,PRIVATE_KEY_JAR_RESOURCE_PATH=src/main/k8s/testing/secretfiles/data_watcher_tls.key,CERT_COLLECTION_JAR_RESOURCE_PATH=src/main/k8s/testing/secretfiles/secure_computation_root.pem,CONTROL_PLANE_TARGET=$SECURE_COMPUTATION_PUBLIC_API_TARGET,CONTROL_PLANE_CERT_HOST=data-watcher.secure-computation.dev.halo-cmm.org"
> 
>       - name: Deploy Requisition Fetcher Function
> ```

Comments from #2218 for discussion:

Craig: "It seems like we need a reusable work flow for deploying and build a cloud function that we can send parameters to. When you add the third one, please refactor it as such."

Sanjay: "Do you mean an Action in this case rather than a Workflow? Meaning extracting the common bits into a separate Action that can be reused?"

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2232)* status: all files reviewed, 3 unresolved discussions (waiting on @kungfucraig)

___
*[`src/main/terraform/gcloud/cmms/edp_aggregator.tf` line 48 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2232#-OOxx9NV3oGFBRb4ks4p:-OOxx9NV3oGFBRb4ks4q:bpzdhec) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/066974b30b650b83313159293b1cb7a05d38c2c1/src/main/terraform/gcloud/cmms/edp_aggregator.tf#L48)):*
From https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2235#-OOxuDPp1zVCrtf-C3yX:

This must be configurable as GCS bucket names are globally unique. This results in the following error:

> The requested bucket name is not available. The bucket namespace is shared by all users of the system. Please select a different name and try again.
___
*[`.github/workflows/build_and_deploy_cloud_functions.yml` line 113 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2232#-OOxwfK14sUIFgBViZ5q:-OOxwfK28fcRnRetegBa:b-iajyc3) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/066974b30b650b83313159293b1cb7a05d38c2c1/.github/workflows/build_and_deploy_cloud_functions.yml#L113)):*
> ```YAML
>           --source="$RUNNER_TEMP/requisition_fetcher_source"
>           --run-service-account=$REQUISITION_FETCHER_SERVICE_ACCOUNT
>           --set-env-vars="KINGDOM_TARGET=$KINGDOM_PUBLIC_API_TARGET,KINGDOM_CERT_HOST=requisition-fetcher.kingdom.dev.halo-cmm.org"
> ```

Cert host shouldn't need to be specified since the target DNS matches a subjectAltName in the certificate.



_Code quote:_
```YAML
KINGDOM_CERT_HOST=requisition-fetcher.kingdom.dev.halo-cmm.org
```

kungfucraig (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2232)* status: all files reviewed, 5 unresolved discussions (waiting on @Marco-Premier and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/requisitionfetcher/RequisitionFetcherFunction.kt` line 41 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2232#-OP1Zewa8bZHGZ7A-5Ft:-OP1Zewa8bZHGZ7A-5Fu:bcqbo4x) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/066974b30b650b83313159293b1cb7a05d38c2c1/src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/requisitionfetcher/RequisitionFetcherFunction.kt#L41)):*
> ```Kotlin
>     for (dataProviderConfig in requisitionFetcherConfig.configsList) {
> 
>       val fileSystemPath = System.getenv("REQUISITION_FILE_SYSTEM_PATH")
> ```

This needs to be in a separate PR.
___
*[`.github/workflows/build_and_deploy_cloud_functions.yml` line 59 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2232#-OP1Zo91C-hZEjdFa9NB:-OP1Zo91C-hZEjdFa9NC:b-uv8cif) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/066974b30b650b83313159293b1cb7a05d38c2c1/.github/workflows/build_and_deploy_cloud_functions.yml#L59)):*
> ```YAML
>           //src/main/kotlin/org/wfanet/measurement/securecomputation/deploy/gcloud/datawatcher:DataWatcherFunction_deploy.jar
> 
>       - name: Copy uber JAR of DataWatcher to temp directory
> ```

Between this and your other PR for the the EG Sync this pattern is being repeated three times in this file. We need to introduce some abstraction.

See: https://docs.github.com/en/actions/sharing-automations/avoiding-duplication

kungfucraig (changes requested):
> Reviewed 3 of 11 files at r1, 9 of 10 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2232)* status: 12 of 13 files reviewed, 12 unresolved discussions (waiting on @Marco-Premier and @SanjayVas)

___
*[`.github/workflows/build_and_push_cloud_function.yml` line 1 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2232#-OPBz_8nARwTuGNKInZc:-OPBz_8nARwTuGNKInZd:b-8xnt05) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3c7edb0f53384cf4c2ce5cd01bfd7aa0664a0223/.github/workflows/build_and_push_cloud_function.yml#L1)):*
> ```YAML
> # Copyright 2025 The Cross-Media Measurement Authors
> ```

rename to "build and deploy" ? That's the verb we're typically using and is consistent with your name below.
___
*[`.github/workflows/build_and_push_cloud_function.yml` line 21 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2232#-OPBzonVCBj0JcO5zlgV:-OPBzonVCBj0JcO5zlgW:b-t6r9mp) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3c7edb0f53384cf4c2ce5cd01bfd7aa0664a0223/.github/workflows/build_and_push_cloud_function.yml#L21)):*
> ```YAML
>     inputs:
>       environment:
>         description: "GitHub environment (dev/stage/prod)"
> ```

I don't think these are the values we use. I believe they are dev, head, and qa, but @SanjayVas would know best.
___
*[`.github/workflows/build_and_push_cloud_function.yml` line 56 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2232#-OPC-4_j9AXnHOFX08aH:-OPC-4_j9AXnHOFX08aI:b-5s3sdk) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3c7edb0f53384cf4c2ce5cd01bfd7aa0664a0223/.github/workflows/build_and_push_cloud_function.yml#L56)):*
> ```YAML
>         required: false
> 
>       # everything below is per-function, so we make them inputs
> ```

Isn't everything above an input too?
___
*[`.github/workflows/build_and_push_cloud_function.yml` line 132 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2232#-OPCDKWm1xhjbQSdgLsm:-OPCDKWm1xhjbQSdgLsn:btc6gwo) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3c7edb0f53384cf4c2ce5cd01bfd7aa0664a0223/.github/workflows/build_and_push_cloud_function.yml#L132)):*
> ```YAML
>             ARGS+=(
>               --trigger-http
>               --allow-unauthenticated
> ```

This seems unsafe. If it's not please add some comments describing why.
___
*[`src/main/terraform/gcloud/cmms/edp_aggregator.tf` line 53 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2232#-OPCE_WZ76aPfZvi5kC9:-OPCE_WZ76aPfZvi5kCA:b-vmv6dg) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3c7edb0f53384cf4c2ce5cd01bfd7aa0664a0223/src/main/terraform/gcloud/cmms/edp_aggregator.tf#L53)):*
> ```HCL
>   data_watcher_trigger_service_account_name = "data-watcher-trigger"
>   terraform_service_account                 = var.terraform_service_account
>   requisition_fetcher_service_account_name  = "requisition-fetcher"
> ```

Did we discuss using the same service account for all EDP Aggregator apps?
___
*[`.github/workflows/deploy_cloud_functions.yml` line 1 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2232#-OPCEOiUEvM-VpYkKGJC:-OPCEOiUEvM-VpYkKGJD:b-qmb1u1) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3c7edb0f53384cf4c2ce5cd01bfd7aa0664a0223/.github/workflows/deploy_cloud_functions.yml#L1)):*
> ```YAML
> # Copyright 2025 The Cross-Media Measurement Authors
> ```

Should this be deploy_edp_aggregator_cloud_functions.yml?
___
*[`.github/workflows/deploy_cloud_functions.yml` line 20 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2232#-OPCE-eG3PZ3u5qV0P03:-OPCE-eG3PZ3u5qV0P04:bngt6d9) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3c7edb0f53384cf4c2ce5cd01bfd7aa0664a0223/.github/workflows/deploy_cloud_functions.yml#L20)):*
> ```YAML
>   push:
>     branches:
>       - marcopremier/add_requisition_fetcher_deployment2
> ```

This doesn't seem right.

Marco-Premier (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2232)* status: 10 of 13 files reviewed, 12 unresolved discussions (waiting on @kungfucraig and @SanjayVas)

___
*[`.github/workflows/build_and_deploy_cloud_functions.yml` line 59 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2232#-OP1Zo91C-hZEjdFa9NB:-OPFOaN4BlWnT_6syJtv:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/066974b30b650b83313159293b1cb7a05d38c2c1/.github/workflows/build_and_deploy_cloud_functions.yml#L59)):*

Between this and your other PR for the the EG Sync this pattern is being repeated three times in this file. We need to introduce some abstraction.

See: https://docs.github.com/en/actions/sharing-automations/avoiding-duplication
</blockquote></details>

Done.
___
*[`.github/workflows/build_and_deploy_cloud_functions.yml` line 113 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2232#-OOxwfK14sUIFgBViZ5q:-OP5OPHYA-43deVdyENF:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/066974b30b650b83313159293b1cb7a05d38c2c1/.github/workflows/build_and_deploy_cloud_functions.yml#L113)):*

Cert host shouldn't need to be specified since the target DNS matches a subjectAltName in the certificate.
</blockquote></details>

Done.
___
*[`.github/workflows/build_and_push_cloud_function.yml` line 1 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2232#-OPBz_8nARwTuGNKInZc:-OPFOpfoEgPAybASDBNc:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3c7edb0f53384cf4c2ce5cd01bfd7aa0664a0223/.github/workflows/build_and_push_cloud_function.yml#L1)):*

rename to "build and deploy" ? That's the verb we're typically using and is consistent with your name below.
</blockquote></details>

Done.
___
*[`.github/workflows/build_and_push_cloud_function.yml` line 21 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2232#-OPBzonVCBj0JcO5zlgV:-OPFP4S16pbRJubltycA:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3c7edb0f53384cf4c2ce5cd01bfd7aa0664a0223/.github/workflows/build_and_push_cloud_function.yml#L21)):*

I don't think these are the values we use. I believe they are dev, head, and qa, but @SanjayVas would know best.
</blockquote></details>

Done.
___
*[`.github/workflows/build_and_push_cloud_function.yml` line 56 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2232#-OPC-4_j9AXnHOFX08aH:-OPFP6gKCn7lXzaJvy8M:b-74elnk) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3c7edb0f53384cf4c2ce5cd01bfd7aa0664a0223/.github/workflows/build_and_push_cloud_function.yml#L56)):*

Isn't everything above an input too?
</blockquote></details>

sorry I was still working on this, I don't think I had published the changes  yet. Now I did :)
___
*[`.github/workflows/build_and_push_cloud_function.yml` line 132 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2232#-OPCDKWm1xhjbQSdgLsm:-OPFV6df-H4ozrJK0flH:b-qfce65) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3c7edb0f53384cf4c2ce5cd01bfd7aa0664a0223/.github/workflows/build_and_push_cloud_function.yml#L132)):*

This seems unsafe. If it's not please add some comments describing why.
</blockquote></details>

Correct, it's not safe. I added a TODO, can I leave it there just for the sake of the initial cloud test?
Currently the requisition fetcher will be called once from the cloud test but then this will be removed in favour of a recurrent polling which will need Cloud Scheduler.
___
*[`.github/workflows/deploy_cloud_functions.yml` line 1 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2232#-OPCEOiUEvM-VpYkKGJC:-OPFPLvT5PiPTM5f72wT:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3c7edb0f53384cf4c2ce5cd01bfd7aa0664a0223/.github/workflows/deploy_cloud_functions.yml#L1)):*

Should this be deploy_edp_aggregator_cloud_functions.yml?
</blockquote></details>

Done.
___
*[`.github/workflows/deploy_cloud_functions.yml` line 20 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2232#-OPCE-eG3PZ3u5qV0P03:-OPFPbcS9sS3VTTkyHdm:bywiqjt) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3c7edb0f53384cf4c2ce5cd01bfd7aa0664a0223/.github/workflows/deploy_cloud_functions.yml#L20)):*

This doesn't seem right.
</blockquote></details>

ditto: as per above I hadn't published changes yet
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/requisitionfetcher/RequisitionFetcherFunction.kt` line 41 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2232#-OP1Zewa8bZHGZ7A-5Ft:-OPFXl6x1-15AFvAwUbi:b-vozdeb) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/066974b30b650b83313159293b1cb7a05d38c2c1/src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/requisitionfetcher/RequisitionFetcherFunction.kt#L41)):*

This needs to be in a separate PR.
</blockquote></details>

Tis fix is to make the deployment working as the 'REQUISITION_FILE_SYSTEM_PATH' is only for testing purposes and I'm not passing that flag with the deployment, which would fail otherwise.

You want me to move it into a separate pr anyway?
___
*[`src/main/terraform/gcloud/cmms/edp_aggregator.tf` line 48 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2232#-OOxx9NV3oGFBRb4ks4p:-OPFQP7s5pJKx9d3TEpU:b-9pev6d) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/066974b30b650b83313159293b1cb7a05d38c2c1/src/main/terraform/gcloud/cmms/edp_aggregator.tf#L48)):*

From https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2235#-OOxuDPp1zVCrtf-C3yX:

This must be configurable as GCS bucket names are globally unique. This results in the following error:

> The requested bucket name is not available. The bucket namespace is shared by all users of the system. Please select a different name and try again.
</blockquote></details>

This was fixed already in # #2235 
___
*[`src/main/terraform/gcloud/cmms/edp_aggregator.tf` line 53 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2232#-OPCE_WZ76aPfZvi5kC9:-OPFQdNmDV8j_jNBqfQ5:blsdw2h) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3c7edb0f53384cf4c2ce5cd01bfd7aa0664a0223/src/main/terraform/gcloud/cmms/edp_aggregator.tf#L53)):*

Did we discuss using the same service account for all EDP Aggregator apps?
</blockquote></details>

Actually no. My thought is that since we can have 2 different types of cloud functions (eg. storage events triggered / standard http) and they needs different roles attached to work, I had preferred a more granular approach, for instance data-watcher need: 'roles/eventarc.eventReceiver' / 'roles/run.invoker'  while event group sync and requisition fetcher does not.
Also I was thinking that in this way each service account can "scale" permissions independently.

You think it's best to only have a single SA that wraps all the permissions needed by all EDP-agg cloud functions?

kungfucraig (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2232)* status: 9 of 14 files reviewed, 12 unresolved discussions (waiting on @Marco-Premier and @SanjayVas)

___
*[`src/main/terraform/gcloud/cmms/edp_aggregator.tf` line 53 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2232#-OPCE_WZ76aPfZvi5kC9:-OPGYN2aEAjVrdfLVvMv:bzffzvp) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3c7edb0f53384cf4c2ce5cd01bfd7aa0664a0223/src/main/terraform/gcloud/cmms/edp_aggregator.tf#L53)):*

Actually no. My thought is that since we can have 2 different types of cloud functions (eg. storage events triggered / standard http) and they needs different roles attached to work, I had preferred a more granular approach, for instance data-watcher need: 'roles/eventarc.eventReceiver' / 'roles/run.invoker'  while event group sync and requisition fetcher does not.
Also I was thinking that in this way each service account can "scale" permissions independently.

You think it's best to only have a single SA that wraps all the permissions needed by all EDP-agg cloud functions?
</blockquote></details>

We can do N services accounts or 1 service account. Anything else is confusing. I agree that the N approach makes things generally more secure and am fine going that route.

However, we need to come up with a convention. Because the data watcher is reusable across apps, we should make sure we're scoping the names to the app not just the reusable components. So in this case, I think we'd want "edpa-data-watcher" and "edpa-requisition-fetcher." Wdyt?

Marco-Premier (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2232)* status: 6 of 16 files reviewed, 12 unresolved discussions (waiting on @kungfucraig and @SanjayVas)

___
*[`src/main/terraform/gcloud/cmms/edp_aggregator.tf` line 53 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2232#-OPCE_WZ76aPfZvi5kC9:-OPUQukV2slPApfGl-qb:b-8fgj5i) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3c7edb0f53384cf4c2ce5cd01bfd7aa0664a0223/src/main/terraform/gcloud/cmms/edp_aggregator.tf#L53)):*

We can do N services accounts or 1 service account. Anything else is confusing. I agree that the N approach makes things generally more secure and am fine going that route.

However, we need to come up with a convention. Because the data watcher is reusable across apps, we should make sure we're scoping the names to the app not just the reusable components. So in this case, I think we'd want "edpa-data-watcher" and "edpa-requisition-fetcher." Wdyt?
</blockquote></details>

yeah, I like that, thanks

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2232)* status: 6 of 16 files reviewed, 16 unresolved discussions (waiting on @kungfucraig)

___
*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2232#-OPWi9T99Vq52sR3b80H:-OPWi9T99Vq52sR3b80I:b-vcoj78) (no related file):*
I'd personally prefer a composite Action for this rather than a reusable (callable) Workflow. I had mentioned this in a comment where @kungfucraig had requested extracting things out to avoid duplication. This does mean that:

1. The parts that rely on secrets have to be done in the Workflow that calls the Action
2. GitHub environment variables need to be passed in as inputs
3. To maximally avoid duplication, the separate functions would be steps rather than jobs, meaning they run in sequence rather than in parallel. The tradeoff here is that there's less overhead, as you only need to do the setup (e.g. authenticating to Google Cloud) once.

I think all of those are worth it for keeping things lighter-weight.
___
*[`.github/workflows/build_and_deploy_cloud_function.yml` line 35 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2232#-OPWhBzs5FYGkHwN6Pt1:-OPWhBzs5FYGkHwN6Pt2:bet0pax) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e7ea24f15bc6e11d0d1bf5cd33f98955447dcd44/.github/workflows/build_and_deploy_cloud_function.yml#L35)):*
> ```YAML
>         required: true
> 
>       source_path:
> ```

This is not a path, but rather a Bazel target label.



_Code quote:_
```YAML
source_path
```
___
*[`.github/workflows/build_and_deploy_cloud_function.yml` line 40 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2232#-OPWhIVP1sYGVF8yplZY:-OPWhIVP1sYGVF8yplZZ:b-9ckiot) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e7ea24f15bc6e11d0d1bf5cd33f98955447dcd44/.github/workflows/build_and_deploy_cloud_function.yml#L40)):*
> ```YAML
>         required: true
> 
>       uber_jar_path:
> ```

You shouldn't need to pass this in. It's discoverable using `bazel cquery`. See https://bazel.build/query/cquery#files-output



_Code quote:_
```YAML
 uber_jar_path
```
___
*[`.github/workflows/build_and_deploy_cloud_function.yml` line 45 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2232#-OPWi2fn0zuiO3bP33UL:-OPWi2fo6kfSro1g5J68:b7henpc) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e7ea24f15bc6e11d0d1bf5cd33f98955447dcd44/.github/workflows/build_and_deploy_cloud_function.yml#L45)):*
> ```YAML
>         required: true
> 
>       trigger_type:
> ```

Do you need this input, or can you just check whether `trigger_bucket` is set?



_Code quote:_
```YAML
trigger_type
```

kungfucraig (changes requested):
> Reviewed 1 of 10 files at r2, 4 of 4 files at r3, 5 of 5 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2232)* status: all files reviewed, 10 unresolved discussions (waiting on @Marco-Premier and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/requisitionfetcher/RequisitionFetcherFunction.kt` line 41 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2232#-OP1Zewa8bZHGZ7A-5Ft:-OPX7Kx2Djappv645OSn:b-d9pzdb) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/066974b30b650b83313159293b1cb7a05d38c2c1/src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/requisitionfetcher/RequisitionFetcherFunction.kt#L41)):*

Tis fix is to make the deployment working as the 'REQUISITION_FILE_SYSTEM_PATH' is only for testing purposes and I'm not passing that flag with the deployment, which would fail otherwise.

You want me to move it into a separate pr anyway?
</blockquote></details>

You can leave it. Do we want a TODO and/or an issue to remove this code, or is this something we want to keep for future testing? If the latter can you add some comments that describes the test use case?
___
*[`.github/workflows/build_and_push_cloud_function.yml` line 132 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2232#-OPCDKWm1xhjbQSdgLsm:-OPX6LKY4WiXX__kUA3w:bljfxnv) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3c7edb0f53384cf4c2ce5cd01bfd7aa0664a0223/.github/workflows/build_and_push_cloud_function.yml#L132)):*

Correct, it's not safe. I added a TODO, can I leave it there just for the sake of the initial cloud test?
Currently the requisition fetcher will be called once from the cloud test but then this will be removed in favour of a recurrent polling which will need Cloud Scheduler.
</blockquote></details>

Is there any access control over these (e.g. call coming within the same network) or would these be unauthenticated and open to the Internet?

Who is the user that is going to be making the call for the initial cloud test? Can we set that user up with permissions?
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/requisitionfetcher/RequisitionFetcherFunction.kt` line 100 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2232#-OPX7bpqAv1Dkzye3s6R:-OPX7bpqAv1Dkzye3s6S:bldeusp) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e7ea24f15bc6e11d0d1bf5cd33f98955447dcd44/src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/requisitionfetcher/RequisitionFetcherFunction.kt#L100)):*
> ```Kotlin
>   companion object {
>     private val kingdomTarget = checkNotEmpty("KINGDOM_TARGET")
>     private val kingdomCertHost: String? = System.getenv("KINGDOM_CERT_HOST")
> ```

Can we set this to a spurious value for testing? Allow it to be null creates an unnecessary failure mode for a real instance.

Marco-Premier (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2232)* status: 12 of 17 files reviewed, 10 unresolved discussions (waiting on @kungfucraig and @SanjayVas)

___
*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2232#-OPWi9T99Vq52sR3b80H:-OP_P3433z9vOF2O-sRV:b-ojm1jw) (no related file):*

I'd personally prefer a composite Action for this rather than a reusable (callable) Workflow. I had mentioned this in a comment where @kungfucraig had requested extracting things out to avoid duplication. This does mean that:

1. The parts that rely on secrets have to be done in the Workflow that calls the Action
2. GitHub environment variables need to be passed in as inputs
3. To maximally avoid duplication, the separate functions would be steps rather than jobs, meaning they run in sequence rather than in parallel. The tradeoff here is that there's less overhead, as you only need to do the setup (e.g. authenticating to Google Cloud) once.

I think all of those are worth it for keeping things lighter-weight.
</blockquote></details>

refactored using a composite action
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/requisitionfetcher/RequisitionFetcherFunction.kt` line 41 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2232#-OP1Zewa8bZHGZ7A-5Ft:-OP_T9M91Wa6UemeFnrw:beh8at9) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/066974b30b650b83313159293b1cb7a05d38c2c1/src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/requisitionfetcher/RequisitionFetcherFunction.kt#L41)):*

You can leave it. Do we want a TODO and/or an issue to remove this code, or is this something we want to keep for future testing? If the latter can you add some comments that describes the test use case?
</blockquote></details>

I added a comment. This code was added in an old pr and here I just had to fix the way the env var was retrieved not to cause nullpointexception

 #2007  
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/requisitionfetcher/RequisitionFetcherFunction.kt` line 100 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2232#-OPX7bpqAv1Dkzye3s6R:-OP_U71RBDsxOH3454QC:bi6frjc) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e7ea24f15bc6e11d0d1bf5cd33f98955447dcd44/src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/requisitionfetcher/RequisitionFetcherFunction.kt#L100)):*

Can we set this to a spurious value for testing? Allow it to be null creates an unnecessary failure mode for a real instance.
</blockquote></details>

`KINGDOM_CERT_HOST` is only needed for testing. In production that value is neither expected nor passed in the git workflow `deploy_edp_aggregator_cloud_functions.yml`.

I had this values always set previously but Sanjay pointed out that is not actually needed as the function `buildMutualTlsChannel` needs to override the authority only when testing and set it to `localhost` 
___
*[`.github/workflows/build_and_deploy_cloud_function.yml` line 35 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2232#-OPWhBzs5FYGkHwN6Pt1:-OP_PPVR0gq9i_eG-PEl:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e7ea24f15bc6e11d0d1bf5cd33f98955447dcd44/.github/workflows/build_and_deploy_cloud_function.yml#L35)):*

This is not a path, but rather a Bazel target label.
</blockquote></details>

Done.
___
*[`.github/workflows/build_and_deploy_cloud_function.yml` line 45 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2232#-OPWi2fn0zuiO3bP33UL:-OP_PsaCCGLA_nUgudeI:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e7ea24f15bc6e11d0d1bf5cd33f98955447dcd44/.github/workflows/build_and_deploy_cloud_function.yml#L45)):*

Do you need this input, or can you just check whether `trigger_bucket` is set?
</blockquote></details>

Done.
___
*[`.github/workflows/build_and_push_cloud_function.yml` line 132 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2232#-OPCDKWm1xhjbQSdgLsm:-OP_Qdkz1fDD2U-Eg6ni:b-kneiw4) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3c7edb0f53384cf4c2ce5cd01bfd7aa0664a0223/.github/workflows/build_and_push_cloud_function.yml#L132)):*

Is there any access control over these (e.g. call coming within the same network) or would these be unauthenticated and open to the Internet?

Who is the user that is going to be making the call for the initial cloud test? Can we set that user up with permissions?
</blockquote></details>

This should not be open to the internet. `event group sync` will be called from the `datawatcher` while `requisition fetcher` will be automatically invoke (polling) by (eg.) `Google Cloud Scheduler`.

So I will indeed remove the `--allow-unauthenticated` and setup proper service account to make those cloud functions reachable ONLY by datawatcher and cloud scheduler. 

For this initial version of the cloud test I wanted to have the endpoint for requisition fetcher open and be able to call it from the cloud test once in order to retrieve requisitions.

I want to prioritize the cloud test itself now over setting up cloud watcher / services accounts, but I'll do that as soon the cloud test is working, is that ok?

SanjayVas (changes requested):
> Reviewed 1 of 10 files at r2, 1 of 4 files at r3, 2 of 5 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2232)* status: 14 of 17 files reviewed, 12 unresolved discussions (waiting on @kungfucraig)

___
*[`.github/actions/build-and-deploy-cloud-function/action.yml` line 40 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2232#-OPamYEXB_mdgnIwZ2ES:-OPamYEXB_mdgnIwZ2ET:b-ee5jih) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0cc277ce4d36c851e4336cdb0d2027c296425483/.github/actions/build-and-deploy-cloud-function/action.yml#L40)):*
> ```YAML
>     required: true
>   bazel_target_label:
>     description: "Bazel target label"
> ```

nit: document that this must be a target that outputs an uber JAR.
___
*[`.github/actions/build-and-deploy-cloud-function/action.yml` line 59 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2232#-OPalYr0DpzD-XCjdhiw:-OPalYr0DpzD-XCjdhix:brgj1wr) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0cc277ce4d36c851e4336cdb0d2027c296425483/.github/actions/build-and-deploy-cloud-function/action.yml#L59)):*
> ```YAML
>   steps:
> 
>     - name: Authenticate to GCP
> ```

Do the auth and setup steps in the workflow outside of the action, as I described in https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2232#-OPWi9T99Vq52sR3b80H
___
*[`.github/actions/build-and-deploy-cloud-function/action.yml` line 79 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2232#-OPamS4P-UYCFi6vAK63:-OPamS4P-UYCFi6vAK64:b-sn74fm) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0cc277ce4d36c851e4336cdb0d2027c296425483/.github/actions/build-and-deploy-cloud-function/action.yml#L79)):*
> ```YAML
>       run: |
>         JAR=$(bazel cquery "${{ inputs.bazel_target_label }}" --output=files \
>             | grep -E '_deploy\.jar$' \
> ```

You shouldn't need this, as `_deploy.jar` targets only have one output.
___
*[`.github/actions/build-and-deploy-cloud-function/action.yml` line 82 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2232#-OPamsWTDE2h1bQZ0IdJ:-OPamsWTDE2h1bQZ0IdK:b-ymzsph) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0cc277ce4d36c851e4336cdb0d2027c296425483/.github/actions/build-and-deploy-cloud-function/action.yml#L82)):*
> ```YAML
>             | head -n1)
> 
>         mkdir -p "$RUNNER_TEMP/source"
> ```

Note that all steps in a job run with the same filesystem, so previous steps may be writing to this same folder if you don't choose a sufficiently unique name. Consider using the `GITHUB_ACTION` environment variable. See https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables#default-environment-variables

kungfucraig (commented):
> Reviewed 4 of 5 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2232)* status: all files reviewed, 9 unresolved discussions (waiting on @Marco-Premier and @SanjayVas)

Marco-Premier (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2232)* status: 15 of 17 files reviewed, 8 unresolved discussions (waiting on @kungfucraig and @SanjayVas)

___
*[`.github/actions/build-and-deploy-cloud-function/action.yml` line 59 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2232#-OPalYr0DpzD-XCjdhiw:-OPeAW7a6fidIZ_xiEaf:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0cc277ce4d36c851e4336cdb0d2027c296425483/.github/actions/build-and-deploy-cloud-function/action.yml#L59)):*

Do the auth and setup steps in the workflow outside of the action, as I described in https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2232#-OPWi9T99Vq52sR3b80H
</blockquote></details>

Done.
___
*[`.github/actions/build-and-deploy-cloud-function/action.yml` line 79 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2232#-OPamS4P-UYCFi6vAK63:-OPe8xkvAOHstgeYBclb:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0cc277ce4d36c851e4336cdb0d2027c296425483/.github/actions/build-and-deploy-cloud-function/action.yml#L79)):*

You shouldn't need this, as `_deploy.jar` targets only have one output.
</blockquote></details>

Done.
___
*[`.github/actions/build-and-deploy-cloud-function/action.yml` line 82 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2232#-OPamsWTDE2h1bQZ0IdJ:-OPeG_8N3Lhicv7w3tpz:b-k5jzq4) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0cc277ce4d36c851e4336cdb0d2027c296425483/.github/actions/build-and-deploy-cloud-function/action.yml#L82)):*

Note that all steps in a job run with the same filesystem, so previous steps may be writing to this same folder if you don't choose a sufficiently unique name. Consider using the `GITHUB_ACTION` environment variable. See https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables#default-environment-variables
</blockquote></details>

I'm passing it now as inputs to the action, I like it more as I can describe what that field is forg
___
*[`.github/workflows/build_and_deploy_cloud_function.yml` line 40 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2232#-OPWhIVP1sYGVF8yplZY:-OPe922Q125CGmWZRIsK:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e7ea24f15bc6e11d0d1bf5cd33f98955447dcd44/.github/workflows/build_and_deploy_cloud_function.yml#L40)):*

You shouldn't need to pass this in. It's discoverable using `bazel cquery`. See https://bazel.build/query/cquery#files-output
</blockquote></details>

Done.

SanjayVas (changes requested):
> Reviewed 2 of 10 files at r2, 1 of 4 files at r3, 4 of 5 files at r4, 3 of 5 files at r5, 2 of 2 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2232)* status: all files reviewed, 9 unresolved discussions (waiting on @kungfucraig and @Marco-Premier)

___
*[`.github/actions/build-and-deploy-cloud-function/action.yml` line 82 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2232#-OPamsWTDE2h1bQZ0IdJ:-OPfiNrgDj0JYuJYZJef:bbhlcni) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0cc277ce4d36c851e4336cdb0d2027c296425483/.github/actions/build-and-deploy-cloud-function/action.yml#L82)):*

I'm passing it now as inputs to the action, I like it more as I can describe what that field is forg
</blockquote></details>

That's leaking an implementation detail of how the action works. The caller of the action shouldn't need to care about the fact that this writes to a temporary directory as part of its implementation. You just need to guarantee that workflow steps don't conflict with each other, which using the `GITHUB_ACTION` environment variable as part of the temp directory path would do.
___
*[`.github/actions/build-and-deploy-cloud-function/action.yml` line 15 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2232#-OPflSwK6uwVMVmzW1J4:-OPflSwK6uwVMVmzW1J5:bswi3du) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/08e189741484ac24a8a64ff3fe132ddaf0b4b4cb/.github/actions/build-and-deploy-cloud-function/action.yml#L15)):*
> ```YAML
> 
> name: "Build and Deploy Cloud Function"
> ```

nit: the product was renamed awhile ago. The overall product is called "Cloud Run", and functions deployed it in are called "Cloud Run functions"



_Suggestion:_
```YAML
Cloud Run function
```
___
*[`.github/actions/build-and-deploy-cloud-function/action.yml` line 16 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2232#-OPfl8v1E_c5B2GEOmA4:-OPfl8v1E_c5B2GEOmA5:b45h) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/08e189741484ac24a8a64ff3fe132ddaf0b4b4cb/.github/actions/build-and-deploy-cloud-function/action.yml#L16)):*
> ```YAML
> 
> name: "Build and Deploy Cloud Function"
> description: "Build Bazel uber-jar and deploy to Google Cloud Functions"
> ```





_Suggestion:_
```YAML
it as a function in Google Cloud Run
```
___
*[`.github/actions/build-and-deploy-cloud-function/action.yml` line 18 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2232#-OPfjxzV3JXDdU_6PuOA:-OPfjxzV3JXDdU_6PuOB:b-xkxfmo) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/08e189741484ac24a8a64ff3fe132ddaf0b4b4cb/.github/actions/build-and-deploy-cloud-function/action.yml#L18)):*
> ```YAML
> description: "Build Bazel uber-jar and deploy to Google Cloud Functions"
> inputs:
>   environment:
> ```

This looks unused?
___
*[`.github/actions/build-and-deploy-cloud-function/action.yml` line 25 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2232#-OPflNe6EO_oKs7cvQNI:-OPflNe6EO_oKs7cvQNJ:b2ni8l0) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/08e189741484ac24a8a64ff3fe132ddaf0b4b4cb/.github/actions/build-and-deploy-cloud-function/action.yml#L25)):*
> ```YAML
>     required: true
>   function_name:
>     description: "GCF function name"
> ```

nit



_Suggestion:_
```YAML
Cloud Run
```
___
*[`.github/actions/build-and-deploy-cloud-function/action.yml` line 40 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2232#-OPfkVHt-qJd2G3eRoTb:-OPfkVHt-qJd2G3eRoTc:b94q2t9) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/08e189741484ac24a8a64ff3fe132ddaf0b4b4cb/.github/actions/build-and-deploy-cloud-function/action.yml#L40)):*
> ```YAML
>     required: true
>   trigger_service_account:
>     description: "Service account to receive the trigger (storage-only)"
> ```

nit: describe this in terms of inputs to the function.



_Suggestion:_
```YAML
. Required if trigger_bucket is specified.
```
___
*[`.github/actions/build-and-deploy-cloud-function/action.yml` line 43 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2232#-OPfkJEdEnfDZqbVtaGF:-OPfkJEdEnfDZqbVtaGG:b-9e9htf) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/08e189741484ac24a8a64ff3fe132ddaf0b4b4cb/.github/actions/build-and-deploy-cloud-function/action.yml#L43)):*
> ```YAML
>     required: false
>   extra_env_vars:
>     description: "Additional --set-env-vars"
> ```

nit: describe this in such a way that the format is clear even if the caller doesn't know anything about the gcloud CLI



_Code quote:_
```YAML
"Additional --set-env-vars"
```
___
*[`.github/actions/build-and-deploy-cloud-function/action.yml` line 61 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2232#-OPfjAys6y1XEFzkZd_8:-OPfjAys6y1XEFzkZd_9:b-mcik71) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/08e189741484ac24a8a64ff3fe132ddaf0b4b4cb/.github/actions/build-and-deploy-cloud-function/action.yml#L61)):*
> ```YAML
>       shell: bash
>       run: |
>         JAR=$(bazel cquery "${{ inputs.bazel_target_label }}" --output=files)
> ```

Map GitHub context variables to the env of any bash step and then use those env vars in the `run` section. This will ensure that it's properly escaped, whereas the approach you're using now would be broken e.g. if there's an unescaped `"` character in an input string.

Marco-Premier (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2232)* status: 15 of 17 files reviewed, 5 unresolved discussions (waiting on @kungfucraig and @SanjayVas)

___
*[`.github/actions/build-and-deploy-cloud-function/action.yml` line 82 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2232#-OPamsWTDE2h1bQZ0IdJ:-OPgCafc30MddwC2eQ0-:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0cc277ce4d36c851e4336cdb0d2027c296425483/.github/actions/build-and-deploy-cloud-function/action.yml#L82)):*

That's leaking an implementation detail of how the action works. The caller of the action shouldn't need to care about the fact that this writes to a temporary directory as part of its implementation. You just need to guarantee that workflow steps don't conflict with each other, which using the `GITHUB_ACTION` environment variable as part of the temp directory path would do.
</blockquote></details>

Done.
___
*[`.github/actions/build-and-deploy-cloud-function/action.yml` line 18 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2232#-OPfjxzV3JXDdU_6PuOA:-OPfvZm0CNGc6R-hI7QM:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/08e189741484ac24a8a64ff3fe132ddaf0b4b4cb/.github/actions/build-and-deploy-cloud-function/action.yml#L18)):*

This looks unused?
</blockquote></details>

Done.
___
*[`.github/actions/build-and-deploy-cloud-function/action.yml` line 61 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2232#-OPfjAys6y1XEFzkZd_8:-OPfxZQkByWvGn76yZ1d:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/08e189741484ac24a8a64ff3fe132ddaf0b4b4cb/.github/actions/build-and-deploy-cloud-function/action.yml#L61)):*

Map GitHub context variables to the env of any bash step and then use those env vars in the `run` section. This will ensure that it's properly escaped, whereas the approach you're using now would be broken e.g. if there's an unescaped `"` character in an input string.
</blockquote></details>

Done.
___
*[`.github/actions/build-and-deploy-cloud-function/action.yml` line 16 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2232#-OPfl8v1E_c5B2GEOmA4:-OPfvPBx3pgcvYklfv2q:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/08e189741484ac24a8a64ff3fe132ddaf0b4b4cb/.github/actions/build-and-deploy-cloud-function/action.yml#L16)):*
> ```YAML
> 
> name: "Build and Deploy Cloud Function"
> description: "Build Bazel uber-jar and deploy to Google Cloud Functions"
> ```

Done.

kungfucraig (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2232)* status: all files reviewed, 4 unresolved discussions (waiting on @SanjayVas)

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2232)* status: all files reviewed, 2 unresolved discussions (waiting on @Marco-Premier)

___
*[`.github/actions/build-and-deploy-cloud-function/action.yml` line 58 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2232#-OPgXnkqEqolnoQOsPaj:-OPgXnkqEqolnoQOsPak:b-ey1dn5) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8f3d5fadc36e644457f9e5933c04bfcdb4aa6c20/.github/actions/build-and-deploy-cloud-function/action.yml#L58)):*
> ```YAML
>       env:
>         BAZEL_TARGET_LABEL: ${{ inputs.bazel_target_label }}
>         GITHUB_ACTION: ${{ github.action }}
> ```

nit: These context variables are automatically in the environment. See https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables#default-environment-variables
___
*[`.github/actions/build-and-deploy-cloud-function/action.yml` line 81 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2232#-OPgY3Ui3MZe5ROksFCI:-OPgY3Ui3MZe5ROksFCJ:b-dyfupo) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8f3d5fadc36e644457f9e5933c04bfcdb4aa6c20/.github/actions/build-and-deploy-cloud-function/action.yml#L81)):*
> ```YAML
>         TRIGGER_SERVICE_ACCOUNT: ${{ inputs.trigger_service_account }}
>         EXTRA_ENV_VARS: ${{ inputs.extra_env_vars }}
>         UBER_JAR_TEMP_DIR: ${{ inputs.uber_jar_temp_dir }}
> ```

This input doesn't exist. It looks like this comes from the environment variable added in the previous step instead.

That said, the better way to do this would be to write a step output rather than polluting the job's environment.

Marco-Premier (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2232)* status: 16 of 17 files reviewed, 1 unresolved discussion (waiting on @kungfucraig and @SanjayVas)

___
*[`.github/actions/build-and-deploy-cloud-function/action.yml` line 81 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2232#-OPgY3Ui3MZe5ROksFCI:-OPjEUHS3WLbFPNEZp7E:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8f3d5fadc36e644457f9e5933c04bfcdb4aa6c20/.github/actions/build-and-deploy-cloud-function/action.yml#L81)):*

This input doesn't exist. It looks like this comes from the environment variable added in the previous step instead.

That said, the better way to do this would be to write a step output rather than polluting the job's environment.
</blockquote></details>

Done.

SanjayVas (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2232)* status: all files reviewed, 1 unresolved discussion (waiting on @kungfucraig)

___
*[`.github/actions/build-and-deploy-cloud-function/action.yml` line 73 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2232#-OPkgZx31nlyJUe9_3LD:-OPkgZx49Pcl6gLFnbvW:bglk0wj) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/db7e1233ebf49e41abe4256a24bb7448f6922120/.github/actions/build-and-deploy-cloud-function/action.yml#L73)):*
> ```YAML
>       shell: bash
>       env:
>         FUNCTION_NAME:            ${{ inputs.function_name }}
> ```

nit: be consistent with existing formatting. Don't align these.

Marco-Premier (commented):
> Reviewed 3 of 11 files at r1, 3 of 10 files at r2, 1 of 4 files at r3, 1 of 5 files at r5, 9 of 9 files at r9, 3 of 3 files at r10, 1 of 1 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2232)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @kungfucraig)

PR #2241, opened 2025-04-30
feat: bind terraform service account to roles/artifactregistry.admin for secure computation artifact registry

Marco-Premier:


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2241)

PR #2171, opened 2025-04-09
feat: Create results fulfiller

jojijac0b:
Creates results fulfiller that will be used by the TEE app to fullfill requisitions. This fulfiller will:
1. Pull in new requisitions from storage
2. Pull in relevant labeled impressions based on collection interval and event group of requisition
3. Filter out labeled impressions based on collection interval, sampling interval, and filter expression in requisition
4. Calculate results using sampled vids(from filtered labeled impressions)
5. Fulfill requistiions to kingdom with result

This PR only handles direct protocol. Other protocols will be added later

PR #2253, opened 2025-05-02
feat: Create utils file for VID calculations that are done in the edp aggregator

jojijac0b:
This pr is a followup to the comment in #2171 to separate out VID calculations into a separate module.


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2253)

SanjayVas (changes requested):
> Reviewed 4 of 7 files at r1.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2253)* status: 4 of 9 files reviewed, 6 unresolved discussions (waiting on @kungfucraig and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt` line 671 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2253#-OPWtsPs3hfcuRroPKUV:-OPWtsPs3hfcuRroPKUW:b-tugv33) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a854dbb4207e03c2904874b363987fa3976a75b6/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt#L671)):*
> ```Kotlin
>                 directProtocolConfig,
>                 try {
>                   VidUtils.selectReachAndFrequencyNoiseMechanism(directNoiseMechanismOptions)
> ```

I don't see this function in the utility object?
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt` line 672 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2253#-OPWtuqJ6J9PxeEIty0C:-OPWtgHp8gIAmfr-K0Aj:b2lz6mk) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a854dbb4207e03c2904874b363987fa3976a75b6/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt#L672)):*
> ```Kotlin
>                 try {
>                   VidUtils.selectReachAndFrequencyNoiseMechanism(directNoiseMechanismOptions)
>                 } catch (e: IllegalArgumentException) {
> ```

Is it true that this is the only condition under which this function might throw an IllegalArgumentException? If so, make sure that's documented on the function. e.g.

```kotlin
/**
 * Selects a noise mechanism for a reach and frequency (R/F) measurement.
 * 
 * @param directNoiseMechanismOptions options for the noise mechanism for a measurement using the direct protocol
 * @throws IlllegalArgumentException if there is no valid noise mechanism indicated in [directNoiseMechanismOptions]
 */
```

If not, then you can't rely on this exception type and need to rely on some other signal (e.g. a custom exception type or return value).
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt` line 674 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2253#-OPWv0jvBSNekIfRu-V9:-OPWv0jvBSNekIfRu-VA:b-p95b1f) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a854dbb4207e03c2904874b363987fa3976a75b6/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt#L674)):*
> ```Kotlin
>                 } catch (e: IllegalArgumentException) {
>                   throw RequisitionRefusalException(
>                     Requisition.Refusal.Justification.SPEC_INVALID,
> ```

Always include the caught exception as the cause when throwing a new exception in response to one that was caught.
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/VidUtils.kt` line 51 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2253#-OPWtLNmAllQ6c2usFTO:-OPWtLNmAllQ6c2usFTP:bsdldb2) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a854dbb4207e03c2904874b363987fa3976a75b6/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/VidUtils.kt#L51)):*
> ```Kotlin
>  * Configuration for storage clients.
>  */
> data class StorageConfig(
> ```

Move this inside of the containing object if it's not used outside of that context. If it is used outside of that context, move it into its own file.



_Code quote:_
```Kotlin
StorageConfig
```
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/VidUtils.kt` line 59 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2253#-OPWnxLiBjmBG2p7qkEa:-OPWnxLiBjmBG2p7qkEb:b-lf30te) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a854dbb4207e03c2904874b363987fa3976a75b6/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/VidUtils.kt#L59)):*
> ```Kotlin
>  * Utility functions for working with VIDs (Virtual IDs) in the EDP Aggregator.
>  */
> object VidUtils {
> ```

We generally avoid "Utils" in the name of utility classes (utility objects in Kotlin) unless they're all operating on a type that's already plural (e.g. a utility class that operates on `Foo` objects might be called `Foos`, but one that operates on `FooStats` objects might be called `FooStatsUtils`). See `java.util.Collections` and `org.wfanet.measurement.common.ProtoReflection` for examples. You would also always include the containing object name at the call-site (e.g. import ProtoReflection and call `ProtoReflection.getTypeUrl` rather than importing `ProtoReflection.getTypeUrl`).

Note for Googlers: this is more clearly described in go/java-practices/utility-classes and go/kotlin/practices/functions

In this case, I think the common thing is that we're dealing with `RequisitonSpec`s.



_Suggestion:_
```Kotlin
RequisitionSpecs
```
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/VidUtils.kt` line 251 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2253#-OPWr9s561sZfL9nM4Xw:-OPWr9s561sZfL9nM4Xx:b-c65uvf) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a854dbb4207e03c2904874b363987fa3976a75b6/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/VidUtils.kt#L251)):*
> ```Kotlin
>    * This is a helper method to extract the descriptor from the TypeRegistry.
>    */
>   private fun getDescriptorForTypeUrl(typeUrl: String, typeRegistry: TypeRegistry): Descriptor {
> ```

TypeRegistry already has a [`getDescriptorForTypeUrl` method](https://protobuf.dev/reference/java/api-docs/com/google/protobuf/TypeRegistry.html#getDescriptorForTypeUrl-java.lang.String-).

kungfucraig (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2253)* status: 4 of 9 files reviewed, 8 unresolved discussions (waiting on @jojijac0b, @SanjayVas, and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/VidUtils.kt` line 59 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2253#-OPWnxLiBjmBG2p7qkEa:-OPXD4qfBxM8QQXmexwc:bnc60px) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a854dbb4207e03c2904874b363987fa3976a75b6/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/VidUtils.kt#L59)):*

We generally avoid "Utils" in the name of utility classes (utility objects in Kotlin) unless they're all operating on a type that's already plural (e.g. a utility class that operates on `Foo` objects might be called `Foos`, but one that operates on `FooStats` objects might be called `FooStatsUtils`). See `java.util.Collections` and `org.wfanet.measurement.common.ProtoReflection` for examples. You would also always include the containing object name at the call-site (e.g. import ProtoReflection and call `ProtoReflection.getTypeUrl` rather than importing `ProtoReflection.getTypeUrl`).

Note for Googlers: this is more clearly described in go/java-practices/utility-classes and go/kotlin/practices/functions

In this case, I think the common thing is that we're dealing with `RequisitonSpec`s.
</blockquote></details>

I agree with Sanjay generally, but would recommend separating what's going on here into several modules. Utils classes are almost always a sign we are not thinking clearly about how to decompose a problem.

The issue here is similar to the issue in the MeasurementResults PR.

Here's a possible decomposition:

1. Start by creating a class called "EventFilter" or "VidFilter."  In it's constructor you should take all of the filter criteria. It should provide a function that takes a flow and returns a flow. Try not to have it depend on the entire Requisition, but just the parts of it that are relevant for the filtering. 

2. "EventReader." Event Reader will take a list of files and 

Then the client can construct an Event Reader and send the flow to the filter which can then be sent to the results creation logic, and sent to a fulfiller, etc.

This should leave us some flexibility with respect to easily replacing certain components in the future (e.g. the reader).

Another option is to pass the filter (as an interface) into the reader (i.e. an application of the strategy pattern). Unless there are performance reasons to prefer the strategy pattern, I prefer the independent pieces approach.
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/VidUtils.kt` line 2 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2253#-OPXAdLv0rKpGDd28eQy:-OPXAdLv0rKpGDd28eQz:bb6ht3a) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a854dbb4207e03c2904874b363987fa3976a75b6/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/VidUtils.kt#L2)):*
> ```Kotlin
> /*
>  * Copyright 2023 The Cross-Media Measurement Authors
> ```

2025
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/VidUtils.kt` line 147 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2253#-OPXCLBZ-Js5cxTiuEOs:-OPXCLBZ-Js5cxTiuEOt:b-pmyeyc) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a854dbb4207e03c2904874b363987fa3976a75b6/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/VidUtils.kt#L147)):*
> ```Kotlin
>       "Invalid vidSamplingIntervalWidth $vidSamplingIntervalWidth"
>     }
>     require(
> ```

You should support wrapping intervals.

jojijac0b (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2253)* status: 3 of 10 files reviewed, 8 unresolved discussions (waiting on @kungfucraig, @SanjayVas, and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/VidUtils.kt` line 2 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2253#-OPXAdLv0rKpGDd28eQy:-OPlF_Fp6ChxqkMxxvDK:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a854dbb4207e03c2904874b363987fa3976a75b6/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/VidUtils.kt#L2)):*

2025
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/VidUtils.kt` line 51 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2253#-OPWtLNmAllQ6c2usFTO:-OPkPyXB3jclfaYpgt-9:b-me31zp) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a854dbb4207e03c2904874b363987fa3976a75b6/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/VidUtils.kt#L51)):*

Move this inside of the containing object if it's not used outside of that context. If it is used outside of that context, move it into its own file.
</blockquote></details>

It will be used by the results fullfiller class in a later pr. Moved out of this file
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/VidUtils.kt` line 251 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2253#-OPWr9s561sZfL9nM4Xw:-OPl6dnTDPSthXVO2Qec:ba3dfob) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a854dbb4207e03c2904874b363987fa3976a75b6/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/VidUtils.kt#L251)):*

TypeRegistry already has a [`getDescriptorForTypeUrl` method](https://protobuf.dev/reference/java/api-docs/com/google/protobuf/TypeRegistry.html#getDescriptorForTypeUrl-java.lang.String-).
</blockquote></details>

Removed
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt` line 671 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2253#-OPWtsPs3hfcuRroPKUV:-OPl6KHJ3VX_7ElqWyiB:bw6iizs) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a854dbb4207e03c2904874b363987fa3976a75b6/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt#L671)):*

I don't see this function in the utility object?
</blockquote></details>

Removing all dependencies in EspSimulator as per @kungfucraig comment in other PR. We will not touch this code

comment: https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2171#-OO5nZFIEJw6P7d8x8kf

SanjayVas (commented):
> Reviewed 3 of 4 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2253)* status: 6 of 10 files reviewed, 6 unresolved discussions (waiting on @jojijac0b, @kungfucraig, and @stevenwarejones)

jojijac0b (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2253)* status: 6 of 10 files reviewed, 6 unresolved discussions (waiting on @kungfucraig, @SanjayVas, and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/VidUtils.kt` line 59 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2253#-OPWnxLiBjmBG2p7qkEa:-OPqTL7xB9ocid_gE4HZ:b6f25ay) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a854dbb4207e03c2904874b363987fa3976a75b6/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/VidUtils.kt#L59)):*

I agree with Sanjay generally, but would recommend separating what's going on here into several modules. Utils classes are almost always a sign we are not thinking clearly about how to decompose a problem.

The issue here is similar to the issue in the MeasurementResults PR.

Here's a possible decomposition:

1. Start by creating a class called "EventFilter" or "VidFilter."  In it's constructor you should take all of the filter criteria. It should provide a function that takes a flow and returns a flow. Try not to have it depend on the entire Requisition, but just the parts of it that are relevant for the filtering. 

2. "EventReader." Event Reader will take a list of files and 

Then the client can construct an Event Reader and send the flow to the filter which can then be sent to the results creation logic, and sent to a fulfiller, etc.

This should leave us some flexibility with respect to easily replacing certain components in the future (e.g. the reader).

Another option is to pass the filter (as an interface) into the reader (i.e. an application of the strategy pattern). Unless there are performance reasons to prefer the strategy pattern, I prefer the independent pieces approach.
</blockquote></details>

@kungfucraig Can you elaborate a little bit on the EventReader? What is the purpose of this class?

kungfucraig (changes requested):
> Reviewed 1 of 4 files at r3, 1 of 9 files at r4, 1 of 2 files at r6, 7 of 10 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2253)* status: 13 of 16 files reviewed, 9 unresolved discussions (waiting on @jojijac0b, @SanjayVas, and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/RequisitionSpecs.kt` line 53 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2253#-OQtuGDN-8_JEcVB81Rc:-OQtuGDN-8_JEcVB81Rd:b-usxebv) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4d4ef362a1f96419314e0895f5c535bdb0f9a8ca/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/RequisitionSpecs.kt#L53)):*
> ```Kotlin
>       "Invalid vidSamplingIntervalWidth $vidSamplingIntervalWidth"
>     }
>     require(
> ```

We should be supporting wrapping intervals (e.g. start = 0.5; width = 0.8).

jojijac0b (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2253)* status: 11 of 16 files reviewed, 9 unresolved discussions (waiting on @kungfucraig, @SanjayVas, and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/EventReader.kt` line 44 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2253#-OQLNKuZ09Fa98R30Mwx:-OQfDx_q24eh1WXV2gDG:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5a6985f29251501cd6ffdf8938d991857332c577/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/EventReader.kt#L44)):*

Maybe call this something different, like "dekStorageConfig" ? It's very unclear what "metadata" means in this context.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/EventReader.kt` line 67 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2253#-OQLOEIN0xj1RKhAK1i2:-OQfDwNs-O-YUGLTd378:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5a6985f29251501cd6ffdf8938d991857332c577/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/EventReader.kt#L67)):*

You should mention that this is the BlobDetails with the dek.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/RequisitionSpecs.kt` line 75 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2253#-OQLMjWl-I7il-pJeZo6:-OQfKriv3J0Nv-UC9Qj_:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5a6985f29251501cd6ffdf8938d991857332c577/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/RequisitionSpecs.kt#L75)):*

Consider passing the EventReader in.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/RequisitionSpecs.kt` line 53 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2253#-OQtuGDN-8_JEcVB81Rc:-OR2VLsd3Z_kZGh-nmnO:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4d4ef362a1f96419314e0895f5c535bdb0f9a8ca/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/RequisitionSpecs.kt#L53)):*

We should be supporting wrapping intervals (e.g. start = 0.5; width = 0.8).
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/VidFilter.kt` line 45 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2253#-OQLL_mP2eRy2UawQ_NG:-OQfPKvPDQVeDiT7Ml5f:bffwaw3) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5a6985f29251501cd6ffdf8938d991857332c577/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/VidFilter.kt#L45)):*

I think you could implement these as two functions a filter function and a mapping function. Then the client code can just use standard map and filter methods to apply these. This allows us to test these filter and map functions on simple inputs and outputs.
</blockquote></details>

@kungfucraig Are you suggesting to split the filterAndExtractVids into 2 separate functions? Or removing the VidFilter class all together?
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/VidFilter.kt` line 48 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2253#-OQLGj_mCXRc3QtdmCvc:-OQfR6OP9E4uQlhfl4I0:bb6un1j) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5a6985f29251501cd6ffdf8938d991857332c577/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/VidFilter.kt#L48)):*

You should do bounds checking on the interval start and end values.
</blockquote></details>

Done
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/VidFilter.kt` line 95 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2253#-OQAg_Cd69NerYAX1px6:-OQfBan07XafrpRPCLiB:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5a6985f29251501cd6ffdf8938d991857332c577/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/VidFilter.kt#L95)):*

won't this be a little faster if you return early once you encounter a false

```
if (!isInSamplingInterval) {
  return false
}
```

</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/VidUtils.kt` line 59 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2253#-OPWnxLiBjmBG2p7qkEa:-OQpYzeb3MVvZIflzIlE:b-i2i6dm) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a854dbb4207e03c2904874b363987fa3976a75b6/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/VidUtils.kt#L59)):*

It looks like you added the other files, should I just review those?
</blockquote></details>

Yes disregard the last comment. EventReader is implemented
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/StorageConfig.kt` line 8 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2253#-OQAgOZOCjVkkrhlJfMC:-OQfMba0ASQPT7y4ve10:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5a6985f29251501cd6ffdf8938d991857332c577/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/StorageConfig.kt#L8)):*

i'd move this to the edpaggregator package as its not specific to the resultsfulfiller
</blockquote></details>

Done.

kungfucraig (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2253)* status: 11 of 16 files reviewed, 9 unresolved discussions (waiting on @jojijac0b, @SanjayVas, and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/RequisitionSpecs.kt` line 53 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2253#-OQtuGDN-8_JEcVB81Rc:-ORDiTDa6mDu7KaCSrrk:b5t8qui) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4d4ef362a1f96419314e0895f5c535bdb0f9a8ca/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/RequisitionSpecs.kt#L53)):*

Done.
</blockquote></details>

In addition width needs to be tested to be <= 1 
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/VidFilter.kt` line 45 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2253#-OQLL_mP2eRy2UawQ_NG:-ORDimW_CA2W2ubIZinU:b-ldc27k) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5a6985f29251501cd6ffdf8938d991857332c577/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/VidFilter.kt#L45)):*

@kungfucraig Are you suggesting to split the filterAndExtractVids into 2 separate functions? Or removing the VidFilter class all together?
</blockquote></details>

There are some "VidFilters" but maybe we don't need the class, just an object that has the filter functions.

I'd like to sort out the issue wrt HMSS and filtering before we go much further though.

jojijac0b (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2253)* status: 11 of 16 files reviewed, 9 unresolved discussions (waiting on @kungfucraig, @SanjayVas, and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/RequisitionSpecs.kt` line 53 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2253#-OQtuGDN-8_JEcVB81Rc:-OREKWzF8BcKn9unCtuU:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4d4ef362a1f96419314e0895f5c535bdb0f9a8ca/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/RequisitionSpecs.kt#L53)):*

In addition width needs to be tested to be <= 1
</blockquote></details>

Done.

jojijac0b (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2253)* status: 11 of 16 files reviewed, 9 unresolved discussions (waiting on @kungfucraig, @SanjayVas, and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/VidFilter.kt` line 52 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2253#-OQLGrY_E83evJZDp9XJ:-ORI8vCR8m-_X2GaKJTc:b-6lfaug) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5a6985f29251501cd6ffdf8938d991857332c577/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/VidFilter.kt#L52)):*

This approach does not work for HMSS. We should discuss this offline. I think we'll probably want to write some additional information into the LabeledImpression rows. @stevenwarejones FYI
</blockquote></details>

@kungfucraig I will use FrequencyVectorBuilder in a subsequent PR when I add the functionality for HMSS protocol. It will be implemented in the same way as EdpSimulator

stevenwarejones (changes requested):
> Reviewed 3 of 10 files at r7, 1 of 1 files at r9, 3 of 7 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2253)* status: 13 of 17 files reviewed, 10 unresolved discussions (waiting on @jojijac0b, @kungfucraig, and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/RequisitionSpecs.kt` line 64 at r11](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2253#-OROEbjq79pVLCNk_MLK:-OROEbjq79pVLCNk_MLL:b-ybvmcj) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/29c1c4e8ffd50103aedbbe233ab95473d123e838/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/RequisitionSpecs.kt#L64)):*
> ```Kotlin
> 
>     // Return a Flow that processes event groups and extracts valid VIDs
>     return requisitionSpec.events.eventGroupsList
> ```

you need something more like
```
    // Return a Flow that processes event groups and extracts valid VIDs
    return requisitionSpec.events.eventGroupsList
      .flatMap { eventGroup ->
        val collectionInterval = eventGroup.value.collectionInterval
        val startDate = LocalDate.ofInstant(collectionInterval.startTime.toInstant(), ZONE_ID)
        val endDate = LocalDate.ofInstant(collectionInterval.endTime.toInstant(), ZONE_ID)
        val dates = startDate.datesUntil(endDate.plusDays(1)).asSequence().toList()
        logger.info("DATES: ${dates.map { it.toString() }.joinToString() }")
        dates.flatMap { date ->
          val blobDetails = getBlobDetails(date.toString(), eventGroup.key)
          val i = getLabeledImpressions(blobDetails).toList()
          logger.info("SIZE1: ${i.size}")
          logger.info(collectionInterval.toString())
          logger.info(vidSamplingIntervalStart.toString())
          logger.info(vidSamplingIntervalWidth.toString())
          val impressions =
            getLabeledImpressions(blobDetails)
              .filter { labeledImpression ->
                isValidImpression(
                  labeledImpression,
                  collectionInterval,
                  eventGroup,
                  vidSamplingIntervalStart,
                  vidSamplingIntervalWidth,
                )
              }
              .map { labeledImpression -> labeledImpression.vid }

          val impressionsList = impressions.toList()
          logger.info("SIZE2: ${impressionsList.size}")
          impressionsList
        }
      }
      .asFlow()
```
```
private suspend fun getBlobDetails(ds: String, eventGroupId: String): BlobDetails {
    val metadataBlobKey = "ds/$ds/event-group-id/$eventGroupId/metadata"
    val metadataBlobUri = "$labeledImpressionMetadataPrefix/$metadataBlobKey"
    val metadataStorageClientUri = SelectedStorageClient.parseBlobUri(metadataBlobUri)
    val impressionsMetadataStorageClient =
      createStorageClient(metadataStorageClientUri, impressionMetadataStorageConfig)
    logger.info("Reading impressions ${metadataStorageClientUri} from $metadataBlobUri")
    // Get EncryptedDek message from storage using the blobKey made up of the ds and eventGroupId
    val metadataBlob =
      checkNotNull(impressionsMetadataStorageClient.getBlob(metadataBlobKey)) {
        "$metadataBlobKey blob cannot be null"
      }
    return BlobDetails.parseFrom(metadataBlob.read().flatten())
  }
```
___
*[`src/test/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/EncryptedMesosStorage.kt` line 32 at r11](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2253#-ORODkznB6sQVJnwW-_-:-ORODkznB6sQVJnwW-_0:bs983e8) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/29c1c4e8ffd50103aedbbe233ab95473d123e838/src/test/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/EncryptedMesosStorage.kt#L32)):*
> ```Kotlin
> 
>   fun createKmsClient(kekUri: String, keyTemplate: String = "AES128_GCM"): KmsClient {
>     val kmsClient = FakeKmsClient()
> ```

anything with fakeKms should be in testing package

kungfucraig (commented):
> Reviewed 1 of 7 files at r11.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2253)* status: 13 of 17 files reviewed, 10 unresolved discussions (waiting on @jojijac0b and @SanjayVas)

jojijac0b (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2253)* status: 12 of 17 files reviewed, 10 unresolved discussions (waiting on @kungfucraig, @SanjayVas, and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/VidFilter.kt` line 45 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2253#-OQLL_mP2eRy2UawQ_NG:-ORTXjFT6tHcDc4NrYG1:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5a6985f29251501cd6ffdf8938d991857332c577/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/VidFilter.kt#L45)):*

There are some "VidFilters" but maybe we don't need the class, just an object that has the filter functions.

I'd like to sort out the issue wrt HMSS and filtering before we go much further though.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/VidFilter.kt` line 60 at r10](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2253#-ORIMQSQ-H7NGacXtb0H:-ORTYV8ZA2lcvYFFqm5D:be33po1) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/dd721e99555d39ada4fe8f0d74eac748179c730a/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/VidFilter.kt#L60)):*

This is redundant with the lines above.

Give the discussion on doing HMSS later, you can have this check that start+width <= 1.0

Wrapping only comes into play with HMSS.
</blockquote></details>

Keeping these checks because I converted VidFilters to an object as previously requested so the initial checks in the init block are no longer done. Added in the extra check of start+width <= 1.0
___
*[`src/test/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/EncryptedMesosStorage.kt` line 32 at r11](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2253#-ORODkznB6sQVJnwW-_-:-ORP7ToUCtctEMds5L4e:b17tf3q) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/29c1c4e8ffd50103aedbbe233ab95473d123e838/src/test/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/EncryptedMesosStorage.kt#L32)):*

anything with fakeKms should be in testing package
</blockquote></details>

This is in test package src/test/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller

kungfucraig (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2253)* status: 12 of 17 files reviewed, 10 unresolved discussions (waiting on @SanjayVas and @stevenwarejones)

kungfucraig (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2253)* status: 10 of 17 files reviewed, 10 unresolved discussions (waiting on @jojijac0b, @SanjayVas, and @stevenwarejones)

___
*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2253#-ORIQJY50Cj628BLE7Do:-ORmQIxc7yNdqw0Jma_z:brdx8v6) (no related file):*

I think we're basically good to go here. Just a few minor details and creating some test utilities.
</blockquote></details>

Do you all have a short description of how you expect this to work overall and how it is going to be deployed. I read the word "server" in the filenames below and get the sense it will run persistently. If this is the case, why not use a cloud function?

kungfucraig (changes requested):
> Reviewed 4 of 7 files at r11, 1 of 2 files at r12, 2 of 3 files at r13, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2253)* status: 16 of 17 files reviewed, 8 unresolved discussions (waiting on @jojijac0b, @SanjayVas, and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/EventReader.kt` line 81 at r13](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2253#-ORmZFSH95G3ydcdC66l:-ORmZFSH95G3ydcdC66m:buii2pl) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/789a16feb82e62292668246c19bf94d89c17df65/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/EventReader.kt#L81)):*
> ```Kotlin
>    * Retrieves blob details for a given collection interval and event group ID.
>    *
>    * @param collectionInterval The time interval for collection
> ```

Given this is operating on a collection interval, should you be returning a BlobDetails for each day entailed in the interval?
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/EventReader.kt` line 101 at r13](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2253#-ORmZ5xh99LZBw73bTQc:-ORmZ5xh99LZBw73bTQd:b-39d1ry) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/789a16feb82e62292668246c19bf94d89c17df65/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/EventReader.kt#L101)):*
> ```Kotlin
>    * Retrieves blob details for a given collection interval and event group ID.
>    *
>    * @param collectionInterval The time interval for collection
> ```

This doesn't take a collection interval.

Also, one of these two functions should call the other one.

stevenwarejones (changes requested):
> Reviewed 1 of 4 files at r3, 1 of 9 files at r4, 1 of 2 files at r6, 1 of 7 files at r11, 1 of 3 files at r13, 4 of 5 files at r14, 3 of 3 files at r15, all commit messages.
> Dismissed @SanjayVas from 3 discussions.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2253)* status: all files reviewed, 5 unresolved discussions (waiting on @kungfucraig and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/VidFilter.kt` line 87 at r15](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2253#-OS4XX097xYte9leXdPF:-OS4XX097xYte9leXdPG:b-m9d764) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/dcaa271fd94c4c39431c777118ea688160416d50/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/VidFilter.kt#L87)):*
> ```Kotlin
>         vidSamplingIntervalStart >= 0 &&
>         vidSamplingIntervalWidth > 0 &&
>         vidSamplingIntervalStart + vidSamplingIntervalWidth <= 1.0
> ```

why are we forbidding wrap-around?

stevenwarejones (commented):
> Dismissed @SanjayVas from a discussion.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2253)* status: all files reviewed, 3 unresolved discussions (waiting on @jojijac0b and @kungfucraig)

kungfucraig (changes requested):
> Reviewed 4 of 5 files at r14, 3 of 3 files at r15, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2253)* status: all files reviewed, 15 unresolved discussions (waiting on @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/VidFilter.kt` line 87 at r15](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2253#-OS4XX097xYte9leXdPF:-OSB9EoO4krGd6g6WWY5:bp06hi8) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/dcaa271fd94c4c39431c777118ea688160416d50/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/VidFilter.kt#L87)):*

why are we forbidding wrap-around?
</blockquote></details>

Only HMSS implements wrapping intervals.
___
*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2253#-OSBBX8K9HIBHPnLlEM8:-OSBBX8K9HIBHPnLlEM9:b-iqgfp) (no related file):*
Please run "bazel coverage" on this PR.
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/RequisitionSpecs.kt` line 77 at r15](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2253#-OSB7B-I2udhSANokknf:-OSB7B-I2udhSANokkng:b-j9mbue) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/dcaa271fd94c4c39431c777118ea688160416d50/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/RequisitionSpecs.kt#L77)):*
> ```Kotlin
>         // Iterates through all dates up to the end date in the collection interval(inclusive)
>         val impressions = dates.flatMapConcat { date ->
>           eventReader.getLabeledImpressionsFlow(date.toString(), eventGroup.key)
> ```

Pass the date all the way down and "toString" it in "getBlobDetails"

It's better to keeper the richer type information for as long as you can.
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/EventReader.kt` line 80 at r15](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2253#-OSB7grc-j8vMhaWvP2v:-OSB7grc-j8vMhaWvP2w:b-w4y2jf) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/dcaa271fd94c4c39431c777118ea688160416d50/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/EventReader.kt#L80)):*
> ```Kotlin
>     )
>     // Get EncryptedDek message from storage using the blobKey made up of the ds and eventGroupId
>     return BlobDetails.parseFrom(impressionsDekStorageClient.getBlob(dekBlobKey)!!.read().flatten())
> ```

Why is casting the null away like this okay? This looks like an NPE in the making to me.
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/EventReader.kt` line 109 at r15](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2253#-OSB86n-CKgezoa1fSUP:-OSB86n-CKgezoa1fSUQ:b-vwbtde) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/dcaa271fd94c4c39431c777118ea688160416d50/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/EventReader.kt#L109)):*
> ```Kotlin
>     val impressionsMesosStorage = MesosRecordIoStorageClient(impressionsAeadStorageClient)
>     val impressionBlob = impressionsMesosStorage.getBlob(storageClientUri.key)
>       ?: throw IllegalStateException("Could not retrieve impression blob from ${storageClientUri.key}")
> ```

This is not the exception type you're looking for. https://docs.oracle.com/javase/8/docs/api/java/lang/IllegalStateException.html

You could use a RuntimeException if you want. Creating an "ImpressionReadException" would be better.
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/EventReader.kt` line 114 at r15](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2253#-OSB8Oyc4DejIJIn1FQE:-OSB8Oyc4DejIJIn1FQF:b2nreh6) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/dcaa271fd94c4c39431c777118ea688160416d50/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/EventReader.kt#L114)):*
> ```Kotlin
>     return impressionBlob.read().map { impressionByteString ->
>       LabeledImpression.parseFrom(impressionByteString)
>         ?: throw IllegalStateException("Failed to parse LabeledImpression from bytes")
> ```

Same as above.
___
*[`src/test/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/EncryptedMesosStorage.kt` line 1 at r15](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2253#-OSB9uzx66x-bPjHa5qu:-OSB9uzyD-WF6_KdjIq0:bkzpkee) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/dcaa271fd94c4c39431c777118ea688160416d50/src/test/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/EncryptedMesosStorage.kt#L1)):*
> ```Kotlin
> package org.wfanet.measurement.edpaggregator.resultsfulfiller
> ```

@stevenwarejones If I recall you have a similar set of utils. It'd be nice to see them merged with this and put in a more common location.

It also occurs to me that some of these are useful outside of tests (e.g. createEncryptedMesosStorage)
___
*[`src/test/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/EncryptedMesosStorage.kt` line 22 at r15](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2253#-OSBAbRg9vk2-wAMyXP_:-OSBAbRg9vk2-wAMyXPa:bssz719) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/dcaa271fd94c4c39431c777118ea688160416d50/src/test/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/EncryptedMesosStorage.kt#L22)):*
> ```Kotlin
> 
> object EncryptedMesosStorage {
>   fun createEncryptedMesosStorage(
> ```

create...StorageClient

It doesn't lo
___
*[`src/test/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/EncryptedMesosStorage.kt` line 27 at r15](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2253#-OSBAkHT5JPlhhHKF4fj:-OSBAkHUAGHPXcSOGMYG:b-d05kov) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/dcaa271fd94c4c39431c777118ea688160416d50/src/test/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/EncryptedMesosStorage.kt#L27)):*
> ```Kotlin
>     kekUri: String,
>     serializedEncryptionKey: ByteString,
>     tinkKeyTemplateType: String = "AES128_GCM_HKDF_1MB"
> ```

It doesn't look like this ever gets passed in (same with functions below). May as well make it a constant in the object.
___
*[`src/test/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/EncryptedMesosStorage.kt` line 35 at r15](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2253#-OSBA9Xr014jclrDgGG7:-OSBA9Xr014jclrDgGG8:b-ppn862) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/dcaa271fd94c4c39431c777118ea688160416d50/src/test/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/EncryptedMesosStorage.kt#L35)):*
> ```Kotlin
>   }
> 
>   fun createKmsClient(kekUri: String, keyTemplate: String = "AES128_GCM"): KmsClient {
> ```

createFakeKmsClient
___
*[`src/test/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/EncryptedMesosStorage.kt` line 74 at r15](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2253#-OSBB4nA6FPu6vF_rE2J:-OSBB4nA6FPu6vF_rE2K:bnu9oaf) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/dcaa271fd94c4c39431c777118ea688160416d50/src/test/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/EncryptedMesosStorage.kt#L74)):*
> ```Kotlin
>   }
> 
>   suspend fun uploadImpressions(
> ```

s/upload/write (here and below)
___
*[`src/test/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/RequisitionSpecsTest.kt` line 58 at r15](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2253#-OSBDYkp9tKzWQhXCcWY:-OSBDYkq8sRuuWfX6a_F:b4h8tyq) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/dcaa271fd94c4c39431c777118ea688160416d50/src/test/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/RequisitionSpecsTest.kt#L58)):*
> ```Kotlin
> import org.wfanet.measurement.storage.filesystem.FileSystemStorageClient
> 
> @RunWith(JUnit4::class)
> ```

Error condition tests on the VID interval would be nice to see.
___
*[`src/test/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/RequisitionSpecsTest.kt` line 183 at r15](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2253#-OSBDmp--8E1VCuOjhYE:-OSBDmp--8E1VCuOjhYF:budr71u) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/dcaa271fd94c4c39431c777118ea688160416d50/src/test/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/RequisitionSpecsTest.kt#L183)):*
> ```Kotlin
>     )
> 
>     assertThat(result.count()).isEqualTo(validImpressionCount1 + validImpressionCount2)
> ```

It would be nice to see another test that writes both sets of valid impressions, but has a requisition that only requires one of them.

Arguably you could change this test to have that behavior and you would not lose anything.
___
*[`src/test/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/EventReaderTest.kt` line 89 at r15](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2253#-OSBCV426OLxtSqjeQHP:-OSBCV426OLxtSqjeQHQ:b-mko6b9) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/dcaa271fd94c4c39431c777118ea688160416d50/src/test/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/EventReaderTest.kt#L89)):*
> ```Kotlin
> 
>   @Test
>   fun `getLabeledImpressionsFlow returns labeled impressions`() = runBlocking {
> ```

Seems like this test could use some of the EncryptedMesos util functions you added. In particular upload/writeImpressions.
___
*[`src/test/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/EventReaderTest.kt` line 89 at r15](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2253#-OSBEHl8Dv0n2gdTEYGV:-OSBEHl8Dv0n2gdTEYGW:b-balhmb) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/dcaa271fd94c4c39431c777118ea688160416d50/src/test/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/EventReaderTest.kt#L89)):*
> ```Kotlin
> 
>   @Test
>   fun `getLabeledImpressionsFlow returns labeled impressions`() = runBlocking {
> ```

Add a test that tries to read from a DS that doesn't exist in the storage.

stevenwarejones (dismissed):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2253)* status: 8 of 23 files reviewed, 14 unresolved discussions (waiting on @jojijac0b, @kungfucraig, and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/EventReader.kt` line 80 at r15](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2253#-OSB7grc-j8vMhaWvP2v:-OSM18cP1V-VRcsSmblT:b96vtfd) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/dcaa271fd94c4c39431c777118ea688160416d50/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/EventReader.kt#L80)):*

Why is casting the null away like this okay? This looks like an NPE in the making to me.
</blockquote></details>

this is going to throw an exception no matter what and should. Changed to more specific exception
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/EventReader.kt` line 114 at r15](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2253#-OSB8Oyc4DejIJIn1FQE:-OSMFom-0O2LDHprmR9T:bb7f2mf) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/dcaa271fd94c4c39431c777118ea688160416d50/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/EventReader.kt#L114)):*

Same as above.
</blockquote></details>

done
___
*[`src/test/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/EncryptedMesosStorage.kt` line 27 at r15](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2253#-OSBAkHT5JPlhhHKF4fj:-OSMF9FTEj88CcM3da-d:b-rewkd7) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/dcaa271fd94c4c39431c777118ea688160416d50/src/test/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/EncryptedMesosStorage.kt#L27)):*

It doesn't look like this ever gets passed in (same with functions below). May as well make it a constant in the object.
</blockquote></details>

i took away the default/deleted some of these params when unused
___
*[`src/test/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/EncryptedMesosStorage.kt` line 35 at r15](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2253#-OSBA9Xr014jclrDgGG7:-OSMF6ysEznkCXqR6BIO:bb7f2mf) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/dcaa271fd94c4c39431c777118ea688160416d50/src/test/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/EncryptedMesosStorage.kt#L35)):*

createFakeKmsClient
</blockquote></details>

done
___
*[`src/test/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/EncryptedMesosStorage.kt` line 74 at r15](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2253#-OSBB4nA6FPu6vF_rE2J:-OSMFCpO72gEdTVudEEP:b-u89r8m) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/dcaa271fd94c4c39431c777118ea688160416d50/src/test/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/EncryptedMesosStorage.kt#L74)):*

s/upload/write (here and below)
</blockquote></details>

moved this to the test itself where I think it better fits

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2253)* status: 8 of 23 files reviewed, 14 unresolved discussions (waiting on @jojijac0b, @kungfucraig, and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/RequisitionSpecs.kt` line 77 at r15](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2253#-OSB7B-I2udhSANokknf:-OSMFqgS0-gg2EE_F3U3:bb7f2mf) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/dcaa271fd94c4c39431c777118ea688160416d50/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/RequisitionSpecs.kt#L77)):*

Pass the date all the way down and "toString" it in "getBlobDetails"

It's better to keeper the richer type information for as long as you can.
</blockquote></details>

done

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2253)* status: 8 of 23 files reviewed, 13 unresolved discussions (waiting on @jojijac0b, @kungfucraig, and @SanjayVas)

___
*[`src/test/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/EncryptedMesosStorage.kt` line 1 at r15](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2253#-OSB9uzx66x-bPjHa5qu:-OSMG0dY1n7ISmJOI0pb:bb7f2mf) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/dcaa271fd94c4c39431c777118ea688160416d50/src/test/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/EncryptedMesosStorage.kt#L1)):*

@stevenwarejones If I recall you have a similar set of utils. It'd be nice to see them merged with this and put in a more common location.

It also occurs to me that some of these are useful outside of tests (e.g. createEncryptedMesosStorage)
</blockquote></details>

done
___
*[`src/test/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/EncryptedMesosStorage.kt` line 22 at r15](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2253#-OSBAbRg9vk2-wAMyXP_:-OSMG1Lc6ASNvqEvqtR4:bb7f2mf) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/dcaa271fd94c4c39431c777118ea688160416d50/src/test/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/EncryptedMesosStorage.kt#L22)):*

create...StorageClient

It doesn't lo
</blockquote></details>

done

kungfucraig (changes requested):
> Reviewed 9 of 15 files at r16, 4 of 6 files at r17, 1 of 1 files at r18, 2 of 2 files at r20, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2253)* status: all files reviewed, 7 unresolved discussions (waiting on @jojijac0b and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/EncryptedStorage.kt` line 30 at r20](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2253#-OSQj8hYD-9r2PEPa9fU:-OSQiWlp1iCdedVky-k1:bk0w184) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c73acf57bb07e6872e3ac37d20177d6498224281/src/main/kotlin/org/wfanet/measurement/edpaggregator/EncryptedStorage.kt#L30)):*
> ```Kotlin
> 
> /** Useful functions for interacting with encrypted storage. */
> object EncryptedStorage {
> ```

Add kdoc on the functions.
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/EncryptedStorage.kt` line 31 at r20](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2253#-OSQifDfEmCmtueuQ-BW:-OSQifDfEmCmtueuQ-BX:bpqms12) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c73acf57bb07e6872e3ac37d20177d6498224281/src/main/kotlin/org/wfanet/measurement/edpaggregator/EncryptedStorage.kt#L31)):*
> ```Kotlin
> /** Useful functions for interacting with encrypted storage. */
> object EncryptedStorage {
>   fun getSerializedEncryptionKey(
> ```

"get" is not the right verb here. Maybe "generate" ?
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/EncryptedStorage.kt` line 42 at r20](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2253#-OSQixjQBhgTdsaam6Ew:-OSQixjQBhgTdsaam6Ex:b-6v7s0s) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c73acf57bb07e6872e3ac37d20177d6498224281/src/main/kotlin/org/wfanet/measurement/edpaggregator/EncryptedStorage.kt#L42)):*
> ```Kotlin
>         keyEncryptionHandle,
>         kmsClient.getAead(kekUri),
>         byteArrayOf(),
> ```

use a keyword argument for this
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/ImpressionReadException.kt` line 17 at r20](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2253#-OSRFrSU6hSyV_K7PgcL:-OSRFrSU6hSyV_K7PgcM:bb7jbea) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c73acf57bb07e6872e3ac37d20177d6498224281/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/ImpressionReadException.kt#L17)):*
> ```Kotlin
> package org.wfanet.measurement.edpaggregator.resultsfulfiller
> 
> class ImpressionReadException(val blobKey: String, val code: Code, message: String? = null) :
> ```

kdoc

stevenwarejones (commented):
> Reviewed 7 of 15 files at r16, 3 of 6 files at r17, 1 of 2 files at r20, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2253)* status: all files reviewed, 7 unresolved discussions (waiting on @jojijac0b and @kungfucraig)

___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/EncryptedStorage.kt` line 30 at r20](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2253#-OSQj8hYD-9r2PEPa9fU:-OSRKlzk8vH_Xjetct3M:bb7f2mf) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c73acf57bb07e6872e3ac37d20177d6498224281/src/main/kotlin/org/wfanet/measurement/edpaggregator/EncryptedStorage.kt#L30)):*

Add kdoc on the functions.
</blockquote></details>

done
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/EncryptedStorage.kt` line 31 at r20](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2253#-OSQifDfEmCmtueuQ-BW:-OSRKlQ0A-AOlUjuawIi:bb7f2mf) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c73acf57bb07e6872e3ac37d20177d6498224281/src/main/kotlin/org/wfanet/measurement/edpaggregator/EncryptedStorage.kt#L31)):*

"get" is not the right verb here. Maybe "generate" ?
</blockquote></details>

done
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/EncryptedStorage.kt` line 42 at r20](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2253#-OSQixjQBhgTdsaam6Ew:-OSRJpy3EKIgAo80VJ-f:b-igly8u) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c73acf57bb07e6872e3ac37d20177d6498224281/src/main/kotlin/org/wfanet/measurement/edpaggregator/EncryptedStorage.kt#L42)):*

use a keyword argument for this
</blockquote></details>

this comes from a java library. I don't believe keyword arguments are possible here. made this a variable to make it more clear what it is.
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/ImpressionReadException.kt` line 17 at r20](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2253#-OSRFrSU6hSyV_K7PgcL:-OSRKwYa4dN_QBpUgmy_:bb7f2mf) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c73acf57bb07e6872e3ac37d20177d6498224281/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/ImpressionReadException.kt#L17)):*

kdoc
</blockquote></details>

done

kungfucraig (commented):
> Reviewed 7 of 7 files at r21, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2253)* status: all files reviewed, 3 unresolved discussions (waiting on @jojijac0b and @stevenwarejones)

kungfucraig (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2253)* status: all files reviewed, 2 unresolved discussions (waiting on @stevenwarejones)

___
*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2253#-OSBBX8K9HIBHPnLlEM8:-OSRPsk_DN3hlol1Bppc:b6k96oj) (no related file):*

Please run "bazel coverage" on this PR.
</blockquote></details>

Please run coverage and make sure you're covering the core logic of the implementation. Once that's done feel free to merge without further review.

stevenwarejones (commented):
> Reviewed 2 of 15 files at r16, 1 of 1 files at r18, 5 of 7 files at r21, 4 of 4 files at r22, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2253)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @jojijac0b)

PR #2258, opened 2025-05-05
feat: Create modules to build measurement results and fulfill requisitions

jojijac0b:
Adds modules that contain functionality to create a measurement result and fulfill a requisition. These modules will be used together in a later PR that creates the results fulfiller


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2258)

SanjayVas (changes requested):
> Reviewed 1 of 8 files at r1, 1 of 4 files at r3.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2258)* status: 2 of 8 files reviewed, 4 unresolved discussions (waiting on @kungfucraig and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/MeasurementHelper.kt` line 39 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2258#-OPWxdFp01C9JbHqhqc0:-OPWxdFp01C9JbHqhqc1:b-fqmuya) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6e9c960109a1a7b82f088a34fc783aa1b9c5c18b/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/MeasurementHelper.kt#L39)):*
> ```Kotlin
>  * Helper functions for measurement operations.
>  */
> open class MeasurementHelper(
> ```

Why is this open? What specialized subclasses do we expect?



_Code quote:_
```Kotlin
open
```
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/MeasurementHelper.kt` line 39 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2258#-OPWxdqKA5_z94z4cu-E:-OPWxdqKA5_z94z4cu-F:b-qeemx) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6e9c960109a1a7b82f088a34fc783aa1b9c5c18b/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/MeasurementHelper.kt#L39)):*
> ```Kotlin
>  * Helper functions for measurement operations.
>  */
> open class MeasurementHelper(
> ```

Some of this looks suspiciously like code we have in the EDP simulator (requisition fulfiller), implying it was copied from there. Can we extract the common code?
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/MeasurementHelper.kt` line 288 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2258#-OPWwVPfEQEbzFNcPbgX:-OPWwVPfEQEbzFNcPbgY:b-8gr5b2) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6e9c960109a1a7b82f088a34fc783aa1b9c5c18b/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/MeasurementHelper.kt#L288)):*
> ```Kotlin
>   }
> 
>   protected open class RequisitionRefusalException(
> ```

I think you want this in its own file. This is not a library for general usage, but rather specifically for the EDPA. As a result, all subclasses should be known. Therefore if this has subclasses, they should be defined as part of a `sealed` class hierarchy rather than having the class being `open`.
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/MeasurementResults.kt` line 23 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2258#-OPWxEI68pjkqFHtds40:-OPWxEI68pjkqFHtds41:bnxxd7z) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6e9c960109a1a7b82f088a34fc783aa1b9c5c18b/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/MeasurementResults.kt#L23)):*
> ```Kotlin
> 
> /** Utilities for computing Measurement results. */
> object MeasurementResults {
> ```

This looks suspiciously like https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/main/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/MeasurementResults.kt. Can it be extracted to a common location?

kungfucraig (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2258)* status: 3 of 8 files reviewed, 4 unresolved discussions (waiting on @jojijac0b and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/MeasurementHelper.kt` line 39 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2258#-OPWxdqKA5_z94z4cu-E:-OPX37iC8IHaU3rNP2FW:b8eyvd9) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6e9c960109a1a7b82f088a34fc783aa1b9c5c18b/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/MeasurementHelper.kt#L39)):*

Some of this looks suspiciously like code we have in the EDP simulator (requisition fulfiller), implying it was copied from there. Can we extract the common code?
</blockquote></details>

If I'm not mistaken, I think we wanted to have a separate TEE App for building the result and another for the communication with the Kingdom. The reason for this is that we think there will need to be some optimization that occurs for creating results whereas for the actual fulfillment there will not be.

 @stevenwarejones can you confirm this?

We should also be thinking of using some EDP Aggregator specific structures to represent a fulfillment/requisition vs. using the Kingdom structures directly (e.g. Requisition). They could wrap the Kingdom structures, however.

For example, we are likely to want to include both the noisy and original value for measurements in the fulfillment payload. The point is that having an independent structure allows the EDP Aggregator to add additional information. 

For this PR I suggest creating the following modules:

1. A module for each of computing reach, reach and frequency, impressions that includes the sampling and noising aspects, and which expose methods for computing noisy and non-noisy values. For sake of correctness we need to be able to see how a metric is computed end to end and to test it holistically. Currently it's spread across at least two places.  

They could all implement:

interface MeaurementResultBuilder {
  buildMeasurementResult(): MeasurementResult
  buildNoisyMeasurementResult(): MeasurementResult
}

The derived class constructors would need to take whatever params are otherwise necessary to build the result. For example ReachResultBuilder might look something like this:

class ReachResultBuilder(float<VID> sampledVids, float samplingRate, AbstractNoiser n) {
}


2. On top of this create a MeasurementResultFactory that behaves similarly to "buildDirectMeasurementResult." It's fine to have a swtich statement inside of a MeasurementResltFactory.buildMeaurementResult() method.  MeasurementResultFactory is probably just an object with a single function.

3. A separate module for dealing with the construction of a noiser using principles similar to 1 and 2 above. We need to file an issue to stop using the Halo built-in noisers and use a proper DP library and this must be fixed before we can release the EDP Aggregator. Please leave a TODO for this.

4. A separate module that performs a direct fuilfillment similar to what "fulfillDirectMeasurement" does, but ideally you would not take in the entire Requisition and MeasurementSpec, but just those parts that are truly required.

I'm happy to discuss this further on call if you'd like. I appreciate there are a lot of ways to decompose problems like this and that the way outlined above could be improved upon.

The main thing is to adhere to the "separation of concerns" principle. Smaller single purpose modules helps to achieve this.

kungfucraig (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2258)* status: 3 of 8 files reviewed, 4 unresolved discussions (waiting on @jojijac0b and @stevenwarejones)

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2258)* status: 0 of 16 files reviewed, 4 unresolved discussions (waiting on @jojijac0b, @kungfucraig, and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/MeasurementHelper.kt` line 39 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2258#-OPWxdqKA5_z94z4cu-E:-OQAebRg3umoSLCFwh9c:bq2oyvu) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6e9c960109a1a7b82f088a34fc783aa1b9c5c18b/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/MeasurementHelper.kt#L39)):*

If I'm not mistaken, I think we wanted to have a separate TEE App for building the result and another for the communication with the Kingdom. The reason for this is that we think there will need to be some optimization that occurs for creating results whereas for the actual fulfillment there will not be.

 @stevenwarejones can you confirm this?

We should also be thinking of using some EDP Aggregator specific structures to represent a fulfillment/requisition vs. using the Kingdom structures directly (e.g. Requisition). They could wrap the Kingdom structures, however.

For example, we are likely to want to include both the noisy and original value for measurements in the fulfillment payload. The point is that having an independent structure allows the EDP Aggregator to add additional information. 

For this PR I suggest creating the following modules:

1. A module for each of computing reach, reach and frequency, impressions that includes the sampling and noising aspects, and which expose methods for computing noisy and non-noisy values. For sake of correctness we need to be able to see how a metric is computed end to end and to test it holistically. Currently it's spread across at least two places.  

They could all implement:

interface MeaurementResultBuilder {
  buildMeasurementResult(): MeasurementResult
  buildNoisyMeasurementResult(): MeasurementResult
}

The derived class constructors would need to take whatever params are otherwise necessary to build the result. For example ReachResultBuilder might look something like this:

class ReachResultBuilder(float<VID> sampledVids, float samplingRate, AbstractNoiser n) {
}


2. On top of this create a MeasurementResultFactory that behaves similarly to "buildDirectMeasurementResult." It's fine to have a swtich statement inside of a MeasurementResltFactory.buildMeaurementResult() method.  MeasurementResultFactory is probably just an object with a single function.

3. A separate module for dealing with the construction of a noiser using principles similar to 1 and 2 above. We need to file an issue to stop using the Halo built-in noisers and use a proper DP library and this must be fixed before we can release the EDP Aggregator. Please leave a TODO for this.

4. A separate module that performs a direct fuilfillment similar to what "fulfillDirectMeasurement" does, but ideally you would not take in the entire Requisition and MeasurementSpec, but just those parts that are truly required.

I'm happy to discuss this further on call if you'd like. I appreciate there are a lot of ways to decompose problems like this and that the way outlined above could be improved upon.

The main thing is to adhere to the "separation of concerns" principle. Smaller single purpose modules helps to achieve this.
</blockquote></details>

I prefer a single app for result computation and fulfillment. Although I think all the separate modules are good ideas.

jojijac0b (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2258)* status: 0 of 40 files reviewed, 4 unresolved discussions (waiting on @kungfucraig, @SanjayVas, and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/MeasurementHelper.kt` line 39 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2258#-OPWxdFp01C9JbHqhqc0:-OQ9QYQG-uRA0RFDkGAQ:bo90blb) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6e9c960109a1a7b82f088a34fc783aa1b9c5c18b/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/MeasurementHelper.kt#L39)):*

Why is this open? What specialized subclasses do we expect?
</blockquote></details>

Removed open
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/MeasurementHelper.kt` line 39 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2258#-OPWxdqKA5_z94z4cu-E:-OQF_Rn7Emt_Sz9h3HUL:b-hgbycn) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6e9c960109a1a7b82f088a34fc783aa1b9c5c18b/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/MeasurementHelper.kt#L39)):*

I prefer a single app for result computation and fulfillment. Although I think all the separate modules are good ideas.
</blockquote></details>

@kungfucraig I've restructured this PR to use the patterns you mentioned. See details below.

Measurement builders
- Created a new package within resultsfulfiller called `compute` that contains the MeasurementResultBuilder interface.
- Within the `compute` package is another package called `protocols` that will contain packages for building measurements based on the protocol(i.e. direct, hmss, etc.)
- Within the `protocols` package is a package called `direct` which contains the implementation of MeasurementResultBuilder for all direct measurements. In this package are the files DirectMeasurementResultFactory, which contains the logic from buildDirectMeasurementResult, and DirectReachAndFrequencyResultBuilder which implements the MeasurementResultBuilder interface. Only direct measurements benefit from the factory pattern. For the other protocols, we can directly call their implementations of the MeasurementResultBuilder interface from the ResultsFulfiller.
- These MeasurementResultBuilders sole responsibility is to create a Measurement.Result to be used for fulfillment

Measurement fulfillment
- Created a package within resultsfulfiller called `fulfillers` that will contain the logic to fulfill requisitions based on the protocol. 
- Within this package is an interface, MeasurementFulfiller, that contains a fulfillRequisition and a requisitionsStub variable. This interface must be implemented by all protocols. I have created one for direct measurements called DirectMeasurementFulfiller. There will be 2 more called HmssMeasurementFulfiller and Llv2MeasurementFulfiller in later PRs.
- These Measurement fulfillers sole responsibility is to fulfill requisitions to the kingdom


I will stitch all of this logic in my ResultsFullfiller class in another PR(#2171). This fulfiller will:
1. Get the filtered, sampled, VIDs using the VidFilter in #2253 
2. Calculate the result using MeasurementResults, and create the Measurement.Result using  resultsfulfiller/compute/protocols module
3. Fulfill the measurement using resultsfulfiller/fulfillers module
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/MeasurementHelper.kt` line 288 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2258#-OPWwVPfEQEbzFNcPbgX:-OQ9OTrOEb5ENVnFeqF6:b35qvcj) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6e9c960109a1a7b82f088a34fc783aa1b9c5c18b/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/MeasurementHelper.kt#L288)):*

I think you want this in its own file. This is not a library for general usage, but rather specifically for the EDPA. As a result, all subclasses should be known. Therefore if this has subclasses, they should be defined as part of a `sealed` class hierarchy rather than having the class being `open`.
</blockquote></details>

Moved RequisitionRefusalException to new class. It is reused by RequisitionFulfiller
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/MeasurementResults.kt` line 23 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2258#-OPWxEI68pjkqFHtds40:-OQ9Sfxc-8raIGxkfj6c:b4sq8og) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6e9c960109a1a7b82f088a34fc783aa1b9c5c18b/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/MeasurementResults.kt#L23)):*

This looks suspiciously like https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/main/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/MeasurementResults.kt. Can it be extracted to a common location?
</blockquote></details>

Consolidated both implementations of MeasurementResults into a single file

kungfucraig (changes requested):
> Reviewed 2 of 7 files at r4, 3 of 11 files at r5, 13 of 34 files at r6, 19 of 20 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2258)* status: 37 of 40 files reviewed, 14 unresolved discussions (waiting on @SanjayVas and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/MeasurementHelper.kt` line 39 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2258#-OPWxdqKA5_z94z4cu-E:-OQKfnlg4a-vBFZVYYwy:b-dtw4t2) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6e9c960109a1a7b82f088a34fc783aa1b9c5c18b/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/MeasurementHelper.kt#L39)):*

@kungfucraig I've restructured this PR to use the patterns you mentioned. See details below.

Measurement builders
- Created a new package within resultsfulfiller called `compute` that contains the MeasurementResultBuilder interface.
- Within the `compute` package is another package called `protocols` that will contain packages for building measurements based on the protocol(i.e. direct, hmss, etc.)
- Within the `protocols` package is a package called `direct` which contains the implementation of MeasurementResultBuilder for all direct measurements. In this package are the files DirectMeasurementResultFactory, which contains the logic from buildDirectMeasurementResult, and DirectReachAndFrequencyResultBuilder which implements the MeasurementResultBuilder interface. Only direct measurements benefit from the factory pattern. For the other protocols, we can directly call their implementations of the MeasurementResultBuilder interface from the ResultsFulfiller.
- These MeasurementResultBuilders sole responsibility is to create a Measurement.Result to be used for fulfillment

Measurement fulfillment
- Created a package within resultsfulfiller called `fulfillers` that will contain the logic to fulfill requisitions based on the protocol. 
- Within this package is an interface, MeasurementFulfiller, that contains a fulfillRequisition and a requisitionsStub variable. This interface must be implemented by all protocols. I have created one for direct measurements called DirectMeasurementFulfiller. There will be 2 more called HmssMeasurementFulfiller and Llv2MeasurementFulfiller in later PRs.
- These Measurement fulfillers sole responsibility is to fulfill requisitions to the kingdom


I will stitch all of this logic in my ResultsFullfiller class in another PR(#2171). This fulfiller will:
1. Get the filtered, sampled, VIDs using the VidFilter in #2253 
2. Calculate the result using MeasurementResults, and create the Measurement.Result using  resultsfulfiller/compute/protocols module
3. Fulfill the measurement using resultsfulfiller/fulfillers module
</blockquote></details>

Thank's @jojijac0b !
___
*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2258#-OQKQrih-G_Y0TLxo6di:-OQKQrih-G_Y0TLxo6dj:b-c5pz9o) (no related file):*
Please update your PR title and description
___
*[`src/test/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/fulfillers/DirectMeasurementFulfillerTest.kt` line 90 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2258#-OQKfdt0C-poUsN-zBgV:-OQKfdt0C-poUsN-zBgW:b-io9dwf) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/564a9adaa3a162a1dacff6229f474d3c190fd534/src/test/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/fulfillers/DirectMeasurementFulfillerTest.kt#L90)):*
> ```Kotlin
> 
>     // Verify the stub was called with the correct parameters
>     verifyProtoArgument(requisitionsServiceMock, RequisitionsCoroutineImplBase::fulfillDirectRequisition)
> ```

You should check the value here.
___
*[`src/main/kotlin/org/wfanet/measurement/dataprovider/MeasurementResults.kt` line 64 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2258#-OQKR5tzE73huoK6-mu8:-OQKR5tzE73huoK6-mu9:b-fdbvtk) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/564a9adaa3a162a1dacff6229f474d3c190fd534/src/main/kotlin/org/wfanet/measurement/dataprovider/MeasurementResults.kt#L64)):*
> ```Kotlin
>    * "deterministic distribution" methodology.
>    */
>   fun computeReachAndFrequency(filteredVids: Iterable<Long>, maxFrequency: Int): ReachAndFrequency {
> ```

You should be able to implement this in terms of the other one by calling "filteredVids.asFlow"
___
*[`src/main/kotlin/org/wfanet/measurement/dataprovider/MeasurementResults.kt` line 122 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2258#-OQKRh-y0HSw4p0j5nUD:-OQKRh-y0HSw4p0j5nUE:b-seapg8) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/564a9adaa3a162a1dacff6229f474d3c190fd534/src/main/kotlin/org/wfanet/measurement/dataprovider/MeasurementResults.kt#L122)):*
> ```Kotlin
> 
>   /** Computes population using the "deterministic count" methodology. */
>   fun computePopulation(
> ```

The EDP Aggregator should not fulfill requisitions for this type of metric. Only the PopulationDataProvider should do so.
___
*[`src/main/kotlin/org/wfanet/measurement/dataprovider/RequisitionRefusalException.kt` line 46 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2258#-OQKSG3x0maEgnMWgS4H:-OQKSG3x0maEgnMWgS4I:b-4twy7x) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/564a9adaa3a162a1dacff6229f474d3c190fd534/src/main/kotlin/org/wfanet/measurement/dataprovider/RequisitionRefusalException.kt#L46)):*
> ```Kotlin
>    * [RequisitionRefusalException] for test EventGroups.
>    */
>   class Test(
> ```

Why do we need the two different versions of this?
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/compute/protocols/direct/DirectReachAndFrequencyResultBuilder.kt` line 57 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2258#-OQKTBFMBy6Kiv4v0qoe:-OQKTBFN5ZoeMCj1uAGy:b-cy2biw) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/564a9adaa3a162a1dacff6229f474d3c190fd534/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/compute/protocols/direct/DirectReachAndFrequencyResultBuilder.kt#L57)):*
> ```Kotlin
>   private val frequencyPrivacyParams: DifferentialPrivacyParams,
>   private val samplingRate: Float,
>   private val directNoiseMechanism: DirectNoiseMechanism = DirectNoiseMechanism.CONTINUOUS_GAUSSIAN,
> ```

Can you remove the defaults? 
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/compute/protocols/direct/DirectReachAndFrequencyResultBuilder.kt` line 67 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2258#-OQKUKSB2H46wI39WWsV:-OQKUKSB2H46wI39WWsW:b59yjor) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/564a9adaa3a162a1dacff6229f474d3c190fd534/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/compute/protocols/direct/DirectReachAndFrequencyResultBuilder.kt#L67)):*
> ```Kotlin
>    * @return The non-noisy reach and frequency measurement result.
>    */
>   override suspend fun buildMeasurementResult(): Measurement.Result {
> ```

I'd recommend structuing this slightly differently. Your client is just going to want to call "buildMeasurementResult" and have the right thing happen. You have a clean way of turning the noise off by setting the noise mechanism to None, and you've separated the core reach computation functions to another module (so they can be tested apart from this).

So here, you'd probably just want to check the noiser and if it's None just return the scaled result. If not call the noiser and then scale. Basically the two versions of this function can be combined.
___
*[`src/test/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/compute/protocols/direct/DirectReachAndFrequencyResultBuilderTest.kt` line 64 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2258#-OQKep573JyInsfyWfj_:-OQKep573JyInsfyWfja:b-lmjlmn) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/564a9adaa3a162a1dacff6229f474d3c190fd534/src/test/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/compute/protocols/direct/DirectReachAndFrequencyResultBuilderTest.kt#L64)):*
> ```Kotlin
>     assertThat(result.frequency.hasDeterministicDistribution()).isTrue()
>     assertThat(result.frequency.relativeFrequencyDistributionMap).isNotEmpty()
>   }
> ```

You need to check the actual values that are returned. 

In the case of the noisy measurement you can run the same test multiple times and average out the noise. You might also be able to play games through the injection of a random number generator. Ask @ple13 how to do this. This is being done already in the correctness test and possibly the EDP Simulator Tests.
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/noise/NoiseBuilder.kt` line 33 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2258#-OQK_z-XAxDNjXWssuGQ:-OQK_z-Y7B6t3sf5wV8H:bocf8h8) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/564a9adaa3a162a1dacff6229f474d3c190fd534/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/noise/NoiseBuilder.kt#L33)):*
> ```Kotlin
>  * Abstract class for building noisers.
>  */
> abstract class NoiseBuilder {
> ```

If you just create a single Noiser class (replacing this one and the one below) you can do the following:

Construct it with the noise mechanism, and random number generator. When you initialize it you can do the checks that ensure the noise mechanism is supported. 

Then supply three functions: 

1. getNoiser (Same as getPublisherNoise below. This is called "publisher noise" because LLv2 had several types of noise, but we should be able to drop the qualifier at this point.) It's not worth supporting Laplace noise.
2. addNoise(value: Int, privacyParams) : Int (replaces add noise for reach)
3. addNoise(value: frequencyMap, privacyParams): frequencyMap (replaces add noise for frequency)
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/noise/NoiseBuilder.kt` line 59 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2258#-OQKd03B4aZZsZWk_aPE:-OQKd03B4aZZsZWk_aPF:buwew38) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/564a9adaa3a162a1dacff6229f474d3c190fd534/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/noise/NoiseBuilder.kt#L59)):*
> ```Kotlin
>     // TODO: Use a proper DP library instead of the Halo built-in noisers.
>     when (directNoiseMechanism) {
>       DirectNoiseMechanism.NONE ->
> ```

Calling this function with None should be an error. If the client detects there is no noise mechanism it should not try to use a noiser.

jojijac0b (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2258)* status: 29 of 41 files reviewed, 14 unresolved discussions (waiting on @kungfucraig, @ple13, @SanjayVas, and @stevenwarejones)

___
*[`src/test/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/compute/protocols/direct/DirectReachAndFrequencyResultBuilderTest.kt` line 64 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2258#-OQKep573JyInsfyWfj_:-OQpFsFhEFxeIe3S5gLg:b-qwwp2c) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/564a9adaa3a162a1dacff6229f474d3c190fd534/src/test/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/compute/protocols/direct/DirectReachAndFrequencyResultBuilderTest.kt#L64)):*

You need to check the actual values that are returned. 

In the case of the noisy measurement you can run the same test multiple times and average out the noise. You might also be able to play games through the injection of a random number generator. Ask @ple13 how to do this. This is being done already in the correctness test and possibly the EDP Simulator Tests.
</blockquote></details>

Reusing same pattern used in InProcessReachMeasurementAccuracyTest by averaging the reach results of 100 runs and testing that they absolute difference compared to the actual reach is less than 5
___
*[`src/test/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/fulfillers/DirectMeasurementFulfillerTest.kt` line 90 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2258#-OQKfdt0C-poUsN-zBgV:-OQpL2Mc4j5tiZSb9vXQ:bif6znt) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/564a9adaa3a162a1dacff6229f474d3c190fd534/src/test/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/fulfillers/DirectMeasurementFulfillerTest.kt#L90)):*

You should check the value here.
</blockquote></details>

The encryptedResult within the fulfillRequisitionRequest will be different every time encryptResult() is called so it is difficult to check this value

kungfucraig (changes requested):
> Reviewed 2 of 34 files at r6, 8 of 10 files at r8, 1 of 1 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2258)* status: 40 of 41 files reviewed, 8 unresolved discussions (waiting on @jojijac0b and @SanjayVas)

___
*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2258#-OQtsJ5LDW8jxF-Yp6jc:-OQtsJ5LDW8jxF-Yp6jd:bt37utv) (no related file):*
This generally looks pretty good. The only thing I see missing is that each module should have its own test module. This means you can focus each layer on testing only the functionality it adds over the previous layer.
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/noise/Noiser.kt` line 41 at r10](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2258#-OQtpEIG-T9BsDkI18e7:-OQtpEIH5b1zONaqw1_c:b-n4g09z) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/58d1cc4c9e9715e2babab381d52b7e8b318e7a74/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/noise/Noiser.kt#L41)):*
> ```Kotlin
> class Noiser(
>   private val noiseMechanism: DirectNoiseMechanism,
>   private val random: SecureRandom = SecureRandom()
> ```

Please remove the default 
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/noise/Noiser.kt` line 60 at r10](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2258#-OQtpVXk0bijDu45AKgJ:-OQtpVXk0bijDu45AKgK:b-n7bjny) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/58d1cc4c9e9715e2babab381d52b7e8b318e7a74/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/noise/Noiser.kt#L60)):*
> ```Kotlin
>    */
>   fun getNoiser(privacyParams: DifferentialPrivacyParams): AbstractNoiser =
>     when (noiseMechanism) {
> ```

Just write this as a single if block:

if (is gaussian) return GaussianNoiser() else throw Refusal(unsupported $noiseMechanism)

jojijac0b (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2258)* status: 40 of 41 files reviewed, 8 unresolved discussions (waiting on @kungfucraig and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/noise/Noiser.kt` line 41 at r10](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2258#-OQtpEIG-T9BsDkI18e7:-OR2LsVUD0ofcBE3KPTW:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/58d1cc4c9e9715e2babab381d52b7e8b318e7a74/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/noise/Noiser.kt#L41)):*

Please remove the default
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/noise/Noiser.kt` line 60 at r10](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2258#-OQtpVXk0bijDu45AKgJ:-OR2MGvVBzBOuVPDXhtQ:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/58d1cc4c9e9715e2babab381d52b7e8b318e7a74/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/noise/Noiser.kt#L60)):*

Just write this as a single if block:

if (is gaussian) return GaussianNoiser() else throw Refusal(unsupported $noiseMechanism)
</blockquote></details>

Done.

jojijac0b (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2258)* status: 36 of 44 files reviewed, 8 unresolved discussions (waiting on @kungfucraig, @SanjayVas, and @stevenwarejones)

___
*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2258#-OQtsJ5LDW8jxF-Yp6jc:-OR2xxDS0IHWkGMgZMd6:b-9czs36) (no related file):*

This generally looks pretty good. The only thing I see missing is that each module should have its own test module. This means you can focus each layer on testing only the functionality it adds over the previous layer.
</blockquote></details>

Added tests for MeasurementResults and DirectMeasurementResultFactory

kungfucraig (approved):
> Reviewed 1 of 1 files at r9, 10 of 10 files at r11, 1 of 1 files at r12, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2258)* status: all files reviewed, 4 unresolved discussions (waiting on @SanjayVas)

PR #2179, opened 2025-04-12
edp aggregator in process test2

stevenwarejones:
Just a draft.

Event Groups are properly registering.
The requisition fetcher successfully calls the kingdom but probably doesn't have the right config to write to storage.
Didn't finish setting up Data Watcher yet. Not much to configure there.

PR #2271, opened 2025-05-06
test: test MC support event groups without encrypted metadata

stevenwarejones:
Background: 
The current in process test supports synthetic event configs embedded in the encrypted metadata of the event group. The EDP Aggregator does not support encrypted metadata.

If encrypted metadata is not present, will use the index present at the end of the event group reference id. For example, event-group-reference-id-eg-11 refers to the SyntheticEventGroupSpec with index 11.

This has been tested by the In Process EDP Aggregator Life Of a Measurement Test in https://github.com/world-federation-of-advertisers/cross-media-measurement/pull/2179


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2271)

kungfucraig (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2271)* status: 0 of 4 files reviewed, 1 unresolved discussion

___
*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2271#-OPbVDnxBazJrnvPL5Ct:-OPbVDny9M0aZO-lNNF6:b-7ofqfx) (no related file):*
I don't understand why the EDP Aggregator has a dependency on the data provider parts that are modified herein.

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2271)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @kungfucraig)

___
*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2271#-OPbVDnxBazJrnvPL5Ct:-OPbWECQBj0hCcNfvi9P:b-5ygt9p) (no related file):*

I don't understand why the EDP Aggregator has a dependency on the data provider parts that are modified herein.
</blockquote></details>

This PR allows the EDP Aggregator to use the same synthetic event generator for its tests. The current Synthetic Event Generator was built and hardcoded to depend on specific encrypted metadata on the Event Group. That field is getting deprecated.

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2271)* status: 0 of 4 files reviewed, 2 unresolved discussions (waiting on @kungfucraig)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EventQuery.kt` line 41 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2271#-OPg4PYU3j_2num40eHI:-OPg4PYU3j_2num40eHJ:b-iqdduy) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/fd6122fe51e356e4cc12ee332bbae6c0cba75d77/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EventQuery.kt#L41)):*
> ```Kotlin
>   fun getLabeledEvents(
>     eventGroupSpec: EventGroupSpec,
>     filterEvents: Boolean,
> ```

1. Optional behavior like this is a code smell. EventQuery is also specifically intended to do the filtering, so this is making a key piece of its behavior optional.
2. EventQuery shouldn't be used for anything new as it's on the chopping block. It's an unnecessary abstraction from when we had to support multiple data sources, i.e. multiple EDP simulator implementations. Now we only have one, so there's no need to have this interface at all.

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2271)* status: 0 of 4 files reviewed, 2 unresolved discussions (waiting on @kungfucraig and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EventQuery.kt` line 41 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2271#-OPg4PYU3j_2num40eHI:-OPkHa6e9SRnQjT8VHBj:bs0p0tx) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/fd6122fe51e356e4cc12ee332bbae6c0cba75d77/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EventQuery.kt#L41)):*

1. Optional behavior like this is a code smell. EventQuery is also specifically intended to do the filtering, so this is making a key piece of its behavior optional.
2. EventQuery shouldn't be used for anything new as it's on the chopping block. It's an unnecessary abstraction from when we had to support multiple data sources, i.e. multiple EDP simulator implementations. Now we only have one, so there's no need to have this interface at all.
</blockquote></details>

This makes sense. I still need the change in `MetadataSyntheticGeneratorEventQuery` in order for the MC to be able to check the result themselves as I understand it but dropped all the dataprovider changes.

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2271)* status: 0 of 4 files reviewed, 2 unresolved discussions (waiting on @kungfucraig and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EventQuery.kt` line 41 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2271#-OPg4PYU3j_2num40eHI:-OPkHyVACEv933KiUz2i:bdf2zfd) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/fd6122fe51e356e4cc12ee332bbae6c0cba75d77/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EventQuery.kt#L41)):*

This makes sense. I still need the change in `MetadataSyntheticGeneratorEventQuery` in order for the MC to be able to check the result themselves as I understand it but dropped all the dataprovider changes.
</blockquote></details>

reverted the rest of the changes except for that.

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2271)* status: 1 of 4 files reviewed, 3 unresolved discussions (waiting on @kungfucraig and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EventQuery.kt` line 41 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2271#-OPg4PYU3j_2num40eHI:-OPkkog_CSsvzWC4sPJm:b-mkhfm0) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/fd6122fe51e356e4cc12ee332bbae6c0cba75d77/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EventQuery.kt#L41)):*
> This makes sense. I still need the change in `MetadataSyntheticGeneratorEventQuery` in order for the MC to be able to check the result themselves

What exactly is this for? You shouldn't be trying to add the EDP Aggregator to the existing correctness test, as that test is specifically designed around the EDP simulators. That is to say that the EventQuery mechanism is strictly a contract between the correctness test and the EDP simulator.

What is your testing plan for EDPA?
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/measurementconsumer/MetadataSyntheticGeneratorEventQuery.kt` line 45 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2271#-OPkjwmH9YV2j2Rt_S6f:-OPkjwmH9YV2j2Rt_S6g:b63gvtu) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e597363b81d9a5773d27e36efeb48419f19a7dd5/src/main/kotlin/org/wfanet/measurement/loadtest/measurementconsumer/MetadataSyntheticGeneratorEventQuery.kt#L45)):*
> ```Kotlin
> 
>   companion object {
>     private fun parseEdpNumber(input: String): Int {
> ```

Display names should always be treated as opaque strings.

PR #2275, opened 2025-05-06
feat: Implement compressed frequency vectors

robinsons:
Example of compressing frequency vectors using bitmaps.

Suppose you have a frequency vector `<r1, r2, r3, ..., rN>` of registers, where each register has a value in the range `0..maxFrequency`. A `CompressedFrequencyVector` is a collection of optimized bitmaps, with one bitmap for each frequency value `1..maxFrequency`. Each bitmap contains the set of registers with the corresponding frequency, for example:

* 1 -> bitmap of registers with frequency == 1
* 2 -> bitmap of registers with frequency == 2
* ...
* maxFrequency -> bitmap of registers with frequency == maxFrequency

We also provide a function `Iterable<CompressedFrequencyVector>.merge()` which produces a (uncompressed) frequency vector obtained by summing the register-wise values of the given `CompressedFrequencyVector`s.

For EDPs who implement fulfillment using a map-reduce pipeline, this can be used like so:

1. For a given requisition, divide impressions into batches with each batch having size `N` (say 100K).
2. For each batch, use `FrequencyVectorBuilder` to build a `FrequencyVector`, and then call `compress()` to turn it into a `CompressedFrequencyVector`.
3. Merge the `CompressedFrequencyVector`s to produce the final `FrequencyVector`.

This implementation uses [RoaringBitmap](https://github.com/RoaringBitmap/RoaringBitmap).


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2275)

stevenwarejones (commented):
> @stevenwarejones reviewed 1 of 6 files at r1.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2275)* status: 1 of 6 files reviewed, 2 unresolved discussions

___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/CompressedFrequencyVector.kt` line 43 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2275#-Oa5auYi-9rkDEB6gTO1:-Oa5auYj6vUlXCg_hJcz:b-ccp8sj) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/efd1c2287fd7e7213e12c20718b2fdc6d0328aa5/src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/CompressedFrequencyVector.kt#L43)):*
> ```Kotlin
>    */
>   val registerBitmaps: List<ByteString>,
> ) : Iterable<RegisterAndFrequency> {
> ```

should extend Serializable
___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/CompressedFrequencyVector.kt` line 65 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2275#-Oa5ay8s-TC_6eUAPfZn:-Oa5ay8s-TC_6eUAPfZo:b-ccp8sj) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/efd1c2287fd7e7213e12c20718b2fdc6d0328aa5/src/main/kotlin/org/wfanet/measurement/eventdataprovider/shareshuffle/CompressedFrequencyVector.kt#L65)):*
> ```Kotlin
> 
> /** Container for a frequency vector [register] and its associated [frequency]. */
> data class RegisterAndFrequency(val register: Int, val frequency: Int)
> ```

should extend Serializable

PR #2276, opened 2025-05-06
fix: submit compliant rfc-1034 ids in control plane

stevenwarejones:
Background: Random UUIDs are not RFC-1034 compliant. This causes error whenever IDs are required to be RFC-1034 compliant like the Secure Computation Control Plane requires

This uses a similar approach to https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5f975e854772a4f871fe84b73e1494adf27e3c26/src/main/kotlin/org/wfanet/measurement/access/common/TlsClientPrincipalMapping.kt#L42


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2276)

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2276)* status: 0 of 3 files reviewed, all discussions resolved (waiting on @kungfucraig)

___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/requisitionfetcher/RequisitionFetcher.kt` line 86 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2276#-OPbRNJC4HuAzXY1FWb1:-OPbRNJD3r91hONiJj1F:b-tt3ptg) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5f975e854772a4f871fe84b73e1494adf27e3c26/src/main/kotlin/org/wfanet/measurement/edpaggregator/requisitionfetcher/RequisitionFetcher.kt#L86)):*
> ```Kotlin
>         .flattenConcat()
> 
>     logger.fine {
> ```

note: this log is useless since the flow hasn't been collected

kungfucraig (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2276)* status: 0 of 3 files reviewed, 1 unresolved discussion

___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt` line 94 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2276#-OPbSxsL4dr2ubZtJdAU:-OPbSxsM7CsvNz8kDSow:byn6r7v) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/32098c2222fe808dac4fcf3e9a3b3eee72be543d/src/main/kotlin/org/wfanet/measurement/securecomputation/teesdk/BaseTeeApplication.kt#L94)):*
> ```Kotlin
>       try {
>         val workItemAttemptId =
>           UUID.randomUUID().toString().replace("/", "-").replace("_", "-").takeLast(63).trim {
> ```

This logic should be in a function that BaseTeeApp and DataWatcher can share with future apps.

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2276)* status: 0 of 3 files reviewed, 2 unresolved discussions (waiting on @kungfucraig and @stevenwarejones)

___
*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2276#-OPbqODDC-y7n13O2jle:-OPbqODDC-y7n13O2jlf:b-v660rj) (no related file):*
> Random UUIDs are not RFC-1034 compliant.

They're not? The string format is 36 characters of hex and dashes. The only thing you need to do is make sure it starts with a letter, which could be as simple as having a prefix. e.g. prefix the random work item UUIDs with`wi-`.

> This uses a similar approach to

Don't copy that approach. That is for the very specific use case of having synthetic resources returned from an existing config file.

SanjayVas (approved):
> Reviewed 1 of 3 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2276)* status: all files reviewed, 2 unresolved discussions (waiting on @kungfucraig)

___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/datawatcher/DataWatcher.kt` line 73 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2276#-OPfg8Dh-ZXQvurySqQ_:-OPfg8Dh-ZXQvurySqQa:b-l1u795) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/2d77aa1d1aa8fd97ae92f69af52445f66a20c9a4/src/main/kotlin/org/wfanet/measurement/securecomputation/datawatcher/DataWatcher.kt#L73)):*
> ```Kotlin
> 
>     val queueConfig = config.controlPlaneQueueSink
>     val workItemId = "work-item-" + UUID.randomUUID().toString()
> ```

nit: you may want to pass in an ID generator function into the constructor so you can make test behavior deterministic. e.g. you can update DataWatcherTest to actually cover the entire CreateWorkItemRequest message rather than just asserting on specific fields.

As a tip if you do that, you can pass a TypeRegistry to ProtoSubject using `unpackingAnyUsing` so it unpacks Any messages for you. This will make is so `isEqualTo`will compare the contents of the Any fields as well.



_Code quote:_
```Kotlin
UUID.randomUUID()
```

stevenwarejones (commented):
> Reviewed 2 of 2 files at r2.
> Dismissed @kungfucraig from a discussion.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2276)* status: all files reviewed, 1 unresolved discussion (waiting on @kungfucraig)

kungfucraig (approved):
> Reviewed 1 of 3 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2276)* status: all files reviewed, 1 unresolved discussion (waiting on @stevenwarejones)

stevenwarejones (commented):
> Reviewed 1 of 3 files at r1, 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2276)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @stevenwarejones)

PR #2278, opened 2025-05-06
feat: Implement EventGroup metadata filters and order by in Kingdom

SanjayVas:
This implements world-federation-of-advertisers/cross-media-measurement-api#233.


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2278)

SanjayVas:
CC @kungfucraig

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2278)* status: 0 of 18 files reviewed, 1 unresolved discussion (waiting on @tristanvuong2021)

___
*[`src/main/resources/kingdom/spanner/add-event-group-metadata.sql` line 54 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2278#-OPbxmrG-Q_5m_ZlBWT4:-OPbxmrG-Q_5m_ZlBWT5:bmrhle6) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d6715a572ecafe964d0e582c690e55f7bdb6d405/src/main/resources/kingdom/spanner/add-event-group-metadata.sql#L54)):*
> ```TSQL
> CREATE SEARCH INDEX EventGroupsByMetadata ON EventGroups(Metadata_Tokens);
> 
> CREATE INDEX EventGroupsByDataAvailabilityStartTime ON EventGroups(DataAvailabilityStartTime DESC);
> ```

@kungfucraig I'm assuming we don't need to shard this index since it's by the availability start time rather than e.g. update time, and therefore we don't have the problem of having a hot spot at the end.

If we do need to shard, my instinct would be to do so naturally by MC ID since ordering by start time is most likely going to happen when also filtering by MC. WDYT?

kungfucraig (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2278)* status: 0 of 18 files reviewed, 1 unresolved discussion (waiting on @SanjayVas and @tristanvuong2021)

___
*[`src/main/resources/kingdom/spanner/add-event-group-metadata.sql` line 54 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2278#-OPbxmrG-Q_5m_ZlBWT4:-OPc5275Ai_YqI1dshO7:bfhtfxf) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d6715a572ecafe964d0e582c690e55f7bdb6d405/src/main/resources/kingdom/spanner/add-event-group-metadata.sql#L54)):*

@kungfucraig I'm assuming we don't need to shard this index since it's by the availability start time rather than e.g. update time, and therefore we don't have the problem of having a hot spot at the end.

If we do need to shard, my instinct would be to do so naturally by MC ID since ordering by start time is most likely going to happen when also filtering by MC. WDYT?
</blockquote></details>

I see no harm in putting an MC prefix before it, especially given current usage where all queries are MC scoped. 

My guess is that we won't have any issues because volume for EG inserts should be fairly low, but you could imagine that if an EDP is inserting new EGs daily that the start times for a particular day are all quite similar.


Did you plan on adding an index for EndTime too?

tristanvuong2021 (changes requested):
> Reviewed 18 of 18 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2278)* status: all files reviewed, 3 unresolved discussions (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/internal/testing/EventGroupsServiceTest.kt` line 756 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2278#-OPg9ZefCrVxHhJnIY-N:-OPg9ZefCrVxHhJnIY-O:bnqmuna) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d6715a572ecafe964d0e582c690e55f7bdb6d405/src/main/kotlin/org/wfanet/measurement/kingdom/service/internal/testing/EventGroupsServiceTest.kt#L756)):*
> ```Kotlin
>       .containsExactlyElementsIn(
>         eventGroups.filter {
>           it.details.metadata.adMetadata.campaignMetadata.campaignName.contains("better", true)
> ```

The documentation states the Spanner SEARCH function is case insensitive and that is what is being used in the implementation, but just looking at the test I cannot tell what the expected response is.
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/internal/testing/EventGroupsServiceTest.kt` line 786 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2278#-OPgAFXx7aoogQvrm-x1:-OPgAFXx7aoogQvrm-x2:bvt2vc2) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d6715a572ecafe964d0e582c690e55f7bdb6d405/src/main/kotlin/org/wfanet/measurement/kingdom/service/internal/testing/EventGroupsServiceTest.kt#L786)):*
> ```Kotlin
>       .ignoringRepeatedFieldOrderOfFields(EventGroup.MEDIA_TYPES_FIELD_NUMBER)
>       .containsExactlyElementsIn(
>         eventGroups.filter {
> ```

Same as campaign name search test.

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2278)* status: 17 of 18 files reviewed, 3 unresolved discussions (waiting on @kungfucraig and @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/internal/testing/EventGroupsServiceTest.kt` line 756 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2278#-OPg9ZefCrVxHhJnIY-N:-OPgRDwlDadr_r-_5zOQ:bkdzry6) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d6715a572ecafe964d0e582c690e55f7bdb6d405/src/main/kotlin/org/wfanet/measurement/kingdom/service/internal/testing/EventGroupsServiceTest.kt#L756)):*

The documentation states the Spanner SEARCH function is case insensitive and that is what is being used in the implementation, but just looking at the test I cannot tell what the expected response is.
</blockquote></details>

This test is indeed verifying the case insensitivity. The actual text ("Better Than Bad") is capitalized and the search text ("better") is lowercase. This is why we use case-insensitive contains for the filter on the expected result.

Did you have a recommendation to make that more clear?
___
*[`src/main/resources/kingdom/spanner/add-event-group-metadata.sql` line 54 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2278#-OPbxmrG-Q_5m_ZlBWT4:-OPh3RAyCVExPtHT82hC:b-qfmrb1) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d6715a572ecafe964d0e582c690e55f7bdb6d405/src/main/resources/kingdom/spanner/add-event-group-metadata.sql#L54)):*

I see no harm in putting an MC prefix before it, especially given current usage where all queries are MC scoped. 

My guess is that we won't have any issues because volume for EG inserts should be fairly low, but you could imagine that if an EDP is inserting new EGs daily that the start times for a particular day are all quite similar.


Did you plan on adding an index for EndTime too?
</blockquote></details>

I'm actually going to leave this index out for now. From testing on dev, I can't get it to actually reduce the number of rows scanned, and the query optimizer never picks this index itself. I think we'll just need to see how this performs once we actually have an idea of the realistic usage on backfilled EG data.

I think it's okay to start like this since DataProviders and EventGroups rows aren't written very frequently, so there should be very little write contention.

tristanvuong2021 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2278)* status: 17 of 18 files reviewed, 3 unresolved discussions (waiting on @kungfucraig and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/internal/testing/EventGroupsServiceTest.kt` line 756 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2278#-OPg9ZefCrVxHhJnIY-N:-OPkFWcOE7LDRSK5c8mO:baqs4p0) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d6715a572ecafe964d0e582c690e55f7bdb6d405/src/main/kotlin/org/wfanet/measurement/kingdom/service/internal/testing/EventGroupsServiceTest.kt#L756)):*

This test is indeed verifying the case insensitivity. The actual text ("Better Than Bad") is capitalized and the search text ("better") is lowercase. This is why we use case-insensitive contains for the filter on the expected result.

Did you have a recommendation to make that more clear?
</blockquote></details>

Maybe just making the argument a named argument so it ends up `contains("better", ignoreCase = true)`

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2278)* status: 16 of 18 files reviewed, 3 unresolved discussions (waiting on @kungfucraig and @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/internal/testing/EventGroupsServiceTest.kt` line 756 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2278#-OPg9ZefCrVxHhJnIY-N:-OPkjBmQEXlR9_JFVqdR:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d6715a572ecafe964d0e582c690e55f7bdb6d405/src/main/kotlin/org/wfanet/measurement/kingdom/service/internal/testing/EventGroupsServiceTest.kt#L756)):*

Maybe just making the argument a named argument so it ends up `contains("better", ignoreCase = true)`
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/internal/testing/EventGroupsServiceTest.kt` line 786 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2278#-OPgAFXx7aoogQvrm-x1:-OPkjC4gBst-TmTS8umX:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d6715a572ecafe964d0e582c690e55f7bdb6d405/src/main/kotlin/org/wfanet/measurement/kingdom/service/internal/testing/EventGroupsServiceTest.kt#L786)):*

Same as campaign name search test.
</blockquote></details>

Done.

tristanvuong2021 (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2278)* status: all files reviewed, 1 unresolved discussion (waiting on @kungfucraig)

kungfucraig (approved):
> Reviewed 16 of 18 files at r1, 1 of 1 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2278)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @stevenwarejones)

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2278)* status: all files reviewed, 1 unresolved discussion (waiting on @stevenwarejones)

___
*[`src/main/resources/kingdom/spanner/add-event-group-metadata.sql` line 21 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2278#-OPqmnbjC-cCdfMGQywr:-OPr3M-m4RBepD_uY21P:bn7wtbi) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8469cead91325ad52a1dd71ee54d29668fff8d0f/src/main/resources/kingdom/spanner/add-event-group-metadata.sql#L21)):*

why does the descriptor need to be in the db?
</blockquote></details>

That's how protocol buffer support in Spanner works. Otherwise Spanner doesn't know how read fields inside of a protocol buffer column. See the README in this directory.

Also note that this PR is just *updating* the descriptor in the Spanner DB since there have been relevant changes to the message types used in Spanner columns, as mentioned in that README.

stevenwarejones (approved):
> Reviewed 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2278)* status: all files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

___
*[`src/main/resources/kingdom/spanner/add-event-group-metadata.sql` line 21 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2278#-OPqmnbjC-cCdfMGQywr:-OQ3vIsXCbtrAcNRRRnO:b-ro15wh) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8469cead91325ad52a1dd71ee54d29668fff8d0f/src/main/resources/kingdom/spanner/add-event-group-metadata.sql#L21)):*

That's how protocol buffer support in Spanner works. Otherwise Spanner doesn't know how read fields inside of a protocol buffer column. See the README in this directory.

Also note that this PR is just *updating* the descriptor in the Spanner DB since there have been relevant changes to the message types used in Spanner columns, as mentioned in that README.
</blockquote></details>

i realized it was just updating. That makes sense that Spanner would need the descriptor. I missed that we had done this in the past when the prior PR came in.

SanjayVas (commented):
> Reviewed 2 of 2 files at r4.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2278)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @SanjayVas)

PR #2317, opened 2025-05-19
feat: Implement public ModelProvidersService

roaminggypsy:


roaminggypsy (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2317)* status: 3 of 7 files reviewed, 4 unresolved discussions (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/ModelProvidersService.kt` line 78 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2317#-ORlhcnN0pHbh8QqhDJL:-ORrg1pn1ybeioyR6uVZ:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/aa4da8dadbcae95749a1d12a6234df741ba8d36f/src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/ModelProvidersService.kt#L78)):*

We're switching to INTERNAL for these and letting UNKNOWN just be used automatically (e.g. for uncaught exception types)
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/ModelProvidersService.kt` line 85 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2317#-ORljbsJ3EHNc9Z3hJmP:-ORrhdEbDa8gscVammkf:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/aa4da8dadbcae95749a1d12a6234df741ba8d36f/src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/ModelProvidersService.kt#L85)):*

This needs to implement max and default page size and use `listResources` to call the internal service, or else the internal service needs to use the same max and default sizes as documented in the public API. Right now this isn't true, as the internal service has a max of 100 whereas the [public service is documented](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/pull/236) as having a max of 1000.

If you change the internal service impl, make sure to update [its documentation](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/main/src/main/proto/wfa/measurement/internal/kingdom/model_providers_service.proto) so it's clear that the max page size is expected for any impl.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/ModelProvidersService.kt` line 94 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2317#-ORlkUDeA6tQQPo57bKn:-ORri9PtEFX3rmMa_TBP:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/aa4da8dadbcae95749a1d12a6234df741ba8d36f/src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/ModelProvidersService.kt#L94)):*

Catch the specific documented exception type(s). If another exception type is thrown, you don't know that it's due to having an invalid page token.
</blockquote></details>

Done.

PR #2339, opened 2025-05-22
feat: add requisition grouper interface and single requisition grouper

stevenwarejones:


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2339)

kungfucraig (changes requested):
> Reviewed 1 of 4 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2339)* status: 1 of 5 files reviewed, 10 unresolved discussions

___
*[`src/main/proto/wfa/measurement/edpaggregator/v1alpha/grouped_requisitions.proto` line 29 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2339#-OQtkkX-9x3gIKNTTfK0:-OQtkkX-9x3gIKNTTfK1:b-a3sdjx) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/eb72d95b7e8df5a16231b8eab73979960db4e899/src/main/proto/wfa/measurement/edpaggregator/v1alpha/grouped_requisitions.proto#L29)):*
> ```proto
> 
> // Message representing a group of requisitions to be processed together.
> message GroupedRequisitions {
> ```

Any reason you didn't extract the CEL filter?
___
*[`src/main/proto/wfa/measurement/edpaggregator/v1alpha/grouped_requisitions.proto` line 32 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2339#-OQtifPu3fhnCEvlgyW1:-OQtifPu3fhnCEvlgyW2:benfyzi) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/eb72d95b7e8df5a16231b8eab73979960db4e899/src/main/proto/wfa/measurement/edpaggregator/v1alpha/grouped_requisitions.proto#L32)):*
> ```proto
>   // The model line for the requisitions
>   string model_line = 1 [
>     (google.api.resource_reference).type = "halo-system.wfanet.org/ModelLine",
> ```

"halo.wfanet.org/ModelLine"
___
*[`src/main/proto/wfa/measurement/edpaggregator/v1alpha/grouped_requisitions.proto` line 35 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2339#-OQtiFrvEHLLBhiDvdOc:-OQtiFrvEHLLBhiDvdOd:b-gk4cnx) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/eb72d95b7e8df5a16231b8eab73979960db4e899/src/main/proto/wfa/measurement/edpaggregator/v1alpha/grouped_requisitions.proto#L35)):*
> ```proto
>     (google.api.field_behavior) = REQUIRED
>   ];
>   // A map of event group resource name to event group reference id. Required
> ```

It is expected that these requisitions have EGs that perfectly overlap those in the map? 

Similarly, what is the relationship between the collection_intervals and the requisitions?
___
*[`src/main/proto/wfa/measurement/edpaggregator/v1alpha/grouped_requisitions.proto` line 37 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2339#-OQtih3MDfd1evU2n_WY:-OQtih3MDfd1evU2n_WZ:b-smv9h4) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/eb72d95b7e8df5a16231b8eab73979960db4e899/src/main/proto/wfa/measurement/edpaggregator/v1alpha/grouped_requisitions.proto#L37)):*
> ```proto
>   // A map of event group resource name to event group reference id. Required
>   // for all event groups present in requisitions.
>   map<string, string> event_group_map = 2
> ```

You should consider doing an unrolled EventGroupMapEntry so we can properly annotate the keys and values. 

For an example see: https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3779f3f887278e2d46c72f348dc55fe931e71ca4/src/main/proto/wfa/measurement/reporting/v2alpha/result_group.proto#L366
___
*[`src/main/proto/wfa/measurement/edpaggregator/v1alpha/grouped_requisitions.proto` line 45 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2339#-OQtj2J_4r1iqe-Wcph2:-OQtj2J_4r1iqe-Wcph3:bw3k0xd) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/eb72d95b7e8df5a16231b8eab73979960db4e899/src/main/proto/wfa/measurement/edpaggregator/v1alpha/grouped_requisitions.proto#L45)):*
> ```proto
>       [(google.api.field_behavior) = REQUIRED];
>   // The requisitions to process together.
>   repeated google.protobuf.Any requisitions = 4 [
> ```

You might provide a link to the Requisition message in a comment.
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/requisitionfetcher/RequisitionGrouper.kt` line 44 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2339#-OQti1woAesV7Pm28gfA:-OQti1woAesV7Pm28gfB:b-aqst2p) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/eb72d95b7e8df5a16231b8eab73979960db4e899/src/main/kotlin/org/wfanet/measurement/edpaggregator/requisitionfetcher/RequisitionGrouper.kt#L44)):*
> ```Kotlin
> ) {
> 
>   fun groupRequisitions(requisitions: List<Requisition>): List<GroupedRequisitions> {
> ```

Can you add a couple sentences of documentation that describe what these functions are supposed to do?
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/requisitionfetcher/RequisitionGrouper.kt` line 53 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2339#-OQtk0Cb2MhIBnZ7RWgN:-OQtk0Cc4npygLyPcX0e:bk076dr) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/eb72d95b7e8df5a16231b8eab73979960db4e899/src/main/kotlin/org/wfanet/measurement/edpaggregator/requisitionfetcher/RequisitionGrouper.kt#L53)):*
> ```Kotlin
>   ): List<GroupedRequisitions>
> 
>   private fun mapRequisitions(requisitions: List<Requisition>): List<GroupedRequisitions> {
> ```

I would create a function called requisitionToGroupedRequisition as that is very clearly doing a simple conversion and then maybe layer this on top of it. Or you could call this requisitionsToGroupedRequisitions.  It took me reading the whole function to understand what it did.
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/requisitionfetcher/RequisitionGrouper.kt` line 55 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2339#-OQtjUoM3pG5SsjjJJ4v:-OQtjUoM3pG5SsjjJJ4w:bvtsmb5) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/eb72d95b7e8df5a16231b8eab73979960db4e899/src/main/kotlin/org/wfanet/measurement/edpaggregator/requisitionfetcher/RequisitionGrouper.kt#L55)):*
> ```Kotlin
>   private fun mapRequisitions(requisitions: List<Requisition>): List<GroupedRequisitions> {
>     return requisitions.mapNotNull { requisition ->
>       val measurementSpec: MeasurementSpec? =
> ```

If this fails we'll basically never fulfill the requisition, right? If so that doesn't seem good. It seems like it would be better to set this requisition up to be refused.
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/requisitionfetcher/RequisitionGrouper.kt` line 65 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2339#-OQtjj2V9hymRHwwWZvt:-OQtjj2V9hymRHwwWZvu:bijpb0c) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/eb72d95b7e8df5a16231b8eab73979960db4e899/src/main/kotlin/org/wfanet/measurement/edpaggregator/requisitionfetcher/RequisitionGrouper.kt#L65)):*
> ```Kotlin
>         null
>       }
>       val requisitionSpec: RequisitionSpec? =
> ```

Similar to above.
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/requisitionfetcher/RequisitionGrouper.kt` line 105 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2339#-OQtkLokD16dXSFr719Y:-OQtkLokD16dXSFr719Z:b4w6hfl) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/eb72d95b7e8df5a16231b8eab73979960db4e899/src/main/kotlin/org/wfanet/measurement/edpaggregator/requisitionfetcher/RequisitionGrouper.kt#L105)):*
> ```Kotlin
> 
>   private suspend fun getEventGroup(name: String): EventGroup {
>     return eventGroupsClient.getEventGroup(getEventGroupRequest { this.name = name })
> ```

This needs to have a retry loop with backoff.

If you have the time I wouldn't mind having a utility abstracts away the retry loop and backoff functionality that allows it to be used with client + method.

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2339)* status: 0 of 9 files reviewed, 9 unresolved discussions (waiting on @kungfucraig)

___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/requisitionfetcher/RequisitionGrouper.kt` line 44 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2339#-OQti1woAesV7Pm28gfA:-ORktMieDFsixORKtEo2:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/eb72d95b7e8df5a16231b8eab73979960db4e899/src/main/kotlin/org/wfanet/measurement/edpaggregator/requisitionfetcher/RequisitionGrouper.kt#L44)):*

Can you add a couple sentences of documentation that describe what these functions are supposed to do?
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/requisitionfetcher/RequisitionGrouper.kt` line 53 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2339#-OQtk0Cb2MhIBnZ7RWgN:-ORktOa-8sARomZmS-H5:br2qacx) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/eb72d95b7e8df5a16231b8eab73979960db4e899/src/main/kotlin/org/wfanet/measurement/edpaggregator/requisitionfetcher/RequisitionGrouper.kt#L53)):*

I would create a function called requisitionToGroupedRequisition as that is very clearly doing a simple conversion and then maybe layer this on top of it. Or you could call this requisitionsToGroupedRequisitions.  It took me reading the whole function to understand what it did.
</blockquote></details>

changed this a simple map 1:1 with comments. lmk if this is more clear.
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/requisitionfetcher/RequisitionGrouper.kt` line 55 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2339#-OQtjUoM3pG5SsjjJJ4v:-ORmZG1O8IjNa8grtD8Q:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/eb72d95b7e8df5a16231b8eab73979960db4e899/src/main/kotlin/org/wfanet/measurement/edpaggregator/requisitionfetcher/RequisitionGrouper.kt#L55)):*

If this fails we'll basically never fulfill the requisition, right? If so that doesn't seem good. It seems like it would be better to set this requisition up to be refused.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/requisitionfetcher/RequisitionGrouper.kt` line 65 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2339#-OQtjj2V9hymRHwwWZvt:-ORmZGzl4Mm4kRPNtRe5:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/eb72d95b7e8df5a16231b8eab73979960db4e899/src/main/kotlin/org/wfanet/measurement/edpaggregator/requisitionfetcher/RequisitionGrouper.kt#L65)):*

Similar to above.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/requisitionfetcher/RequisitionGrouper.kt` line 105 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2339#-OQtkLokD16dXSFr719Y:-ORkuNcj1xDTQDJSK8pI:bgmb1ch) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/eb72d95b7e8df5a16231b8eab73979960db4e899/src/main/kotlin/org/wfanet/measurement/edpaggregator/requisitionfetcher/RequisitionGrouper.kt#L105)):*

This needs to have a retry loop with backoff.

If you have the time I wouldn't mind having a utility abstracts away the retry loop and backoff functionality that allows it to be used with client + method.
</blockquote></details>

i added in a throttler. Do we have retries on public to internal API for the kingdom? If not, I'm not sure I see the point of having it here. Nonetheless, created a Github issue.

https://github.com/world-federation-of-advertisers/cross-media-measurement/issues/2372

kungfucraig (changes requested):
> Reviewed 1 of 5 files at r1, 3 of 7 files at r3, 5 of 5 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2339)* status: all files reviewed, 5 unresolved discussions (waiting on @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/edpaggregator/v1alpha/grouped_requisitions.proto` line 37 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2339#-ORsgeOcAWHMbP1IPSh_:-ORsgeOd5jdjUGhOsw31:b-77kqtf) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f27cc26b5e17d5d296a26aedfcfcdfb92d982aaf/src/main/proto/wfa/measurement/edpaggregator/v1alpha/grouped_requisitions.proto#L37)):*
> ```proto
>   ];
>   // A single entry mapping an Event Group to an Event Group Reference Id.
>   message EventGroupMapEntry {
> ```

Why is the whole EG the key instead of the EG reference ID? Is this really a map or just a list of pairs?
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/requisitionfetcher/RequisitionGrouper.kt` line 86 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2339#-ORsdhaq2MdEF_o_021J:-ORsdhaq2MdEF_o_021K:bfzdeja) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f27cc26b5e17d5d296a26aedfcfcdfb92d982aaf/src/main/kotlin/org/wfanet/measurement/edpaggregator/requisitionfetcher/RequisitionGrouper.kt#L86)):*
> ```Kotlin
>   private fun mapRequisition(requisition: Requisition): GroupedRequisitions? {
> 
>     val measurementSpec: MeasurementSpec? =
> ```

You should consider a helper function/class that unpacks the requisition by getting and validating everything you need and returning the various null checked subobjects. I think this could include the EventGroup queries too.

Basically it would be a lot cleaner to do the validation in a separate function, throw an exception and have a single catch block that turns that exception into a refusal.

It's also a lot easier to test your validation logic because you can just assert that the proper is exception is thrown instead of mocking refuse requisition. 
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/requisitionfetcher/RequisitionGrouper.kt` line 160 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2339#-ORsejIZ8XzJ688MzO6A:-ORsejIZ8XzJ688MzO6B:b1r2efn) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f27cc26b5e17d5d296a26aedfcfcdfb92d982aaf/src/main/kotlin/org/wfanet/measurement/edpaggregator/requisitionfetcher/RequisitionGrouper.kt#L160)):*
> ```Kotlin
>     for (eventGroupEntry in requisitionSpec!!.events.eventGroupsList) {
>       val eventGroupName = eventGroupEntry.key
>       val eventGroup = runBlocking { getEventGroup(eventGroupName) }
> ```

You should really implement batch get for Event Groups. I think this has already come up as part of the EG Sync work.
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/requisitionfetcher/RequisitionGrouper.kt` line 161 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2339#-ORsdALAC6Dr_vbcDDzB:-ORsdALAC6Dr_vbcDDzC:b-yn57bt) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f27cc26b5e17d5d296a26aedfcfcdfb92d982aaf/src/main/kotlin/org/wfanet/measurement/edpaggregator/requisitionfetcher/RequisitionGrouper.kt#L161)):*
> ```Kotlin
>       val eventGroupName = eventGroupEntry.key
>       val eventGroup = runBlocking { getEventGroup(eventGroupName) }
>       if (eventGroup == null) {
> ```

You don't want to fail the requisition for a single RPC failure that could be retried. 

The overall strategy needs to be more forgiving. We shouldn't refuse a requisition due to communication failures with the Kingdom. You should retry inline when appropriate, and you could also return a list of requisitions that could be retried some time later and the fetcher could retry them as a batch. It's okay to refuse the requisition if we get an permanent error from the EG service though.
___
*[`src/test/kotlin/org/wfanet/measurement/edpaggregator/requisitionfetcher/AbstractRequisitionGrouperTest.kt` line 371 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2339#-ORsi2TgEbOn_pchNAr4:-ORsi2TgEbOn_pchNAr5:b-3kmbph) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f27cc26b5e17d5d296a26aedfcfcdfb92d982aaf/src/test/kotlin/org/wfanet/measurement/edpaggregator/requisitionfetcher/AbstractRequisitionGrouperTest.kt#L371)):*
> ```Kotlin
>         protocols +=
>           ProtocolConfigKt.protocol {
>             liquidLegionsV2 =
> ```

Should probably use HMSS.

kungfucraig (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2339)* status: all files reviewed, 5 unresolved discussions (waiting on @Marco-Premier and @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/edpaggregator/v1alpha/grouped_requisitions.proto` line 60 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2339#-ORuoN_29dkJ_MA0iKXZ:-ORwmnHW4Hkow7ekCYlO:bcnhjbp) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f27cc26b5e17d5d296a26aedfcfcdfb92d982aaf/src/main/proto/wfa/measurement/edpaggregator/v1alpha/grouped_requisitions.proto#L60)):*

I think our practice has been to use Any in these cases. wdyt @kungfucraig ?
</blockquote></details>

Yeah, we don't want to import the CMMS protos in the EDP Aggregator protos. Ideally you'd take a copy of just the parts you need in the fetcher and use separate EDPA protos throughout the entire pipeline.

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2339)* status: 2 of 14 files reviewed, 4 unresolved discussions (waiting on @kungfucraig and @Marco-Premier)

___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/requisitionfetcher/RequisitionGrouper.kt` line 86 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2339#-ORsdhaq2MdEF_o_021J:-OS1N-Au1o3PvbNDvNxk:be7r32w) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f27cc26b5e17d5d296a26aedfcfcdfb92d982aaf/src/main/kotlin/org/wfanet/measurement/edpaggregator/requisitionfetcher/RequisitionGrouper.kt#L86)):*

Completely agree. see the github issue I put in the in the class comments. We need a real verifier set of utils that are common to the ResultsFulfiller and the RequisitionFetcher/Grouper. Since this needs a refactor at that point anyways, are okay with leaving as-is?
</blockquote></details>

added a simple verifier class
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/requisitionfetcher/RequisitionGrouper.kt` line 161 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2339#-ORsdALAC6Dr_vbcDDzB:-OS-iiNr9srvSD4XEpDv:b-s4r82g) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f27cc26b5e17d5d296a26aedfcfcdfb92d982aaf/src/main/kotlin/org/wfanet/measurement/edpaggregator/requisitionfetcher/RequisitionGrouper.kt#L161)):*

This isn't just about RPC retries. You really need to have a way to deal with the fact the server could just be down for a period of time.

For example the fetcher should probably consider three streams of requisitions one of each for 1) fulfillment; 2) refusal; 3) retry and be able to deal with those accordingly. I would think that means handling refusals in a more robust way too.
</blockquote></details>

this means that the Fetcher was able to fetch requisitions, but seconds later the kingdom was down. Until we have the metadata store, there are two options:
1. Fail
2. Just skip the requisition entirely. It will be fetched again the next time.
I went ahead with the 2nd for now.
___
*[`src/test/kotlin/org/wfanet/measurement/edpaggregator/requisitionfetcher/AbstractRequisitionGrouperTest.kt` line 371 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2339#-ORsi2TgEbOn_pchNAr4:-OS-ckxX5awJkzuklyvQ:bt4m15t) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f27cc26b5e17d5d296a26aedfcfcdfb92d982aaf/src/test/kotlin/org/wfanet/measurement/edpaggregator/requisitionfetcher/AbstractRequisitionGrouperTest.kt#L371)):*

Should probably use HMSS.
</blockquote></details>

just changing this to direct for now.

kungfucraig (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2339)* status: 11 of 14 files reviewed, 9 unresolved discussions (waiting on @Marco-Premier and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/requisitionfetcher/GroupedRequisitionsValidator.kt` line 44 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2339#-OS69zt1-FLfA8f40xOd:-OS6Txek9ZMd4XiEaocM:b-xm65ve) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0f1dcd8528592112e89cacde727783546577f97d/src/main/kotlin/org/wfanet/measurement/edpaggregator/requisitionfetcher/GroupedRequisitionsValidator.kt#L44)):*

You should throw an exception from the validator and have the client do the refusal. That way you decouple what to do when validation fails from the actual validation. It also makes it much easier to test because you can avoid mocking.

I'm sorry to be so difficult about this... We should try to adhere to SOLID (https://en.wikipedia.org/wiki/SOLID), in this case the "S" part of it.

This also means you can turn this class into an object.
</blockquote></details>

Nvm on making it an object. You still need the encryption keys...

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2339)* status: 4 of 14 files reviewed, 8 unresolved discussions (waiting on @kungfucraig and @Marco-Premier)

___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/requisitionfetcher/RequisitionGrouper.kt` line 51 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2339#-OS6AdbMEH2iiccRpNL3:-OSK6p0i-wLftHhNnEFQ:b-bpyea5) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0f1dcd8528592112e89cacde727783546577f97d/src/main/kotlin/org/wfanet/measurement/edpaggregator/requisitionfetcher/RequisitionGrouper.kt#L51)):*

If you follow the suggestions from the validator comments you can drop this argument.
</blockquote></details>

its still needed since you need to pass in the decryption key as you said elsewhere
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/requisitionfetcher/RequisitionGrouper.kt` line 87 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2339#-OS6BYOjBtjbTc6L707t:-OSK8VcZE9Ki_5zE4L3B:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0f1dcd8528592112e89cacde727783546577f97d/src/main/kotlin/org/wfanet/measurement/edpaggregator/requisitionfetcher/RequisitionGrouper.kt#L87)):*

Log the requisition name.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/requisitionfetcher/RequisitionGrouper.kt` line 122 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2339#-OS6Bx8G5xpP4l7B6LyM:-OSK8hi38OmE0z1QRkVp:b-tro1ng) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0f1dcd8528592112e89cacde727783546577f97d/src/main/kotlin/org/wfanet/measurement/edpaggregator/requisitionfetcher/RequisitionGrouper.kt#L122)):*

"collectionIntervals"  or "newCollectionIntervalList" ?

Or can you avoid the temporary and append the collection interval to the list and then call sort on it?
</blockquote></details>

renamed the variable.
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/requisitionfetcher/RequisitionGrouper.kt` line 131 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2339#-OS6CkIH37SWmIrQOLYs:-OSK8lsB8cZINodxOrrn:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0f1dcd8528592112e89cacde727783546577f97d/src/main/kotlin/org/wfanet/measurement/edpaggregator/requisitionfetcher/RequisitionGrouper.kt#L131)):*

I don't think you should use runBlocking like this, but make this a suspend function.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/requisitionfetcher/testing/TestRequisitionData.kt` line 81 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2339#-OS6DMZ15tpLsFSgzBgn:-OSK8rxp6vPXjsAsSjNb:b-7z4kar) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0f1dcd8528592112e89cacde727783546577f97d/src/main/kotlin/org/wfanet/measurement/edpaggregator/requisitionfetcher/testing/TestRequisitionData.kt#L81)):*

I think this should just be "ELGAMAL_PUBLIC_KEY" which is signed by the DUCHY key as part of the consent signaling flow.
</blockquote></details>

i just deleted the llv2 references for now
___
*[`src/main/proto/wfa/measurement/edpaggregator/v1alpha/grouped_requisitions.proto` line 62 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2339#-OS6EAXe535sespo7RCs:-OSK99XFDr_ljr_3yLOE:b-cpsucl) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0f1dcd8528592112e89cacde727783546577f97d/src/main/proto/wfa/measurement/edpaggregator/v1alpha/grouped_requisitions.proto#L62)):*

I only see one event_group in this message. Did you mean to have the collection interval or EG Details?
</blockquote></details>

i forgot to delete this message. It is not used.
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/requisitionfetcher/GroupedRequisitionsValidator.kt` line 43 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2339#-OS69qZI0qPo_opmW-g0:-OSK69i8DDnybQodAmZ-:b-7gmwza) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0f1dcd8528592112e89cacde727783546577f97d/src/main/kotlin/org/wfanet/measurement/edpaggregator/requisitionfetcher/GroupedRequisitionsValidator.kt#L43)):*

Is this really specific to "grouped" requisitions vs. just being a requisition validator?
</blockquote></details>

i named it this because in the follow up PR it adds a function to validated grouped requistions. But I will refactor that slightly to not use that message and instead just process requistision.

Done
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/requisitionfetcher/GroupedRequisitionsValidator.kt` line 44 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2339#-OS69zt1-FLfA8f40xOd:-OSK6RGr-bsafE7n3CTT:b-qgdjlx) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0f1dcd8528592112e89cacde727783546577f97d/src/main/kotlin/org/wfanet/measurement/edpaggregator/requisitionfetcher/GroupedRequisitionsValidator.kt#L44)):*

Nvm on making it an object. You still need the encryption keys...
</blockquote></details>

I added in a lambda that is to be called instead. does that work?

kungfucraig (changes requested):
> Reviewed 10 of 10 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2339)* status: all files reviewed, 3 unresolved discussions (waiting on @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/requisitionfetcher/GroupedRequisitionsValidator.kt` line 44 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2339#-OS69zt1-FLfA8f40xOd:-OSQa-Nm4YA06JjBF-PV:b6ya7a3) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0f1dcd8528592112e89cacde727783546577f97d/src/main/kotlin/org/wfanet/measurement/edpaggregator/requisitionfetcher/GroupedRequisitionsValidator.kt#L44)):*

I added in a lambda that is to be called instead. does that work?
</blockquote></details>

It may be a bit too clever, but let's go with it and see how it ages.
___
*[`src/test/kotlin/org/wfanet/measurement/edpaggregator/requisitionfetcher/RequisitionsValidatorTest.kt` line 99 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2339#-OSQaenrEcbzSIGz3Thf:-OSQaensEzfyqn7tustl:b-e91fry) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c37e818122c6052de2525829b762536c43436e38/src/test/kotlin/org/wfanet/measurement/edpaggregator/requisitionfetcher/RequisitionsValidatorTest.kt#L99)):*
> ```Kotlin
> 
> @RunWith(JUnit4::class)
> class RequisitionsValidatorTest {
> ```

Presumably this is the test that you mentioned needs to be updated still?
___
*[`src/test/kotlin/org/wfanet/measurement/edpaggregator/requisitionfetcher/SingleRequisitionGrouperTest.kt` line 99 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2339#-OSQawsO-OqG4hyRQIy5:-OSQawsO-OqG4hyRQIy6:b-lg01s9) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c37e818122c6052de2525829b762536c43436e38/src/test/kotlin/org/wfanet/measurement/edpaggregator/requisitionfetcher/SingleRequisitionGrouperTest.kt#L99)):*
> ```Kotlin
>   }
> 
>   private fun refuseRequisition(requisition: Requisition, refusal: Requisition.Refusal) =
> ```

I still think that 
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/requisitionfetcher/RequisitionGrouper.kt` line 78 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2339#-OSQb8vS08NkXFndAwyy:-OSQb8vS08NkXFndAwyz:b-s4lmaw) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c37e818122c6052de2525829b762536c43436e38/src/main/kotlin/org/wfanet/measurement/edpaggregator/requisitionfetcher/RequisitionGrouper.kt#L78)):*
> ```Kotlin
>   private suspend fun mapRequisition(requisition: Requisition): GroupedRequisitions? {
> 
>     val measurementSpec: MeasurementSpec =
> ```

So if the requisition is invalid what happens to it? It doesn't look like it gets refused or percolated up to the caller.

My preference is that we would return to the caller a sequence or flow of requisitions to refuse, requisitions to retry, and requisitions to fulfill (i.e. as grouped requisitions).

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2339)* status: all files reviewed, 3 unresolved discussions (waiting on @kungfucraig)

___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/requisitionfetcher/RequisitionGrouper.kt` line 78 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2339#-OSQb8vS08NkXFndAwyy:-OSQnRxMAvdZx0gQLAy0:bg5rwm5) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c37e818122c6052de2525829b762536c43436e38/src/main/kotlin/org/wfanet/measurement/edpaggregator/requisitionfetcher/RequisitionGrouper.kt#L78)):*

So if the requisition is invalid what happens to it? It doesn't look like it gets refused or percolated up to the caller.

My preference is that we would return to the caller a sequence or flow of requisitions to refuse, requisitions to retry, and requisitions to fulfill (i.e. as grouped requisitions).
</blockquote></details>

this is not handled by the RequisitionGrouper but by the RequisitionValidator (actually, its determined by the fatalRequisitionErrorPredicate function)

kungfucraig (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2339)* status: all files reviewed, 3 unresolved discussions (waiting on @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/requisitionfetcher/RequisitionGrouper.kt` line 78 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2339#-OSQb8vS08NkXFndAwyy:-OSRIUmw6dpvaPGUAhFY:b-sgk5ry) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c37e818122c6052de2525829b762536c43436e38/src/main/kotlin/org/wfanet/measurement/edpaggregator/requisitionfetcher/RequisitionGrouper.kt#L78)):*

this is not handled by the RequisitionGrouper but by the RequisitionValidator (actually, its determined by the fatalRequisitionErrorPredicate function)
</blockquote></details>

Great. So you're test just happens to want to call refuse.
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/requisitionfetcher/RequisitionGrouper.kt` line 84 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2339#-OSQsuyHA3WBFm7a2u4d:-OSRJ3zW8eigsFrjAZd8:brhcu0x) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1d382fcea1022cccd66b9c0af86316d625477b43/src/main/kotlin/org/wfanet/measurement/edpaggregator/requisitionfetcher/RequisitionGrouper.kt#L84)):*

would you prefer something more like this @kungfucraig ?
</blockquote></details>

I think I like relying on the callback better. I lost track of the fact that the caller is passing the validator in and can install a callback of its choice.
___
*[`src/test/kotlin/org/wfanet/measurement/edpaggregator/requisitionfetcher/SingleRequisitionGrouperTest.kt` line 99 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2339#-OSQawsO-OqG4hyRQIy5:-OSRJcDvEbB7evQrbKls:b6a8ii3) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c37e818122c6052de2525829b762536c43436e38/src/test/kotlin/org/wfanet/measurement/edpaggregator/requisitionfetcher/SingleRequisitionGrouperTest.kt#L99)):*

what did you mean here?
</blockquote></details>

Sorry, I think that for the purposes of testing using a mock and a refuseCall is unnecessary. You can do something much simpler to test the behavior. You're call though. I don't feel too strongly about it.
___
*[`src/main/proto/wfa/measurement/edpaggregator/v1alpha/grouped_requisitions.proto` line 75 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2339#-OSRJZ9J8zXAiB_-A3se:-OSRJZ9J8zXAiB_-A3sf:b-lnp86x) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1d382fcea1022cccd66b9c0af86316d625477b43/src/main/proto/wfa/measurement/edpaggregator/v1alpha/grouped_requisitions.proto#L75)):*
> ```proto
> 
>   // The validation status of the grouped requisitions
>   message ValidationStatus {
> ```

Per above, let's not have this

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2339)* status: all files reviewed, 2 unresolved discussions (waiting on @kungfucraig)

___
*[`src/main/proto/wfa/measurement/edpaggregator/v1alpha/grouped_requisitions.proto` line 75 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2339#-OSRJZ9J8zXAiB_-A3se:-OSRNRFMD-ViQ_ddv2ZP:bvth56w) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1d382fcea1022cccd66b9c0af86316d625477b43/src/main/proto/wfa/measurement/edpaggregator/v1alpha/grouped_requisitions.proto#L75)):*

Per above, let's not have this
</blockquote></details>

reverted
___
*[`src/test/kotlin/org/wfanet/measurement/edpaggregator/requisitionfetcher/SingleRequisitionGrouperTest.kt` line 99 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2339#-OSQawsO-OqG4hyRQIy5:-OSRRldDC8oknYU3bQvK:b-7cht9f) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c37e818122c6052de2525829b762536c43436e38/src/test/kotlin/org/wfanet/measurement/edpaggregator/requisitionfetcher/SingleRequisitionGrouperTest.kt#L99)):*

Sorry, I think that for the purposes of testing using a mock and a refuseCall is unnecessary. You can do something much simpler to test the behavior. You're call though. I don't feel too strongly about it.
</blockquote></details>

Simplified the tests. I agree they were overly complicated

PR #2355, opened 2025-05-27
fix: Replace LocalComputationIdGenerator with RandomIdGenerator

### LINKED ISSUE #1720: False ALREADY_EXISTS during Computation creation in Duchy Herald
**Describe the bug**
Both aggregator and worker1 showed the LLv2 computation is existing during creation by the herald. However, there is no entry in the database given the global computation id. 
Checking the log, the computation was not created at all. By re-runing the native query in gcp spanner for the reading operation of  ComputationToken, it shows no items as well.

**Steps to reproduce**
Run stress test and there is chance to happen.

**Component(s) affected**
e.g. Kingdom, Duchy, Reporting Server, testing simulators

**Version**
v0.5.7-rc2

**Environment**
QA env

**Additional context**
global computation id: "eGyFuwTOhdk"

kungfucraig:
This PR should fix #1720 which manifested as duplicate ComputationId when inserting a new Computation. Two things are addressed by this PR. 

- We replaced the old ID generator with one that uses a random number. This matches the behavior of the rest of the system.
- We now check for the newly generated ID before inserting the new Computation and continue to generate IDs until one is found that does not exist.


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2355)

SanjayVas (changes requested):
> Reviewed 4 of 6 files at r1, 3 of 4 files at r2, 4 of 4 files at r3, 2 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2355)* status: all files reviewed, 2 unresolved discussions (waiting on @renjiezh)

___
*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2355#-ORJAltY9F9P8fsyx8v_:-ORJAltY9F9P8fsyx8va:b-5f09nt) (no related file):*
Maybe explain what was wrong in the PR description?
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/gcloud/spanner/computation/GcpSpannerComputationsDatabaseTransactor.kt` line 81 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2355#-ORJAITi2nJNNwTIySjc:-ORJAITi2nJNNwTIySjd:b2adcwk) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/14acbdacf231d2c290b62328a968dd01a78d676f/src/main/kotlin/org/wfanet/measurement/duchy/deploy/gcloud/spanner/computation/GcpSpannerComputationsDatabaseTransactor.kt#L81)):*
> ```Kotlin
> 
>     val localComputationId: Long =
>       computationIdGenerator.generateNewId { id ->
> ```

I see you found my function for ensuring unique IDs :)



_Code quote:_
```Kotlin
generateNewId
```
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/gcloud/spanner/computation/GcpSpannerComputationsDatabaseTransactor.kt` line 83 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2355#-ORJ9o8zE5ayJ1AAsvJA:-ORJ9o8zE5ayJ1AAsvJB:bnc3gx) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/14acbdacf231d2c290b62328a968dd01a78d676f/src/main/kotlin/org/wfanet/measurement/duchy/deploy/gcloud/spanner/computation/GcpSpannerComputationsDatabaseTransactor.kt#L83)):*
> ```Kotlin
>       computationIdGenerator.generateNewId { id ->
>         databaseClient
>           .singleUseReadOnlyTransaction()
> ```

I don't think we want to create a separate transaction for this. Instead, we want to do it in the same transaction as the write. This means using AsyncDatabaseClient.readWriteTransaction instead of calling AsyncDatabaseClient.write, as that creates and executes its own transaction.

kungfucraig (commented):
> Reviewed 6 of 6 files at r1, 4 of 4 files at r2, 3 of 4 files at r3, 2 of 2 files at r4, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2355)* status: all files reviewed, 1 unresolved discussion (waiting on @renjiezh and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/gcloud/spanner/computation/GcpSpannerComputationsDatabaseTransactor.kt` line 81 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2355#-ORJAITi2nJNNwTIySjc:-ORJDx9E-3H09Gk_UYiS:bo6ap9z) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/14acbdacf231d2c290b62328a968dd01a78d676f/src/main/kotlin/org/wfanet/measurement/duchy/deploy/gcloud/spanner/computation/GcpSpannerComputationsDatabaseTransactor.kt#L81)):*

I see you found my function for ensuring unique IDs :)
</blockquote></details>

Indeed!
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/gcloud/spanner/computation/GcpSpannerComputationsDatabaseTransactor.kt` line 83 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2355#-ORJ9o8zE5ayJ1AAsvJA:-ORJGgg25zkQlHDwl2xF:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/14acbdacf231d2c290b62328a968dd01a78d676f/src/main/kotlin/org/wfanet/measurement/duchy/deploy/gcloud/spanner/computation/GcpSpannerComputationsDatabaseTransactor.kt#L83)):*

I don't think we want to create a separate transaction for this. Instead, we want to do it in the same transaction as the write. This means using AsyncDatabaseClient.readWriteTransaction instead of calling AsyncDatabaseClient.write, as that creates and executes its own transaction.
</blockquote></details>

Done.

PR #2381, opened 2025-06-04
refactor: Switch to Committed Write Stream with Offsets for the Operational Metrics Export

tristanvuong2021:
Currently, the Operational Metrics Export uses the BigQuery default stream, but that does not have any handling for duplicates. To prevent duplicates, an application-created stream must be used to make use of offsets.


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2381)

PR #2391, opened 2025-06-06
Add private network

georgi:


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2391)

Marco-Premier (approved):
New VPC would require a forwarding rule that is not supported only globally using "all-apis". Switching over regional one would require google-beta provider which implies unwanted changes on the terraform-cmms.yml workflow.

So this PR only adds a new private subnetwork for the MIG to access privately google services and an NAT for outbound traffic and access the kingdom api.

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2391#-:-O_XDWjW2_YRDqFxNk3u:buh04ql)* status: 0 of 11 files reviewed, all discussions resolved (waiting on @kungfucraig)

Marco-Premier (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2391)* status: 5 of 11 files reviewed, 2 unresolved discussions (waiting on @SanjayVas)

___
*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2391#-O_YJUt75Y-4wDsj3I8y:-O_kUFnj-dtGc9A9hW4f:b-y8nbp5) (no related file):*

Is there any way to do this with IPv6 instead? Depending on the configuration, it may avoid some unnecessary NAT.
</blockquote></details>

I don't think that's need as **Private Google Access (PGA)**  API calls IPv4 already bypasses Cloud NAT The nat in this case is only needed to reach the kingdom as the VM has no public IP. 
___
*[`src/main/terraform/gcloud/modules/edp-aggregator/main.tf` line 20 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2391#-O_YJI4R0EsIcfZZxZu8:-O_kWvCv9f1fDPXjksGb:bb6un1j) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f220b31e87a16c5989c7c489239ef8bc2516234f/src/main/terraform/gcloud/modules/edp-aggregator/main.tf#L20)):*

Document where they came from
</blockquote></details>

Done

PR #2447, opened 2025-06-16
feat!: add Cloud Scheduler for requisition fetcher with 1-minute interval

georgi:
Issue: #2398 
BREAKING-CHANGE: EDPA deployment requires a new Terraform module called "cloud-scheduler" for automatically launching the RequisitionFetcher Cloud Run Function.

Integrate Cloud Scheduler to automatically trigger the requisition fetcher Cloud Function every 1 minute. The scheduler uses OIDC authentication with a dedicated service account and dynamically constructs the function URL based on the current GCP project and region.

- Add requisition_fetcher_scheduler_config to edp-aggregator module
- Create Cloud Scheduler job with 1-minute cron schedule
- Create dedicated service account for scheduler authentication
- Grant cloudfunctions.invoker role to scheduler service account
- Configure scheduler to always deploy (non-optional)


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2447)

georgi (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2447)* status: 2 of 3 files reviewed, 1 unresolved discussion (waiting on @Marco-Premier)

___
*[`src/main/terraform/gcloud/cmms/edp_aggregator.tf` line 112 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2447#-OStKTP2DLruA42TsJoM:-OSxXo7v1HrAfpVg2X8w:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/99e865b7d427928e46d9a35b06ce8e03e49a845b/src/main/terraform/gcloud/cmms/edp_aggregator.tf#L112)):*

I think Cloud Scheduler only supports 5-field cron syntax.  Meaning it does not support per-second scheduling.
Also a second thought here, 5 seconds may be to stringent anyway. 
Can you please start with  1 minute interval between invocation?
</blockquote></details>

Done.

PR #2465, opened 2025-06-18
fix: wrong secret name

Marco-Premier:


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2465)

PR #2478, opened 2025-06-25
feat: Implement Validation for CreateBasicReport

tristanvuong2021:
This only adds validation. The whole method is not implemented. This helps break up the full implementation.


SanjayVas (changes requested):
> Reviewed 9 of 12 files at r1, 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2478)* status: all files reviewed, 1 unresolved discussion (waiting on @kungfucraig)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/BasicReportsService.kt` line 274 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2478#-OTd6tbV-5FDIR7iCgh_:-OTd6tbV-5FDIR7iCgha:b-zfaze) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0a74514b18b6db7d6bfcb4ab349af66fede7d4c4/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/BasicReportsService.kt#L274)):*
> ```Kotlin
>           .asStatusRuntimeException(Status.Code.INVALID_ARGUMENT)
> 
>     val campaignGroup: ReportingSet =
> ```

Extract out the campaign group get rather than returning it from the validate method.

tristanvuong2021 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2478)* status: 11 of 12 files reviewed, 1 unresolved discussion (waiting on @kungfucraig and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/BasicReportsService.kt` line 274 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2478#-OTd6tbV-5FDIR7iCgh_:-OTdGVLzDKke5l1g3YxR:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/660664152ea5175c46c5047799e06ba8f51fdabf/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/BasicReportsService.kt#L274)):*

Extract out the campaign group get rather than returning it from the validate method.
</blockquote></details>

Done.

tristanvuong2021 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2478)* status: 9 of 13 files reviewed, 16 unresolved discussions (waiting on @kungfucraig and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/BasicReportsService.kt` line 112 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2478#-OTdSeWfEGRUwUt7LlJV:-OTiDEiR456krvjM8Awc:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6a163a4785e3b6f7e4f04c5cb849a765a34fddbd/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/BasicReportsService.kt#L112)):*

This structure is awkward if the when expression doesn't always return a status. The newer structure is to parse the error code. See how the public ReportingSetsService does it.

Note that this is why I said things get messy if you have a method start mixing the older ReportingInternalException with the newer ServiceException.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/BasicReportsService.kt` line 115 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2478#-OTdRqgL47_Kwpj_Ve0i:-OThdNJK0JGjvlClhA2R:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6a163a4785e3b6f7e4f04c5cb849a765a34fddbd/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/BasicReportsService.kt#L115)):*

This would be a FAILED_PRECONDITION which should not use InvalidFieldValueException. Instead, we use specific NotFoundException types.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/BasicReportsService.kt` line 119 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2478#-OTdR6i2CHTTbkulmQzE:-OThbMAlDVFJ3zcngrdo:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6a163a4785e3b6f7e4f04c5cb849a765a34fddbd/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/BasicReportsService.kt#L119)):*

We just use INTERNAL now for anything unhandled from the backend. It should also be an exhaustive when on the error code extraction.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/BasicReportsService.kt` line 99 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2478#-OThs8ww6XvEvkEozAhB:-OTiGoCQ4rSUbEo6ro8b:b-2rwq93) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/544a46d9d956f08a1034303902520d76fef8fe6e/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/BasicReportsService.kt#L99)):*

put this in a function: "validateAndGetCampaignGroup"
</blockquote></details>

It was one function originally, but it was changed to separate validation and other actions.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/BasicReportsService.kt` line 303 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2478#-OTht-zR-cZ9suAIj7mm:-OTiIDtpER68RECwakcG:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/544a46d9d956f08a1034303902520d76fef8fe6e/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/BasicReportsService.kt#L303)):*

Can you put the code for validating the reporting interval into its own function?
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/BasicReportsService.kt` line 326 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2478#-OThtKE1BCxHmMNBPZBf:-OTiHSid8PQLWS8Wzn38:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/544a46d9d956f08a1034303902520d76fef8fe6e/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/BasicReportsService.kt#L326)):*

Describe in the message in detail what check(s) failed.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/BasicReportsService.kt` line 335 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2478#-OThta3Z1WQLqfg8e7cF:-OTiHTHP77jLD5T1g639:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/544a46d9d956f08a1034303902520d76fef8fe6e/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/BasicReportsService.kt#L335)):*

Add details to the error message.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/BasicReportsService.kt` line 338 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2478#-OThtebv2FEg4Cb5mjGu:-OTiL7U6AbgeS9HMttSq:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/544a46d9d956f08a1034303902520d76fef8fe6e/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/BasicReportsService.kt#L338)):*

Add a function called "validateImpressionQualificationFilters"
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/BasicReportsService.kt` line 360 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2478#-OThuSswCoJiHQUQ2ZDt:-OTiKW5M6VzplqFk9wjM:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/544a46d9d956f08a1034303902520d76fef8fe6e/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/BasicReportsService.kt#L360)):*

Add a function called "validateCustomImpressionQualificationFilterSpec"
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/BasicReportsService.kt` line 367 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2478#-OThu7d6CwBS7rfWjZ5a:-OTiJBP0EvLZZdR-COg6:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/544a46d9d956f08a1034303902520d76fef8fe6e/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/BasicReportsService.kt#L367)):*

You should say which media type(s) it is.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/BasicReportsService.kt` line 388 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2478#-OThusPk9nlmhv85z6Hk:-OTiIfB84cKWPjNPTzak:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/544a46d9d956f08a1034303902520d76fef8fe6e/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/BasicReportsService.kt#L388)):*

These need to be checked against the actual event template, which should be read as configuration into the server. I'd be okay leaving that as a TODO for now.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/BasicReportsService.kt` line 415 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2478#-OThvW4n2wecQiRhLtSl:-OTiLr8l6J3ucSun2eTu:bgkhhky) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/544a46d9d956f08a1034303902520d76fef8fe6e/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/BasicReportsService.kt#L415)):*

Add a function to validate the result group list and another function to validate each individual result group.
</blockquote></details>

There isn't enough to separate this way, but I did create functions to separate it differently
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/BasicReportsService.kt` line 433 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2478#-OThvuecDpnTJdg_0YM5:-OTiOGdNALHWVSw3YBX0:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/544a46d9d956f08a1034303902520d76fef8fe6e/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/BasicReportsService.kt#L433)):*

Add a function to validate a reporting unit
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/BasicReportsService.kt` line 480 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2478#-OThw2D51SkBQ2LkQU7l:-OTiIgV10G6hltr_wvrh:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/544a46d9d956f08a1034303902520d76fef8fe6e/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/BasicReportsService.kt#L480)):*

These need to be check against the event template and we need to ensure the proper ET annotations are set. I'll send a PR on Test Event as soon as I can. As with the above, I think a TODO is fine here.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/BasicReportsService.kt` line 544 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2478#-OThwci_3whENCQo6WH_:-OTiGKHp8W5l4tZyOTyH:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/544a46d9d956f08a1034303902520d76fef8fe6e/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/BasicReportsService.kt#L544)):*

Add a function for validating the metric spec. That should allow you to call it multiple times instead of repeating nearly identical blocks of code.
</blockquote></details>

Done.

tristanvuong2021 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2478)* status: 10 of 13 files reviewed, 17 unresolved discussions (waiting on @kungfucraig and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/Errors.kt` line 33 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2478#-OTib_OF8bsRxNOabpoN:-OTipkMAEHHIFEdx7Sra:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/9db17f9f6fca8abc4c278dc5decc2f8e7c75e754/src/main/kotlin/org/wfanet/measurement/reporting/service/api/Errors.kt#L33)):*

Technically this is just a ReportingSet not being found. The only reason that there's a separate one for CAMPAIGN_GROUP_INVALID is because it's specifically an error indicating that a ReportingSet is not a valid Campaign Group.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/BasicReportsService.kt` line 177 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2478#-OTidiVq3TDoqmWcjzVA:-OTiplyj6LIDIXMALvRi:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/9db17f9f6fca8abc4c278dc5decc2f8e7c75e754/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/BasicReportsService.kt#L177)):*

Same here re: it should be its own not found exception rather than InvalidFieldValueException and be a FAILED_PRECONDITION
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/BasicReportsService.kt` line 460 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2478#-OTiexhk8-_makSPIh2P:-OTioP6u85Lg2VZ5DSZ1:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/9db17f9f6fca8abc4c278dc5decc2f8e7c75e754/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/BasicReportsService.kt#L460)):*

You **must** include the field name in the message. Otherwise you're using the wrong exception type.

Note also that *all* variables in any error info message must have entries in the metadata map.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/BasicReportsService.kt` line 461 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2478#-OTie6bu1VElDSRSiRWm:-OTioOcj5dIM-RnOV9Y2:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/9db17f9f6fca8abc4c278dc5decc2f8e7c75e754/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/BasicReportsService.kt#L461)):*

INVALID_ARGUMENT means the request is *always* invalid regardless of state, meaning that if you were to try the same request at a later time it could *never* work.
</blockquote></details>

Done.

SanjayVas (approved):
> Reviewed 2 of 3 files at r6, 1 of 1 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2478)* status: all files reviewed, 14 unresolved discussions (waiting on @kungfucraig and @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/Errors.kt` line 107 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2478#-OTmRQfFCnuI8s_tJ7TA:-OTmRQfFCnuI8s_tJ7TB:b2ni8l0) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/78879cde7e95da0601223eb46dd672cce71e3926/src/main/kotlin/org/wfanet/measurement/reporting/service/api/Errors.kt#L107)):*
> ```Kotlin
>   ServiceException(
>     Errors.Reason.REPORTING_SET_NOT_FOUND,
>     "Reporting Set $name not found",
> ```

nit



_Suggestion:_
```Kotlin
ReportingSet
```

tristanvuong2021 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2478)* status: 12 of 13 files reviewed, 13 unresolved discussions (waiting on @kungfucraig and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/BasicReportsService.kt` line 76 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2478#-OThehc9CV-BxZFt4WCS:-OTmToT30BvLDVod-d5d:bgwfd0o) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/544a46d9d956f08a1034303902520d76fef8fe6e/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/BasicReportsService.kt#L76)):*

What do you think of moving most/all of the validation logic into a separate BasicReportValidator file?
</blockquote></details>

I think for now I will leave it all in this file because the validation also throws gRPC exceptions so I would have to separate that from the validation file, but if the file ends up really big later, then I will reconsider.

kungfucraig (changes requested):
> Reviewed 1 of 4 files at r5, 1 of 3 files at r6, 1 of 1 files at r7, 1 of 1 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2478)* status: all files reviewed, 3 unresolved discussions (waiting on @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/BasicReportsService.kt` line 76 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2478#-OThehc9CV-BxZFt4WCS:-OTnfyNf4dP9IQaUfwBd:brn5vdq) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/544a46d9d956f08a1034303902520d76fef8fe6e/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/BasicReportsService.kt#L76)):*

I think for now I will leave it all in this file because the validation also throws gRPC exceptions so I would have to separate that from the validation file, but if the file ends up really big later, then I will reconsider.
</blockquote></details>

I've found the reporting server code pretty hard to read because of file sizes and lack of modularization. I can see how this file is going to get huge. Consider that MetricsService is 3000+ lines of code... 

I'm not opposed to doing this later, but it'll never be easier to do it than it is right now.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/BasicReportsService.kt` line 99 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2478#-OThs8ww6XvEvkEozAhB:-OTngoOABtFJ8OpCUTf0:b-g0ao92) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/544a46d9d956f08a1034303902520d76fef8fe6e/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/BasicReportsService.kt#L99)):*

It was one function originally, but it was changed to separate validation and other actions.
</blockquote></details>

What I was suggesting is that lines 110-149 be factored out.

More generally, what I'm objecting to is functions that require a lot of scrolling to read and that are performing more than a single task.

Each block of code here is more or less independent and that independence should be expressed as a decomposition into smaller units. 

It really makes things a lot easier to read, test, and reason about.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/BasicReportsService.kt` line 415 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2478#-OThvW4n2wecQiRhLtSl:-OTnl42e4X0otra86fcm:brxp9sd) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/544a46d9d956f08a1034303902520d76fef8fe6e/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/BasicReportsService.kt#L415)):*

There isn't enough to separate this way, but I did create functions to separate it differently
</blockquote></details>

Sounds good!
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/BasicReportsService.kt` line 86 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2478#-OTni17P9cnRJUBqO_SO:-OTni17QCycP6ZrO0FFf:bcu88c6) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/305c6798f711032f8dbf6b879175df3469457db4/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/BasicReportsService.kt#L86)):*
> ```Kotlin
> ) : BasicReportsCoroutineImplBase(coroutineContext) {
> 
>   override suspend fun createBasicReport(request: CreateBasicReportRequest): Operation {
> ```

Add kdoc to the functions. Even just to write:

/**
 * Creates a BasicReport
*/

Or 

/**
 * Validates an X
 */

it provides a nice visual cue when browsing the code.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/BasicReportsService.kt` line 300 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2478#-OTnk14w6h30VOjxFYNA:-OTnk14w6h30VOjxFYNB:bheb825) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/305c6798f711032f8dbf6b879175df3469457db4/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/BasicReportsService.kt#L300)):*
> ```Kotlin
>    * @throws [StatusRuntimeException] when validation fails
>    */
>   private fun validateCreateBasicReportRequest(
> ```

I can live with it, but I think the extension functions make this harder to read vs. functions named "validateX(x: X)"

kungfucraig (approved):
> Reviewed 5 of 5 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2478)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @tristanvuong2021)

PR #2480, opened 2025-06-25
fix!: Make DimensionSpecSummary grouping repeated

tristanvuong2021:
BREAKING CHANGE!

There can be more than 1 so the field needs to represent a list. The field is now repeated.


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2480)

Issue #2470, opened 2025-06-19
SyntheticDataGeneration.generateEvents returns Flow despite no coroutine usage


SanjayVas:
#2290 incorrectly changed SyntheticDataGeneration to return a Flow rather than a Sequence. There are no coroutines used in event generation. Therefore returning a Sequence is correct, as it allows the function to be used by non-coroutine code. Indeed, this change also introduced a `runBlocking` to call it from SyntheticGeneratorEventQuery. `runBlocking` should only be called from the program entry point (`main` or unit test methods).

For coroutine code that operates on Flows, it's trivial to convert a Sequence to a Flow using the `.asFlow()` extension function. See https://kotlinlang.org/docs/flow.html#flow-builders

PR #2487, opened 2025-06-27
fix: Remove runBlocking call from MeasurementResults

SanjayVas:
It is incorrect to call runBlocking anywhere but the single transition point into coroutine code. For example, `main` or a test method.

Issue: #2470


stevenwarejones (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2487)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @SanjayVas)

PR #2538, opened 2025-07-01
feature: Add CampaignGroup filter to Internal Reporting Sets Service Stream Method

tristanvuong2021:
This is useful for MC API Phase 2 to look up relevant Reporting Sets.


tristanvuong2021 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2538)* status: 4 of 5 files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

___
*[`src/main/proto/wfa/measurement/internal/reporting/v2/reporting_sets_service.proto` line 58 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2538#-OUBn2WP6lyJr4tvqy8Q:-OUkDQpLAnihF-j_U1-0:bdangwc) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f8ef6d9c1b91b637c9c7fba346280ab3b2607f64/src/main/proto/wfa/measurement/internal/reporting/v2/reporting_sets_service.proto#L58)):*

Don't you need to add this to the page token too (assuming you're checking previous request params)?
</blockquote></details>

This is only updating the internal method. I only need the internal method updated. If the public method is also updated later, then the page token will be updated then.

PR #2551, opened 2025-07-08
feat: add check to skip empty file processing in DataWatcher

Marco-Premier:
Prevent DataWatcher from being triggered multiple times for the same file.

### DIFF
--- a/src/main/kotlin/org/wfanet/measurement/securecomputation/deploy/gcloud/datawatcher/DataWatcherFunction.kt
+++ b/src/main/kotlin/org/wfanet/measurement/securecomputation/deploy/gcloud/datawatcher/DataWatcherFunction.kt
@@ -50,10 +50,16 @@ class DataWatcherFunction : CloudEventsFunction {
       StorageObjectData.newBuilder()
         .apply { JsonFormat.parser().merge(cloudEventData, this) }
         .build()
+
     val blobKey: String = data.getName()
     val bucket: String = data.getBucket()
     val path = "$scheme://$bucket/$blobKey"
     logger.info("Receiving path $path")
+    val size = data.size
+    if (size == 0L) {
+      logger.info("Skipping processing: file '${path}' is empty")
+      return
+    }
     runBlocking { dataWatcher.receivePath(path) }
   }
 


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2551)

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2551)* status: all files reviewed, 1 unresolved discussion (waiting on @Marco-Premier)

___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/deploy/gcloud/datawatcher/DataWatcherFunction.kt` line 59 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2551#-OUf3RO9DZ8lSJyDimOO:-OUfdzEg9zae3RrCjLVg:b-scf6bj) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/429ee22d7c273f1702a3705f85b13d298f74af33/src/main/kotlin/org/wfanet/measurement/securecomputation/deploy/gcloud/datawatcher/DataWatcherFunction.kt#L59)):*

Can you file an issue and leave a TODO indicating we need monitoring for this. I think if we are seeing a lot of empty files it could indicate some kind of upstream bug.
</blockquote></details>

what if a ds has zero impressions? I would prefer an empty file be uploaded for data availability purposes. There must be a better way of doing this.

Marco-Premier (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2551)* status: all files reviewed, 2 unresolved discussions (waiting on @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/deploy/gcloud/datawatcher/DataWatcherFunction.kt` line 61 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2551#-OUjmNuT13ckF5EEIaJk:-OUpDoq96F5cB2T8uosS:b-pfe8dw) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/429ee22d7c273f1702a3705f85b13d298f74af33/src/main/kotlin/org/wfanet/measurement/securecomputation/deploy/gcloud/datawatcher/DataWatcherFunction.kt#L61)):*

can you add a TODO and a github issue to instead of returning on empty files to make sure the metadata store updates that there is availability on this date?
</blockquote></details>

Yes I can definitely do that. Are you against having this workaround merged till we have the metadata store? Cause it causes problems with the cloud test where multiple TEE apps "compete" to fulfill same requisitions

Marco-Premier (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2551)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Marco-Premier)

___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/deploy/gcloud/datawatcher/DataWatcherFunction.kt` line 61 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2551#-OUjmNuT13ckF5EEIaJk:-OV8sXag4mCkiciokY7s:bbmuikd) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/429ee22d7c273f1702a3705f85b13d298f74af33/src/main/kotlin/org/wfanet/measurement/securecomputation/deploy/gcloud/datawatcher/DataWatcherFunction.kt#L61)):*

just do the standard open a github issue, track it
</blockquote></details>

#2568

Issue #2211, opened 2025-04-22
Support for signed builds to support secure compute attestation.


mikkokotila:
See: https://codelabs.developers.google.com/signed-container-image-codelab#0 

This is necessary to have a reasonable CI/CD process with our secure compute images.

Signed images will allow EDPs to provide KMS keys that are permissioned to images that are built with the Halo build process.

- This means that they are basically agreeing to trust the Halo development process.
- It may require that we implement OWNERS file in GitHub and allow EDPs to proactively review code as it is coming in. This is optional as they could simply choose to trust the existing development process instead.
- If we have to expand the necessary set of reviews for secure compute use cases we would be looking at doing this only for select parts of the repo.
- This is one of several reasons to keep the code that runs in the secure compute environments of minimal size. 

Without signed images, CI/CD will be nearly impossible since we’ll need to gather permissions from each EDP for each version of each secure compute image.

@renjiezh can advise on the KMS integration. He may also have a demo of how to do signed builds already.
@SanjayVas  can advise how best to manage the key for signing the builds.

The design should consider key rotation.

Thanks @kungfucraig for putting this together.

renjiezh:
Thank you @mikkokotila for the writing! 

Yes, we are planing to have the signed image for the TEE implementation. It will be covered in my implementation design doc. I'd like to leave some comments regarding the current ideas.

1. Signed Image vs Image Digest
EDPs need a way to verify that the image running in the TEE env can be trusted, either verify image digest or image signature.
However, image digest requires EDP operators to update/add the digest for every image release. It will definitely delay the delivery for new releases given the requirement that all EDPs need extra effort to allow the new release.
Thus, signed image allow EDPs automatically trust new releases as long as it is signed by a trusted key. Key rotation is also easy to achieve by adding trusted key instead of replacing trusted key during the transition.

2. Github OWNERS or not
Adding OWNERS with special process for code review of secure compute could help to guarantee the integrity. However, it will inevitably introduce extra effort to review the code and delay the development and delivery process. Another shortcoming is that, the code running in TEE might depend on other CMM libraries, e.g. `common-jvm`. It implies that any code change to `common-jvm` should also be treated as secure compute related code requiring stricter reviews.
The alternative is to agree to trust the entire CMM project as all code of it requires at least two parties to approve to merge. Besides, Github has all history of code changes for audit purpose. 
We don't have a final solution regarding this. But we will discuss to find the most suitable choice.

[...]

PR #2552, opened 2025-07-08
ci: Add workflow for image signing

laureanobrs:
Issue: #2211


SanjayVas (changes requested):
> Reviewed 1 of 8 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2552)* status: 1 of 8 files reviewed, 12 unresolved discussions (waiting on @laureanobrs)

___
*[`.github/workflows/maybe-push-images.yml` line 80 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2552#-OWMsdJ85I4BCMF4FWyx:-OWMsdJ85I4BCMF4FWyy:b-ys5vi5) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/fc4e0de49422c9dd416aa14fb5fc1d0a7d955114/.github/workflows/maybe-push-images.yml#L80)):*
> ```YAML
>     uses: ./.github/workflows/sign-images.yml
>     needs: [get-image-tag, push-images]
>     if: ${{ github.event_name == 'release' }}
> ```

I don't think the event name will ever be `release` here, since the only declared events for this workflow are `workflow_dispatch` and `workflow_call`.
___
*[`src/main/docker/sign_images.sh.template` line 15 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2552#-OWMt39S2DbjbKwLX1Ox:-OWMt39S2DbjbKwLX1Oy:b15053p) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/fc4e0de49422c9dd416aa14fb5fc1d0a7d955114/src/main/docker/sign_images.sh.template#L15)):*
> ```
> 
> # install cosign
> COSIGN_VERSION="2.5.0"
> ```

This script shouldn't install cosign. You want to do that from a repo rule referenced by the MODULE. For something like this, you would generally define a whole `rules_cosign` module.

Is there a reason you can't use the `cosign_sign` rule from rules_oci? The comment in the README saying it's not part of the public API is inaccurate, as it's available at https://github.com/bazel-contrib/rules_oci/blob/main/cosign/defs.bzl

laureanobrs (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2552)* status: 1 of 8 files reviewed, 12 unresolved discussions (waiting on @SanjayVas)

___
*[`.github/workflows/maybe-push-images.yml` line 80 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2552#-OWMsdJ85I4BCMF4FWyx:-OWNDHXs-2MXuC_IkaiW:bko55ur) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/fc4e0de49422c9dd416aa14fb5fc1d0a7d955114/.github/workflows/maybe-push-images.yml#L80)):*

I don't think the event name will ever be `release` here, since the only declared events for this workflow are `workflow_dispatch` and `workflow_call`.
</blockquote></details>

I think it will. From the documentation:
https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#workflow_call
"When a workflow is triggered with the `workflow_call` event, the event payload in the called workflow is the same event payload from the calling workflow."

I created a sample repo to test that:
https://github.com/laureanobrs/test-workflow-events/actions
1 - "print-event" workflow dispatched, does not run the job conditional to the release event
2 - "print-event" workflow called from other workflow on release, actually does run that job

Does it make sense?
___
*[`src/main/docker/sign_images.sh.template` line 15 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2552#-OWMt39S2DbjbKwLX1Ox:-OWNEv4kD0kYG3vuXMNy:bpd2izr) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/fc4e0de49422c9dd416aa14fb5fc1d0a7d955114/src/main/docker/sign_images.sh.template#L15)):*
> define a whole `rules_cosign` module

Just in case you have it handy — do you have an example of how to do that? (So I can keep it in mind for a future case)

> Is there a reason you can't use the `cosign_sign` rule from rules\_oci?

I followed Renjie’s doc, which talked about installing Cosign, and when I saw that note in rules_oci, I decided not to try it. But I’ll give it a try now.

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2552)* status: 1 of 8 files reviewed, 11 unresolved discussions (waiting on @laureanobrs)

___
*[`.github/workflows/maybe-push-images.yml` line 80 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2552#-OWMsdJ85I4BCMF4FWyx:-OWNIkis4dthD56g_xw6:bqvsw68) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/fc4e0de49422c9dd416aa14fb5fc1d0a7d955114/.github/workflows/maybe-push-images.yml#L80)):*

I think it will. From the documentation:
https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#workflow_call
"When a workflow is triggered with the `workflow_call` event, the event payload in the called workflow is the same event payload from the calling workflow."

I created a sample repo to test that:
https://github.com/laureanobrs/test-workflow-events/actions
1 - "print-event" workflow dispatched, does not run the job conditional to the release event
2 - "print-event" workflow called from other workflow on release, actually does run that job

Does it make sense?
</blockquote></details>

Huh, I'm kind of surprised. Thanks for doing the experiment.
___
*[`src/main/docker/sign_images.sh.template` line 15 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2552#-OWMt39S2DbjbKwLX1Ox:-OWNIwTlAqlz6QxL1X6Z:bb2c6on) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/fc4e0de49422c9dd416aa14fb5fc1d0a7d955114/src/main/docker/sign_images.sh.template#L15)):*
> Just in case you have it handy — do you have an example of how to do that?

Our own [rules_cue](https://github.com/world-federation-of-advertisers/rules_cue) would be an example. It's only a bit complicated because it's supporting MacOS Darwin as well, which we don't really need to do. (There was a point in time where we were trying to support developing on MacOS, which we gave up on).

Taking another look at this, you could actually go a bit simpler than creating a whole module and just use the `http_file` repo rule from `@bazel_tools//tools/build_defs/repo:http.bzl`.

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2552)* status: 1 of 8 files reviewed, 11 unresolved discussions (waiting on @laureanobrs)

___
*[`src/main/docker/sign_images.sh.template` line 15 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2552#-OWMt39S2DbjbKwLX1Ox:-OWNKNkJ7fizn3Y8kARC:b-f6rgtw) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/fc4e0de49422c9dd416aa14fb5fc1d0a7d955114/src/main/docker/sign_images.sh.template#L15)):*

> Just in case you have it handy — do you have an example of how to do that?

Our own [rules_cue](https://github.com/world-federation-of-advertisers/rules_cue) would be an example. It's only a bit complicated because it's supporting MacOS Darwin as well, which we don't really need to do. (There was a point in time where we were trying to support developing on MacOS, which we gave up on).

Taking another look at this, you could actually go a bit simpler than creating a whole module and just use the `http_file` repo rule from `@bazel_tools//tools/build_defs/repo:http.bzl`.
</blockquote></details>

FWIW, the proper way of doing it would be to create a custom toolchain, but that's a real pain when you only care about Linux on amd64.

SanjayVas (changes requested):
> Reviewed 1 of 3 files at r3, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2552)* status: 3 of 9 files reviewed, 8 unresolved discussions (waiting on @Copilot and @laureanobrs)

___
*[`src/main/docker/sign_images.bzl` line 15 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2552#-OWaXHUB5DZpUwouX_NT:-OWaXHUC1XWjC02RAGMy:b-z2usyo) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6a5f1a8d141cefb4867ea53e76b280379b62212f/src/main/docker/sign_images.bzl#L15)):*
> ```Starlark
> 
> """ Image signing scripts """
> ```

Move this file somewhere under the //build tree
___
*[`src/main/docker/sign_images.bzl` line 18 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2552#-OWaXOL_38-98KO3quMV:-OWaXOL_38-98KO3quMW:btsh21v) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6a5f1a8d141cefb4867ea53e76b280379b62212f/src/main/docker/sign_images.bzl#L18)):*
> ```Starlark
> 
> load("//build:variables.bzl", "IMAGE_REPOSITORY_SETTINGS")
> load("//src/main/docker/panel_exchange_client:images.bzl", "ALL_IMAGES")
> ```

Don't reference the images themselves here. They should be taken as rule attributes.

Basically this rule should still be written as if it were in a separate module. Referencing IMAGE_REPOSITORY_SETTINGS is okay since those are just string constants.
___
*[`src/main/docker/sign_images.bzl` line 59 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2552#-OWaXpRY1OKrnTrYst4-:-OWaXpRY1OKrnTrYst40:b-ymp4nw) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6a5f1a8d141cefb4867ea53e76b280379b62212f/src/main/docker/sign_images.bzl#L59)):*
> ```Starlark
>             allow_single_file = True,
>         ),
>         "cosign": attr.label(
> ```

The caller shouldn't be able to override the executable.



_Suggestion:_
```Starlark
_cosign
```
___
*[`WORKSPACE` line 2 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2552#-OWaWSqL9KKNBd_eaGFt:-OWaWSqL9KKNBd_eaGFu:b7laf4h) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6a5f1a8d141cefb4867ea53e76b280379b62212f/WORKSPACE#L2)):*
> ```Starlark
> workspace(name = "wfa_measurement_system")
> ```

This goes into the MODULE.bazel file, not the WORKSPACE. WORKSPACE is deprecated and will be dropped when we upgrade to Bazel 8. Instead of the load call, you use 

```Starlark
http_file = use_repo_rule("@bazel_tools//:http.bzl", "http_file")
```

laureanobrs (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2552)* status: 0 of 15 files reviewed, 2 unresolved discussions (waiting on @SanjayVas)

___
*[`build/cosign/sign_images.bzl` line 62 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2552#-OWbUClo-HnpOJk8a90a:-OWv4AmI8WksNAg_Mehp:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4c02d12a05c3b2d2b587a28b6cde8b8f982c34ee/build/cosign/sign_images.bzl#L62)):*

This should be the label of the underlying oci_image target so you can grab the digest without having to call `docker` in your script. This means that you'll need the repository to be passed separately, which also means that this rule has to be singular.

Basically, the rule should be the same as `cosign_sign` but able to expand Make variables within the repository option. Taking a look at that implementation, you could pretty much just copy it and not even need to use http_file since rules_oci exposes the toolchains for cosign and jq. You could probably even directly reference the shell script templates rather than copying since they are exported.
</blockquote></details>

Done.

SanjayVas (changes requested):
> Reviewed 13 of 13 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2552)* status: all files reviewed, 6 unresolved discussions (waiting on @laureanobrs)

___
*[`src/main/docker/panel_exchange_client/images.bzl` line 48 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2552#-OWvMuDq5h2MfJt8cYea:-OWwIdIG9Sshos9qTg1E:bf0um81) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/afa6b58d0c52a9e47a18656b236cf4293117898c/src/main/docker/panel_exchange_client/images.bzl#L48)):*

Done.
</blockquote></details>

I still see a new ALL_IMAGES that includes LOCAL_IMAGES here?
___
*[`build/cosign/sign.bzl` line 6 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2552#-OWv6kL73CDrNi679-p_:-OWwJm948yuP3eB2Z8NL:bc2ozbf) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/afa6b58d0c52a9e47a18656b236cf4293117898c/build/cosign/sign.bzl#L6)):*

Done. The original file does not have a license header. I added "# Copyright 2022 The Bazel Contrib Authors"
</blockquote></details>

I don't think there are general authors for bazel-contrib. In this case there is no existing license header, so I don't know how to apply the process I shared above.

Perhaps the better approach here is to do this as a patch using single_version_override in MODULE.bazel. See https://bazel.build/rules/lib/globals/module#single_version_override
___
*[`build/cosign/extensions.bzl` line 19 at r9](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2552#-OWwGIWGAGPVHHSBJqNJ:-OWwGIWGAGPVHHSBJqNK:b-alq8qq) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/10046f356dd10867d036e56c00e6819273cb6d1e/build/cosign/extensions.bzl#L19)):*
> ```Starlark
> load("@rules_oci//cosign:repositories.bzl", "cosign_register_toolchains")
> 
> def _cosign_extension(module_ctx):
> ```

nit: Match conventions for extension definition



_Suggestion:_
```Starlark
_cosign_impl
```
___
*[`build/cosign/extensions.bzl` line 20 at r9](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2552#-OWwH25_A9Cd5doBKD8y:-OWwH25_A9Cd5doBKD8z:bhbf3fr) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/10046f356dd10867d036e56c00e6819273cb6d1e/build/cosign/extensions.bzl#L20)):*
> ```Starlark
> 
> def _cosign_extension(module_ctx):
>     cosign_register_toolchains(name = "cosign", register = False)
> ```

nit: use the same name that rules_oci uses in its WORKSPACE



_Suggestion:_
```Starlark
oci_cosign
```
___
*[`build/cosign/extensions.bzl` line 26 at r9](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2552#-OWwHErmEEyPnpeE-PCW:-OWwHErmEEyPnpeE-PCX:b7twma4) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/10046f356dd10867d036e56c00e6819273cb6d1e/build/cosign/extensions.bzl#L26)):*
> ```Starlark
>     implementation = _cosign_extension,
>     tag_classes = {
>         "toolchains": tag_class(),
> ```

Drop this tag class, as it doesn't do anything.

SanjayVas (changes requested):
> Reviewed 9 of 13 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2552)* status: 16 of 20 files reviewed, 4 unresolved discussions (waiting on @Copilot and @laureanobrs)

___
*[`build/cosign/sign.sh.tpl` line 25 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2552#-OWv6Gy96vxWfRSvU2uY:-OX-SlMl9B03BS4cj3JY:b-ep9pb) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/afa6b58d0c52a9e47a18656b236cf4293117898c/build/cosign/sign.sh.tpl#L25)):*

Is it OK to keep the file now that we use patches instead of modified copy-paste files?
</blockquote></details>

Fine for initial debugging, but you should drop it to keep the patches to a minimum. We need to maintain the patches, so the fewer changed files/lines the better.
___
*[`MODULE.bazel` line 207 at r10](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2552#-OX-S5he-qnHoW7FB_zE:-OX-S5he-qnHoW7FB_zF:bai8pne) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3db84b0a9d2f9869808e9e2fdfdfe4a93509c908/MODULE.bazel#L207)):*
> ```Starlark
> single_version_override(
>     module_name = "rules_oci",
>     version = "2.2.1",
> ```

Leave out the version from the override
___
*[`MODULE.bazel` line 209 at r10](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2552#-OX-S-T82PFnwAcJtEcG:-OX-S-T82PFnwAcJtEcH:b-8jnsin) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3db84b0a9d2f9869808e9e2fdfdfe4a93509c908/MODULE.bazel#L209)):*
> ```Starlark
>     version = "2.2.1",
>     patches = [
>         "//patches/rules_oci:sign_dot_bzl.patch",
> ```

Keep the patch files under the //build tree.
___
*[`patches/rules_oci/sign_dot_bzl.patch` line 11 at r10](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2552#-OX-SQrq1UEBH4hU1iF6:-OX-SQrrCwrwF9EiOLZ7:bb00jk2) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3db84b0a9d2f9869808e9e2fdfdfe4a93509c908/patches/rules_oci/sign_dot_bzl.patch#L11)):*
> ```Diff
> -        Digests and tags are not allowed.
> -    """),
> +    "repository_url": attr.string(mandatory = True, doc = "Image repository url subject to make variable expansion"),
> ```

Keep the existing attr name/meaning. Just add the make var expansion.

SanjayVas (changes requested):
> Reviewed 3 of 13 files at r10, 8 of 8 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2552)* status: all files reviewed, 1 unresolved discussion (waiting on @laureanobrs)

___
*[`.github/workflows/maybe-push-images.yml` line 76 at r11](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2552#-OX-j37xA-TrgcxhsdJJ:-OX-j37xA-TrgcxhsdJK:bm0bibf) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a7ee382adf0a646b883da6a2c6d6c3708690f35f/.github/workflows/maybe-push-images.yml#L76)):*
> ```YAML
>       # kms-key-resource-name: "gcpkms://projects/halo-cmm-poc/locations/global/keyRings/key-ring-signed-builds/cryptoKeys/key-signed-builds/cryptoKeyVersions/3"
>       #WIP - kms-key-resource-name: "gcpkms://projects/halo-cmm-poc/locations/global/keyRings/key-ring-signed-builds/cryptoKeys/key-signed-builds-ed25519"
>       kms-key-resource-name: "gcpkms://projects/halo-cmm-poc/locations/global/keyRings/key-ring-signed-builds/cryptoKeys/key-signed-build-hardware/cryptoKeyVersions/1"
> ```

Generally LGTM. Just waiting for these WIP bits to be resolved with the final key locations and we're good.

laureanobrs (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2552)* status: 21 of 22 files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

___
*[`.github/workflows/maybe-push-images.yml` line 76 at r11](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2552#-OX-j37xA-TrgcxhsdJJ:-OX5bA8LDbjonhNltZux:b90sz2a) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a7ee382adf0a646b883da6a2c6d6c3708690f35f/.github/workflows/maybe-push-images.yml#L76)):*

Generally LGTM. Just waiting for these WIP bits to be resolved with the final key locations and we're good.
</blockquote></details>

Done.

Successful run: https://github.com/world-federation-of-advertisers/cross-media-measurement/actions/runs/16816568167/job/47634681166

SanjayVas (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2552)* status: all files reviewed, 1 unresolved discussion (waiting on @laureanobrs)

___
*[`.github/workflows/maybe-push-images.yml` line 74 at r12](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2552#-OX5gKir0-4JwXQw-xif:-OX5gKir0-4JwXQw-xig:bk2p6qg) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7a84d4775bfa41c5342487af98f06c394f762657/.github/workflows/maybe-push-images.yml#L74)):*
> ```YAML
>       image-tag: ${{ needs.get-image-tag.outputs.image-tag }}
>       kms-key-resource-name: "gcpkms://${{ vars.IMAGE_SIGNING_DEV_KEY_RESOURCE_NAME }}"
>       service-account: ${{ vars.IMAGE_SIGNING_DEV_SA }}
> ```

nit: I thought we were going with "test" instead of "dev"?



_Code quote:_
```YAML
_DEV
```

Issue #2423, opened 2025-06-06
MeasurementConsumerSimulator does not respect specified eventRange


SanjayVas:
The `eventRange` parameter of `MeasurementConsumerSimulator` indicates what range to use for the collection interval in Measurement requests. #2321 incorrectly changed the behavior of MeasurementConsumerSimulator to ignore this parameter and instead rely on EventGroup data availability. This broke some tests that rely on `MeasurementConsumerSimulator` as a test harness.

#2390 provides a stopgap fix for the tests, but the class still does not behave as expected. This will break again e.g. once #2333 is implemented.

PR #2554, opened 2025-07-09
fix: Respect eventRange constructor parameter in EventQueryMeasurementConsumerSimulator

stevenwarejones:
Issue: #2423

### DIFF
--- a/src/main/kotlin/org/wfanet/measurement/loadtest/measurementconsumer/MeasurementConsumerSimulator.kt
+++ b/src/main/kotlin/org/wfanet/measurement/loadtest/measurementconsumer/MeasurementConsumerSimulator.kt
@@ -242,7 +242,7 @@ abstract class MeasurementConsumerSimulator(
       )
     val reachTolerance = computeErrorMargin(reachVariance)
     if (expectedResult.reach.value.toDouble() < reachTolerance) {
-      throw IllegalStateException("Expected result cannot be less and tolerance")
+      throw IllegalStateException("Expected result cannot be less than tolerance")
     }
 
     if (requiredCapabilities.honestMajorityShareShuffleSupported) {
@@ -366,7 +366,7 @@ abstract class MeasurementConsumerSimulator(
         )
       val reachTolerance = computeErrorMargin(reachVariance)
       if (expectedResult.reach.value.toDouble() < reachTolerance) {
-        throw IllegalStateException("Expected result cannot be less and tolerance")
+        throw IllegalStateException("Expected result cannot be less than tolerance")
       }
 
       assertThat(reachAndFrequencyResult)
@@ -433,7 +433,7 @@ abstract class MeasurementConsumerSimulator(
         )
       val reachTolerance = computeErrorMargin(reachVariance)
       if (expectedResult.reach.value.toDouble() < reachTolerance) {
-        throw IllegalStateException("Expected result cannot be less and tolerance")
+        throw IllegalStateException("Expected result cannot be less than tolerance")
       }
 
       assertThat(reachResult).reachValue().isWithin(reachTolerance).of(expectedResult.reach.value)
@@ -537,7 +537,7 @@ abstract class MeasurementConsumerSimulator(
       )
     val reachTolerance = computeErrorMargin(reachVariance)
     if (result.expectedResult.reach.value.toDouble() < reachTolerance) {
-      throw IllegalStateException("Expected result cannot be less and tolerance")
+      throw IllegalStateException("Expected result cannot be less than tolerance")
     }
 
     if (requiredCapabilities.honestMajorityShareShuffleSupported) {
@@ -593,8 +593,8 @@ abstract class MeasurementConsumerSimulator(
 
       val variance = computeImpressionVariance(result, measurementInfo.measurementSpec, protocol)
       val tolerance = computeErrorMargin(variance)
-      if (expectedResult.reach.value.toDouble() < tolerance) {
-        throw IllegalStateException("Expected result cannot be less and tolerance")
+      if (expectedResult.impression.value.toDouble() < tolerance) {
+        throw IllegalStateException("Expected impressions cannot be less than tolerance")
       }
       assertThat(result.impression.hasDeterministicCount()).isTrue()
       assertThat(result.impression.noiseMechanism).isEqualTo(expectedDirectNoiseMechanism)


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2554)

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2554)* status: 0 of 3 files reviewed, 2 unresolved discussions (waiting on @kungfuc, @kungfucraig, and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/measurementconsumer/EdpAggregatorMeasurementConsumerSimulator.kt` line 67 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2554#-OUju_1l1NPT09ouXA2v:-OUk0Dtd7H6QO4oMn3UE:b-z7ajsl) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/98a1f14302d5cb1550eaa0b30b83ab3819822f27/src/main/kotlin/org/wfanet/measurement/loadtest/measurementconsumer/EdpAggregatorMeasurementConsumerSimulator.kt#L67)):*

Rename this if it has different behavior than expected
</blockquote></details>

just deleted - can add back later if needed.
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/measurementconsumer/EventQueryMeasurementConsumerSimulator.kt` line 71 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2554#-OUjufgt4Ph4Y8lKJ8j0:-OUk0IDpAqmRp0S5lQ2w:bwmdwu5) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/98a1f14302d5cb1550eaa0b30b83ab3819822f27/src/main/kotlin/org/wfanet/measurement/loadtest/measurementconsumer/EventQueryMeasurementConsumerSimulator.kt#L71)):*

Drop this part. It's misleading, as it implies the behavior is that the dataAvailabilityInterval *should* be used instead of this,
</blockquote></details>

an MC should always consider dataAvailability when making a measurement request. Do you have a rewording that can capture that  as well? Or I can open an github issue and link to it that we need to respect dataAvailability here

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2554)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @kungfucraig and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/measurementconsumer/EventQueryMeasurementConsumerSimulator.kt` line 71 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2554#-OUjufgt4Ph4Y8lKJ8j0:-OUknE3Y5hPV5Rbeop0L:b-8trul) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/98a1f14302d5cb1550eaa0b30b83ab3819822f27/src/main/kotlin/org/wfanet/measurement/loadtest/measurementconsumer/EventQueryMeasurementConsumerSimulator.kt#L71)):*
> an MC should always consider dataAvailability when making a measurement request

This statement also doesn't match what's happening here. It'll only be an error to request times outside of the *DataProvider* data availability. An MC is free to request outside of the EventGroup availability - they should just expect 0 events outside of the EG availability.

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2554)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @kungfucraig and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/measurementconsumer/EventQueryMeasurementConsumerSimulator.kt` line 71 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2554#-OUjufgt4Ph4Y8lKJ8j0:-OUl5ugTE7mkVTKpfqUA:b-diym7s) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/98a1f14302d5cb1550eaa0b30b83ab3819822f27/src/main/kotlin/org/wfanet/measurement/loadtest/measurementconsumer/EventQueryMeasurementConsumerSimulator.kt#L71)):*
> i would think an MC should consider event group availability

This is incorrect. Copying my comment from Slack:

> I think you may be misunderstanding the use of EG data availability w.r.t. MCs. MCs only need to consider DataProvider availability when making a request. The EG availability is just to sync expectations with the EDP - i.e. whether to expect events outside of that range. MCs are likely to use a collection interval that **is not** within the availability range of every EG. For example, doing a report for a month with _some_ campaigns having started partway through the month.

Separately, the mechanism for choosing an event range that the test covers also need not consider EG data availability.

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2554)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @kungfucraig and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/measurementconsumer/EventQueryMeasurementConsumerSimulator.kt` line 71 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2554#-OUjufgt4Ph4Y8lKJ8j0:-OV9ezSd1my30HRds4pD:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/98a1f14302d5cb1550eaa0b30b83ab3819822f27/src/main/kotlin/org/wfanet/measurement/loadtest/measurementconsumer/EventQueryMeasurementConsumerSimulator.kt#L71)):*

It's again not used *instead* of any availability field. The comment can state that this must be within a data provider's availability interval, but it's still correct to specify an explicit interval here.
</blockquote></details>

Done.

SanjayVas (changes requested):
> Reviewed 1 of 3 files at r1, 1 of 1 files at r3, 1 of 1 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2554)* status: all files reviewed, 1 unresolved discussion (waiting on @kungfucraig)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/measurementconsumer/MeasurementConsumerSimulator.kt` line 252 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2554#-OV9seJ52ysOei0-OR9f:-OV9seJ52ysOei0-OR9g:b4k4ez8) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3319194bd79c501efd8b3c793370ff15069d3432/src/main/kotlin/org/wfanet/measurement/loadtest/measurementconsumer/MeasurementConsumerSimulator.kt#L252)):*
> ```Kotlin
>       .of(expectedResult.reach.value)
> 
>     assertThat(reachAndFrequencyResult.reach.value.toDouble()).isGreaterThan(reachTolerance)
> ```

Sorry, I hadn't looked at this part before.

I don't think this is the right way to handle the bug where we ended up with a low number of events in the test data. That's really more of a test setup/infra failure, which we can accomplish by checking the *expected* result. That should also throw some other exception rather than being an assertion to indicate it's a problem in the test setup rather than the SUT.

The line above this is already doing the correct assertion on the reach value.

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2554)* status: 2 of 3 files reviewed, 1 unresolved discussion (waiting on @kungfucraig and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/measurementconsumer/MeasurementConsumerSimulator.kt` line 252 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2554#-OV9seJ52ysOei0-OR9f:-OVDjaTN-HIR2QK2sO3M:b-nk17k8) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3319194bd79c501efd8b3c793370ff15069d3432/src/main/kotlin/org/wfanet/measurement/loadtest/measurementconsumer/MeasurementConsumerSimulator.kt#L252)):*

Sorry, I hadn't looked at this part before.

I don't think this is the right way to handle the bug where we ended up with a low number of events in the test data. That's really more of a test setup/infra failure, which we can accomplish by checking the *expected* result. That should also throw some other exception rather than being an assertion to indicate it's a problem in the test setup rather than the SUT.

The line above this is already doing the correct assertion on the reach value.
</blockquote></details>

updated this to throw an illegal state exception against the expected result. I think this should more intuitively communicate what to fix when it breaks

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2554)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @stevenwarejones)

PR #2558, opened 2025-07-09
fix: Increase memory limit of computation control service deployment

SanjayVas:
OOMKilled events were observed when running the correctness test in Halo-managed environments. The increase appears to be from classes, as there's a jump in Metaspace usage starting with nightly/20250626.1.


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2558)

PR #2560, opened 2025-07-10
feat: deploy edpa cloud functions using terraform

Marco-Premier:


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2560)

Marco-Premier (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2560)* status: 0 of 11 files reviewed, 3 unresolved discussions (waiting on @SanjayVas)

___
*[`src/main/terraform/gcloud/modules/gcs-bucket-cloud-function/main.tf` line 103 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2560#-OUpT1Xp0X5xyfW0q1Z5:-OUtZGfp7awPxFCkbOmj:b-ac9znf) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/44b832fc703b018bb5357c598e3c1e65ff2c1950/src/main/terraform/gcloud/modules/gcs-bucket-cloud-function/main.tf#L103)):*

Is there a reason why `--output=files` wouldn't work here?
</blockquote></details>

It works. changed it to `--output=files`
___
*[`src/main/terraform/gcloud/modules/gcs-bucket-cloud-function/main.tf` line 129 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2560#-OUpTGCY1IeFhoN51Ius:-OUtZLCACOmFQ3gSGLsO:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/44b832fc703b018bb5357c598e3c1e65ff2c1950/src/main/terraform/gcloud/modules/gcs-bucket-cloud-function/main.tf#L129)):*

Never use eval. You can build up the arguments as a Bash array instead. You can see how some of our other Bash scripts do this, such as https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/main/tools/bazel-container
</blockquote></details>

Done.
___
*[`src/main/terraform/gcloud/modules/gcs-bucket-cloud-function/variable.tf` line 63 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2560#-OUpTqEt1kwIrTHNMCRM:-OUtZM-u7ZC2kJzE-xQl:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/44b832fc703b018bb5357c598e3c1e65ff2c1950/src/main/terraform/gcloud/modules/gcs-bucket-cloud-function/variable.tf#L63)):*

Our Terraform modules must be usable by operators who are just referencing the module source [via the GitHub repo](https://developer.hashicorp.com/terraform/language/modules/sources#github) and therefore may not have the source code checked out or Bazel installed. Therefore the module needs to take in the uber jar path.

For Halo-managed environments, the uber JAR can be built and found using Bazel like this in the GitHub workflow. For real operators, we'll need to include the uber JARs as release artifacts that they can download if they don't want to build them on their own.
</blockquote></details>

Done.

SanjayVas (changes requested):
> Reviewed 2 of 11 files at r2, 4 of 5 files at r3.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2560)* status: 8 of 13 files reviewed, 4 unresolved discussions

___
*[`.github/workflows/terraform-cmms.yml` line 104 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2560#-OVEFJI2EXoAujh6v7LT:-OVEFJI3-ULRkDX2Zi4k:bta5a1r) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/448a88792862d92850d87605b6e825ce907f64a8/.github/workflows/terraform-cmms.yml#L104)):*
> ```YAML
>         EOF
> 
>     - name: Download Uber JARs using Maven
> ```

I don't think you need to install to a local repository. You should be able to use the [`dependency:copy` goal](https://maven.apache.org/plugins/maven-dependency-plugin/copy-mojo.html) and specify the output directory.
___
*[`.github/workflows/build_and_push_uber_jars.yml` line 43 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2560#-OVE974h5yJ1zBOonPjD:-OVE8Ekx6C_Glb5HjcOF:b-2boal6) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/448a88792862d92850d87605b6e825ce907f64a8/.github/workflows/build_and_push_uber_jars.yml#L43)):*
> ```YAML
>       - name: Build and Deploy Data Watcher
>         env:
>           IMAGE_TAG: ${{ inputs.image-tag }}
> ```

Maven needs some extra logic around versions because unlike image tags, they have an ordering and use a specific scheme. This means that you'll want to append "-SNAPSHOT" to the version if it's not from a release. Furthermore if it's not a semver version (e.g. when it's a commit hash like for manual builds), you'd use "0.0.0-${IMAGE_TAG}-SNAPSHOT" since it needs to start with semver.
___
*[`.github/actions/build-and-push-uber-jar/action.yml` line 71 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2560#-OVE8ioI5wpgjqJA1q_r:-OVE1ky-6aYqsiil8w-1:b-iknyky) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/448a88792862d92850d87605b6e825ce907f64a8/.github/actions/build-and-push-uber-jar/action.yml#L71)):*
> ```YAML
>         JAR_PATH=$(bazel cquery "$BAZEL_TARGET_LABEL" --output=files)
>               
>         mvn -B -e deploy:deploy-file \
> ```

Does this throw any errors if the same artifact version is published more than once? Or does it just overwrite? I came across that issue with publishing using the Bazel rule which requires being careful about versions. See other comment.

Marco-Premier (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2560)* status: 3 of 13 files reviewed, 4 unresolved discussions (waiting on @SanjayVas)

___
*[`.github/actions/build-and-push-uber-jar/action.yml` line 71 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2560#-OVE8ioI5wpgjqJA1q_r:-OVIG4vREj0b2ukUEvLC:b-at32sk) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/448a88792862d92850d87605b6e825ce907f64a8/.github/actions/build-and-push-uber-jar/action.yml#L71)):*

Does this throw any errors if the same artifact version is published more than once? Or does it just overwrite? I came across that issue with publishing using the Bazel rule which requires being careful about versions. See other comment.
</blockquote></details>

It doesn't throw. It adds assets to the same version, eg. https://github.com/world-federation-of-advertisers/cross-media-measurement/packages/2581416
___
*[`.github/workflows/build_and_push_uber_jars.yml` line 43 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2560#-OVE974h5yJ1zBOonPjD:-OVIFxYrABzgzV7AcQmr:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/448a88792862d92850d87605b6e825ce907f64a8/.github/workflows/build_and_push_uber_jars.yml#L43)):*

Maven needs some extra logic around versions because unlike image tags, they have an ordering and use a specific scheme. This means that you'll want to append "-SNAPSHOT" to the version if it's not from a release. Furthermore if it's not a semver version (e.g. when it's a commit hash like for manual builds), you'd use "0.0.0-${IMAGE_TAG}-SNAPSHOT" since it needs to start with semver.
</blockquote></details>

Done.
___
*[`.github/workflows/terraform-cmms.yml` line 104 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2560#-OVEFJI2EXoAujh6v7LT:-OVIGEE71UapQklurAz3:b-tlfx5i) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/448a88792862d92850d87605b6e825ce907f64a8/.github/workflows/terraform-cmms.yml#L104)):*

I don't think you need to install to a local repository. You should be able to use the [`dependency:copy` goal](https://maven.apache.org/plugins/maven-dependency-plugin/copy-mojo.html) and specify the output directory.
</blockquote></details>

I've been trying that first but copy goal doesn't support `remoteRepositories` which should be specified writing a temporary `pom.xml`. I didn't like that much and also I was getting errors when copy the jar.

Marco-Premier (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2560)* status: 6 of 13 files reviewed, 5 unresolved discussions (waiting on @SanjayVas)

___
*[`.github/workflows/build_and_push_uber_jars.yml` line 21 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2560#-OVJqZXd4J1orEtik0De:-OVg0di74VkBskxcsHVs:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d7ba0bea02c765b4df8903a80bb66b1c0fc3a7a3/.github/workflows/build_and_push_uber_jars.yml#L21)):*

This must *either* be called directly on release or be called by another workflow using workflow_call, but not both. I assume that what you want is the latter, since you need it to be done as part of the larger update-cmms workflow.
</blockquote></details>

Done.
___
*[`.github/workflows/build_and_push_uber_jars.yml` line 46 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2560#-OVJpQMN3UG-SAUkHraY:-OVg0rLD1wV8OKsJuw5w:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d7ba0bea02c765b4df8903a80bb66b1c0fc3a7a3/.github/workflows/build_and_push_uber_jars.yml#L46)):*

This says `base-version`, but the input for workflow_call is called `package-version`.
</blockquote></details>

Done.
___
*[`.github/workflows/build_and_push_uber_jars.yml` line 53 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2560#-OVJqtWA6iMx_qIt1N7p:-OVg2sDG-34SVZGVWc2Z:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d7ba0bea02c765b4df8903a80bb66b1c0fc3a7a3/.github/workflows/build_and_push_uber_jars.yml#L53)):*

There's technically three cases:

1. Called by workflow_call for a non-release build, where the input version will be a commit hash. In this case, you'd want to use the 0.0.0 prefix.
2. Called by workflow_call for a release build, where the version will be SemVer.
3. Called manually by workflow_dispatch, where the version should be Semver but it needs the -SNAPSHOT suffix.

Note that the publish-maven-artifacts workflow only needs to handle (2) and (3) since it's not called by the update-cmms workflow and is instead triggered directly on release. You will not be able to use the event name to differentiate between (1) and (2), so you may instead need to parse the input to determine whether it's already in SemVer format.
</blockquote></details>

Done.
___
*[`.github/workflows/terraform-cmms.yml` line 109 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2560#-OVJsNzt91xoOBZk-tnJ:-OVg331-5EuKys-BkUyq:b-18sv4r) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d7ba0bea02c765b4df8903a80bb66b1c0fc3a7a3/.github/workflows/terraform-cmms.yml#L109)):*

The event name will never be release here, since this workflow isn't triggered directly on release. This version must always be based on the image-tag input. You may want to extract a common action for this so  you can use the same logic in the workflow that pushes to Maven.
</blockquote></details>

Removed this entirely, now it's triggered only for `workflow_call` and check the `$PACKAGE_VERSION`

SanjayVas (changes requested):
> Reviewed 1 of 10 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2560)* status: 6 of 13 files reviewed, 2 unresolved discussions (waiting on @Marco-Premier)

___
*[`.github/workflows/terraform-cmms.yml` line 109 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2560#-OVJsNzt91xoOBZk-tnJ:-OVimhPmCOakB5Z12ErP:b-u7djij) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d7ba0bea02c765b4df8903a80bb66b1c0fc3a7a3/.github/workflows/terraform-cmms.yml#L109)):*

Removed this entirely, now it's triggered only for `workflow_call` and check the `$PACKAGE_VERSION`
</blockquote></details>

I'm still seeing the check for release here?

Marco-Premier (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2560)* status: 7 of 14 files reviewed, 3 unresolved discussions (waiting on @SanjayVas)

___
*[`.github/workflows/publish-cloud-run-function-uber-jars.yml` line 18 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2560#-OVnEMKm9o5WyQlRsEy2:-OVsONaNDOnDIfD2jnPk:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a0e1d8af5fb1b91f842ac92db4761a15c77eeb7e/.github/workflows/publish-cloud-run-function-uber-jars.yml#L18)):*

This still needs a workflow_dispatch so it can be called manually for testing.
</blockquote></details>

Done.
___
*[`.github/workflows/publish-cloud-run-function-uber-jars.yml` line 18 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2560#-OVnFt7029TzbDERNT2N:-OVsONz681A6T42Betpf:b-6mrzdo) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a0e1d8af5fb1b91f842ac92db4761a15c77eeb7e/.github/workflows/publish-cloud-run-function-uber-jars.yml#L18)):*

I'm not spotting where this is called from. I assume it should be called from update-cmms when the existing image-tag is not specified.
</blockquote></details>

Separate PR for that
___
*[`.github/workflows/terraform-cmms.yml` line 107 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2560#-OVnEFGJ54LiHpDYWsx3:-OVsPHTQA4l9ORvIVkOr:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a0e1d8af5fb1b91f842ac92db4761a15c77eeb7e/.github/workflows/terraform-cmms.yml#L107)):*
> ```YAML
>       uses: ./.github/actions/get-maven-artifact-version
>       with:
>         package-version: ${{ inputs.package-version }}
> ```

Done.

SanjayVas (changes requested):
> Reviewed 5 of 10 files at r4, 1 of 2 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2560)* status: 13 of 14 files reviewed, 1 unresolved discussion (waiting on @Marco-Premier)

___
*[`src/main/terraform/gcloud/modules/gcs-bucket-cloud-function/variable.tf` line 69 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2560#-OVscHhiCu0c0fbhaK0R:-OVscHhiCu0c0fbhaK0S:b-r7bf7h) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/bbef80e27469e1a5ae2023299cb3712a3a53fe4b/src/main/terraform/gcloud/modules/gcs-bucket-cloud-function/variable.tf#L69)):*
> ```HCL
> }
> 
> variable "uber_jar_dir" {
> ```

Don't pass a separate param for this. You can determine it trivially from the uber jar path above.

SanjayVas (approved):
> Reviewed 1 of 2 files at r7, 8 of 8 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2560)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Marco-Premier)

Marco-Premier (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2560)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Marco-Premier)

PR #2561, opened 2025-07-10
ci: Avoid triggering code-based workflows when PR description is edited

SanjayVas:
This relies on the default types for the `pull_request` event (`opened`, `synchronize`, `reopened`). See https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request

### DIFF
--- a/.github/workflows/api-lint.yml
+++ b/.github/workflows/api-lint.yml
@@ -19,7 +19,6 @@ on:
     branches:
       - main
       - 'releases/**'
-    types: [opened, synchronize, edited]
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -23,7 +23,6 @@ on:
     branches:
       - main
       - 'releases/**'
-    types: [opened, synchronize, edited]
   workflow_dispatch:
 
 concurrency:
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -5,7 +5,6 @@ on:
     branches:
       - main
       - 'releases/**'
-    types: [opened, synchronize, edited]
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}


SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2561)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @laureanobrs)

___
*[`.github/workflows/api-lint.yml` line 22 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2561#-OUpXv863LX0V0oKHBZP:-OUuVILvEywYSWxE4j6Q:b-2giuk8) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e681da35a784d7e656711630be786006c00df2b8/.github/workflows/api-lint.yml#L22)):*

Summary: I think removing “edited” also prevents the workflow from triggering when the base branch of a pull request is changed. Are we okay with that?

I understand that by default, a workflow only runs when a pull_request event’s activity type is `opened`, `synchronize`, or `reopened`. In the documentation I see:

**synchronize:**  
“A pull request’s head branch was updated. For example, the head branch was updated from the base branch or new commits were pushed to the head branch.”  
[https://docs.github.com/en/webhooks/webhook-events-and-payloads?actionType=synchronize#pull\_request](https://docs.github.com/en/webhooks/webhook-events-and-payloads?actionType=synchronize#pull_request)

**edited:**  
“The title or body of a pull request was edited, or the base branch of a pull request was changed.”  
[https://docs.github.com/en/webhooks/webhook-events-and-payloads?actionType=edited#pull\_request](https://docs.github.com/en/webhooks/webhook-events-and-payloads?actionType=edited#pull_request)
</blockquote></details>

This PR is both removing `edited` which we don't actually want for this case and also adding the missing `reopened` which happens very seldom but still makes sense. It's using the defaults to ensure that we keep in sync with any new types that GitHub determines are appropriate.

laureanobrs (approved):
> Reviewed 3 of 3 files at r1.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2561)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @SanjayVas)

PR #2565, opened 2025-07-13
feat: Create or Reuse required Primitive ReportingSets for CreateBasicReport

tristanvuong2021:
CreateBasicReport requires specific Primitive ReportingSets. If the ReportingSets already exist, then they can be reused. If they don't, then they are created.


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2565)

SanjayVas (changes requested):
> Reviewed 1 of 4 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2565)* status: 1 of 5 files reviewed, 1 unresolved discussion

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/BasicReportsService.kt` line 80 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2565#-OV9GxNABbC1kulnK0A3:-OV9GxNABbC1kulnK0A4:bf31zom) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f3a01b7f8cf80aa649fd22c087d3e2a09e145ccd/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/BasicReportsService.kt#L80)):*
> ```Kotlin
>     ImpressionQualificationFiltersCoroutineStub,
>   private val internalReportingSetsStub: InternalReportingSetsCoroutineStub,
>   private val publicReportingSetsService: ReportingSetsCoroutineImplBase,
> ```

This class already has access to the ReportingSets service via the internal service stub. It shouldn't also access the the public service impl. More clearly, service impls should never be called directly without using a stub, as otherwise the gRPC infra code gets skipped.

If you want to reuse some code from the public service impl, factor it out and have this class use that.

tristanvuong2021 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2565)* status: 0 of 8 files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/BasicReportsService.kt` line 80 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2565#-OV9GxNABbC1kulnK0A3:-OVIYfhTA7pff6m345r_:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f0f5e4664e1ab51e127c082d8d654b4799751d2e/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/BasicReportsService.kt#L80)):*

This class already has access to the ReportingSets service via the internal service stub. It shouldn't also access the the public service impl. More clearly, service impls should never be called directly without using a stub, as otherwise the gRPC infra code gets skipped.

If you want to reuse some code from the public service impl, factor it out and have this class use that.
</blockquote></details>

Done.

SanjayVas (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2565)* status: all files reviewed, 2 unresolved discussions (waiting on @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/BasicReportsService.kt` line 173 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2565#-OVJM1s18n8yB94Edo1h:-OVJM1s18n8yB94Edo1i:bwshuw) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/413ce39586f3cee1cbfb130b028f509fe4da6609/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/BasicReportsService.kt#L173)):*
> ```Kotlin
>       }
> 
>     val dataProviderEventGroupsMap: Map<String, MutableList<String>> = buildMap {
> ```

nit: this function is very long. Can you split it up?
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/ProtoConversions.kt` line 120 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2565#-OVJIi5d-kihwKGmX3rc:-OVJIi5d-kihwKGmX3rd:bj0x6hs) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/413ce39586f3cee1cbfb130b028f509fe4da6609/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/ProtoConversions.kt#L120)):*
> ```Kotlin
> import org.wfanet.measurement.reporting.v2alpha.timeIntervals
> 
> private val setExpressionCompiler = SetExpressionCompiler()
> ```

nit: wrap this inside a ProtoConverters object. We want to avoid top-level properties in new code.

stevenwarejones (approved):
> Reviewed 6 of 8 files at r3, 2 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2565)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @tristanvuong2021)

PR #2585, opened 2025-07-17
fix: Reduce max heap size of computation control server

SanjayVas:
OOMKilled events were still observed despite the changes in #2558. After reviewing monitoring, it appears that the max heap size is set much higher than necessary. This means that the container memory limit can also be reverted back to 512 MiB.

RELNOTES: Operators will likely want to decrease the max heap size of the Duchy computation control service to 192MiB to accomodate dependency updates.


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2585)

PR #2591, opened 2025-07-21
feat: Implement Trustee RequisitionRequestBuilder

roaminggypsy:


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2591)

renjiezh (changes requested):
> Reviewed 10 of 27 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2591)* status: 10 of 27 files reviewed, 4 unresolved discussions (waiting on @kungfucraig)

___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/trustee/FulfillRequisitionRequestBuilder.kt` line 45 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2591#-OVizWg6-9xptFwE_FQu:-OVizWg6-9xptFwE_FQv:b9m7gar) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/b66e23c12a513ac8af978e425c746f9daac9b5a8/src/main/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/trustee/FulfillRequisitionRequestBuilder.kt#L45)):*
> ```Kotlin
>   private val frequencyVector: FrequencyVector,
>   private val kmsClient: KmsClient,
>   private val kmsKekUriVal: String,
> ```

Can we rename these variable names without `val` or `bytes`.
___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/trustee/FulfillRequisitionRequestBuilder.kt` line 71 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2591#-OVj-9JFAQxlGuM26kPk:-OVj-9JFAQxlGuM26kPl:ber01se) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/b66e23c12a513ac8af978e425c746f9daac9b5a8/src/main/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/trustee/FulfillRequisitionRequestBuilder.kt#L71)):*
> ```Kotlin
>       encryptedFrequencyVectorBytes = encryptedData.toByteString()
> 
>       val kekAead = kmsClient.getAead(kmsKekUriVal)
> ```

This involves an rpc call. So we need a unit test for it. You can use FakeKmsClient to produce this scenario.
___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/trustee/FulfillRequisitionRequestBuilder.kt` line 73 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2591#-OVj0Nng3MMLyEJVw7qC:-OVj0Nng3MMLyEJVw7qD:byepn8y) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/b66e23c12a513ac8af978e425c746f9daac9b5a8/src/main/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/trustee/FulfillRequisitionRequestBuilder.kt#L73)):*
> ```Kotlin
>       val kekAead = kmsClient.getAead(kmsKekUriVal)
>       val outputStream = ByteArrayOutputStream()
>       dekHandle.write(BinaryKeysetWriter.withOutputStream(outputStream), kekAead)
> ```

This is also an rpc call. Please try to manipulate the aead from FakeKmsClient to produce the error in unit test.
___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/trustee/FulfillRequisitionRequestBuilder.kt` line 77 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2591#-OVj6J83-4fHSjEEoqM7:-OVj6J83-4fHSjEEoqM8:bwv92px) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/b66e23c12a513ac8af978e425c746f9daac9b5a8/src/main/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/trustee/FulfillRequisitionRequestBuilder.kt#L77)):*
> ```Kotlin
>     } catch (e: GeneralSecurityException) {
>       throw IllegalStateException("Failed to build trustee request due to a cryptographic error", e)
>     }
> ```

UNAVAILABLE,
ABORT,

renjiezh (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2591)* status: 10 of 27 files reviewed, 5 unresolved discussions (waiting on @kungfucraig and @roaminggypsy)

___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/trustee/FulfillRequisitionRequestBuilder.kt` line 68 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2591#-OVjCsRx4l6lEjPY6fkH:-OVjCsRx4l6lEjPY6fkI:bqzzfrm) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/b66e23c12a513ac8af978e425c746f9daac9b5a8/src/main/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/trustee/FulfillRequisitionRequestBuilder.kt#L68)):*
> ```Kotlin
>       val dekHandle = KeysetHandle.generateNew(KeyTemplates.get("AES128_GCM"))
>       val dekAead = dekHandle.getPrimitive(Aead::class.java)
>       val encryptedData = dekAead.encrypt(frequencyVector.toByteArray(), null)
> ```

We need to rewrite this part.
To align with the public api, the format of data should be bytes of integers array. [Link](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/commit/3553b2a16b642a7ecba8f7388a523b6a23fe3172#diff-5fc5dde7ee15fb113eb7facc34902a2310c13b024bede9ef16dff9808eb8b910R127)
So you need to access `frequencyVector.frequencyData`and convert into bytes (big_endian). Please add a helper function in the class to do that.

kungfucraig (changes requested):
It would have been nice to see the file moving, etc. (i.e. the prefactor) in a separate PR from the new code. This time let's just leave it, but in the future please try to use multiple PRs for this type of thing. PRs should be single purpose.

> Reviewed 23 of 27 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2591#-:-OVjWD3XBB8zfg6ZfqFW:b5x5hp7)* status: 26 of 27 files reviewed, 12 unresolved discussions (waiting on @renjiezh and @roaminggypsy)

___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/trustee/FulfillRequisitionRequestBuilder.kt` line 45 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2591#-OVizWg6-9xptFwE_FQu:-OVoJvD0Erf3ySsI--yd:b2ngthx) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/b66e23c12a513ac8af978e425c746f9daac9b5a8/src/main/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/trustee/FulfillRequisitionRequestBuilder.kt#L45)):*

Can we rename these variable names without `val` or `bytes`.
</blockquote></details>

+1 
___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/trustee/FulfillRequisitionRequestBuilder.kt` line 68 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2591#-OVjCsRx4l6lEjPY6fkH:-OVoKKRgEM-9kypBW79E:b-jp6b86) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/b66e23c12a513ac8af978e425c746f9daac9b5a8/src/main/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/trustee/FulfillRequisitionRequestBuilder.kt#L68)):*

We need to rewrite this part.
To align with the public api, the format of data should be bytes of integers array. [Link](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/commit/3553b2a16b642a7ecba8f7388a523b6a23fe3172#diff-5fc5dde7ee15fb113eb7facc34902a2310c13b024bede9ef16dff9808eb8b910R127)
So you need to access `frequencyVector.frequencyData`and convert into bytes (big_endian). Please add a helper function in the class to do that.
</blockquote></details>

We need to send an array of bytes, where each byte is a single element of the FV. Thus, you need something like this:

val output : ByteArray
for index, item in fv:
  if item > 255:
    throw
   else:
     output[index]  = item.truncatedToSingleByte()


This will result in a payload that is 25% of the size of using an integer.     
  
___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/trustee/FulfillRequisitionRequestBuilder.kt` line 40 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2591#-OVo43ECAn5xNkjzi83B:-OVo43ECAn5xNkjzi83C:bfuuxwx) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/b66e23c12a513ac8af978e425c746f9daac9b5a8/src/main/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/trustee/FulfillRequisitionRequestBuilder.kt#L40)):*
> ```Kotlin
> import org.wfanet.measurement.consent.client.dataprovider.computeRequisitionFingerprint
> 
> class FulfillRequisitionRequestBuilder(
> ```

Add kdoc
___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/trustee/FulfillRequisitionRequestBuilder.kt` line 76 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2591#-OVoLXN87Ei0WMHRUuvL:-OVoLXN87Ei0WMHRUuvM:b4co0jm) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/b66e23c12a513ac8af978e425c746f9daac9b5a8/src/main/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/trustee/FulfillRequisitionRequestBuilder.kt#L76)):*
> ```Kotlin
>       encryptedDekBytes = outputStream.toByteArray().toByteString()
>     } catch (e: GeneralSecurityException) {
>       throw IllegalStateException("Failed to build trustee request due to a cryptographic error", e)
> ```

s/request/FulfillRequisitionRequest
s/trustee/TrusTEE

Also I think just a regular old RuntimeException would be better. Or you could consider a custom exception "FulfllRequisitionRequestBuildException" derived from RuntimeException.

Note that the Java doc says the following about the IllegalStateException: "This exception is thrown when a method is invoked at an illegal or inappropriate time or if the provider is not in an appropriate state for the requested operation. For example, this exception must be thrown if `Session.commit` is called on a non-transacted session. This exception is also called when a domain inappropriate method is called, such as calling `TopicSession.CreateQueueBrowser`."
___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/trustee/FulfillRequisitionRequestBuilder.kt` line 102 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2591#-OVoMZv3-vcVATCywf_1:-OVoMZv3-vcVATCywf_2:b-6tkxz3) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/b66e23c12a513ac8af978e425c746f9daac9b5a8/src/main/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/trustee/FulfillRequisitionRequestBuilder.kt#L102)):*
> ```Kotlin
>                 impersonatedServiceAccount = impersonatedServiceAccountVal
>               }
>             // TODO: populationSpecFingerprint
> ```

We should talk about the algorithm for this and maybe you can code it up? cc: @renjiezh 
___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/shareshuffle/BUILD.bazel` line 27 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2591#-OVjW68o-V-OurTO2E1y:-OVjW68o-V-OurTO2E1z:b-azm1k2) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/b66e23c12a513ac8af978e425c746f9daac9b5a8/src/main/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/shareshuffle/BUILD.bazel#L27)):*
> ```Starlark
> 
> maven_export(
>     name = "requisition_shareshuffle_maven",
> ```

As mentioned above, just a single maven package is required.
___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/frequencyvector/v2alpha/FrequencyVectorBuilder.kt` line 15 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2591#-OVjVQ8FE3ju-usFIj06:-OVjVQ8FE3ju-usFIj07:b4d7w10) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/b66e23c12a513ac8af978e425c746f9daac9b5a8/src/main/kotlin/org/wfanet/measurement/eventdataprovider/frequencyvector/v2alpha/FrequencyVectorBuilder.kt#L15)):*
> ```Kotlin
> 
> package org.wfanet.measurement.eventdataprovider.frequencyvector.v2alpha
> ```

I'm inclined to have this in requisition/common. Or if this class cannot be reused for TrusTEE it should live in "shareshuffle"
___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/trustee/BUILD.bazel` line 25 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2591#-OVjW9rT7UD-gYHancMJ:-OVjW9rT7UD-gYHancMK:b-azm1k2) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/b66e23c12a513ac8af978e425c746f9daac9b5a8/src/main/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/trustee/BUILD.bazel#L25)):*
> ```Starlark
> 
> maven_export(
>     name = "requisition_trustee_maven",
> ```

As mentioned above, just a single maven package is required.
___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/frequencyvector/v2alpha/BUILD.bazel` line 21 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2591#-OVjVUfH6oj2hlPI2Cod:-OVjVUfH6oj2hlPI2Coe:b-vie7hv) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/b66e23c12a513ac8af978e425c746f9daac9b5a8/src/main/kotlin/org/wfanet/measurement/eventdataprovider/frequencyvector/v2alpha/BUILD.bazel#L21)):*
> ```Starlark
> 
> maven_export(
>     name = "frequencyvector_maven",
> ```

We should just have a single maven package for requisition.
___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/common/BUILD.bazel` line 25 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2591#-OVjW11mEFMvElSmGLIG:-OVjW11mEFMvElSmGLIH:b-azm1k2) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/b66e23c12a513ac8af978e425c746f9daac9b5a8/src/main/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/common/BUILD.bazel#L25)):*
> ```Starlark
> 
> maven_export(
>     name = "requisition_common_maven",
> ```

As mentioned above, just a single maven package is required.

kungfucraig (changes requested):
> Reviewed 1 of 27 files at r1, 4 of 17 files at r2, 1 of 1 files at r3, 1 of 1 files at r4, 2 of 3 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2591)* status: 26 of 27 files reviewed, 13 unresolved discussions (waiting on @renjiezh)

___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/common/FrequencyVectorBuilder.kt` line 129 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2591#-OWLSBDZ2COHUsyI5l_H:-OWLSBDZ2COHUsyI5l_I:bakrwre) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/68ccdaa80ce1dd13eeb8249c17697fa13f69bb01/src/main/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/common/FrequencyVectorBuilder.kt#L129)):*
> ```Kotlin
> 
>   /** The accumulated frequency data */
>   private val frequencyData: IntArray
> ```

@roaminggypsy Would you mind changing this to be a ByteArray and providing an additional build method that 
___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/trustee/FulfillRequisitionRequestBuilder.kt` line 47 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2591#-OWLJPsyAjRl8ikjWrQQ:-OWLJPsyAjRl8ikjWrQR:b-tonjki) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/68ccdaa80ce1dd13eeb8249c17697fa13f69bb01/src/main/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/trustee/FulfillRequisitionRequestBuilder.kt#L47)):*
> ```Kotlin
>  * @param kmsClient The key management system client
>  * @param kmsKekUri The key encryption key uri
>  * @param workloadIdentityProvider The resource name of the workload identity provider
> ```

Give an example of this and of the service account name.
___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/trustee/FulfillRequisitionRequestBuilder.kt` line 52 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2591#-OWLThy73Q4Vt4VOa9lC:-OWLThy73Q4Vt4VOa9lD:b-yj6bid) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/68ccdaa80ce1dd13eeb8249c17697fa13f69bb01/src/main/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/trustee/FulfillRequisitionRequestBuilder.kt#L52)):*
> ```Kotlin
>  * @throws RuntimeException if a cryptographic error happens
>  */
> class FulfillRequisitionRequestBuilder(
> ```

We need to have an option to not encrypt the fulfillment payload.

I'd recommend adding a data class called "EncryptionParams" that encapsulates the various params needed for encryption and allow that as an optional input to the constructor. 

Then have "build" require the encrypted params and add another function called "buildUnencrypted" that will build the request payload without encrypting it.
___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/trustee/FulfillRequisitionRequestBuilder.kt` line 55 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2591#-OWLT5fq3CVx91tjbSPx:-OWLRyLJDmMnaw432fRR:b-ixa0q) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/68ccdaa80ce1dd13eeb8249c17697fa13f69bb01/src/main/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/trustee/FulfillRequisitionRequestBuilder.kt#L55)):*
> ```Kotlin
>   private val requisition: Requisition,
>   private val requisitionNonce: Long,
>   private val frequencyVector: FrequencyVector,
> ```

Can you leave a file and issue and leave a TODO to refit this to take a ByteArray and to update FrequencyVectorBuilder to operate on a ByteArray?
___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/trustee/FulfillRequisitionRequestBuilder.kt` line 143 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2591#-OWLPDeL2-FemJzYlcGi:-OWLPDeL2-FemJzYlcGj:b-eyzt3e) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/68ccdaa80ce1dd13eeb8249c17697fa13f69bb01/src/main/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/trustee/FulfillRequisitionRequestBuilder.kt#L143)):*
> ```Kotlin
>     }
> 
>     private fun integersToBytes(integers: List<Int>): ByteArray {
> ```

This is more idiomatic:

ByteArray(integers.size) {index ->
   require(integers[index] in 0..255)
   integers[index].toByte()
  }

Really you don't need this function at all. You could just inline the above.

SanjayVas (changes requested):
> Reviewed 4 of 17 files at r2.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2591)* status: 25 of 27 files reviewed, 14 unresolved discussions (waiting on @kungfucraig and @renjiezh)

___
*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2591#-OWNHwWa1Xbc4j7k0dlP:-OWNHwWa1Xbc4j7k0dlQ:bat9eyk) (no related file):*
> It would have been nice to see the file moving, etc. (i.e. the prefactor) in a separate PR from the new code. This time let's just leave it, but in the future please try to use multiple PRs for this type of thing. PRs should be single purpose.

I'm going to go further and state that the refactor for renaming the packages needs to be split off into a separate PR. It's just too messy otherwise. That PR should have a type of `refactor!` and include a `BREAKING-CHANGE` trailer explaining what the breaking change is. This trailer can be repeated to indicate multiple breaking changes.

roaminggypsy:
> > Reviewed 4 of 17 files at r2.
> > _[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2591)_ status: 25 of 27 files reviewed, 14 unresolved discussions (waiting on @kungfucraig and @renjiezh)
> 
> _[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2591#-OWNHwWa1Xbc4j7k0dlP:-OWNHwWa1Xbc4j7k0dlQ:bat9eyk) (no related file):_
> 
> > It would have been nice to see the file moving, etc. (i.e. the prefactor) in a separate PR from the new code. This time let's just leave it, but in the future please try to use multiple PRs for this type of thing. PRs should be single purpose.
> 
> I'm going to go further and state that the refactor for renaming the packages needs to be split off into a separate PR. It's just too messy otherwise. That PR should have a type of `refactor!` and include a `BREAKING-CHANGE` trailer explaining what the breaking change is. This trailer can be repeated to indicate multiple breaking changes.

yes. 

I'm going to keep this one around for any reference, and make a new one for refactoring.

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2591)* status: 25 of 27 files reviewed, 14 unresolved discussions (waiting on @kungfucraig and @SanjayVas)

___
*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2591#-OWNHwWa1Xbc4j7k0dlP:-OWNR0xq--UUEnnIQKGQ:b-ba7jcz) (no related file):*

yes. 

I'm going to keep this one around for any reference, and make a new one for refactoring.

</blockquote></details>

Sounds good to me.

kungfucraig (commented):
> Reviewed 5 of 17 files at r2, 1 of 1 files at r5, 1 of 2 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2591)* status: 26 of 27 files reviewed, 9 unresolved discussions (waiting on @renjiezh and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/common/FrequencyVectorBuilder.kt` line 129 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2591#-OWLSBDZ2COHUsyI5l_H:-OWNK0XF6bzoIHwUoEsc:b-ifqvxl) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/68ccdaa80ce1dd13eeb8249c17697fa13f69bb01/src/main/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/common/FrequencyVectorBuilder.kt#L129)):*

I feel like this PR is becoming too many things. I will address this in another PR for #2705
</blockquote></details>

That sounds great. Thanks for pushing back. :)

SanjayVas (changes requested):
> Reviewed 1 of 2 files at r7, 22 of 23 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2591)* status: 28 of 29 files reviewed, 15 unresolved discussions (waiting on @kungfucraig and @renjiezh)

___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/trustee/FulfillRequisitionRequestBuilder.kt` line 48 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2591#-OWbYiph1mrVH0_bHTen:-OWbYiph1mrVH0_bHTeo:b3xkyh7) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0682151dfc815c0c30d5f141c2c2b5847277225a/src/main/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/trustee/FulfillRequisitionRequestBuilder.kt#L48)):*
> ```Kotlin
>  *   encrypted.
>  * @throws IllegalArgumentException if the requisition is malformed or the frequency vector is empty
>  * @throws RuntimeException if a cryptographic error happens
> ```

Probably want to use GeneralSecurityException (or a relevant subclass of it) instead.
___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/trustee/FulfillRequisitionRequestBuilder.kt` line 77 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2591#-OWbZREK9cih52CqCRA4:-OWbZREK9cih52CqCRA5:b-tehyck) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0682151dfc815c0c30d5f141c2c2b5847277225a/src/main/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/trustee/FulfillRequisitionRequestBuilder.kt#L77)):*
> ```Kotlin
>   private val protocolConfig: TrusTee
>   private val frequencyVectorBytes: ByteArray
>   private var encryptedFrequencyVector: ByteString? = null
> ```

Why do we have mutable instance variables here? This is not a true builder in the sense that it doesn't have setters and everything is specified in the constructor. Therefore, the only mutable variables you'd need would be local to the `build` method.
___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/trustee/FulfillRequisitionRequestBuilder.kt` line 100 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2591#-OWbZi_xCsuuOGfxxhZ5:-OWbZi_xCsuuOGfxxhZ6:bbjk39j) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0682151dfc815c0c30d5f141c2c2b5847277225a/src/main/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/trustee/FulfillRequisitionRequestBuilder.kt#L100)):*
> ```Kotlin
>       }
> 
>     encryptionParams?.let { params ->
> ```

Use a regular`if` rather than `?.let`
___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/trustee/FulfillRequisitionRequestBuilder.kt` line 101 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2591#-OWbZt8z6BL3vA0l35vG:-OWbZt8z6BL3vA0l35vH:bltlybe) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0682151dfc815c0c30d5f141c2c2b5847277225a/src/main/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/trustee/FulfillRequisitionRequestBuilder.kt#L101)):*
> ```Kotlin
> 
>     encryptionParams?.let { params ->
>       val dekHandle = KeysetHandle.generateNew(KeyTemplates.get("AES256_GCM"))
> ```

Extract a constant for the key template



_Code quote:_
```Kotlin
AES256_GCM
```
___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/trustee/FulfillRequisitionRequestBuilder.kt` line 104 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2591#-OWb_0Di1NHLAF9YbssU:-OWb_0DjD1J_LJzKRBSF:b-2j33m6) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0682151dfc815c0c30d5f141c2c2b5847277225a/src/main/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/trustee/FulfillRequisitionRequestBuilder.kt#L104)):*
> ```Kotlin
>       val dekAead = dekHandle.getPrimitive(Aead::class.java)
> 
>       val encryptedData = dekAead.encrypt(frequencyVectorBytes, null)
> ```

Explicit types
___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/trustee/FulfillRequisitionRequestBuilder.kt` line 115 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2591#-OWbYzCD9vLJa-PDnOkp:-OWbYzCE2VOqw95UTNTn:b-13t01e) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0682151dfc815c0c30d5f141c2c2b5847277225a/src/main/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/trustee/FulfillRequisitionRequestBuilder.kt#L115)):*
> ```Kotlin
> 
>   /** Builds the Sequence of requests with encrypted payload. */
>   fun buildEncrypted(): Sequence<FulfillRequisitionRequest> = sequence {
> ```

You shouldn't need separate public methods since you already know whether to encrypt by the virtue of `encryptionParams` being `null`.
___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/trustee/FulfillRequisitionRequestBuilder.kt` line 206 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2591#-OWb_VxU6fIAZz0SJsjN:-OWb_VxU6fIAZz0SJsjO:b-cuqi5p) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0682151dfc815c0c30d5f141c2c2b5847277225a/src/main/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/trustee/FulfillRequisitionRequestBuilder.kt#L206)):*
> ```Kotlin
>       requisitionNonce: Long,
>       frequencyVector: FrequencyVector,
>       kmsClient: KmsClient,
> ```

The point of a params object like EncryptionParams is to reduce the number of parameters in public method signatures.

kungfucraig (changes requested):
> Reviewed 4 of 17 files at r2, 21 of 23 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2591)* status: 27 of 29 files reviewed, 15 unresolved discussions (waiting on @renjiezh, @roaminggypsy, and @SanjayVas)

roaminggypsy (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2591#-:-OWnWfpY6IntNF0yUDdQ:bnfp4nl)* status: 26 of 29 files reviewed, 15 unresolved discussions (waiting on @kungfucraig, @renjiezh, and @SanjayVas)

___
*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2591#-OWNHwWa1Xbc4j7k0dlP:-OWbUjhH7Am4G3r-1sFY:b-896fix) (no related file):*

Sounds good to me.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/trustee/FulfillRequisitionRequestBuilder.kt` line 48 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2591#-OWbYiph1mrVH0_bHTen:-OWnA4ek2RLABpRYbcE7:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0682151dfc815c0c30d5f141c2c2b5847277225a/src/main/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/trustee/FulfillRequisitionRequestBuilder.kt#L48)):*

Probably want to use GeneralSecurityException (or a relevant subclass of it) instead.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/trustee/FulfillRequisitionRequestBuilder.kt` line 77 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2591#-OWbZREK9cih52CqCRA4:-OWnWhO27AgrSrcFFcBD:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0682151dfc815c0c30d5f141c2c2b5847277225a/src/main/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/trustee/FulfillRequisitionRequestBuilder.kt#L77)):*

Why do we have mutable instance variables here? This is not a true builder in the sense that it doesn't have setters and everything is specified in the constructor. Therefore, the only mutable variables you'd need would be local to the `build` method.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/trustee/FulfillRequisitionRequestBuilder.kt` line 100 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2591#-OWbZi_xCsuuOGfxxhZ5:-OWnLywY9zmmpdVNM3eR:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0682151dfc815c0c30d5f141c2c2b5847277225a/src/main/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/trustee/FulfillRequisitionRequestBuilder.kt#L100)):*

Use a regular`if` rather than `?.let`
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/trustee/FulfillRequisitionRequestBuilder.kt` line 101 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2591#-OWbZt8z6BL3vA0l35vG:-OWnD8tZ0OFDxPd_OPv3:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0682151dfc815c0c30d5f141c2c2b5847277225a/src/main/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/trustee/FulfillRequisitionRequestBuilder.kt#L101)):*

Extract a constant for the key template
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/trustee/FulfillRequisitionRequestBuilder.kt` line 104 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2591#-OWb_0Di1NHLAF9YbssU:-OWnDn_MD0frk6hTCkEi:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0682151dfc815c0c30d5f141c2c2b5847277225a/src/main/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/trustee/FulfillRequisitionRequestBuilder.kt#L104)):*

Explicit types
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/trustee/FulfillRequisitionRequestBuilder.kt` line 115 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2591#-OWbYzCD9vLJa-PDnOkp:-OWnWmsWC53asOiVw-mr:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0682151dfc815c0c30d5f141c2c2b5847277225a/src/main/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/trustee/FulfillRequisitionRequestBuilder.kt#L115)):*

You shouldn't need separate public methods since you already know whether to encrypt by the virtue of `encryptionParams` being `null`.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/trustee/FulfillRequisitionRequestBuilder.kt` line 206 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2591#-OWb_VxU6fIAZz0SJsjN:-OWnOIWX5dAhfzCJuu-E:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0682151dfc815c0c30d5f141c2c2b5847277225a/src/main/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/trustee/FulfillRequisitionRequestBuilder.kt#L206)):*

The point of a params object like EncryptionParams is to reduce the number of parameters in public method signatures.
</blockquote></details>

Done.

kungfucraig (changes requested):
This is on the right track. Feel free to merge it without my approval once you get an approval from Sanjay and Renjie.

> Reviewed 1 of 1 files at r9, 2 of 4 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2591#-:-OWrhciUC6BVUMlpQmjR:b-gjkyrh)* status: 29 of 31 files reviewed, 17 unresolved discussions (waiting on @renjiezh and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/requisition/Native.kt` line 16 at r10](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2591#-OWrh00E6RVNpfqpjh3X:-OWrh00E6RVNpfqpjh3Y:b-ql8ph5) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/cf6b47a7edf8a505c66bad3fdd19f323da13ad43/src/main/kotlin/org/wfanet/measurement/eventdataprovider/requisition/Native.kt#L16)):*
> ```Kotlin
>  * limitations under the License.
>  */
> ```

Any reason why you're making this change? Generally you don't want to include whitespace only changes in unrelated PRs.
___
*[`src/test/kotlin/org/wfanet/measurement/eventdataprovider/requisition/NativeTest.kt` line 16 at r10](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2591#-OWrhYYQ7dY_wWaGfK7S:-OWrhYYQ7dY_wWaGfK7T:bb73xfv) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/cf6b47a7edf8a505c66bad3fdd19f323da13ad43/src/test/kotlin/org/wfanet/measurement/eventdataprovider/requisition/NativeTest.kt#L16)):*
> ```Kotlin
>  * limitations under the License.
>  */
> ```

Same

renjiezh (approved):
> Reviewed 21 of 23 files at r8, 1 of 1 files at r9, 4 of 4 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2591)* status: all files reviewed, 9 unresolved discussions (waiting on @roaminggypsy and @SanjayVas)

SanjayVas (changes requested):
> Reviewed 1 of 17 files at r2, 1 of 1 files at r9, 4 of 4 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2591)* status: all files reviewed, 6 unresolved discussions (waiting on @roaminggypsy)

___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/trustee/BUILD.bazel` line 16 at r10](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2591#-OWvtEOO5CqIQiS02kYB:-OWvtEOP10UXEBNaYoUN:b-djc6q1) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c05f495a8950f53c79b9e9c0583f777581ef9221/src/main/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/trustee/BUILD.bazel#L16)):*
> ```Starlark
>     srcs = ["FulfillRequisitionRequestBuilder.kt"],
>     tags = ["maven_coordinates=" + MAVEN_COORDINATES],
>     visibility = [
> ```

nit: move this to the default package visibility. Note that it already covers the test
___
*[`src/test/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/trustee/BUILD.bazel` line 5 at r10](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2591#-OWwELTLAR0Mj3TzSycB:-OWwELTLAR0Mj3TzSycC:b-owjjog) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c05f495a8950f53c79b9e9c0583f777581ef9221/src/test/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/trustee/BUILD.bazel#L5)):*
> ```Starlark
> kt_jvm_test(
>     name = "fulfill_requisition_request_builder_test",
>     testonly = True,
> ```

nit: You don't need to mark `_test` targets as testonly since they are inherently that way
___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/trustee/FulfillRequisitionRequestBuilder.kt` line 96 at r10](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2591#-OWw9WiTBS9kIImC2CZG:-OWw6McX6q0zC3z_EUCW:b-rndl6e) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c05f495a8950f53c79b9e9c0583f777581ef9221/src/main/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/trustee/FulfillRequisitionRequestBuilder.kt#L96)):*
> ```Kotlin
>   /** Builds the Sequence of requests with payload. */
>   fun build(): Sequence<FulfillRequisitionRequest> = sequence {
>     var payload = frequencyVectorBytes.toByteString()
> ```

Refactor this to not need the mutable payload.
___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/trustee/FulfillRequisitionRequestBuilder.kt` line 114 at r10](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2591#-OWw6dhKB7lHUfzGrcp0:-OWw6dhKB7lHUfzGrcp1:b-2uqa2u) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c05f495a8950f53c79b9e9c0583f777581ef9221/src/main/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/trustee/FulfillRequisitionRequestBuilder.kt#L114)):*
> ```Kotlin
>             val dekAead = dekHandle.getPrimitive(Aead::class.java)
> 
>             val encryptedData: ByteArray = dekAead.encrypt(frequencyVectorBytes, null)
> ```

Are we not using a streaming key type for the DEK? If we do that, then we can encrypt in chunks as we emit them. i.e. use StreamingAead instead of Aead.

SanjayVas (changes requested):
> Reviewed 3 of 8 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2591)* status: 26 of 31 files reviewed, 6 unresolved discussions (waiting on @kungfucraig, @renjiezh, and @roaminggypsy)

___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/trustee/FulfillRequisitionRequestBuilder.kt` line 114 at r10](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2591#-OWw6dhKB7lHUfzGrcp0:-OX5c0S77qXxF2x2TeaX:bov16zu) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c05f495a8950f53c79b9e9c0583f777581ef9221/src/main/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/trustee/FulfillRequisitionRequestBuilder.kt#L114)):*

Use StreamingAead now but 

1) I did not do "encrypt in chunks as we emit them". I tried something like the snippet but it just doesn't look as intuitive as the current implementation (and currently this snippet doesn't work)

- A write might not produce any readable ciphertext immediately.
- A write might produce less or (more likely) more ciphertext than the size of the plaintext chunk due to headers, nonces, and authentication tags. 
- Multiple write calls might be needed before the primitive flushes a segment of ciphertext.

2) are we sure about not needing coroutine? Although the current implementation unit tests pass, the deadlock could happen because: 
1. The writer starts. plaintextChannel.write(...) begins encrypting data and pushing the ciphertext into the pipe's sink.
2. A Pipe has a very small internal buffer (often just a few kilobytes). This buffer will fill up almost instantly.
3. Once the buffer is full, the write call blocks (pauses). It will wait indefinitely for a reader to consume data from the other end of the pipe to make space.
4. However, the reader is in the next block of code on the same thread. It will never get a chance to run because the writer is blocked, holding the thread.
</blockquote></details>

What you basically want is a buffer of of size RPC_CHUNK_SIZE_BYTES which you will copy your ciphertext bytes into. Once that buffer is full, you yield a body chunk until you get to the end of the data in which you can yield a chunk with fewer than RPC_CHUNK_SIZE_BYTES in the data.

Your outer loop is over writing all the plaintext data into the encrypting channel.

Therefore the loop iterations look something like:
* Write some data into writablebytechannel
* Within the WriteableByteChannel impl you've passed, copy what you get into your buffer
* If it's full, yield the body chunk
* If it's not full and you've written all the plaintext data, yield the last body chunk
___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/trustee/FulfillRequisitionRequestBuilder.kt` line 123 at r11](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2591#-OX5YyQm2eeN8qQjUEgg:-OX5YyQm2eeN8qQjUEgh:b1jgl4v) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/250ccbdf9f5d3caa16dcbef6df2a8256a09e2da4/src/main/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/trustee/FulfillRequisitionRequestBuilder.kt#L123)):*
> ```Kotlin
>   }
> 
>   private suspend fun SequenceScope<FulfillRequisitionRequest>.yieldBodyChunks(
> ```

Just return the Sequence of body chunks and call `yieldAll` from the caller



_Suggestion:_
```Kotlin
fun buildEncryptedBodyChunks(dekHandle: KeysetHandle, plainText: ByteArray): Sequence<FulfillRequisitionRequest>
```
___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/trustee/FulfillRequisitionRequestBuilder.kt` line 123 at r11](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2591#-OX5ZMKu1_WWCPGiFP1k:-OX5ZMKu1_WWCPGiFP1l:b-af4v31) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/250ccbdf9f5d3caa16dcbef6df2a8256a09e2da4/src/main/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/trustee/FulfillRequisitionRequestBuilder.kt#L123)):*
> ```Kotlin
>   }
> 
>   private suspend fun SequenceScope<FulfillRequisitionRequest>.yieldBodyChunks(
> ```

This shouldn't be a suspend function



_Suggestion:_
```Kotlin
private fun
```
___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/common/BUILD.bazel` line 9 at r11](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2591#-OX5YG4D4TydG4mXlQgU:-OX5YG4D4TydG4mXlQgV:b2tv1gk) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/250ccbdf9f5d3caa16dcbef6df2a8256a09e2da4/src/main/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/common/BUILD.bazel#L9)):*
> ```Starlark
>     test_target(":__pkg__"),
>     "//src/main/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha:__pkg__",
>     "//src/test/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/shareshuffle:__pkg__",
> ```

nit: common can just be visible to all siblings (subpackages of parent)



_Suggestion:_
```Starlark
    "//src/main/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha:__subpackages__",
    "//src/test/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha:__subpackages__",
```

SanjayVas (changes requested):
> Reviewed 4 of 8 files at r11, 6 of 6 files at r12, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2591)* status: all files reviewed, 5 unresolved discussions (waiting on @kungfucraig and @renjiezh)

___
*[`src/test/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/trustee/FulfillRequisitionRequestBuilderTest.kt` line 231 at r12](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2591#-OXVcr0p-NbMaJc2PNwb:-OXVcr0qEr5az2Ts6oxT:bom0i6t) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5f514349cc5d36de61225aa5a96dc61145aeb080/src/test/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/trustee/FulfillRequisitionRequestBuilderTest.kt#L231)):*
> ```Kotlin
>         byteArrayOf(),
>       )
>     val decryptedData = ByteArrayOutputStream()
> ```

nit: just for the test, you can use a decrypting stream to make it easier to read all of the decrypted data
___
*[`MODULE.bazel` line 401 at r12](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2591#-OXVbqIw8EEcOy2HPSMF:-OXVbqIx5jeqIaYbHCl0:b-z0bnma) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5f514349cc5d36de61225aa5a96dc61145aeb080/MODULE.bazel#L401)):*
> ```Starlark
> )
> 
> archive_override(
> ```

nit: Add a DO_NOT_SUBMIT comment referencing the pending PR until you switch to the version.
___
*[`src/test/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulatorTest.kt` line 21 at r12](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2591#-OXVe1V98MGjshfPLOuX:-OXVe1V98MGjshfPLOuY:b6vqkik) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5f514349cc5d36de61225aa5a96dc61145aeb080/src/test/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulatorTest.kt#L21)):*
> ```Kotlin
> import com.google.common.truth.extensions.proto.ProtoTruth.assertThat
> import com.google.protobuf.ByteString
> import com.google.protobuf.kotlin.toByteStringUtf8
> ```

Unrelated change?
___
*[`src/test/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/trustee/BUILD.bazel` line 4 at r12](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2591#-OXVcV40Ct31zUVLCwuH:-OXVcV40Ct31zUVLCwuI:b-7wri3y) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5f514349cc5d36de61225aa5a96dc61145aeb080/src/test/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/trustee/BUILD.bazel#L4)):*
> ```Starlark
> 
> kt_jvm_test(
>     name = "fulfill_requisition_request_builder_test",
> ```

JVM test targets have UpperCamelCase names.



_Suggestion:_
```Starlark
FulfillRequisitionRequestBuilderTest
```
___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/trustee/FulfillRequisitionRequestBuilder.kt` line 125 at r12](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2591#-OXVcAHI8AS0Bb0Edp3m:-OXVcAHI8AS0Bb0Edp3n:b-u3us7b) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5f514349cc5d36de61225aa5a96dc61145aeb080/src/main/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/trustee/FulfillRequisitionRequestBuilder.kt#L125)):*
> ```Kotlin
>             null,
>           )
>           .toList()
> ```

Drop the toList, otherwise you're losing the whole point of streaming and getting all the ciphertext chunks in memory.



_Code quote:_
```Kotlin
.toList()
```

roaminggypsy (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2591)* status: 26 of 31 files reviewed, 3 unresolved discussions (waiting on @kungfucraig, @renjiezh, and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/trustee/FulfillRequisitionRequestBuilder.kt` line 125 at r12](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2591#-OXVcAHI8AS0Bb0Edp3m:-OXZvthMCIgBiUSUdCuB:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5f514349cc5d36de61225aa5a96dc61145aeb080/src/main/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/trustee/FulfillRequisitionRequestBuilder.kt#L125)):*

Drop the toList, otherwise you're losing the whole point of streaming and getting all the ciphertext chunks in memory.
</blockquote></details>

Done.
___
*[`src/test/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/trustee/BUILD.bazel` line 4 at r12](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2591#-OXVcV40Ct31zUVLCwuH:-OXZw7HW6bN9I1qbSv4R:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5f514349cc5d36de61225aa5a96dc61145aeb080/src/test/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/trustee/BUILD.bazel#L4)):*

JVM test targets have UpperCamelCase names.
</blockquote></details>

Done.
___
*[`src/test/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulatorTest.kt` line 21 at r12](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2591#-OXVe1V98MGjshfPLOuX:-OX_-6uJ5t1ROZYD85VT:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5f514349cc5d36de61225aa5a96dc61145aeb080/src/test/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulatorTest.kt#L21)):*

Unrelated change?
</blockquote></details>

Done.

roaminggypsy (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2591)* status: 28 of 32 files reviewed, 1 unresolved discussion (waiting on @kungfucraig, @renjiezh, and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/common/BUILD.bazel` line 8 at r13](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2591#-OXywtfj0VOhjZFx86UJ:-OY3ZwKlBB3wS9UlpU_Z:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/586d455d22011821979d6087ede811fbb1a680a8/src/main/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/common/BUILD.bazel#L8)):*

I thought that outside of tests and inter-dependencies within the requisition package, we were going to have everything depend on the top-level target (`//src/main/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha`). Therefore this shouldn't need visibility expansion.
</blockquote></details>

Done.

SanjayVas (approved):
> @SanjayVas reviewed 5 of 5 files at r14, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2591)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @kungfucraig and @renjiezh)

roaminggypsy (commented):
> @roaminggypsy reviewed 2 of 17 files at r2, 18 of 23 files at r8, 2 of 8 files at r11, 4 of 6 files at r13, 4 of 5 files at r14.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2591)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @roaminggypsy)

PR #2626, opened 2025-07-23
ci: Add slack notificacions of nightly build runs

laureanobrs:


SanjayVas (changes requested):
> Reviewed 2 of 4 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2626)* status: 2 of 5 files reviewed, 4 unresolved discussions (waiting on @laureanobrs)

___
*[`.github/workflows/test-slack-integration.yml` line 15 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2626#-OX5HBZN-F3DDM5EVRrZ:-OXQcWVo5Xqn_ZeMKISl:b328wtd) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8fcd55f46359b51881272ca3efbeaa039b4041d2/.github/workflows/test-slack-integration.yml#L15)):*

I have mixed feelings about keeping this file. On the one hand, I’d prefer to delete it before merging; on the other, I’d rather keep it so we can test potential future changes. What do you think?
</blockquote></details>

We generally don't merge testing workflows. We do exactly this where we add a temporary workflow that doesn't get merged.
___
*[`.github/actions/send-nightly-build-notification/action.yml` line 21 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2626#-OXQeCpcDEhKtLtskxoL:-OXQeCpcDEhKtLtskxoM:b-3bybdi) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8fcd55f46359b51881272ca3efbeaa039b4041d2/.github/actions/send-nightly-build-notification/action.yml#L21)):*
> ```YAML
>     type: string
>     required: true
>   housekeeping_channel_id:
> ```

nit: use hyphens for action inputs



_Suggestion:_
```YAML
housekeeping-channel-id
```
___
*[`.github/workflows/update-cmms.yml` line 52 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2626#-OXQd3V4CXlJjZrUWotu:-OXQd3V4CXlJjZrUWotv:b-qusglc) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8fcd55f46359b51881272ca3efbeaa039b4041d2/.github/workflows/update-cmms.yml#L52)):*
> ```YAML
>         description: "Tag of existing published images (skips push)"
>         type: string
>       send-slack-notification:
> ```

If you have this input here, anyone can check it when manually running this workflow. I don't think we want that possibility.

SanjayVas (changes requested):
> Reviewed 1 of 4 files at r3, 1 of 3 files at r5.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2626)* status: 3 of 5 files reviewed, 3 unresolved discussions (waiting on @laureanobrs)

___
*[`.github/workflows/update-cmms.yml` line 52 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2626#-OXQd3V4CXlJjZrUWotu:-OXV1JOXEebN5LAPXfJP:bx8wbuk) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8fcd55f46359b51881272ca3efbeaa039b4041d2/.github/workflows/update-cmms.yml#L52)):*
> Right now it would send a notification both when deploy-nightly is run manually and when it’s run by the schedule.  
I understand we don’t want notifications for manual runs.

I'm actually fine with notifications for manually triggered runs. e.g. if you trigger a rerun or run at a new commit, it's probably still fine to notify everyone that nightly was pushed or failed. The main thing is only notifying from the deploy-nightly workflow, as opposed to someone running update-cmms that just happens to be running it from the nightly tag.

Therefore (2) is my preferred solution.

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2626)* status: 4 of 5 files reviewed, 1 unresolved discussion (waiting on @laureanobrs)

___
*[`.github/workflows/update-cmms.yml` line 169 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2626#-OXyLOOw87zbM7KwJpwt:-OXzj4oW9pstjsylFDmW:bqikixd) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/32893280d8b2249313e49595733897350cafc5a2/.github/workflows/update-cmms.yml#L169)):*

I’m doing a quick check-in so I don’t show up with something completely different without having discussed it first.

I’m not finding a function that would let me do something like `allMatch("success", needs.*.result)`  .
The alternative would be to check that it doesn’t contain any of the other statuses, like:
```
${{ !contains(needs.*.result, 'failure')
     && !contains(needs.*.result, 'cancelled')
     && !contains(needs.*.result, 'skipped') }}
```

But that would mean if a new status is introduced, the notifications could be incorrect. If that sounds good enough, I’ll make that change.

Otherwise, another option would be to split this job into a separate workflow, with `event = workflow_run`. I think it could work, but I haven’t tested it.

What do you think?
</blockquote></details>

You only need to check success. It's fine IMO to enumerate the individual needs such that if another need gets introduced the condition would need to be updated.

e.g.

```yaml
success: ${{ needs.publish-images.result == 'success' && needs.run-tests.result == 'success' && ... }}
```

laureanobrs (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2626)* status: 3 of 5 files reviewed, 1 unresolved discussion (waiting on @Copilot and @SanjayVas)

___
*[`.github/workflows/update-cmms.yml` line 169 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2626#-OXyLOOw87zbM7KwJpwt:-OY26ped4woZAB07hX6U:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/32893280d8b2249313e49595733897350cafc5a2/.github/workflows/update-cmms.yml#L169)):*

You only need to check success. It's fine IMO to enumerate the individual needs such that if another need gets introduced the condition would need to be updated.

e.g.

```yaml
success: ${{ needs.publish-images.result == 'success' && needs.run-tests.result == 'success' && ... }}
```
</blockquote></details>

Done.

SanjayVas (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2626)* status: all files reviewed, 1 unresolved discussion (waiting on @laureanobrs)

___
*[`.github/workflows/slack-nightly-build-notification.yml` line 21 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2626#-OY2MNgj3hWEwjAg5Ot1:-OY2MNgj3hWEwjAg5Ot2:bx83myh) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1e54343d350f2ad113200f7d9412ce9b88f728c7/.github/workflows/slack-nightly-build-notification.yml#L21)):*
> ```YAML
>     inputs:
>       image-tag:
>         description: "Nightly build tag"
> ```

nit: Disambiguate between the image tag and the Git tag for the nightly build. The Git tag has a slash, where the image tag has a hyphen.



_Suggestion:_
```YAML
Image tag for nightly build
```

laureanobrs (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2626)* status: 2 of 5 files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

___
*[`.github/workflows/slack-nightly-build-notification.yml` line 21 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2626#-OY2MNgj3hWEwjAg5Ot1:-OY2nChE5SHrnYP2j6tV:b-khgb5) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1e54343d350f2ad113200f7d9412ce9b88f728c7/.github/workflows/slack-nightly-build-notification.yml#L21)):*

nit: Disambiguate between the image tag and the Git tag for the nightly build. The Git tag has a slash, where the image tag has a hyphen.
</blockquote></details>

Good point! Actually, it was a bug: we need the git tag because we are including some links to GH on the notification body.

PR #2627, opened 2025-07-23
feat: implement TrusTEE cryptor

renjiezh:


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627)

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627)* status: 13 of 22 files reviewed, 12 unresolved discussions (waiting on @kungfucraig, @ple13, and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/TrusTeeMill.kt` line 149 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627#-OWuQ_mxEiCYKTOn_J3f:-OWvSSr400OKPWxmjRIj:b-lgrzxp) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a6b88fc223a040ebfd36da029c9e2087ae3aa98d/src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/TrusTeeMill.kt#L149)):*

When do we expect the dpParams to be null?
</blockquote></details>

Good question. 
This is basically for test purpose. The public API requires dp params.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/TrusTeeMill.kt` line 230 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627#-OWuS4eBDAT009oXPr1o:-OWwQ9vu4PDFt1ISunyb:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a6b88fc223a040ebfd36da029c9e2087ae3aa98d/src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/TrusTeeMill.kt#L230)):*

You should do this conversion on the fly in TrusTeeProcessorImpl on line 70.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorImpl.kt` line 34 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627#-OWuRFTF0KiPolr92un0:-OWwQB8VB0b60wvr5mmI:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a6b88fc223a040ebfd36da029c9e2087ae3aa98d/src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorImpl.kt#L34)):*

You could use a byte array here and require max frequency <= 127.

I realize that for a single computation the memory savings doesn't matter that much, but we will likely reuse this code when we build out the "multi-measurement" functionality. So it sets a good precedent.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorImpl.kt` line 52 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627#-OWuRa0-50fkpsG-WL7t:-OWvTQWW5KEJmKK88VRr:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a6b88fc223a040ebfd36da029c9e2087ae3aa98d/src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorImpl.kt#L52)):*

require >=2 for frequency computations?
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorImpl.kt` line 58 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627#-OWv8LFU7bwuN_HWaMyf:-OWvTqFs7as3FBE4iS0x:bi3ppc6) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a6b88fc223a040ebfd36da029c9e2087ae3aa98d/src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorImpl.kt#L58)):*

That could happen as part of the addition loop.
</blockquote></details>

Done.
We don't need to worry about maxFrequency. Processor with cap the max. 
But I do need to check if it is negative. I added that during conversion from bytes to int.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorImpl.kt` line 58 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627#-OWuSjrk0BE-ox4t-Uid:-OWwQCB25heYlAn_vv1k:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a6b88fc223a040ebfd36da029c9e2087ae3aa98d/src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorImpl.kt#L58)):*

Pass a byte array or byte string here.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorImpl.kt` line 71 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627#-OWv6HETDu6JPAcPkxXh:-OWvUAhl68MRq7zU8IEp:bp1r27n) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a6b88fc223a040ebfd36da029c9e2087ae3aa98d/src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorImpl.kt#L71)):*

While in C++ this is true, in Java I believe the output of the addition of two bytes is an Integer, so we don't need to worry about overflow. Also, in Java bytes are signed, so the max value is 127.
</blockquote></details>

We only use 8-bit for data in transmission.  AggregatedFrequencyVector is still 32-bit. 
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorImpl.kt` line 109 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627#-OWuUj577z-jrF8vXUDX:-OWwQhTuEowHAiaow_Fs:b-kfmf8o) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a6b88fc223a040ebfd36da029c9e2087ae3aa98d/src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorImpl.kt#L109)):*

You should have explicit unit tests for computeReach and computeFrequencyDistribution. That will mean pulling them out into a separate module.
</blockquote></details>

Done.
Extracted them in Utility class and also added test for it.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorImpl.kt` line 110 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627#-OWvBWGq1EU5e75KjKoX:-OWwd0QM0AzHePLG8hGu:b-r7t2to) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a6b88fc223a040ebfd36da029c9e2087ae3aa98d/src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorImpl.kt#L110)):*

nit: rawReach may be misleading as this is not reach.
</blockquote></details>

Done.
Renamed to reachInSample.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorImpl.kt` line 116 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627#-OWuVFHk1ViJBEEphOYt:-OWwTlS_3YKzLIuNg2P9:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a6b88fc223a040ebfd36da029c9e2087ae3aa98d/src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorImpl.kt#L116)):*

Use keyword args for the constants or pull them out as vals.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorImpl.kt` line 116 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627#-OWvC2WaBGMHSrQ8GkIN:-OWwdFI-EM3yU2HeomTi:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a6b88fc223a040ebfd36da029c9e2087ae3aa98d/src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorImpl.kt#L116)):*

nit: noiseReach may be misleading as well.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorImpl.kt` line 118 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627#-OWvAdP58hMxWE40xPid:-OWwbb6YCEL75JmyI9ti:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a6b88fc223a040ebfd36da029c9e2087ae3aa98d/src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorImpl.kt#L118)):*

clip the noisedReach to the range [0, length of frequency vector].
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorImpl.kt` line 134 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627#-OWvCVzIEnljefr_4yAP:-OWwdHSM2eW9OdeXKew6:b-sxe4an) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a6b88fc223a040ebfd36da029c9e2087ae3aa98d/src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorImpl.kt#L134)):*

activeRegisters?
</blockquote></details>

Done.
Renamed as totalSampledUsers
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorImpl.kt` line 137 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627#-OWuggmF25PPoX3ytuYr:-OWwbkJM72D0CLmXUywq:b863gl8) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a6b88fc223a040ebfd36da029c9e2087ae3aa98d/src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorImpl.kt#L137)):*

+1
</blockquote></details>

the histogram has values for frequency from 0 to max_freq: [0, max_freq].
Why do we want to remove 0's count? 
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorImpl.kt` line 144 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627#-OWuVP-E8rA9MFZ7XXQH:-OWwTkd35TW1x3hwPGJ_:b3pz36q) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a6b88fc223a040ebfd36da029c9e2087ae3aa98d/src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorImpl.kt#L144)):*

Keyword args here too.
</blockquote></details>

Done.
Interestingly, the java library does not keep the keyword so I can only use comment for that.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorImpl.kt` line 158 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627#-OWufFUc4trH12Yq5nQI:-OWwYXjGDCe5YG3HGz79:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a6b88fc223a040ebfd36da029c9e2087ae3aa98d/src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorImpl.kt#L158)):*

This one should get its own unit tests too.
</blockquote></details>

Done.

SanjayVas (changes requested):
> Reviewed 3 of 8 files at r3, 12 of 17 files at r4, 1 of 9 files at r5.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627)* status: 14 of 22 files reviewed, 17 unresolved discussions (waiting on @kungfucraig and @ple13)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorImpl.kt` line 34 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627#-OWuRFTF0KiPolr92un0:-OWwdlpx4txxBSdBi_tf:b-58wsyf) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a6b88fc223a040ebfd36da029c9e2087ae3aa98d/src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorImpl.kt#L34)):*

You could use a byte array here and require max frequency <= 127.

I realize that for a single computation the memory savings doesn't matter that much, but we will likely reuse this code when we build out the "multi-measurement" functionality. So it sets a good precedent.
</blockquote></details>

If you want to do things with bytes, use a ByteBuffer instead so it's expandable. This has the added benefit of being able to declare the property as a `val`.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorImpl.kt` line 109 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627#-OWuUj577z-jrF8vXUDX:-OWwgfba1dVv2yntKxDS:b-tf6n9y) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a6b88fc223a040ebfd36da029c9e2087ae3aa98d/src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorImpl.kt#L109)):*

Done.
Extracted them in Utility class and also added test for it.
</blockquote></details>

Why can't these be covered by exercising the public API? Obviously if this code is intended to be used separately that's a different story.

Just making sure we're not extracting something that should be considered an implementation detail.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/herald/TrusTeeStarter.kt` line 136 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627#-OWwbvBu8JhfUN0Y-tyA:-OWwbvBu8JhfUN0Y-tyB:b4fyekb) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a6b88fc223a040ebfd36da029c9e2087ae3aa98d/src/main/kotlin/org/wfanet/measurement/duchy/herald/TrusTeeStarter.kt#L136)):*
> ```Kotlin
>   }
> 
>   private fun MeasurementSpec.toTrusTeeType(): TrusTee.ComputationDetails.Type {
> ```

nit: Use the lowest receiver type that makes sense



_Suggestion:_
```Kotlin
MeasurementSpec.MeasurementTypeCase
```
___
*[`src/test/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/BUILD.bazel` line 20 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627#-OWwgH4k7QEXGCC-zSD1:-OWwgH4k7QEXGCC-zSD2:bjl2z6b) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a6b88fc223a040ebfd36da029c9e2087ae3aa98d/src/test/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/BUILD.bazel#L20)):*
> ```Starlark
> )
> 
> java_test(
> ```

If you're using `java_test` to deal with the native deps bug, add a TODO that references it. Otherwise use `kt_jvm_test` instead of the separate library and test targets.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/TrusTeeMill.kt` line 122 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627#-OWwcAbH6ZQA1uDj-Jwr:-OWwcAbH6ZQA1uDj-Jws:b941c13) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a6b88fc223a040ebfd36da029c9e2087ae3aa98d/src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/TrusTeeMill.kt#L122)):*
> ```Kotlin
> 
>   private suspend fun computingPhase(token: ComputationToken): ComputationToken {
>     val trusTeeDetails = token.computationDetails.trusTee
> ```

nit: explicit types
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorImpl.kt` line 36 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627#-OWwegNI5A1crwfm-tTf:-OWwegNI5A1crwfm-tTg:be7pofl) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a6b88fc223a040ebfd36da029c9e2087ae3aa98d/src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorImpl.kt#L36)):*
> ```Kotlin
>   private var aggregatedFrequencyVector: IntArray? = null
> 
>   private var maxFrequency by Delegates.notNull<Int>()
> ```

Why is this a `var` if it's set within the `init` block?
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorUtility.kt` line 42 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627#-OWwhk853m4KDix15xHT:-OWwhk853m4KDix15xHU:b-giwbo9) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/87e44f63c55172c83e4c7a2adf4aa46ad18c3438/src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorUtility.kt#L42)):*
> ```Kotlin
>  *   is the count of users with that frequency.
>  */
> fun buildHistogram(frequencyVector: IntArray, maxFrequency: Int): LongArray {
> ```

No top-level functions that are not extensions in new code. Put this in an object.

Generally if you have a "Utility" or "Utils" file in our Kotlin code, it's a smell.

ple13 (changes requested):
> Reviewed 9 of 9 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627)* status: all files reviewed, 17 unresolved discussions (waiting on @kungfucraig and @renjiezh)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorImpl.kt` line 137 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627#-OWuggmF25PPoX3ytuYr:-OWywVxZE0bApKusWD-M:bc5jmwz) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a6b88fc223a040ebfd36da029c9e2087ae3aa98d/src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorImpl.kt#L137)):*

the histogram has values for frequency from 0 to max_freq: [0, max_freq].
Why do we want to remove 0's count?
</blockquote></details>

By definition, reach equals sum of frequency counts. So, the frequency starts at 1.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorUtility.kt` line 25 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627#-OX-UkqZAbeoUti4i0UN:-OX-UkqZAbeoUti4i0UO:bqgag19) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/87e44f63c55172c83e4c7a2adf4aa46ad18c3438/src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorUtility.kt#L25)):*
> ```Kotlin
>   return this.map { byte ->
>       val frequency = byte.toInt()
>       require(frequency >= 0 && frequency < 127) {
> ```

nit: declare a constant for 127?
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorImpl.kt` line 176 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627#-OX-U9me7L2GukTzO_vv:-OX-U9mf6nLMjMRO6sP8:bpvk9fr) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/87e44f63c55172c83e4c7a2adf4aa46ad18c3438/src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorImpl.kt#L176)):*
> ```Kotlin
>     val totalNoisedCount = noisedHistogram.sum()
>     if (totalNoisedCount == 0L) {
>       return (0..maxFrequency).associate { it.toLong() to 0.0 }
> ```

return (1..maxFrequency).associate { it.toLong() to 0.0 }



_Code quote:_
```Kotlin
return (0..maxFrequency).associate { it.toLong() to 0.0 }
```

ple13 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627)* status: all files reviewed, 17 unresolved discussions (waiting on @kungfucraig and @renjiezh)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorImpl.kt` line 137 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627#-OWuggmF25PPoX3ytuYr:-OX-X3pL6WHyt3vnOTtT:b5z9hjz) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a6b88fc223a040ebfd36da029c9e2087ae3aa98d/src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorImpl.kt#L137)):*

I see. thanks. 
The histogram is not necessarily defined as starting at 1. Starting from 0 makes it easier to get totalSampledUsers and freq ratio. I think we can keep the histogram as-is. But I will update the frequency result to not include freq 0. Wdyt?
</blockquote></details>

The metrics service takes input as a map of i to frequency[i] / (frequency[1] + ... + frequency[maxF]) for i = 1...maxF. So, we need to exclude frequency 0 before computing the relative frequency histogram.

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627)* status: all files reviewed, 17 unresolved discussions (waiting on @kungfucraig, @ple13, and @renjiezh)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorUtility.kt` line 25 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627#-OX-UkqZAbeoUti4i0UN:-OX0Gy_g2xnx_2jclQzT:b-gosfmk) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/87e44f63c55172c83e4c7a2adf4aa46ad18c3438/src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorUtility.kt#L25)):*

nit: declare a constant for 127?
</blockquote></details>

Said constant already exists: `Byte.MAX_VALUE`

kungfucraig (changes requested):
> Reviewed 3 of 17 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627)* status: all files reviewed, 15 unresolved discussions (waiting on @ple13, @renjiezh, and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/TrusTeeMill.kt` line 149 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627#-OWuQ_mxEiCYKTOn_J3f:-OX04uNqBgVdPwR2Fz0N:b3isxlo) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a6b88fc223a040ebfd36da029c9e2087ae3aa98d/src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/TrusTeeMill.kt#L149)):*

Good question. 
This is basically for test purpose. The public API requires dp params.
</blockquote></details>

I'd recommend that we require dpParams in this module, but that 
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorImpl.kt` line 34 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627#-OWuRFTF0KiPolr92un0:-OX0ATX_9vI9vewTbja9:bfne8ui) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a6b88fc223a040ebfd36da029c9e2087ae3aa98d/src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorImpl.kt#L34)):*

If you want to do things with bytes, use a ByteBuffer instead so it's expandable. This has the added benefit of being able to declare the property as a `val`.
</blockquote></details>

I don't think ByteBuffer is the right abstraction here since we are dealing strictly with an array of numbers that we are doing arithmetic on. Thinking about it further... IntArray is probably the best so we can avoid the computational churn of byte-to-int conversions, which we would incur both when aggregating and during histogram construction.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorImpl.kt` line 109 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627#-OWuUj577z-jrF8vXUDX:-OX0l6CvCg-ipbqhabCE:b-i9eiln) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a6b88fc223a040ebfd36da029c9e2087ae3aa98d/src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorImpl.kt#L109)):*

Why can't these be covered by exercising the public API? Obviously if this code is intended to be used separately that's a different story.

Just making sure we're not extracting something that should be considered an implementation detail.
</blockquote></details>

I see three reasons to extract these:

1. They should get reused by the EDP Aggregator direct measurement functions. Currently the EDP Aggregator has implementations that use the Halo noise libraries, which have known issues. 

2. We should as a matter of policy isolate any code that does anything numerical and ensure it has its own tests. This allows us to be much more through when we test, possibly bypassing bounds checks that clients do, etc.

3. While, we can, and for testability should, make the DP params optional in the utility module, we should make the DP params required in the TrustTEE mill. This is more a matter of ensuring there's no way to trick the system level code into computing these things without DP. 
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorImpl.kt` line 134 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627#-OWvCVzIEnljefr_4yAP:-OX0kjYb1YJkSH1q7XGr:bvj2oe5) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a6b88fc223a040ebfd36da029c9e2087ae3aa98d/src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorImpl.kt#L134)):*

Done.
Renamed as totalSampledUsers
</blockquote></details>

I hate to nit pick... but totalSampledVids would be more accurate.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorImpl.kt` line 137 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627#-OWuggmF25PPoX3ytuYr:-OX0kuLA0Bn5piaB9QFm:b-owt1kf) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a6b88fc223a040ebfd36da029c9e2087ae3aa98d/src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorImpl.kt#L137)):*

Oh no. I didn't even know that. Cool. I will update the histogram.
</blockquote></details>

Also if we expose frequency 0, we'd be exposing a second estimate of reach and thus implicitly consuming more privacy budget.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorImpl.kt` line 44 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627#-OX0A01J8MeT8lbP9oVg:-OX0A01K4QpXK2xwLfhC:b-y77anj) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/87e44f63c55172c83e4c7a2adf4aa46ad18c3438/src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorImpl.kt#L44)):*
> ```Kotlin
>       is TrusTeeReachAndFrequencyParams -> {
>         maxFrequency = trusTeeParams.maximumFrequency
>         require(maxFrequency >= 2) { "Invalid max frequency: $maxFrequency" }
> ```

state that max frequency must be greater than 2 in the message'''

Also require max frequency <= 127.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorImpl.kt` line 61 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627#-OX0X8-w-8xqiZ_nWR6t:-OX0X8-w-8xqiZ_nWR6u:blzrihe) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/87e44f63c55172c83e4c7a2adf4aa46ad18c3438/src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorImpl.kt#L61)):*
> ```Kotlin
>     require(bytes.isNotEmpty()) { "Input frequency vector cannot be empty." }
> 
>     val vector = bytes.toIntArray()
> ```

You don't need to convert this.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorUtility.kt` line 1 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627#-OX0nQbn-1VttuNBEb-J:-OX0nQbn-1VttuNBEb-K:b-4f52ub) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/87e44f63c55172c83e4c7a2adf4aa46ad18c3438/src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorUtility.kt#L1)):*
> ```Kotlin
> // Copyright 2025 The Cross-Media Measurement Authors
> ```

Maybe rename this file to "ReachAndFrequencyComputations" and have an object with the same name that encapsulates the functions?

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627)* status: 14 of 24 files reviewed, 12 unresolved discussions (waiting on @kungfucraig, @ple13, and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/TrusTeeMill.kt` line 149 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627#-OWuQ_mxEiCYKTOn_J3f:-OX1wTPTCCynwjghha41:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a6b88fc223a040ebfd36da029c9e2087ae3aa98d/src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/TrusTeeMill.kt#L149)):*

I'd recommend that we require dpParams in this module, but that
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorImpl.kt` line 34 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627#-OWuRFTF0KiPolr92un0:-OX1wVYY7i914jWt--bl:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a6b88fc223a040ebfd36da029c9e2087ae3aa98d/src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorImpl.kt#L34)):*

I don't think ByteBuffer is the right abstraction here since we are dealing strictly with an array of numbers that we are doing arithmetic on. Thinking about it further... IntArray is probably the best so we can avoid the computational churn of byte-to-int conversions, which we would incur both when aggregating and during histogram construction.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorImpl.kt` line 36 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627#-OWwegNI5A1crwfm-tTf:-OX0NnrX4XiKUsnWZlOb:bxq2hv9) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a6b88fc223a040ebfd36da029c9e2087ae3aa98d/src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorImpl.kt#L36)):*

Why is this a `var` if it's set within the `init` block?
</blockquote></details>

Done.
Thanks for catching it.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorImpl.kt` line 109 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627#-OWuUj577z-jrF8vXUDX:-OX1wdcm06dVlv0MsCwU:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a6b88fc223a040ebfd36da029c9e2087ae3aa98d/src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorImpl.kt#L109)):*

I see three reasons to extract these:

1. They should get reused by the EDP Aggregator direct measurement functions. Currently the EDP Aggregator has implementations that use the Halo noise libraries, which have known issues. 

2. We should as a matter of policy isolate any code that does anything numerical and ensure it has its own tests. This allows us to be much more through when we test, possibly bypassing bounds checks that clients do, etc.

3. While, we can, and for testability should, make the DP params optional in the utility module, we should make the DP params required in the TrustTEE mill. This is more a matter of ensuring there's no way to trick the system level code into computing these things without DP.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorImpl.kt` line 137 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627#-OWuggmF25PPoX3ytuYr:-OX1wN2K9AZZXcrG1W-_:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a6b88fc223a040ebfd36da029c9e2087ae3aa98d/src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorImpl.kt#L137)):*

Also if we expose frequency 0, we'd be exposing a second estimate of reach and thus implicitly consuming more privacy budget.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorImpl.kt` line 44 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627#-OX0A01J8MeT8lbP9oVg:-OX1x7Ta3w9CUlrDfpDf:b2ni02g) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/87e44f63c55172c83e4c7a2adf4aa46ad18c3438/src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorImpl.kt#L44)):*

state that max frequency must be greater than 2 in the message'''

Also require max frequency <= 127.
</blockquote></details>

byt
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorImpl.kt` line 61 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627#-OX0X8-w-8xqiZ_nWR6t:-OX1wcUM3VpsNEQjlf9O:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/87e44f63c55172c83e4c7a2adf4aa46ad18c3438/src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorImpl.kt#L61)):*

You don't need to convert this.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorImpl.kt` line 176 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627#-OX-U9me7L2GukTzO_vv:-OX1wNmpCS1u1OnumufJ:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/87e44f63c55172c83e4c7a2adf4aa46ad18c3438/src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorImpl.kt#L176)):*

return (1..maxFrequency).associate { it.toLong() to 0.0 }
</blockquote></details>

Done.
___
*[`src/test/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/BUILD.bazel` line 20 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627#-OWwgH4k7QEXGCC-zSD1:-OX0B_vZ9khd61uidv1I:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a6b88fc223a040ebfd36da029c9e2087ae3aa98d/src/test/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/BUILD.bazel#L20)):*

If you're using `java_test` to deal with the native deps bug, add a TODO that references it. Otherwise use `kt_jvm_test` instead of the separate library and test targets.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorUtility.kt` line 1 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627#-OX0nQbn-1VttuNBEb-J:-OX1wP662_nxISMpLT71:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/87e44f63c55172c83e4c7a2adf4aa46ad18c3438/src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorUtility.kt#L1)):*

Maybe rename this file to "ReachAndFrequencyComputations" and have an object with the same name that encapsulates the functions?
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorUtility.kt` line 42 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627#-OWwhk853m4KDix15xHT:-OX1wQMu50-7uQImSTNo:bstqrkd) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/87e44f63c55172c83e4c7a2adf4aa46ad18c3438/src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorUtility.kt#L42)):*

No top-level functions that are not extensions in new code. Put this in an object.

Generally if you have a "Utility" or "Utils" file in our Kotlin code, it's a smell.
</blockquote></details>

Done.
Refactored and renamed.

SanjayVas (approved):
> Reviewed 2 of 17 files at r4, 1 of 9 files at r5, 10 of 10 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627)* status: all files reviewed, 14 unresolved discussions (waiting on @kungfucraig and @ple13)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/ReachAndFrequencyCalculator.kt` line 21 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627#-OX4xyXx7AgYumWWFKd4:-OX4xyXx7AgYumWWFKd5:b9u93kd) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/bf86c0cc1b4905d1936e3ee0023a6d11addbd05a/src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/ReachAndFrequencyCalculator.kt#L21)):*
> ```Kotlin
> import org.wfanet.measurement.internal.duchy.DifferentialPrivacyParams
> 
> object ReachAndFrequencyCalculator {
> ```

nit: Avoid naming utility objects as "agent nouns". The equivalent object for deterministic methodologies used by the EDP simulator is `MeasurementResults`.



_Suggestion:_
```Kotlin
ReachAndFrequencyComputations
```
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessor.kt` line 62 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627#-OX4zDH16NEDrwkobQUD:-OX4zDH16NEDrwkobQUE:b-i0b7jd) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/bf86c0cc1b4905d1936e3ee0023a6d11addbd05a/src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessor.kt#L62)):*
> ```Kotlin
>    *   non-negative 8-bit signed integer representing a frequency.
>    */
>   fun addFrequencyVectorBytes(bytes: ByteArray)
> ```

nit: could just call this `addFrequencyVector` and name the parameter. The parameter type already tells you that it's bytes.



_Suggestion:_
```Kotlin
fun addFrequencyVector(frequencyVector: ByteArray)
```
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorImpl.kt` line 31 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627#-OX4Vjd614soI_t4wzdE:-OX4Vjd614soI_t4wzdF:b-9a9s63) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/bf86c0cc1b4905d1936e3ee0023a6d11addbd05a/src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorImpl.kt#L31)):*
> ```Kotlin
>    * Subsequent calls add to this vector. It is null until the first vector is added.
>    */
>   private var aggregatedFrequencyVector: IntArray? = null
> ```

nit: consider using a lateinit var if it goes from null to non-null exactly once.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorImpl.kt` line 40 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627#-OX4z_pVD5vGcbghFUJa:-OX4z_pVD5vGcbghFUJb:buvlfu9) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/bf86c0cc1b4905d1936e3ee0023a6d11addbd05a/src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorImpl.kt#L40)):*
> ```Kotlin
>       is TrusTeeReachAndFrequencyParams -> {
>         maxFrequency = trusTeeParams.maximumFrequency
>         require(maxFrequency >= 2 && maxFrequency <= Byte.MAX_VALUE) {
> ```

nit: Prefer idiomatic Kotlin for closed ranges



_Suggestion:_
```Kotlin
maxFrequency in 2..Byte.MAX_VALUE
```
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorImpl.kt` line 83 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627#-OX5-pNj5iADcJhJXwEW:-OX5-pNk3b1-8saZ9Sc0:b-3e3b13) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/bf86c0cc1b4905d1936e3ee0023a6d11addbd05a/src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorImpl.kt#L83)):*
> ```Kotlin
>         ?: throw IllegalStateException(
>           "addFrequencyVectorBytes must be called before computeResult."
>         )
> ```

nit: idiomatic Kotlin

Though note that this changes if you follow my lateinit var suggestion, as you'd check isInitialized instead.



_Suggestion:_
```Kotlin
    val frequencyVector = requireNotNull(aggregatedFrequencyVector) { "addFrequencyVectorBytes must be called before computeResult." }
```

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627)* status: 15 of 24 files reviewed, 10 unresolved discussions (waiting on @kungfucraig, @ple13, and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/ReachAndFrequencyComputations.kt` line 15 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627#-OX9OriK7Ic18U83FJU1:-OX9OriK7Ic18U83FJU2:bvv1iz5) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0f146d848a40fd232d1d0340c5ff3a40ed309e36/src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/ReachAndFrequencyComputations.kt#L15)):*
> ```Kotlin
> 
> package org.wfanet.measurement.duchy.mill.trustee.processor
> ```

this needs a more generic package
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/ReachAndFrequencyComputations.kt` line 21 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627#-OX9OhKJByFPlfvxUAFs:-OX9OhKJByFPlfvxUAFt:br25vui) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0f146d848a40fd232d1d0340c5ff3a40ed309e36/src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/ReachAndFrequencyComputations.kt#L21)):*
> ```Kotlin
> import org.wfanet.measurement.internal.duchy.DifferentialPrivacyParams
> 
> object ReachAndFrequencyComputations {
> ```

can this get pulled into a separate pr?

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627)* status: 15 of 24 files reviewed, 10 unresolved discussions (waiting on @kungfucraig, @ple13, @renjiezh, and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/ReachAndFrequencyComputations.kt` line 15 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627#-OX9OriK7Ic18U83FJU1:-OX9Zx-l847GzRhPjCNJ:b-nru59m) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0f146d848a40fd232d1d0340c5ff3a40ed309e36/src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/ReachAndFrequencyComputations.kt#L15)):*

this needs a more generic package
</blockquote></details>

Working on it.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/ReachAndFrequencyComputations.kt` line 21 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627#-OX9OhKJByFPlfvxUAFs:-OX9Zw--Dpmbf93V9NKg:b-nru59m) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0f146d848a40fd232d1d0340c5ff3a40ed309e36/src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/ReachAndFrequencyComputations.kt#L21)):*

can this get pulled into a separate pr?
</blockquote></details>

Working on it.

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627)* status: 15 of 24 files reviewed, 10 unresolved discussions (waiting on @kungfucraig, @ple13, @renjiezh, and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/ReachAndFrequencyComputations.kt` line 15 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627#-OX9OriK7Ic18U83FJU1:-OX9yCR59rwkDVinV2de:bo8zcm9) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0f146d848a40fd232d1d0340c5ff3a40ed309e36/src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/ReachAndFrequencyComputations.kt#L15)):*

Working on it.
</blockquote></details>

org.wfanet.measurement.computation

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627)* status: 15 of 24 files reviewed, 11 unresolved discussions (waiting on @kungfucraig, @ple13, @renjiezh, and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/ReachAndFrequencyComputations.kt` line 42 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627#-OXAhAiw2kNAcKPepZ7J:-OXAhAiw2kNAcKPepZ7K:b-okhx2c) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0f146d848a40fd232d1d0340c5ff3a40ed309e36/src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/ReachAndFrequencyComputations.kt#L42)):*
> ```Kotlin
> 
>     // The histogram is built only from non-zero frequencies, so its sum is the reach in the sample.
>     val reachInSample = rawHistogram.sum()
> ```

maybe I'm misunderstanding the data structure. Isn't sum going to be the reach? I would think you would want `rawHistogram.count { it != 0L }`

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627)* status: 15 of 24 files reviewed, 11 unresolved discussions (waiting on @kungfucraig, @ple13, @renjiezh, and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/ReachAndFrequencyComputations.kt` line 42 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627#-OXAhAiw2kNAcKPepZ7J:-OXAhXtx2GYJ3tGSlboG:b-bk69ll) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0f146d848a40fd232d1d0340c5ff3a40ed309e36/src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/ReachAndFrequencyComputations.kt#L42)):*

maybe I'm misunderstanding the data structure. Isn't sum going to be the reach? I would think you would want `rawHistogram.count { it != 0L }`
</blockquote></details>

sorry. what I typed wasn't clear. I mean sum is the total impressions. reach would be `.count { it != 0 }`

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627)* status: 15 of 24 files reviewed, 10 unresolved discussions (waiting on @kungfucraig, @ple13, @renjiezh, and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/ReachAndFrequencyComputations.kt` line 42 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627#-OXAhAiw2kNAcKPepZ7J:-OXAhorBA0A4a8srFS-P:buwfxi) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0f146d848a40fd232d1d0340c5ff3a40ed309e36/src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/ReachAndFrequencyComputations.kt#L42)):*

sorry. what I typed wasn't clear. I mean sum is the total impressions. reach would be `.count { it != 0 }`
</blockquote></details>

nvm - i was thinking the LongArray was the frequencyVector. I see better now from the tests.

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627)* status: 15 of 24 files reviewed, 13 unresolved discussions (waiting on @kungfucraig, @ple13, @renjiezh, and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorImpl.kt` line 34 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627#-OWuRFTF0KiPolr92un0:-OXAqaTyCpiH8hNZ26st:b-4h5gk5) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a6b88fc223a040ebfd36da029c9e2087ae3aa98d/src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorImpl.kt#L34)):*

Done.
</blockquote></details>

I've commented elsewhere but I do have concerns about frequency > 127
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/ReachAndFrequencyComputations.kt` line 35 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627#-OXArMC7BKVNgVM0-VUU:-OXArMC7BKVNgVM0-VUV:b-obc4wb) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0f146d848a40fd232d1d0340c5ff3a40ed309e36/src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/ReachAndFrequencyComputations.kt#L35)):*
> ```Kotlin
>   fun computeReach(
>     rawHistogram: LongArray,
>     vectorSize: Int,
> ```

Can you add vector size to the kdoc?
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/ReachAndFrequencyComputations.kt` line 46 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627#-OXAriRqAW0pICYxhbRr:-OXAriRqAW0pICYxhbRs:b-99ddfz) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0f146d848a40fd232d1d0340c5ff3a40ed309e36/src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/ReachAndFrequencyComputations.kt#L46)):*
> ```Kotlin
>     if (dpParams == null) {
>       val scaledReach = (reachInSample / vidSamplingIntervalWidth).toLong()
>       return min(scaledReach, maxPossibleScaledReach)
> ```

when would the scaledReach exceed the maxPossibleScaledReach. Is this just in the case where scaledReach w/ Noise is greater than the population?
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/ReachAndFrequencyComputations.kt` line 129 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627#-OXAq594Etf--wVss2y7:-OXAq594Etf--wVss2y8:b-a3i1im) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0f146d848a40fd232d1d0340c5ff3a40ed309e36/src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/ReachAndFrequencyComputations.kt#L129)):*
> ```Kotlin
>    *   frequency `k`.
>    */
>   fun buildHistogram(frequencyVector: IntArray, maxFrequency: Int): LongArray {
> ```

i capped histogram is also helpful for counting impressions. wdyt of moving this to its own file `HistogramComputations`

stevenwarejones (commented):
> Reviewed 2 of 20 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627)* status: 18 of 39 files reviewed, 14 unresolved discussions (waiting on @kungfucraig, @ple13, @renjiezh, and @SanjayVas)

stevenwarejones (changes requested):
> Reviewed 1 of 12 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627)* status: 18 of 44 files reviewed, 15 unresolved discussions (waiting on @kungfucraig, @ple13, @renjiezh, and @SanjayVas)

___
*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627#-OXTXAPe90bCadZAvzBM:-OXTXAPf5YN5OsW4_PqV:b-xp57u7) (no related file):*
can you change the base branch to `renjiez-trustee-computations`?

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627)* status: 41 of 44 files reviewed, 17 unresolved discussions (waiting on @kungfucraig, @SanjayVas, and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/TrusTeeMill.kt` line 221 at r11](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627#-OXzT6UK0FiKanHE8Zy7:-OY3grhk8tTea6BKX06Q:b-rgq97i) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8fa1b56c00b03700576bcbd694e6020c551a0ef8/src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/TrusTeeMill.kt#L221)):*

Use the StreamingAead primitive for the actual data. That's what is being used to encrypt in #2591, so you need to do the same here even if you were decrypting the entire ciphertext at once.

Of course, since this is read from blob storage and our StorageClient offers a streaming API, you should be doing the same here rather than reading the entire blob then decrypting. See [StreamingAeadStorageClient from common-jvm](https://github.com/world-federation-of-advertisers/common-jvm/blob/main/src/main/kotlin/org/wfanet/measurement/common/crypto/tink/StreamingAeadStorageClient.kt).
</blockquote></details>

Added a TODO to address this in a future PR.
https://github.com/world-federation-of-advertisers/cross-media-measurement/issues/2800
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorImpl.kt` line 116 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627#-OWuVFHk1ViJBEEphOYt:-OY3cPXr0_OSI2GTkq1B:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a6b88fc223a040ebfd36da029c9e2087ae3aa98d/src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorImpl.kt#L116)):*

See my other comment about when keyword args ([named arguments](https://kotlinlang.org/docs/functions.html#named-arguments)) should/shouldn't be used.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorImpl.kt` line 66 at r11](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627#-OXzVIT5BE9LEe4rinTO:-OY3btQP8L47svU7Uc7Z:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8fa1b56c00b03700576bcbd694e6020c551a0ef8/src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorImpl.kt#L66)):*

aggregatedFrequencyVector can never be null
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorImpl.kt` line 126 at r11](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627#-OXzWu9NAAHz1gmmhZ5d:-OY3cQIN6aX9sMVH48Rj:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8fa1b56c00b03700576bcbd694e6020c551a0ef8/src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorImpl.kt#L126)):*

Don't unnecessarily use named arguments when the variable names or types make it obvious what the arguments are.

In general, named arguments should only be used when necessary (e.g. for optional params) or to avoid ambiguity (e.g. when an argument is an unnamed literal, especially a boolean)
</blockquote></details>

Done.

SanjayVas (changes requested):
> @SanjayVas reviewed 6 of 6 files at r12, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627)* status: all files reviewed, 16 unresolved discussions (waiting on @kungfucraig, @renjiezh, and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/TrusTeeMill.kt` line 221 at r11](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627#-OXzT6UK0FiKanHE8Zy7:-OY3jsP49wv2kcpjexzZ:b-q33tnz) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8fa1b56c00b03700576bcbd694e6020c551a0ef8/src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/TrusTeeMill.kt#L221)):*

Added a TODO to address this in a future PR.
https://github.com/world-federation-of-advertisers/cross-media-measurement/issues/2800
</blockquote></details>

I don't see the TODO?
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorImpl.kt` line 80 at r12](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627#-OY3kEbwAEDPSAXPJmbS:-OY3kEbwAEDPSAXPJmbT:b-7jslk2) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/531231517a831e6fc40b212f7c9c3fa501b6e8b7/src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorImpl.kt#L80)):*
> ```Kotlin
> 
>   override fun computeResult(): ComputationResult {
>     check(!::aggregatedFrequencyVector.isInitialized) {
> ```

The condition is backwards here. You want to check that the vector is initialized.

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627)* status: 43 of 44 files reviewed, 16 unresolved discussions (waiting on @kungfucraig, @SanjayVas, and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/TrusTeeMill.kt` line 221 at r11](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627#-OXzT6UK0FiKanHE8Zy7:-OY3loJf2_TlLYge0C95:b-gvmkgf) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8fa1b56c00b03700576bcbd694e6020c551a0ef8/src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/TrusTeeMill.kt#L221)):*

I don't see the TODO?
</blockquote></details>

I placed it in the trustee mill main body where this function is called.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorImpl.kt` line 80 at r12](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627#-OY3kEbwAEDPSAXPJmbS:-OY3ldAM2tku5qKhNpML:bgiua5f) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/531231517a831e6fc40b212f7c9c3fa501b6e8b7/src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorImpl.kt#L80)):*

The condition is backwards here. You want to check that the vector is initialized.
</blockquote></details>

Oh my.
Thanks for catching it. 
Glad we have a test unit to guard this.

SanjayVas (approved):
> @SanjayVas reviewed 1 of 1 files at r13, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627)* status: all files reviewed, 14 unresolved discussions (waiting on @kungfucraig and @stevenwarejones)

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627)* status: all files reviewed, 14 unresolved discussions (waiting on @georgi, @kungfucraig, and @renjiezh)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorImpl.kt` line 58 at r13](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627#-OY7dwMPD__Se-B9Nr_k:-OY7eRmI4TnyelV1PsZ6:b-ut0yy5) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/b229833b349b2ab29dbcab30524f96fcb4603e63/src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorImpl.kt#L58)):*

does this design only allow adding one frequency vector at a time? otherwise, we need a log on the aggregatedFrequencyVEctor
</blockquote></details>

maybe just document in the kdoc on the class that its not thread-safe

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627)* status: all files reviewed, 14 unresolved discussions (waiting on @georgi, @kungfucraig, and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorImpl.kt` line 58 at r13](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627#-OY7dwMPD__Se-B9Nr_k:-OY7rxqv3z0e6ikTPsSb:bk03uiu) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/b229833b349b2ab29dbcab30524f96fcb4603e63/src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/processor/TrusTeeProcessorImpl.kt#L58)):*

maybe just document in the kdoc on the class that its not thread-safe
</blockquote></details>

It is not thread-safe. A note was added on the interface`TrusTeeProcessor`

stevenwarejones (approved):
> @stevenwarejones reviewed 3 of 8 files at r3, 1 of 17 files at r4, 2 of 10 files at r6, 1 of 9 files at r7, 1 of 17 files at r8, 16 of 20 files at r9, 6 of 10 files at r11, 3 of 6 files at r12.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627)* status: all files reviewed, 8 unresolved discussions (waiting on @georgi and @kungfucraig)

renjiezh (commented):
> @renjiezh dismissed @kungfucraig from a discussion.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2627)* status: all files reviewed, 7 unresolved discussions (waiting on @kungfucraig)

PR #2709, opened 2025-07-30
feat: Storage Event Source

georgi:
This implementation wraps EventReader to provide events from storage, conforming to the EventSource interface. It combines event batches from multiple event readers across different dates and event group reference IDs, processing them in parallel and streaming the results as they become available.

The event source iterates through each date in the specified date range and each event group reference ID, launching parallel coroutines to read events from storage for each date/event group combination. Event batches from all combinations are merged into a single output flow.

### DIFF
--- a/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/StorageEventSource.kt
+++ b/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/StorageEventSource.kt
@@ -1,0 +1,162 @@
+/*
+ * Copyright 2025 The Cross-Media Measurement Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.wfanet.measurement.edpaggregator.resultsfulfiller
+
+import com.google.crypto.tink.KmsClient
+import com.google.protobuf.Message
+import com.google.protobuf.TypeRegistry
+import java.time.LocalDate
+import java.util.logging.Logger
+import kotlin.coroutines.CoroutineContext
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.asFlow
+import kotlinx.coroutines.flow.channelFlow
+import kotlinx.coroutines.flow.toList
+import kotlinx.coroutines.launch
+import org.wfanet.measurement.common.OpenEndTimeRange
+import org.wfanet.measurement.edpaggregator.StorageConfig
+
+/**
+ * Event source implementation that reads events from storage using EventReader.
+ *
+ * This implementation wraps EventReader to provide events from storage,
+ * conforming to the EventSource interface. It combines event batches from multiple
+ * event readers across different dates and event group reference IDs, processing
+ * them in parallel and streaming the results as they become available.
+ *
+ * The event source iterates through each date in the specified date range and
+ * each event group reference ID, launching parallel coroutines to read events
+ * from storage for each date/event group combination. Event batches from all
+ * combinations are merged into a single output flow.
+ *
+ * @param eventReader The EventReader instance to use for reading events
+ * @param dateRange The date range for which to read events (inclusive)
+ * @param eventGroupReferenceIds The event group reference IDs to read
+ * @param batchSize The size of event batches to return
+ */
+class StorageEventSource(
+  private val eventReader: EventReader,
+  private val dateRange: ClosedRange<LocalDate>,
+  private val eventGroupReferenceIds: List<String>,
+  private val batchSize: Int = DEFAULT_BATCH_SIZE
+) : EventSource {
+
+  companion object {
+    private val logger = Logger.getLogger(StorageEventSource::class.java.name)
+    const val DEFAULT_BATCH_SIZE = 256
+  }
+
+  /**
+   * Secondary constructor that creates an EventReader internally.
+   */
+  constructor(
+    kmsClient: KmsClient?,
+    impressionsStorageConfig: StorageConfig,
+    impressionDekStorageConfig: StorageConfig,
+    labeledImpressionsDekPrefix: String,
+    typeRegistry: TypeRegistry,
+    dateRange: ClosedRange<LocalDate>,
+    eventGroupReferenceIds: List<String>,
+    batchSize: Int = DEFAULT_BATCH_SIZE
+  ) : this(
+    EventReader(
+      kmsClient,
+      impressionsStorageConfig,
+      impressionDekStorageConfig,
+      labeledImpressionsDekPrefix,
+      typeRegistry,
+      batchSize
+    ),
+    dateRange,
+    eventGroupReferenceIds,
+    batchSize
+  )
+
+  override suspend fun generateEventBatches(
+    dispatcher: CoroutineContext
+  ): Flow<List<LabeledEvent<out Message>>> {
+    logger.info("Starting storage-based event generation with batching")
+
+    return channelFlow {
+      // Process each date and event group combination in parallel
+      val processingJobs = mutableListOf<kotlinx.coroutines.Job>()
+
+      // Calculate total combinations for progress tracking
+      val totalDays = java.time.temporal.ChronoUnit.DAYS.between(dateRange.start, dateRange.endInclusive) + 1
+      val totalCombinations = totalDays * eventGroupReferenceIds.size
+      var processedCombinations = 0L
+      var totalEventsRead = 0L
+      var totalBatchesSent = 0L
+
+      logger.info("Processing $totalCombinations date/event group combinations (${totalDays} days × ${eventGroupReferenceIds.size} event groups)")
+
+      var currentDate = dateRange.start
+      while (currentDate <= dateRange.endInclusive) {
+        val date = currentDate
+        for (eventGroupReferenceId in eventGroupReferenceIds) {
+          val job = launch(dispatcher) {
+            logger.fine("Reading events for date $date and event group $eventGroupReferenceId")
+
+            try {
+              var batchCount = 0
+              var eventCount = 0
+
+              // Use the new batched API from EventReader
+              eventReader.getLabeledEventsBatched(date, eventGroupReferenceId, batchSize).collect { batch ->
+                send(batch)
+                batchCount++
+                eventCount += batch.size
+              }
+
+              logger.fine("Read $eventCount events in $batchCount batches for date $date and event group $eventGroupReferenceId")
+
+              // Update progress
+              synchronized(this@channelFlow) {
+                processedCombinations++
+                totalEventsRead += eventCount
+                totalBatchesSent += batchCount
+
+                // Log progress every 10% or at least every 10 combinations
+                if (processedCombinations % maxOf(1L, totalCombinations / 10) == 0L || processedCombinations == totalCombinations) {
+                  val progressPercent = (processedCombinations * 100) / totalCombinations
+                  logger.info("Progress: $progressPercent% ($processedCombinations/$totalCombinations combinations) - Total events: $totalEventsRead, Batches sent: $totalBatchesSent")
+                }
+              }
+            } catch (e: Exception) {
+              logger.warning("Failed to read events for date $date and event group $eventGroupReferenceId: ${e.message}")
+              // Continue with other dates/event groups
+              synchronized(this@channelFlow) {
+                processedCombinations++
+              }
+            }
+          }
+
+          processingJobs.add(job)
+        }
+        currentDate = currentDate.plusDays(1)
+      }
+
+      logger.info("Launched ${processingJobs.size} storage reading jobs")
+
+      // Wait for all processing to complete
+      processingJobs.forEach { it.join() }
+
+      logger.info("Completed storage event generation - Total events: $totalEventsRead, Total batches: $totalBatchesSent")
+    }
+  }
+}
--- a/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/BUILD.bazel
+++ b/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/BUILD.bazel
@@ -87,6 +87,43 @@ kt_jvm_library(
 )
 
 kt_jvm_library(
+    name = "event_source",
+    srcs = ["EventSource.kt"],
+    deps = [
+        ":labeled_event",
+        "@wfa_common_jvm//imports/java/com/google/protobuf",
+        "@wfa_common_jvm//imports/kotlin/kotlinx/coroutines:core",
+    ],
+)
+
+kt_jvm_library(
+    name = "storage_event_source",
+    srcs = ["StorageEventSource.kt"],
+    deps = [
+        ":event_reader",
+        ":event_source",
+        ":labeled_event",
+        "//src/main/kotlin/org/wfanet/measurement/edpaggregator:storage_config",
+        "@wfa_common_jvm//imports/java/com/google/protobuf",
+        "@wfa_common_jvm//imports/kotlin/kotlinx/coroutines:core",
+        "@wfa_common_jvm//src/main/kotlin/org/wfanet/measurement/common/crypto/tink",
+        "@wfa_common_jvm//src/main/kotlin/org/wfanet/measurement/storage:selected_storage_client",
+    ],
+)
+
+kt_jvm_library(
+    name = "filter_processor",
+    srcs = ["FilterProcessor.kt"],
+    deps = [
+        ":event_batch",
+        ":labeled_event",
+        "//src/main/kotlin/org/wfanet/measurement/eventdataprovider/eventfiltration:event_filters",
+        "@wfa_common_jvm//imports/java/com/google/protobuf",
+        "@wfa_common_jvm//imports/kotlin/kotlinx/coroutines:core",
+    ],
+)
+
+kt_jvm_library(
     name = "filter_spec",
     srcs = ["FilterSpec.kt"],
     deps = [
--- a/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/EventSource.kt
+++ b/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/EventSource.kt
@@ -1,0 +1,40 @@
+/*
+ * Copyright 2025 The Cross-Media Measurement Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.wfanet.measurement.edpaggregator.resultsfulfiller
+
+import com.google.protobuf.Message
+import kotlin.coroutines.CoroutineContext
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.flow.Flow
+
+/**
+ * Interface for event sources that provide events for processing.
+ * 
+ * This interface abstracts the source of events, allowing for different implementations
+ * such as synthetic generation or reading from storage.
+ */
+interface EventSource {
+  /**
+   * Generates a flow of event batches.
+   * 
+   * @param dispatcher The coroutine context for parallel processing
+   * @return A flow of event batches, where each batch is a list of labeled events
+   */
+  suspend fun generateEventBatches(
+    dispatcher: CoroutineContext = Dispatchers.Default
+  ): Flow<List<LabeledEvent<out Message>>>
+}
--- a/src/test/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/BUILD.bazel
+++ b/src/test/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/BUILD.bazel
@@ -188,3 +188,20 @@ kt_jvm_test(
         "@wfa_common_jvm//imports/kotlin/kotlinx/coroutines:core",
     ],
 )
+
+kt_jvm_test(
+    name = "StorageEventSourceTest",
+    srcs = ["StorageEventSourceTest.kt"],
+    test_class = "org.wfanet.measurement.edpaggregator.resultsfulfiller.StorageEventSourceTest",
+    deps = [
+        "//src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller:storage_event_source",
+        "//src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller:labeled_event",
+        "//src/main/proto/wfa/measurement/api/v2alpha/event_templates/testing:test_event_kt_jvm_proto",
+        "@wfa_common_jvm//imports/java/com/google/common/truth",
+        "@wfa_common_jvm//imports/java/org/junit",
+        "@wfa_common_jvm//imports/kotlin/kotlinx/coroutines:core",
+        "@wfa_common_jvm//imports/kotlin/org/mockito/kotlin",
+        "@wfa_common_jvm//src/main/kotlin/org/wfanet/measurement/storage:selected_storage_client",
+        "@wfa_common_jvm//src/main/kotlin/org/wfanet/measurement/storage/testing",
+    ],
+)
--- a/src/test/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/StorageEventSourceTest.kt
+++ b/src/test/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/StorageEventSourceTest.kt
@@ -1,0 +1,268 @@
+/*
+ * Copyright 2025 The Cross-Media Measurement Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.wfanet.measurement.edpaggregator.resultsfulfiller
+
+import com.google.common.truth.Truth.assertThat
+import com.google.crypto.tink.KmsClient
+import com.google.protobuf.Any
+import com.google.protobuf.TypeRegistry
+import java.time.LocalDate
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.flowOf
+import kotlinx.coroutines.flow.toList
+import kotlinx.coroutines.runBlocking
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.junit.runners.JUnit4
+import org.mockito.kotlin.any
+import org.mockito.kotlin.doThrow
+import org.mockito.kotlin.eq
+import org.mockito.kotlin.mock
+import org.mockito.kotlin.whenever
+import org.wfanet.measurement.api.v2alpha.event_templates.testing.Person
+import org.wfanet.measurement.api.v2alpha.event_templates.testing.person
+import org.wfanet.measurement.api.v2alpha.event_templates.testing.testEvent
+import org.wfanet.measurement.common.pack
+import org.wfanet.measurement.edpaggregator.StorageConfig
+import java.time.Instant
+
+@RunWith(JUnit4::class)
+class StorageEventSourceTest {
+
+  private val mockEventReader: EventReader = mock()
+  private val mockKmsClient: KmsClient = mock()
+  private val mockStorageConfig: StorageConfig = StorageConfig()
+  private val mockTypeRegistry: TypeRegistry = TypeRegistry.getEmptyTypeRegistry()
+
+  @Test
+  fun `generateEventBatches returns batched events from storage`() {
+    runBlocking {
+    val dateRange = LocalDate.of(2025, 1, 1)..LocalDate.of(2025, 1, 2)
+    val eventGroupReferenceIds = listOf("event-group-1", "event-group-2")
+    val batchSize = 2
+
+    val testEvent1 = testEvent { person = person { gender = Person.Gender.MALE } }
+    val testEvent2 = testEvent { person = person { gender = Person.Gender.FEMALE } }
+    
+    val labeledEvent1 = LabeledEvent(
+      timestamp = Instant.now(),
+      vid = 1L,
+      message = testEvent1,
+      eventGroupReferenceId = "event-group-1"
+    )
+    val labeledEvent2 = LabeledEvent(
+      timestamp = Instant.now(),
+      vid = 2L,
+      message = testEvent2,
+      eventGroupReferenceId = "event-group-1"
+    )
+    val labeledEvent3 = LabeledEvent(
+      timestamp = Instant.now(),
+      vid = 3L,
+      message = testEvent1,
+      eventGroupReferenceId = "event-group-2"
+    )
+
+    whenever(mockEventReader.getLabeledEventsBatched(eq(LocalDate.of(2025, 1, 1)), eq("event-group-1"), eq(batchSize)))
+      .thenReturn(flowOf(listOf(labeledEvent1, labeledEvent2)))
+    
+    whenever(mockEventReader.getLabeledEventsBatched(eq(LocalDate.of(2025, 1, 1)), eq("event-group-2"), eq(batchSize)))
+      .thenReturn(flowOf(listOf(labeledEvent3)))
+    
+    whenever(mockEventReader.getLabeledEventsBatched(eq(LocalDate.of(2025, 1, 2)), any(), eq(batchSize)))
+      .thenReturn(flowOf())
+
+    val eventSource = StorageEventSource(
+      eventReader = mockEventReader,
+      dateRange = dateRange,
+      eventGroupReferenceIds = eventGroupReferenceIds,
+      batchSize = batchSize
+    )
+
+    val batches = eventSource.generateEventBatches().toList()
+
+    // Due to parallel processing, batches may be sent in different orders
+    // We should verify the total number of events across all batches
+    val totalEvents = batches.flatten()
+    assertThat(totalEvents).hasSize(3)
+    
+    val vids = totalEvents.map { it.vid }.sorted()
+    assertThat(vids).containsExactly(1L, 2L, 3L)
+    }
+  }
+
+  @Test
+  fun `generateEventBatches handles empty date range`() = runBlocking {
+    val dateRange = LocalDate.of(2025, 1, 2)..LocalDate.of(2025, 1, 1)
+    val eventGroupReferenceIds = listOf("event-group-1")
+
+    val eventSource = StorageEventSource(
+      eventReader = mockEventReader,
+      dateRange = dateRange,
+      eventGroupReferenceIds = eventGroupReferenceIds
+    )
+
+    val batches = eventSource.generateEventBatches().toList()
+
+    assertThat(batches).isEmpty()
+  }
+
+  @Test
+  fun `generateEventBatches continues processing when ImpressionReadException occurs`() = runBlocking {
+    val dateRange = LocalDate.of(2025, 1, 1)..LocalDate.of(2025, 1, 1)
+    val eventGroupReferenceIds = listOf("event-group-1", "event-group-2")
+    val batchSize = 10
+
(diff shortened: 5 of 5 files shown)


src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/StorageEventSource.kt
```diff
@@ -0,0 +1,362 @@
...
+     * 
+     * This function processes each event group's collection intervals and expands them
+     * into individual date ranges, deduplicating any overlapping periods.
+     * 
+     * @param eventGroupDetailsList List of event groups with their collection intervals
+     * @return List of unique date ranges for each event group
+     */
+    fun expandEventGroupDateRanges(
+      eventGroupDetailsList: List<EventGroupDetails>
+    ): List<EventGroupDateRange> {
```
georgi, line 170: no range needed.

src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/StorageEventSource.kt
```diff
@@ -0,0 +1,362 @@
...
+      return eventGroupDetailsList
+        .flatMap { eventGroupDetails ->
+          eventGroupDetails.collectionIntervalsList.flatMap { interval ->
+            val startDate = interval.startTime.toLocalDate()
```
georgi, line 174: date in the metadata store is based on what? assume utc for now

stevenwarejones (commented):
> @stevenwarejones reviewed 1 of 3 files at r2, 4 of 8 files at r3, 5 of 5 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2709)* status: all files reviewed, 7 unresolved discussions

___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/StorageEventSource.kt` line 196 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2709#-OYW_PBO6oH8V8S6-mNe:-OYW_PBO6oH8V8S6-mNf:brmlwai) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6ce07e6a3515f65a589a8c00aa06150453f6d469/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/StorageEventSource.kt#L196)):*
> ```Kotlin
>     private fun com.google.protobuf.Timestamp.toLocalDate(): LocalDate {
>       return Instant.ofEpochSecond(this.seconds, this.nanos.toLong())
>         .atZone(ZoneOffset.UTC)
> ```

add this as a constructor param by default set to UTC to give it visibility
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/StorageEventSource.kt` line 208 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2709#-OYW_V5r2WHAvWV8bERo:-OYW_V5r2WHAvWV8bERp:byzlfsl) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6ce07e6a3515f65a589a8c00aa06150453f6d469/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/StorageEventSource.kt#L208)):*
> ```Kotlin
>    * path resolution or event reader creation logic.
>    *
>    * @param kmsClient Client for Key Management Service operations (nullable for unencrypted storage)
> ```

note why and when its nullable
___
*[`src/test/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/MockEventReader.kt` line 17 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2709#-OYW_c148-V2ed7z5914:-OYW_c148-V2ed7z5915:b-ecjti6) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6ce07e6a3515f65a589a8c00aa06150453f6d469/src/test/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/MockEventReader.kt#L17)):*
> ```Kotlin
>  */
> 
> package org.wfanet.measurement.edpaggregator.resultsfulfiller
> ```

this should be in a testing class marked as a testonly package
___
*[`src/test/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/MockEventReader.kt` line 35 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2709#-OYW_jJ4EL7tjiMxQbpm:-OYW_jJ4EL7tjiMxQbpn:b-54vv4c) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6ce07e6a3515f65a589a8c00aa06150453f6d469/src/test/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/MockEventReader.kt#L35)):*
> ```Kotlin
>  * @param exception Optional exception to throw when readEvents() is called
>  */
> class MockEventReader(
> ```

nit: I prefer `FakeEventReader` since you aren't actually using any mock
___
*[`src/test/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/MockEventReaderTest.kt` line 17 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2709#-OYW_uUB4IQS2LmcpGvs:-OYW_uUB4IQS2LmcpGvt:be215wq) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6ce07e6a3515f65a589a8c00aa06150453f6d469/src/test/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/MockEventReaderTest.kt#L17)):*
> ```Kotlin
>  */
> 
> package org.wfanet.measurement.edpaggregator.resultsfulfiller
> ```

ditt on moving to `org.wfanet.measurement.edpaggregator.resultsfulfiller.testing`

stevenwarejones (commented):
> @stevenwarejones reviewed 7 of 7 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2709)* status: all files reviewed, 3 unresolved discussions (waiting on @georgi)

georgi (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2709)* status: all files reviewed, 1 unresolved discussion (waiting on @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/StorageEventSource.kt` line 208 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2709#-OYW_V5r2WHAvWV8bERo:-OY_oJi82w1H4-L3d4-b:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6ce07e6a3515f65a589a8c00aa06150453f6d469/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/StorageEventSource.kt#L208)):*

note why and when its nullable
</blockquote></details>

Done.

stevenwarejones (commented):
> @stevenwarejones reviewed 5 of 5 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2709)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @georgi)

renjiezh (approved):
> @renjiezh reviewed 1 of 8 files at r3, 1 of 5 files at r4, 4 of 7 files at r5, 5 of 5 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2709)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @georgi)

stevenwarejones (approved):
> @stevenwarejones reviewed 2 of 2 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2709)* status: all files reviewed, 3 unresolved discussions (waiting on @georgi)

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2709)* status: all files reviewed, 1 unresolved discussion (waiting on @georgi)

___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/StorageEventSource.kt` line 89 at r10](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2709#-OYpzIhS5iwel4BJ_lzt:-OYpzIhS5iwel4BJ_lzu:b-gj3z8) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/cc3eea157a494816e66fb522dbc032ee830186c5/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/StorageEventSource.kt#L89)):*
> ```Kotlin
> data class EventGroupDate(
>   val eventGroupReferenceId: String,
>   val date: LocalDate
> ```

sorry for the late comment - but looking further up your stack - wdyt of making this a timeinterval instead

stevenwarejones (changes requested):
> @stevenwarejones reviewed 8 of 14 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2709)* status: 12 of 17 files reviewed, 7 unresolved discussions (waiting on @renjiezh)

___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/StorageImpressionMetadataService.kt` line 76 at r11](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2709#-OZAiTHI9lKtvlprG7fV:-OZAiTHI9lKtvlprG7fW:bjrh98m) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/551f7899c7c254477488ff7f2567a52a9c3cd5f7/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/StorageImpressionMetadataService.kt#L76)):*
> ```Kotlin
> 
>       ImpressionDataSource(
>         interval = Interval.newBuilder()
> ```

nit: use the kotlin dsl
```
import com.google.type.interval
interval {
   startTime = ...
   ...
).build()
```

___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/StorageImpressionMetadataService.kt` line 95 at r11](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2709#-OZAijDyDMrxK7queswI:-OZAijDyDMrxK7queswJ:b-1cn0t5) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/551f7899c7c254477488ff7f2567a52a9c3cd5f7/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/StorageImpressionMetadataService.kt#L95)):*
> ```Kotlin
>    *   metadata contents.
>    */
>   private suspend fun readBlobDetails(metadataPath: String): BlobDetails {
> ```

i think this could be protected in the interface
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/StorageEventSource.kt` line 190 at r11](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2709#-OZASwfT7OwKWTix194E:-OZASwfT7OwKWTix194F:bz72kkh) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/551f7899c7c254477488ff7f2567a52a9c3cd5f7/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/StorageEventSource.kt#L190)):*
> ```Kotlin
>   /** Creates EventReader instances for all data sources provided by the metadata service. */
>   private suspend fun createEventReaders(): List<EventReaderInfo> {
>     val allSources = mutableListOf<Pair<String, ImpressionDataSource>>()
> ```

nit: don't use pairs - use a data class name meaningfully documented
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/StorageEventSource.kt` line 200 at r11](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2709#-OZATR7QDcb2fkY9DSzl:-OZATR7QDcb2fkY9DSzm:bmu93uf) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/551f7899c7c254477488ff7f2567a52a9c3cd5f7/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/StorageEventSource.kt#L200)):*
> ```Kotlin
>         sources.forEach { src -> allSources += details.eventGroupReferenceId to src }
>       }
>     }
> ```

nit: no need for a mutable map
```
val allSources: Map<String, ImpressionDataSource> =
  eventGroupDetailsList
    .flatMap { details ->
      details.collectionIntervalsList.flatMap { interval ->
        impressionMetadataService
          .listImpressionDataSources(details.eventGroupReferenceId, interval)
          .map { src -> details.eventGroupReferenceId to src }
      }
    }
    .toMap()
```

___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/StorageEventSource.kt` line 202 at r11](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2709#-OZAT2Yp9BM4YnSoeSrK:-OZAT2Yq2HJpaPJ8E7ow:b-zfevsi) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/551f7899c7c254477488ff7f2567a52a9c3cd5f7/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/StorageEventSource.kt#L202)):*
> ```Kotlin
>     }
> 
>     // Deduplicate sources by blob URI to avoid double-reading overlapping intervals.
> ```

i think this should actually throw - if there are overlapping sources, something is broken
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/ImpressionMetadataService.kt` line 48 at r11](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2709#-OZASS6TEO1OHjAkvHmm:-OZASS6TEO1OHjAkvHmn:b-7cl6f) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/551f7899c7c254477488ff7f2567a52a9c3cd5f7/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/ImpressionMetadataService.kt#L48)):*
> ```Kotlin
>    *
>    * @param eventGroupReferenceId event group reference identifier.
>    * @param period time period (closed-open) in UTC.
> ```

nit: protobuf interval is timezone independent
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/testing/PathEchoImpressionMetadataService.kt` line 40 at r11](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2709#-OZAj9rS3jiBYwiw84Dp:-OZAj9rS3jiBYwiw84Dq:b5pqjgc) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/551f7899c7c254477488ff7f2567a52a9c3cd5f7/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/testing/PathEchoImpressionMetadataService.kt#L40)):*
> ```Kotlin
>  * metadata path, with an empty [EncryptedDek]. No storage I/O is performed.
>  */
> class PathEchoImpressionMetadataService(
> ```

why do we need this? why not just use StorageImpressionMetadataService in the tests

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2709)* status: 11 of 17 files reviewed, 8 unresolved discussions (waiting on @georgi and @renjiezh)

___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/StorageImpressionMetadataService.kt` line 39 at r12](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2709#-OZC5W0F7DA3hQGgGymI:-OZC5W0F7DA3hQGgGymJ:bb6xpuw) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d799443d28032bbf72485dddd1f122cd437062ba/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/StorageImpressionMetadataService.kt#L39)):*
> ```Kotlin
>  * storage, and emits sources with UTC day intervals. Date expansion uses [zoneIdForDates].
>  */
> class StorageImpressionMetadataService(
> ```

i was thinking about this more.... the path resolver is essentially the same thing as your ImpressionMetadataService. I think they should be combined. 
It should probably be something more like
```
class StorageImpressionMetadataService(
  private val impressionDekStorageConfig: StorageConfig,
) : ImpressionMetadataService {

override fun listImpressionDataSources(
    eventGroupReferenceId: String,
    period: Interval,
    modelLine: String
  ): List<ImpressionDataSource> {
    // although I question the need for the EventPathResolver interface at all at this point - could just be a private function in this class
    val pathResolver = HardCodedPathEventPathResolver(....)
    val paths = pathResolver.resolvePaths(period, eventGroupReferenceId, modelLine)
    return paths.map { path ->
          val blobDetails = readBlobDetails(path.metadataPath)
          ImpressionDataSource(...
          
```

georgi (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2709)* status: 3 of 31 files reviewed, 6 unresolved discussions (waiting on @renjiezh and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/StorageEventSource.kt` line 202 at r11](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2709#-OZAT2Yp9BM4YnSoeSrK:-OZFvZlv19XAYGwl_NRW:b-cpguxn) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/551f7899c7c254477488ff7f2567a52a9c3cd5f7/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/StorageEventSource.kt#L202)):*

i think this should actually throw - if there are overlapping sources, something is broken
</blockquote></details>

Done. overlapping sources are expected in grouped requisitions, as discussed
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/StorageImpressionMetadataService.kt` line 95 at r11](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2709#-OZAijDyDMrxK7queswI:-OZFyiOgDEYIdOLxSoGq:b-myxx8t) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/551f7899c7c254477488ff7f2567a52a9c3cd5f7/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/StorageImpressionMetadataService.kt#L95)):*

i think this could be protected in the interface
</blockquote></details>

sorry, I don't understand what you mean
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/StorageImpressionMetadataService.kt` line 39 at r12](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2709#-OZC5W0F7DA3hQGgGymI:-OZFvhb3Cs1aJAJzgVlB:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d799443d28032bbf72485dddd1f122cd437062ba/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/StorageImpressionMetadataService.kt#L39)):*

i was thinking about this more.... the path resolver is essentially the same thing as your ImpressionMetadataService. I think they should be combined. 
It should probably be something more like
```
class StorageImpressionMetadataService(
  private val impressionDekStorageConfig: StorageConfig,
) : ImpressionMetadataService {

override fun listImpressionDataSources(
    eventGroupReferenceId: String,
    period: Interval,
    modelLine: String
  ): List<ImpressionDataSource> {
    // although I question the need for the EventPathResolver interface at all at this point - could just be a private function in this class
    val pathResolver = HardCodedPathEventPathResolver(....)
    val paths = pathResolver.resolvePaths(period, eventGroupReferenceId, modelLine)
    return paths.map { path ->
          val blobDetails = readBlobDetails(path.metadataPath)
          ImpressionDataSource(...
```

</blockquote></details>

Done.

stevenwarejones (approved):
> @stevenwarejones reviewed 1 of 14 files at r11, 28 of 28 files at r14, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2709)* status: all files reviewed, 1 unresolved discussion (waiting on @georgi)

___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/StorageImpressionMetadataService.kt` line 95 at r11](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2709#-OZAijDyDMrxK7queswI:-OZGUr6P39dNkwSo5W7e:bc1raak) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/551f7899c7c254477488ff7f2567a52a9c3cd5f7/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/StorageImpressionMetadataService.kt#L95)):*

sorry, I don't understand what you mean
</blockquote></details>

i think that this same function will be needed by the GrpcImpressionMetadataService. But we can make that change later.

stevenwarejones (changes requested):
> @stevenwarejones reviewed 24 of 24 files at r15, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2709)* status: all files reviewed, 5 unresolved discussions (waiting on @kungfucraig and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/ImpressionMetadataService.kt` line 43 at r14](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2709#-OZG_6jb-SS5IfG8VAzp:-OZK6WcdA2nscldVoy2i:bki7xaj) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/45ca238ffefb85b8c975c62dbb9813c49f5895a6/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/ImpressionMetadataService.kt#L43)):*

This API does not exist yet. We have this interface to communicate what the future API dependency should look like.
</blockquote></details>

i'd prefer to merge this and potentially delete it once we update to the ImpressionMetadataService. For now, it does not use the grpc impl.
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/StorageImpressionMetadataService.kt` line 43 at r14](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2709#-OZGaADt2iwaiZliqCww:-OZK6ktf5V-ihymU8xWl:b-hg6gnt) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/45ca238ffefb85b8c975c62dbb9813c49f5895a6/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/StorageImpressionMetadataService.kt#L43)):*

No this is here until we have the actual API.
</blockquote></details>

as stated above, this is the current workaround until the api work is finished. I'd prefer to get this merged in since it already works with the current cloud test. upgrading to the grpc version now will break the cloud test.
___
*[`src/test/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/BUILD.bazel` line 197 at r15](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2709#-OZK6yS-EvaPracUQU5U:-OZK6yS-EvaPracUQU5V:b-6vn7rj) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e93964f65a804754ab3e9831af75c75eeb3b5d40/src/test/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/BUILD.bazel#L197)):*
> ```Starlark
>         "//src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller:labeled_event",
>         "//src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller:storage_event_source",
>         "//src/main/proto/wfa/measurement/api/v2alpha/event_templates/testing:test_event_kt_jvm_proto",
> ```

this should not be a dependency here. it should be marked as testonly and this shouldn't compile.

georgi (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2709)* status: all files reviewed, 5 unresolved discussions (waiting on @kungfucraig, @SanjayVas, and @stevenwarejones)

___
*[`src/test/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/BUILD.bazel` line 197 at r15](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2709#-OZK6yS-EvaPracUQU5U:-OZK8KdtAIcXfZvldqL_:b-qhj0s0) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e93964f65a804754ab3e9831af75c75eeb3b5d40/src/test/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/BUILD.bazel#L197)):*

this should not be a dependency here. it should be marked as testonly and this shouldn't compile.
</blockquote></details>

I don't understand. This is a test target and the dependency is testonly. What's missing?

stevenwarejones (approved):
> @stevenwarejones reviewed all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2709)* status: all files reviewed, 4 unresolved discussions (waiting on @kungfucraig and @SanjayVas)

stevenwarejones (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2709)* status: all files reviewed, 8 unresolved discussions (waiting on @georgi, @kungfucraig, and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/ImpressionMetadataService.kt` line 43 at r14](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2709#-OZG_6jb-SS5IfG8VAzp:-OZLC_Cw-p1aOoMzh9Zj:b7892gs) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/45ca238ffefb85b8c975c62dbb9813c49f5895a6/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/ImpressionMetadataService.kt#L43)):*

The API definition was merged in #2776 and can already be coded against. That's no need to wait for the implementation as you can swap in another impl using the gRPC in-process transport. The main thing is using that API definition as the interface rather than wrapping in another one and passing the client stub where needed.

Writing client wrappers around gRPC client stubs is usually an anti-pattern. The exceptional cases are those like maintaining more complex client libraries (in multiple languages) like how Google Cloud does, or creating a different abstraction (e.g. how we use StorageClient to abstract many storage impls, only one of which is the gRPC ForwardedStorage). There's no need for a different abstraction here.

The options are:

1. Update this PR to code against that API definition, even if you currently use a different impl along with the gRPC in-process transport.
2. Add a TODO to delete the interface and make it a fast-follow. Note that this doesn't require waiting for the API to be implemented, and could just mean doing (1) in an immediate follow-up PR.
</blockquote></details>

i prefer 2. I agree we could write an in-process test. I'm more concerned that this will break the cloud test. To not break the current cloud test, we need to do (2) imo.
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/StorageEventSource.kt` line 192 at r16](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2709#-OZL3rRr6Ijs0FdKEAqJ:-OZLCr6S6wuoDeQLNDEy:b32v57) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e93964f65a804754ab3e9831af75c75eeb3b5d40/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/StorageEventSource.kt#L192)):*

Avoid `Pair`. Use specific data classes instead.
</blockquote></details>

+1
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/StorageImpressionMetadataService.kt` line 43 at r14](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2709#-OZGaADt2iwaiZliqCww:-OZLCv6Z7lrtNS9_g43e:b-7ob65v) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/45ca238ffefb85b8c975c62dbb9813c49f5895a6/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/StorageImpressionMetadataService.kt#L43)):*

See above comment re: not needing to wait for the service impl to switch to the API definition as the interface.

Note that I personally think this PR should have waited for the implementation of the metadata API unless this particular implementation will be useful for testing going forward. This seems like throwaway code to me.
</blockquote></details>

i agree its throwaway work at this point. but its already written. It should just be a fast-follow delete.

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2709)* status: all files reviewed, 8 unresolved discussions (waiting on @georgi and @kungfucraig)

___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/ImpressionMetadataService.kt` line 43 at r14](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2709#-OZG_6jb-SS5IfG8VAzp:-OZLHX9j1iqc7Lccf-xJ:bqtuedk) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/45ca238ffefb85b8c975c62dbb9813c49f5895a6/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/ImpressionMetadataService.kt#L43)):*
> I agree we could write an in-process test

In-process gRPC transport != in-process test. The change that needs to happen eventually is to drop this interface and use the API definition instead. This means passing in the client stub wherever you currently pass objects of this interface type.

In order to use the storage-backed implementation, you could just have it extend the corresponding CoroutineImplBase instead of implementing this interface. Then you can run an in-process server to serve that implementation so you can obtain a client stub. When the real implementation is ready, you can drop that code without having to change any code that calls the service.

kungfucraig (approved):
> @kungfucraig reviewed 1 of 5 files at r4, 1 of 14 files at r11, 5 of 28 files at r14, 24 of 24 files at r15, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2709)* status: all files reviewed, 9 unresolved discussions (waiting on @georgi and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/StorageEventSource.kt` line 97 at r16](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2709#-OZLZIv43mgiKdLt3o_-:-OZLZIv43mgiKdLt3o_0:b-3vhc4x) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e93964f65a804754ab3e9831af75c75eeb3b5d40/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/StorageEventSource.kt#L97)):*
> ```Kotlin
>   private val kmsClient: KmsClient?,
>   private val impressionsStorageConfig: StorageConfig,
>   private val descriptor: Descriptors.Descriptor,
> ```

nit: eventDescriptor

georgi (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2709)* status: 12 of 32 files reviewed, 4 unresolved discussions (waiting on @kungfucraig, @renjiezh, @SanjayVas, and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/StorageEventSource.kt` line 192 at r16](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2709#-OZL3rRr6Ijs0FdKEAqJ:-OZQi1iI5rjGlKn_G7Jd:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e93964f65a804754ab3e9831af75c75eeb3b5d40/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/StorageEventSource.kt#L192)):*

+1
</blockquote></details>

Done.

SanjayVas (changes requested):
> @SanjayVas reviewed 1 of 5 files at r4, 2 of 28 files at r14, 7 of 24 files at r15, 5 of 22 files at r18, 2 of 3 files at r19, 14 of 14 files at r20.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2709)* status: 30 of 32 files reviewed, 5 unresolved discussions (waiting on @georgi, @kungfucraig, and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/ImpressionMetadataService.kt` line 38 at r20](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2709#-OZQkeJZ8PXlpdAVUoGI:-OZQkeJZ8PXlpdAVUoGJ:bmt8tht) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1f2273803e1b5d59d3fb9b6fe2b7dad4b79a580a/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/ImpressionMetadataService.kt#L38)):*
> ```Kotlin
>  * Provides impression metadata for event groups over time periods.
>  *
>  * TODO: will be replaced by the gRPC client later
> ```

nit: Follow the TODO format as specified in our [style guide](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/main/docs/code-style.md#todos) to include a context reference. Ideally this would be a GitHub issue so we can keep track of this needing to be done.

Additionally, make this more specific as to the action to be taken.



_Suggestion:_
```Kotlin
TODO(world-federation-of-advertisers/cross-media-measurement#<issue number>): Delete this interface in favor of using gRPC client stubs.
```
___
*[`src/test/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/StorageEventSourceTest.kt` line 374 at r20](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2709#-OZQpNa00fddftu2WUFK:-OZQpNa00fddftu2WUFL:b-1r7xps) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1f2273803e1b5d59d3fb9b6fe2b7dad4b79a580a/src/test/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/StorageEventSourceTest.kt#L374)):*
> ```Kotlin
>       )
> 
>     assertFailsWith<Exception> {
> ```

Is there not a more specific exception type here? The assertions need to minimally disambiguate between this and other exceptions that might be thrown for other reasons.



_Code quote:_
```Kotlin
Exception
```
___
*[`src/test/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/StorageEventSourceTest.kt` line 538 at r20](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2709#-OZQpcyZAvdrYH-viTrz:-OZQpcyZAvdrYH-viTs-:bvoit9a) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1f2273803e1b5d59d3fb9b6fe2b7dad4b79a580a/src/test/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/StorageEventSourceTest.kt#L538)):*
> ```Kotlin
>       person = person {
>         ageGroup =
>           org.wfanet.measurement.api.v2alpha.event_templates.testing.Person.AgeGroup.YEARS_18_TO_34
> ```

nit: add import



_Code quote:_
```Kotlin
org.wfanet.measurement.api.v2alpha.event_templates.testing.Person
```
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/StorageEventSource.kt` line 43 at r20](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2709#-OZQlUARBXW7MTFfgMcS:-OZQlUARBXW7MTFfgMcT:b-enucv3) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1f2273803e1b5d59d3fb9b6fe2b7dad4b79a580a/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/StorageEventSource.kt#L43)):*
> ```Kotlin
>   private var totalEventsRead = 0L
>   private var totalBatchesSent = 0L
>   private var isClosed = false
> ```

nit: just `closed`.

The "is" prefix is generally for accessor methods which aren't needed in Kotlin. It's to satisfy the fact that function/method names should be verb phrases.



_Suggestion:_
```Kotlin
closed
```
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/StorageEventSource.kt` line 77 at r20](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2709#-OZQlpUP52HnIePyZeJ0:-OZQlpUP52HnIePyZeJ1:b-sxdkd9) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1f2273803e1b5d59d3fb9b6fe2b7dad4b79a580a/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/StorageEventSource.kt#L77)):*
> ```Kotlin
>   }
> 
>   fun logCompletion() {
> ```

nit: drop this method and just have the caller call close (preferrably from the `use` extension, but manually if neeed).

georgi (commented):
> @georgi reviewed 1 of 24 files at r15, 1 of 1 files at r24, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2709)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @georgi)

georgi (commented):
> @georgi reviewed 2 of 2 files at r25, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2709)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @georgi)

PR #2711, opened 2025-07-31
feat: add flags for res fulfiller using entrypoint

Marco-Premier:


SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2711)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @kungfucraig)

___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/BUILD.bazel` line 180 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2711#-OWWTeI98Qo_2M85og6S:-OWWVP5GCRXg_KdjbHU1:b-yhl8ge) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/43c3b4659c95817f655f4360e6ebc1ad063f3024/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/BUILD.bazel#L180)):*
Certainly this approach is problematic. If we can't avoid image customization for each environment, whatever customization needs to be applied must be achievable for operators without having to do their own Bazel builds.

> Even the API targets could be passed via the work item.

I'm not sure that we can (or should try to) pass everything via the WorkItem. I assume we need some amount of startup configuration.

PR #2713, opened 2025-08-01
feat: add support for multiple EDP cert collections

Marco-Premier:
- Add a new set of flags for each EDP that has to be part of the system.
- Pull certs from secrets for each EDP and store them locally


Marco-Premier (commented):
Done.

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2713#-:-OWuWW-XDDt02yNlQHax:b-896fix)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @kungfucraig)

___
*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2713#-OWkweK50QUiElyy-F8-:-OWuaTAN3DtfED9t8Ldk:b-896fix) (no related file):*

Similar to the previous PR. I'm skeptical that tests were not added/modified.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/ResultsFulfillerAppRunner.kt` line 343 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2713#-OWkvhup2N1VAwnFDdri:-OWu_kmn2dELYmZ60GWy:b-eiiokd) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d4ba81b9ebe5cf93e4993ebfdf2440b9d00d3480/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/ResultsFulfillerAppRunner.kt#L343)):*

Can you leave a comment that these paths must line up with what is in the ResultFulillerParams sent with the WorkItem.

Also, I think it's probably worth updating the ConsentParams to just have a single value, which corresponds to the "%s" in the paths here. It just seems strange that the config is reference what are essentially hard coded assumptions. Would you mind filing a bug for that?
</blockquote></details>

Agreed. I removed dynamic formatting. The full path must be passed now

Marco-Premier (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2713)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @kungfucraig)

PR #2564, opened 2025-07-11
feat!: Set plaintext metadata and data availability on EventGroups created by EDP Simulators

SanjayVas:
Issue: #2333
BREAKING-CHANGE: There are now two EDP Simulator images: simulator/legacy-metadata-edp and simulator/edp. The former sets legacy EventGroup metadata while the latter sets plaintext EventGroup metadata. Note that the former is not covered by Halo K8s tests.

Issue #2332, opened 2025-05-21
Remove MetadataSyntheticGeneratorEventQuery


SanjayVas:
The EventQuery mechanism is a holdover from when there were multiple EDP simulator implementations. It is unnecessarily complex now that there is only one implementation. A different mechanism can be used to share ground truth data between EDP simulators and the correctness test, such as a shared config file. The current mechanism relies on custom EventGroup metadata, which is deprecated.

This is a prerequisite to supporting the new EventGroup fields such as plaintext metadata and data availability in the EDP simulator.

SanjayVas:
Blocked by #2470

PR #2722, opened 2025-08-05
feat: update EdpSimulator to support TrusTee

roaminggypsy:


roaminggypsy:
@renjiezh I will add more unit tests later on, but let's focus on the main changes in simulator itself now.

At least two more things to add (either in this PR or another one?):
1. DataProviderCapabilities for Trustee
2. Trustee publicKey

renjiezh (changes requested):
1. Yes, please update DataProviderCapabilities for Trustee in the next PR. It requires public API change.
2. Trustee does not need the public key as of HMSS>

Besides, please add a happy path test for Trustee fulfillment.

> Reviewed 5 of 6 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2722#-:-OX5e5OxAnU8RhR11EGm:bcnjoij)* status: 5 of 6 files reviewed, 1 unresolved discussion

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt` line 189 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2722#-OX5ePyg2fUPiVLSlY_5:-OX5ePyh344Cqjg0UJIp:bhzdq4z) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/2ee1a94b844e90d6ad27ff7f86bc5e6d22b9ae92/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulator.kt#L189)):*
> ```Kotlin
>   private val logSketchDetails: Boolean = false,
>   private val health: SettableHealth = SettableHealth(),
>   private val kmsKekUri: String = "",
> ```

Use a `data class` to wrap these three. Maybe call it `TrusTeeEncryptionParams`. And allow them nullable.
`private val trusTeeEncryptionParams: TrusTeeEncryptionParams? =  null`
Also update the runner accordingly.

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2722)* status: 5 of 6 files reviewed, 2 unresolved discussions (waiting on @renjiezh and @roaminggypsy)

___
*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2722#-OX5ikbDBNSi8cK1_RzL:-OX5ikbEBKzlaLk1lAH8:bdwuhxg) (no related file):*
Base this on #2564, which needs to go in for v0.5.23

SanjayVas (changes requested):
> Reviewed 2 of 7 files at r4, 9 of 9 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2722)* status: all files reviewed, 7 unresolved discussions (waiting on @renjiezh and @roaminggypsy)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/AbstractEdpSimulatorRunner.kt` line 180 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2722#-OY35T52Bv_LI-blEOeM:-OY35T52Bv_LI-blEOeN:b6gc1lk) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/556582c4a6486e296acbccfe1e34bfef69017ed9/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/AbstractEdpSimulatorRunner.kt#L180)):*
> ```Kotlin
>         MinimumIntervalThrottler(Clock.systemUTC(), flags.throttlerMinimumInterval),
>         health,
>         AbstractEdpSimulator.TrusTeeParams(
> ```

You should only be building the params object if the flags are set. Otherwise pass in null.

You need to check if flags.trusTeeParams is initialized.
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/AbstractEdpSimulator.kt` line 172 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2722#-OY33CoZE0zA0HojlWA1:-OY33Co_3YZ33GDSA-8i:b-oynbon) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/556582c4a6486e296acbccfe1e34bfef69017ed9/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/AbstractEdpSimulator.kt#L172)):*
> ```Kotlin
>   private val logSketchDetails: Boolean,
>   private val health: SettableHealth,
>   private val trusTeeParams: TrusTeeParams? = null,
> ```

1. Optional params (those with defaults) go last
2. We generally require all the params at the AbstractEdpSimulator level and put the defaults, if any, on the concrete impls.
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/AbstractEdpSimulator.kt` line 173 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2722#-OY33qks1BLcaq5J21LO:-OY33qks1BLcaq5J21LP:b-jxuya2) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/556582c4a6486e296acbccfe1e34bfef69017ed9/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/AbstractEdpSimulator.kt#L173)):*
> ```Kotlin
>   private val health: SettableHealth,
>   private val trusTeeParams: TrusTeeParams? = null,
>   private val kmsClientFactory: KmsClientFactory<GCloudWifCredentials>? = null,
> ```

Looking at how these two params are used, maybe just take in a `TrusTeeRequisitionRequestBuilder.EncryptionParams` instead?
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/AbstractEdpSimulator.kt` line 1627 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2722#-OY34cNh8e-z7pkJ08L7:-OY34cNh8e-z7pkJ08L8:bqvt1po) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/556582c4a6486e296acbccfe1e34bfef69017ed9/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/AbstractEdpSimulator.kt#L1627)):*
> ```Kotlin
>     private const val DECLINED_EVENT_GROUP_ID = "declined"
> 
>     private const val OAUTH_TOKEN_TYPE_ID_TOKEN = "urn:ietf:params:oauth:token-type:id_token"
> ```

@renjiezh how much are these parameters expected to differ for different use cases? If these are always the same, maybe they should be hard-coded by the factory rather than constructor params of GCloudWifCredentials ?
___
*[`src/test/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulatorTest.kt` line 744 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2722#-OY36a7V5HqKc2tBIYKw:-OY36a7V5HqKc2tBIYKx:byrsk2c) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/556582c4a6486e296acbccfe1e34bfef69017ed9/src/test/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulatorTest.kt#L744)):*
> ```Kotlin
> 
>     val allEvents = matchingEvents + nonMatchingEvents
>     val eventQuery = InMemoryEventQuery(allEvents)
> ```

Avoid using InMemoryEventQuery for new test methods. Use a SyntheticGeneratorEventQuery instead.

We want to get rid of it eventually. See #2332

roaminggypsy (commented):
based on our discussion, we only verify the headers (since encryption part is tested in trustee request builder unit tests )

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2722#-:-OYDGCt94y52Dw9e3K_7:b-64t4wx)* status: 6 of 11 files reviewed, 7 unresolved discussions (waiting on @re, @renjiezh, and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/AbstractEdpSimulator.kt` line 172 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2722#-OY33CoZE0zA0HojlWA1:-OY8WL__030jHSmVLjj4:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/556582c4a6486e296acbccfe1e34bfef69017ed9/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/AbstractEdpSimulator.kt#L172)):*

1. Optional params (those with defaults) go last
2. We generally require all the params at the AbstractEdpSimulator level and put the defaults, if any, on the concrete impls.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/AbstractEdpSimulator.kt` line 173 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2722#-OY33qks1BLcaq5J21LO:-OY8d84kCEgwJe8sqcLd:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/556582c4a6486e296acbccfe1e34bfef69017ed9/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/AbstractEdpSimulator.kt#L173)):*

Looking at how these two params are used, maybe just take in a `TrusTeeRequisitionRequestBuilder.EncryptionParams` instead?
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/AbstractEdpSimulator.kt` line 1627 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2722#-OY34cNh8e-z7pkJ08L7:-OY8ctLQ6r3wPztMEMD4:b-o5ziw5) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/556582c4a6486e296acbccfe1e34bfef69017ed9/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/AbstractEdpSimulator.kt#L1627)):*

@renjiezh how much are these parameters expected to differ for different use cases? If these are always the same, maybe they should be hard-coded by the factory rather than constructor params of GCloudWifCredentials ?
</blockquote></details>

We will just use GcpKmsClient() (it will use default credentials)
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/AbstractEdpSimulatorRunner.kt` line 180 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2722#-OY35T52Bv_LI-blEOeM:-OY8eAz80QpL-wW5hN0Y:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/556582c4a6486e296acbccfe1e34bfef69017ed9/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/AbstractEdpSimulatorRunner.kt#L180)):*

You should only be building the params object if the flags are set. Otherwise pass in null.

You need to check if flags.trusTeeParams is initialized.
</blockquote></details>

Done.
___
*[`src/test/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulatorTest.kt` line 744 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2722#-OY36a7V5HqKc2tBIYKw:-OYDF9YbB1kFewRo7CT-:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/556582c4a6486e296acbccfe1e34bfef69017ed9/src/test/kotlin/org/wfanet/measurement/loadtest/dataprovider/EdpSimulatorTest.kt#L744)):*

Avoid using InMemoryEventQuery for new test methods. Use a SyntheticGeneratorEventQuery instead.

We want to get rid of it eventually. See #2332
</blockquote></details>

Done.

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2722)* status: all files reviewed, 3 unresolved discussions (waiting on @roaminggypsy and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/AbstractEdpSimulatorRunner.kt` line 173 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2722#-OYDb4FC7KHtERVUPR2G:-OYDgcKz8jfLEiYcQZiL:b-elrb2d) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/082120be58c2bff7a8904e34c9251d906f15c820/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/AbstractEdpSimulatorRunner.kt#L173)):*

@renjiezh How do we handle running locally, e.g. local cluster or in-process test? Off the top of my head, the options I can think of are

1. Use `FakeKmsClient` with a KEK URI that uses the `fake-kms` scheme.

    I think this is a no-go, as it requires the TrusTEE mill implementation to also use FakeKmsClient which is test code.
2. No encryption when running locally (leaving that path unexercised)
3. Implement our own simple (insecure) key management service with a client that implements `KmsClient`.

    This would be in place of a Google Cloud KMS emulator which doesn't exist. It's similar to what we do with ForwardedStorage service, which acts as our storage emulator when running locally.

Whatever we decide, I'm okay adding a TODO for now and coming back to it later. I just wanted to make sure there's a plan.
</blockquote></details>

Thanks for bringing this up.

I was thinking to use `FakeKmsClient` for all in-process tests.
But you are right it is not a good idea for local cluster run. I can create a simple ForwardedKms for local cluster runs. How about that?

roaminggypsy (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2722)* status: 10 of 11 files reviewed, 2 unresolved discussions (waiting on @renjiezh and @SanjayVas)

___
*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2722#-OX5ikbDBNSi8cK1_RzL:-OYIH0HZAkCW6a_0S9ek:b-896fix) (no related file):*

Base this on #2564, which needs to go in for v0.5.23
</blockquote></details>

Done.

stevenwarejones (changes requested):
> @stevenwarejones reviewed 2 of 7 files at r4, 4 of 9 files at r6, 4 of 5 files at r7, 1 of 1 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2722)* status: all files reviewed, 3 unresolved discussions (waiting on @renjiezh and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/AbstractEdpSimulatorRunner.kt` line 174 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2722#-OYWKhfk2iY_LsxDXW62:-OYWKhfk2iY_LsxDXW63:b-7j195u) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1f5e702be041fe792a6814865757630d25e41839/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/AbstractEdpSimulatorRunner.kt#L174)):*
> ```Kotlin
>         // TODO(lindreamdeyi): Swap the KMS client for local cluster runs.
>         TrusTeeFulfillRequisitionRequestBuilder.EncryptionParams(
>           GcpKmsClient(),
> ```

shouldn't this be using the `kmsClientFactory` to build this client? or is that what the TODO was referring to? the todo isn't so clear to me.

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2722)* status: all files reviewed, 3 unresolved discussions (waiting on @SanjayVas and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/AbstractEdpSimulatorRunner.kt` line 174 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2722#-OYWKhfk2iY_LsxDXW62:-OYX18b-Dt938YblQJ6V:b-cau0e6) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1f5e702be041fe792a6814865757630d25e41839/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/AbstractEdpSimulatorRunner.kt#L174)):*

shouldn't this be using the `kmsClientFactory` to build this client? or is that what the TODO was referring to? the todo isn't so clear to me.
</blockquote></details>

The EDP should own the KMS on cloud so they have the credential to access the KMS directly. EDP can create the KmsClient in the EDP runner, most likely with the [Tink KMS library](https://github.com/tink-crypto/tink/tree/master/java_src/src/main/java/com/google/crypto/tink/integration). Thus, the EDP simulator can get the client without a factory.
The todo is for the EDP runner. At beginning we plan to use FakeKmsClient for all testing runners, but the local cluster test should not use a in-process FakeKmsClient. Instead, a Kms emulator should be deployed. However, there is no official KMS emulator, thus we have to write a ForwardedKms which is similar to the GCS service in local cluster.

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2722)* status: all files reviewed, 3 unresolved discussions (waiting on @SanjayVas and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/AbstractEdpSimulatorRunner.kt` line 174 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2722#-OYWKhfk2iY_LsxDXW62:-OYYAjpC8EnWCiSWUkSj:b-8mo14f) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/9ca3b59d02d6052c6bef462425284f587a3911de/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/AbstractEdpSimulatorRunner.kt#L174)):*

so, the k8s test is going to fail until this todo is done?
</blockquote></details>

Nope. TrusTEE has two modes, with or without encryption. If no KmsClient is provided,
```
trusTeeEncryptionParams: TrusTeeFulfillRequisitionRequestBuilder.EncryptionParams? = null,
```
it just send plaintext data (for TechEdge use case). Before the todo addressed, we cannot enable the encryption mode in EDP.

stevenwarejones (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2722)* status: all files reviewed, 2 unresolved discussions (waiting on @renjiezh and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/AbstractEdpSimulatorRunner.kt` line 174 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2722#-OYWKhfk2iY_LsxDXW62:-OYYE3qf7h0jtZ7Fa1LT:bllhwmk) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/9ca3b59d02d6052c6bef462425284f587a3911de/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/AbstractEdpSimulatorRunner.kt#L174)):*

Nope. TrusTEE has two modes, with or without encryption. If no KmsClient is provided,
```
trusTeeEncryptionParams: TrusTeeFulfillRequisitionRequestBuilder.EncryptionParams? = null,
```
it just send plaintext data (for TechEdge use case). Before the todo addressed, we cannot enable the encryption mode in EDP.
</blockquote></details>

cc: @Marco-Premier

renjiezh (changes requested):
> @renjiezh reviewed 1 of 7 files at r4, 4 of 9 files at r6, 1 of 5 files at r7, 10 of 11 files at r10, 10 of 10 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2722)* status: all files reviewed, 2 unresolved discussions (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/AbstractEdpSimulator.kt` line 1120 at r11](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2722#-OYc2fwX5qu5PmfxFiTZ:-OYc2fwX5qu5PmfxFiT_:berkg9q) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5fc9212c54dabc0835a7417bff391663b154d197/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/AbstractEdpSimulator.kt#L1120)):*
> ```Kotlin
>     }
> 
>     chargeIndirectPrivacyBudget(
> ```

If no trusTeeEncryption, don't need to charge privacy budget.

renjiezh (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2722)* status: all files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

roaminggypsy (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2722)* status: all files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/AbstractEdpSimulator.kt` line 1120 at r11](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2722#-OYc2fwX5qu5PmfxFiTZ:-OYgP_co5EZX__oK7Kt3:b-5ezt1t) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5fc9212c54dabc0835a7417bff391663b154d197/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/AbstractEdpSimulator.kt#L1120)):*

If no trusTeeEncryption, don't need to charge privacy budget.
</blockquote></details>

(as discussed offline, let's keep charging budget in this case)

PR #2747, opened 2025-08-08
add ImpressionComputations

stevenwarejones:


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2747)

kungfucraig (dismissed):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2747)* status: 0 of 7 files reviewed, 1 unresolved discussion (waiting on @renjiezh)

___
*[`src/main/kotlin/org/wfanet/measurement/computation/ImpressionComputations.kt` line 22 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2747#-OXAy_jS-DrhW4R2UDeD:-OXAy_jTAwt3Y14i-15a:byvq3l7) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/97ebdb2d562b1e7e59795b10c2b35bc8279697d1/src/main/kotlin/org/wfanet/measurement/computation/ImpressionComputations.kt#L22)):*
> ```Kotlin
> 
> object ImpressionComputations {
>   private const val L0_SENSITIVITY = 1
> ```

We need to discuss how impressions get computed in some detail. 

This might be fine to start with, but we can't launch it, as it's not applying the proper amount of noise. Sensitivity is data dependent. We need to consider implementing dynamic clipping or perhaps have some options for this that each EDP can configure?

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2747)* status: 0 of 7 files reviewed, 1 unresolved discussion (waiting on @kungfucraig and @renjiezh)

___
*[`src/main/kotlin/org/wfanet/measurement/computation/ImpressionComputations.kt` line 22 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2747#-OXAy_jS-DrhW4R2UDeD:-OXB1f1M6SpHWYr3Rx3H:b-n4ar85) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/97ebdb2d562b1e7e59795b10c2b35bc8279697d1/src/main/kotlin/org/wfanet/measurement/computation/ImpressionComputations.kt#L22)):*

We need to discuss how impressions get computed in some detail. 

This might be fine to start with, but we can't launch it, as it's not applying the proper amount of noise. Sensitivity is data dependent. We need to consider implementing dynamic clipping or perhaps have some options for this that each EDP can configure?
</blockquote></details>

i'll make it configurable. would that work?

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2747)* status: 0 of 7 files reviewed, 1 unresolved discussion (waiting on @kungfucraig and @renjiezh)

___
*[`src/main/kotlin/org/wfanet/measurement/computation/ImpressionComputations.kt` line 22 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2747#-OXAy_jS-DrhW4R2UDeD:-OXB1o0z7TsVYCoThdCy:b-v92akr) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/97ebdb2d562b1e7e59795b10c2b35bc8279697d1/src/main/kotlin/org/wfanet/measurement/computation/ImpressionComputations.kt#L22)):*

i'll make it configurable. would that work?
</blockquote></details>

also, wouldn't the same things apply to reach?

stevenwarejones (commented):
> Reviewed 4 of 4 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2747)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @renjiezh)

PR #2767, opened 2025-08-11
add k-anonymity to ImpressionComputations

stevenwarejones:


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2767)

ple13 (changes requested):
> Reviewed 2 of 4 files at r1, 3 of 4 files at r3, 7 of 7 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2767)* status: all files reviewed, 1 unresolved discussion (waiting on @renjiezh)

___
*[`src/main/kotlin/org/wfanet/measurement/computation/ImpressionComputations.kt` line 67 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2767#-OXVn4LJ4AaiIDHW2958:-OXVn4LJ4AaiIDHW2959:b-o70d8s) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c83d888659598cc859508da394249c91db67a75f/src/main/kotlin/org/wfanet/measurement/computation/ImpressionComputations.kt#L67)):*
> ```Kotlin
>       } else {
>         if (
>           rawImpressionCount < kAnonymityParams.minImpressions ||
> ```

I wonder if we should use the noisy values for the comparisons (when DP Params are specified) instead of the true values. Decisions made based on true values may leak information about users' privacy.

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2767)* status: 6 of 8 files reviewed, 1 unresolved discussion (waiting on @ple13 and @renjiezh)

___
*[`src/main/kotlin/org/wfanet/measurement/computation/ImpressionComputations.kt` line 67 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2767#-OXVn4LJ4AaiIDHW2958:-OX_Z_1a0hdE3O59BCu0:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c83d888659598cc859508da394249c91db67a75f/src/main/kotlin/org/wfanet/measurement/computation/ImpressionComputations.kt#L67)):*

I wonder if we should use the noisy values for the comparisons (when DP Params are specified) instead of the true values. Decisions made based on true values may leak information about users' privacy.
</blockquote></details>

Done.

ple13 (changes requested):
> Reviewed 1 of 2 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2767)* status: 7 of 8 files reviewed, 2 unresolved discussions (waiting on @renjiezh)

___
*[`src/main/kotlin/org/wfanet/measurement/computation/ImpressionComputations.kt` line 85 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2767#-OX_cUkS2lSpInrQ0BHh:-OX_cUkS2lSpInrQ0BHi:b-ilx6c8) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e7349f0a6928c63141f9006c7450ed086639a984/src/main/kotlin/org/wfanet/measurement/computation/ImpressionComputations.kt#L85)):*
> ```Kotlin
>         if (
>           scaledImpressionCount < kAnonymityParams.minImpressions ||
>             (rawHistogram.sum() / vidSamplingIntervalWidth) < kAnonymityParams.minUsers
> ```

We need to add noise to rawHistogram.sum() too. Assume two databases D and D' = D U {some user} where histogram(D).sum() = minUsers - 1 and histogram(D').sum() = minUsers. Then, without DP noises, kAnonymityImpressionCount will be 0 for D, and scaledImpressionCount for D'. A distinguisher can easily tell whether D or D' is used in the computation.

noisedHistogramCount = noise.addNoise(raw histogram.sum(), l0Sen=1, lInfSen=1, eps, delta)

We will need to decide how much privacy budget we will charge for the computation of impressions when kAnonymityParams exists. I guess we could have a GitHub issue to track this?

PR #2773, opened 2025-08-12
chore: Rename GH vars used for test signed builds

laureanobrs:
Issue: https://github.com/world-federation-of-advertisers/cross-media-measurement/issues/2211


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2773)

PR #2781, opened 2025-08-14
add k-anonymity for reach and frequency computations

stevenwarejones:


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2781)

ple13 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2781)* status: all files reviewed, 1 unresolved discussion (waiting on @renjiezh and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/computation/ReachAndFrequencyComputations.kt` line 81 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2781#-OXgJlsv2bIEClzROKkz:-OXk3xHN2JQkRpAnKM6z:bqflmx3) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a168101bbedc01db1ea7e6f5b4a41ccc803db642/src/main/kotlin/org/wfanet/measurement/computation/ReachAndFrequencyComputations.kt#L81)):*

that's true but the histogram needs to computed based on an unknown max frequency.  We could use 128 but that means a lot of dp is applied potentially zeroing out cases that didn't need to be. I'd prefer to keep it this way. Otherwise, the dp noise >> min number of impressions for k-anonmity. If the minimum number of impressions is 1000 but the dp noise is 10k, then this is problematic from my pov.
</blockquote></details>

How about using a `maxFrequency` as in the case of the function computeFrequencyDistribution?

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2781)* status: all files reviewed, 1 unresolved discussion (waiting on @ple13 and @renjiezh)

___
*[`src/main/kotlin/org/wfanet/measurement/computation/ReachAndFrequencyComputations.kt` line 81 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2781#-OXgJlsv2bIEClzROKkz:-OXkQlEM02C0VGiBY5f6:bwe42qo) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a168101bbedc01db1ea7e6f5b4a41ccc803db642/src/main/kotlin/org/wfanet/measurement/computation/ReachAndFrequencyComputations.kt#L81)):*

How about using a `maxFrequency` as in the case of the function computeFrequencyDistribution?
</blockquote></details>

the frequency distributions already have a built in max frequency that is going to be revealed:
https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/0c869e2a26532f01415f680035190693b0b9d91e/src/main/proto/wfa/measurement/api/v2alpha/measurement_spec.proto#L94

Impressions already have it too as part of the API:
https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/0c869e2a26532f01415f680035190693b0b9d91e/src/main/proto/wfa/measurement/api/v2alpha/measurement_spec.proto#L107

The problem is that reach only measurements don't have this param of what the max frequency will be. In fact, the max frequency is unneeded for reach completely. Its only revealed as a side-channel by using k-anonymity.

I'd prefer to add to the current github issue and document the issue.

stevenwarejones (commented):
> Reviewed 9 of 9 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2781)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @renjiezh)

PR #2812, opened 2025-08-21
ci: Add cosign annotations needed for Confidential Space

laureanobrs:


SanjayVas (approved):
> @SanjayVas reviewed 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2812)* status: all files reviewed, 4 unresolved discussions (waiting on @Marco-Premier)

___
*[`.github/workflows/sign-images.yml` line 88 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2812#-OYDYw6RD4aq3IZhNeX2:-OYDYw6RD4aq3IZhNeX3:b-cflpa9) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3df8f77e07fe47408d8c717d36a4928548783e22/.github/workflows/sign-images.yml#L88)):*
> ```YAML
>           GCP_KMS_KEY_RESOURCE_NAME: ${{ inputs.kms-key-resource-name }}
>         run: |
>           set -euo pipefail
> ```

nit: Do you want to just skip this step and use the public key we already have checked into the repo? Basically extract what's in the README into files that you link from the README. You'd have another input parameter for the key path.

laureanobrs (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2812)* status: all files reviewed, 4 unresolved discussions (waiting on @Copilot, @Marco-Premier, and @SanjayVas)

___
*[`.github/workflows/sign-images.yml` line 88 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2812#-OYDYw6RD4aq3IZhNeX2:-OYDgfp30c_5iK2p5if8:b-f3nh3t) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3df8f77e07fe47408d8c717d36a4928548783e22/.github/workflows/sign-images.yml#L88)):*

nit: Do you want to just skip this step and use the public key we already have checked into the repo? Basically extract what's in the README into files that you link from the README. You'd have another input parameter for the key path.
</blockquote></details>

I think the correctness of the signing process is now dependent on a kind of duplicated logic and a single source of truth.

In the alternative version, with files in the repo, the 'duplicated' logic is eliminated, but two sources of truth appear (cosign does the signing with the provided KMS key, and we add the annotation with a public key that's in a file).

I also don't think the future implementation of key rotation sways the decision toward either option.

I don't have a strong opinion on this. If you think the alternative is better for avoiding this code, let's go with that one.

PR #2788, opened 2025-08-15
feat: Implement system API for RequisitionMetadata service 

renjiezh:
Issue: https://github.com/world-federation-of-advertisers/cross-media-measurement/issues/2826

PR #2849, opened 2025-08-28
feat: Add Spanner schema for RequisitionMetadata

renjiezh:
Issue: https://github.com/world-federation-of-advertisers/cross-media-measurement/issues/2827


SanjayVas (changes requested):
> @SanjayVas reviewed 2 of 25 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2849)* status: 2 of 25 files reviewed, 2 unresolved discussions

___
*[`src/main/resources/edpaggregator/spanner/create-requisition-metadata-schema.sql` line 21 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2849#-OYmehYfCKv5Rc-Md82s:-OYmehYfCKv5Rc-Md82t:b-kofqdc) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/35965e936a09a59a2b7cbad45a2bca506b7544cf/src/main/resources/edpaggregator/spanner/create-requisition-metadata-schema.sql#L21)):*
> ```TSQL
> -- Cloud Spanner database schema for the EDP Aggregator.
> --
> -- Table hierarchy:
> ```

nit: we don't actually need these top-level hierarchy comments as they go out-of-date the moment that you add another scheme change
___
*[`src/main/resources/edpaggregator/spanner/create-requisition-metadata-schema.sql` line 25 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2849#-OYmepQ22fSwE6zdMY9W:-OYmepQ22fSwE6zdMY9X:b-l5kb2b) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/35965e936a09a59a2b7cbad45a2bca506b7544cf/src/main/resources/edpaggregator/spanner/create-requisition-metadata-schema.sql#L25)):*
> ```TSQL
> --   └── RequisitionMetadata
> --       └── RequisitionMetadataActions
> ```

You'll want to add the proto descriptor and define a proto bundle for the enums here so you can use the protobuf enum types in the schema. See my comment in #2788.

Note that generally the sequencing for creating a new service is:

1. Define the public API
2. Define the internal API
3. Add the DB schema and implement the internal API
4. Implement the public API

Issue #2840, opened 2025-08-26
Serve PopulationSpecs directly from the CMMS API


SanjayVas:
The `Population` resource in the CMMS API currently has a blob URI for the `PopulationSpec`. This design assumed that only model provider(s) and the population provider needed access to a given `PopulationSpec`. The design predates protocols such as HMSS which require Event Data Providers (EDPs) to also have direct access to the `PopulationSpec`. The coordination required to ensure everyone has access is complex, and we are already seeing problems where EDPs have mismatched `PopulationSpec`s.

The design can be greatly simplified by just having the `PopulationSpec` be a field in `Population`. For now, every EDP and MP in the system should have access (see #2715). In the future, we may want to add support for restricting which MPs or EDPs can access a given `PopulationSpec`.

This would preempt #2832 .

PR #2860, opened 2025-09-03
Add validation for whether population fields are set in PopulationSpec.

SanjayVas:
Issue: #2840


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2860)

PR #2878, opened 2025-09-05
fix: change data_availability_interval field description

Marco-Premier:


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2878)

stevenwarejones (approved):

Issue #2863, opened 2025-09-03
Create population DataProvider "pdp1" on Halo-managed environments


SanjayVas:
- [x] Create resources
- [x] Set PDP_NAME and PDP_CERT_NAME GitHub environment variables
- Update AKID mapping configs
  - [x] dev
  - [x] head
  - [x] qa

SanjayVas:
Created on halo-cmm-dev:

`dataProviders/KnvDaDC7QKA`
`dataProviders/KnvDaDC7QKA/certificates/EY3XMTC7QPE`

SanjayVas:
## halo-cmm-head

DataProvider name: `dataProviders/TMB_hM88nzQ`
Certificate name: `dataProviders/TMB_hM88nzQ/certificates/VmksqzDDYTU`

PR #2904, opened 2025-09-10
feat: Update population requisition fulfiller in Halo-managed environments

SanjayVas:
Issue: #2907


SanjayVas:
Depends on #2863

PR #2913, opened 2025-09-12
feat: get model line availability

Marco-Premier:


SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2913)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @kungfucraig and @renjiezh)

___
*[`src/main/proto/wfa/measurement/edpaggregator/v1alpha/impression_metadata_service.proto` line 58 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2913#-O_JGRJ_5d2YUn4TsmXR:-O_JGRJ_5d2YUn4TsmXS:b-5rwy00) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/43d5af29db5eee0de7f94470f04a988db6510575/src/main/proto/wfa/measurement/edpaggregator/v1alpha/impression_metadata_service.proto#L58)):*
> ```proto
> 
>   // Computes the availability intervals for the given ModelLines.
>   rpc ComputeModelLinesAvailability(ComputeModelLinesAvailabilityRequest)
> ```

How are we guaranteeing that we have the impressions for all of the EDP's EventGroups for a given (ModelLine, Interval) pair before this is called?

This method's results can only be valid if that condition has been met.

kungfucraig (approved):
> @kungfucraig reviewed 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2913)* status: all files reviewed, 1 unresolved discussion (waiting on @renjiezh and @SanjayVas)

___
*[`src/main/proto/wfa/measurement/edpaggregator/v1alpha/impression_metadata_service.proto` line 58 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2913#-O_JGRJ_5d2YUn4TsmXR:-O_JWX2S2iyHDICiyBGq:b-llvkya) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/43d5af29db5eee0de7f94470f04a988db6510575/src/main/proto/wfa/measurement/edpaggregator/v1alpha/impression_metadata_service.proto#L58)):*

How are we guaranteeing that we have the impressions for all of the EDP's EventGroups for a given (ModelLine, Interval) pair before this is called?

This method's results can only be valid if that condition has been met.
</blockquote></details>

The caller has to be able to determine this, which is fine since the caller of this method is also the one the crawls the uploaded data and creates to impression metadata entries. 

Marco had previously implemented this function client side, but we really need it to be implemented over the database for it to be efficient.

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2913)* status: all files reviewed, 1 unresolved discussion (waiting on @kungfucraig and @renjiezh)

___
*[`src/main/proto/wfa/measurement/edpaggregator/v1alpha/impression_metadata_service.proto` line 58 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2913#-O_JGRJ_5d2YUn4TsmXR:-O_JekqtB8bQo7LSe0zF:b-hu7yk3) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/43d5af29db5eee0de7f94470f04a988db6510575/src/main/proto/wfa/measurement/edpaggregator/v1alpha/impression_metadata_service.proto#L58)):*

The caller has to be able to determine this, which is fine since the caller of this method is also the one the crawls the uploaded data and creates to impression metadata entries. 

Marco had previously implemented this function client side, but we really need it to be implemented over the database for it to be efficient.
</blockquote></details>

Let's at least document what this actually returns so there's no ambiguity. Specifically, that this returns the available ranges for each ModelLine for which there is ImpressionMetadata, but that it is not equivalent to the data availability for a given (DataProvider, ModelLine) pair unless there are no EventGroups missing.

I'm just worried about this somehow being called when not all ImpressionMetadata entries have been created and therefore the result gets misinterpreted.

kungfucraig (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2913)* status: all files reviewed, 1 unresolved discussion (waiting on @renjiezh and @SanjayVas)

___
*[`src/main/proto/wfa/measurement/edpaggregator/v1alpha/impression_metadata_service.proto` line 58 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2913#-O_JGRJ_5d2YUn4TsmXR:-O_Jl7Ja7XT3y1rLZnyi:b-1qf9p2) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/43d5af29db5eee0de7f94470f04a988db6510575/src/main/proto/wfa/measurement/edpaggregator/v1alpha/impression_metadata_service.proto#L58)):*

Let's at least document what this actually returns so there's no ambiguity. Specifically, that this returns the available ranges for each ModelLine for which there is ImpressionMetadata, but that it is not equivalent to the data availability for a given (DataProvider, ModelLine) pair unless there are no EventGroups missing.

I'm just worried about this somehow being called when not all ImpressionMetadata entries have been created and therefore the result gets misinterpreted.
</blockquote></details>

Yeah, documenting that caveat makes sense to me.

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2913)* status: all files reviewed, 1 unresolved discussion (waiting on @kungfucraig, @Marco-Premier, and @renjiezh)

___
*[`src/main/proto/wfa/measurement/edpaggregator/v1alpha/impression_metadata_service.proto` line 58 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2913#-O_JGRJ_5d2YUn4TsmXR:-O_Nm8Pf6RXwx57hwUux:bxmltn4) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/43d5af29db5eee0de7f94470f04a988db6510575/src/main/proto/wfa/measurement/edpaggregator/v1alpha/impression_metadata_service.proto#L58)):*

> this returns the available ranges for each ModelLine

I'm fully aware that's an issue @SanjayVas . I'm not sure what you mean by ranges, like a list of intervals, meaning changing the `ComputeModelLinesAvailabilityResponse.ModelLineAvailability` object?

If so which interval should I be using when updating the kingdom? As it accepts a single interval:
https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/main/src/main/proto/wfa/measurement/api/v2alpha/data_provider.proto#L109
</blockquote></details>

I just mean updating the documentation of this method to ensure that a caller doesn't misinterpret the response. In particular, the return value is not valid to use in `DataProvider.data_availability_intervals` if there are any `ImpressionMetadata` resources missing within the interval when this is called. The service has no way to detect a case such as the `ImpressionMetadata` for one `EventGroup` being missing, as it doesn't know the entire set of `EventGroup`s.

I'm not sure of the best way to phrase the above. A start would be to be explicit about what the method actually returns, which I believe is the largest interval for which all `ImpressionMetadata` are `ACTIVE` for each `ModelLine`. As noted above, this is distinctly different from the meaning of an interval in `DataProvider.data_availability_intervals`, and is only the same when there are no `EventGroup`s for which `ImpressionMetadata` is absent.

kungfucraig (approved):
> @kungfucraig reviewed 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2913)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @renjiezh)

Issue #2920, opened 2025-09-15
Allow ModelLine resources to be created with active_start_time in the past


SanjayVas:
We're already in a case where we need to create `ModelLine`s retroactively. This should be fine as EDPs now specify their data availability per-`ModelLine`, so there won't be any requirement to support the entire active range.

SanjayVas:
In order to create a ModelLine that was already active, we also need to allow a ModelRollout that already occurred.

PR #2921, opened 2025-09-16
feat: Allow retroactive creation of ModelLines and ModelRollouts

SanjayVas:
This implements world-federation-of-advertisers/cross-media-measurement-api#251

Issue: #2920

### DIFF
--- a/src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/SpannerModelLinesService.kt
+++ b/src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/SpannerModelLinesService.kt
@@ -28,7 +28,6 @@ import org.wfanet.measurement.common.grpc.failGrpc
 import org.wfanet.measurement.common.grpc.grpcRequire
 import org.wfanet.measurement.common.identity.ExternalId
 import org.wfanet.measurement.common.identity.IdGenerator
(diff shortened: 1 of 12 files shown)


SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2921)* status: 6 of 12 files reviewed, all discussions resolved (waiting on @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/SpannerModelLinesService.kt` line 68 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2921#-O_Mvsds98OkoRmBE43q:-O_Nu6sD3QN-vCaIABS7:b-nawji5) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3979e21dde9da8bf5f7c83c1ffa47269207422bb/src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/SpannerModelLinesService.kt#L68)):*

do you think there should be any sanity checking? ie you shouldn't be specifying a date a year in the past.
</blockquote></details>

I don't want to try to be too smart in service implementations, only validating what we can guarantee is incorrect. Additionally for synthetic data testing this may actually be over a year old. For example, our current synthetic event group specs for the EDP simulators generate events from 2021.

PR #2935, opened 2025-09-18
refactor: allow data watcher to process empty blobs if are called 'done'

Marco-Premier:


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2935)

Issue #2831, opened 2025-08-25
Support running Population Data Provider (PDP) requisition fulfiller in cloud


SanjayVas:
The PDP requisition fulfiller implementation is incomplete and it cannot yet be run in a cloud environment. Having a running PDP fulfiller is necessary for future MC API phases.

PR #2937, opened 2025-09-18
perf: Increase memory to Population requisition fulfiller to avoid OOMKilled

SanjayVas:
Issue: #2831


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2937)

PR #2951, opened 2025-09-22
perf: Reduce heap size of Requisition Fulfillment server

SanjayVas:
This avoids OOMKilled when running K8s tests on Halo-managed environments. Real environments will likely want to use more memory depending on fulfillment volume.


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2951)

SanjayVas:
Observed peaks after running K8s correctness test with these values:

Heap used: 102MiB
Heap used after last GC: 61MiB
Container memory used: 400MiB

PR #2965, opened 2025-09-24
fix: error that prevent mig re-creation when instance template changed

Marco-Premier:


Marco-Premier (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2965)* status: 1 of 5 files reviewed, 2 unresolved discussions (waiting on @SanjayVas)

___
*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2965#-O_wxBnJ5qbJ_ykQQaK4:-Oa-crOp6BQ_HbBF-DDg:b-896fix) (no related file):*

I don't see any `Issue` trailer.
</blockquote></details>

Done.
___
*[`src/main/terraform/gcloud/cmms/edp_aggregator.tf` line 94 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2965#-O_wvaeR83idXqnmbIGv:-Oa-_NXn4xVsq-IF-T-v:b-bbw6fc) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7f2a0a3993356114fc1a3042fab8ea83b0fd3383/src/main/terraform/gcloud/cmms/edp_aggregator.tf#L94)):*

We can't have pushing the same config with no other changes result in a MIG being recreated.

If we need different names, we need to document in the reusable module that changing variable X requires changing variable Y and do it ourselves manually when we make such a change in our `cmms` module. If we make a change that requires users of the reusable module to change their variables, that's a breaking change that needs to be called out.
</blockquote></details>

Agree. I added documentation, and a variable to set the version of the mig name so that it can keep the same "base name" and have a suffix that must change in case of network changes. 

I mark this PR as breaking change too.

I've updated that github variable already in all 3 environments.

SanjayVas (changes requested):
> @SanjayVas reviewed 3 of 4 files at r2.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2965)* status: 4 of 5 files reviewed, 2 unresolved discussions (waiting on @Marco-Premier)

___
*[`src/main/terraform/gcloud/cmms/edp_aggregator.tf` line 94 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2965#-O_wvaeR83idXqnmbIGv:-Oa1hBfI7X4DC9VsxJcS:bwpu101) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7f2a0a3993356114fc1a3042fab8ea83b0fd3383/src/main/terraform/gcloud/cmms/edp_aggregator.tf#L94)):*

Agree. I added documentation, and a variable to set the version of the mig name so that it can keep the same "base name" and have a suffix that must change in case of network changes. 

I mark this PR as breaking change too.

I've updated that github variable already in all 3 environments.
</blockquote></details>

If we need a version, it should be hard-coded in the cmms module rather than an environment variable as it needs to happen in the PR along with whatever code change requires a new name.

Issue #2643, opened 2025-07-25
Enable google storage private access


Marco-Premier:
Enable private access for lower latency when fetching impressions from gcp storage

PR #2978, opened 2025-09-26
feat!: add private network

Marco-Premier:
Add Private Network Infrastructure. Enable instance template updates by requiring a new MIG version.

BREAKING-CHANGE: The edp_aggregator Terraform module has had changes to its network configuration that require recreating the MIG with a new name. A new value must be specified for the managed_instance_group_name variable.

Issue: #2643 #2966


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2978)

Marco-Premier (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2978)* status: 3 of 8 files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

___
*[`src/main/terraform/gcloud/modules/edp-aggregator/main.tf` line 362 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2978#-Oa6-QmbDX7_BZqx9drV:-OaKry7uEumGKFwe1M7F:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/866acf98d6f1ebf9344978a03a6312ba17bbcd14/src/main/terraform/gcloud/modules/edp-aggregator/main.tf#L362)):*

If the issue is potential overlap here, we need to make sure that *anyone* using the module can avoid overlap. Therefore this should come from a variable, but we can of course have  "192.168.0.0/16" as the default.

The network should also probably be a variable in case someone doesn't want to attach this to the existing default network (e.g. they've created another one or have made other changes to the default one).
</blockquote></details>

Done.

SanjayVas (approved):
> @SanjayVas reviewed 3 of 8 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/2978)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Marco-Premier)

PR #3007, opened 2025-10-02
test: Add Basic Report creation to correctness test

tristanvuong2021:
The EmptyClusterCorrectnessTest and the SyntheticGeneratorCorrectnessTest both verify Basic Report Creation through the gateway

Issue: #3006

### DIFF
--- a/src/main/proto/wfa/measurement/integration/k8s/testing/correctness_test_config.proto
+++ b/src/main/proto/wfa/measurement/integration/k8s/testing/correctness_test_config.proto
@@ -67,11 +67,11 @@ message CorrectnessTestConfig {
   string access_public_api_target = 11;
 
   // jwt audience for Reporting Server.
-  string reporting_token_audience = 13;
+  string reporting_token_audience = 12;
 
   // Expected hostname (DNS-ID) in the access public API server's TLS
   // certificate.
   //
   // If not specified, standard TLS DNS-ID derivation will be used.
-  string access_public_api_cert_host = 12;
+  string access_public_api_cert_host = 13;
 }


SanjayVas (changes requested):
> @SanjayVas reviewed 10 of 17 files at r1, 7 of 7 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3007)* status: all files reviewed, 5 unresolved discussions (waiting on @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/reporting/ReportingUserSimulator.kt` line 322 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3007#-Ob-QL8G8nNLx2AdS5js:-Ob-QL8G8nNLx2AdS5jt:bjbguoe) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8df153f25d94377effb2676aa043574bdb187488/src/main/kotlin/org/wfanet/measurement/loadtest/reporting/ReportingUserSimulator.kt#L322)):*
> ```Kotlin
>     logger.info("Basic Report retrieval succeeded")
> 
>     val retrievedBasicReportBuilder = BasicReport.newBuilder()
> ```

Avoid assigning variable to builder in order to minimize mutability

e.g.

```kotlin
val retrievedBasicReport = BasicReport.newBuilder().also {
  JsonFormat.parser()
      .ignoringUnknownFields()
      .merge(retrievedBasicReportJson, it)
}.build()
```
___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/reporting/ReportingUserSimulator.kt` line 346 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3007#-Ob-RlPZ0DIclV0gmlzE:-Ob-RlPZ0DIclV0gmlzF:b731gjl) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8df153f25d94377effb2676aa043574bdb187488/src/main/kotlin/org/wfanet/measurement/loadtest/reporting/ReportingUserSimulator.kt#L346)):*
> ```Kotlin
> 
>   private suspend fun getEventGroup(): EventGroup {
>     var response: ListEventGroupsResponse = ListEventGroupsResponse.getDefaultInstance()
> ```

I think you can still use `listResources` here. See src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/EventGroupsService.kt for an example of how to use `listResources` with post-filtering.
___
*[`.github/workflows/run-k8s-tests.yml` line 56 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3007#-Ob-PVm0-EccrKL_rGog:-Ob-PVm0-EccrKL_rGoh:b976rbj) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8df153f25d94377effb2676aa043574bdb187488/.github/workflows/run-k8s-tests.yml#L56)):*
> ```YAML
>         GCLOUD_PROJECT: ${{ vars.GCLOUD_PROJECT }}
>         REPORTING_PUBLIC_API_TARGET: ${{ vars.REPORTING_PUBLIC_API_TARGET }}
>         REPORTING_GATEWAY_TARGET: ${{ vars.REPORTING_GATEWAY_TARGET }}
> ```

I don't think we have these environment variables. Do you need me to create them?
___
*[`MODULE.bazel` line 242 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3007#-Ob-PBAr8z602jcNoebY:-Ob-PBAr8z602jcNoebZ:bmtjgmr) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8df153f25d94377effb2676aa043574bdb187488/MODULE.bazel#L242)):*
> ```Starlark
>         "com.google.crypto.tink:tink-gcpkms:1.9.0",
>         "com.squareup.okhttp3:okhttp:4.12.0",
>         "com.squareup.okhttp3:okhttp-tls:4.12.0",
> ```

Drop the versions here since these are now coming from the BOM.



_Suggestion:_
```Starlark
        "com.squareup.okhttp3:okhttp",
        "com.squareup.okhttp3:okhttp-tls",
```
___
*[`src/test/kotlin/org/wfanet/measurement/integration/k8s/AbstractCorrectnessTest.kt` line 391 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3007#-Ob-Tht87mqIdROFGhz-:-Ob-Tht87mqIdROFGhz0:b-c9vaic) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8df153f25d94377effb2676aa043574bdb187488/src/test/kotlin/org/wfanet/measurement/integration/k8s/AbstractCorrectnessTest.kt#L391)):*
> ```Kotlin
>       }
>       if (mcUserRole.name.isEmpty()) {
>         mcUserRole = rolesStub.createRole(createRoleRequest)
> ```

This is going to throw an ALREADY_EXISTS error if the `mcUser` role exists with different permissions. I assume we don't want a bunch of different roles (e.g. mcUser-1, mcUser-2, etc.) so you probably want something like the following:

Check if the `mcUser` role exists

1. If it does
    a. If it has the right permissions, use that
    b. If not, delete the role and go to (2). This will have a side effect of removing all policy bindings, but that should be okay on test environments.
2. If it does not exist
   a. Create it

PR #3017, opened 2025-10-07
fix(report-processor): add threshold for large correction.

ple13:
Noise correction may modify the unnoised metrics a bit (e.g. 1e-9). However, this is enough to flag these updates as `large correction` as the check (abs(updated value - original value) > 7*standard_deviation) where standard_deviation = 0.

In order to avoid these false positives, we update the check for a large correction to (abs(updated value - original value) > max(7*standard_deviation, 1.0).


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3017)

PR #3066, opened 2025-10-14
feat: add telemetry module

georgi:


georgi (commented):
> @georgi reviewed 1 of 3 files at r2, 11 of 11 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3066)* status: all files reviewed, 6 unresolved discussions (waiting on @SanjayVas)

SanjayVas (changes requested):
> @SanjayVas reviewed 1 of 3 files at r2, 11 of 11 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3066)* status: all files reviewed, 8 unresolved discussions (waiting on @georgi)

___
*[`MODULE.bazel` line 252 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3066#-ObYvkq77mQc2AzgU9bV:-ObeFDps3o_ZVtbNPbn3:b-7vgtir) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/9826529f4cf5884a98c731ef6440c7e9cbf92b86/MODULE.bazel#L252)):*

that does not work: java.lang.RuntimeException: org.eclipse.aether.transfer.ArtifactNotFoundException: Could not find artifact io.opentelemetry.semconv:opentelemetry-semconv:pom: in
</blockquote></details>

Sorry, I had the wrong BOM. We're using `opentelemetry-bom` in common-jvm whereas this is in `opentelemetry-instrumentation-bom`
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/telemetry/TracedOperation.kt` line 28 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3066#-ObeKIcV84iz-9AM584v:-ObeKIcV84iz-9AM584w:b-v4ns8v) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7501640513f7725e04d5deab4dc1fb2d18f81929/src/main/kotlin/org/wfanet/measurement/edpaggregator/telemetry/TracedOperation.kt#L28)):*
> ```Kotlin
>  * Components create their own traced operations using these utilities.
>  */
> object TracedOperation {
> ```

nit: maybe `Tracing` or `TracedOperations`?



_Suggestion:_
```Kotlin
Tracing
```
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/telemetry/TracedOperation.kt` line 36 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3066#-ObeKPkR7yZyLaTx3JRD:-ObeKPkR7yZyLaTx3JRE:b-5d7yjw) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7501640513f7725e04d5deab4dc1fb2d18f81929/src/main/kotlin/org/wfanet/measurement/edpaggregator/telemetry/TracedOperation.kt#L36)):*
> ```Kotlin
>    * @param block Code to execute
>    */
>   suspend fun <T> trace(
> ```

Can you make this inline so you can drop the suspend modifiers? You'll probably need to make `withSpanContext` public or annotate it with `@PublishedApi`.

See https://kotlinlang.org/docs/inline-functions.html#restrictions-for-public-api-inline-functions
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/telemetry/EdpaTelemetry.kt` line 94 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3066#-ObeHZWH5bc4lk7Bwo6w:-ObeHZWH5bc4lk7Bwo6x:b-2z8dwq) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7501640513f7725e04d5deab4dc1fb2d18f81929/src/main/kotlin/org/wfanet/measurement/edpaggregator/telemetry/EdpaTelemetry.kt#L94)):*
> ```Kotlin
>    */
>   fun initialize() {
>     if (initialized) {
> ```

You can simplify this by just having a no-op method like `ensureInitialized` and then just rely on the `init` block. This way you don't need to keep track of the `initialized` state and can rely on initialization guarantees for thread safety. Otherwise you'd need to do some more work to make this thread safe.
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/telemetry/EdpaTelemetry.kt` line 130 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3066#-ObeI3PiBqnS8sxONjR7:-ObeI3PiBqnS8sxONjR8:b3sdu74) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7501640513f7725e04d5deab4dc1fb2d18f81929/src/main/kotlin/org/wfanet/measurement/edpaggregator/telemetry/EdpaTelemetry.kt#L130)):*
> ```Kotlin
> 
>     // Install JVM runtime metrics instrumentation
>     Classes.registerObservers(openTelemetry)
> ```

Do we still need to do this with autoconfigure?
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/telemetry/EdpaTelemetry.kt` line 182 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3066#-ObeITS8-0IMv16T5Iyf:-ObeITS8-0IMv16T5Iyg:b-cxinue) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7501640513f7725e04d5deab4dc1fb2d18f81929/src/main/kotlin/org/wfanet/measurement/edpaggregator/telemetry/EdpaTelemetry.kt#L182)):*
> ```Kotlin
>    */
>   fun flush(timeout: Duration = Duration.ofSeconds(5)): Boolean {
>     if (!initialized) return true
> ```

This can't be true when initialization happens from the `init` block.
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/telemetry/EdpaTelemetry.kt` line 201 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3066#-ObeItoy4DoxvFhilV88:-ObeItoy4DoxvFhilV89:blufpon) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7501640513f7725e04d5deab4dc1fb2d18f81929/src/main/kotlin/org/wfanet/measurement/edpaggregator/telemetry/EdpaTelemetry.kt#L201)):*
> ```Kotlin
>       }
>       logger.warning("Error during OpenTelemetry flush: ${e.message}")
> ```

Whenever logging due to an exception, use the logging method that takes in an exception rather than manually logging the exception message. That way you also get the stack trace and cause.
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/telemetry/EdpaTelemetry.kt` line 216 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3066#-ObeIK7SBUs-xrQapEq9:-ObeIK7T1GbB2CiXRzp9:bxydmdt) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7501640513f7725e04d5deab4dc1fb2d18f81929/src/main/kotlin/org/wfanet/measurement/edpaggregator/telemetry/EdpaTelemetry.kt#L216)):*
> ```Kotlin
> 
>     try {
>       meterProvider.shutdown().join(10, TimeUnit.SECONDS)
> ```

I believe autoconfigure takes care of this.
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/telemetry/EdpaTelemetry.kt` line 258 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3066#-ObeJ_WM1JYldVdPz3V-:-ObeJ_WN6YDAMzOolpJW:be5o1ka) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7501640513f7725e04d5deab4dc1fb2d18f81929/src/main/kotlin/org/wfanet/measurement/edpaggregator/telemetry/EdpaTelemetry.kt#L258)):*
> ```Kotlin
>  * @throws Exception Any exception thrown by the block (except it's properly recorded in the span)
>  */
> inline suspend fun <T> withSpan(
> ```

If you drop the suspend from both this and the lambda then it will inherently be suspending iff the lamdba is suspending. That's the benefit of using `inline` here.



_Code quote:_
```Kotlin
suspend
```

georgi (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3066)* status: 8 of 12 files reviewed, 7 unresolved discussions (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/telemetry/EdpaTelemetry.kt` line 94 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3066#-ObeHZWH5bc4lk7Bwo6w:-ObhbV2N876v0-YhmUWj:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7501640513f7725e04d5deab4dc1fb2d18f81929/src/main/kotlin/org/wfanet/measurement/edpaggregator/telemetry/EdpaTelemetry.kt#L94)):*

You can simplify this by just having a no-op method like `ensureInitialized` and then just rely on the `init` block. This way you don't need to keep track of the `initialized` state and can rely on initialization guarantees for thread safety. Otherwise you'd need to do some more work to make this thread safe.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/telemetry/EdpaTelemetry.kt` line 130 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3066#-ObeI3PiBqnS8sxONjR7:-Obhtfax1dA2ETkvRNhp:bubkrm5) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7501640513f7725e04d5deab4dc1fb2d18f81929/src/main/kotlin/org/wfanet/measurement/edpaggregator/telemetry/EdpaTelemetry.kt#L130)):*

Do we still need to do this with autoconfigure?
</blockquote></details>

As these are coming from the opentelemetry instrumentation library it would not be possible for the SDK to call it? (without depending on it) 
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/telemetry/EdpaTelemetry.kt` line 182 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3066#-ObeITS8-0IMv16T5Iyf:-ObhbWQR7h4hDJoF2Imz:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7501640513f7725e04d5deab4dc1fb2d18f81929/src/main/kotlin/org/wfanet/measurement/edpaggregator/telemetry/EdpaTelemetry.kt#L182)):*

This can't be true when initialization happens from the `init` block.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/telemetry/EdpaTelemetry.kt` line 201 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3066#-ObeItoy4DoxvFhilV88:-Obhgppu6W-B0Fgd0hnN:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7501640513f7725e04d5deab4dc1fb2d18f81929/src/main/kotlin/org/wfanet/measurement/edpaggregator/telemetry/EdpaTelemetry.kt#L201)):*

Whenever logging due to an exception, use the logging method that takes in an exception rather than manually logging the exception message. That way you also get the stack trace and cause.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/telemetry/EdpaTelemetry.kt` line 216 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3066#-ObeIK7SBUs-xrQapEq9:-ObhkPELABupIC6__a5g:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7501640513f7725e04d5deab4dc1fb2d18f81929/src/main/kotlin/org/wfanet/measurement/edpaggregator/telemetry/EdpaTelemetry.kt#L216)):*

I believe autoconfigure takes care of this.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/telemetry/EdpaTelemetry.kt` line 258 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3066#-ObeJ_WM1JYldVdPz3V-:-ObhkVa59BMUFfKPDbul:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7501640513f7725e04d5deab4dc1fb2d18f81929/src/main/kotlin/org/wfanet/measurement/edpaggregator/telemetry/EdpaTelemetry.kt#L258)):*

If you drop the suspend from both this and the lambda then it will inherently be suspending iff the lamdba is suspending. That's the benefit of using `inline` here.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/telemetry/TracedOperation.kt` line 36 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3066#-ObeKPkR7yZyLaTx3JRD:-ObhtNhl9fnKZutrK5R9:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7501640513f7725e04d5deab4dc1fb2d18f81929/src/main/kotlin/org/wfanet/measurement/edpaggregator/telemetry/TracedOperation.kt#L36)):*

Can you make this inline so you can drop the suspend modifiers? You'll probably need to make `withSpanContext` public or annotate it with `@PublishedApi`.

See https://kotlinlang.org/docs/inline-functions.html#restrictions-for-public-api-inline-functions
</blockquote></details>

Done.

georgi (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3066)* status: 8 of 12 files reviewed, 7 unresolved discussions (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/telemetry/EdpaTelemetry.kt` line 216 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3066#-ObeIK7SBUs-xrQapEq9:-Obi-Pk39zOn81Xi0rUv:b-9jiywt) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7501640513f7725e04d5deab4dc1fb2d18f81929/src/main/kotlin/org/wfanet/measurement/edpaggregator/telemetry/EdpaTelemetry.kt#L216)):*

Done.
</blockquote></details>

I removed the hook but left this function for testing.

georgi (commented):
> @georgi reviewed 4 of 4 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3066)* status: all files reviewed, 7 unresolved discussions (waiting on @SanjayVas)

georgi (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3066)* status: all files reviewed, 2 unresolved discussions (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/telemetry/EdpaTelemetry.kt` line 94 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3066#-ObeHZWH5bc4lk7Bwo6w:-Obi4r1cAo6nGGMx2nmx:b7wv9b3) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7501640513f7725e04d5deab4dc1fb2d18f81929/src/main/kotlin/org/wfanet/measurement/edpaggregator/telemetry/EdpaTelemetry.kt#L94)):*

I'm still seeing the public `initialize` method?

What I was suggesting was move everything from the `initialize` method into the `init` block and have a *public* `ensureInitialized` that does nothing aside from getting the calling code to reference `EdpaTelemetry` such that initialization is triggered.
</blockquote></details>

Isn't a no-op function optimized away by the compiler?

georgi (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3066)* status: 11 of 12 files reviewed, 2 unresolved discussions (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/telemetry/EdpaTelemetry.kt` line 94 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3066#-ObeHZWH5bc4lk7Bwo6w:-Obi5TQv2xNwwFyoP12A:b9w3r3m) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7501640513f7725e04d5deab4dc1fb2d18f81929/src/main/kotlin/org/wfanet/measurement/edpaggregator/telemetry/EdpaTelemetry.kt#L94)):*

Isn't a no-op function optimized away by the compiler?
</blockquote></details>

Either way, we would reference any of the methods here so the class should get loaded .
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/telemetry/EdpaTelemetry.kt` line 182 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3066#-ObeITS8-0IMv16T5Iyf:-Obi5Qgl72AK7MGYk1et:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7501640513f7725e04d5deab4dc1fb2d18f81929/src/main/kotlin/org/wfanet/measurement/edpaggregator/telemetry/EdpaTelemetry.kt#L182)):*

What I mean is that the `initialized` property is not needed as it can never be false within the context of any method on this object.
</blockquote></details>

Done.

SanjayVas (changes requested):
> @SanjayVas reviewed 1 of 1 files at r6.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3066)* status: all files reviewed, 1 unresolved discussion (waiting on @georgi)

___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/telemetry/EdpaTelemetry.kt` line 94 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3066#-ObeHZWH5bc4lk7Bwo6w:-Obi61Ky9Y94GIlK6cOW:b-3tsxu6) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7501640513f7725e04d5deab4dc1fb2d18f81929/src/main/kotlin/org/wfanet/measurement/edpaggregator/telemetry/EdpaTelemetry.kt#L94)):*

Either way, we would reference any of the methods here so the class should get loaded .
</blockquote></details>

The trigger to initialize the object shouldn't get optimized away.

Recall that initialization of this object needs to occur *before* anything can access the global OpenTelemetry instance, which itself may happen as early as the loading of other classes. The safest way I can think do this is to ensure something on this object is referenced early, such as within the `init` block of the main class for each Cloud Run function. This way even if the object reference is abstracted to another dependency of the Cloud Run function, it's still loaded earlier.

georgi (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3066)* status: 11 of 12 files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/telemetry/EdpaTelemetry.kt` line 94 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3066#-ObeHZWH5bc4lk7Bwo6w:-ObieuwD2nVoHDVKezh5:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7501640513f7725e04d5deab4dc1fb2d18f81929/src/main/kotlin/org/wfanet/measurement/edpaggregator/telemetry/EdpaTelemetry.kt#L94)):*

The trigger to initialize the object shouldn't get optimized away.

Recall that initialization of this object needs to occur *before* anything can access the global OpenTelemetry instance, which itself may happen as early as the loading of other classes. The safest way I can think do this is to ensure something on this object is referenced early, such as within the `init` block of the main class for each Cloud Run function. This way even if the object reference is abstracted to another dependency of the Cloud Run function, it's still loaded earlier.
</blockquote></details>

Done.

georgi (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3066)* status: 11 of 12 files reviewed, 4 unresolved discussions (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/telemetry/EdpaTelemetry.kt` line 92 at r9](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3066#-Oc6ucLx9j-R3J9qCk5S:-Oc7HO78ALaDuZmmdmq4:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/73a6fccbd66faf4398fc5cc2f0fb43599b9b3482/src/main/kotlin/org/wfanet/measurement/edpaggregator/telemetry/EdpaTelemetry.kt#L92)):*

If I'm understanding the problem correctly, you want one of the following three scenarios:

1. Real SDK instance.
2. No-op.
3. Test SDK instance.


For (2), I think specifying the `otel.sdk.disabled` system property give you this without needing anything else. My understanding is that you may be able to be more efficient in skipping some other steps, but that shouldn't be necessary. This means that (2) and (1) are effectively the same.

For (3), it depends on what you're trying to verify. If trying to test manual instrumentation of some class, then it should be sufficient to just replace the global instance with your own SDK instance.

The only case where you need to actually get this object to use a different SDK instance is in tests for this object itself, e.g. if you want to exercise the `flush` method. If you really need that, then having the instance being a simple var with synchronized access should do. Then you just have some method which sets that var for testing to replace the SDK instance.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/telemetry/EdpaTelemetry.kt` line 123 at r9](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3066#-Oc6tFCJ5_B8LWreJfJC:-Oc7HbeLAkgVIAvfK6aG:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/73a6fccbd66faf4398fc5cc2f0fb43599b9b3482/src/main/kotlin/org/wfanet/measurement/edpaggregator/telemetry/EdpaTelemetry.kt#L123)):*

All logging *must* be through java.util.logging (JUL). If we really wanted to use otel logging, we'd configure it as a backend for JUL.

I don't see why we even want to use otel logging. We just need to do whatever is necessary to ensure any logs from JUL get into Google Cloud Logging, which I assume there's some simple way to do. e.g. for GKE, stderr logs go into Google Cloud Logging by default.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/telemetry/EdpaTelemetry.kt` line 267 at r9](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3066#-Oc6zcvBBX7Ow5aIiIL6:-Oc7HeUZ3Sbd2-cChKgg:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/73a6fccbd66faf4398fc5cc2f0fb43599b9b3482/src/main/kotlin/org/wfanet/measurement/edpaggregator/telemetry/EdpaTelemetry.kt#L267)):*

Let's have all system properties configured outside of code. This means that you won't get any exporters unless configured in JVM options, which I think is what we want.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/telemetry/EdpaTelemetry.kt` line 282 at r9](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3066#-Oc6xkg-94e7M9VZjaLF:-Oc7HexX6IbjSni_W0BW:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/73a6fccbd66faf4398fc5cc2f0fb43599b9b3482/src/main/kotlin/org/wfanet/measurement/edpaggregator/telemetry/EdpaTelemetry.kt#L282)):*
> ```Kotlin
> 
>     val autoConfiguredSdk =
>       AutoConfiguredOpenTelemetrySdk.builder().setResultAsGlobal().build()
> ```

Done.

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3066)* status: 11 of 13 files reviewed, 3 unresolved discussions (waiting on @georgi)

___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/telemetry/Tracing.kt` line 49 at r12](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3066#-OcC_n53EeRfqQ1Jq4Pk:-OcHCqAE3xlOE4TGUZzo:b-mvxyp5) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/77428e390d5fee0261688bec8347899fcb97de93/src/main/kotlin/org/wfanet/measurement/edpaggregator/telemetry/Tracing.kt#L49)):*

This function suspends which means it doesn't handle non-suspendable code. That is to say it doesn't handle being called from code that doesn't suspend and therefore may not have a coroutine context.

It's possible you simply don't have that use case, which is fine. If you do however, you'll need a version of this function that doesn't suspend.
</blockquote></details>

The comment is still incorrect in that it only handles non-suspending code now.
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/telemetry/Tracing.kt` line 93 at r13](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3066#-OcHCzRA8x2TjENHcFgF:-OcHCzRA8x2TjENHcFgG:b91e9vn) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/39e3c98e309e76b29e6cd5ecff52a95b2294481c/src/main/kotlin/org/wfanet/measurement/edpaggregator/telemetry/Tracing.kt#L93)):*
> ```Kotlin
>    * @param block Code to execute
>    */
>   suspend fun <T> traceSuspending(
> ```

Keep this the same name as the other function but add a CoroutineScope receiver. This way when you're in a suspend function, this overload will automatically be selected over the other one and therefore prevent mistakes of calling the other one within suspend functions.



_Suggestion:_
```Kotlin
suspend fun <T> CoroutineScope.trace
```

georgi (commented):
> @georgi reviewed 4 of 4 files at r16, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3066)* status: all files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

SanjayVas (approved):
> @SanjayVas reviewed 2 of 4 files at r16, 2 of 2 files at r17, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3066)* status: all files reviewed, 1 unresolved discussion (waiting on @georgi)

___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/telemetry/Tracing.kt` line 207 at r17](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3066#-OcwQBoA0odTPJQ37zcB:-OcwQBoA0odTPJQ37zcC:b7ca2z3) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/92874ef3d1d54305329fa903f8225f0ae741e75f/src/main/kotlin/org/wfanet/measurement/edpaggregator/telemetry/Tracing.kt#L207)):*
> ```Kotlin
> 
>   override fun get(carrier: HttpRequest?, key: String): String? =
>     carrier
> ```

nit: optimize for the case where the case matches by doing an initial map lookup with the key first. Then you can follow up with the O(n) linear search.

SanjayVas (approved):
> @SanjayVas reviewed 2 of 2 files at r19, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3066)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @georgi)

SanjayVas (approved):
Approving assuming comments are resolved

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3066#-:-OdLPtHD-N9O0NTxTnjs:baepo6s)* status: all files reviewed, 4 unresolved discussions (waiting on @georgi)

PR #3070, opened 2025-10-15
feat: Transform ReportResults into BasicReport ResultGroups

tristanvuong2021:
Transform ReportResults into BasicReport ResultGroups. Does not include updating GetBasicReport to return the ResultGroups from the transformation.

Issue: #2557

### DIFF
--- a/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/gcloud/spanner/BasicReportNoiseCorrectedResultsTransformation.kt
+++ b/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/gcloud/spanner/BasicReportNoiseCorrectedResultsTransformation.kt
@@ -346,13 +346,15 @@ object BasicReportNoiseCorrectedResultsTransformation {
           ResultGroupKt.metricMetadata {
             this.reportingUnitSummary = reportingUnitSummary
 
-            if (reportingWindowResults.key.nonCumulativeWindowStartDate != null) {
+            val nonCumulativeWindowStartDate =
+              reportingWindowResults.key.nonCumulativeWindowStartDate
+            if (nonCumulativeWindowStartDate != null) {
               nonCumulativeMetricStartTime =
                 reportStart
                   .copy {
-                    day = reportingWindowResults.key.nonCumulativeWindowStartDate!!.day
-                    month = reportingWindowResults.key.nonCumulativeWindowStartDate!!.month
-                    year = reportingWindowResults.key.nonCumulativeWindowStartDate!!.year
+                    day = nonCumulativeWindowStartDate.day
+                    month = nonCumulativeWindowStartDate.month
+                    year = nonCumulativeWindowStartDate.year
                   }
                   .toTimestamp()
             }


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3070)

SanjayVas (changes requested):
> @SanjayVas reviewed 2 of 14 files at r1, 1 of 6 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3070)* status: 3 of 14 files reviewed, 4 unresolved discussions

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/gcloud/spanner/SpannerBasicReportsService.kt` line 618 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3070#-Oc8LAcC5PvaN5YJ-8KK:-Oc8LAcC5PvaN5YJ-8KL:b-fihg3t) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/59be683ec6bf1eb004e5ec4fb1d142c3010884c8/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/gcloud/spanner/SpannerBasicReportsService.kt#L618)):*
> ```Kotlin
>         )
>         .associate {
>           it.copy {
> ```

Passing around incomplete protobuf messages leads to trouble. This is why we tend to be careful when referencing resource messages in Create requests lest someone thinks it's a full message.

I'd recommend using a separate data class for the map key here for the parts you want to compare. Also be careful when comparing lists if order doesn't matter. The repeated field inside of Primitive _should_ be documented as UNORDERED_LIST, and I don't know if we've put any ORDER BYs in the query that would guarantee a stable order of event groups. DBs don't tend to guarantee result order without that, which means that it may be possible for the order of event groups in a primitive ReportingSet to change.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/gcloud/spanner/SpannerBasicReportsService.kt` line 627 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3070#-Oc8LfOLDDxOWtUeK1yW:-Oc8LfOLDDxOWtUeK1yX:b-tfbzlx) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/59be683ec6bf1eb004e5ec4fb1d142c3010884c8/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/gcloud/spanner/SpannerBasicReportsService.kt#L627)):*
> ```Kotlin
> 
>     val eventGroupKeysByDataProviderId:
>       Map<String, MutableList<ReportingSet.Primitive.EventGroupKey>> =
> ```

Does this need to be mutable?



_Code quote:_
```Kotlin
MutableList
```
___
*[`src/main/kotlin/org/wfanet/measurement/common/Timestamps.kt` line 27 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3070#-Oc8KTcw7nyeyOnirD6Y:-Oc8KTcw7nyeyOnirD6Z:b9x5a7q) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/59be683ec6bf1eb004e5ec4fb1d142c3010884c8/src/main/kotlin/org/wfanet/measurement/common/Timestamps.kt#L27)):*
> ```Kotlin
> 
> fun DateTime.toTimestamp(): Timestamp {
>   val source = this
> ```

nit: Maybe a check to make sure at least one of these is set?
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/gcloud/spanner/BasicReportNoiseCorrectedResultsTransformation.kt` line 38 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3070#-Oc8Kekz0ScwDRciMiP9:-Oc8Kekz0ScwDRciMiPA:b-hw4u3l) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/59be683ec6bf1eb004e5ec4fb1d142c3010884c8/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/gcloud/spanner/BasicReportNoiseCorrectedResultsTransformation.kt#L38)):*
> ```Kotlin
> import org.wfanet.measurement.internal.reporting.v2.reportingImpressionQualificationFilter
> import org.wfanet.measurement.internal.reporting.v2.resultGroup
> ```

Don't create top-level non-extension functions in new code. Wrap these in an object instead.

tristanvuong2021 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3070)* status: 1 of 16 files reviewed, 3 unresolved discussions (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/gcloud/spanner/BasicReportNoiseCorrectedResultsTransformation.kt` line 38 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3070#-Oc8Kekz0ScwDRciMiP9:-OcBju4i9vDW2oNy3ES_:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/fd6e48a9ceb8b074ed378d09f388661c4b1cc13f/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/gcloud/spanner/BasicReportNoiseCorrectedResultsTransformation.kt#L38)):*

Don't create top-level non-extension functions in new code. Wrap these in an object instead.

</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/gcloud/spanner/SpannerBasicReportsService.kt` line 618 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3070#-Oc8LAcC5PvaN5YJ-8KK:-OcBqrq5CHVOdcgZotnX:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/fd6e48a9ceb8b074ed378d09f388661c4b1cc13f/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/gcloud/spanner/SpannerBasicReportsService.kt#L618)):*

Passing around incomplete protobuf messages leads to trouble. This is why we tend to be careful when referencing resource messages in Create requests lest someone thinks it's a full message.

I'd recommend using a separate data class for the map key here for the parts you want to compare. Also be careful when comparing lists if order doesn't matter. The repeated field inside of Primitive _should_ be documented as UNORDERED_LIST, and I don't know if we've put any ORDER BYs in the query that would guarantee a stable order of event groups. DBs don't tend to guarantee result order without that, which means that it may be possible for the order of event groups in a primitive ReportingSet to change.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/gcloud/spanner/SpannerBasicReportsService.kt` line 627 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3070#-Oc8LfOLDDxOWtUeK1yW:-OcBjW5o3dnhOaim5ueA:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/fd6e48a9ceb8b074ed378d09f388661c4b1cc13f/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/gcloud/spanner/SpannerBasicReportsService.kt#L627)):*

Does this need to be mutable?
</blockquote></details>

Done.

SanjayVas (changes requested):
> @SanjayVas reviewed 3 of 14 files at r1, 1 of 6 files at r2, 3 of 8 files at r3.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3070)* status: 8 of 16 files reviewed, 5 unresolved discussions

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/gcloud/spanner/SpannerBasicReportsService.kt` line 636 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3070#-OcCgsNV4LJkIsuwM61d:-OcCgsNV4LJkIsuwM61e:bsnirjs) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c14ad9c6c2cf5327398fbb4ef0eb76378ba9a748/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/gcloud/spanner/SpannerBasicReportsService.kt#L636)):*
> ```Kotlin
>         for (dataProviderId in eventGroupKeysByDataProviderId.keys) {
>           val primitiveEventGroupKeys =
>             eventGroupKeysByDataProviderId.getValue(dataProviderId).toHashSet()
> ```

Avoid mutability and specifying impls where not necessary



_Suggestion:_
```Kotlin
.toSet()
```
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/gcloud/spanner/SpannerBasicReportsService.kt` line 675 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3070#-OcCfOgFDZRLKUDjmV12:-OcCfOgFDZRLKUDjmV13:b-ptamah) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c14ad9c6c2cf5327398fbb4ef0eb76378ba9a748/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/gcloud/spanner/SpannerBasicReportsService.kt#L675)):*
> ```Kotlin
>     )
>   }
> }
> ```

Use [sealed classes](https://kotlinlang.org/docs/sealed-classes.html) to represent oneofs

Also typically something like this would be defined inside of the parent class rather than its companion object. (`Foo.Bar` vs. `Foo.Companion.Bar`)



_Suggestion:_
```Kotlin
    private sealed class ReportingSetKey {
      data class Primitive(val eventGroupKeys: Set<ReportingSet.Primitive.EventGroupKey>) : ReportingSetKey()
      data class Composite(val setExpression: ReportingSet.SetExpression) : ReportingSetKey()
    }
  }
}
```
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/job/BasicReportsReportsJob.kt` line 689 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3070#-OcChxd35OARGwjdIhEp:-OcChxd35OARGwjdIhEq:bb9o83y) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c14ad9c6c2cf5327398fbb4ef0eb76378ba9a748/src/main/kotlin/org/wfanet/measurement/reporting/job/BasicReportsReportsJob.kt#L689)):*
> ```Kotlin
>     val externalReportingSetId: String,
>     val filter: String,
>     val groupingPredicates: HashSet<String>,
> ```

Does this need to be both

1. Mutable and
2. A HashSet (as opposed to not caring about the Set impl)

If not, use `Set`



_Suggestion:_
```Kotlin
Set
```

SanjayVas (approved):
> @SanjayVas reviewed 4 of 4 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3070)* status: all files reviewed, 1 unresolved discussion (waiting on @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/gcloud/spanner/BasicReportNoiseCorrectedResultsTransformation.kt` line 353 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3070#-OcgRz2oASHJu2ShHZle:-OcgRz2oASHJu2ShHZlf:bgu74dy) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/bed52e255ad9a69e829cbd651f4b202bfe686623/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/gcloud/spanner/BasicReportNoiseCorrectedResultsTransformation.kt#L353)):*
> ```Kotlin
>                 reportStart
>                   .copy {
>                     day = reportingWindowResults.key.nonCumulativeWindowStartDate!!.day
> ```

nit: extract variable for repeatedly used expression, which will also allow smart casts to work



_Code quote:_
```Kotlin
reportingWindowResults.key.nonCumulativeWindowStartDate!!
```

PR #3076, opened 2025-10-16
feat: Implement ImpressionMetadata batch delete

roaminggypsy:
Issue: #3089

### DIFF
--- a/src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/spanner/SpannerImpressionMetadataService.kt
+++ b/src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/spanner/SpannerImpressionMetadataService.kt
@@ -257,10 +257,9 @@ class SpannerImpressionMetadataService(
 
           if (result.impressionMetadata.state == State.IMPRESSION_METADATA_STATE_DELETED) {
             throw ImpressionMetadataNotFoundException(
-                request.dataProviderResourceId,
-                request.impressionMetadataResourceId,
-              )
-              .asStatusRuntimeException(Status.Code.NOT_FOUND)
+              request.dataProviderResourceId,
+              request.impressionMetadataResourceId,
+            )
           }
 
           txn.updateImpressionMetadataState(


roaminggypsy (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3076)* status: 1 of 7 files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

___
*[`src/main/proto/wfa/measurement/internal/edpaggregator/impression_metadata_service.proto` line 45 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3076#-Obj9EvX8lzvbRuD9KfL:-ObjL-vP26EKyqcnwUMF:byai5m9) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e39b1d96a108f7804467e23a4c6709d387c31a5a/src/main/proto/wfa/measurement/internal/edpaggregator/impression_metadata_service.proto#L45)):*

I plan to get rid of singular internal create too. I'm in the middle of updating that. The reason is that I want to centralize deletion logic in batch method, and internal singular deletion seems redundant.
</blockquote></details>

if the other approach is preferred, I can change this and create to use the other approach.

roaminggypsy (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3076)* status: 0 of 7 files reviewed, 2 unresolved discussions (waiting on @roaminggypsy and @SanjayVas)

___
*[`src/main/proto/wfa/measurement/internal/edpaggregator/impression_metadata_service.proto` line 45 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3076#-Obj9EvX8lzvbRuD9KfL:-Oc2fSYQ8FjdG9J7Dwwh:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e39b1d96a108f7804467e23a4c6709d387c31a5a/src/main/proto/wfa/measurement/internal/edpaggregator/impression_metadata_service.proto#L45)):*

Just be aware that if anything is already using these services that this may cause a temporary breakage during upgrade. For example, if the public API server gets updated before the internal API server or vice-versa. One benefit of conserving the internal API is that you don't risk that breakage.

This may be less of a risk for Delete than Create, as I believe the EDPA is already calling Create but not Delete. Regardless, we should be consistent with the approaches
</blockquote></details>

Done.
___
*[`src/main/proto/wfa/measurement/internal/edpaggregator/impression_metadata_service.proto` line 46 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3076#-ObjW28o9ViXSfBQxjJ8:-Oc2WGsjD7c0DcTUKFtp:b-8g5kyp) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6254fd6237fdbb1828d1434b196f32d7d3d22c13/src/main/proto/wfa/measurement/internal/edpaggregator/impression_metadata_service.proto#L46)):*

BatchDelete should have empty response. That should simplify your code quite a bit.
</blockquote></details>

I think we should keep the response according to [AIP#235](https://google.aip.dev/235): In the case where a response message is necessary because the resource is soft-deleted, the response **should** be specified with the following pattern: 



_Code snippet:_
```proto
message BatchDeleteBooksResponse {
  // Books deleted.
  repeated Book books = 1;
}

The response message must include one repeated field corresponding to the resources that were soft-deleted.
```

SanjayVas (commented):
Still reviewing, just getting these replies out.

> @SanjayVas reviewed 1 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3076#-:-Oc2rfQHDUd_c5vWKFb5:bvjpu3o)* status: 1 of 7 files reviewed, all discussions resolved (waiting on @roaminggypsy)

___
*[`src/main/proto/wfa/measurement/internal/edpaggregator/impression_metadata_service.proto` line 46 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3076#-ObjW28o9ViXSfBQxjJ8:-Oc2rZ-q5cxP8UndZHED:b6zrrkh) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6254fd6237fdbb1828d1434b196f32d7d3d22c13/src/main/proto/wfa/measurement/internal/edpaggregator/impression_metadata_service.proto#L46)):*

I think we should keep the response according to [AIP#235](https://google.aip.dev/235): In the case where a response message is necessary because the resource is soft-deleted, the response **should** be specified with the following pattern:
</blockquote></details>

Good catch.

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3076)* status: 1 of 7 files reviewed, 1 unresolved discussion (waiting on @roaminggypsy)

___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/service/v1alpha/ImpressionMetadataService.kt` line 176 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3076#-OcHQdUIBTb1EtjLPW5H:-OcHQdUIBTb1EtjLPW5I:b-rbt0p9) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ee24a97f744716480152700f96365fcc470aed8b/src/main/kotlin/org/wfanet/measurement/edpaggregator/service/v1alpha/ImpressionMetadataService.kt#L176)):*
> ```Kotlin
>     return try {
>       internalImpressionMetadataStub
>         .batchDeleteImpressionMetadata(
> ```

Continue calling the singular version on the internal service since it exists.

Marco-Premier (approved):
> @Marco-Premier reviewed 1 of 7 files at r1, 1 of 3 files at r2, 2 of 2 files at r3, 2 of 3 files at r4, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3076)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @stevenwarejones)

PR #3081, opened 2025-10-20
feat: resultsfulfiller telemetry

georgi:


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3081)

kungfucraig (approved):
> @kungfucraig reviewed 3 of 7 files at r1, 4 of 4 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3081)* status: all files reviewed, 2 unresolved discussions (waiting on @Marco-Premier)

___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/ResultsFulfiller.kt` line 136 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3081#-Ocrac_x6IHjLMQ1h-cg:-Ocrac_x6IHjLMQ1h-ch:bibca3o) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/180181455bece5a5e0ce911336bee4e43829b26b/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/ResultsFulfiller.kt#L136)):*
> ```Kotlin
>     val requisitionMetadataByName: Map<String, RequisitionMetadata> =
>       requisitionsMetadata.associateBy { it.cmmsRequisition }
>     val reportId: String? = requisitionsMetadata.firstOrNull()?.report
> ```

nit: If either of these end up null, it seems like there's a problem. Any reason not to force them to exist?

PR #3082, opened 2025-10-20
perf: Increase container memory for Reporting internal API server

SanjayVas:
Seeing restarts that look like OOMKilled even at only 35MiB of heap usage.


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3082)

PR #3087, opened 2025-10-20
feat: Implement impression metadata batch create

roaminggypsy:
Issue: #3088

### DIFF
--- a/src/test/kotlin/org/wfanet/measurement/edpaggregator/service/v1alpha/ImpressionMetadataServiceTest.kt
+++ b/src/test/kotlin/org/wfanet/measurement/edpaggregator/service/v1alpha/ImpressionMetadataServiceTest.kt
@@ -51,10 +51,10 @@ import org.wfanet.measurement.edpaggregator.v1alpha.GetImpressionMetadataRequest
 import org.wfanet.measurement.edpaggregator.v1alpha.ImpressionMetadata
 import org.wfanet.measurement.edpaggregator.v1alpha.ListImpressionMetadataRequest
 import org.wfanet.measurement.edpaggregator.v1alpha.ListImpressionMetadataRequestKt
-import org.wfanet.measurement.edpaggregator.v1alpha.batchDeleteImpressionMetadataRequest
-import org.wfanet.measurement.edpaggregator.v1alpha.batchDeleteImpressionMetadataResponse
 import org.wfanet.measurement.edpaggregator.v1alpha.batchCreateImpressionMetadataRequest
 import org.wfanet.measurement.edpaggregator.v1alpha.batchCreateImpressionMetadataResponse
+import org.wfanet.measurement.edpaggregator.v1alpha.batchDeleteImpressionMetadataRequest
+import org.wfanet.measurement.edpaggregator.v1alpha.batchDeleteImpressionMetadataResponse
 import org.wfanet.measurement.edpaggregator.v1alpha.computeModelLineBoundsRequest
 import org.wfanet.measurement.edpaggregator.v1alpha.computeModelLineBoundsResponse
 import org.wfanet.measurement.edpaggregator.v1alpha.copy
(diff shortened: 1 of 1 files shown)


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3087)

roaminggypsy (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3087)* status: 0 of 13 files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/service/v1alpha/ImpressionMetadataService.kt` line 117 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3087#-Oc2Sule4YrcdrNXlOp7:-Oc2Sule4YrcdrNXlOp8:bipn3q4) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6c18d9090e0cc25dcc177b99540e3f814d65f799/src/main/kotlin/org/wfanet/measurement/edpaggregator/service/v1alpha/ImpressionMetadataService.kt#L117)):*
> ```Kotlin
>     request: CreateImpressionMetadataRequest
>   ): ImpressionMetadata {
>     return batchCreateImpressionMetadata(
> ```

The original option was "have the public one call the internal BatchCreate", but "have the public one call the PUBLIC BatchCreate" is probably better because they share some common field checks. Otherwise it's going to be lots of similar repeated code.

roaminggypsy (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3087)* status: 5 of 14 files reviewed, 2 unresolved discussions (waiting on @roaminggypsy and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/service/Errors.kt` line 246 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3087#-Oc2TVam1lX_L4hExext:-Oc7zc498dCxL4Gacgw9:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6c18d9090e0cc25dcc177b99540e3f814d65f799/src/main/kotlin/org/wfanet/measurement/edpaggregator/service/Errors.kt#L246)):*

This doesn't work as the field paths are different for internal vs. public API requests. InvalidFieldValue basically can't be mapped this way. You may need to use a more specific error reason.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/service/v1alpha/ImpressionMetadataService.kt` line 117 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3087#-Oc2Sule4YrcdrNXlOp7:-Oc81nyOA3bprpaxtklZ:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6c18d9090e0cc25dcc177b99540e3f814d65f799/src/main/kotlin/org/wfanet/measurement/edpaggregator/service/v1alpha/ImpressionMetadataService.kt#L117)):*

> "have the public one call the PUBLIC BatchCreate"

This doesn't work due to needing error translation. e.g. the field paths on BatchCreate are different than the ones for singular Create.

I also don't think you resolved the comment on the original PR about the option of preserving the internal API to avoid errors during update. That is to say keeping both methods on the internal API and just sharing some code on the internal API.
</blockquote></details>

Done.

roaminggypsy (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3087)* status: 13 of 14 files reviewed, all discussions resolved (waiting on @Marco-Premier, @SanjayVas, and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/service/internal/testing/ImpressionMetadataServiceTest.kt` line 175 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3087#-Ocaea8V3ohSLhxhQM2C:-Oclt06y1bXEax6hs4_s:b5masc4) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/00a3a3c2851cd39c57ee5a386f4f5e0f24f49ca8/src/main/kotlin/org/wfanet/measurement/edpaggregator/service/internal/testing/ImpressionMetadataServiceTest.kt#L175)):*

I'm talking about code changes that appear to be unnecessary. The way this was written before was correct.
</blockquote></details>

It was using batchCreate, which was correct, but then I think that I want to make 
1) Use singular create for test cases that check field values that only affects one instance creation use singular create  (in this case, it's just singular missing request_id
2) Use batch create for test cases where some field values affect multiple creations, or need to check the duplication (e.g. same blob uri within the batch), consistency (e.g. same parent) against other values.

PR #3115, opened 2025-10-30
docs: add edpa deployment guide

Marco-Premier:
Issue: #2528

### DIFF
--- a/src/main/kotlin/org/wfanet/measurement/reporting/postprocessing/v2alpha/ReportConversion.kt
+++ b/src/main/kotlin/org/wfanet/measurement/reporting/postprocessing/v2alpha/ReportConversion.kt
@@ -362,9 +362,9 @@ fun Report.toReportSummaries(): List<ReportSummary> {
                                           resultAttribute.metricResult.reachAndFrequency.reach
                                             .value > 0
                                         ) {
-                                          bin.binResult.value.toLong()
+                                          bin.binResult.value
                                         } else {
-                                          0
+                                          0.0
                                         }
                                       // TODO(@ple13): Read the standard deviations directly from
                                       // the frequency buckets when the report populates the
--- a/src/main/kotlin/org/wfanet/measurement/reporting/postprocessing/v2alpha/ReportProcessor.kt
+++ b/src/main/kotlin/org/wfanet/measurement/reporting/postprocessing/v2alpha/ReportProcessor.kt
@@ -28,6 +28,7 @@ import java.time.format.DateTimeFormatter
 import java.util.Locale
 import java.util.logging.Logger
 import kotlin.io.path.name
+import kotlin.math.roundToLong
 import org.wfanet.measurement.common.getJarResourcePath
 import org.wfanet.measurement.common.toJson
 import org.wfanet.measurement.gcloud.gcs.GcsStorageClient
@@ -255,7 +256,7 @@ interface ReportProcessor {
      */
     private fun processReport(report: Report, verbose: Boolean = false): ReportProcessingOutput {
       val reportSummaries = report.toReportSummaries()
-      val correctedMeasurementsMap = mutableMapOf<String, Long>()
+      val correctedMeasurementsMap = mutableMapOf<String, Double>()
       val resultMap = mutableMapOf<String, ReportPostProcessorResult>()
       val foundIssues = mutableSetOf<ReportPostProcessorIssue>()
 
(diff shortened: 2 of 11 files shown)


SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3115)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @kungfucraig, @Marco-Premier, and @stevenwarejones)

___
*[`docs/edpaggregator/deployment-guide.md` line 3 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3115#-Od9kwoj3Uz-IV0aupq5:-OdF9iFvCxRxWOaEc--E:bqzeg1j) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c90df705b9c939b658dd81e1a577bfdb61f5af35/docs/edpaggregator/deployment-guide.md#L3)):*

Halo DCR is the EDP Aggregator.  I can rename it to "EDP Aggregator". I'm not sure what part you are suggesting to split out on its own deployment guide?
</blockquote></details>

Yes, stick to "EDP Aggregator" since that's the name we've been using rather than introducing a new name now. If we were going to rename it, the better name we had come up with was "EDP Proxy".

What I was suggesting splitting out was the secure computation system which EDPA uses to its own deployment guide. That's intended to be generic rather than being specific to EDPA, and we may eventually swap TrusTEE to make use of it.
___
*[`docs/edpaggregator/deployment-guide.md` line 790 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3115#-OdFBvQl85YGtwVmnQj3:-OdFBvQl85YGtwVmnQj4:b-w2iyep) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/06d20960c7b2343432648efaef26e80ca9e2daa1/docs/edpaggregator/deployment-guide.md#L790)):*
> ```Markdown
> ## Step 1 \- Deploy infra using Terraform
> 
> A Terraform example that sets up the EDP Aggregator infrastructure can be found [here](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/main/src/main/terraform/gcloud/cmms/edp_aggregator.tf).
> ```

Write up a separate example root module in src/main/terraform/gcloud/examples that doesn't rely on anything else from the cmms root module. As a root module, it should have everything necessary for usage from an otherwise empty project.

Since the intent is to deploy it in the same project/cluster as the Kingdom, you may want to start with the [Kingdom example](https://github.com/world-federation-of-advertisers/cross-media-measurement/tree/main/src/main/terraform/gcloud/examples/kingdom) as your base, e.g. symlink `examples/kingdom/main.tf` as `examples/edp-aggregator/kingdom.tf`

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3115)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @kungfucraig, @Marco-Premier, and @SanjayVas)

___
*[`docs/edpaggregator/deployment-guide.md` line 3 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3115#-Od9kwoj3Uz-IV0aupq5:-OdnzJ_bCfK-zYej1Zqj:b2zmesp) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c90df705b9c939b658dd81e1a577bfdb61f5af35/docs/edpaggregator/deployment-guide.md#L3)):*

Yes, stick to "EDP Aggregator" since that's the name we've been using rather than introducing a new name now. If we were going to rename it, the better name we had come up with was "EDP Proxy".

What I was suggesting splitting out was the secure computation system which EDPA uses to its own deployment guide. That's intended to be generic rather than being specific to EDPA, and we may eventually swap TrusTEE to make use of it.
</blockquote></details>

I added a new doc just focused secure computation @SanjayVas  - if that looks okay, I can delete the references in this doc and point to the other doc as a pre-req.

SanjayVas (commented):
> @SanjayVas reviewed 1 of 2 files at r3, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3115)* status: all files reviewed, 1 unresolved discussion (waiting on @kungfucraig and @Marco-Premier)

PR #3121, opened 2025-10-31
Fix: bug in the RequisitionFetcher where existing requisition metadata with state QUEUED or PROCESSING where treated as new

Marco-Premier:
# Summary

Treat QUEUED and PROCESSING requisition metadata states as existing instead of new

# Description

This change resolves a bug in RequisitionFetcher where existing requisition metadata in state QUEUED or PROCESSING was incorrectly treated as new. As a result, the system attempted to re-group and recreate metadata for requisitions already in progress, leading to a gRPC ALREADY_EXISTS exception.

# What’s Fixed

QUEUED and PROCESSING metadata entries are now recognized as existing, same as STORED.

# Tests

Added a test to ensure existing metadata in QUEUED and PROCESSING states prevents creation of a new grouped requisition

Issue: #3135 
Closes: #3135

### DIFF
--- a/src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/spanner/BUILD.bazel
+++ b/src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/spanner/BUILD.bazel
@@ -15,7 +15,6 @@ kt_jvm_library(
     name = "spanner_impression_metadata_service",
     srcs = ["SpannerImpressionMetadataService.kt"],
     deps = [
-        "//src/main/kotlin/org/wfanet/measurement/common:id_generator",
         "//src/main/kotlin/org/wfanet/measurement/common/api:etags",
         "//src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/spanner/db",
         "//src/main/kotlin/org/wfanet/measurement/edpaggregator/service/internal:errors",
--- a/src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/spanner/InternalApiServices.kt
+++ b/src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/spanner/InternalApiServices.kt
@@ -29,7 +29,7 @@ object InternalApiServices {
   ): Services {
     return Services(
       SpannerRequisitionMetadataService(databaseClient, coroutineContext, idGenerator),
-      SpannerImpressionMetadataService(databaseClient, coroutineContext, idGenerator),
+      SpannerImpressionMetadataService(databaseClient, coroutineContext),
     )
   }
 }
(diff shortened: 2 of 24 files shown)


kungfucraig (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3121)* status: all files reviewed, 2 unresolved discussions (waiting on @Marco-Premier)

___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/requisitionfetcher/RequisitionGrouperByReportId.kt` line 460 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3121#-Ocv2rY04QUI71GRdC4E:-OcvxhiKAgdMx8O_5nzT:be9xnct) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a417a5ba2fc1c453e39637009cec293c7df5dd93/src/main/kotlin/org/wfanet/measurement/edpaggregator/requisitionfetcher/RequisitionGrouperByReportId.kt#L460)):*

yeah - @Marco-Premier   can you file an issue to do that as a follow-up? It could alternatively be a not in state(s) filter
</blockquote></details>

Happy to do this as a follow-up. Please add a todo.

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3121)* status: all files reviewed, 3 unresolved discussions (waiting on @Marco-Premier)

___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/requisitionfetcher/RequisitionGrouperByReportId.kt` line 465 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3121#-Od4GHqr-VGfNbQOQrB1:-Od4GHqr-VGfNbQOQrB2:b-8edrgj) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a417a5ba2fc1c453e39637009cec293c7df5dd93/src/main/kotlin/org/wfanet/measurement/edpaggregator/requisitionfetcher/RequisitionGrouperByReportId.kt#L465)):*
> ```Kotlin
>           RequisitionMetadata.State.STORED,
>           RequisitionMetadata.State.QUEUED,
>           RequisitionMetadata.State.PROCESSING,
> ```

there is a race condition where you might need FULFILLED too. Why filter this list at all? I think you can just return the unfiltered list queried by report name.

kungfucraig (approved):
> @kungfucraig reviewed 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3121)* status: all files reviewed, 1 unresolved discussion (waiting on @stevenwarejones)

PR #3139, opened 2025-11-05
feat: set workers based on cpu count

georgi:


georgi (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3139)* status: all files reviewed, 1 unresolved discussion (waiting on @kungfucraig)

___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/ResultsFulfillerApp.kt` line 245 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3139#-OdL6Dhk9pUqMOIepzq5:-OdL6zEm7K_C50BW4nr9:bxtykpe) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/93e708dc44e1fd4c14199392a56a75f63ebbc9a1/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/ResultsFulfillerApp.kt#L245)):*

nit: If you think that each thread is going to consume 100% of a CPU this is fine, but oftentimes I've used a multiplier of the number of CPUs.
</blockquote></details>

even with 192 cores I could sature CPU at 100%, so this should be good in most cases.

Issue #3084, opened 2025-10-20
Implement BigQuery Authorized View Workflow


callcarlos:
This issue tracks the implementation of a new exchange workflow leveraging Google BigQuery Authorized Views for the join operation between the Model  Provider's join keys and the Data Provider's event data.

See the [design doc](https://docs.google.com/document/d/1bugx5RjymhxEzn7TnmM7Zz0aSfQcB7xnQSZU1L1H2sg/edit?usp=sharing).

PR #3164, opened 2025-11-10
feat: Implement Authorized view workflow Preprocess Events

callcarlos:
Implement the event preprocessing task for the BigQuery authorized view  workflow. This adds the preprocessor that encrypts events before they're written to BigQuery tables.

Key changes:
- Add PreprocessEventsTask to encrypt join keys and event data
- Implement AesGcmCryptoHelper for AES-GCM encryption/decryption
- Define EncryptedMatchedEvent proto format for encrypted events
- Wire up the preprocessor in ProductionExchangeTaskMapper

The preprocessor:
1. Reads UnprocessedEvent protos (plaintext join key + event data)
2. Encrypts join keys using EDP's commutative deterministic cipher
3. Encrypts event data using AES-GCM with encrypted join key
4. Outputs length-delimited EncryptedMatchedEvent protos

Issue: #3084

### DIFF
--- a/src/main/kotlin/org/wfanet/panelmatch/client/authorizedview/BUILD.bazel
+++ b/src/main/kotlin/org/wfanet/panelmatch/client/authorizedview/BUILD.bazel
@@ -1,0 +1,17 @@
+load("@wfa_rules_kotlin_jvm//kotlin:defs.bzl", "kt_jvm_library")
+
+package(default_visibility = [
+    "//src/main/kotlin/org/wfanet/panelmatch:__subpackages__",
+    "//src/test/kotlin/org/wfanet/panelmatch:__subpackages__",
+])
+
+kt_jvm_library(
+    name = "authorizedview",
+    srcs = glob(["*.kt"]),
+    deps = [
+        "//src/main/kotlin/org/wfanet/panelmatch/common",
+        "@wfa_common_jvm//imports/java/com/google/common:guava",
+        "@wfa_common_jvm//imports/java/com/google/crypto/tink",
+        "@wfa_common_jvm//imports/java/com/google/protobuf",
+    ],
+)
--- a/src/main/kotlin/org/wfanet/panelmatch/client/authorizedview/TinkAeadHelper.kt
+++ b/src/main/kotlin/org/wfanet/panelmatch/client/authorizedview/TinkAeadHelper.kt
@@ -1,0 +1,120 @@
+// Copyright 2025 The Cross-Media Measurement Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package org.wfanet.panelmatch.client.authorizedview
+
+import com.google.crypto.tink.Aead
+import com.google.crypto.tink.BinaryKeysetReader
+import com.google.crypto.tink.CleartextKeysetHandle
+import com.google.crypto.tink.aead.AeadConfig
+import com.google.crypto.tink.proto.AesGcmKey
+import com.google.crypto.tink.proto.KeyData
+import com.google.crypto.tink.proto.KeyStatusType
+import com.google.crypto.tink.proto.Keyset
+import com.google.crypto.tink.proto.OutputPrefixType
+import com.google.protobuf.ByteString
+import com.google.protobuf.kotlin.toByteString
+import java.security.SecureRandom
+import kotlin.math.absoluteValue
+
+/**
+ * Helper object for AES-GCM encryption operations using Tink AEAD primitives.
+ *
+ * This class creates Tink AEAD instances from raw key bytes for use in the authorized view
+ * workflow. The encrypted data format matches standard AES-GCM: IV (12 bytes) || ciphertext || auth
+ * tag (16 bytes).
+ *
+ * Uses [OutputPrefixType.RAW] to ensure compatibility with standard AES-GCM format without Tink's
+ * key ID prefix.
+ *
+ * TODO(world-federation-of-advertisers/common-jvm#365): Move this utility to common-jvm.
+ */
+object TinkAeadHelper {
+  private const val AES_KEY_SIZE = 32 // 256 bits for AES-256
+  private const val AES_GCM_KEY_TYPE_URL = "type.googleapis.com/google.crypto.tink.AesGcmKey"
+  private val NO_ASSOCIATED_DATA = byteArrayOf()
+  private val secureRandom = SecureRandom()
+
+  init {
+    AeadConfig.register()
+  }
+
+  /**
+   * Creates a Tink [Aead] primitive from raw key bytes.
+   *
+   * @param key The raw key bytes (must be at least 32 bytes; only first 32 bytes are used)
+   * @return An [Aead] instance for encryption/decryption
+   * @throws IllegalArgumentException if key has fewer than 32 bytes
+   */
+  fun createAead(key: ByteString): Aead {
+    require(key.size() >= AES_KEY_SIZE) {
+      "Key must be at least $AES_KEY_SIZE bytes, got ${key.size()}"
+    }
+    val truncatedKey = key.substring(0, AES_KEY_SIZE)
+
+    val aesGcmKey = AesGcmKey.newBuilder().setVersion(0).setKeyValue(truncatedKey).build()
+
+    val keyData =
+      KeyData.newBuilder()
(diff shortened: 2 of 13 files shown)


callcarlos (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3164)* status: 4 of 14 files reviewed, 3 unresolved discussions (waiting on @emadlivani and @robinsons)

___
*[`src/main/proto/wfa/panelmatch/client/authorizedview/encrypted_event.proto` line 48 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3164#-Odt2P-80HgwXfftVa6z:-Oe3MKUKEj5DUnaU0EsH:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7519554a9c2bd1a336e141322733e0dabd76bd6d/src/main/proto/wfa/panelmatch/client/authorizedview/encrypted_event.proto#L48)):*

Since this is a known to be a serialized message, can we just use that message as the type instead of `bytes`?
</blockquote></details>

Done.

callcarlos (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3164)* status: 4 of 14 files reviewed, 3 unresolved discussions (waiting on @emadlivani and @robinsons)

___
*[`src/main/kotlin/org/wfanet/panelmatch/client/authorizedview/AesGcmCryptoHelper.kt` line 34 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3164#-Odss17gEKbDJIhIvSHI:-OeN0qUW6VVp4Yu5jwhu:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7519554a9c2bd1a336e141322733e0dabd76bd6d/src/main/kotlin/org/wfanet/panelmatch/client/authorizedview/AesGcmCryptoHelper.kt#L34)):*

@emadlivani  Can you please review this? I think you're more familiar with the desired encryption scheme.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/PreprocessSourceEventsTask.kt` line 102 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3164#-Odt0rPv24_TwztlpRjL:-OeN0p2XEzagV3pICxpQ:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7519554a9c2bd1a336e141322733e0dabd76bd6d/src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/PreprocessSourceEventsTask.kt#L102)):*

`PreprocessEventsRequest` has several fields besides the unprocessed events. Here, you're only making use of the unprocessed events. My question is: Do we need those other fields? For example, `identifier_hash_pepper` and `hkdf_pepper` - are these not necessary for the new encryption scheme? Perhaps a question for @emadlivani .

Second, if we don't need those other fields, then I think it would be more appropriate for the `raw-events` input to use a different format. Length-delimited `UnprocessedEvent` would work.
</blockquote></details>

Done.

robinsons (commented):
> @robinsons reviewed 5 of 5 files at r3, 1 of 1 files at r4, 4 of 4 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3164)* status: all files reviewed, 1 unresolved discussion (waiting on @emadlivani)

___
*[`src/main/kotlin/org/wfanet/panelmatch/client/common/HashExtension.kt` line 14 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3164#-OeTEel10FHZdQWA5Uhe:-OeTEel10FHZdQWA5Uhf:bq1l9xn) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/bddefb998f66efb29eb88a45509777b47865857e/src/main/kotlin/org/wfanet/panelmatch/client/common/HashExtension.kt#L14)):*
> ```Kotlin
>  * @return A new [ByteString] representing the SHA-256 hash.
>  */
> fun ByteString.sha256(): ByteString {
> ```

FYI this functionality is already available here: https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/main/src/main/kotlin/org/wfanet/panelmatch/common/Fingerprinters.kt

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3164)* status: all files reviewed, 1 unresolved discussion (waiting on @callcarlos).

___
*[`src/main/kotlin/org/wfanet/panelmatch/client/authorizedview/AesGcmCryptoHelper.kt` line 35 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3164#-OgqtJ0DDki7tpMQop8M:-OgqtJ0DDki7tpMQop8N:bkn1lu3) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/73360bf0d486eff8558a009775a6a132931a6ee1/src/main/kotlin/org/wfanet/panelmatch/client/authorizedview/AesGcmCryptoHelper.kt#L35)):*
> ```Kotlin
>   private const val GCM_IV_SIZE = 12 // 96 bits (recommended for AES-GCM)
>   private const val GCM_TAG_SIZE = 128 // 128 bits auth tag
>   private const val CIPHER_ALGORITHM = "AES/GCM/NoPadding"
> ```

are other encryption schemes supported? see common-jvm - I prefer to use Streaming AEAD and tink

kungfucraig (commented):
> @kungfucraig made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3164)* status: all files reviewed, 1 unresolved discussion (waiting on @stevenwarejones).

___
*[`src/main/kotlin/org/wfanet/panelmatch/client/authorizedview/AesGcmCryptoHelper.kt` line 35 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3164#-OgqtJ0DDki7tpMQop8M:-OgsrzTqEhsmlDluqHPI:b-dr6wwa) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/73360bf0d486eff8558a009775a6a132931a6ee1/src/main/kotlin/org/wfanet/panelmatch/client/authorizedview/AesGcmCryptoHelper.kt#L35)):*
<details><summary><i>Previously, emadlivani (Emad Livani) wrote…</i></summary><blockquote>

It might be tough to make it work with other encryption methods given the nature of the double blind encryption we are using. There is dependency on our end as well. AES-GCM has been working well in other use cases like this for us.
</blockquote></details>

I agree with Emad we should not change the encryption scheme that is being used here.

We already have a working system we are fitting this into.

Would it make sense to file an issue and leave a TODO to move this to common JVM, or perhaps we could move it there now?

callcarlos (commented):
> @callcarlos made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3164)* status: 8 of 16 files reviewed, 1 unresolved discussion (waiting on @callcarlos, @robinsons, and @stevenwarejones).

___
*[`src/main/kotlin/org/wfanet/panelmatch/client/authorizedview/AesGcmCryptoHelper.kt` line 35 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3164#-OgqtJ0DDki7tpMQop8M:-OhfU79qCco3Kx9eZLDQ:b-5oasdd) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/73360bf0d486eff8558a009775a6a132931a6ee1/src/main/kotlin/org/wfanet/panelmatch/client/authorizedview/AesGcmCryptoHelper.kt#L35)):*

then imo it should be using https://github.com/world-federation-of-advertisers/common-jvm/blob/db70b45a1f2b801c08171655b260fca2a29ca20b/src/main/kotlin/org/wfanet/measurement/common/crypto/tink/AeadStorageClient.kt

and if not, at least use Tink
</blockquote></details>

AeadStorageClient wouldn't really fit in this implementation, but I can use tink inside this helper class will rename to TinkAeadHelper.kt, and try to keep the same interface.

callcarlos (commented):
> @callcarlos made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3164)* status: 14 of 16 files reviewed, 1 unresolved discussion (waiting on @stevenwarejones).

___
*[`src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/PreprocessEventsTask.kt` line 77 at r9](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3164#-OiFDPiI1JR9qVsVgIwV:-OiP-a213RVunt-XXnx_:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/4eb79f67a3749fe5a0ade477b52a4208c85aed93/src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/PreprocessEventsTask.kt#L77)):*

shouldn't this throw exceptions?
</blockquote></details>

Done.

PR #3166, opened 2025-11-11
fix: edpa - calculate correct frequency histogram when reach is zero

stevenwarejones:


stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3166)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @kungfucraig)

___
*[`src/test/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/ResultsFulfillerTest.kt` line 633 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3166#-Odo1A0k306F4C4nZjGI:-OdoBzk6C2BxAfXxrzXI:bnv6lel) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/49175ac3d305087fbfc358d0d52c64f4499fa1fb/src/test/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/ResultsFulfillerTest.kt#L633)):*

We want these to be zeroes don't we?

See: https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1503ca13cd3c2707a60ab7c397a45c26104a75e7/src/main/kotlin/org/wfanet/measurement/dataprovider/MeasurementResults.kt#L82
</blockquote></details>

yep - that's a bug in ReachAndFrequencyComputations. I added a test there and put in a fix. Note: this affected any 0 reach result. Also there may be other places in the code that assume non-zero reach, which is rare when DP is applied but potentially common for no dp use cases.

kungfucraig (approved):
> @kungfucraig reviewed 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3166)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @stevenwarejones)

stevenwarejones (commented):
> @stevenwarejones reviewed 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3166)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @stevenwarejones)

PR #3175, opened 2025-11-12
fix(edpa): Reduce fetcher frequency to every 15 minutes to avoid race conditions

georgi:
RELNOTES: Halo-managed environments have reduced the frequency of the Requisition Fetcher component EDPA to every 15 minutes to avoid problems from having more than one instance running concurrently. Operators may wish to do the same. This can be configured using the `requisition_fetcher_scheduler_config` variable of the `edp-aggregator` Terraform module.

### DIFF
--- a/src/main/k8s/dev/duchy_gke.cue
+++ b/src/main/k8s/dev/duchy_gke.cue
@@ -72,13 +72,13 @@ _duchy_cert_name: "duchies/\(_duchy_name)/certificates/\(_certificateId)"
 #FulfillmentResourceRequirements: ResourceRequirements=#ResourceRequirements & {
 	requests: {
 		cpu:    "200m"
-		memory: "512Mi"
+		memory: "1Gi"
 	}
 	limits: {
 		memory: ResourceRequirements.requests.memory
 	}
 }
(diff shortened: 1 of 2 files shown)


georgi (commented):
Do you mean task queue between kingdom and requisition fetcher, or adding another service in-between to schedule tasks?

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3175#-:-OduBqwr8_Pc99wHA73Q:b-iy3wd2)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @georgi)

SanjayVas (approved):
I think it'd be task queue between the scheduler and the fetcher. Basically scheduler schedules into task queue which then can have a max tasks of 1.

This is just a rough idea from searching for options. It can be designed later.

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3175#-:-Odu_I2FBnu4RBjDEsAY:b-ehzj7h)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @georgi)

Issue #3178, opened 2025-11-12
Move ModelLine creation to ResourceSetup


tristanvuong2021:
The EdpSimulators update their DataAvailiabilityIntervals for every ModelLine they have access to. Currently, when this is done, the ModelLine needed for testing hasn't been created yet. To resolve this, the ModelLine creation is moved to ResourceSetup.

PR #3179, opened 2025-11-12
refactor: ModelLine creation is moved to ResourceSetup

tristanvuong2021:
When the ModelLine is created after everything else is created, the EdpSimulator does not update its DataAvailabilityInterval to cover the created ModelLine. Therefore, the ModelLine needs to be created earlier.

Issue: #3178


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3179)

PR #3180, opened 2025-11-12
fix(report-processor): use double type for frequency histogram value.

### LINKED ISSUE #3159: Erroneous large corrections due to rounding error
I think we’ve spotted a bug in the noise corrector. It seems to be rounding frequency values down, which can result in large corrections for an EDP measurements.

I saw the below scenario (note that I’ve changed the values of EDP measurements):

The pre-noise corrected values for frequency 13, 14 and 15+ for EDP direct Reach&Frequency measurement are provided, respectively, as:

1200000.9999999998
1100000.9999999998
1.2000000999999996E7

The noise corrector appears to round these down to 1,200,000, 1,100,000, and 12,000,000.

It then sees a discrepancy between the sum of the frequency values and the reach value and performs a correction.

And because this EDP has a standard deviation of 0, even a correction of 1 is considered a large correction, because the threshold for large corrections is 7 * standard deviation.

The noise corrector log shows similar to the below where a noise correction of 1 has been applied.

```
"largeCorrections": [{
        "metricTitle": "XXXXX",
        "originalValue": "40000001",
        "correctedValue": "40000000"
      }]
    }
  },
  "issues": ["HAS_LARGE_CORRECTIONS"]
```

It also still seems to get it wrong, because it only corrects the reach value by 1 and does not adjust any of the frequency values, so there is still a discrepancy between the sum of the frequency histogram and the reach after correction.

ple13:
Currently, the type of frequency histogram bin value is double. However, report processor assumes that these values are integers. The assignment from report proto to report summary proto has an implicit conversion from double to Long, which is equivalent to a rounding down operation. Due to the rounding down issue, the reach becomes larger than the sum of the frequency counts, and the report processor tends to correct it. The problem appears when the EDP doesn't add noises to the measurements, thus, the change to the reach could be flagged as "large correction". To avoid this issue, the report processor now supports the double type for frequency histogram bin values.

Closes #3159

Issue: #3159

### DIFF
--- a/src/test/python/wfa/measurement/reporting/postprocessing/tools/post_process_origin_report_test.py
+++ b/src/test/python/wfa/measurement/reporting/postprocessing/tools/post_process_origin_report_test.py
@@ -448,10 +448,11 @@ class TestOriginReport(unittest.TestCase):
             'cumulative/mrc/edp1_edp2_' + str(i).zfill(2)],
           TOLERANCE
       )
-      self.assertLessEqual(
+      self._assertFuzzyLessEqual(
           corrected_measurements_map['cumulative/mrc/edp2_' + str(i).zfill(2)],
           corrected_measurements_map[
             'cumulative/mrc/edp1_edp2_' + str(i).zfill(2)],
+          TOLERANCE
       )
       self._assertFuzzyLessEqual(
           corrected_measurements_map[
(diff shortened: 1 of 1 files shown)


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3180)

kungfucraig (approved):
> @kungfucraig reviewed 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3180)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @ple13)

PR #3191, opened 2025-11-17
fix(report-processor): invalidate solution with negative results.

### LINKED ISSUE #3185: Noise corrector can return solutions with negative frequency values
**Describe the bug**
We have seen two instances of noise-corrected reports where the higher N+ (13+, 14+, 15+) direct reach values are calculated to be small (single digit) negative values. These were calculated from direct EDP reach and frequency measurements. We have validated that the pre-noise corrected higher frequency histogram buckets (13, 14, 15+) for this measurement are small (less than 10,000)/ zero. The EDP did not return any negative values.

**Steps to reproduce**
The two occurrences of this issue we've seen are where:
1. The direct EDP higher frequency values (5 to 15+ buckets) pre-noise correction are small (less than 10,000) / zero.
2. The measurement requires noise correction because the corresponding direct EDP impressionCount metric is slightly (~50,000) lower compared with the sum of the frequency bucket products.
3. Using a US market population spec

**Component(s) affected**
Noise corrector

**Version**
5.28-rc1

**Environment**
Aquila production

**Additional context**

ple13:
Sometimes the quadratic solver cannot find a solution that meets all constraints, it returns a partial solution. In that case, we need to verify that none of the results are less than a NEGATIVE_THRESHOLD of -1e-1. The threshold is acceptable as all the output values will be rounded to their nearest integer and small negative values will become 0.

Closes #3185

Issue: #3185

### DIFF
--- a/src/main/kotlin/org/wfanet/measurement/api/v2alpha/EventTemplates.kt
+++ b/src/main/kotlin/org/wfanet/measurement/api/v2alpha/EventTemplates.kt
@@ -67,4 +67,12 @@ object EventTemplates {
 
     return templateType.options.getExtension(EventAnnotationsProto.eventTemplate)
   }
+
+  fun getEventDescriptor(eventMessageDescriptor: Descriptors.Descriptor): EventDescriptor {
+    require(eventMessageDescriptor.options.hasExtension(EventAnnotationsProto.event)) {
+      "${eventMessageDescriptor.fullName} it not an event message type"
+    }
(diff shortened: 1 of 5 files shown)


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3191)

stevenwarejones (approved):

PR #3194, opened 2025-11-18
feat!: Deploy TrusTEE mill in MIG by Terraform

renjiezh:
BREAKING-CHANGE: The `duchy` Terraform module now includes a subnetwork for TrusTEE Mill instances. The CIDR range is controlled by the `trustee_mill_subnetwork_cidr_range` variable. This may need to be adjusted from the default if it conflicts with any existing ranges.
RELNOTES: The `duchy` Terraform module includes new variables for TrusTEE. These will be explained in future documentation.

### DIFF
--- a/src/main/kotlin/org/wfanet/measurement/duchy/deploy/gcloud/daemon/mill/trustee/GcsTrusTeeMillDaemon.kt
+++ b/src/main/kotlin/org/wfanet/measurement/duchy/deploy/gcloud/daemon/mill/trustee/GcsTrusTeeMillDaemon.kt
@@ -20,6 +20,7 @@ import com.google.cloud.secretmanager.v1.SecretVersionName
 import com.google.protobuf.ByteString
 import java.io.File
 import java.nio.file.Files
+import java.nio.file.StandardOpenOption
 import org.wfanet.measurement.common.commandLineMain
 import org.wfanet.measurement.duchy.deploy.common.daemon.mill.trustee.TrusTeeMillDaemon
 import org.wfanet.measurement.gcloud.gcs.GcsFromFlags
@@ -111,11 +112,17 @@ class GcsTrusTeeMillDaemon : TrusTeeMillDaemon() {
     val file = File(path)
     file.parentFile?.mkdirs()
     val buffer = bytes.asReadOnlyByteBuffer()
-    Files.newByteChannel(file.toPath()).use { channel ->
-      while (buffer.hasRemaining()) {
-        channel.write(buffer)
+    Files.newByteChannel(
+        file.toPath(),
+        StandardOpenOption.CREATE,
+        StandardOpenOption.WRITE,
+        StandardOpenOption.TRUNCATE_EXISTING,
+      )
+      .use { channel ->
+        while (buffer.hasRemaining()) {
+          channel.write(buffer)
+        }
       }
-    }
   }
 
   private fun accessSecret(projectId: String, secretId: String, version: String): ByteString {


renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3194)* status: 16 of 26 files reviewed, 6 unresolved discussions (waiting on @georgi, @Marco-Premier, and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/gcloud/daemon/mill/trustee/BUILD.bazel` line 42 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3194#-OgcPnhG-9BOOYVF43e8:-Ogeolng8LRJK1SIrFln:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/10b788194af60271298afb9e18391517d6c6e053/src/main/kotlin/org/wfanet/measurement/duchy/deploy/gcloud/daemon/mill/trustee/BUILD.bazel#L42)):*

This is for adding a glibc constraint. You only need it for images which use JNI.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/deploy/gcloud/daemon/mill/trustee/GcsTrusTeeMillDaemon.kt` line 121 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3194#-OgcSDYJ7eXPTfqBzPE0:-Ogf6bzM2Rxr8086cRln:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/10b788194af60271298afb9e18391517d6c6e053/src/main/kotlin/org/wfanet/measurement/duchy/deploy/gcloud/daemon/mill/trustee/GcsTrusTeeMillDaemon.kt#L121)):*

If this is an immutable ByteString, don't unnecessarily copy it to a mutable ByteArray. You can then handle the write using nio as

```kotlin
val buffer = bytes.asReadOnlyByteBuffer()
Files.newByteChannel(path).use { channel ->
  while (buffer.hasRemaining()) {
    channel.write(buffer)
  }
}
```

This should do a buffered write with as few copies as possible.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/duchy/mill/MillFlags.kt` line 56 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3194#-OgcXIN30sjq2cT_hXK6:-Ogf6eh79dCvk9xNdwY6:byg433c) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/10b788194af60271298afb9e18391517d6c6e053/src/main/kotlin/org/wfanet/measurement/duchy/mill/MillFlags.kt#L56)):*

I don't think you can use this with a flag that isn't required, as you end up with ambiguity or errors.

You instead likely want to specify `defaultValue = CommandLine.Option.NULL_VALUE` and have this be nullable so that you can distinguish when it's unspecified.
</blockquote></details>

Refactored.
___
*[`src/main/kotlin/org/wfanet/measurement/common/identity/DuchyInfo.kt` line 70 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3194#-OgcP574621O29dUdvlW:-Ogf6gZg8BbWffuSpBrW:bsrqz6o) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/10b788194af60271298afb9e18391517d6c6e053/src/main/kotlin/org/wfanet/measurement/common/identity/DuchyInfo.kt#L70)):*

Why this change? The only time it doesn't need to be set is when calling `setForTest`, in which case the flags aren't used.
</blockquote></details>

Done.
Reverted back to true and moved its usage for only llv2 and hmss.

SanjayVas (commented):
> @SanjayVas reviewed 7 of 9 files at r3, 2 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3194)* status: 24 of 26 files reviewed, 2 unresolved discussions (waiting on @georgi, @Marco-Premier, and @renjiezh)

PR #1535, opened 2024-03-25
Add CreateUiReport to the CLI

bdomen-ggl:
Update the Reporting CLI tool to create a UI readable report. It is moderately hardcoded to limit the amount of configuration required by the user. There are a number of steps to complete this process:
1. Create 3 primitive Reporting Sets (with a number of event groups).
2. Create 1 union Reporting Set - the union of all 3 primitives.
3. Create 3 unique Reporting Sets (difference between the union and the union of the other two. eg. unique A = ABC - BC)
4. Create 1 metric calculation spec for the primitives.
5. Create 1 metric calculation spec for the composites.
6. Create a report mapping the reporting sets and metric calculation specs and include the other report parameters (like date range and frequency).

All the reporting sets and the report will be tagged appropriately so the Reporting BFF can process the report correctly.

PR #1495, opened 2024-02-21
Add UI features from Demo work

bdomen-ggl:
Several small changes from added features during the demo:
* Added the "demo" banner at the top of the site and made it sticky.
* Updated some card labels in the Report view and adjusted the styles so the columns aren't fixed. Didn't specifically get feedback from UX on the behavior but this was requested from users.
* Also made the summary cards more flexible for different content.
* Added an overflow to the card in case the user is on a small screen. This was noticeable in the summary table.
* Removed "on target reach" from the charts.
* Updated the title.
* Added Google Analytics.

Bug Fixes:
* Added a margin at the bottom of the site so the feedback button isn't stuck over things at the bottom like charts.
* Fixed fonts. There was an extra "./" in the font src that would work locally but screw up webpack with routes on the server. Also removed the extra font styles and just left sans serif.
* Added more cases to the number to magnitude converter including handling the 0 case to make the logic simpler.
* Updated BUILD files for some ChartGroups. These charts were removed from the current UI so they didn't fail the build.

PR #1447, opened 2024-01-29
Update the Reporting UI Report Models, Views, and ViewModels to match the BFF model

bdomen-ggl:
Change Notes:
* Report model updated to match the BFF report model changes.
* Report View Model updated to consume and process the new BFF report model.
* A lot of files have small changes due to consolidating chart specific props into a general prop (ChartGroup). This makes generating the charts with D3 and building the objects in the view model a lot easier.
* The charts and views changed a considerable amount since the data we're collecting in the view model is a lot different.

PR #3197, opened 2025-11-18
chore: Delete experimental ui

laureanobrs:
This PR deletes the experimental UI considering the changes introduced throughout the following PRs:
#1535
#1495
#1459
#1447
#1427
#1360
#1369
#1370
#1345
#1352
#1340
#1333
#1218
#1203
#1204
#1198
#1174
#1143
#1141
#1139
#1085
#1033
#1006

Issue: #3108

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -152,10 +152,6 @@ bazel_dep(
     repo_name = "com_google_protobuf",
 )
 bazel_dep(
-    name = "aspect_bazel_lib",
-    version = "2.7.7",
-)
-bazel_dep(
     name = "abseil-cpp",
     version = "20240722.0.bcr.2",
     repo_name = "com_google_absl",
--- a/MODULE.bazel.lock
+++ b/MODULE.bazel.lock
@@ -21,8 +21,7 @@
     "https://bcr.bazel.build/modules/apple_support/1.5.0/MODULE.bazel": "50341a62efbc483e8a2a6aec30994a58749bd7b885e18dd96aa8c33031e558ef",
     "https://bcr.bazel.build/modules/aspect_bazel_lib/1.36.0/MODULE.bazel": "710d3560d8891d209f7985f3e4223011c3fefed0cd4d23d3e7b77b0f8287ef64",
     "https://bcr.bazel.build/modules/aspect_bazel_lib/2.7.2/MODULE.bazel": "780d1a6522b28f5edb7ea09630748720721dfe27690d65a2d33aa7509de77e07",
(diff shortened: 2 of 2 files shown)


laureanobrs (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3197)* status: 0 of 154 files reviewed, 1 unresolved discussion

___
*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3197#-OeOVyZxB9mhvWSxGLTL:-OeOVyZxB9mhvWSxGLTM:b-pstpu6) (no related file):*
In PRs #1360 and #1006 some grpc_gateway-related stuff were added that I believe are currently used by other components, so I didn’t remove anything related to that.  
In PR #1535 the CreateUIReport was added to the CLI. It’s not clear to me whether it should be removed.  
WDYT?

laureanobrs (commented):
> @laureanobrs reviewed 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3197)* status: all files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

___
*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3197#-OeOVyZxB9mhvWSxGLTL:-OeSx9sRDpA2G0vxs5qQ:bkllrht) (no related file):*

You're right that the gRPC gateway parts are used elsewhere now and need to stay.

CreateUiReport can be removed, as it was specifically to make a Report that's compatible with this UI. There's no need for it otherwise.
</blockquote></details>

Removed CreateUiReport (manually reverted PR #1535).
___
*[`MODULE.bazel` line 155 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3197#-OeSEM5j1ehUKGiSrt_H:-OeSwSDO8L_drZHFNJHc:b-4qh01x) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/469e31717628f9d7fc8ab14625d2c5ba5a8f721d/MODULE.bazel#L155)):*

Is this still used? I thought it was only there to make sure we were using a sufficiently new version for the other aspect modules.
</blockquote></details>

I didn’t find any direct uses of `aspect_bazel_lib` in our code.

Looking at `bazel mod graph | grep "aspect\|"`, I saw that it appears as a dependency of the bazel_dep entries matching `aspect_.*` that we removed in this PR, and also of `rules_oci`. See 
[https://github.com/bazel-contrib/rules\_oci/blob/main/MODULE.bazel#L8](https://github.com/bazel-contrib/rules_oci/blob/main/MODULE.bazel#L8).

It’s not clear to me whether it needs to be kept because of `rules_oci`.

Issue #3096, opened 2025-10-23
Implement internal ReportResults service


SanjayVas:
For MC API Phase II

PR #3206, opened 2025-11-20
refactor: Refer to unprocessed/processed instead of noisy/de-noised

SanjayVas:
In the Reporting internal API, the terms "noisy" and "de-noised" have been used incorrectly. The Report Processor does not de-noise, but rather processes Report results to add derived metrics and ensure constraints are met. Similarly in the case where differential privacy is not used, the unprocessed results aren't technically "noisy".

This updates references within that API. To avoid churn, it avoids any cases that would require a database schema update.

Issue: #3096


SanjayVas (commented):
> @SanjayVas reviewed 3 of 6 files at r3.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3206)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @stevenwarejones and @tristanvuong2021)

PR #3228, opened 2025-11-28
fix: use max page size for fetching event groups

georgi:
Issue: #3243 
Closes: #3243


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3228)

PR #3230, opened 2025-11-28
refactor: increase event groups list page size

Marco-Premier:
# SUMMARY

Increase the page size used when fetching CMMS EventGroups from the Public API to reduce the total number of gRPC calls required during synchronization. Previously, the page size was implicitly defaulting to 10, leading to large quantities of paginated requests.

This change explicitly sets pageSize = 500 in the listEventGroupsRequest, and defines LIST_EVENT_GROUPS_PAGE_SIZE as a constant for maintainability.


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3230)

PR #3231, opened 2025-11-28
fix: share grpc channel across data availability requests

Marco-Premier:
# SUMMARY

Multiple requests to DataAvailabilitySync are handled by the same Cloud Function instance. However, the existing callback responsible for closing the gRPC connection never executes because the JVM remains up and running between invocations. This leads to resource starvation and eventually requires a restart of the Cloud Function.

This PR introduces a fix by explicitly shutting down the channel every time the request is completed, ensuring proper cleanup and preventing starvation.

Issue: #3241 
Closes: #3241


Marco-Premier (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3231)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @georgi and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/dataavailability/DataAvailabilitySyncFunction.kt` line 188 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3231#-OfQOTXO77ph7TSd9Xyd:-OfZRY6D74EufJ8CS41-:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7fd31ab6ec275f6992eb389aa5b6321503ce83d9/src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/dataavailability/DataAvailabilitySyncFunction.kt#L188)):*

Use `lateinit var` with a non-null type rather than a nullable `var`.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/dataavailability/DataAvailabilitySyncFunction.kt` line 232 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3231#-OfQPa8b9cjKWHvMUHAE:-OfZRYVB1omDFG8IMadb:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7fd31ab6ec275f6992eb389aa5b6321503ce83d9/src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/dataavailability/DataAvailabilitySyncFunction.kt#L232)):*

The config isn't actually constant/static. Since the design has the config passed in the request, the TLS params are allowed to differ for each request. You technically therefore need a separate channel per unique TLS params.
</blockquote></details>

Done.

PR #3233, opened 2025-11-30
feat: Implement internal EventGroup batch create

roaminggypsy:
Issue: https://github.com/world-federation-of-advertisers/cross-media-measurement/issues/2450


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3233)

Issue #3216, opened 2025-11-25
Support configuration for default IQFs for CreateBasicReport


Saarahx:
**Describe the bug**
The MC API Phase II does not comply with the measurement policy requirements. Specifically:

- AMI and MRC filters are not applied by default for Video and Display.
- The API requires the caller to explicitly specify an Impression Qualification Filter (IQF), which contradicts the requirement that defaults should be applied automatically.
- The user should then additionally be able to apply Custom filters.

**Steps to reproduce**
N/A - Based on review of code and chat with Sanjay.

**Component(s) affected**

MC API Phase II
Impression Qualification Filters Configurations

Version
Phase II release

**Environment**
N/A

**Expected defaults**

**Video:**
AMI
MRC Standard: 100% in view, 2 continuous seconds

**Display:**
AMI
MRC Standard: 50% in view, 1 continuous second

**Actual behavior**
Caller must specify at least one IQF (impression_qualification_filter_id).
No default applied; API errors if field is empty.
AMI and MRC exist in configuration but are not auto-applied.

**Additional context**
Requirement: AMI must be default for both Video and Display without user intervention.

Requirements found here - https://docs.google.com/document/d/1gn3mjZGGid_ALFqMQewI6NhCwtPVm4v4mtIdc4A1j0Y/edit?tab=t.0#heading=h.1xkpval2ltml

<img width="368" height="62" alt="Image" src="https://github.com/user-attachments/assets/fa600cd0-1ab2-415c-ba56-e2c8f96bd304" />

**Developer notes confirm:**
AMI and MRC IDs exist but must be explicitly passed by the caller.
API does not support leaving IQF empty.

PR #3268, opened 2025-12-03
feat: Add base ImpressionQualificationFilters flags

tristanvuong2021:
Add base ImpressionQualificationFilters that will be used alongside user-set ImpressionQualificationFilters in BasicReport.

Issue: #3216 
RELNOTES: The Reporting public API server has a new `--base-impression-qualification-filter` option to specify the resource name of an ImpressionQualificationFilter resource which is always included in BasicReport creation. It can be specified multiple times.

### DIFF
--- a/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/common/server/V2AlphaPublicApiServer.kt
+++ b/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/common/server/V2AlphaPublicApiServer.kt
@@ -22,7 +22,7 @@ import com.google.protobuf.util.JsonFormat
 import io.grpc.Channel
 import io.grpc.ServerServiceDefinition
 import io.grpc.Status
-import io.grpc.StatusException
+import io.grpc.StatusRuntimeException
 import io.grpc.inprocess.InProcessChannelBuilder
 import java.io.File
 import java.security.SecureRandom
@@ -199,7 +199,7 @@ private object V2AlphaPublicApiServer {
             cmmsMeasurementConsumerId = measurementConsumerKey.measurementConsumerId
           }
         )
-      } catch (e: StatusException) {
+      } catch (e: StatusRuntimeException) {
         when (e.status.code) {
           Status.Code.ALREADY_EXISTS -> {}
           else -> throw e
@@ -225,7 +225,7 @@ private object V2AlphaPublicApiServer {
                 }
               )
             )
-          } catch (e: StatusException) {
+          } catch (e: StatusRuntimeException) {
             when (e.status.code) {
               Status.Code.NOT_FOUND -> {
                 throw IllegalArgumentException(


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3268)

SanjayVas (changes requested):
> @SanjayVas reviewed 11 of 18 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3268)* status: 11 of 18 files reviewed, 1 unresolved discussion (waiting on @kungfucraig and @tristanvuong2021)

___
*[`src/main/proto/wfa/measurement/internal/reporting/v2/basic_report_details.proto` line 35 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3268#-OffulUR4MUjhsl4LG5i:-OffulUS5LkhhpecSL31:bend5et) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7edf8c3331b978a464ae026b361d885563072af1/src/main/proto/wfa/measurement/internal/reporting/v2/basic_report_details.proto#L35)):*
> ```proto
> 
>   // The impression qualification filters for the BasicReport.
>   repeated ReportingImpressionQualificationFilter
> ```

The invariant should be that `effective_impression_qualification_filters` always contains at least `impression_qualification_filters`. The only time this won't be the case is for BasicReport entities that were written before this change. You therefore have three options for that case:

1. Explicitly document that the invariant might not hold for older entities, so that everyone reading this field needs to handle it.
2. Satisfy the invariant in the internal API by copying from the other field when it's read from the DB.
3. Ensure all existing BasicReports are deleted, since it has no real users.

tristanvuong2021 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3268)* status: 19 of 20 files reviewed, 3 unresolved discussions (waiting on @kungfucraig and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/BasicReportsService.kt` line 247 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3268#-OfkHxLm1EeRPYmnOlZK:-OfkN5yJ7bBjNzmZZmD7:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7d346aa0b0a127fa5973024cdb981284dbb7f68e/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/BasicReportsService.kt#L247)):*

Do we maybe want to do this on server startup? The base config doesn't change at runtime, so it seems wasteful to re-fetch them during CreateBasicReport.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/BasicReportsService.kt` line 249 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3268#-OfkH_DV7gNvr-vMh71f:-OfkN5fh9JP25SQ68WxB:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7d346aa0b0a127fa5973024cdb981284dbb7f68e/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/BasicReportsService.kt#L249)):*

Move this up to the initialization of `key` so that its type is non-nullable
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/gcloud/spanner/SpannerBasicReportsService.kt` line 360 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3268#-OfkH6PtEcjpwxsv0rOT:-OfkM0gE0uuFa1Gvx2e1:b-mj34h7) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7d346aa0b0a127fa5973024cdb981284dbb7f68e/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/gcloud/spanner/SpannerBasicReportsService.kt#L360)):*

nit: add a comment that this is to handle old DB entries
</blockquote></details>

I ended up removing this. I realized this doesn't make sense because when this is merged, new BasicReports created through the public API will always have it set. It wouldn't be set if this is directly called, which I do use for the new test case added to SpannerBasicReportsServiceTest, but wouldn't be done otherwise.

tristanvuong2021 (commented):
> @tristanvuong2021 reviewed 4 of 4 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3268)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @kungfucraig and @Marco-Premier)

PR #3270, opened 2025-12-03
build: Update deps for googleapis

### LINKED ISSUE #3266: Use googleapis module from Bazel Central Registry
At the time that this Bazel project was migrated to the Bzlmod dependency system, the googleapis dependency did not have a module published in the Bazel Central Registry (BCR). We therefore had to write our own in the [WFA Halo Bazel Registry](https://github.com/world-federation-of-advertisers/bazel-registry). Since then, it has been published to BCR with a slight incompatibility in usage that our transitive dependencies such as gRPC may rely on.

SanjayVas:
Closes #3266

Issue: #3266


SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3270)* status: 0 of 4 files reviewed, all discussions resolved (waiting on @laureanobrs)

___
*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3270#-OfWzXvjCEDBSgkDrVKI:-Of_YCCN7TuQun2RfYL4:b-896fix) (no related file):*

I'm hitting https://github.com/bazel-contrib/rules_jvm_external/issues/1498, so I think I need to downgrade everything back to rules_jvm_external 6.8. Hopefully that doesn't break anything else I did...
</blockquote></details>

Done.

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3270)* status: 2 of 4 files reviewed, 1 unresolved discussion (waiting on @laureanobrs)

___
*[`MODULE.bazel` line 296 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3270#-Of_lqGF7fQWUaWBUOc6:-Of_zSI17aPPnnJQHZHS:bep8m32) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/dce76f5cbf24295667d2811820d47d23e66dfdf4/MODULE.bazel#L296)):*

Are we switching from maven to coursier? If that’s the case, I think we should document the rationale.
</blockquote></details>

Oops, that was unintentional while I was debugging something. Switched back to the Maven resolver. That should also make the maven_install.json diff smaller.

Issue #3276, opened 2025-12-04
ImpressionQualificationFilterConfig is not Validated against the Event Descriptor


tristanvuong2021:
**Describe the bug**
The ImpressionQualificationFilterConfig isn't validated against the Event Descriptor.

**Steps to reproduce**
1. Use ImpressionQualificationFilterConfig in testing folder

**Component(s) affected**
Reporting Server

**Version**
5.28

**Environment**
GitHub  CorrectnessTest

PR #3277, opened 2025-12-04
fix: Validate IQF config against Event Descriptor

tristanvuong2021:
ImpressionQualificationFilterConfig is now validated against the Event descriptor.

Issue: #3276

### DIFF
--- a/src/main/docker/images.bzl
+++ b/src/main/docker/images.bzl
@@ -294,6 +294,11 @@ REPORTING_V2_COMMON_IMAGES = [
         repository = _PREFIX + "/reporting/v2/basic-reports-reports",
     ),
     struct(
+        name = "report_result_post_processor_image",
+        image = "//src/main/python/wfa/measurement/reporting/deploy/v2/common/job:post_process_report_result_job_executor_image",
+        repository = _PREFIX + "/reporting/v2/report-result-post-processor",
+    ),
+    struct(
         name = "reporting_spanner_update_schema_image",
         image = "//src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/gcloud/spanner/tools:update_schema_image",
         repository = _PREFIX + "/reporting/v2/spanner-update-schema",
--- a/src/main/python/wfa/measurement/reporting/deploy/v2/common/job/BUILD.bazel
+++ b/src/main/python/wfa/measurement/reporting/deploy/v2/common/job/BUILD.bazel
@@ -1,0 +1,21 @@
+load("@pip//:requirements.bzl", "requirement")
+load("@rules_python//python:defs.bzl", "py_binary")
+load("//build/rules_oci:defs.bzl", "py_image")
+
+package(default_visibility = ["//src:__subpackages__"])
+
+py_binary(
+    name = "post_process_report_result_job_executor",
+    srcs = ["post_process_report_result_job_executor.py"],
+    imports = ["../"],
+    deps = [
+        requirement("absl-py"),
+        "//src/main/python/wfa/measurement/reporting/job:post_process_report_result_job",
+    ],
+)
+
+py_image(
+    name = "post_process_report_result_job_executor_image",
+    binary = ":post_process_report_result_job_executor",
+    visibility = ["//src:docker_image_deployment"],
+)
--- a/src/main/python/wfa/measurement/reporting/deploy/v2/common/job/post_process_report_result_job_executor.py
+++ b/src/main/python/wfa/measurement/reporting/deploy/v2/common/job/post_process_report_result_job_executor.py
@@ -1,0 +1,147 @@
+# Copyright 2025 The Cross-Media Measurement Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""A one-shot job for fetching, correcting, and updating a report."""
+
+import os
+import grpc
+
+from absl import app
+from absl import flags
+from absl import logging
+
+from job import post_process_report_result_job
+
+_INTERNAL_REPORTING_TARGET = flags.DEFINE_string(
+    "internal_reporting_target",
+    None,
+    "The target for the internal reporting API.",
+    required=True,
+)
+
+_TLS_CERT_FILE = flags.DEFINE_string(
+    "tls_cert_file",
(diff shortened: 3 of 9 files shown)


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3277)

SanjayVas (changes requested):
> @SanjayVas reviewed 4 of 34 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3277)* status: 4 of 34 files reviewed, 4 unresolved discussions (waiting on @tristanvuong2021)

___
*[`src/test/kotlin/org/wfanet/measurement/common/EventDescriptorTest.kt` line 19 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3277#-OffzbvTD8ao3MDo8HNH:-OffzbvTD8ao3MDo8HNI:b-lvm3pc) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/827c35e408609d243b4c155b6b7c43a3da0ffcc8/src/test/kotlin/org/wfanet/measurement/common/EventDescriptorTest.kt#L19)):*
> ```Kotlin
> package org.wfanet.measurement.common
> 
> import com.google.common.truth.Truth
> ```

Revert this import change
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/internal/ImpressionQualificationFilterMapping.kt` line 33 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3277#-Offye3S6ZY4aOGW6uzl:-Offye3S6ZY4aOGW6uzm:b-36ly1a) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/827c35e408609d243b4c155b6b7c43a3da0ffcc8/src/main/kotlin/org/wfanet/measurement/reporting/service/internal/ImpressionQualificationFilterMapping.kt#L33)):*
> ```Kotlin
> class ImpressionQualificationFilterMapping(
>   config: ImpressionQualificationFilterConfig,
>   eventDescriptor: EventDescriptor?,
> ```

Why is this nullable? It seems that any case that we have ImpressionQualificationFilters we'd need to know which event message type they're for.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/internal/ImpressionQualificationFilterMapping.kt` line 33 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3277#-Ofg-KgQDdFGaux1b-lZ:-Ofg-KgQDdFGaux1b-l_:b-swjjpz) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/827c35e408609d243b4c155b6b7c43a3da0ffcc8/src/main/kotlin/org/wfanet/measurement/reporting/service/internal/ImpressionQualificationFilterMapping.kt#L33)):*
> ```Kotlin
> class ImpressionQualificationFilterMapping(
>   config: ImpressionQualificationFilterConfig,
>   eventDescriptor: EventDescriptor?,
> ```

There's now some redundancy in the SpannerReportResultsService constructor as it takes in both the event message descriptor and the ImpressionQualificationFilterMapping which in turn has a constructor parameter that derives from it.

The easy way to resolve it is to expose the underlying `Descriptors.Descriptor` as a public property from here. You could either do so by exposing it as a property on EventDescriptor, or just have this class take in the `Descriptors.Descriptor` and construct the `EventDescriptor` privately if nothing else needs that type.
___
*[`src/main/kotlin/org/wfanet/measurement/common/EventDescriptor.kt` line 17 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3277#-Offwj2L5_WNvUbAWZPG:-Offwj2M0FWAXsXHmcT3:bax62nz) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/827c35e408609d243b4c155b6b7c43a3da0ffcc8/src/main/kotlin/org/wfanet/measurement/common/EventDescriptor.kt#L17)):*
> ```Kotlin
>  */
> 
> package org.wfanet.measurement.common
> ```

This depends on API protos and therefore must not be in `common`. That package is for things that would otherwise be in common-jvm, i.e. not specific to any Halo API or component. `org.wfanet.measurement.api.v2alpha` would be more appropriate.

Speaking of which, there's an EventTemplates object there that has some similar code. Maybe add a TODO to reduce the duplication between the two? e.g. have EventDescriptor call EventTemplates, as the latter operates directly on the protobuf descriptor. Or just drop EventDescriptor entirely and expose a getTemplateFields function in EventTemplates similar to its existing `getGroupableFields` but without the GROUPABLE filter.

SanjayVas (changes requested):
> @SanjayVas reviewed 3 of 34 files at r1, 39 of 39 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3277)* status: all files reviewed, 2 unresolved discussions (waiting on @tristanvuong2021)

___
*[`src/test/kotlin/org/wfanet/measurement/reporting/service/internal/ImpressionQualificationFilterMappingTest.kt` line 34 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3277#-OfkDggwCtM-UexmU8QA:-OfkDggwCtM-UexmU8QB:buae4yv) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/68e66039fe7496ffac657fbad8c4156f0e52f83e/src/test/kotlin/org/wfanet/measurement/reporting/service/internal/ImpressionQualificationFilterMappingTest.kt#L34)):*
> ```Kotlin
> class ImpressionQualificationFilterMappingTest {
>   @Test
>   fun `processing valid config is successful`() {
> ```

nit: Avoid "is successful" and "succeeds" in test methods as it's not sufficiently specific



_Suggestion:_
```Kotlin
`constructor returns instance when config is valid`
```
___
*[`src/test/kotlin/org/wfanet/measurement/reporting/service/internal/ImpressionQualificationFilterMappingTest.kt` line 83 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3277#-OfkDIk70CybveDg-7KN:-OfkDIk70CybveDg-7KO:b3dtqj9) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/68e66039fe7496ffac657fbad8c4156f0e52f83e/src/test/kotlin/org/wfanet/measurement/reporting/service/internal/ImpressionQualificationFilterMappingTest.kt#L83)):*
> ```Kotlin
>     )
> 
>     assert(true)
> ```

1. You don't need this assertion.
2. We don't use any assertions other than Truth and assertFailsWith

SanjayVas (approved):
> @SanjayVas reviewed 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3277)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @tristanvuong2021)

PR #3235, opened 2025-12-01
refactor: accept max frequency for impression count

Marco-Premier:
…ler param

PR #3280, opened 2025-12-04
fix: missing frequency cap value passed to the fulfiller selector

Marco-Premier:
The previous PR for overriding the impression max frequency (#3235) had a bug that only allowed it be lowered, not raised.

Issue: #3246
Closes: #3246

### DIFF
--- a/src/main/k8s/dev/duchy_gke.cue
+++ b/src/main/k8s/dev/duchy_gke.cue
@@ -117,7 +117,7 @@ duchy: #SpannerDuchy & {
 	}
 	_kingdom_system_api_target:       #KingdomSystemApiTarget
 	_kingdom_public_api_target:       #KingdomPublicApiTarget
-	_trusteeMillSubnetworkCidrRange:  _trusteeMillSubnetworkCidrRange
+	_trusteeMillSubnetworkRange:      _trusteeMillSubnetworkCidrRange
 	_blob_storage_flags:              _cloudStorageConfig.flags
 	_verbose_grpc_logging:            "false"
 	_duchyMillParallelism:            4
--- a/src/main/k8s/duchy.cue
+++ b/src/main/k8s/duchy.cue
@@ -34,7 +34,7 @@ import ("strings")
 	_shareShuffleWorkLockDuration?:    string
 	_kingdom_system_api_target:        string
 	_kingdom_public_api_target:        string
-	_trusteeMillSubnetworkCidrRange?:  string
+	_trusteeMillSubnetworkRange?:      string
 	_blob_storage_flags: [...string]
 	_verbose_grpc_logging: "true" | "false"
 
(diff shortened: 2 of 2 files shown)


SanjayVas (changes requested):
> @SanjayVas reviewed 2 of 4 files at r1, 1 of 1 files at r2, 1 of 2 files at r3, 2 of 2 files at r4.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3280)* status: 7 of 10 files reviewed, 1 unresolved discussion (waiting on @kungfucraig and @Marco-Premier)

___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/common/FrequencyVectorBuilder.kt` line 63 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3280#-OffaW0W9lhBUFfZbAYi:-OffaW0W9lhBUFfZbAYj:b-xy8qew) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0a4d5c54569f35831fff369cfedeab1d96e9f9e6/src/main/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/common/FrequencyVectorBuilder.kt#L63)):*
> ```Kotlin
>   val populationSpec: PopulationSpec,
>   val measurementSpec: MeasurementSpec,
>   val overrideImpressionMaxFrequencyPerUser: Int?,
> ```

I'm not seeing where this is used by the builder itself. There shouldn't be public properties here just to have the builder hold them. I'm not sure why the other properties of the builder are public.

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3280)* status: 7 of 10 files reviewed, 2 unresolved discussions (waiting on @kungfucraig and @Marco-Premier)

___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/common/FrequencyVectorBuilder.kt` line 63 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3280#-OffaW0W9lhBUFfZbAYi:-OffddW9A1BmhCquafAX:b-ruxtt5) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0a4d5c54569f35831fff369cfedeab1d96e9f9e6/src/main/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/common/FrequencyVectorBuilder.kt#L63)):*

Nevermind, I see that it was already in use.

nit: make the property private
</blockquote></details>

Or rather, make it not a property at all since it looks like it's only used by the `init` block.
___
*[`src/main/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/common/FrequencyVectorBuilder.kt` line 90 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3280#-OffdtHV9O12DPTU7I6d:-OffdtHV9O12DPTU7I6e:b-fhl84d) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/85c8911e6391851e5242d674ab8d3eb336c12aba/src/main/kotlin/org/wfanet/measurement/eventdataprovider/requisition/v2alpha/common/FrequencyVectorBuilder.kt#L90)):*
> ```Kotlin
>       }
>     }
> ```

nit: validate that overrideImpressionMaxFrequencyPerUser is >=1 if specified

Issue #3290, opened 2025-12-05
Population Metric does not create a Population Measurement


tristanvuong2021:
**Describe the bug**
The Reporting Metrics Service does not actually create a Population Measurement. The RequisiitonSpec created does not set population.

**Steps to reproduce**
1. Create a Population Metric
2. Check that state is always FAILED after processing.

**Component(s) affected**
Reporting Server

**Version**
5.28

**Environment**
local

PR #3291, opened 2025-12-05
fix!: Metrics Service now correctly creates a Population Measurement for a Population Metric

tristanvuong2021:
The Metrics Service is fixed to create a RequisitionSpec with population set.

Issue: #3290
BREAKING-CHANGE: The Reporting public API server has a new required `--pdp-name` option. This is the resource name of the Population Data Provider (PDP) corresponding to the running Population Requisition Fulfiller instance.


stevenwarejones (approved):
> @stevenwarejones reviewed 20 of 22 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3291)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @tristanvuong2021)

PR #3292, opened 2025-12-05
fix: set max frequency below the ring modulus

stevenwarejones:
The default max frequency is limited by two things:
1. The size in the frequency vector (Byte.MAX_VALUE)
2. The Ring Modulus used for HMShuffle

### DIFF
--- a/MODULE.bazel.lock
+++ b/MODULE.bazel.lock
@@ -2546,7 +2546,7 @@
         "recordedFileInputs": {
           "@@rules_python~//tools/publish/requirements_linux.txt": "d576e0d8542df61396a9b38deeaa183c24135ed5e8e73bb9622f298f2671811e",
           "@@rules_fuzzing~//fuzzing/requirements.txt": "ab04664be026b632a0d2a2446c4f65982b7654f5b6851d2f9d399a19b7242a5b",
-          "@@//requirements.txt": "f406321045147912c3bc34beacb815a8162809ca28a7e6af32258f52998032b1",
+          "@@//requirements.txt": "30923dfae55cd1f9f22232c3e47572919853847e709f37841e695bdccec31311",
           "@@rules_python~//tools/publish/requirements_windows.txt": "d18538a3982beab378fd5687f4db33162ee1ece69801f9a451661b1b64286b76",
           "@@protobuf~//python/requirements.txt": "983be60d3cec4b319dcab6d48aeb3f5b2f7c3350f26b3a9e97486c37967c73c5",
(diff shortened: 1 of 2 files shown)


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3292)

kungfucraig (changes requested):
> @kungfucraig reviewed 2 of 2 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3292)* status: all files reviewed, 1 unresolved discussion (waiting on @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/computation/KAnonymityParams.kt` line 32 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3292#-OfjmA0BBo_hHLteGc2n:-OfjmA0BBo_hHLteGc2o:b-4hwza1) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/245a9c50f51bc91453efd991fcfc7b765ffa4f11/src/main/kotlin/org/wfanet/measurement/computation/KAnonymityParams.kt#L32)):*
> ```Kotlin
>   val minUsers: Int,
>   val minImpressions: Int,
>   val reachMaxFrequencyPerUser: Int = 126,
> ```

Can these two things reference the same constant, perhaps defined in a third file?

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3292)* status: 0 of 5 files reviewed, 1 unresolved discussion (waiting on @kungfucraig)

___
*[`src/main/kotlin/org/wfanet/measurement/computation/KAnonymityParams.kt` line 32 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3292#-OfjmA0BBo_hHLteGc2n:-OfjwnnkChQESVNqjrbe:b-dktbsg) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/245a9c50f51bc91453efd991fcfc7b765ffa4f11/src/main/kotlin/org/wfanet/measurement/computation/KAnonymityParams.kt#L32)):*

Unfortunately, this is part of the protocol config configured in the kingdom. I'd think we'd need to think about how to maintain consistency across these configs. we could add a param for the kingdom ring modulus. I don't want to have to add that param since we are moving to TrusTee anyways.
</blockquote></details>

actually - went ahead and put it in a separate params file

PR #3297, opened 2025-12-05
ci: Fix Update-cmms dependency graph

### LINKED ISSUE #3296: update-edp-aggregator-services workflow sometimes fails with `deployments.apps "default-collector" not found`
Example failing run: https://github.com/world-federation-of-advertisers/cross-media-measurement/actions/runs/19973044129/job/57284040328

This is likely because the rollout check lists all Deployments and then checks the rollout status of each. However the EDP aggregator services live in the same cluster as the Kingdom, meaning there are Deployments that may be being updated as part of a concurrently-running update-kingdom workflow

It should be sufficient to have this workflow depend on the update-kingdom workflow to ensure that the rollouts have already completed for the other Deployments in the cluster.

laureanobrs:
Closes #3296 

Issue: #3296

### DIFF
--- a/.github/workflows/update-cmms.yml
+++ b/.github/workflows/update-cmms.yml
@@ -177,7 +177,7 @@ jobs:
 
   update-edp-aggregator-service:
     uses: ./.github/workflows/configure-edp-aggregator.yml
-    needs: [publish-images, terraform]
+    needs: [publish-images, terraform, update-kingdom]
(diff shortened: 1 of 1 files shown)


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3297)

Issue #3302, opened 2025-12-08
CreateBasicReport does not request Population correctly


tristanvuong2021:
**Describe the bug**
CreateBasicReport includes IQF filters for MetricCalculationSpecs that have a Population MetricSpec. It should only use DimensionSpec filters for Population MetricSpec.

**Steps to reproduce**
1. Call CreateBasicReport

**Component(s) affected**
Reporting Server

**Version**
v5.29

**Environment**
local

PR #3303, opened 2025-12-08
fix: CreateBasicReport does not include IQF filters in MetricCalculationSpecs that have a Population MetricSpec

tristanvuong2021:
CreateBasicReport does not include IQF filters in MetricCalculationSpecs that have a Population MetricSpec. DimensionSpec filters only MetricCalculationSpecs are used for Population MetricSpecs.

Issue: #3302

### DIFF
--- a/src/main/resources/kingdom/spanner/changelog.yaml
+++ b/src/main/resources/kingdom/spanner/changelog.yaml
@@ -110,3 +110,6 @@ databaseChangeLog:
 - include:
     file: add-population-details.sql
     relativeToChangeLogFile: true
+- include:
+    file: create-event-group-activities-table.sql
+    relativeToChangeLogFile: true
--- a/src/main/resources/kingdom/spanner/create-event-group-activities-table.sql
+++ b/src/main/resources/kingdom/spanner/create-event-group-activities-table.sql
@@ -1,0 +1,39 @@
+-- liquibase formatted sql
+
+-- Copyright 2025 The Cross-Media Measurement Authors
+--
+-- Licensed under the Apache License, Version 2.0 (the "License");
+-- you may not use this file except in compliance with the License.
+-- You may obtain a copy of the License at
+--
+--     http://www.apache.org/licenses/LICENSE-2.0
+--
+-- Unless required by applicable law or agreed to in writing, software
+-- distributed under the License is distributed on an "AS IS" BASIS,
+-- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+-- See the License for the specific language governing permissions and
+-- limitations under the License.
+
+-- changeset lindreamdeyi:32 dbms:cloudspanner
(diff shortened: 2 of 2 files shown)


SanjayVas (approved):
> @SanjayVas reviewed 2 of 7 files at r1, 5 of 5 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3303)* status: all files reviewed, 1 unresolved discussion (waiting on @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/job/BasicReportsReportsJob.kt` line 276 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3303#-Og-WdJD2t5gKt7E-QaM:-Og-WdJD2t5gKt7E-QaN:bo8xzuc) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d81ef342118d42ec25ae7fbac6494e7f835cac87/src/main/kotlin/org/wfanet/measurement/reporting/job/BasicReportsReportsJob.kt#L276)):*
> ```Kotlin
>           // Population Metrics only use DimensionSpec because ImpressionQualificationFilter filters
>           // includes non-Population fields, and the ReportingSet is irrelevant so the ReportingSet
>           // to use to lookup the result in the other map is unknown.
> ```

nit: this is hard to read



_Suggestion:_
```Kotlin
          // Population results are not keyed by IQF and ReportingSet, as IQFs include non-Population fields and the ReportingSet EventGroups are irrelevant to the Population Measurement.
        
```

stevenwarejones (approved):
> @stevenwarejones reviewed 2 of 7 files at r1, 4 of 5 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3303)* status: all files reviewed, 1 unresolved discussion (waiting on @Marco-Premier and @tristanvuong2021)

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/job/BasicReportsReportsJob.kt` line 793 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3303#-OgA4B1L5k51qBZuG6t3:-OgA4B1L5k51qBZuG6t4:b-zh7673) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7d9c3a17c334618bd35fe6c854aca67ee0748e03/src/main/kotlin/org/wfanet/measurement/reporting/job/BasicReportsReportsJob.kt#L793)):*
> ```Kotlin
>     val dimensionSpecFilters: List<EventFilter>,
>     // CEL String created from dimensionSpecFilters
>     val dimensionSpecFilter: String,
> ```

if this can be derived from the dimensionSpecFilters, is it just here for convenience?

PR #3305, opened 2025-12-09
refactor: use join key as request id when creating an event group

Marco-Premier:
unique request ids should be created based on the joined key of the MC ID and the event group reference id.

Issue: #3321


tristanvuong2021 (approved):
> @tristanvuong2021 reviewed 1 of 1 files at r1.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3305)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @SanjayVas)

PR #3311, opened 2025-12-11
build: Use Python gRPC Bazel rules from grpc rather than rules_proto_grpc_python

### LINKED ISSUE #3327: Incompatible protobuf gencode/runtime versions in report result processor
Error when running the report result processor container image for MC API Phase II:

```
google.protobuf.runtime_version.VersionError: Detected incompatible Protobuf Gencode/Runtime versions when loading wfa/measurement/internal/reporting/v2/basic_report.proto: gencode 5.29.5 runtime 5.27.2. Runtime version cannot be older than the linked gencode version. See Protobuf version guarantees at https://protobuf.dev/support/cross-version-runtime-guarantee.
```

This seems to stem from Python dependency management. We're currently using the third-party `rules_proto_grpc_python` rather than first-party gRPC rules for Python from the `protobuf` Bazel dep, which may be causing some issues. We also have a relatively old version of the `protobuf` dep. Finally, we're not using a lockfile for PyPi dependencies like we do for Maven.

SanjayVas:
This addresses an issue with mismatched codegen and runtime protobuf versions in Python images which use gRPC.

Closes #3327

Issue: #3327


SanjayVas:
CC @ple13

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3311)* status: all files reviewed, 2 unresolved discussions (waiting on @laureanobrs)

___
*[`BUILD.bazel` line 13 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3311#-OgEyft_6LM0C0KS1b2u:-OgF-7RVBXy8A-fuC6Mg:btkmlpo) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/b6853efc64ccea410b5b29a3ead38f4a1d7da065/BUILD.bazel#L13)):*

How should we re-generate the lock when modifying the pyproject.toml?

Should we run
`bazel run //:requirements.update`
?
</blockquote></details>

Yes. Those instructions are included at the top of the generated file.
___
*[`MODULE.bazel` line 432 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3311#-OgEw3Sj7bJ83SLPLR42:-OgF-NJj23i9VMbZwbCQ:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/b6853efc64ccea410b5b29a3ead38f4a1d7da065/MODULE.bazel#L432)):*

DO_NOT_SUBMIT instead of TODO?
</blockquote></details>

Done.

laureanobrs (changes requested):
> @laureanobrs reviewed 4 of 6 files at r4, 12 of 13 files at r5, 1 of 1 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3311)* status: all files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/measurementconsumer/stats/Variances.kt` line 26 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3311#-OgJMZHC3vyXS2wOQt9M:-OgJTcaA98emz6YBDsBI:b-xx5pip) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/afbd5bfe2f55158ab3eab73809d31c3eeabdd4b0/src/main/kotlin/org/wfanet/measurement/measurementconsumer/stats/Variances.kt#L26)):*

FYI This was incorrectly relying on a transitive Guava dependency, which resulted in breakage when the transitive dependency was removed. Since the library already relies on Apache Commons Numbers and Apache Commons Math3, I swapped to that instead.
</blockquote></details>

Cool. 
Quick question, I think the new alternative handles NaNs in a different way. Is that relevant for this case?

SanjayVas (commented):
> @SanjayVas reviewed 1 of 1 files at r6.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3311)* status: all files reviewed, 1 unresolved discussion (waiting on @laureanobrs)

___
*[`src/main/kotlin/org/wfanet/measurement/measurementconsumer/stats/Variances.kt` line 26 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3311#-OgJMZHC3vyXS2wOQt9M:-OgJ_u7QACWgIPZEA1E_:brpsr8f) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/afbd5bfe2f55158ab3eab73809d31c3eeabdd4b0/src/main/kotlin/org/wfanet/measurement/measurementconsumer/stats/Variances.kt#L26)):*

Cool. 
Quick question, I think the new alternative handles NaNs in a different way. Is that relevant for this case?
</blockquote></details>

No. I can't think of a way for the sum to be NaN.

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3311)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @SanjayVas)

___
*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3311#-OgY_AFG2hLqdN0L5cdD:-OgY_rRz-NkoepxRD3wt:b-u34rur) (no related file):*

The last time I checked the PR, these tests were failing:  
```
//src/test/python/wfa/measurement/reporting/postprocessing/noiseninja:solver_test  
//src/test/python/wfa/measurement/reporting/postprocessing/report:report_test
```

Were they fixed, or would the tests potentially be flaky?
</blockquote></details>

These were fixed by specifying the right versions of the solver dependencies.

laureanobrs (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3311)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @SanjayVas)

PR #3318, opened 2025-12-12
fix(report-processor): fix large correction test for unnoised edps.

ple13:
This uses a mock for the Solver to avoid having to maintain inputs that get the exact desired response from the real Solver. That would otherwise be difficult to keep up-to-date with changes to the underlying solver implementations.

### DIFF
--- a/src/test/python/wfa/measurement/reporting/postprocessing/report/report_test.py
+++ b/src/test/python/wfa/measurement/reporting/postprocessing/report/report_test.py
@@ -14,10 +14,13 @@
 
 import unittest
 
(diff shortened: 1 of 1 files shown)


stevenwarejones (approved):
> @stevenwarejones reviewed 1 of 1 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3318)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @ple13)

PR #3341, opened 2025-12-16
fix: Update TrusTEE mill to use StreamingAead

renjiezh:


SanjayVas (approved):
> @SanjayVas reviewed 5 of 5 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3341)* status: all files reviewed, 1 unresolved discussion (waiting on @renjiezh)

___
*[`src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/TrusTeeMill.kt` line 224 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3341#-Ogd1sq87yQtUE4DgVsm:-Ogd1sq87yQtUE4DgVsn:b9f9xcb) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e8393860aa90f24080c09f7429ef4619db3a62ad/src/main/kotlin/org/wfanet/measurement/duchy/mill/trustee/TrusTeeMill.kt#L224)):*
> ```Kotlin
>   }
> 
>   private fun decryptRequisitionData(dek: KeysetHandle, data: ByteString): ByteArray {
> ```

nit: This isn't really taking advantage of the streaming aspect. The idea is to pass in each chunk while reading the blob, meaning you'd need to change getRequisitionData. See StreamingAead.decrypt in common-jvm.

Of course, there's less benefit if you need the whole plaintext in memory, but it still avoids having both the full plaintext and ciphertext in memory at the same time.

Maybe leave a TODO?

Issue #2345, opened 2025-05-22
Make model_line required in metric_calculation_spec proto and metric.proto


tristanvuong2021:
These fields will be added as part of [#2308](https://github.com/world-federation-of-advertisers/cross-media-measurement/pull/2308), but they won't be required initially. They will need to be required when the time comes.

PR #3402, opened 2026-01-01
feat! Make model_line required in Metric and MetricCalculationSpec

tristanvuong2021:
model_line is now required in both Metric and MetricCalculationSpec.

BREAKING CHANGE: Two flags, "--default-vid-model-line" and "--measurement-consumer-model-line" have been removed, and the model_line field in both Metric and MetricCalculationSpec are now required.

Issue: #2345


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3402)

PR #3418, opened 2026-01-07
fix: Include media type predicates in MetricCalculationSpec filters from IQFs

### LINKED ISSUE #3417: Presence check for ImpressionQualificationFilter applied to term field instead of its parent
### Bug description

For example, if an IQF has a filter with the following term:

```textproto
path: "display.viewable"
value {
  bool_value: true
}
```

the resulting CEL expression contains `has(display.viewable) && display.viewable == true` where it's expected to be `has(display) && display.viewable == true`

### Steps To Reproduce

1. Call CreateBasicReport referencing an IQF with at least one filter.

### Component(s) affected

Reporting (MC API)

### Version

v0.5.30

### Environment

Local

### Additional context

_No response_

SanjayVas:
This also removes unnecessary empty terms and parentheses from the resulting CEL expression in MetricCalculationSpec.

Closes #3417

Issue: #3417

### DIFF
--- a/src/main/kotlin/org/wfanet/measurement/eventdataprovider/eventfiltration/validation/EventFilterValidationException.kt
+++ b/src/main/kotlin/org/wfanet/measurement/eventdataprovider/eventfiltration/validation/EventFilterValidationException.kt
@@ -14,7 +14,8 @@
 
 package org.wfanet.measurement.eventdataprovider.eventfiltration.validation
 
-class EventFilterValidationException(val code: Code, message: String) : Exception(message) {
+class EventFilterValidationException(val code: Code, message: String, cause: Throwable? = null) :
+  Exception(message, cause) {
 
   enum class Code(val description: String) {
     INVALID_CEL_EXPRESSION(
--- a/src/main/kotlin/org/wfanet/measurement/eventdataprovider/eventfiltration/validation/EventFilterValidator.kt
+++ b/src/main/kotlin/org/wfanet/measurement/eventdataprovider/eventfiltration/validation/EventFilterValidator.kt
@@ -220,7 +220,8 @@ object EventFilterValidator {
       } catch (e: Exception) {
         throw EventFilterValidationException(
           EventFilterValidationException.Code.INVALID_CEL_EXPRESSION,
-          e.message ?: "Compiling expression threw an unexpected exception",
+          "Failed to compile CEL expression $celExpression",
+          e,
         )
       }
     failOnInvalidExpression(astAndIssues.issues)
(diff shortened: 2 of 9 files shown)


tristanvuong2021 (commented):
> @tristanvuong2021 reviewed 3 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3418)* status: 3 of 5 files reviewed, 1 unresolved discussion (waiting on @SanjayVas).

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/BasicReportTransformations.kt` line 184 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3418#-OiL-oipCOb-vlrasGyI:-OiL5OdJ-0S8dfnMuAiy:bp0p949) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f6085433bba9a798d137558bfd12f1080d375808/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/BasicReportTransformations.kt#L184)):*

Though actually if there is in fact media type filtering already in place, then we theoretically shouldn't even need the presence check here. If there isn't other media type filtering, then we should be applying this at the filter spec level rather than the filter level (i.e. putting a presence check for every template that has the matching media type).
</blockquote></details>

The media type filtering is relying on this. So there would need to be a mapping of media type to parent field to use here.

tristanvuong2021 (approved):
> @tristanvuong2021 made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3418)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @SanjayVas).

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/BasicReportTransformations.kt` line 182 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3418#-OiO7KI-2HwKCZQK4DVe:-OiOUMmL3AOTPacLJfOy:bamrox) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ee1a12bcf4b5a8273d59a1a1f18d9ba09e02b308/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/BasicReportTransformations.kt#L182)):*

nit: This does end up being called after the filter specs media type is validated against the filter since all the validation is done previously, so only the add portion is actually necessary.
</blockquote></details>

Nevermind, I read this wrong

SanjayVas (commented):
> @SanjayVas made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3418)* status: all files reviewed, 1 unresolved discussion (waiting on @SanjayVas).

___
*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3418#-OiQUn0w7LK6OkhGS93F:-OiQUn0x5sx_X8RIaHZ6:bovcc9b) (no related file):*
Ran into a snag: the `has()` macro does not work for top-level fields. We can't use that to implement media type checks. Working on an alternate solution...

tristanvuong2021 (approved):
> @tristanvuong2021 reviewed 7 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3418)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @stevenwarejones and @tristanvuong2021).

PR #3456, opened 2026-01-13
feat: data watcher propagates resource id headers

georgi:
This change is required to propagate the resource id header to the deletion function.
The resource name of the impression metadata is stored on the object metadata of the blob.
This avoids unnecessary API lookups, which are only needed when we don't know the resource name on the API.

Issue: #3469


georgi (commented):
> @georgi made 1 comment and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3456)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @Marco-Premier).

___
*[`src/main/kotlin/org/wfanet/measurement/securecomputation/deploy/gcloud/datawatcher/DataWatcherFunction.kt` line 63 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3456#-Oiwi2qHBPhn5Nnz_IWM:-OjR9MWdAeeUVT7xtQ29:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/12e605788682f20733e5ad30595e03e30c35a091/src/main/kotlin/org/wfanet/measurement/securecomputation/deploy/gcloud/datawatcher/DataWatcherFunction.kt#L63)):*

is the `DataAvailabilityCleanupFunction` the function responsible for receiving the objectDelete event? 

I guess you set up that one in the terraform pr
</blockquote></details>

Done.

kungfucraig (approved):
> @kungfucraig reviewed 7 files and all commit messages, and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3456)* status: all files reviewed, 1 unresolved discussion (waiting on @Marco-Premier).

PR #3464, opened 2026-01-15
fix: Always specify methodology for Measurement results from Duchy

SanjayVas:
The required methodology fields in Measurement.Result were only being specified for the HMSS protocol. This meant that no methodology was specified for TrusTEE, which prevented the correctness test harness (MeasurementConsumerSimulator) from being able to compute variance when that protocol was being used.

Issue: #2592


stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3464)* status: 14 of 15 files reviewed, 2 unresolved discussions (waiting on @kungfucraig and @SanjayVas).

___
*[`src/main/kotlin/org/wfanet/measurement/measurementconsumer/stats/Covariances.kt` line 108 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3464#-Oj2CDGC6XNLmBRVaCGD:-Oj2CDGC6XNLmBRVaCGE:b-xzcoxn) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/de8cb3db40ffe9cf8aec71c08a73f84f243c6d45/src/main/kotlin/org/wfanet/measurement/measurementconsumer/stats/Covariances.kt#L108)):*
> ```Kotlin
>         }
>         is DeterministicMethodology,
>         is HonestMajorityShareShuffleMethodology -> {
> ```

why would this be needed for HMSS now?

SanjayVas (commented):
> @SanjayVas made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3464)* status: all files reviewed, 2 unresolved discussions (waiting on @kungfucraig and @stevenwarejones).

___
*[`src/main/kotlin/org/wfanet/measurement/measurementconsumer/stats/Covariances.kt` line 108 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3464#-Oj2CDGC6XNLmBRVaCGD:-Oj6u-Bw7GSbZTV4cKy7:b-t6m3bg) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/de8cb3db40ffe9cf8aec71c08a73f84f243c6d45/src/main/kotlin/org/wfanet/measurement/measurementconsumer/stats/Covariances.kt#L108)):*

it just seems inconsistent. the approach for methodology for hmss vs trustee - either I'm missing some design change or some other context - maybe we can discuss on the Monday call?
</blockquote></details>

1. The line of code this comment is referencing for covariance indeed uses the same approach for both HMSS and TrusTEE.
2. This PR does not actually make any changes to how variance and covariance are computed. It only changes how the methodology is communicated when setting a Measurement result from a Duchy. It fixes a bug where the methodology was only being set for HMMS rather than for all protocols.

Now if you're referring to the fact that the methodology is different between TrusTEE and HMSS for variance (see Variances.kt rather than this file), this is true and was already true before this PR. HMMS variance depends on the frequency vector size, whereas TrusTEE variance depends on sampling interval. You can ask @kungfucraig for any explanation as I don't actually know why that's true, but it's also irrelevant to this PR.

I will note that even if there is an unintentional inconsistency - e.g. if for some reason we *could* use the same approach to compute HMMS variance, I don't think we'd bother with the API changes necessary to fix it now. We'd just wait for HMSS to be deprecated, which it will be once TrusTEE is rolled out everywhere.

PR #3485, opened 2026-01-21
Add model line mapping support for EDP aggregator

stevenwarejones:
Covers two use cases:
1. Markets desiring that more than one model line point to the same piece of data
2. A model line name change after impression data has already been placed into storage

Supports both results fulfillment mapping as well as data availability mapping.

Issue: #3492


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3485)

PR #3491, opened 2026-01-23
feat: implement spanner persistence layer for ClientAccounts

getina:
implements SpannerClientAccountsService with create, delete, get, and stream operations. includes reader, writers, and tests

Issue: #3594

### DIFF
--- a/.github/workflows/publish-cloud-run-function-uber-jars.yml
+++ b/.github/workflows/publish-cloud-run-function-uber-jars.yml
@@ -118,3 +118,14 @@ jobs:
           artifact-group-id: org.wfanet.measurement.edpaggregator.deploy.gcloud.dataavailability
           bazel_target_label: //src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/dataavailability:DataAvailabilitySyncFunction_deploy.jar
 
+      - name: Build and Deploy Data Availability Cleanup
+        env:
+          MAVEN_USERNAME: ${{ github.actor }}
+          MAVEN_PASSWORD: ${{ secrets.GITHUB_TOKEN }}
+        uses: ./.github/actions/build-and-push-uber-jar
+        with:
+          package-version: ${{ steps.get-artifact-version.outputs.artifact-version }}
+          artifact-id: data-availability-cleanup
+          artifact-group-id: org.wfanet.measurement.edpaggregator.deploy.gcloud.dataavailability
+          bazel_target_label: //src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/dataavailability:DataAvailabilityCleanupFunction_deploy.jar
+
--- a/.github/workflows/terraform-cmms.yml
+++ b/.github/workflows/terraform-cmms.yml
@@ -62,6 +62,7 @@ jobs:
       GCLOUD_REGION: ${{ vars.GCLOUD_REGION }}
       GCLOUD_ZONE: ${{ vars.GCLOUD_ZONE }}
       DATA_WATCHER_CONFIG_CONTENT: ${{ vars.DATA_WATCHER_CONFIG_CONTENT }}
+      DATA_WATCHER_DELETE_CONFIG_CONTENT: ${{ vars.DATA_WATCHER_DELETE_CONFIG_CONTENT }}
       REQUISITION_FETCHER_CONFIG_CONTENT: ${{ vars.REQUISITION_FETCHER_CONFIG_CONTENT }}
       EDPA_EDPS_CONFIG: ${{ vars.EDPA_EDPS_CONFIG }}
       RESULTS_FULFILLER_POPULATION_SPEC_CONTENT: ${{ vars.RESULTS_FULFILLER_POPULATION_SPEC_CONTENT }}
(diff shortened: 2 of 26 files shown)


getina (commented):
> @getina made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3491)* status: 0 of 16 files reviewed, 2 unresolved discussions (waiting on @Marco-Premier and @stevenwarejones).

___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/SpannerClientAccountsService.kt` line 76 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3491#-Ojz_qIDDNB25A-JzAu-:-Ok0TM_t-bTwdDHlZr40:bb7f2mf) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/89cd479d9f8b5751ce34e76a961a1f4dbe97135f/src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/SpannerClientAccountsService.kt#L76)):*

The `client_account_service.proto` now defines GetClientAccountRequest with a oneof `external_parent_id` that allows either `external_measurement_consumer_id` or `external_data_provider_id` as the parent, but this implementation only handles the MeasurementConsumer case.

As per the existing pattern in `SpannerEventGroupsService.getEventGroup()`, this should use a when expression on request.externalParentIdCase to handle both parent types.

```
@Suppress("WHEN_ENUM_CAN_BE_NULL_IN_JAVA")
 return when (request.externalParentIdCase) {
   GetClientAccountRequest.ExternalParentIdCase.EXTERNAL_MEASUREMENT_CONSUMER_ID -> {
     // lookup by MeasurementConsumer
   }
   GetClientAccountRequest.ExternalParentIdCase.EXTERNAL_DATA_PROVIDER_ID -> {
     // lookup by DataProvider using ClientAccountsByDataProviderAndExternalId index
   }
   GetClientAccountRequest.ExternalParentIdCase.EXTERNALPARENTID_NOT_SET ->
     throw Status.INVALID_ARGUMENT.withDescription("external_parent_id not specified")
       .asRuntimeException()
 }
```
</blockquote></details>

done
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/SpannerClientAccountsService.kt` line 98 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3491#-OjzaNmD18TzSoYzhsZI:-Ok0TN02DTx7s31HhOxI:bb7f2mf) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/89cd479d9f8b5751ce34e76a961a1f4dbe97135f/src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/SpannerClientAccountsService.kt#L98)):*

Same issue as `getClientAccount`. The proto defines `DeleteClientAccountRequest` with a oneof `external_parent_id` supporting both MeasurementConsumer and DataProvider parents, but only the MeasurementConsumer case is handled here.
</blockquote></details>

done

getina (commented):
> @getina reviewed 1 file.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3491)* status: 8 of 17 files reviewed, 6 unresolved discussions (waiting on @Marco-Premier, @SanjayVas, and @stevenwarejones).

SanjayVas (approved):
> @SanjayVas partially reviewed 11 files and all commit messages, made 2 comments, and resolved 4 discussions.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3491)* status: all files reviewed, 4 unresolved discussions (waiting on @getina, @Marco-Premier, and @stevenwarejones).

___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/writers/CreateClientAccount.kt` line 70 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3491#-OkVglEi3p5Ua_16Adlx:-OkVglEi3p5Ua_16Adly:bz1cnm3) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/71ef6746968c59d944d2501b049dad769833d85f/src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/writers/CreateClientAccount.kt#L70)):*
> ```Kotlin
> 
>     do {
>       attempts++
> ```

nit: This can be simplified a bit.

You don't actually need a max number of attempts. It's an int64 ID space, so even if we only use positive integers that's still 63 bits to find something non-conflicting. Most times it'll find good IDs on the first try.

You can therefore end up with something like

```kotlin
var clientAcccountId: InternalId
do {
  clientAccountId = idGenerator.generateInternalId()
} while (ClientAccountReader.existsWithInternalId(transactionContext, measurementConsumerId, clientAccountId))
```

and then the same for the external ID.

If desired, you could even add generateNewInternalId and generateNewExternalId extension functions for this IdGenerator type similar to the [`generateNewId` method](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/9b31f0641eaa3494d203687987b94b2e24e862cd/src/main/kotlin/org/wfanet/measurement/common/IdGenerator.kt#L40) on org.wfanet.measurement.common.IdGenerator.
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/readers/ClientAccountReader.kt` line 124 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3491#-OkVhEhS5FwPOcdKHAPk:-OkVhEhS5FwPOcdKHAPl:bskd0yy) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/71ef6746968c59d944d2501b049dad769833d85f/src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/readers/ClientAccountReader.kt#L124)):*
> ```Kotlin
> 
>   /** Checks if a ClientAccount exists with the given internal ID for the MeasurementConsumer. */
>   suspend fun readByInternalId(
> ```

nit: You don't need to read the whole entity just to determine if it exists. Instead, add a method  to the companion object. just to check the existence of the row. You can use `AsyncDatabaseClient.ReadContext.readRow` to read the row using the primary key, passing any column since you won't be using the result.

You can do the same for the external ID using `readRowUsingIndex` specifying the `ClientAccountsByMeasurementConsumerAndExternalId` index.

PR #3504, opened 2026-01-26
feat: add client account key classes

getina:
adds client account key classes, which are used to parse and construct client account resource names

Issue: #3596

### DIFF
--- a/src/main/terraform/gcloud/modules/edp-aggregator/moved.tf
+++ b/src/main/terraform/gcloud/modules/edp-aggregator/moved.tf
@@ -15,6 +15,9 @@
 # Moved blocks for refactoring google_secret_manager_secret_iam_member resources
 # from edp-aggregator module into gcs-bucket-cloud-function and http-cloud-function modules.
 # See: https://developer.hashicorp.com/terraform/language/modules/develop/refactoring
+#
+# NOTE: Terraform moved blocks require static expressions - variable references are not
+# allowed. These values must match the secret_id values passed to the module.
 
 # =============================================================================
 # data_watcher_cloud_function (gcs-bucket-cloud-function module)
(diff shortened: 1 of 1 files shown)


SanjayVas (changes requested):
> @SanjayVas reviewed 22 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3504)* status: all files reviewed, 1 unresolved discussion (waiting on @getina and @stevenwarejones).

___
*[`src/main/kotlin/org/wfanet/measurement/api/v2alpha/DataProviderClientAccountKey.kt` line 28 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3504#-OkFWDuSBWJ3yK9Mxz7J:-OkFWDuSBWJ3yK9Mxz7K:b-gtb6c8) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/59466d70d249fccf50f51f9dbda7859743ea6123/src/main/kotlin/org/wfanet/measurement/api/v2alpha/DataProviderClientAccountKey.kt#L28)):*
> ```Kotlin
>   override val clientAccountId: String,
> ) : ClientAccountKey {
>   override val parentKey = DataProviderKey(dataProviderId)
> ```

Invert the handling of parentKey vs. dataProviderId:

1. Take in the parentKey in the primary constructor
2. Override the dataProviderId property to get the ID from parentKey
3. (Optional) Add a secondary constructor that takes in dataProviderId instead of parentKey

This way if the caller already has the parentKey instance you don't end up creating a new object for it.

SanjayVas (approved):
> @SanjayVas reviewed 9 files and all commit messages, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3504)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @stevenwarejones).

PR #2441, opened 2025-06-13
feat: Add CLI tools for manual load testing of EventGroup reads and writes

SanjayVas:
This was used for load testing the EventGroup search API using plaintext metadata, and is useful for other manual testing of EventGroups.

PR #3510, opened 2026-01-29
test: EGA performance test

roaminggypsy:


roaminggypsy (commented):
Kinda of stuck at this. Not completely ready for review but want to get some feedback on overall perf test set-up because I feel like I'm not doing this right

1) The choice of InProcessKingdom/InProcessReportingServer vs Common Server: Existing InProcessServer probably requires less set-up, but it doesn't expose port like Common Server, so I have to inject stubs to the load test tools, and I can't configure apiChannel with serviceConfig. Common Server would be slower because of network overhead? which will make it even harder to pass the test (I guess)

2) test cases: right now the first test case can pass locally, but for the second one, our target is 10k EventGroups, but even just for 100 EventGroups, it failed to create all the activities 

```Failed to create activities for dataProviders/D0Hq_QvuFTs/eventGroups/c0PF6vQR4Tw from 2024-01-01 to 2024-12-30: io.grpc.StatusException: DEADLINE_EXCEEDED: CallOptions deadline exceeded after 29.999996540s. Name resolution delay 0.000000000 seconds)```

> @roaminggypsy made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3510#-:-OkAu9lYDvNvT7VJo32K:bzgcvk3)* status: 0 of 14 files reviewed, all discussions resolved (waiting on @SanjayVas).

SanjayVas (changes requested):
> @SanjayVas reviewed 2 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3510)* status: 2 of 15 files reviewed, 1 unresolved discussion (waiting on @kungfucraig and @roaminggypsy).

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/tools/EventGroupActivitiesPerformanceTest.kt` line 41 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3510#-OkG6S800I91RmKVRaMR:-OkG6S800I91RmKVRaMS:b-5szebo) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c41b7a17612c0c909aaa8389144b4766149068a4/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/tools/EventGroupActivitiesPerformanceTest.kt#L41)):*
> ```Kotlin
>   subcommands = [CommandLine.HelpCommand::class],
> )
> class EventGroupActivitiesPerformanceTest : Runnable {
> ```

I don't think this wrapper is really that necessary. You should be able to do what you need by just updating the CLI tools from #2441 and calling them with the params you want. This way it's also easier to do the clean up to restore the environment.

SanjayVas (changes requested):
> @SanjayVas made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3510)* status: 2 of 21 files reviewed, 2 unresolved discussions (waiting on @kungfucraig and @roaminggypsy).

___
*[`src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/tools/EventGroupActivitiesPerformanceTest.kt` line 121 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3510#-OkG88fmAxYF5BI0EK12:-OkG88fmAxYF5BI0EK13:b538wko) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c41b7a17612c0c909aaa8389144b4766149068a4/src/main/kotlin/org/wfanet/measurement/loadtest/dataprovider/tools/EventGroupActivitiesPerformanceTest.kt#L121)):*
> ```Kotlin
>         CommandLine(WriteEventGroups())
>           .registerConverter(Duration::class.java) { it.toDuration() }
>           .execute(
> ```

Rather than having this extra infra, you can just write up an "@file" with the right args. See https://picocli.info/#AtFiles

Issue #3469, opened 2026-01-15
feat: Implement ImpressionMetadata Lifecycle Sync


georgi:
### Suggest an idea

Integrate the lifecycle logic into the main sync flow and create the cleanup function to handle deletions.

### Tasks
- [ ] **Sync Flow:** Update `DataAvailabilitySync.kt` to call `setObjectMetadata` after successful Spanner creation.
- [ ] **Cleanup Function:** Implement `DataAvailabilityCleanupFunction.kt` to handle `DeleteImpressionMetadata` gRPC calls.
- [ ] **Idempotency:** Ensure the cleanup function handles `NOT_FOUND` (if record already deleted).

### Decision Log
- We are using `x-goog-meta-impression-metadata-resource-id` to avoid a DB lookup during cleanup.

PR #3515, opened 2026-01-30
refactor: move secret access iam binding into cloud function modules

georgi:
This refactoring breaks a circular dependency in the Terraform configuration. Previously, the secret IAM bindings were defined in the parent module but needed to reference service accounts created by child modules, which created ordering issues. By moving the IAM bindings into the cloud function modules themselves:

1. The deploy provisioner now properly waits for IAM bindings to complete
2. Each module is self-contained with its own IAM requirements
3. The dependency graph is clearer and avoids circular references

### Changes
- Moves `google_secret_manager_secret_iam_member` resource from `edp-aggregator` module into `gcs-bucket-cloud-function` and `http-cloud-function` modules
- Adds `secrets_to_access` variable to cloud function modules to specify which secrets the service account needs access to
- Makes cloud function modules depend on `module.secrets` to ensure secrets are created before the IAM bindings

Issue: #3469 

BREAKING-CHANGE: The gcs-bucket-cloud-function and http-cloud-function modules now manage their own google_secret_manager_secret_iam_member resources via the new secrets_to_access variable. Callers of these modules outside of edp-aggregator must pass secrets_to_access to grant secret access. Existing edp-aggregator deployments must apply with moved.tf, adjusted for the actual names, before removing them to migrate state without resource recreation.

### DIFF
--- a/.github/workflows/publish-cloud-run-function-uber-jars.yml
+++ b/.github/workflows/publish-cloud-run-function-uber-jars.yml
@@ -118,3 +118,14 @@ jobs:
           artifact-group-id: org.wfanet.measurement.edpaggregator.deploy.gcloud.dataavailability
           bazel_target_label: //src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/dataavailability:DataAvailabilitySyncFunction_deploy.jar
 
+      - name: Build and Deploy Data Availability Cleanup
+        env:
+          MAVEN_USERNAME: ${{ github.actor }}
+          MAVEN_PASSWORD: ${{ secrets.GITHUB_TOKEN }}
+        uses: ./.github/actions/build-and-push-uber-jar
+        with:
+          package-version: ${{ steps.get-artifact-version.outputs.artifact-version }}
+          artifact-id: data-availability-cleanup
+          artifact-group-id: org.wfanet.measurement.edpaggregator.deploy.gcloud.dataavailability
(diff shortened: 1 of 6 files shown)


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3515)

georgi (commented):
> @georgi made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3515)* status: all files reviewed, 2 unresolved discussions (waiting on @SanjayVas).

___
*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3515#-OkFYG8v8COpcStrdRX2:-OkV6aDX16upk68oyBy2:b-896fix) (no related file):*

Add BREAKING-CHANGE trailers to explain breaking changes in any reusable Terraform modules, such as new required variables.
</blockquote></details>

Done.
___
*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3515#-OkFZCTw1n_wTR-LqVyg:-OkV6agz6wd-8J7vh3xi:b-896fix) (no related file):*

If this hasn't already been applied, make sure to add the appropriate `moved` blocks to prevent unnecessary resource creation/deletion. See https://developer.hashicorp.com/terraform/language/modules/develop/refactoring
</blockquote></details>

Done.

Issue #3468, opened 2026-01-15
feat: Extend StorageClient and DataWatcher for Object Metadata Support


georgi:
### Suggest an idea

Modify the internal storage interfaces and the DataWatcher trigger to handle `Custom-Time` and custom metadata headers.

### Tasks
- [ ] **StorageClient:** Add `setObjectMetadata(blobKey, customTime, metadata)` to the interface.
- [ ] **GcsStorageClient:** Implement using `BlobInfo.newBuilder()`.
- [ ] **DataWatcher:** Update `DataWatcherFunction.kt` to extract `data.metadataMap` from CloudEvents.
- [ ] **DataWatcher:** Update `sendToHttpEndpoint` to pass `X-Impression-Metadata-Resource-Id` header.

### Unit Tests
- [ ] Verify `customTime` is correctly formatted for GCS API.
- [ ] Verify `DataWatcher` successfully extracts custom metadata from `StorageObjectData`.

PR #3524, opened 2026-02-03
fix: terraform refactoring statements 

georgi:
Terraform moved blocks require static expressions, variable references are not allowed. 

Issue: #3468


georgi (commented):
> @georgi made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3524)* status: all files reviewed, 1 unresolved discussion (waiting on @SanjayVas).

___
*[`src/main/terraform/gcloud/modules/edp-aggregator/moved.tf` line 19 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3524#-OkZm4BjBNbY4lfJ-v9s:-Ok_K_dW22zGAMoZai6Y:b-d46ff0) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/b7d2b3155777d6e1dd8692accbf4f09204e9ada3/src/main/terraform/gcloud/modules/edp-aggregator/moved.tf#L19)):*

This is an instruction needed to be given to operators who use the module, no? In which case we can't have the moved blocks for them and instead need to update the BREAKING-CHANGE from the previous PR to explain that things were moved and that they need to do their own moved blocks.
</blockquote></details>

I updated the BREAKING-CHANGE to include that information.

Issue #3467, opened 2026-01-15
Update the Terraform storage module to support object lifecycle


georgi:
### Suggest an idea

Update the Terraform storage module to support retention based on `Custom-Time` and deploy the necessary Cloud Functions for event-driven cleanup.

### Acceptance Criteria
- [ ] Update `src/main/terraform/gcloud/modules/storage-bucket/main.tf` with `lifecycle_rule`.
- [ ] Implement `days_since_custom_time` condition.
- [ ] Deploy second `DataWatcher` instance for `google.cloud.storage.object.v1.deleted` events.
- [ ] Deploy `DataAvailabilityCleanupFunction` HTTP endpoint.

### Technical Notes
Add an age-based fallback (90 days from creation) for objects missing `Custom-Time`.

PR #3525, opened 2026-02-03
fix: define trigger event type for bucket cloud function

georgi:
Data watcher and Data watcher delete functions need to react to different event trigger types.

Issue: #3467 

BREAKING-CHANGE: The gcs-bucket-cloud-function module now requires the trigger_event_type variable to be explicitly set. Valid values are "finalized" or "deleted"


SanjayVas (approved):
> @SanjayVas reviewed 3 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3525)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @georgi).

PR #3545, opened 2026-02-05
fix: always compute data availability for all active model lines

stevenwarejones:
The current DataAvailabilitySync only syncs correctly if the done blob is written for all model lines. This fix updates the DataAvailabilitySync to work in scenarios where the done blob is written per model line.

Additionally, this fixes an issue where delete blobs were included in availability.

Issue: #3551


stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3545)* status: 6 of 12 files reviewed, all discussions resolved (waiting on @kungfucraig and @Marco-Premier).

PR #3560, opened 2026-02-12
fix: Ensure thread safety of SetExpressionCompiler caches

### LINKED ISSUE #3555: Intermittent failure of HMSS reach case in GCloudInProcessLifeOfAReportV2IntegrationTest
The `report with HMSS union reach across 2 edps has the expected result` method of `//src/test/kotlin/org/wfanet/measurement/integration/deploy/gcloud:GCloudInProcessLifeOfAReportV2IntegrationTest` sometimes fails with the actual reach value being outside of the tolerance of the expected reach value. Sometimes it's only a small amount, but the more likely failure case is that the actual reach is ~50% of the expected value.

For example:

```
value of:
    measurementResult.reach.value.getValue
expected:
    4001.0
but was:
    1997.0
outside tolerance:
    24.695726670410156
measurementResult was:
    reach {
      value: 1997
    }
```

This appears to have started occurring after #3459, which changed the test method in question to use synthetic `edp2` rather than `edp3`. `edp3` has a significantly larger number of impressions than `edp2`. The failure is not observed when the test is changed back to use `edp3`. This should not affect whether the test passes, and instead may be surfacing a pre-existing bug.

In the cases where the reach is half the expected value, there's a primitive ReportingSet basis missing from the composite ReportingSet:

```diff
       primitive_reporting_set_bases {
-        external_reporting_set_id: "abc1"
-        filters: "person.age_group == 3"
-      }
-      primitive_reporting_set_bases {
         external_reporting_set_id: "abc0"
         filters: "person.age_group == 1"
       }
```

```diff
       details {
         results {
           reach {
-            value: 3966
-            noise_mechanism: DISCRETE_GAUSSIAN
-            honest_majority_share_shuffle {
-              frequency_vector_size: 99999
+            value: 1997
+            noise_mechanism: CONTINUOUS_GAUSSIAN
+            deterministic_count_distinct {
             }
           }
         }
-        data_provider_count: 2
+        data_provider_count: 1
       }
     }
   }
```

[half-expected-reach.log](https://github.com/user-attachments/files/25195385/half-expected-reach.log)

[outside-tolerance.log](https://github.com/user-attachments/files/25195387/outside-tolerance.log)

SanjayVas:
In addition to using safe operations for cache updates, this removes the attempted parallelism in set expression compilation which is likely what caused the observed concurrent writes. This also improves test failure messages for GCloudInProcessLifeOfAReportV2IntegrationTest.

Closes #3555

Issue: #3555

### DIFF
--- a/src/main/kotlin/org/wfanet/measurement/integration/common/reporting/v2/InProcessLifeOfAReportIntegrationTest.kt
+++ b/src/main/kotlin/org/wfanet/measurement/integration/common/reporting/v2/InProcessLifeOfAReportIntegrationTest.kt
@@ -17,6 +17,7 @@
 package org.wfanet.measurement.integration.common.reporting.v2
 
 import com.google.common.truth.Truth.assertThat
+import com.google.common.truth.Truth.assertWithMessage
 import com.google.common.truth.extensions.proto.ProtoTruth.assertThat
 import com.google.protobuf.timestamp
 import com.google.type.DayOfWeek
@@ -63,6 +64,7 @@ import org.wfanet.measurement.api.v2alpha.Measurement
 import org.wfanet.measurement.api.v2alpha.MeasurementConsumerKey
 import org.wfanet.measurement.api.v2alpha.MeasurementConsumersGrpcKt.MeasurementConsumersCoroutineStub
 import org.wfanet.measurement.api.v2alpha.MeasurementKt
+import org.wfanet.measurement.api.v2alpha.MeasurementSpec
 import org.wfanet.measurement.api.v2alpha.MeasurementsGrpcKt.MeasurementsCoroutineStub
 import org.wfanet.measurement.api.v2alpha.RequisitionSpecKt
 import org.wfanet.measurement.api.v2alpha.eventGroup as cmmsEventGroup
@@ -72,6 +74,7 @@ import org.wfanet.measurement.api.v2alpha.getDataProviderRequest
 import org.wfanet.measurement.api.v2alpha.getMeasurementConsumerRequest
 import org.wfanet.measurement.api.v2alpha.listMeasurementsRequest
 import org.wfanet.measurement.api.v2alpha.testing.MeasurementResultSubject.Companion.assertThat
+import org.wfanet.measurement.api.v2alpha.unpack
 import org.wfanet.measurement.api.withAuthenticationKey
 import org.wfanet.measurement.common.OpenEndTimeRange
 import org.wfanet.measurement.common.base64UrlEncode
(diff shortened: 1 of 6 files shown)


SanjayVas (commented):
> @SanjayVas made 2 comments and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3560)* status: 3 of 4 files reviewed, all discussions resolved (waiting on stevenwarejones).

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/SetExpressionCompiler.kt` line 91 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3560#-OlNSPhmEYRlJEE7UGuw:-Oli1edl4bRuCMRzR3Gl:b-g9yy0b) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/baedb8034c87e7a253d7f0cc1fc30718d11ad5cb/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/SetExpressionCompiler.kt#L91)):*

Is there a case where MutablePrimitiveRegionToUnionSetCoefficientMap also needs to be a thread safe map?
</blockquote></details>

Done. Switched it to also be a ConcurrentMap just in case.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/SetExpressionCompiler.kt` line 98 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3560#-OlNTKoA8WekyI_-DRnh:-Oli0nFf8o4LIMmKBbIp:btfp0hl) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/baedb8034c87e7a253d7f0cc1fc30718d11ad5cb/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/SetExpressionCompiler.kt#L98)):*

ditto on VennDiagramDecomposition? does it need to be thread-safe --- something like CopyOnWriteArrayList<thread-safe set>
</blockquote></details>

VennDiagramDecomposition is unmodifiable. Values of that type in the cache are effectively immutable due to only being set to new instances with no modifiable references kept.

SanjayVas (commented):
> @SanjayVas reviewed 3 files.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3560)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on stevenwarejones).

PR #3570, opened 2026-02-18
feat: Create RawImpressionBatchFile Table and Proto Files

jojijac0b:
Create a new Spanner table RawImpressionBatchFile to track raw impression files grouped into batches for VID labeling.

Issue: #3584

### DIFF
--- a/src/main/kotlin/org/wfanet/measurement/edpaggregator/eventgroups/EventGroupSync.kt
+++ b/src/main/kotlin/org/wfanet/measurement/edpaggregator/eventgroups/EventGroupSync.kt
@@ -75,10 +75,11 @@ data class EventGroupKey(val eventGroupReferenceId: String, val measurementConsu
  * Syncs event groups with the CMMS Public API.
  * 1. **Creates** any EventGroups that exist in the input flow but not in CMMS.
  * 2. **Updates** existing EventGroups in CMMS if their data has changed.
- * 3. **Deletes** EventGroups that exist in CMMS but are no longer present
- *    in the input [eventGroups] flow.
+ * 3. **Deletes** EventGroups whose state is [EventGroup.State.DELETED] from CMMS.
  * 4. **Returns** a flow of [MappedEventGroup], one element for each successfully
- *    created or updated EventGroup.
+ *    created or updated EventGroup (deleted EventGroups are not included).
+ *
+ * EventGroups that exist in CMMS but are absent from the input flow are left as-is.
  */
 class EventGroupSync(
   private val edpName: String,
@@ -109,8 +110,11 @@ class EventGroupSync(
    * For EventGroups with `client_account_reference_id`, resolves all associated
    * MeasurementConsumers and creates separate EventGroup resources for each.
    *
+   * EventGroups with [EventGroup.State.DELETED] are deleted from CMMS if they exist. EventGroups
+   * absent from the input flow are left unchanged.
+   *
    * @return Flow of [MappedEventGroup] for each successfully synced EventGroup. Failed syncs are
-   *   skipped and logged.
+   *   skipped and logged. Deleted EventGroups are not included.
    */
   suspend fun sync(): Flow<MappedEventGroup> = flow {
     withSpan(
@@ -131,10 +135,28 @@ class EventGroupSync(
 
       val edpEventGroupsList = eventGroups.toList()
 
-      // Track keys of successfully resolved EventGroups to prevent deletion
-      val resolvedEventGroupKeys = mutableSetOf<EventGroupKey>()
-
       for (eventGroup in edpEventGroupsList) {
+        if (eventGroup.state == EventGroup.State.DELETED) {
+          try {
+            validateDeletedEventGroup(eventGroup)
+          } catch (e: Exception) {
+            logger.log(Level.SEVERE, e) {
+              "Skipping deleted Event Group ${eventGroup.eventGroupReferenceId}: " +
+                "Validation failed"
+            }
+            metrics.invalidEventGroupFailure.add(1, metricAttributes())
+            continue
+          }
+
+          val key = EventGroupKey(eventGroup.eventGroupReferenceId, eventGroup.measurementConsumer)
+          val cmmsEventGroup = cmmsEventGroups[key]
+          if (cmmsEventGroup != null) {
+            deleteCmmsEventGroup(cmmsEventGroup)
+            metrics.syncDeleted.add(1, metricAttributes())
+          }
+          continue
+        }
+
         try {
           validateEventGroup(eventGroup)
         } catch (e: Exception) {
@@ -148,24 +170,11 @@ class EventGroupSync(
         val measurementConsumerKeys: Set<MeasurementConsumerKey> =
           resolveMeasurementConsumers(eventGroup)
 
-        // Process event group for each measurement consumer
         for (measurementConsumerKey in measurementConsumerKeys) {
           val eventGroupWithConsumer =
             eventGroup.copy { measurementConsumer = measurementConsumerKey.toName() }
-          resolvedEventGroupKeys.add(
-            EventGroupKey(
-              eventGroupWithConsumer.eventGroupReferenceId,
-              eventGroupWithConsumer.measurementConsumer,
-            )
-          )
           syncEventGroupItem(eventGroupWithConsumer, cmmsEventGroups)?.let { emit(it) }
         }
-      }
-
-      val keysToDelete = cmmsEventGroups.keys - resolvedEventGroupKeys
-      for (key in keysToDelete) {
-        val eventGroup = cmmsEventGroups.getValue(key)
-        deleteCmmsEventGroup(eventGroup)
       }
     }
   }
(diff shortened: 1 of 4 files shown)


Marco-Premier (changes requested):
> @Marco-Premier reviewed 5 files and all commit messages, and made 3 comments.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3570)* status: all files reviewed, 3 unresolved discussions (waiting on jojijac0b and stevenwarejones).

___
*[`src/main/resources/edpaggregator/spanner/create-raw-impression-batch-file-schema.sql` line 32 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3570#-OluZObN1RGY3GOZWizJ:-OluZObN1RGY3GOZWizK:b-64uvvi) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/21fd8125f8301c61eeec7d17eb8eef49669fc77d/src/main/resources/edpaggregator/spanner/create-raw-impression-batch-file-schema.sql#L32)):*
> ```TSQL
> 
> -- Tracks raw impression files grouped into batches for VID labeling.
> -- Hierarchy: DataProvider -> Upload -> Batch -> File
> ```

Since raw impressions are partition by date I think we should add a new field: 

```
-- The date this upload covers.
  UploadDate DATE NOT NULL
```

and add an index:

```
-- Supports: "Find the ACTIVE upload for a specific date"
-- Query: WHERE DataProviderResourceId = @dp AND UploadDate = @date AND State = 'ACTIVE'
CREATE INDEX RawImpressionBatchFileByUploadDateAndState
  ON RawImpressionBatchFile(DataProviderResourceId, UploadDate, State);
```
___
*[`src/main/resources/edpaggregator/spanner/create-raw-impression-batch-file-schema.sql` line 33 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3570#-OluYgp53dsx-_w7XUTj:-OluYgp53dsx-_w7XUTk:b-l0p6ea) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/21fd8125f8301c61eeec7d17eb8eef49669fc77d/src/main/resources/edpaggregator/spanner/create-raw-impression-batch-file-schema.sql#L33)):*
> ```TSQL
> -- Tracks raw impression files grouped into batches for VID labeling.
> -- Hierarchy: DataProvider -> Upload -> Batch -> File
> CREATE TABLE RawImpressionBatchFile (
> ```

I actually think that `RawImpressionMetadata` would be better naming and align with the existing `ImpressionMetadata` table



_Code quote:_
```TSQL
CREATE TABLE RawImpressionBatchFile (
```
___
*[`src/main/resources/edpaggregator/spanner/create-raw-impression-batch-file-schema.sql` line 41 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3570#-OluYHcE3mAZzddyCWUV:-OluYHcE3mAZzddyCWUW:b-rybxrl) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/21fd8125f8301c61eeec7d17eb8eef49669fc77d/src/main/resources/edpaggregator/spanner/create-raw-impression-batch-file-schema.sql#L41)):*
> ```TSQL
>   BatchIndex INT64 NOT NULL,
>   -- Zero-based index of the file within the batch.
>   FileIndex INT64 NOT NULL,
> ```

second thought here, don't think this is needed as each file is already identified by the BlobUri which has a unique index already. Remove it also from the `PRIMARY KEY`



_Code quote:_
```TSQL
  -- Zero-based index of the file within the batch.
  FileIndex INT64 NOT NULL,
```

SanjayVas (commented):
What would the resource look like in the system API? (I generally design these top-down, defining the highest-level API first). It's sounding like a `RawImpressionMetadata` resource parented by DataProvider? I see that there's two other components to the primary key though, so I'm not certain what the resource ID would be. Maybe you have something in-between to represent the batch so the pattern would be something like `dataProviders/{data_provider}/rawImpressionBatches/{batch}/metadata/{batch_index}`?

> what you think about this denormalized table?

Which part specifically is denormalized? Is it that there's not an intermediate table for the batch? That would be fine.

> @SanjayVas reviewed all commit messages and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3570#-:-OlwPTOc6b6OlUrMvB48:bjibnwz)* status: 2 of 5 files reviewed, 3 unresolved discussions (waiting on Marco-Premier and stevenwarejones).

jojijac0b:
Hey @SanjayVas. The resource would look like: dataProviders/{data_provider}/rawImpressionUploads/{raw_impression_upload}/batches/{batch}/files/{file}. So if this table was normalized, the intermediate tables would be: Upload table / Batch table / File table

SanjayVas (commented):
I'm just trying to figure out what the resource IDs are in this case. It sounds like `UploadId` is the resource ID for the `RawImpressionUpload` resource. I'm not sure from your description what the resource IDs are for the batch and the file. I'd expect this table to have one row per leaf resource.

> > Which column corresponds to the file ID? Is it the upload date?
>
> No, it's the `BlobUri` which is unique.

The blob URI is too long for a resource ID, which usually should conform to RFC-1034. See https://google.aip.dev/122.

We usually have separate columns for the DB primary key vs. the API resource IDs. Notice how RequisitionMetadata has both `RequisitionMetadataId INT64` and `RequisitionMetadataResourceId STRING(63)`. Therefore I think this needs separate columns for the STRING(63) upload resource ID vs. the internal INT64 upload ID. Additionally you'll want a separate resource ID for the file, where you may have a separate method for looking up a raw impression file resource via blob URI.

> @SanjayVas made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3570#-:-OmBqaRD0bnjyJ7amYeP:bt0b21t)* status: 2 of 5 files reviewed, 3 unresolved discussions (waiting on Marco-Premier and stevenwarejones).

stevenwarejones (approved):
> @stevenwarejones partially reviewed 2 files and all commit messages, and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3570)* status: all files reviewed, 3 unresolved discussions (waiting on Marco-Premier).

Marco-Premier (approved):
LGTM

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3570#-:-OmQpvki2K5_MfXCoc7Q:bb6ywmx)* status: all files reviewed, 3 unresolved discussions (waiting on jojijac0b).

SanjayVas (changes requested):
> @SanjayVas reviewed 4 files and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3570)* status: all files reviewed, 2 unresolved discussions (waiting on jojijac0b).

___
*[`src/main/resources/edpaggregator/spanner/create-raw-impression-metadata-schema.sql` line 45 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3570#-OmRRrKEAxx7DsrNX781:-OmRRrKEAxx7DsrNX782:bo2eeax) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/82aeb7dc17ba1aeeee07433e1d10099b2b0364ad/src/main/resources/edpaggregator/spanner/create-raw-impression-metadata-schema.sql#L45)):*
> ```TSQL
>   -- Corresponds to {batch} in the resource pattern.
>   BatchIndex INT64 NOT NULL,
>   -- Internal DB key for this file, unique per DataProvider.
> ```

Why do we need FileId and FileResourceId to be unique per DataProvider? Generally each ID segment of a resource ID need only be unique within the context of its parent. Therefore I'd expect this just to be unique for a given batch.



_Code quote:_
```TSQL
unique per DataProvider.
```
___
*[`src/main/resources/edpaggregator/spanner/create-raw-impression-metadata-schema.sql` line 51 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3570#-OmROW7_0P3bG7b48YDm:-OmROW7_0P3bG7b48YDn:b-7obfi6) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/82aeb7dc17ba1aeeee07433e1d10099b2b0364ad/src/main/resources/edpaggregator/spanner/create-raw-impression-metadata-schema.sql#L51)):*
> ```TSQL
>   -- Indexed via RawImpressionMetadataByFileResourceId for resource-ID-based lookups.
>   FileResourceId STRING(63) NOT NULL,
>   -- The date this upload covers. Has a 1:1 relationship with UploadId/UploadResourceId
> ```

If this has a 1:1 relationship then you don't need this many columns.

Unless you have concerns about hot-spotting, you can use this in the primary key. If you intend to expose the date, you can also use it as the resource ID (you'd use the ISO-8601 string format in the public API).

PR #3574, opened 2026-02-19
feat: Define VID labeling config and params protos for dispatcher architecture

jojijac0b:
Define the proto messages for the VID Labeling Dispatcher architecture:

- `VidLabelingConfig` (config): Static per-EDP configuration loaded by the Cloud Function dispatcher at startup, with a `VidLabelingConfigs` wrapper for multi-EDP support.
- `VidLabelingDispatcherParams` (v1alpha): Slim message sent by the DataWatcher to the Cloud Function, containing only the `data_provider` used to select the correct config.
- `VidLabelerParams` (v1alpha): TEE application parameters built by the Cloud Function from config + per-invocation data, passed as an `Any` to the Secure Computation API.

Issue: #3588


jojijac0b (commented):
Done — restored all existing field numbers (fields 1-7 match main). New dispatcher-specific fields use 8+.

jojijac0b:
*[`src/main/proto/wfa/measurement/edpaggregator/v1alpha/vid_labeler_params.proto` line 70 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3574#-OmG_c5oDs7ooz_m5hRu:-OmpAE0kEUKxGZtyfeOw:b9h683s) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/b3cf825da67ec2a4b9e22404c847a3ddb44eb4fc/src/main/proto/wfa/measurement/edpaggregator/v1alpha/vid_labeler_params.proto#L70)):*

isn't it just one set of TransportLayerSecurityParams? Its messy to add a proto field we intend to immediately deprecate.
</blockquote></details>

@stevenwarejones  is this comment still relevant after the decision not to merge TransportLayerSecurityParams?

jojijac0b:
*[`src/main/proto/wfa/measurement/config/edpaggregator/vid_labeling_dispatcher_config.proto` line 42 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3574#-Olung-F-X7F5eS_R7X8:-Omp2sIF-EV4TQm-X9zW:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/b6ed3ccb16eab77e2daadd0753b36653c9fe52b6/src/main/proto/wfa/measurement/config/edpaggregator/vid_labeling_dispatcher_config.proto#L42)):*

I agree. Sanjay's question got me thinking if we even need this proto. A lot is duplicated across these two messages. I suggest combining. Its an implementation detail that the work is split between a cloud function and MIGs. It could be done a different way in the future.
</blockquote></details>

Done.

SanjayVas (approved):
> @SanjayVas reviewed 3 files and all commit messages, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3574)* status: all files reviewed, 1 unresolved discussion (waiting on Marco-Premier and stevenwarejones).

PR #3575, opened 2026-02-21
feat: Add CLI tool to detect Reports with invalid ReportingSets

SanjayVas:
This detects the usage of ReportingSets that are invalid due to the underlying cause of #3555


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3575)

PR #3572, opened 2026-02-19
feat: Create VID labeler params proto

jojijac0b:
Define the VidLabelerParams proto message used to pass parameters to the VID Labeling TEE application.

Issue: #3588

PR #3589, opened 2026-02-25
fix: Merge transport layer security params protos

jojijac0b:
This PR merges the two existing TransportLayerSecurityParams protos—one file system-based (used by Cloud Functions) and one resource-based (used by TEE applications)—into a single, unified proto. This consolidation was identified while developing the VID Labeler params and VID Dispatcher config, which we decided to merge into a single proto in #3574. This PR is a prerequisite for that work, as it establishes a cleaner pattern that simplifies the design change.

### DIFF
--- a/src/main/kotlin/org/wfanet/measurement/edpaggregator/eventgroups/EventGroupSync.kt
+++ b/src/main/kotlin/org/wfanet/measurement/edpaggregator/eventgroups/EventGroupSync.kt
@@ -75,10 +75,11 @@ data class EventGroupKey(val eventGroupReferenceId: String, val measurementConsu
  * Syncs event groups with the CMMS Public API.
  * 1. **Creates** any EventGroups that exist in the input flow but not in CMMS.
  * 2. **Updates** existing EventGroups in CMMS if their data has changed.
- * 3. **Deletes** EventGroups that exist in CMMS but are no longer present
- *    in the input [eventGroups] flow.
+ * 3. **Deletes** EventGroups whose state is [EventGroup.State.DELETED] from CMMS.
  * 4. **Returns** a flow of [MappedEventGroup], one element for each successfully
- *    created or updated EventGroup.
+ *    created or updated EventGroup (deleted EventGroups are not included).
+ *
+ * EventGroups that exist in CMMS but are absent from the input flow are left as-is.
  */
 class EventGroupSync(
   private val edpName: String,
@@ -109,8 +110,11 @@ class EventGroupSync(
    * For EventGroups with `client_account_reference_id`, resolves all associated
    * MeasurementConsumers and creates separate EventGroup resources for each.
    *
+   * EventGroups with [EventGroup.State.DELETED] are deleted from CMMS if they exist. EventGroups
+   * absent from the input flow are left unchanged.
+   *
    * @return Flow of [MappedEventGroup] for each successfully synced EventGroup. Failed syncs are
-   *   skipped and logged.
+   *   skipped and logged. Deleted EventGroups are not included.
    */
   suspend fun sync(): Flow<MappedEventGroup> = flow {
     withSpan(
@@ -131,10 +135,28 @@ class EventGroupSync(
 
       val edpEventGroupsList = eventGroups.toList()
 
-      // Track keys of successfully resolved EventGroups to prevent deletion
-      val resolvedEventGroupKeys = mutableSetOf<EventGroupKey>()
-
       for (eventGroup in edpEventGroupsList) {
+        if (eventGroup.state == EventGroup.State.DELETED) {
+          try {
+            validateDeletedEventGroup(eventGroup)
+          } catch (e: Exception) {
+            logger.log(Level.SEVERE, e) {
+              "Skipping deleted Event Group ${eventGroup.eventGroupReferenceId}: " +
+                "Validation failed"
+            }
+            metrics.invalidEventGroupFailure.add(1, metricAttributes())
+            continue
+          }
+
+          val key = EventGroupKey(eventGroup.eventGroupReferenceId, eventGroup.measurementConsumer)
+          val cmmsEventGroup = cmmsEventGroups[key]
+          if (cmmsEventGroup != null) {
+            deleteCmmsEventGroup(cmmsEventGroup)
+            metrics.syncDeleted.add(1, metricAttributes())
+          }
+          continue
+        }
+
         try {
           validateEventGroup(eventGroup)
         } catch (e: Exception) {
@@ -148,24 +170,11 @@ class EventGroupSync(
         val measurementConsumerKeys: Set<MeasurementConsumerKey> =
           resolveMeasurementConsumers(eventGroup)
 
-        // Process event group for each measurement consumer
         for (measurementConsumerKey in measurementConsumerKeys) {
           val eventGroupWithConsumer =
             eventGroup.copy { measurementConsumer = measurementConsumerKey.toName() }
-          resolvedEventGroupKeys.add(
-            EventGroupKey(
-              eventGroupWithConsumer.eventGroupReferenceId,
-              eventGroupWithConsumer.measurementConsumer,
-            )
-          )
           syncEventGroupItem(eventGroupWithConsumer, cmmsEventGroups)?.let { emit(it) }
         }
-      }
-
-      val keysToDelete = cmmsEventGroups.keys - resolvedEventGroupKeys
-      for (key in keysToDelete) {
-        val eventGroup = cmmsEventGroups.getValue(key)
-        deleteCmmsEventGroup(eventGroup)
       }
     }
   }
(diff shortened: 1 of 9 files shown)


stevenwarejones (changes requested):
> @stevenwarejones reviewed 16 files and all commit messages, and made 4 comments.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3589)* status: all files reviewed, 4 unresolved discussions (waiting on jojijac0b and Marco-Premier).

___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/requisitionfetcher/RequisitionFetcherFunction.kt` line 338 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3589#-OmKWcsV7hNpxCHnaEK6:-OmKWcsV7hNpxCHnaEK7:b-r7yvwh) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/2e6d5f2de4bcbad6d10f356da14dbd6990f726ef/src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/requisitionfetcher/RequisitionFetcherFunction.kt#L338)):*
> ```Kotlin
>       return SigningCerts.fromPemFiles(
>         certificateFile =
>           checkNotNull(File(transportLayerSecurityParams.cloudParams.certFilePath)),
> ```

this needs to be backwards compatible if the old field is present. It should also check if `hasCloudParams` otherwise - check the old value
___
*[`src/main/proto/wfa/measurement/config/edpaggregator/data_availability_sync_config.proto` line 39 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3589#-OmKWqAXEduvrCvQpXZq:-OmKWqAXEduvrCvQpXZr:bnrzj25) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/2e6d5f2de4bcbad6d10f356da14dbd6990f726ef/src/main/proto/wfa/measurement/config/edpaggregator/data_availability_sync_config.proto#L39)):*
> ```proto
> 
>   // How the DataAvailabilitySync connects to the CMMS Public API
>   wfa.measurement.edpaggregator.v1alpha.TransportLayerSecurityParams
> ```

preserve the old field and mark as deprecated. create a new field
___
*[`src/main/proto/wfa/measurement/edpaggregator/v1alpha/transport_layer_security_params.proto` line 30 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3589#-OmKUD-R70MYH3tFzHIO:-OmKUD-R70MYH3tFzHIP:b-feez6f) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/2e6d5f2de4bcbad6d10f356da14dbd6990f726ef/src/main/proto/wfa/measurement/edpaggregator/v1alpha/transport_layer_security_params.proto#L30)):*
> ```proto
> message TransportLayerSecurityParams {
>   // TLS params for cloud functions using filesystem-mounted certificates.
>   message CloudParams {
> ```

nit: `FileSystemParams`
___
*[`src/main/proto/wfa/measurement/edpaggregator/v1alpha/transport_layer_security_params.proto` line 40 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3589#-OmKUHuj8b2HLHRVU7SW:-OmKUHuj8b2HLHRVU7SX:bacy72e) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/2e6d5f2de4bcbad6d10f356da14dbd6990f726ef/src/main/proto/wfa/measurement/edpaggregator/v1alpha/transport_layer_security_params.proto#L40)):*
> ```proto
> 
>   // TLS params for TEE applications using resource key paths.
>   message TeeParams {
> ```

nit: `ResourceParams`

jojijac0b:
*[`src/main/proto/wfa/measurement/edpaggregator/v1alpha/results_fulfiller_params.proto` line 83 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3589#-OmQfkPTBrHoFVVU_ki3:-OmV0vH10Hr-IfUR-zNw:b-d61pul) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f3ab0c8cf6648a5f5c78c212eee8b84a64e40b7f/src/main/proto/wfa/measurement/edpaggregator/v1alpha/results_fulfiller_params.proto#L83)):*

Done.
</blockquote></details>

Reverted back to the structure that was in place before https://github.com/world-federation-of-advertisers/cross-media-measurement/pull/3572

jojijac0b (commented):
> @jojijac0b made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3589)* status: 21 of 22 files reviewed, 7 unresolved discussions (waiting on Marco-Premier and stevenwarejones).

___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/requisitionfetcher/RequisitionFetcherFunction.kt` line 433 at r9](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3589#-OmVNQDi2bgHPU8I5W1I:-OmVdzXz39XqT-ZhLkkk:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ace5589c594efbd69ba64b8ad80eb5f697a768ce/src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/requisitionfetcher/RequisitionFetcherFunction.kt#L433)):*

just throw an exception
</blockquote></details>

Done.

stevenwarejones (commented):
> @stevenwarejones reviewed 6 files and all commit messages, and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3589)* status: all files reviewed, 5 unresolved discussions (waiting on Marco-Premier).

jojijac0b (commented):
> @jojijac0b made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3589)* status: all files reviewed, 7 unresolved discussions (waiting on Marco-Premier, SanjayVas, and stevenwarejones).

___
*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3589#-OmkgdSs9K9yIpLZjr4j:-OmlUqWc3qyCt0v7mq1L:bgw5zfv) (no related file):*

To clarify what I understood from the discussion: What it sounded like was that there were components that needed to be versioned eventually or updated to clearly handle multiple versions. That can be handled in a future PR. What cannot be violated in this PR is using API messages where they shouldn't be used, e.g. in config messages. That is to say that the config itself should not be versioned.
</blockquote></details>

@stevenwarejones Agreed. I'm working on this in a separate PR based on jojijacob-create-dispatcher-proto. The new PR creates FileSystemTransportLayerSecurityParams, DataAvailabilitySyncParams, and EventGroupSyncParams in v1alpha/, and updates the Cloud Function Kotlin code to accept both old and new formats with backwards-compatible parsing. The "unified" approach from this PR will not carry over.

@SanjayVas Understood — I'll handle the proper versioning of wire params in a separate PR. The new approach creates separate v1alpha messages (DataAvailabilitySyncParams, EventGroupSyncParams) for what goes over the wire, with a FileSystemTransportLayerSecurityParams (not "unified") for the TLS use case. The config messages will remain independent from the versioned API. The Cloud Functions will use a converter from the v1alpha wire params to the internal config format, following your recommendation of option 1.

Versioning changes can be found here: https://github.com/world-federation-of-advertisers/cross-media-measurement/pull/3622.
___
*[`src/main/proto/wfa/measurement/config/edpaggregator/data_availability_sync_config.proto` line 64 at r12](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3589#-Oml7WArEt3ux5kqACXt:-OmlUizX-3s3Dipc8a_b:bxi55td) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7bc86d759d2ee087af6ec93780fcab8ca8abfdb4/src/main/proto/wfa/measurement/config/edpaggregator/data_availability_sync_config.proto#L64)):*

This config cannot be tied to a specific API version. Assuming this is trying to configure a component which will itself send the API params, there are the following options in order of preference:

1. Use a separate message in the config and convert to the appropriate versioned API message.
2. Use an Any field. This may potentially need to be repeated if the component that uses this config may handle multiple API versions.
3. Have a separate field for each API version. If the component that uses this config will itself be versioned, this can be a oneof. If it instead will handle multiple API versions then you may need to have multiple set and cannot use a oneof.
</blockquote></details>

Acknowledged. The new PR follows option 1: the Cloud Functions receive the v1alpha wire params, convert them to the internal config format, and all downstream code operates on the config type. The config messages won't reference any v1alpha types directly.

PR #3599, opened 2026-02-26
Add VerifySyntheticData tool for validating encrypted synthetic data

stevenwarejones:
Decrypts and parses generated impressions using AWS/GCP/Fake KMS to verify end-to-end envelope encryption integrity.

Issue: #3615


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3599)

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3599)* status: 3 of 6 files reviewed, 1 unresolved discussion (waiting on kungfucraig, Marco-Premier, and SanjayVas).

___
*[`src/test/kotlin/org/wfanet/measurement/loadtest/edpaggregator/tools/GenerateSyntheticDataTest.kt` line 36 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3599#-OmVWRdpCVlSbuEmwZ3I:-OmVrjRU57l5EtHUjybb:b-l05c7s) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/679c4d6027f7536650083d4d8bca2e64e1893d22/src/test/kotlin/org/wfanet/measurement/loadtest/edpaggregator/tools/GenerateSyntheticDataTest.kt#L36)):*

I do disagree. The parsing is handled by Picocli which we don't need to exercise and should be able to assume just works as documented. The only tests we should have are ones which call the entry point (`main` in this case).
</blockquote></details>

removed the tests since in that case there isn't much to test

PR #3600, opened 2026-02-26
edpa should not delete event groups

stevenwarejones:
Currently, the EDPA deletes event groups not re-uploaded by the EDP. The intended behavior was that these event groups be marked as INACTIVE. This PR does not introduce an INACTIVE state but, instead, removes the incorrect deletion.


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3600)

PR #3613, opened 2026-03-01
 fix: add TrusTee protocol support in Reporting Server proto conversion

### LINKED ISSUE #3612: Reporting Server fails to convert TrusTee measurement results with UNKNOWN gRPC error
### Bug description

 When a TrusTee measurement completes successfully in the duchy, the Reporting Server throws an UNKNOWN gRPC error when attempting to persist the result. The measurement itself succeeds end-to-end, but the Reporting
  Server's proto conversion layer has no handler for the TrusTee protocol, causing it to crash during batchSetInternalMeasurementResults.

### Steps To Reproduce

1. Deploy a Duchy TEE
2. Create a Report
3. Fulfill it with EDPs
4. Let the Duchy, Kingdom, RS complete the workflows
5. Call ListReport or GetReport from the RS

### Component(s) affected

Reporting Server

### Version

0.5.34

### Environment

0.5.34 Self Deployement

### Additional context

Root Cause

  ProtoConversions.kt in the Reporting Server (reporting/service/api/v2alpha/) contains two functions — Measurement.Result.Reach.toInternal() and Measurement.Result.Frequency.toInternal() — that convert public
  Measurement results to internal protos. Both functions handle Direct, LiquidLegionsV2, ReachOnlyLiquidLegionsV2, and HonestMajorityShareShuffle protocols via else if branches, but neither has a branch for TrusTee. Any
   measurement using TrusTee falls through to the else { error(...) } clause.

  ---
  Affected Flow

  1. TrusTee duchy computation completes successfully
  2. Kingdom stores and serves the result correctly (Kingdom's own ProtoConversions.kt handles TrusTee)
  3. Reporting Server receives the result, decrypts it successfully
  4. Reporting Server calls source.reach.toInternal(protocolConfig) — crashes here
  5. Exception is caught and re-thrown as Status.UNKNOWN to the caller

ugur47:
The Reporting Server's ProtoConversions.kt was missing a hasTrusTee() branch in both Measurement.Result.Reach.toInternal() and Measurement.Result.Frequency.toInternal(). This caused any successfully completed TrusTee
  measurement to fail with Status.UNKNOWN: Measurement protocol is not set or not supported when the Reporting Server attempted to persist the result — even though the duchy computation and Kingdom serving were both
  correct.

  This PR adds the missing branch in both conversion functions and covers the fix with unit tests for reach-only and reach-and-frequency TrusTee metrics in MetricsServiceTest.

fixes #3612

Issue: #3612

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -39,7 +39,7 @@ bazel_dep(
 )
 bazel_dep(
     name = "common-jvm",
-    version = "0.137.0",
+    version = "0.138.0",
     repo_name = "wfa_common_jvm",
 )
 bazel_dep(
@@ -385,14 +385,14 @@ oci = use_extension("@rules_oci//oci:extensions.bzl", "oci")
 oci.pull(
     name = "go_image_base",
     # Digest of `nonroot` tag.
-    digest = "sha256:cd961bbef4ecc70d2b2ff41074dd1c932af3f141f2fc00e4d91a03a832e3a658",
+    digest = "sha256:8b9f2e503e55aff85b79d6b22c7a63a65170e8698ae80de680e3f5ea600977bf",
     image = "gcr.io/distroless/base-debian12",
     platforms = ["linux/amd64"],
 )
 oci.pull(
     name = "py_image_base",
     # Digest of `nonroot` tag.
-    digest = "sha256:498e3cb48826d99bad1301994da60caa126b25c0ce03dc4e6a47a7c10fe02493",
+    digest = "sha256:3aa5d1eb6e9f83a4ed806174bb40296e04758ab95f53b61f976d3df2b57ca289",
     image = "gcr.io/distroless/python3-debian12",
     platforms = ["linux/amd64"],
 )
@@ -401,7 +401,7 @@ oci.pull(
 oci.pull(
     name = "java_image_base_root",
     # Digest of `latest` tag.
-    digest = "sha256:fd925ba431f3a6c1f1c8114ce1999ca38803220baf0fdf25a4c71b38db8af67f",
+    digest = "sha256:0210d5b77dfbe851916184405ddfc8297e7bc652029f7a485274cb71ea2462bc",
     image = "gcr.io/distroless/java17-debian12",
     platforms = ["linux/amd64"],
 )
--- a/MODULE.bazel.lock
+++ b/MODULE.bazel.lock
@@ -538,8 +538,8 @@
     "https://raw.githubusercontent.com/world-federation-of-advertisers/bazel-registry/main/modules/common-cpp/0.14.0/MODULE.bazel": "1bfc0048abf7c87aba25d041caaf8f96e0252848d79de6b442cd56ace3189f2e",
     "https://raw.githubusercontent.com/world-federation-of-advertisers/bazel-registry/main/modules/common-cpp/0.14.0/source.json": "abc2a96da0f5af0a85672811998798a55a8dce68083a470cfd341c68e06d2262",
     "https://raw.githubusercontent.com/world-federation-of-advertisers/bazel-registry/main/modules/common-jvm/0.129.0/MODULE.bazel": "2bf570f12d4cd8b5d743bc3d619b7bd554f7f18d8179114c33a30f39abb7e447",
-    "https://raw.githubusercontent.com/world-federation-of-advertisers/bazel-registry/main/modules/common-jvm/0.137.0/MODULE.bazel": "a1c2b142de329ed2ba19f3c938d4917ad776ad8652c66e128ee4889913998271",
-    "https://raw.githubusercontent.com/world-federation-of-advertisers/bazel-registry/main/modules/common-jvm/0.137.0/source.json": "428ad65e76686804095f5c2805d0d5cbe96abe5588d7ed93ab4ea2a319def92c",
+    "https://raw.githubusercontent.com/world-federation-of-advertisers/bazel-registry/main/modules/common-jvm/0.138.0/MODULE.bazel": "fcc9ff6b99766295495a604a81dcb059607c61caf1fcded6cbb76b1db99dba48",
+    "https://raw.githubusercontent.com/world-federation-of-advertisers/bazel-registry/main/modules/common-jvm/0.138.0/source.json": "d6fd2dec624842d48bdcb21b770368832c87f208e7d0056766cda468c48a622c",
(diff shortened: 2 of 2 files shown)


stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3613)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on SanjayVas and ugur47).

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/ProtoConversions.kt` line 756 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3613#-OmfDFaL9mlRU2qE-R1i:-OmfDFaL9mlRU2qE-R1j:b-6ntiw8) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/038cc48a89a8dc69a4d02752249c7d4b3385d672/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/ProtoConversions.kt#L756)):*
> ```Kotlin
>         Measurement.Result.Reach.MethodologyCase.METHODOLOGY_NOT_SET -> {}
>       }
>     } else if (protocolConfig.protocolsList.any { it.hasLiquidLegionsV2() }) {
> ```

trustee should be the priority check

stevenwarejones (changes requested):
> @stevenwarejones reviewed 2 files and all commit messages, made 2 comments, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3613)* status: all files reviewed, 2 unresolved discussions (waiting on SanjayVas and ugur47).

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/ProtoConversions.kt` line 746 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3613#-OmjQRYH4-M-U7Y_-V9-:-OmjQRYH4-M-U7Y_-V90:b18v85v) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a919e30d4b858a7278e4fb4ef442b86afde95f81/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/ProtoConversions.kt#L746)):*
> ```Kotlin
>       // source for frequency_vector_size and the default instance is the only option.
>       // TODO: update once the public API Measurement.Result exposes a TrusTee methodology field.
>       trusTee = TrusTee.getDefaultInstance()
> ```

```
source.trusTee.toInternal()
```

___
*[`src/test/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsServiceTest.kt` line 9150 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3613#-OmjQa_fAbUwBFyW1_KQ:-OmjQa_g5kRVaXYvbUYW:b5ez9oy) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a919e30d4b858a7278e4fb4ef442b86afde95f81/src/test/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/MetricsServiceTest.kt#L9150)):*
> ```Kotlin
>                       value = REACH_FREQUENCY_REACH_VALUE
>                       noiseMechanism = NoiseMechanism.CONTINUOUS_GAUSSIAN
>                       trusTee = TrusTee.getDefaultInstance()
> ```

need a test case without default instance

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3613)* status: all files reviewed, 2 unresolved discussions (waiting on SanjayVas and ugur47).

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/ProtoConversions.kt` line 746 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3613#-OmjQRYH4-M-U7Y_-V9-:-OmjkfcZEHg2TQtIOiFX:byfnfj3) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a919e30d4b858a7278e4fb4ef442b86afde95f81/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/ProtoConversions.kt#L746)):*

The existing TODO comment tracks this.  Here is a more in depth explanation:

There are two separate proto objects involved here, and the gap is between them:


1. protocolConfig is Measurement.protocol_config (field #5 in measurement.proto), whose type is ProtocolConfig from protocol_config.proto. The Kingdom sets this OUTPUT_ONLY to record which protocol was selected.
  ProtocolConfig.Protocol does have TrusTee trus_tee = 5 (see protocol_config.proto https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/main/src/main/proto/wfa/measurement/api/v2alpha/protocol_config.proto#L223), so it.hasTrusTee() and reading noiseMechanism from it are both valid.


  2. source is Measurement.Result.Frequency / Measurement.Result.Reach from measurement.proto (https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/main/src/main/proto/wfa/measurement/api/v2alpha/measurement.proto) — the decrypted Duchy output. The methodology oneof in both messages lists custom_direct_methodology, deterministic_*,
  liquid_legions_*, and honest_majority_share_shuffle, but has no TrusTee field at v0.89.0. So source.trusTee does not exist and would be a compile error.

  The HMSS case works because HonestMajorityShareShuffleMethodology is present in both places: in ProtocolConfig.Protocol to identify the protocol, and in Measurement.Result.{Reach,Frequency}.methodology to
  carry the frequency_vector_size from the Duchy result. TrusTee only has the first half — the protocol config side — but nothing on the result side yet.

  Until the public API adds a TrusTee methodology field to Measurement.Result, TrusTee.getDefaultInstance() is the only option, and the test must match that.
</blockquote></details>

can you take care of that public api change then and then add it here?

ugur47 (commented):
> @ugur47 resolved 1 discussion and dismissed @stevenwarejones from a discussion.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3613)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on kungfucraig, SanjayVas, and stevenwarejones).

SanjayVas (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3613)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on kungfucraig).

PR #3637, opened 2026-03-09
feat: Make HMSS and Trustee produce identical results

### LINKED ISSUE #3636: TrusTee and HMSS produce different results due to Float vs Double precision in vidSamplingIntervalWidth
### Suggest an idea

Summary
The vid_sampling_interval_width proto field is defined as float (32-bit) across all protocol protos (trus_tee.proto, honest_majority_share_shuffle_methods.proto, liquid_legions_v2_encryption_methods.proto). When this value is used for reach/frequency scaling, TrusTee and HMSS produce slightly different results for the same input data due to differences in floating-point precision during division.

Root Cause
TrusTee (Kotlin): The ReachAndFrequencyComputations.computeReach function accepted vidSamplingIntervalWidth as a Float. The key division reachInSample / vidSamplingIntervalWidth was performed in 32-bit float precision.
HMSS (C++): The C++ crypto worker reads the proto float field but promotes it to double (64-bit) for internal computation. The equivalent division is performed in 64-bit double precision.
For vidSamplingIntervalWidth values that are not exactly representable in binary (e.g., 0.3), the 32-bit and 64-bit representations differ slightly. This causes the final toLong() truncation to produce different integer results.

Example
With test data using vidSamplingIntervalWidth = 0.2 (from MetricSpecConfig):

TrusTee (Float): cross-publisher reach = 5369, impressions = 9122
HMSS (Double): cross-publisher reach = 5371, impressions = 9124
The discrepancy disappears when vidSamplingIntervalWidth = 1.0, since 1.0 is exactly representable in both 32-bit and 64-bit IEEE 754.

stevenwarejones:
Closes #3636

Issue: #3636


kungfucraig (approved):
> @kungfucraig reviewed 13 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3637)* status: all files reviewed, 1 unresolved discussion (waiting on SanjayVas and stevenwarejones).

___
*[`src/main/kotlin/org/wfanet/measurement/integration/common/reporting/v2/InProcessEdpAggregatorLifeOfAReportTest.kt` line 1118 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3637#-OnY6Uql7zfrpm0Tgwts:-OnY6Uql7zfrpm0Tgwtt:bh5dibb) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f4d7984a0d433f141598ad0c2057fbfa9de9ad4b/src/main/kotlin/org/wfanet/measurement/integration/common/reporting/v2/InProcessEdpAggregatorLifeOfAReportTest.kt#L1118)):*
> ```Kotlin
>         }
>     }
>     private const val EXPECTED_CROSS_PUBLISHER_REACH = 5371L
> ```

nit: please make a note that we expect all ways of computing the measurements to produce exactly the same results

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3637)* status: 12 of 15 files reviewed, all discussions resolved (waiting on kungfucraig and SanjayVas).

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3637)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on SanjayVas).

Issue #3578, opened 2026-02-23
VID labeler for the EDPA


TNATALI:
### Suggest an idea

A new EDP Aggregator component that takes encrypted EDP Raw Impressions and apply VID models to generate VID labeled impressions

TNATALI:
There are a few Sub Items still open and under development. Re-opening this parent FR

TNATALI:
All sub-issue are completed. Closing this parent Feature

PR #3682, opened 2026-04-07
fix: Split VidLabeler storage params in config and params protos

jojijac0b:
## Summary

Raw impressions may need to be hosted on a different GCS bucket not owned by the operator. To support this, the `VidLabelerParams` proto needs separate storage configurations for reading raw impressions vs writing VID-labeled output, rather than a single shared `storage_params` field.

**Storage params split:**
- Renames `storage_params` to `vid_labeled_impressions_storage_params` for output storage
- Adds `raw_impressions_storage_params` for input storage configuration
- Renames `labeled_impressions_blob_prefix` → `impressions_blob_prefix` in `StorageParams` to make it generic for both use cases
- Splits `storage_params` in `VidLabelingConfig` into `raw_impressions_storage_params` and `vid_labeled_impressions_storage_params` so the dispatcher can assign different GCP projects for reading vs writing
- Renumbers fields sequentially in both `VidLabelingConfig` and `VidLabelerParams` (nothing is deployed yet)

**KMS config restructuring:**
- Removes `vid_labeled_impressions_kek_uri` from both `VidLabelerParams` and `VidLabelingConfig` — KMS config does not belong in the dispatcher path
- Adds `raw_impression_kms_config` (field 5) to `EventDataProviderConfig` for decrypting raw impression data, only set for EDPs that use the VID Labeler
- The existing `kms_config` (field 2) in `EventDataProviderConfig` handles labeled impression encryption (shared with ResultsFulfiller)
- Removes all KEK URI wiring from `VidLabelingDispatcher`, `VidLabelingDispatcherFunction`, and tests

**Updates:**
- Updates `VidLabelingDispatcher`, `VidLabelingDispatcherFunction`, and tests to use the new field names

## Test plan

- [ ] Existing `VidLabelingDispatcherTest` passes with updated field names
- [ ] Proto compiles successfully

Issue: #3683

PR #3684, opened 2026-04-08
feat: add VidLabeler implementation and VidLabelerMetrics

jojijac0b:
## Summary

Adds the `VidLabeler` class and `VidLabelerMetrics` for the VID labeling pipeline.

**VidLabeler** implements the core pipeline for labeling raw impressions with VIDs:
- `listBatchFileUris()` — paginated listing of files in a batch via `RawImpressionMetadataBatchFileService`
- `readAndDecryptRawImpressions()` — reads `BlobDetails` from GCS, decrypts via `EncryptedStorage`
- `resolveModelLines()` — filters model line configs by override list
- `labelImpressions()` — TODO, pending VidModelCache/VidModelResolver
- `encryptAndWriteLabeledImpressions()` — groups by (model line, event group), envelope-encrypts, writes to GCS with BlobDetails metadata
- `writeOutputMetadata()` — batch creates `ImpressionMetadata` entries via gRPC
- `markBatchProcessed()` / `markBatchFailed()` — updates batch state

**VidLabelerMetrics** follows the `ResultsFulfillerMetrics` pattern:
- 6 duration histograms (one per pipeline stage)
- 2 counters (batches processed, impressions labeled) with success/failure attributes
- Shared `MetricsUtil.measured()` extension function for timing code blocks
- `create()` companion factory using `Instrumentation.meter`

Issue: #3578

## Test plan

- [x] `VidLabeler` and `VidLabelerMetrics` BUILD targets compile
- [x] `labelBatch marks batch as FAILED when an error occurs` — verifies error handling path
- [x] `labelBatch throws NotImplementedError when batch has no files` — verifies empty batch behavior
- [x] `envelope encryption round-trips labeled impressions through write and read` — verifies encrypt/decrypt pattern
- [x] `readBlobDetails parses binary proto metadata` — verifies binary protobuf parsing path
- [x] `readBlobDetails falls back to JSON when binary parsing fails` — verifies JSON fallback path


jojijac0b:
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/vidlabeler/VidLabeler.kt` line 262 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3684#-OpfOZKk78YKv11qjj5d:-OpfOZKk78YKv11qjj5e:blus05m) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ec8c0b9d23cc3b197de70f937376f87333c8f5f8/src/main/kotlin/org/wfanet/measurement/edpaggregator/vidlabeler/VidLabeler.kt#L262)):*
> ```Kotlin
>     labeledImpressions: List<LabeledImpressionOutput>,
>   ): List<BlobDetails> {
>     // TODO: Implement encryptAndWriteLabeledImpressions using EncryptedStorage.
> ```

Blocked on #3682 merging — the encrypt KMS client will come from `EventDataProviderConfig.kms_config`, wired through `VidLabelerAppRunner` at startup (KEK URI was removed from `VidLabelerParams` in #3682).

jojijac0b:
## Design Update — Proto & Architecture Changes

This PR now reflects the final design after the storage params split landed in #3682:

**Proto changes (from #3682, now on main):**
- `VidLabelerParams.storage_params` → split into `raw_impressions_storage_params` (field 2) and `vid_labeled_impressions_storage_params` (field 3)
- `StorageParams.labeled_impressions_blob_prefix` → renamed to `impressions_blob_prefix` (generic for both read/write)
- KEK URI removed from `VidLabelingConfig` — KMS config now lives in `EventDataProviderConfig` per Marco's suggestion (Sanjay +1'd). The existing `kms_config` (field 2) handles labeled impression encryption, and the new `raw_impression_kms_config` (field 5) handles raw impression decryption.

**What this PR adds on top:**
- `VidLabeler.encryptAndWriteLabeledImpressions` — groups impressions by (modelLine, eventGroupReferenceId), generates a DEK per group via envelope encryption, writes encrypted data + BlobDetails metadata
- `encryptKekUri: String` constructor parameter on `VidLabeler` — needed for DEK generation (the `KmsClient` alone doesn't expose its bound KEK URI)
- Tests for the write path using `FakeKmsClient` + `FileSystemStorageClient` for local validation

**VidLabelerApp** validates both storage params and `raw_impression_metadata_batch` before delegating to `VidLabeler`. The `VidLabeler` dependency itself is deferred (TODO) — it will be wired in the AppRunner PR (#3685).

stevenwarejones (changes requested):
1. dont reference internal task names like `T263656471` in the description.
2. Title is Add VidLabelerAppRunner and KMS config. Should be feat: Add VidLabelerAppRunner and KMS config.   

> @stevenwarejones reviewed 2 files and made 6 comments.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3684#-:-Orn7PEz5cQiWMk_BOGO:b-20r2ul)* status: 2 of 7 files reviewed, 6 unresolved discussions (waiting on jojijac0b, Marco-Premier, and SanjayVas).

___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/vidlabeler/VidLabeler.kt` line 194 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3684#-Orn7mCL4pvHYNVAffy9:-Orn7mCM1fCcwJ-M2jH_:b-d0c0yh) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/228916b17ed1a153f72e79ddccb16e00023874fc/src/main/kotlin/org/wfanet/measurement/edpaggregator/vidlabeler/VidLabeler.kt#L194)):*
> ```Kotlin
>   }
> 
>   /** Reads a [BlobDetails] proto from a metadata blob URI. */
> ```

document the json fallback behavior
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/vidlabeler/VidLabeler.kt` line 207 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3684#-Orn7bUzEYIlT6zypXco:-Orn7bUzEYIlT6zypXcp:bwbqvff) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/228916b17ed1a153f72e79ddccb16e00023874fc/src/main/kotlin/org/wfanet/measurement/edpaggregator/vidlabeler/VidLabeler.kt#L207)):*
> ```Kotlin
>     } catch (e: com.google.protobuf.InvalidProtocolBufferException) {
>       val builder = BlobDetails.newBuilder()
>       JsonFormat.parser().ignoringUnknownFields().merge(bytes.toString(Charsets.UTF_8), builder)
> ```

i don't think we should be ignoring unknown fields here, right?
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/vidlabeler/VidLabelerMetrics.kt` line 107 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3684#-Orn7wXM7CXDRgG2albQ:-Orn7wXM7CXDRgG2albR:b34n8yv) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/228916b17ed1a153f72e79ddccb16e00023874fc/src/main/kotlin/org/wfanet/measurement/edpaggregator/vidlabeler/VidLabelerMetrics.kt#L107)):*
> ```Kotlin
>         block()
>       } finally {
>         this.record(timer.elapsedNow().inWholeNanoseconds / 1_000_000_000.0)
> ```

this is in a few places. can you combine the two existing and this one to use a common helper function?
___
*[`src/test/kotlin/org/wfanet/measurement/edpaggregator/vidlabeler/VidLabelerTest.kt` line 171 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3684#-Orn8JcG52pnw7Wck_Bs:-Orn8JcG52pnw7Wck_Bt:bg2hr0r) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/228916b17ed1a153f72e79ddccb16e00023874fc/src/test/kotlin/org/wfanet/measurement/edpaggregator/vidlabeler/VidLabelerTest.kt#L171)):*
> ```Kotlin
> 
>   @Test
>   fun `encryptAndWriteLabeledImpressions round-trips through read`() = runBlocking {
> ```

this is an awkward test. it doesn't call the very method the test title states it tests.
___
*[`src/test/kotlin/org/wfanet/measurement/edpaggregator/vidlabeler/VidLabelerTest.kt` line 263 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3684#-Orn96Bw2ZL6vkwVt2Zj:-Orn96Bx398-Iml7s-by:b-uyxk89) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/228916b17ed1a153f72e79ddccb16e00023874fc/src/test/kotlin/org/wfanet/measurement/edpaggregator/vidlabeler/VidLabelerTest.kt#L263)):*
> ```Kotlin
>       mapOf(
>         MODEL_LINE_1 to
>           VidLabelerParams.ModelLineConfig.newBuilder()
> ```

use kotlin dsls

stevenwarejones (changes requested):
> @stevenwarejones made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3684)* status: 2 of 7 files reviewed, 7 unresolved discussions (waiting on jojijac0b, Marco-Premier, and SanjayVas).

___
*[`src/test/kotlin/org/wfanet/measurement/edpaggregator/vidlabeler/VidLabelerTest.kt` line 171 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3684#-Orn8JcG52pnw7Wck_Bs:-Orn9xPZ7JzuvB6i6RxC:bykizlf) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/228916b17ed1a153f72e79ddccb16e00023874fc/src/test/kotlin/org/wfanet/measurement/edpaggregator/vidlabeler/VidLabelerTest.kt#L171)):*

this is an awkward test. it doesn't call the very method the test title states it tests.
</blockquote></details>

additionally, i think we need a test that tests 
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/vidlabeler/VidLabeler.kt` line 253 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3684#-OrnAUcfApOA4SOfZHsk:-OrnAUcfApOA4SOfZHsl:b-oc8en2) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/228916b17ed1a153f72e79ddccb16e00023874fc/src/main/kotlin/org/wfanet/measurement/edpaggregator/vidlabeler/VidLabeler.kt#L253)):*
> ```Kotlin
>     //  4. Run inference to get VID.
>     //  5. Build LabeledImpression with VID, event_time, event, and event_group_reference_id.
>     throw NotImplementedError("VID model inference not yet implemented")
> ```

shouldn't the PR for implementing labelImpressions be done first?

stevenwarejones (commented):
1. NotImplementedError bypasses the catch block — labelBatch() catches Exception, but NotImplementedError extends Error, not Exception. If a production batch hits the unimplemented labelImpressions, markBatchFailed() is never
  called and the batch gets stuck. The test even acknowledges this: "NotImplementedError extends Error, not Exception, so VidLabeler's catch block does not invoke markBatchFailed." Should either catch Throwable or guard against
  this explicitly.

  2. Round-trip test still doesn't test actual code (previous comment #6) — The test was renamed to envelope encryption round-trips labeled impressions through write and read but it still manually replicates the encryption
  pattern rather than calling encryptAndWriteLabeledImpressions. It proves Tink works, not that VidLabeler works. The method is private, so either make it internal/@VisibleForTesting, or test it through labelBatch() once
  labelImpressions is implemented.
  
  3. Misleading test name: labelBatch throws NotImplementedError when batch has no files — The NotImplementedError is thrown because labelImpressions is unimplemented, not because the batch has no files. A batch WITH files would
  hit the same error after reading them. The test name implies the error is caused by the empty batch.

  4. readBlobDetails tests are indirect and fragile — Both the binary and JSON tests work by calling labelBatch() and asserting it fails at a later stage ("Raw impression blob not found"). If any upstream logic changes, these
  tests break even though readBlobDetails is fine. They also don't assert on the parsed BlobDetails field values. Better to make readBlobDetails testable directly.

  5. No test for resolveModelLines — Neither the default path (use all model lines) nor the override path (filter + requireNotNull) is tested. The override with an invalid model line name should also be tested.

  6. No test for encryptAndWriteLabeledImpressions grouping logic — The method groups by (modelLine, eventGroupReferenceId) and writes separate blobs per group. This grouping is the most complex logic in the write path and has no
   coverage.

  7. Existing callers not refactored to use MetricsUtil — Your previous comment said "combine the two existing and this one." MetricsUtil.kt was created, but I don't see existing callers being updated in this diff. The old inline
   measured() implementations should be removed and replaced with MetricsUtil.measured.

> @stevenwarejones reviewed 4 files and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3684#-:-Ot_wagrAhrYa_FFA3J3:bykqeob)* status: 4 of 9 files reviewed, 7 unresolved discussions (waiting on jojijac0b, Marco-Premier, and SanjayVas).

PR #3698, opened 2026-04-14
feat: add invalid deletion checks in DataAvailabilityMonitor

### LINKED ISSUE #3699: Add spurious deletion detection to DataAvailabilityMonitor
### Suggest an idea

Three scenarios cause spurious deletions:
                                                                                                                                                                                                                                                      
  1. Delete-then-recreate at the same path: A file is explicitly deleted, then a new file is uploaded to the same path. The OBJECT_DELETE event triggers DataAvailabilityCleanup to soft-delete the ImpressionMetadata entry. The subsequent          
  OBJECT_FINALIZE attempts to create a new entry, but the blobUri-based requestId was already consumed by the original creation, so it returns the existing (now deleted) record idempotently. This happens regardless of whether object versioning is
   enabled.                                                                                                                                                                                                                                           
  2. File overwrite without versioning: A file is overwritten, which GCS treats as a delete of the old version plus creation of the new version. Both OBJECT_DELETE and OBJECT_FINALIZE events fire independently with no ordering guarantee, so the
  delete handler can run after the finalize handler, permanently marking the entry as deleted even though the blob exists. Enabling object versioning prevents this scenario by firing OBJECT_ARCHIVE instead of OBJECT_DELETE on overwrite.          
  3. Noncurrent version deletion with versioning enabled: When object versioning is enabled and a noncurrent version is permanently deleted (e.g., by lifecycle rules), GCS fires OBJECT_DELETE for that version. The event contains the same object
  path as the live version, so the delete handler soft-deletes the Kingdom entry even though the live version still exists on the bucket.

stevenwarejones:
Closes #3699

Issue: #3699


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3698)

src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/dataavailability/DataAvailabilityMonitorFunction.kt
```diff
@@ -119,15 +126,39 @@ class DataAvailabilityMonitorFunction : HttpFunction {
...
+    val spuriousDeletionConfig = config.spuriousDeletionCheck
+    val impressionMetadataStub: ImpressionMetadataServiceCoroutineStub? =
+      if (spuriousDeletionConfig.enabled) {
+        require(config.hasImpressionMetadataConnection()) {
+          "impression_metadata_connection must be set when spurious_deletion_check is enabled"
+        }
+        require(config.dataProviderName.isNotEmpty()) {
+          "data_provider_name must be set when spurious_deletion_check is enabled"
+        }
+        val channel = createImpressionMetadataChannel(config.impressionMetadataConnection)
```
georgi: When `spurious_deletion_check` is enabled this allocates a new `ManagedChannel` per invocation/config, but nothing closes or reuses it. In a warm Cloud Functions JVM that will leak gRPC transports and threads across requests. Can we cache this like `DataAvailabilitySyncFunction`, or explicitly close the channel after the monitor run?

src/main/kotlin/org/wfanet/measurement/edpaggregator/dataavailability/DataAvailabilityMonitor.kt
```diff
@@ -279,14 +320,101 @@ class DataAvailabilityMonitor(
...
+    }
+
+    require(spuriousDeletionLookbackDays > 0) { "spuriousDeletionLookbackDays must be > 0" }
+    require(dataProviderName.isNotBlank()) { "dataProviderName must not be blank" }
+
+    val modelLineName = modelLineKey.toName()
+    val today = clock()
+    val cutoffDate = today.minusDays(spuriousDeletionLookbackDays.toLong())
+    val cutoffEpochSeconds = cutoffDate.atStartOfDay(ZoneOffset.UTC).toEpochSecond()
+    val nowEpochSeconds = Instant.now().epochSecond
```
georgi: This lookback window is not anchored to the same notion of "today" used by `checkFullStatus()`: it recomputes `clock()` per model line, converts the start boundary to UTC midnight, and uses `Instant.now()` for the end boundary. Around midnight or with a non-UTC `time_zone`, different model lines can be queried with different windows and boundary rows can flip in or out. Please derive the full interval once from the invocation's configured date/time zone and reuse it for every model line.

src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/dataavailability/DataAvailabilityMonitorFunction.kt
```diff
@@ -203,11 +242,31 @@ class DataAvailabilityMonitorFunction : HttpFunction {
...
+    return buildMutualTlsChannel(impressionMetadataTarget, signingCerts, impressionMetadataCertHost)
+      .withShutdownTimeout(Duration.ofSeconds(channelShutdownDurationSeconds))
+  }
+
   companion object {
     private val logger: Logger = Logger.getLogger(this::class.java.name)
 
     private val fileSystemPath: String? = System.getenv("DATA_AVAILABILITY_FILE_SYSTEM_PATH")
 
+    private val impressionMetadataTarget: String = System.getenv("IMPRESSION_METADATA_TARGET") ?: ""
```
georgi: `IMPRESSION_METADATA_TARGET` becomes required once `spurious_deletion_check` is enabled, but this defaults it to an empty string instead of failing fast. That means a misconfigured deployment will not fail until channel construction or the first request, and the error will be much less actionable. `DataAvailabilitySyncFunction` already uses `EnvVars.checkNotNullOrEmpty("IMPRESSION_METADATA_TARGET")`; this should probably do the same.

src/main/kotlin/org/wfanet/measurement/edpaggregator/dataavailability/DataAvailabilityMonitor.kt
stevenwarejones: Done.

src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/dataavailability/DataAvailabilityMonitorFunction.kt
stevenwarejones: added it. this is a scheduled job so I'm not really sure its needed though.

src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/dataavailability/DataAvailabilityMonitorFunction.kt
stevenwarejones: Done.

stevenwarejones (commented):
> @stevenwarejones partially reviewed 9 files and made 3 comments.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3698)* status: 9 of 10 files reviewed, 3 unresolved discussions (waiting on georgi and Marco-Premier).

kungfucraig:
Please add the Google team after you've gotten an LGTM from someone in your org. @SanjayVas can take care of this one once it's ready.

Marco-Premier (approved):
> @Marco-Premier reviewed 10 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3698)* status: all files reviewed, 3 unresolved discussions (waiting on georgi, kungfucraig, and SanjayVas).

SanjayVas (changes requested):
> @SanjayVas reviewed 8 files and all commit messages, and made 4 comments.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3698)* status: all files reviewed, 7 unresolved discussions (waiting on georgi, kungfucraig, and stevenwarejones).

___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/dataavailability/DataAvailabilityMonitor.kt` line 68 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3698#-OqMODDv4vT3ssYTqKuu:-OqMODDv4vT3ssYTqKuv:b2wyycb) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/91d245c87cc7fd23d83bf609bbf3671332ad7a05/src/main/kotlin/org/wfanet/measurement/edpaggregator/dataavailability/DataAvailabilityMonitor.kt#L68)):*
> ```Kotlin
>  * @property activeModelLines Set of model line keys that are expected to upload daily.
>  * @property impressionMetadataStub Optional gRPC stub for the ImpressionMetadata service. Required
>  *   only for [checkSpuriousDeletions].
> ```

I'm not spotting a symbol with this name in this PR. You can probably just say that it's used for checking for spurious deletions. See my other comment on the behavior of `checkFullStatus`.



_Code quote:_
```Kotlin
only for [checkSpuriousDeletions]
```
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/dataavailability/DataAvailabilityMonitor.kt` line 137 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3698#-OqMOs5xEg7u7ow4AdUL:-OqMOs5xEg7u7ow4AdUM:banq4uc) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/91d245c87cc7fd23d83bf609bbf3671332ad7a05/src/main/kotlin/org/wfanet/measurement/edpaggregator/dataavailability/DataAvailabilityMonitor.kt#L137)):*
> ```Kotlin
>    *   considered stale.
>    * @param clock Provides the current date for staleness checks.
>    * @param spuriousDeletionLookbackDays If set and [impressionMetadataStub] is available, also
> ```

Having `impressionMetadataStub` be non-null should be a precondition for specifying a non-null value for the `spuriousDeletionLookbackDays` rather than having conditional behavior. Have this method throw an error in this case instead.



_Code quote:_
```Kotlin
 and [impressionMetadataStub] is available
```
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/dataavailability/DataAvailabilityMonitor.kt` line 353 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3698#-OqMPP4lCiTyNalfA3_D:-OqMPP4lCiTyNalfA3_E:bmgnyrn) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/91d245c87cc7fd23d83bf609bbf3671332ad7a05/src/main/kotlin/org/wfanet/measurement/edpaggregator/dataavailability/DataAvailabilityMonitor.kt#L353)):*
> ```Kotlin
> 
>     if (
>       spuriousDeletionInterval == null || impressionMetadataStub == null || dataProviderName == null
> ```

Same note here re: preconditions.
___
*[`src/main/proto/wfa/measurement/config/edpaggregator/data_availability_monitor_config.proto` line 70 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3698#-OqMPxm91JgxYtsPprbl:-OqMPxm91JgxYtsPprbm:bt4w621) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/91d245c87cc7fd23d83bf609bbf3671332ad7a05/src/main/proto/wfa/measurement/config/edpaggregator/data_availability_monitor_config.proto#L70)):*
> ```proto
>   // Whether to enable the spurious deletion check. When false or unset,
>   // the check is skipped entirely.
>   bool enabled = 1;
> ```

You already have the presence of the field to indicate whether this is enabled. Even that is not technically necessary as `lookback_days` not being specified is sufficient (0 is not a valid value).

At least drop the boolean and consider lookback_days to be required if `spurious_deletion_check` is specified. If you don't expect to need other parameters to the spurious deletion check, you can drop the message type entirely and just inline the lookback days field into the parent message type where it does nothing if not specified.

stevenwarejones (commented):
> @stevenwarejones partially reviewed 3 files and made 4 comments.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3698)* status: 8 of 11 files reviewed, 7 unresolved discussions (waiting on georgi, kungfucraig, Marco-Premier, and SanjayVas).

___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/dataavailability/DataAvailabilityMonitor.kt` line 68 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3698#-OqMODDv4vT3ssYTqKuu:-OqMxJhCCLIG_wsZzU_V:b-kd3ane) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/91d245c87cc7fd23d83bf609bbf3671332ad7a05/src/main/kotlin/org/wfanet/measurement/edpaggregator/dataavailability/DataAvailabilityMonitor.kt#L68)):*

I'm not spotting a symbol with this name in this PR. You can probably just say that it's used for checking for spurious deletions. See my other comment on the behavior of `checkFullStatus`.
</blockquote></details>

yeah - sorry - I refactored along the way and didn't update this comment. thx for pointing out.
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/dataavailability/DataAvailabilityMonitor.kt` line 137 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3698#-OqMOs5xEg7u7ow4AdUL:-OqMxSrK5iuwEizdAueg:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/91d245c87cc7fd23d83bf609bbf3671332ad7a05/src/main/kotlin/org/wfanet/measurement/edpaggregator/dataavailability/DataAvailabilityMonitor.kt#L137)):*

Having `impressionMetadataStub` be non-null should be a precondition for specifying a non-null value for the `spuriousDeletionLookbackDays` rather than having conditional behavior. Have this method throw an error in this case instead.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/dataavailability/DataAvailabilityMonitor.kt` line 353 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3698#-OqMPP4lCiTyNalfA3_D:-OqMzFCP2jXnGpumfJui:b2ecdfe) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/91d245c87cc7fd23d83bf609bbf3671332ad7a05/src/main/kotlin/org/wfanet/measurement/edpaggregator/dataavailability/DataAvailabilityMonitor.kt#L353)):*

Same note here re: preconditions.
</blockquote></details>

anything else you want changed here? I wasn't sure I understood.
___
*[`src/main/proto/wfa/measurement/config/edpaggregator/data_availability_monitor_config.proto` line 70 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3698#-OqMPxm91JgxYtsPprbl:-OqMx_UEDi88M6_7cCc3:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/91d245c87cc7fd23d83bf609bbf3671332ad7a05/src/main/proto/wfa/measurement/config/edpaggregator/data_availability_monitor_config.proto#L70)):*

You already have the presence of the field to indicate whether this is enabled. Even that is not technically necessary as `lookback_days` not being specified is sufficient (0 is not a valid value).

At least drop the boolean and consider lookback_days to be required if `spurious_deletion_check` is specified. If you don't expect to need other parameters to the spurious deletion check, you can drop the message type entirely and just inline the lookback days field into the parent message type where it does nothing if not specified.
</blockquote></details>

Done.

SanjayVas (approved):
> @SanjayVas reviewed 6 files and all commit messages, made 4 comments, and resolved 3 discussions.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3698)* status: all files reviewed, 7 unresolved discussions (waiting on georgi, kungfucraig, and stevenwarejones).

___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/dataavailability/DataAvailabilityMonitor.kt` line 353 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3698#-OqMPP4lCiTyNalfA3_D:-OqN372W2nnNdhWPHPsn:b-1ccpqz) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/91d245c87cc7fd23d83bf609bbf3671332ad7a05/src/main/kotlin/org/wfanet/measurement/edpaggregator/dataavailability/DataAvailabilityMonitor.kt#L353)):*

anything else you want changed here? I wasn't sure I understood.
</blockquote></details>

Use precondition functions for `impressionMetadataStub` and `dataProviderName`. e.g. `checkNotNull`.
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/dataavailability/DataAvailabilityMonitor.kt` line 158 at r9](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3698#-OqN3ZbY0kngfaPHbkMT:-OqN3ZbY0kngfaPHbkMU:b-lz68em) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/b1bfea40662db5a26bce1cf5abab5795eb6f2909/src/main/kotlin/org/wfanet/measurement/edpaggregator/dataavailability/DataAvailabilityMonitor.kt#L158)):*
> ```Kotlin
>         ?.also {
>           require(it > 0) { "spuriousDeletionLookbackDays must be > 0" }
>           requireNotNull(impressionMetadataStub) {
> ```

nit: `check` rather than `require` for preconditions relating to instance state (IllegalStateException instead of IllegalArgumentException for `impressionMetadataStub` and `dataProviderName`



_Code quote:_
```Kotlin
require
```
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/dataavailability/DataAvailabilityMonitor.kt` line 161 at r9](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3698#-OqN3why6YtfuDe5JB_l:-OqN3why6YtfuDe5JB_m:btwrjgd) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/b1bfea40662db5a26bce1cf5abab5795eb6f2909/src/main/kotlin/org/wfanet/measurement/edpaggregator/dataavailability/DataAvailabilityMonitor.kt#L161)):*
> ```Kotlin
>             "impressionMetadataStub must be set when spuriousDeletionLookbackDays is specified"
>           }
>           require(!dataProviderName.isNullOrBlank()) {
> ```

nit: just check for null here. Invalid values such as blank should be checked in the constructor



_Code quote:_
```Kotlin
isNullOrBlank()
```
___
*[`src/main/proto/wfa/measurement/config/edpaggregator/data_availability_monitor_config.proto` line 57 at r9](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3698#-OqN46HV9nswwoFF9c4s:-OqN46HV9nswwoFF9c4t:b-h8pahd) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/b1bfea40662db5a26bce1cf5abab5795eb6f2909/src/main/proto/wfa/measurement/config/edpaggregator/data_availability_monitor_config.proto#L57)):*
> ```proto
>   // spurious deletions (entries marked deleted whose blobs still exist on
>   // the bucket). When set to a value > 0, the spurious deletion check is
>   // enabled. When unset or 0, the check is skipped.
> ```

nit: no need to distinguish as these mean the exact same thing



_Suggestion:_
```proto
unspecified
```

stevenwarejones (commented):
> @stevenwarejones reviewed 4 files and all commit messages, and resolved 3 discussions.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3698)* status: all files reviewed, 4 unresolved discussions (waiting on georgi and kungfucraig).

stevenwarejones (commented):
> @stevenwarejones made 1 comment, resolved 3 discussions, and dismissed @SanjayVas and @georgi from 3 discussions.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3698)* status: all files reviewed, 1 unresolved discussion (waiting on georgi and kungfucraig).

___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/dataavailability/DataAvailabilityMonitor.kt` line 353 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3698#-OqMPP4lCiTyNalfA3_D:-OqNA3ZI2cHsXRBVTtz5:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/91d245c87cc7fd23d83bf609bbf3671332ad7a05/src/main/kotlin/org/wfanet/measurement/edpaggregator/dataavailability/DataAvailabilityMonitor.kt#L353)):*

Use precondition functions for `impressionMetadataStub` and `dataProviderName`. e.g. `checkNotNull`.
</blockquote></details>

Done.

stevenwarejones (commented):
> @stevenwarejones resolved 1 discussion and dismissed @georgi from a discussion.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3698)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on georgi and kungfucraig).

PR #275, opened 2021-09-14
Only check the error type when the error exist in the log.

wangyaopw:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/275)

PR #3705, opened 2026-04-20
feat: add entity_key and entity_metadata to internal kingdom EventGroup schema

### LINKED ISSUE #3708: Support Entity Type and Entity Id in Kingdom
Propagate the API changes made in [275](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/pull/275/changes) to kingdom. 

Support uniqueness and search

ugur47:
Schema + proto only. Wires the new fields from [cross-media-measurement-api #275](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/pull/275) through the internal Kingdom layer.

#### Placement: `entity_key` and `entity_metadata` at the top level of the internal `EventGroup` proto

Initial instinct was to nest `entity_metadata` inside `EventGroupDetails.EventGroupMetadata` alongside the existing `ad_metadata` / `campaign_metadata`. That placement would require `ALTER PROTO BUNDLE UPDATE` on `EventGroupDetails`, which Spanner rejects while that column has indexed stored-generated dependents. `EventGroupDetails` already has three (`BrandName_Tokens`, `CampaignName_Tokens`, `Metadata_Tokens`) plus the `EventGroupsByMetadata` search index, all from `add-event-group-metadata.sql`.

Google's docs confirm this is intended Spanner behavior, not an emulator-only quirk: *"For STORED or generated columns that are indexed, you can't change the data type of the column, or of any columns that the generated column references"* ([Create and manage generated columns](https://cloud.google.com/spanner/docs/generated-column/how-to)). An `ALTER PROTO BUNDLE UPDATE` on `EventGroupDetails` is treated as a data-type change to the locked column.

To nest, we would have had to drop the three TOKENLIST columns plus the search index in one changeset, bundle-update in a second, recreate them in a third.

Putting the fields at the top level of the internal `EventGroup` proto avoids that entire category of churn:

- No `ALTER PROTO BUNDLE UPDATE` — `EventGroupDetails` is untouched.
- No drop-and-recreate of the metadata-search infrastructure.
- Single migration changeset.

#### Column types

`EntityType` (`STRING(MAX)`), `EntityId` (`STRING(MAX)`), and `EntityMetadata` (`` `google.protobuf.Struct` ``) map to top-level Spanner columns on `EventGroups`. `EntityMetadata` is declared as a **proto-typed column, not `BYTES`**.

The proto bundle is extended via `` ALTER PROTO BUNDLE INSERT (`google.protobuf.Struct`) `` — an `INSERT` (not `UPDATE`), so no column is locked against it.

#### Defaults and nullability

`EntityType` is declared `NOT NULL DEFAULT ("campaign")` so Spanner backfills every existing row as part of the `ADD COLUMN` — no separate DML changeset needed — and new rows created without an `entity_key` inherit the same default, matching the public-API filter default of `["campaign"]`. `EntityId` stays `NULL` for legacy rows, which is what the `UNIQUE NULL_FILTERED INDEX EventGroupsByEntityKey` needs to exempt them from the uniqueness constraint.

  Issue: #3708


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3705)

SanjayVas (commented):
> @SanjayVas reviewed 6 files and all commit messages, and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3705)* status: all files reviewed, 2 unresolved discussions (waiting on ugur47).

___
*[`src/main/proto/wfa/measurement/internal/kingdom/event_group.proto` line 105 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3705#-Oqfu0IH4pzZ_vv_cdEp:-Oqfu0IH4pzZ_vv_cdEq:b-6snpj7) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/659b937359fb32c6b80b9e9749be4d6ec667b365/src/main/proto/wfa/measurement/internal/kingdom/event_group.proto#L105)):*
> ```proto
>   // specific to the (DataProvider, EntityKey.entity_type) pair. Stored
>   // serialized in the EntityMetadata column on the EventGroups Spanner table.
>   google.protobuf.Struct entity_metadata = 17;
> ```

While entity type and ID should indeed be top-level and map to columns, `entity_metadata` should be inside EventGroupDetails.
___
*[`src/main/proto/wfa/measurement/internal/kingdom/event_groups_service.proto` line 136 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3705#-OqfuRY2Bc_mhvko4SIB:-OqfuRY2Bc_mhvko4SIC:b-snl7wh) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/659b937359fb32c6b80b9e9749be4d6ec667b365/src/main/proto/wfa/measurement/internal/kingdom/event_groups_service.proto#L136)):*
> ```proto
>     // in. Empty means no filtering at this layer; request-level default
>     // handling happens in the public-API service.
>     repeated string entity_type_in = 17;
> ```

Hold off on this internal service definition change until you have an implementation for it.

SanjayVas (commented):
> That placement would require`ALTER PROTO BUNDLE UPDATE` on `EventGroupDetails`, which Spanner rejects while that column has any dependent stored/tokenlist generated columns. `EventGroupDetails` already has three such dependents (`BrandName_Tokens`, `CampaignName_Tokens`, `Metadata_Tokens`) from `add-event-group-metadata.sql`.

You're saying we can never update EventGroupDetails now? That doesn't sound right.

> @SanjayVas made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3705#-:-OqfuwFT-v4DhqOial32:b-5j5ynz)* status: all files reviewed, 2 unresolved discussions (waiting on ugur47).

SanjayVas (commented):
Can you share the error you're getting or the documentation around it? Is the error from the Spanner emulator or from a real Cloud Spanner instance?

> @SanjayVas made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3705#-:-OqfvAAu9FnMRRMzRXFH:b-arl8za)* status: all files reviewed, 2 unresolved discussions (waiting on ugur47).

ugur47 (commented):
I agree, I don't like this. And we can go the EventGroupDetails route. We would have 3 changesets though. Let me know if you are ok with it. Happy to take that route.   

Here is the buildbuddy logs:   https://halo-cmm.buildbuddy.io/invocation/253c6d50-9e59-40bc-9dce-ebc4cb4e47f7

- SpannerEmulatorRule.createDatabase / SpannerEmulatorDatabaseRule. The failure happens at RUN BATCH batch validator rejects the ALTER PROTO BUNDLE UPDATE wholesale because of the existing BrandName_Tokens / CampaignName_Tokens / Metadata_Tokens dependents, even when the batch also drops those dependents first.

This is spanner Emulator. Not sure about real spanner. What is the protocol to deal with Emulator vs possibly different real spanner behaviour?

> @ugur47 made 3 comments.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3705#-:-Oqg0qiX9Jr_zWKnwLEm:by7izcy)* status: all files reviewed, 2 unresolved discussions (waiting on SanjayVas).

___
*[`src/main/proto/wfa/measurement/internal/kingdom/event_group.proto` line 105 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3705#-Oqfu0IH4pzZ_vv_cdEp:-Oqg2-BGC8guDZzsXMFi:b-xl55yh) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/659b937359fb32c6b80b9e9749be4d6ec667b365/src/main/proto/wfa/measurement/internal/kingdom/event_group.proto#L105)):*

While entity type and ID should indeed be top-level and map to columns, `entity_metadata` should be inside EventGroupDetails.
</blockquote></details>

I agree. lets resolve the other discussion. Im good if you want to go the other way
___
*[`src/main/proto/wfa/measurement/internal/kingdom/event_groups_service.proto` line 136 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3705#-OqfuRY2Bc_mhvko4SIB:-Oqg2QQm6com0RMXA1al:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/659b937359fb32c6b80b9e9749be4d6ec667b365/src/main/proto/wfa/measurement/internal/kingdom/event_groups_service.proto#L136)):*

Hold off on this internal service definition change until you have an implementation for it.
</blockquote></details>

Done.

SanjayVas (commented):
> What is the protocol to deal with Emulator vs possibly different real spanner behaviour?

The first is to try to get a minimal reproduction in a real Cloud Spanner instance. If we get a similar error in the real impl, then we're stuck and have to go about it the long way regardless. If it works in Cloud Spanner then we can take a look to see if there's a chance the issue has already been fixed in the emulator if we update to a newer version. If not then we file a bug but may still have to continue with the long approach.

> @SanjayVas made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3705#-:-Oqg2oTxEtrNRTOkwnsQ:b-etjdbx)* status: 5 of 6 files reviewed, 2 unresolved discussions (waiting on ugur47).

ugur47 (commented):
Google's documentation confirms this is intended Spanner behavior, not an emulator-only restriction.

  **From [Create and manage generated columns](https://cloud.google.com/spanner/docs/generated-column/how-to):**
  > "For `STORED` or generated columns that are indexed, you can't change the data type of the column, or of any columns that the generated column references."

  **From [Make schema updates](https://cloud.google.com/spanner/docs/schema-updates):**
  Proto message type changes are allowed "provided that modified fields of those messages are not used as keys in any table" — but combined with the generated-column restriction above, a proto column whose fields are referenced by indexed/stored generated columns is locked down.

  `EventGroupDetails` is a proto-typed column; the existing `BrandName_Tokens`, `CampaignName_Tokens`, `Metadata_Tokens` are `TOKENLIST` stored generated columns indexed via `EventGroupsByMetadata`. Any `ALTER
  PROTO BUNDLE UPDATE` on `EventGroupDetails` is treated by Spanner as a data-type change to a locked-down column. The emulator's error message (`Cannot change the data type of column EventGroupDetails, which has
  a dependent stored generated column`) matches the documented restriction verbatim.

  There's a broader structural question I think is worth surfacing: those TOKENLIST dependents lock down `EventGroupDetails` against all future bundle updates, not just this one. Every time a new field is added inside `EventGroupDetails` — not only `entity_metadata` — we'd need the same drop-and-recreate-the-search-infrastructure dance (three Liquibase changesets, brief search-index gap during deploy, etc.). Spanner's check isn't "additive" vs "destructive"; any bundle update on a locked column is blocked.

  Given that, I'd propose putting `entity_metadata` at the top level of the internal `EventGroup` proto and storing it as a plain column (`BYTES`-encoded `Struct`), alongside `entity_key`'s top-level placement. The proto shape no longer one-to-one mirrors the public API nesting, but the internal proto is only consumed by Kingdom's public↔internal conversion (which already maps fields individually) and by Spanner (which treats the column as opaque bytes plus two narrow projections). It sidesteps the bundle update cleanly, and also avoids the recurring cost the next time we want to add metadata.

  Happy to do the three-changeset variant if you still prefer `entity_metadata` inside `EventGroupDetails` — just wanted to surface the tradeoff explicitly.

> @ugur47 made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3705#-:-OqjxBpb6KgUXzItm5Wt:bpdjgyo)* status: 5 of 6 files reviewed, 2 unresolved discussions (waiting on SanjayVas).

SanjayVas (commented):
It's really unfortunate to me that this restriction exists, as our intended pattern was to use the protobuf Details messages as a lightweight way to add data without needing manual schema updates (proto bundle updates are partially automated in the sense that the descriptor is generated).

My instinct is to either

1. Deal with the churn to put this into EventGroupDetails (dropping and recreating the indexes) so everything is consistent.
2. Add a protobuf column for entity metadata.

We don't ever want to use `BYTES` when we know it's a protobuf. That is to say we're explicitly okay with having to do proto bundle updates.

@kungfucraig do you have an opinion here? 

> @SanjayVas made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3705#-:-OqlSA4E69pa9LlUsONv:b-8qds4a)* status: 5 of 6 files reviewed, 2 unresolved discussions (waiting on ugur47).

kungfucraig (commented):
What do you think of this:

1. Add a JSON column for all of the things we want indexed, this will be duplicative of some, but perhaps not all fields, in the EventGroupDetails.
2. Update EventGroupDetails and drop the indexes
3. Recreate the indexes on the JSON column

> @kungfucraig made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3705#-:-OquuzEC0A2GkYRsbE0f:bx7xlx9)* status: 5 of 6 files reviewed, 2 unresolved discussions (waiting on SanjayVas and stevenwarejones).

ugur47 (commented):
This has some long term benefits but is a bigger lift than either of the two options already on the table. 

  Roughly 3–4 changesets (vs. 1 for a top-level proto-typed column, 3 for nesting in EventGroupDetails without the JSON column):

  1. DDL — drop EventGroupsByMetadata + the three TOKENLIST generated columns (BrandName_Tokens, CampaignName_Tokens, Metadata_Tokens).
  2. DDL — ALTER PROTO BUNDLE UPDATE EventGroupDetails (to nest entity_metadata inside EventGroupMetadata) with SET PROTO_DESCRIPTORS, ADD COLUMN SearchableMetadata JSON, add EntityType / EntityId, create the two new entity-key indexes.
  3. DML — backfill SearchableMetadata for existing rows from EventGroupDetails (brand / campaign).
  4. DDL — recreate BrandName_Tokens / CampaignName_Tokens / Metadata_Tokens sourced via JSON_VALUE(SearchableMetadata, …), recreate EventGroupsByMetadata.

**More importantly**

**5. It also introduces a two-sources-of-truth pattern: on every EventGroup write, Kingdom has to compose SearchableMetadata JSON from EventGroupDetails and write it alongside. New code, new tests, new ongoing invariant — none of which exists today**

  Not my preference given the above but  let me know if its is your preference, and I'll go ahead and implement this as soon as possible.

> @ugur47 made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3705#-:-Oqv4Uig8wHwKeoyyYdQ:b-unb8wv)* status: 5 of 6 files reviewed, 2 unresolved discussions (waiting on SanjayVas and stevenwarejones).

SanjayVas (commented):
Looping back from Slack conversation: @kungfucraig and I have agreed on my option 2:

> Add a protobuf column for entity metadata.

> @SanjayVas reviewed 1 file, made 2 comments, and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3705#-:-OqwLHvZCIw1jLDFaHJy:brv1sop)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on stevenwarejones).

___
*[`src/main/proto/wfa/measurement/internal/kingdom/event_group.proto` line 105 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3705#-Oqfu0IH4pzZ_vv_cdEp:-OqwL_qh4cx6Jd3KRT1_:b-sy91r9) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/659b937359fb32c6b80b9e9749be4d6ec667b365/src/main/proto/wfa/measurement/internal/kingdom/event_group.proto#L105)):*

I agree. lets resolve the other discussion. Im good if you want to go the other way
</blockquote></details>

From other discussion, putting this inside of EventGroupDetails is messier than we'd like

ugur47 (commented):
Thanks @SanjayVas @kungfucraig I implemented it. ptal

> @ugur47 made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3705#-:-Oqzp1KU61yRDa0gH9EO:bn3058b)* status: 4 of 6 files reviewed, all discussions resolved (waiting on SanjayVas and stevenwarejones).

kungfucraig (approved):
> @kungfucraig reviewed 6 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3705)* status: all files reviewed, 1 unresolved discussion (waiting on SanjayVas, stevenwarejones, and ugur47).

___
*[`src/main/resources/kingdom/spanner/add-event-group-entity-key.sql` line 51 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3705#-Or-3GpDAGo8Bfqz-Mgt:-Or-3GpDAGo8Bfqz-Mgu:b-38hukn) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/48cc88d7f8a1449927b920d631fddce10013b471/src/main/resources/kingdom/spanner/add-event-group-entity-key.sql#L51)):*
> ```TSQL
> 
> -- Supports the entity_type_in filter in ListEventGroups.
> CREATE INDEX EventGroupsByEntityType
> ```

nit: Do you really need this index given the one below?

ugur47 (commented):
> @ugur47 made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3705)* status: all files reviewed, 1 unresolved discussion (waiting on kungfucraig, SanjayVas, and stevenwarejones).

___
*[`src/main/resources/kingdom/spanner/add-event-group-entity-key.sql` line 51 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3705#-Or-3GpDAGo8Bfqz-Mgt:-Or-EW8F9I63SLTFROPk:b-66ltq3) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/48cc88d7f8a1449927b920d631fddce10013b471/src/main/resources/kingdom/spanner/add-event-group-entity-key.sql#L51)):*

nit: Do you really need this index given the one below?
</blockquote></details>

Yes - NULL_FILTERED unique index  excludes rows where EntityId is NULL - these are the legacy ones . They will all have 'campaign' as entity type and prefix scan on EventGroupsByEntityKey would miss them

ugur47 (commented):
> @ugur47 resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3705)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on SanjayVas and stevenwarejones).

SanjayVas (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3705)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on stevenwarejones).

SanjayVas (approved):
Make sure to add an `Issue` trailer for the GitHub issue that this is for. See [Dev Standards](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/main/docs/dev-standards.md)

> @SanjayVas made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3705#-:-Or-_wd00uhEu5EyJeoT:b-xnolm6)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on stevenwarejones).

stevenwarejones (approved):
> @stevenwarejones reviewed 6 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3705)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on ugur47).

Issue #3727, opened 2026-04-24
Add End-to-end support for EntityKey across impressions, event groups, and metadata (EDPA)


stevenwarejones:
### Suggest an idea

Add support for `EntityKey` (entity_type + entity_id, per
`wfa.measurement.api.v2alpha.EventGroup.EntityKey`) across the EDP
Aggregator so that impressions and event groups can be tagged, stored,
queried, and filtered by the `DataProvider`'s entity taxonomy (e.g.
`creative`, `ad-set`, `campaign`).
Today the CMMS public API's `EventGroup` already carries an `EntityKey`,
but the EDP Aggregator does not propagate or honor it end-to-end:
- `edpaggregator.v1alpha.LabeledImpression` now has an `entity_keys` field
  (#<PR-for-the-proto-change>) but nothing produces or consumes it yet.
- `edpaggregator.eventgroups.v1alpha.EventGroup` has no `entity_key` field.
- `ImpressionMetadata` cannot be queried by entity key.
- `ResultsFulfiller`'s `FilterSpec` / `FilterProcessor` only filter by
  event-group-reference-id, time range, and CEL.

PR #3743, opened 2026-04-30
feat: filter ResultsFulfiller events by entity_keys

Marco-Premier:
## Summary                                                                                                                                                                                               
   
  Adds an entity-key selector to the ResultsFulfiller pipeline so callers                                                                                                                                  
  can restrict events to a chosen set of `LabeledImpression.entity_keys`.
  Pure logic + tests; wiring from the requisition flow lands in a follow-up. 

Issue: #3727


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3743)

stevenwarejones (commented):
The entity-key filter is wired as an additional AND on top of event group reference ID filtering, but the intent is that entity keys replace event group reference IDs as the selection mechanism when the event group has an entity_key set.                                                                                                                                                                                      
                                    
  Each EventGroup has at most one entity_key (see EventGroup.entity_key in the API). When an event group has one, the entity key — not the event group reference ID — should be used to identify which events belong to it. The two are alternative selectors for the same purpose.                                                                                                                                                             
  
  What needs to change:                                                                                                                                                                                                   
                                                                  
  1. FilterProcessor — use entity-key matching instead of event group reference ID matching, not in addition to it.
  2. FilterSpec.init — relax require(eventGroupReferenceIds.isNotEmpty()) to allow entity keys as the sole selector.
  3. Upstream construction — can you outline in the PR description how the follow-up will decide which selector to populate on FilterSpec? Want to make sure this abstraction fits. 

> @stevenwarejones reviewed 6 files and all commit messages, and made 4 comments.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3743#-:-OrTpQ4X2RvU2ONd_gJ2:b-vbj776)* status: 6 of 7 files reviewed, 3 unresolved discussions (waiting on Marco-Premier).

___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/FilterSpec.kt` line 41 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3743#-OrTnU_m6vE6h_MVeyH1:-OrTnU_m6vE6h_MVeyH2:bjqpuep) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1deca0a0ba84c3a8b98631460d23f4cac1108dee/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/FilterSpec.kt#L41)):*
> ```Kotlin
>   val collectionInterval: Interval,
>   val eventGroupReferenceIds: List<String>,
>   val entityKeys: Set<LabeledImpression.EntityKey> = emptySet(),
> ```

nit: no default value please.
___
*[`src/test/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/FilterProcessorTest.kt` line 70 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3743#-OrTo6W1AHIhCxBUn3RM:-OrTo6W1AHIhCxBUn3RN:b-uvxfqh) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1deca0a0ba84c3a8b98631460d23f4cac1108dee/src/test/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/FilterProcessorTest.kt#L70)):*
> ```Kotlin
>     ageGroup: Person.AgeGroup = Person.AgeGroup.YEARS_18_TO_34,
>     gender: Person.Gender = Person.Gender.MALE,
>     entityKeys: List<LabeledImpression.EntityKey> = emptyList(),
> ```

nit: no empty value here
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/LabeledEvent.kt` line 47 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3743#-OrTo0be0RN5lD6G5ewK:-OrTo0be0RN5lD6G5ewL:b-es0gd8) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1deca0a0ba84c3a8b98631460d23f4cac1108dee/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/LabeledEvent.kt#L47)):*
> ```Kotlin
>   val vid: Long,
>   val message: T,
>   val entityKeys: List<LabeledImpression.EntityKey> = emptyList(),
> ```

nit: ditto on no default value


<!-- Sent from Reviewable.io -->

Marco-Premier (commented):
I think this implementation now follows the suggestions 

> @Marco-Premier made 1 comment and resolved 3 discussions.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3743#-:-OrULbOT9ezJMjEU6ifh:b-old8le)* status: 2 of 17 files reviewed, all discussions resolved (waiting on stevenwarejones).

stevenwarejones (changes requested):
See #3744 that adds entity_keys to BlobDetails, the filtering in FilterProcessor can work at two levels — matching how event_group_reference_id works today:
                                                                                                                                                                                                                        
  Batch level (from BlobDetails.entity_keys): The EventBatch should carry the blob's EntityKeyGroup list (populated from BlobDetails when reading the blob). When entityKeys is the active selector, FilterProcessor can short-circuit an entire batch if none of the blob's entity key groups match the filter — same optimization as the current event_group_reference_id batch-level check.                                                
                                                                  
  Impression level (from LabeledImpression.entity_keys): Within a matching batch, the per-event entity key check stays as-is — each LabeledEvent carries its specific entity key(s) and must intersect the filter.      
                                                                  
  Concretely, what needs to change in this PR:                                                                                                                                                                          
                                                                  
  1. EventBatch — add an entityKeys field (populated from BlobDetails.entity_keys via StorageEventReader).                                                                                                              
  2. FilterProcessor.processBatch — when entityKeyFilterEnabled, do a batch-level check against EventBatch.entityKeys first (does any group in the batch match the filter?). If no match, short-circuit the whole batch.
   If it matches, fall through to the per-event LabeledImpression.entity_keys check.                                                                                                                                    
  3. FilterSpec.init — already addressed (either selector must be non-empty).
                                                                                                                                                                                                                        
  This gives you the same two-tier pattern as the existing event_group_reference_id path: fast batch rejection, then precise per-event filtering

> @stevenwarejones reviewed 4 files and all commit messages, and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3743#-:-OrUXSdk33ybZCnjItYx:b-vs8km0)* status: 6 of 17 files reviewed, 1 unresolved discussion (waiting on Marco-Premier).

___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/EventProcessingOrchestrator.kt` line 115 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3743#-OrULkQQ1F5-MxVokJBo:-OrULkQQ1F5-MxVokJBp:b8z1zvi) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/85fa2b426321390c7b2dfe745110be3bc59e0b77/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/EventProcessingOrchestrator.kt#L115)):*
> ```Kotlin
>             collectionInterval = firstEventGroup.value.collectionInterval,
>             eventGroupReferenceIds = eventGroupReferenceIds,
>             entityKeys = emptySet(),
> ```

no defaults please


<!-- Sent from Reviewable.io -->

stevenwarejones (commented):
The entity key filtering path still isn't wired up. PR #3744 added entity_keys to BlobDetails, but nothing downstream consumes it yet. Specifically:
                                                                                                                                                                                                                    
  1. EventBatch — still only has eventGroupReferenceId. Needs an entityKeys: List<EntityKeyGroup> field populated from BlobDetails.entity_keys via StorageEventReader.
  2. FilterSpec — still only accepts eventGroupReferenceIds. Needs an entityKeys selector, and init should require that one of the two selectors is non-empty (not both required).                                  
  3. FilterProcessor.processBatch — no batch-level entity key check. When entity keys are the active selector, it should short-circuit the batch if none of EventBatch.entityKeys match the filter, mirroring the   
  existing eventGroupReferenceIds check on line 92.                                                                                                                                                                 
  4. LabeledEvent — no per-event entity keys for the impression-level check within a matching batch.                                                                                                                
                                                                                                                                                                                                                    
  The two-tier pattern (fast batch rejection → per-event filtering) that works today for event_group_reference_id needs to be replicated for entity keys. Right now the proto field from #3744 is defined but dead —
   nothing reads it. 

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3743#-:-OrnXJUC6hDIDwbfNS_s:b-s56had)* status: 6 of 17 files reviewed, 1 unresolved discussion (waiting on Marco-Premier).



<!-- Sent from Reviewable.io -->

Marco-Premier (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3743)* status: 3 of 24 files reviewed, 1 unresolved discussion (waiting on stevenwarejones).

___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/EventProcessingOrchestrator.kt` line 115 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3743#-OrULkQQ1F5-MxVokJBo:-OrrACAY8fUXqCa9rI2g:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/85fa2b426321390c7b2dfe745110be3bc59e0b77/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/EventProcessingOrchestrator.kt#L115)):*

no defaults please
</blockquote></details>

Done.

stevenwarejones (commented):
The sealed-class FilterSpec is good, but fromRequisitions() still hardcodes ByEventGroupReferenceIds — there's no selection logic to decide which variant to use.
                                                                                                                                                                                                                    
  The rule is: look at all event groups in the requisition and pick the selector based on what's populated, with entity key taking precedence:
                                                                                                                                                                                                                    
  1. If any event group has entity_key set, all must have it (fail if mixed). Build FilterSpec.ByEntityKeys from the collected entity keys.                                                                         
  2. If none have entity_key, all must have a reference ID in the map (already enforced by getValue()). Build FilterSpec.ByEventGroupReferenceIds as today.                                                         
                                                                                                                                                                                                                    
  This is all-or-nothing across the event groups — not a per-group decision. Every sink ends up configured for one selector or the other, never a mix.                                                              
                                                                                                                                                                                                                    
  Concretely, fromRequisitions() needs:                                                                                                                                                                             
  - A new parameter providing entity keys per event group (e.g. eventGroupEntityKeyMap: Map<String, LabeledImpression.EntityKey>).
  - A check across all groups: val hasEntityKeys = eventGroups.any { eventGroupEntityKeyMap.containsKey(it.key) }.                                                                                                  
  - A validation that it's consistent: if hasEntityKeys, require all groups have an entity key; if not, require all have a reference ID.
  - Branch to construct the matching FilterSpec variant.                                                                                                                                                            
                                                                                                                                                                                                                    
  Tests should cover: all entity keys present, all reference IDs present, and the mixed case (should throw).  

> @stevenwarejones reviewed 3 files and all commit messages, made 2 comments, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3743#-:-OrsO6pE-tFEsSsgoUzA:b-now09i)* status: 6 of 24 files reviewed, 1 unresolved discussion (waiting on Marco-Premier).

___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/EventBatch.kt` line 36 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3743#-OrsMbY76Om73HIeL4GW:-OrsMbY76Om73HIeL4GX:b-awho3v) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/09c53fa25d2b87ffbd83ed67a7d98d687ba8ed5d/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/EventBatch.kt#L36)):*
> ```Kotlin
>  *   propagated from `BlobDetails.entity_keys`. Enables fast batch-level rejection in
>  *   `FilterProcessor` when entity keys are the active selector — same two-tier optimisation as the
>  *   existing `eventGroupReferenceId` path.
> ```

nit: over-specify what it contains... `will have at least one impression for each entity key. All impressions will be associated with at least one of the entity keys. Impressions may not necessarily be associated with all entity keys.` this is the updated language on the proto


<!-- Sent from Reviewable.io -->

Marco-Premier (commented):
should eventGroupEntityKeyMap come from a CMMS lookup the orchestrator does itself, or from a parameter passed into the orchestrator from above?

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3743#-:-Orsnnzf8vV3wuXRCElb:b-nrgrmr)* status: 6 of 24 files reviewed, 1 unresolved discussion (waiting on stevenwarejones).

stevenwarejones (commented):
Passed in as a parameter to fromRequisitions(), same pattern as eventGroupReferenceIdMap. The caller already has the EventGroup resources — extract entity keys there and pass them alongside the reference ID
  map. The orchestrator shouldn't make CMMS calls itself.

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3743#-:-OrswPD21aoKoKrP2TUh:b3ziaal)* status: 6 of 24 files reviewed, 1 unresolved discussion (waiting on Marco-Premier).

stevenwarejones (commented):
see https://github.com/world-federation-of-advertisers/cross-media-measurement/pull/3752

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3743#-:-Ort5HDoAkofJcCf-jnZ:bwerknw)* status: 6 of 24 files reviewed, 1 unresolved discussion (waiting on Marco-Premier).

stevenwarejones (changes requested):
> @stevenwarejones reviewed 21 files and all commit messages, made 10 comments, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3743)* status: all files reviewed, 10 unresolved discussions (waiting on Marco-Premier).

___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/FilterProcessor.kt` line 116 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3743#-OrxboGfBJ0d8pPMsQBp:-OrxboGfBJ0d8pPMsQBq:b-u53kyv) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/23843af6bc0c84f279d54f92cebe7e67a19bee46/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/FilterProcessor.kt#L116)):*
> ```Kotlin
>     }
> 
>     val entityKeyFilter: Set<LabeledImpressionEntityKey>? =
> ```

nit: this is redundant. you can combine with above
```
  val entityKeyFilter: Set<LabeledImpressionEntityKey>? =                                                                                                                                                     
    when (val spec = filterSpec) {                                                                                                                                                                            
      is FilterSpec.ByEventGroupReferenceIds -> {                                                                                                                                                             
        if (!spec.eventGroupReferenceIds.contains(batch.eventGroupReferenceId)) {                                                                                                                             
          return emptyBatchLike(batch)                                                                                                                                                                        
        }                                                                                                                                                                                                     
        null                                                                                                                                                                                                  
      }                                                                                                                                                                                                       
      is FilterSpec.ByEntityKeys -> {
        if (!batchEntityKeysOverlap(batch, spec.entityKeys)) {                                                                                                                                                
          return emptyBatchLike(batch)                            
        }                                                                                                                                                                                                     
        spec.entityKeys
      }                                                                                                                                                                                                       
    }       
```

___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/FilterProcessor.kt` line 205 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3743#-OrxbAI36v6EEGyR-iNd:-OrxbAI36v6EEGyR-iNe:boocios) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/23843af6bc0c84f279d54f92cebe7e67a19bee46/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/FilterProcessor.kt#L205)):*
> ```Kotlin
>    * A batch with empty `entityKeys` always fails this check.
>    */
>   private fun batchEntityKeysOverlap(
> ```

i think something like his would be faster for batches with a large number of entity keys vs a list contains

```
  private data class EntityKeyPair(val entityType: String, val entityId: String)                                                                                                                              
   
  private fun batchEntityKeysOverlap(                                                                                                                                                                         
      batch: EventBatch<T>,                                       
      filter: Set<LabeledImpressionEntityKey>,                                                                                                                                                                
    ): Boolean {                                                                                                                                                                                              
      if (batch.entityKeys.isEmpty()) return false                                                                                                                                                            
      val batchKeyPairs: Set<EntityKeyPair> =                                                                                                                                                                 
        batch.entityKeys.flatMap { g ->                                                                                                                                                                       
          g.entityIdsList.map { id -> EntityKeyPair(g.entityType, id) }                                                                                                                                       
        }.toSet()                                                                                                                                                                                             
      return filter.any { fk -> EntityKeyPair(fk.entityType, fk.entityId) in batchKeyPairs }                                                                                                                  
    }    
```

___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/FilterProcessor.kt` line 218 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3743#-OrxcCa_46E9T_yc36jq:-OrxcCa_46E9T_yc36jr:b7coxn1) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/23843af6bc0c84f279d54f92cebe7e67a19bee46/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/FilterProcessor.kt#L218)):*
> ```Kotlin
> }
> 
> private typealias LabeledImpressionEntityKey =
> ```

nit: Replace private typealias LabeledImpressionEntityKey = ... with a normal import and use LabeledImpression.EntityKey directly.  
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/FilterSpec.kt` line 56 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3743#-Orxc0DH5lTTkIZLsKMt:-Orxc0DH5lTTkIZLsKMu:b-8snep4) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/23843af6bc0c84f279d54f92cebe7e67a19bee46/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/FilterSpec.kt#L56)):*
> ```Kotlin
>       require(eventGroupReferenceIds.isNotEmpty()) { "eventGroupReferenceIds must not be empty" }
>       requireValidCollectionInterval(collectionInterval)
>     }
> ```

nit: put a message on the require
___
*[`src/test/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/FilterProcessorTest.kt` line 96 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3743#-OrxcY638qyHuU4BGoAp:-OrxcY638qyHuU4BGoAq:b-d0gm7s) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/23843af6bc0c84f279d54f92cebe7e67a19bee46/src/test/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/FilterProcessorTest.kt#L96)):*
> ```Kotlin
>       minTime,
>       maxTime,
>       eventGroupReferenceId = "test-group",
> ```

nit: pass this in the constructor for the function similar to entitykeys
___
*[`src/test/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/FilterProcessorTest.kt` line 116 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3743#-OrxcauX5D7Q22Dyfvzy:-OrxcauX5D7Q22Dyfvzz:beozfn5) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/23843af6bc0c84f279d54f92cebe7e67a19bee46/src/test/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/FilterProcessorTest.kt#L116)):*
> ```Kotlin
>     val startTime = Instant.parse("2050-01-01T00:00:00Z")
>     val endTime = Instant.parse("2000-01-01T00:00:00Z")
>     return Interval.newBuilder()
> ```

nit: use kotlin dsl
___
*[`src/test/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/FilterProcessorTest.kt` line 585 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3743#-OrxcqQREWQKXaPMXO7u:-OrxcqQREWQKXaPMXO7v:b6b4s12) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/23843af6bc0c84f279d54f92cebe7e67a19bee46/src/test/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/FilterProcessorTest.kt#L585)):*
> ```Kotlin
> 
>   @Test
>   fun `processBatch with entityKeys filter drops events whose entityKeys are empty`() {
> ```

this should throw an exception
___
*[`src/test/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/FilterProcessorTest.kt` line 679 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3743#-Orxd-0c7hvUCvrn7cKV:-Orxd-0c7hvUCvrn7cKW:bsm24tb) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/23843af6bc0c84f279d54f92cebe7e67a19bee46/src/test/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/FilterProcessorTest.kt#L679)):*
> ```Kotlin
> 
>   @Test
>   fun `FilterSpec ByEntityKeys init throws when entityKeys is empty`() {
> ```

nit: this is testing on the batch level, right? be a little more specific is this is batch level or event level
___
*[`src/test/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/FilterProcessorTest.kt` line 723 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3743#-Orxd8wLDMDNblOMS8C4:-Orxd8wMD1BhwJGioPbW:bifo2cj) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/23843af6bc0c84f279d54f92cebe7e67a19bee46/src/test/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/FilterProcessorTest.kt#L723)):*
> ```Kotlin
>     assertThat(exception)
>       .hasMessageThat()
>       .contains("collectionInterval startTime must be before endTime")
> ```

nit: we don't generally check messages. might be better to bubble up specific exception types.
___
*[`src/test/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/FilterProcessorTest.kt` line 764 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3743#-OrxdHbg9poO2xVMenKL:-OrxdHbg9poO2xVMenKM:bv92uz9) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/23843af6bc0c84f279d54f92cebe7e67a19bee46/src/test/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/FilterProcessorTest.kt#L764)):*
> ```Kotlin
> 
>   @Test
>   fun `processBatch with batch entityKeys not overlapping the filter short-circuits`() {
> ```

short-circuits? do you mean returns empty?


<!-- Sent from Reviewable.io -->

Marco-Premier (commented):
> @Marco-Premier made 6 comments and resolved 4 discussions.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3743)* status: 22 of 25 files reviewed, 6 unresolved discussions (waiting on stevenwarejones).

___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/FilterProcessor.kt` line 205 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3743#-OrxbAI36v6EEGyR-iNd:-Os0cYa_1KM50AcrWBea:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/23843af6bc0c84f279d54f92cebe7e67a19bee46/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/FilterProcessor.kt#L205)):*

i think something like his would be faster for batches with a large number of entity keys vs a list contains

```
  private data class EntityKeyPair(val entityType: String, val entityId: String)                                                                                                                              
   
  private fun batchEntityKeysOverlap(                                                                                                                                                                         
      batch: EventBatch<T>,                                       
      filter: Set<LabeledImpressionEntityKey>,                                                                                                                                                                
    ): Boolean {                                                                                                                                                                                              
      if (batch.entityKeys.isEmpty()) return false                                                                                                                                                            
      val batchKeyPairs: Set<EntityKeyPair> =                                                                                                                                                                 
        batch.entityKeys.flatMap { g ->                                                                                                                                                                       
          g.entityIdsList.map { id -> EntityKeyPair(g.entityType, id) }                                                                                                                                       
        }.toSet()                                                                                                                                                                                             
      return filter.any { fk -> EntityKeyPair(fk.entityType, fk.entityId) in batchKeyPairs }                                                                                                                  
    }
```

</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/FilterSpec.kt` line 56 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3743#-Orxc0DH5lTTkIZLsKMt:-Os0fSGY1end6KGLuZyX:b-yxqfyy) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/23843af6bc0c84f279d54f92cebe7e67a19bee46/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/FilterSpec.kt#L56)):*

nit: put a message on the require
</blockquote></details>

if you refer to `requireValidCollectionInterval` the message is present inside the function:

```
protected fun requireValidCollectionInterval(interval: Interval) {
    require(interval.startTime.toInstant().isBefore(interval.endTime.toInstant())) {
      "collectionInterval startTime must be before endTime"
    }
  }
``` 
___
*[`src/test/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/FilterProcessorTest.kt` line 585 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3743#-OrxcqQREWQKXaPMXO7u:-Os0ilE23uMn96JnVF0D:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/23843af6bc0c84f279d54f92cebe7e67a19bee46/src/test/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/FilterProcessorTest.kt#L585)):*

this should throw an exception
</blockquote></details>

Done.
___
*[`src/test/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/FilterProcessorTest.kt` line 679 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3743#-Orxd-0c7hvUCvrn7cKV:-Os0kI1s9eevBQN5sdOa:bn1q75x) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/23843af6bc0c84f279d54f92cebe7e67a19bee46/src/test/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/FilterProcessorTest.kt#L679)):*

nit: this is testing on the batch level, right? be a little more specific is this is batch level or event level
</blockquote></details>

this test is actually at neither the batch level nor the event level. It's exercising the_ **FilterSpec.ByEntityKeys** **constructor** itself, asserting the init  block rejects an empty selector set._
___
*[`src/test/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/FilterProcessorTest.kt` line 723 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3743#-Orxd8wLDMDNblOMS8C4:-Os0mJFiE1LEmvWYTycJ:b-8din87) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/23843af6bc0c84f279d54f92cebe7e67a19bee46/src/test/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/FilterProcessorTest.kt#L723)):*

nit: we don't generally check messages. might be better to bubble up specific exception types.
</blockquote></details>

This is thrown by `require(interval.startTime.toInstant().isBefore(interval.endTime.toInstant())) {` which only throws IllegalArgumentException. Are you suggesting to create a custom exception for it for the sake of the tests?
___
*[`src/test/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/FilterProcessorTest.kt` line 764 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3743#-OrxdHbg9poO2xVMenKL:-Os0mUTD3MhWThjQvtx1:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/23843af6bc0c84f279d54f92cebe7e67a19bee46/src/test/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/FilterProcessorTest.kt#L764)):*

short-circuits? do you mean returns empty?
</blockquote></details>

Done.

stevenwarejones (approved):
> @stevenwarejones reviewed 3 files and all commit messages, made 1 comment, and resolved 5 discussions.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3743)* status: all files reviewed, 1 unresolved discussion (waiting on Marco-Premier).

___
*[`src/test/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/FilterProcessorTest.kt` line 723 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3743#-Orxd8wLDMDNblOMS8C4:-Os21BDk0hTq-W1Dqepj:b-pou7js) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/23843af6bc0c84f279d54f92cebe7e67a19bee46/src/test/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/FilterProcessorTest.kt#L723)):*

This is thrown by `require(interval.startTime.toInstant().isBefore(interval.endTime.toInstant())) {` which only throws IllegalArgumentException. Are you suggesting to create a custom exception for it for the sake of the tests?
</blockquote></details>

yes - create a custom exception type

stevenwarejones (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3743)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on SanjayVas).

SanjayVas (approved):
> @SanjayVas reviewed 25 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3743)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on Marco-Premier).

stevenwarejones (changes requested):
The sealed class approach for FilterSpec is solid — making the selector variants type-level exclusive is the right call. The filtering logic in FilterProcessor is well-optimized (batch-level short-circuits, pre-flattened
  entity key sets). But I share your reservations about a few patterns, especially around the override/precedence mechanism.                                                                                                              
                                                                  
  The "override" pattern — entity keys vs reference IDs                                                                                                                                                                                   
                                                                  
  The pattern you're probably side-eyeing is in FilterSpecIndex.fromRequisitions:                                                                                                                                                         
  
  val hasEntityKeys = eventGroups.any { eventGroupEntityKeyMap.containsKey(it.key) }                                                                                                                                                      
                                                                                                                                                                                                                                          
  val filterSpec: FilterSpec =
    if (hasEntityKeys) {                                                                                                                                                                                                                  
      // ByEntityKeys path                                                                                                                                                                                                                
    } else {
      // ByEventGroupReferenceIds path (legacy fallback)                                                                                                                                                                                  
    }                                                                                                                                                                                                                                     
  
  This is a runtime precedence check on two parallel maps that both get threaded all the way from ResultsFulfiller → EventProcessingOrchestrator.run() → FilterSpecIndex.fromRequisitions. A few concerns:                                
                                                                  
  1. Two parallel maps for one concept. eventGroupReferenceIdMap and eventGroupEntityKeyMap are both built from the same eventGroupMapList, passed side-by-side through two layers, and then the selector variant decision happens deep   
  inside fromRequisitions. This means the caller builds both maps unconditionally, doesn't know which one will be used, and can't reason about it from the call site.
  2. Late consistency validation. The "all-or-nothing" rule (if any event group has entity keys, all must) is only enforced inside fromRequisitions with a requireNotNull after the any check. If RequisitionGrouper upstream populates   
  entity keys for 3 of 4 event groups, you get a runtime exception deep in the pipeline. This invariant should arguably be pushed closer to the data source — either at RequisitionGrouper time or at ResultsFulfiller map construction.  
  3. Implicit precedence. Entity keys silently win over reference IDs when both are present. The eventGroupReferenceIdMap is always built (it doesn't filter like eventGroupEntityKeyMap does), so a reference ID exists for every event
  group. If entity keys are present, the reference IDs are silently ignored. This is correct per the design, but it's implicit — a reader of ResultsFulfiller.fulfillRequisitions() sees both maps passed to orchestrator.run() and has to
   trace into FilterSpecIndex to understand that entity keys take precedence.
                                                                                                                                                                                                                                          
  What I'd consider instead: Rather than passing two parallel maps and deciding downstream, make the selector decision once at the ResultsFulfiller level and pass a single discriminated type:                                           
  
  sealed class EventGroupSelector {                                                                                                                                                                                                       
    data class ByReferenceIds(val map: Map<String, String>) : EventGroupSelector()                                                                                                                                                        
    data class ByEntityKeys(val map: Map<String, LabeledImpression.EntityKey>) : EventGroupSelector()                                                                                                                                     
  }                                                                                                                                                                                                                                       
                                                                                                                                                                                                                                          
  This pushes the precedence decision to the call site (where both maps are constructed and context is clear), eliminates the dead eventGroupReferenceIdMap when entity keys are present, and makes the fromRequisitions signature        
  unambiguous.  

> @stevenwarejones reviewed all commit messages and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3743#-:-OsNFMPFCW8Kamc8_YM1:b-908819)* status: 25 of 27 files reviewed, 1 unresolved discussion (waiting on Marco-Premier and SanjayVas).

___
*[`src/main/kotlin/org/wfanet/measurement/integration/common/InProcessEdpAggregatorComponents.kt` line 579 at r11](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3743#-OsNHu-ZCMyjVtW9S3Pg:-OsNHu-ZCMyjVtW9S3Ph:b-78563a) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7238c99e8e2f18efb3035436a862c0a5bc110692/src/main/kotlin/org/wfanet/measurement/integration/common/InProcessEdpAggregatorComponents.kt#L579)):*
> ```Kotlin
>       // default (campaign entity_type, empty entity_id) so the orchestrator's ByEntityKeys
>       // selector overlaps with the batch and the FilterProcessor guard does not trip.
>       val override: EventGroupEntityOverride? =
> ```

  Issues with the entityOverridesByEdp pattern                                                                                                                                                                                            
    
```                                                                                                                                                                                                                                      
  val override: EventGroupEntityOverride? =                       
      entityOverridesByEdp[edpAggregatorShortName]?.get(mappedEventGroup.eventGroupReferenceId)                                                                                                                                           
  val blobEntityKeys: List<EntityKey> =                                                                                                                                                                                                   
      listOf(                                                                                                                                                                                                                             
        EntityKey(                                                                                                                                                                                                                        
          entityType = override?.entityKey?.entityType ?: KINGDOM_DEFAULT_ENTITY_TYPE,                                                                                                                                                    
          entityId = override?.entityKey?.entityId.orEmpty(),                                                                                                                                                                             
        )                                                                                                                                                                                                                                 
      )                                                                                                                                                                                                                                 
```                                                                                                                                                                                                                                         
  1. Silent fabrication of phantom entity keys when no override exists.                                                                                                                                                                   
  When override is null (no override configured for this EDP + reference ID), the code doesn't skip or signal "no entity key" — it fabricates one: EntityKey("kingdom_default_type", ""). That empty-string entity ID is a real value that
   flows downstream into FilterSpec.ByEntityKeys matching. Every non-overridden event group ends up with the same synthetic entity key, so they'll all match each other in the entity-key intersection logic. This collapses distinct     
  event groups into one bucket.

  2. orEmpty() masks a null that should be an error.
  If override?.entityKey?.entityId is null, that means the override was configured but has no entity ID. That's almost certainly a config error. .orEmpty() silently converts it to "", producing the same phantom key as the no-override
  case. An override that maps to empty should fail loudly, not silently degrade to the default.

  3. The override is keyed by eventGroupReferenceId — the thing entity keys are replacing.
  The whole point of entity keys is to provide a new selector mechanism independent of reference IDs. But the override lookup uses reference IDs as the key:
  entityOverridesByEdp[edpAggregatorShortName]?.get(mappedEventGroup.eventGroupReferenceId). This creates a permanent dependency on the legacy system. You can't remove reference IDs without breaking the override mechanism.

  4. Always produces exactly one entity key (listOf(...)).
  The result is hardcoded to a single-element list. If an event group should have zero entity keys (no override, legacy mode) or multiple entity keys (composite entity), this code can't express either. The single-element list forces a
   1:1 mapping between event groups and entity keys, which the downstream FilterSpec.ByEntityKeys (using a Set) seems designed to avoid.

  - EDP not in the override map at all → no override configured (maybe intentional)
  - EDP present but reference ID not found → configuration gap (probably a bug)
  - Override exists but entityKey.entityType is null → broken override config (definitely a bug)

  All three silently produce KINGDOM_DEFAULT_ENTITY_TYPE. The second and third cases should probably throw.

  What I'd suggest

  If no override is configured, the entity key should be absent (null/empty list), not a fabricated default. Let the downstream code (like FilterSpecIndex.fromRequisitions) see that there's no entity key and select
  ByEventGroupReferenceIds for that event group. If an override is present but malformed, fail immediately. Something like:
```
  val override: EventGroupEntityOverride? =
      entityOverridesByEdp[edpAggregatorShortName]?.get(mappedEventGroup.eventGroupReferenceId)
  val blobEntityKeys: List<EntityKey> =
      if (override != null) {
        listOf(
          EntityKey(
            entityType = requireNotNull(override.entityKey?.entityType) {
              "Override for ${mappedEventGroup.eventGroupReferenceId} has no entityType"
            },
            entityId = requireNotNull(override.entityKey?.entityId) {
              "Override for ${mappedEventGroup.eventGroupReferenceId} has no entityId"
            },
          )
        )
      } else {
        emptyList()
      }
```
  This makes the no-override case explicit (empty list, no entity key), validates overrides when present, and keeps the downstream all-or-nothing consistency check in FilterSpecIndex honest — it sees real data rather than synthetic
  defaults.


<!-- Sent from Reviewable.io -->

stevenwarejones (commented):
Another issue based on other merged code
```
  // EventGroupReader.kt — always sets entityKey now                                                                                                                                                                                      
  entityKey = entityKey {                                                                                                                                                                                                                 
      entityType = struct.getString("EntityType")  // never null, defaults to "campaign"
      if (!struct.isNull("EntityId")) {                                                                                                                                                                                                   
        entityId = struct.getString("EntityId")                   
      }                                                                                                                                                                                                                                   
  }                                                               
```
  This means hasEntityKey() on the internal proto is now always true. In ResultsFulfiller, the filter that builds eventGroupEntityKeyMap:                                                                                                 
```  
  val eventGroupEntityKeyMap =                                                                                                                                                                                                            
      groupedRequisitions.eventGroupMapList                       
          .filter { it.details.hasEntityKey() }  // now always true
          .associate { ... }                                                                                                                                                                                                           
```
  ...will include every event group, even legacy ones that never had entity keys. Combined with the all-or-nothing check in FilterSpecIndex.fromRequisitions, this means FilterSpec.ByEntityKeys will be selected for all requisitions —  
  even ones where entity keys were never intentionally configured. And those phantom keys (entityType="campaign", no entityId) need to match something on the blob side, which brings us to MissingBatchEntityKeysException when blobs
  lack entity keys.                                                                                                                                                                                                                       
                                                                  
  This is worth flagging because it's an interaction between c1f1a49 (Kingdom defaults) and PR #3743 (filtering logic). Neither is wrong in isolation, but together they create a situation where legacy event groups silently switch     
  filtering strategies.

I suggest a one-line change — require a non-empty entityId to consider it a real entity key:                                                                                                                                                        

```                                                            
  val eventGroupEntityKeyMap: Map<String, LabeledImpression.EntityKey> =                                                                                                                                                                  
      groupedRequisitions.eventGroupMapList                                                                                                                                                                                               
          .filter { it.details.hasEntityKey() && it.details.entityKey.entityId.isNotEmpty() }
          .associate { ... }  
```
and add a test to breaks without this change and works with it.

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3743#-:-OsNKMR1-z0Jp0hH9UC6:b-u3261c)* status: 25 of 27 files reviewed, 1 unresolved discussion (waiting on Marco-Premier and SanjayVas).



<!-- Sent from Reviewable.io -->

Marco-Premier (commented):
Regarding the first comment:

**1\. Where should `sealed class EventGroupSelector` live?** Nested inside `FilterSpecIndex` (co-located with sole consumer), top-level in `EventProcessingOrchestrator.kt` (same file as the public `run(...)` API), or its own file? Lean nested-in-`FilterSpecIndex` — let me know if you'd prefer otherwise.

**2\. Where should the precedence decision land?** `ResultsFulfiller.fulfillRequisitions` (where both maps are currently built — smaller diff, contained scope) or further upstream in `RequisitionGrouper` / `RequisitionsValidator` (more correct, but expands scope to other consumers of those types)? Lean `ResultsFulfiller` for this PR.

**3\. Should I drop the defensive `requireNotNull` inside `fromRequisitions`?** With the discriminated type the asymmetric case is unrepresentable, and `RequisitionsValidator.kt:142-144` already enforces all-or-nothing upstream — so the inner check becomes dead code. Or keep it as belt-and-braces?

While regarding the "and add test to breaks without this change and works with it":

On the regression test for the `eventGroupEntityKeyMap` filter — your one-line change is inline at `ResultsFulfiller.kt:147` inside `fulfillRequisitions`. To test it I see two approaches, both with trade-offs — which do you prefer?

**A. Integration test in `ResultsFulfillerTest.kt`.** Construct a `groupedRequisitions` with a default-sentinel event group (`entityKey = (campaign, "")`), wire up the existing mocks (KMS, impression/requisition metadata services), call `fulfillRequisitions()` and assert it doesn't throw `MissingBatchEntityKeysException`. Real regression test, but ~100 lines of mock setup copied from the existing patterns in that file.

**B. Extract the filter into a `@VisibleForTesting` companion-object helper** (`buildEventGroupEntityKeyMap`) and unit-test it directly. ~10-line test with real regression value, but it's a small structural change beyond your "one-line" suggestion and adds a public testing surface to `ResultsFulfiller`.

Lean **B** for the maintenance cost, but happy to do **A** if you'd rather keep the production surface minimal. Let me know.

> @Marco-Premier made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3743#-:-OsSP75Z0U84lfs3o5WE:b-gkf6d3)* status: 24 of 27 files reviewed, 1 unresolved discussion (waiting on raimundoltdf, SanjayVas, and stevenwarejones).

___
*[`src/main/kotlin/org/wfanet/measurement/integration/common/InProcessEdpAggregatorComponents.kt` line 579 at r11](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3743#-OsNHu-ZCMyjVtW9S3Pg:-OsSRgBPDCqAWNX7oUlj:b-1r3bqy) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7238c99e8e2f18efb3035436a862c0a5bc110692/src/main/kotlin/org/wfanet/measurement/integration/common/InProcessEdpAggregatorComponents.kt#L579)):*

Issues with the entityOverridesByEdp pattern                                                                                                                                                                                            
    
```                                                                                                                                                                                                                                      
  val override: EventGroupEntityOverride? =                       
      entityOverridesByEdp[edpAggregatorShortName]?.get(mappedEventGroup.eventGroupReferenceId)                                                                                                                                           
  val blobEntityKeys: List<EntityKey> =                                                                                                                                                                                                   
      listOf(                                                                                                                                                                                                                             
        EntityKey(                                                                                                                                                                                                                        
          entityType = override?.entityKey?.entityType ?: KINGDOM_DEFAULT_ENTITY_TYPE,                                                                                                                                                    
          entityId = override?.entityKey?.entityId.orEmpty(),                                                                                                                                                                             
        )                                                                                                                                                                                                                                 
      )                                                                                                                                                                                                                                 
```                                                                                                                                                                                                                                         
  1. Silent fabrication of phantom entity keys when no override exists.                                                                                                                                                                   
  When override is null (no override configured for this EDP + reference ID), the code doesn't skip or signal "no entity key" — it fabricates one: EntityKey("kingdom_default_type", ""). That empty-string entity ID is a real value that
   flows downstream into FilterSpec.ByEntityKeys matching. Every non-overridden event group ends up with the same synthetic entity key, so they'll all match each other in the entity-key intersection logic. This collapses distinct     
  event groups into one bucket.

  2. orEmpty() masks a null that should be an error.
  If override?.entityKey?.entityId is null, that means the override was configured but has no entity ID. That's almost certainly a config error. .orEmpty() silently converts it to "", producing the same phantom key as the no-override
  case. An override that maps to empty should fail loudly, not silently degrade to the default.

  3. The override is keyed by eventGroupReferenceId — the thing entity keys are replacing.
  The whole point of entity keys is to provide a new selector mechanism independent of reference IDs. But the override lookup uses reference IDs as the key:
  entityOverridesByEdp[edpAggregatorShortName]?.get(mappedEventGroup.eventGroupReferenceId). This creates a permanent dependency on the legacy system. You can't remove reference IDs without breaking the override mechanism.

  4. Always produces exactly one entity key (listOf(...)).
  The result is hardcoded to a single-element list. If an event group should have zero entity keys (no override, legacy mode) or multiple entity keys (composite entity), this code can't express either. The single-element list forces a
   1:1 mapping between event groups and entity keys, which the downstream FilterSpec.ByEntityKeys (using a Set) seems designed to avoid.

  - EDP not in the override map at all → no override configured (maybe intentional)
  - EDP present but reference ID not found → configuration gap (probably a bug)
  - Override exists but entityKey.entityType is null → broken override config (definitely a bug)

  All three silently produce KINGDOM_DEFAULT_ENTITY_TYPE. The second and third cases should probably throw.

  What I'd suggest

  If no override is configured, the entity key should be absent (null/empty list), not a fabricated default. Let the downstream code (like FilterSpecIndex.fromRequisitions) see that there's no entity key and select
  ByEventGroupReferenceIds for that event group. If an override is present but malformed, fail immediately. Something like:
```
  val override: EventGroupEntityOverride? =
      entityOverridesByEdp[edpAggregatorShortName]?.get(mappedEventGroup.eventGroupReferenceId)
  val blobEntityKeys: List<EntityKey> =
      if (override != null) {
        listOf(
          EntityKey(
            entityType = requireNotNull(override.entityKey?.entityType) {
              "Override for ${mappedEventGroup.eventGroupReferenceId} has no entityType"
            },
            entityId = requireNotNull(override.entityKey?.entityId) {
              "Override for ${mappedEventGroup.eventGroupReferenceId} has no entityId"
            },
          )
        )
      } else {
        emptyList()
      }
```
  This makes the no-override case explicit (empty list, no entity key), validates overrides when present, and keeps the downstream all-or-nothing consistency check in FilterSpecIndex honest — it sees real data rather than synthetic
  defaults.
</blockquote></details>

I implemented your suggested code snippet character-for-character, which addresses the phantom-key fabrication, the `orEmpty()` masking, and the malformed-override-should-throw issues. Three of the concerns from your prose aren't covered by the snippet though, so flagging to confirm scope:

1.  1.
    
    **Override is keyed by `eventGroupReferenceId` (legacy coupling).** Decoupling this means changing the override map's key type and reworking how tests configure overrides — non-trivial. In scope for this PR or follow-up?
    
2.  2.
    
    **Hardcoded `listOf(...)` — can't express 0 or many entity keys.** I now produce `emptyList()` when no override is configured, but when an override exists I still produce exactly one element. Should `EventGroupEntityOverride.entityKey` become `List<EntityKey>` (or similar) so it can express composites? In scope for this PR or follow-up?
    
3.  3.
    
    **"EDP present but refId not found" should throw.** Currently both "EDP not in map" and "EDP in map but refId missing" silently route to `emptyList()`. To distinguish them I'd add a check like `if (edpOverrides != null) requireNotNull(edpOverrides[refId]) { … }`. Want me to add that here?


<!-- Sent from Reviewable.io -->

stevenwarejones (changes requested):
ResultsFulfiller — tests for how eventGroupEntityKeyMap is constructed from groupedRequisitions:                                                                                                                                    
  1. All event groups have Kingdom-default entity keys (entityType="campaign", entityId="") → entityId.isNotEmpty() filter produces empty map → ByEventGroupReferenceIds selected
  2. All event groups have real entity keys (both entityType and entityId populated) → map is fully populated → ByEntityKeys selected                                                                                                 
  3. Mix of real and default entity keys → only real ones in map → fromRequisitions hits all-or-nothing check and throws             
                                                                                                                                                                                                                                      
  EventProcessingIntegrationTest — currently every test passes eventGroupEntityKeyMap = emptyMap(), so there's zero coverage of the ByEntityKeys path through FilterSpecIndex.fromRequisitions. Need:                                 
  1. Entity-keys-only: all event groups have entity keys with matching blob-level entity keys → ByEntityKeys produces correct frequency vectors                                                                                       
  2. Mixed: some event groups have entity keys, some don't → fromRequisitions throws from the all-or-nothing consistency check

> @stevenwarejones reviewed 3 files and all commit messages, made 2 comments, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3743#-:-OsU5d0G3Qo1li24ywB4:b352won)* status: all files reviewed, 1 unresolved discussion (waiting on Marco-Premier, raimundoltdf, and SanjayVas).

___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/ResultsFulfiller.kt` line 145 at r12](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3743#-OsSU6Zb9biViX6rTZ8j:-OsSU6Zb9biViX6rTZ8k:b-4vjvvc) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/6de286371b8f00c15e45514eff50db8aa13a8702/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/ResultsFulfiller.kt#L145)):*
> ```Kotlin
>     val eventGroupEntityKeyMap: Map<String, LabeledImpression.EntityKey> =
>       groupedRequisitions.eventGroupMapList
>         .filter { it.details.hasEntityKey() && it.details.entityKey.entityId.isNotEmpty() }
> ```

  nit: Worth a short comment here explaining why we require non-empty entityId — the Kingdom schema defaults entity_type to "campaign" for all rows, so hasEntityKey() alone is always true. The entityId check distinguishes explicitly configured entity keys from schema defaults.


<!-- Sent from Reviewable.io -->

stevenwarejones (commented):
For the EventGroupSelector sealed class:
                                                                                                                                                                                                                                      
  1. Location: Agree with nested-in-FilterSpecIndex. It's the sole consumer, and if it ever needs broader visibility you can promote it later. No need to create a file for a two-variant sealed class.
  2. Precedence decision: Agree with ResultsFulfiller for this PR. Moving it upstream to RequisitionGrouper is architecturally cleaner but increases scope and risk. Do it in ResultsFulfiller now, file an issue to push it upstream 
  later if the team wants that.                                                                                                                                                                                                       
  3. requireNotNull inside fromRequisitions: Keep it. The upstream validation and the type-level guarantee make it technically dead code, but it's a one-line safety net at a system boundary (encrypted requisition specs are being  
  unpacked). If someone adds a new caller of fromRequisitions that doesn't go through RequisitionsValidator, the check catches it. Belt-and-braces is the right call here — remove it when there's a reason to, not preemptively.     
                                                                  
  For the regression test: Go with B. Extracting buildEventGroupEntityKeyMap as a @VisibleForTesting companion helper is a minimal structural change and gives you a focused unit test that directly exercises the                    
  entityId.isNotEmpty() guard. The mock setup cost of option A isn't worth it for testing a map filter. The testing surface it adds to ResultsFulfiller is a single pure function with no side effects — that's fine.

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3743#-:-OsU5r4MA_7c6mV5cQwS:bizstrw)* status: all files reviewed, 1 unresolved discussion (waiting on Marco-Premier, raimundoltdf, and SanjayVas).



<!-- Sent from Reviewable.io -->

stevenwarejones (commented):
for the two tests i mentioned, we should add test coverage for the entity key path in those two places.

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3743#-:-OsU60Cq1N4Vw6tr3EE6:b-une0up)* status: all files reviewed, 1 unresolved discussion (waiting on Marco-Premier, raimundoltdf, and SanjayVas).

Marco-Premier (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3743#-:-OsWFn4iEOL0TrPj48y9:bnfp4nl)* status: 23 of 27 files reviewed, 1 unresolved discussion (waiting on raimundoltdf, SanjayVas, and stevenwarejones).

stevenwarejones (changes requested):
> @stevenwarejones reviewed 4 files and all commit messages, made 1 comment, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3743)* status: all files reviewed, 1 unresolved discussion (waiting on Marco-Premier, raimundoltdf, and SanjayVas).

___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/ResultsFulfiller.kt` line 43 at r13](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3743#-OsWQkhY8yDYJhvTgpG5:-OsWQkhY8yDYJhvTgpG6:bhdiezq) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/da431b6909a31c10f8467b1a86a28d0989d0ba1e/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/ResultsFulfiller.kt#L43)):*
> ```Kotlin
> import kotlinx.coroutines.flow.toList
> import kotlinx.coroutines.withContext
> import org.jetbrains.annotations.VisibleForTesting
> ```

use guava: `com.google.common.annotations.VisibleForTesting`

stevenwarejones (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3743)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on raimundoltdf and SanjayVas).

___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/ResultsFulfiller.kt` line 43 at r13](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3743#-OsWQkhY8yDYJhvTgpG5:-OsWRPpCC5CtUG-AxCIW:b-tz0cv2) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/da431b6909a31c10f8467b1a86a28d0989d0ba1e/src/main/kotlin/org/wfanet/measurement/edpaggregator/resultsfulfiller/ResultsFulfiller.kt#L43)):*

use guava: `com.google.common.annotations.VisibleForTesting`
</blockquote></details>

nvm - not i guess we use the jetbrains one

Marco-Premier (commented):
> @Marco-Premier reviewed 11 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3743)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on raimundoltdf and SanjayVas).

PR #3799, opened 2026-05-18
fix: synchronous flush and batch getBlob for BufferedDataAvailabilityCleanup

### LINKED ISSUE #3798: fix: BufferedDataAvailabilityCleanup flush never fires on Cloud Run
## Problem

BufferedDataAvailabilityCleanup uses a ScheduledExecutorService background timer to periodically flush queued delete events. On Cloud Run (Gen2 Cloud Functions), instances are killed as soon as all HTTP requests complete, so the background timer thread never fires.

Buffered delete events are silently lost when the instance shuts down.

Additionally, each metadata blob corresponding impression blob is checked via individual getBlob() calls in DataAvailabilitySync, creating O(N) storage RPCs per sync cycle.

## Fix

- Replace ScheduledExecutorService with synchronous lock-based flush
- Every enqueue() flushes synchronously under a ReentrantLock
- Concurrent requests coalesce: first thread drains queue, others find it empty
- Batch-resolve blob URIs via ListImpressionMetadata with blobUris filter
- Cap batch delete RPCs at 100 records (configurable via batchSize)
- NOT_FOUND fallback: if batch delete fails, retry individually
- Replace N individual getBlob() calls with batch prefix listing in DataAvailabilitySync

stevenwarejones:
## Summary

Fixes #3798

- Replace ScheduledExecutorService background timer with synchronous lock-based flush in BufferedDataAvailabilityCleanup. Cloud Run kills instances when all HTTP requests complete, so background timers never fire.
- Every enqueue() call flushes synchronously under a ReentrantLock. Concurrent requests coalesce naturally (first thread drains queue, others find it empty).
- Batch-resolve blob URIs via ListImpressionMetadata with blobUris filter instead of individual lookups.
- Cap batch delete RPCs at configurable batchSize (default 100). Add NOT_FOUND fallback with individual delete retry.
- Replace N individual getBlob() calls with batch prefix listing in DataAvailabilitySync, reducing O(N) to O(K) storage RPCs.

## Test plan

- [x] Unit tests: BufferedDataAvailabilityCleanupTest (sequential + concurrent flush)
- [x] Integration tests: DataAvailabilityCleanupFunctionTest (unbuffered, buffered sequential, buffered concurrent)
- [x] Cloud test: 1330 campaigns (2660 GCS objects) through full EventArc pipeline with batch_size=100

### DIFF
--- a/src/main/kotlin/org/wfanet/measurement/edpaggregator/dataavailability/BUILD.bazel
+++ b/src/main/kotlin/org/wfanet/measurement/edpaggregator/dataavailability/BUILD.bazel
@@ -9,6 +9,7 @@ package(default_visibility = [
 kt_jvm_library(
     name = "data_availability_sync",
     srcs = [
+        "BufferedDataAvailabilityCleanup.kt",
         "DataAvailabilityCleanup.kt",
         "DataAvailabilityCleanupMetrics.kt",
         "DataAvailabilitySync.kt",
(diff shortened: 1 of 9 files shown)


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3799)

PR #3801, opened 2026-05-18
fix: batch getBlob in DataAvailabilitySync

### LINKED ISSUE #3800: fix: DataAvailabilitySync makes O(N) individual getBlob calls per sync cycle
## Problem

`DataAvailabilitySync.createModelLineToImpressionMetadataMap()` checks whether each impression blob exists by calling `storageClient.getBlob()` individually for every metadata file. With N metadata files, this creates O(N) storage RPCs per sync cycle.

## Fix

Replace N individual `getBlob()` calls with a two-pass approach:
1. Parse all metadata to extract unique impression blob directory prefixes
2. Batch-list those prefixes via `listBlobs()` to get all existing blob keys
3. Check set membership instead of individual RPCs

This reduces storage RPCs from O(N) to O(K) where K is the number of unique directory prefixes (typically 1 per sync cycle).

stevenwarejones:
## Summary

Fixes #3800

Replace O(N) individual `getBlob()` calls with batch prefix listing in `DataAvailabilitySync.createModelLineToImpressionMetadataMap()`. Parses all metadata to extract unique impression blob directory prefixes, batch-lists via `listBlobs()`, then checks set membership. Reduces storage RPCs from O(N) to O(K) where K is the number of unique directory prefixes (typically 1).

## Test plan

- [x] Existing DataAvailabilitySync tests pass (logic is equivalent)
- [x] Cloud-tested with 1330 campaigns through full pipeline

### DIFF
--- a/src/main/kotlin/org/wfanet/measurement/edpaggregator/dataavailability/DataAvailabilitySync.kt
+++ b/src/main/kotlin/org/wfanet/measurement/edpaggregator/dataavailability/DataAvailabilitySync.kt
@@ -29,8 +29,6 @@ import java.util.logging.Logger
 import kotlin.text.Charsets.UTF_8
 import kotlin.time.TimeSource
 import kotlinx.coroutines.ExperimentalCoroutinesApi
-import kotlinx.coroutines.flow.Flow
-import kotlinx.coroutines.flow.filter
 import kotlinx.coroutines.flow.toList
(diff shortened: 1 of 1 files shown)


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3801)

Issue #3790, opened 2026-05-15
Automate dependency updates


laureanobrs:
## Context

As part of the dependency update automation work, we are going to use Renovate.

In the first stage, Renovate will run in Silent mode. This means that dependency update PRs need to be triggered manually from the Developer Portal:

https://developer.mend.io/github/world-federation-of-advertisers/cross-media-measurement

## Plan

We will iterate on the Renovate configuration to cover the different dependency types, version constraints, dependency grouping into a single PR, immortal PRs, and other related behavior.

## Initial scope

The first stage of testing and configuration is focused on integrating Renovate and updating the GitHub Actions used in our workflows.

- https://github.com/world-federation-of-advertisers/cross-media-measurement/pull/3791
- https://github.com/world-federation-of-advertisers/cross-media-measurement/pull/3805
- https://github.com/world-federation-of-advertisers/cross-media-measurement/pull/3806
- https://github.com/world-federation-of-advertisers/cross-media-measurement/pull/3796
- https://github.com/world-federation-of-advertisers/actions/pull/52
- https://github.com/world-federation-of-advertisers/cross-media-measurement/pull/3807

PR #3805, opened 2026-05-19
build: Set both tag and digest in oci.pull rules

laureanobrs:
This PR sets the tag, in addition to the digest, in the oci_pull usages. This is only allowed if `reproducible = True` is also declared; see https://github.com/bazel-contrib/rules_oci/pull/587.

This allows tools like Renovate to correctly update the digests.

Issue: #3790

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -389,14 +389,16 @@ use_repo(pip, "pip")
 oci = use_extension("@rules_oci//oci:extensions.bzl", "oci")
 oci.pull(
     name = "go_image_base",
-    # Digest of `nonroot` tag.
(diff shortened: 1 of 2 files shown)


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3805)

SanjayVas (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3805)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on laureanobrs).

PR #3809, opened 2026-05-19
fix: exclude multi-entity-key event group from direct RF/reach tests

### LINKED ISSUE #3808: Direct RF integration test flaky due to incorrect entity key chunking hack
## Bug

`GCloudEdpAggregatorLifeOfAMeasurementIntegrationTest` "create a direct RF measurement" test fails intermittently with reach values approximately 2x the expected value (e.g., expected 993, actual 2036).

## Root Cause

The integration test used a `subList(0, chunkSize)` hack to approximate entity key behavior for multi-entity-key event groups. This hack split events by position — e.g., with 2000 events and 2 entity keys, events 0–999 were tagged with key A and events 1000–1999 with key B.

Both the blob writer (`InProcessEdpAggregatorComponents`) and the expected-value calculator (`EdpAggregatorMeasurementConsumerSimulator`) used this positional split. However, the EDPA's `FilterProcessor` performs real per-event entity key intersection filtering, which didn't match the hack's assumptions.

The test was **flaky** rather than always-failing because the measurement protocol adds `CONTINUOUS_GAUSSIAN` differential privacy noise. The expected value was consistently wrong, but random DP noise sometimes shifted the actual result close enough to fall within the tolerance window (pass) and sometimes pushed it further away (fail).

## Fix

PR #3810: Remove the chunking hack from both places. Tag all events with all entity keys in the blob (matching real EDPA data generation via `GenerateSyntheticData`) and use all events for expected value calculation. Multi-entity-key event groups still exercise two different entity keys in the same blob file.

stevenwarejones:
## Summary

- Add `eventGroupFilter` to direct RF and reach-only tests to exclude the multi-entity-key event group
- All other tests in `InProcessEdpAggregatorLifeOfAMeasurementIntegrationTest` already exclude this event group; these two were the only ones missing it
- Without the filter, non-deterministic event group selection can pick the multi-entity-key group, causing expected/actual value divergence

Fixes #3808

## Test plan

- [x] `GCloudEdpAggregatorLifeOfAMeasurementIntegrationTest` passes 3/3 runs with `--runs_per_test=3`

### DIFF
--- a/src/main/kotlin/org/wfanet/measurement/integration/common/InProcessEdpAggregatorLifeOfAMeasurementIntegrationTest.kt
+++ b/src/main/kotlin/org/wfanet/measurement/integration/common/InProcessEdpAggregatorLifeOfAMeasurementIntegrationTest.kt
@@ -306,7 +306,11 @@ abstract class InProcessEdpAggregatorLifeOfAMeasurementIntegrationTest(
       // Use frontend simulator to create a direct reach and frequency measurement and verify
       // its
       // result.
-      mcSimulator.testDirectReachAndFrequency(runId = "1234", numMeasurements = 1)
+      mcSimulator.testDirectReachAndFrequency(
+        runId = "1234",
+        numMeasurements = 1,
+        eventGroupFilter = { it.eventGroupReferenceId != MULTI_ENTITY_KEY_EVENT_GROUP_REF_ID },
+      )
     }
 
   @Test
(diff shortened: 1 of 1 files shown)


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3809)

PR #3815, opened 2026-05-20
fix: move Gaussian noise EDP restriction tests to no-noise subclass

### LINKED ISSUE #3814: Gaussian noise EDP restriction tests fail in noise subclasses
The Gaussian noise EDP restriction tests (HMSS/TrusTee no noise basic report fails when EDP requires Gaussian noise) are defined in the abstract base class InProcessEdpAggregatorLifeOfAReportTest, which means they run in every subclass implementation. These tests should not be in the abstract class — they only apply to implementations that use NoiseMechanism.NONE in their protocol config.

In subclasses that configure the protocol with CONTINUOUS_GAUSSIAN, the EDP noise restriction is satisfied, the report correctly succeeds, and the assertion expecting FAILED breaks. The tests are not universally valid across all implementations.

Fix: extract the test bodies into protected helper methods in the base class, and add @Test methods only in the subclass implementations where NoiseMechanism.NONE is configured.

stevenwarejones:
Fixes #3814

## Summary
- The Gaussian noise EDP restriction tests should not be in the abstract base class — they only apply to implementations that use NoiseMechanism.NONE in their protocol config. In implementations that use CONTINUOUS_GAUSSIAN, the EDP restriction is satisfied and the report correctly succeeds, so the tests are not valid there.
- Extract the test bodies into protected helper methods in InProcessEdpAggregatorLifeOfAReportTest so implementations can opt in
- Add @Test methods in GCloudEdpAggregatorLifeOfAReportTest (which uses NoiseMechanism.NONE) to exercise the restriction

## Test plan
- [x] Both tests pass in GCloudEdpAggregatorLifeOfAReportTest (no-noise config)
- [x] Build succeeds for all subclasses

Issue: #3814

### DIFF
--- a/src/main/kotlin/org/wfanet/measurement/integration/common/reporting/v2/InProcessEdpAggregatorLifeOfAReportTest.kt
+++ b/src/main/kotlin/org/wfanet/measurement/integration/common/reporting/v2/InProcessEdpAggregatorLifeOfAReportTest.kt
@@ -636,8 +636,7 @@ abstract class InProcessEdpAggregatorLifeOfAReportTest(
     )
   }
 
-  @Test
-  fun `HMSS no noise basic report fails when EDP requires Gaussian noise`() = runBlocking {
+  protected suspend fun assertHmssReportFailsWhenEdpRequiresGaussianNoise() {
     val hmssEventGroups = getHmssEventGroupsIncludingRestrictedEdp()
     check(hmssEventGroups.size > 1)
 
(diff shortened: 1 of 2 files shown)


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3815)

SanjayVas (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3815)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on stevenwarejones).

PR #3818, opened 2026-05-21
feat: Add RawImpressionUpload proto definitions for internal and public API

### LINKED ISSUE #3880: Add RawImpressionUpload proto and gRPC service definitions
Add proto message and service definitions for `RawImpressionUpload`, the top-level parent resource for the VID labeling pipeline. One `RawImpressionUpload` is created per `DataProvider` impression-upload event.

Scope:
- Public proto under `src/main/proto/wfa/measurement/edpaggregator/v1alpha/`:
  - `raw_impression_upload.proto` — resource message with `State` enum, `done_blob_uri`, lifecycle timestamps.
  - `raw_impression_upload_service.proto` — `RawImpressionUploadService` with `CreateRawImpressionUpload`, `GetRawImpressionUpload`, `ListRawImpressionUploads` plus any state-transition RPCs (AIP-216).
- Internal proto under `src/main/proto/wfa/measurement/internal/edpaggregator/`:
  - `raw_impression_upload.proto` — internal representation using `data_provider_resource_id` / `raw_impression_upload_resource_id`.
  - `raw_impression_upload_state.proto` — top-level `RawImpressionUploadState` enum.
  - `raw_impression_upload_service.proto` — internal `RawImpressionUploadService` mirroring the public surface.
- `BUILD.bazel` updates in both packages.

Conventions per parent #3879:
- Public ↔ internal parity on Filter fields, request fields, RPCs, resource fields, and State enums.
- Any transitional State value (i.e. not the initial-on-Create and not DELETED) requires a dedicated transition RPC.
- AIPs are authoritative over local precedent.

Tracked by PR #3818.

getina:
## Summary
- Add proto message and service definitions for RawImpressionUpload, the top-level parent resource for the new VID labeling pipeline.
- Includes create, get, list, and state transition (Mark Active/Completed/Failed) methods for both internal and public (v1alpha) APIs.
- Both internal and public proto definitions in a single PR per task guidance.
- No delete RPC per design doc (parent table uses cascade delete at the DB level).

## Test plan
- bazel build passes for all proto targets (internal message, internal service, internal state, public message, public service)

Closes: #3880 
Issue: #3880


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3818)

stevenwarejones (changes requested):
The existing RawImpressionMetadataBatch carries a State field. The new design moves state to the child RawImpressionUploadModelLine table, which makes sense since state is now per-model-line. But this PR ships only the parent
  resource with no state and no child service, so there's no way to query whether an upload is processing, complete, or failed.

  Can we either (a) add a denormalized State to RawImpressionUpload that rolls up from its children, or (b) land the RawImpressionUploadModelLine child resource/service in this same PR so callers can query state at that level?
  Shipping the parent alone without state is a regression in observability vs RawImpressionMetadataBatch.

  What's the plan for RawImpressionUploadFile and RawImpressionUploadModelLine services? If they're follow-up PRs, can you add tracking issues?

> @stevenwarejones made 5 comments.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3818#-:-OtFi4PmEn0HBl90TS4c:bh9noqt)* status: 0 of 6 files reviewed, 4 unresolved discussions (waiting on getina and Marco-Premier).

___
*[`src/main/proto/wfa/measurement/edpaggregator/v1alpha/raw_impression_upload_service.proto` line 56 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3818#-OtFhp6_8AP5eXgaZ7e_:-OtFhp6_8AP5eXgaZ7ea:bano094) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ed7d520fa11ab199be76b7935ffa2c7bf5def26a/src/main/proto/wfa/measurement/edpaggregator/v1alpha/raw_impression_upload_service.proto#L56)):*
> ```proto
>   // The RawImpressionUpload to create.
>   // The `name` and output-only fields are ignored.
>   RawImpressionUpload raw_impression_upload = 2
> ```

  The resource has no client-settable fields besides name (which is IDENTIFIER and ignored on create). This means CreateRawImpressionUploadRequest.raw_impression_upload is effectively empty. Is there anything the caller should set
  at create time -- e.g., the done-blob URI that triggered the upload, or a dispatch ID for correlation? If not, consider whether the raw_impression_upload body field is needed on Create at all vs just parent + request_id.
___
*[`src/main/proto/wfa/measurement/edpaggregator/v1alpha/raw_impression_upload_service.proto` line 96 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3818#-OtFhc8k7Wm4-Ko9xNDB:-OtFhc8k7Wm4-Ko9xNDC:bcq4vqc) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ed7d520fa11ab199be76b7935ffa2c7bf5def26a/src/main/proto/wfa/measurement/edpaggregator/v1alpha/raw_impression_upload_service.proto#L96)):*
> ```proto
> 
>   // A page token, received from a previous call.
>   string page_token = 3 [(google.api.field_behavior) = OPTIONAL];
> ```

  No filter support on List. The existing ListImpressionMetadataRequest has a structured Filter message with typed fields (model_line, state, time interval, etc.). Without at least a time-range filter, this RPC isn't useful for
  operators or the dispatcher to find recent uploads. Even if state isn't on this resource yet, a create_time interval filter is the minimum.
___
*[`src/main/proto/wfa/measurement/internal/edpaggregator/raw_impression_upload.proto` line 30 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3818#-OtFhBTW71tLZbUsOE_P:-OtFhBTW71tLZbUsOE_Q:b-if1n3p) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ed7d520fa11ab199be76b7935ffa2c7bf5def26a/src/main/proto/wfa/measurement/internal/edpaggregator/raw_impression_upload.proto#L30)):*
> ```proto
> // impressions and writes a "done" blob, which triggers the
> // VidLabelingDispatcher. All downstream pipeline state cascades from here.
> message RawImpressionUpload {
> ```

Missing create_request_id field. The existing ImpressionMetadata Spanner row persists the create request ID for idempotency enforcement at the DB layer. The request_id on the Create RPC is enough for the service layer, but the
  server needs to store it on the resource to detect duplicate creates across retries. See the existing CreateRequestId column on ImpressionMetadata and RawImpressionMetadataBatch.
___
*[`src/main/proto/wfa/measurement/internal/edpaggregator/raw_impression_upload_service.proto` line 47 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3818#-OtFhOwtEEgJ4FXKkpTQ:-OtFhOwtEEgJ4FXKkpTR:b6m3vbe) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ed7d520fa11ab199be76b7935ffa2c7bf5def26a/src/main/proto/wfa/measurement/internal/edpaggregator/raw_impression_upload_service.proto#L47)):*
> ```proto
> 
>   // Resource ID for the new upload.
>   string raw_impression_upload_id = 2;
> ```

  data_provider_resource_id (field 1) and raw_impression_upload_id (field 2) are redundant with the same fields inside the nested raw_impression_upload message (field 3). The existing internal Create pattern puts the parent ID on
  the request and lets the server generate the resource ID -- see CreateImpressionMetadataRequest. Suggest dropping field 2 and having the server generate it, consistent with the existing IdGenerator pattern. If the dispatcher
  needs deterministic IDs, document why.

Marco-Premier (changes requested):
> @Marco-Premier reviewed 6 files and all commit messages, and made 6 comments.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3818)* status: all files reviewed, 6 unresolved discussions (waiting on getina and stevenwarejones).

___
*[`src/main/proto/wfa/measurement/edpaggregator/v1alpha/raw_impression_upload_service.proto` line 56 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3818#-OtFhp6_8AP5eXgaZ7e_:-Ou7mbcgABSsFsU3Tcmt:bb14jhd) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ed7d520fa11ab199be76b7935ffa2c7bf5def26a/src/main/proto/wfa/measurement/edpaggregator/v1alpha/raw_impression_upload_service.proto#L56)):*

The resource has no client-settable fields besides name (which is IDENTIFIER and ignored on create). This means CreateRawImpressionUploadRequest.raw_impression_upload is effectively empty. Is there anything the caller should set
  at create time -- e.g., the done-blob URI that triggered the upload, or a dispatch ID for correlation? If not, consider whether the raw_impression_upload body field is needed on Create at all vs just parent + request_id.
</blockquote></details>

as per comment above I am happy to add the done blob uri
___
*[`src/main/proto/wfa/measurement/edpaggregator/v1alpha/raw_impression_upload_service.proto` line 96 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3818#-OtFhc8k7Wm4-Ko9xNDB:-Ou7nMYU4Uhdigh0Vsj6:b-i1royd) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ed7d520fa11ab199be76b7935ffa2c7bf5def26a/src/main/proto/wfa/measurement/edpaggregator/v1alpha/raw_impression_upload_service.proto#L96)):*

No filter support on List. The existing ListImpressionMetadataRequest has a structured Filter message with typed fields (model_line, state, time interval, etc.). Without at least a time-range filter, this RPC isn't useful for
  operators or the dispatcher to find recent uploads. Even if state isn't on this resource yet, a create_time interval filter is the minimum.
</blockquote></details>

TODO: add a filter that let query by `RawImpressionUpload` and / or `create_time`
___
*[`src/main/proto/wfa/measurement/internal/edpaggregator/raw_impression_upload.proto` line 30 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3818#-OtFhBTW71tLZbUsOE_P:-Ou7oJtB9PHGLITbmK0o:bz9mf6f) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ed7d520fa11ab199be76b7935ffa2c7bf5def26a/src/main/proto/wfa/measurement/internal/edpaggregator/raw_impression_upload.proto#L30)):*

Missing create_request_id field. The existing ImpressionMetadata Spanner row persists the create request ID for idempotency enforcement at the DB layer. The request_id on the Create RPC is enough for the service layer, but the
  server needs to store it on the resource to detect duplicate creates across retries. See the existing CreateRequestId column on ImpressionMetadata and RawImpressionMetadataBatch.
</blockquote></details>

Same comment on public api:

Let's add 2 new fields:

```
RawImpressionUpload {
    STATE_UNSPECIFIED = 0;
    CREATED = 1; // the upload was created but model line are not processing yet
    ACTIVE = 2;      // at least one model line is still processing
    COMPLETED = 3;   // all model lines completed successfully
    FAILED = 4;      // at least one model line failed
  }
```

```
string done_blob_uri
```

As per @stevenwarejones  comment the request id is set in the `CreateRawImpressionUploadRequest` so there is no need to add it here.
___
*[`src/main/proto/wfa/measurement/internal/edpaggregator/raw_impression_upload_service.proto` line 47 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3818#-OtFhOwtEEgJ4FXKkpTQ:-Ou7oxdz8HaRdq7hRu8i:b-ccatjx) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ed7d520fa11ab199be76b7935ffa2c7bf5def26a/src/main/proto/wfa/measurement/internal/edpaggregator/raw_impression_upload_service.proto#L47)):*

data_provider_resource_id (field 1) and raw_impression_upload_id (field 2) are redundant with the same fields inside the nested raw_impression_upload message (field 3). The existing internal Create pattern puts the parent ID on
  the request and lets the server generate the resource ID -- see CreateImpressionMetadataRequest. Suggest dropping field 2 and having the server generate it, consistent with the existing IdGenerator pattern. If the dispatcher
  needs deterministic IDs, document why.
</blockquote></details>

+1 drop the upload it and resource id
___
*[`src/main/proto/wfa/measurement/edpaggregator/v1alpha/raw_impression_upload.proto` line 41 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3818#-Ou7lLYK91OQ5DD3RISl:-Ou7lLYK91OQ5DD3RISm:b51okhj) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ed7d520fa11ab199be76b7935ffa2c7bf5def26a/src/main/proto/wfa/measurement/edpaggregator/v1alpha/raw_impression_upload.proto#L41)):*
> ```proto
>   // Resource name.
>   string name = 1 [(google.api.field_behavior) = IDENTIFIER];
> ```

Let's add 2 new fields:

```
RawImpressionUpload {
    STATE_UNSPECIFIED = 0;
    CREATED = 1; // the upload was created but model line are not processing yet
    ACTIVE = 2;      // at least one model line is still processing
    COMPLETED = 3;   // all model lines completed successfully
    FAILED = 4;      // at least one model line failed
  }
```

```
string done_blob_uri
```
___
*[`src/main/proto/wfa/measurement/internal/edpaggregator/raw_impression_upload_service.proto` line 82 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3818#-Ou7oeZr7yusrXv4KTM_:-Ou7oeZr7yusrXv4KTMa:b-bp3va4) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ed7d520fa11ab199be76b7935ffa2c7bf5def26a/src/main/proto/wfa/measurement/internal/edpaggregator/raw_impression_upload_service.proto#L82)):*
> ```proto
> 
> // Response message for ListRawImpressionUploads.
> message ListRawImpressionUploadsResponse {
> ```

as per public api comment, let's add a filter here to filter by `RawImpressionUpload` and / or `create_time`

Marco-Premier (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3818)* status: all files reviewed, 6 unresolved discussions (waiting on getina and stevenwarejones).

___
*[`src/main/proto/wfa/measurement/edpaggregator/v1alpha/raw_impression_upload.proto` line 41 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3818#-Ou7lLYK91OQ5DD3RISl:-Ou7uGhV0181rTLPLFLF:b-9c31r) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ed7d520fa11ab199be76b7935ffa2c7bf5def26a/src/main/proto/wfa/measurement/edpaggregator/v1alpha/raw_impression_upload.proto#L41)):*

Let's add 2 new fields:

```
RawImpressionUpload {
    STATE_UNSPECIFIED = 0;
    CREATED = 1; // the upload was created but model line are not processing yet
    ACTIVE = 2;      // at least one model line is still processing
    COMPLETED = 3;   // all model lines completed successfully
    FAILED = 4;      // at least one model line failed
  }
```

```
string done_blob_uri
```
</blockquote></details>

the state field is called: `RawImpressionUploadState`

Marco-Premier (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3818)* status: all files reviewed, 6 unresolved discussions (waiting on getina and stevenwarejones).

___
*[`src/main/proto/wfa/measurement/edpaggregator/v1alpha/raw_impression_upload_service.proto` line 96 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3818#-OtFhc8k7Wm4-Ko9xNDB:-Ou7vbHq-_fqewX3dCQt:b-g6uh62) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ed7d520fa11ab199be76b7935ffa2c7bf5def26a/src/main/proto/wfa/measurement/edpaggregator/v1alpha/raw_impression_upload_service.proto#L96)):*

TODO: add a filter that let query by `RawImpressionUpload` and / or `create_time`
</blockquote></details>

`RawImpressionUploadState` and or `create_time`

getina (commented):
RawImpressionUploadFile and RawImpressionUploadModelLine services will be in follow-up PRs. I will add tracking issues to these!

> @getina made 7 comments.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3818#-:-Ou9X66D2J_APUbNQVvq:be26pe3)* status: all files reviewed, 6 unresolved discussions (waiting on Marco-Premier and stevenwarejones).

___
*[`src/main/proto/wfa/measurement/edpaggregator/v1alpha/raw_impression_upload.proto` line 41 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3818#-Ou7lLYK91OQ5DD3RISl:-Ou9QkljEMS3jep4sz5m:b-kph0hz) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ed7d520fa11ab199be76b7935ffa2c7bf5def26a/src/main/proto/wfa/measurement/edpaggregator/v1alpha/raw_impression_upload.proto#L41)):*

the state field is called: `RawImpressionUploadState`
</blockquote></details>

Every public proto in the codebase uses enum State, not enum RawImpressionMetadataBatchState or any long-form name. Added done_blob_uri and state to the public proto
___
*[`src/main/proto/wfa/measurement/edpaggregator/v1alpha/raw_impression_upload_service.proto` line 56 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3818#-OtFhp6_8AP5eXgaZ7e_:-Ou9VAf_A9E0AZvkiFVw:bb7f2mf) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ed7d520fa11ab199be76b7935ffa2c7bf5def26a/src/main/proto/wfa/measurement/edpaggregator/v1alpha/raw_impression_upload_service.proto#L56)):*

as per comment above I am happy to add the done blob uri
</blockquote></details>

done
___
*[`src/main/proto/wfa/measurement/edpaggregator/v1alpha/raw_impression_upload_service.proto` line 96 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3818#-OtFhc8k7Wm4-Ko9xNDB:-Ou9V9U80PP4NIDAhiDj:bb7f2mf) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ed7d520fa11ab199be76b7935ffa2c7bf5def26a/src/main/proto/wfa/measurement/edpaggregator/v1alpha/raw_impression_upload_service.proto#L96)):*

`RawImpressionUploadState` and or `create_time`
</blockquote></details>

done
___
*[`src/main/proto/wfa/measurement/internal/edpaggregator/raw_impression_upload.proto` line 30 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3818#-OtFhBTW71tLZbUsOE_P:-Ou9VRZQ4XJcwLhN-EMg:bb7f2mf) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ed7d520fa11ab199be76b7935ffa2c7bf5def26a/src/main/proto/wfa/measurement/internal/edpaggregator/raw_impression_upload.proto#L30)):*

Same comment on public api:

Let's add 2 new fields:

```
RawImpressionUpload {
    STATE_UNSPECIFIED = 0;
    CREATED = 1; // the upload was created but model line are not processing yet
    ACTIVE = 2;      // at least one model line is still processing
    COMPLETED = 3;   // all model lines completed successfully
    FAILED = 4;      // at least one model line failed
  }
```

```
string done_blob_uri
```

As per @stevenwarejones  comment the request id is set in the `CreateRawImpressionUploadRequest` so there is no need to add it here.
</blockquote></details>

done
___
*[`src/main/proto/wfa/measurement/internal/edpaggregator/raw_impression_upload_service.proto` line 47 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3818#-OtFhOwtEEgJ4FXKkpTQ:-Ou9Vt1R8MGFzs5v0COw:bb7f2mf) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ed7d520fa11ab199be76b7935ffa2c7bf5def26a/src/main/proto/wfa/measurement/internal/edpaggregator/raw_impression_upload_service.proto#L47)):*

+1 drop the upload it and resource id
</blockquote></details>

done
___
*[`src/main/proto/wfa/measurement/internal/edpaggregator/raw_impression_upload_service.proto` line 82 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3818#-Ou7oeZr7yusrXv4KTM_:-Ou9Vtdo2ZY-2TVIXOch:bb7f2mf) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ed7d520fa11ab199be76b7935ffa2c7bf5def26a/src/main/proto/wfa/measurement/internal/edpaggregator/raw_impression_upload_service.proto#L82)):*

as per public api comment, let's add a filter here to filter by `RawImpressionUpload` and / or `create_time`
</blockquote></details>

done

Marco-Premier (changes requested):
> @Marco-Premier reviewed 7 files and all commit messages, made 1 comment, and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3818)* status: all files reviewed, 5 unresolved discussions (waiting on getina and stevenwarejones).

___
*[`src/main/proto/wfa/measurement/internal/edpaggregator/raw_impression_upload_service.proto` line 78 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3818#-OuBXnMGC0dAuiuMjsUn:-OuBXnMGC0dAuiuMjsUo:b-da0ikn) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f84d80b30a7336d16ff727ad81e6ac53c4ff1419/src/main/proto/wfa/measurement/internal/edpaggregator/raw_impression_upload_service.proto#L78)):*
> ```proto
>   message Filter {
>     // Filter by pipeline state.
>     RawImpressionUploadState state = 1;
> ```

public filter is using two fields. Add the missing one here: 

```
google.type.Interval create_time_interval = 2
        [(google.api.field_behavior) = OPTIONAL];
```

Marco-Premier (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3818)* status: all files reviewed, 6 unresolved discussions (waiting on getina and stevenwarejones).

___
*[`src/main/proto/wfa/measurement/edpaggregator/v1alpha/raw_impression_upload_service.proto` line 42 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3818#-OuBaSG82nl6mqZ_QZFh:-OuBaSG82nl6mqZ_QZFi:b-8fgag2) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f84d80b30a7336d16ff727ad81e6ac53c4ff1419/src/main/proto/wfa/measurement/edpaggregator/v1alpha/raw_impression_upload_service.proto#L42)):*
> ```proto
>   rpc ListRawImpressionUploads(ListRawImpressionUploadsRequest)
>       returns (ListRawImpressionUploadsResponse) {}
> }
> ```

one things missing:

Since we added the state we need now grpc methods to alter the state.
When an upload is created it will automatically set to `CREATED` we need methods to update the state to:

- `ACTIVE`
- `COMPLETED`
- `FAILED`

An example can be found here: https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/main/src/main/proto/wfa/measurement/edpaggregator/v1alpha/raw_impression_metadata_batch_service.proto#L47-L54

**same applies to the internal api**

stevenwarejones (commented):
The internal service is missing state transition RPCs. The existing RawImpressionMetadataBatchService has MarkRawImpressionMetadataBatchProcessed and MarkRawImpressionMetadataBatchFailed, and RequisitionMetadataService has
  QueueRequisitionMetadata, StartProcessingRequisitionMetadata, FulfillRequisitionMetadata, etc.

  For the CREATED → ACTIVE → COMPLETED/FAILED state machine on this resource, the analogous RPCs would be:
```
  rpc ActivateRawImpressionUpload(ActivateRawImpressionUploadRequest)
      returns (RawImpressionUpload);

  rpc CompleteRawImpressionUpload(CompleteRawImpressionUploadRequest)
      returns (RawImpressionUpload);

  rpc FailRawImpressionUpload(FailRawImpressionUploadRequest)
      returns (RawImpressionUpload);
```
  Each request would take data_provider_resource_id and raw_impression_upload_resource_id. The server enforces valid transitions (e.g., only CREATED → ACTIVE, only ACTIVE → COMPLETED/FAILED). These can be follow-up PRs but should have tracking
  issues since they're on the critical path for the dispatcher and pipeline phase transitions.

> @stevenwarejones reviewed 7 files and all commit messages, made 2 comments, and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3818#-:-OuCgdEA31uD61NU-qHH:bzdwtfv)* status: all files reviewed, 4 unresolved discussions (waiting on getina and Marco-Premier).

___
*[`src/main/proto/wfa/measurement/internal/edpaggregator/raw_impression_upload_service.proto` line 78 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3818#-OuBXnMGC0dAuiuMjsUn:-OuCfz8M50cjXte5tZtW:b32v57) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f84d80b30a7336d16ff727ad81e6ac53c4ff1419/src/main/proto/wfa/measurement/internal/edpaggregator/raw_impression_upload_service.proto#L78)):*

public filter is using two fields. Add the missing one here: 

```
google.type.Interval create_time_interval = 2
        [(google.api.field_behavior) = OPTIONAL];
```
</blockquote></details>

+1

getina (commented):
Added MarkRawImpressionUploadActive/MarkRawImpressionUploadCompleted/MarkRawImpressionUploadFailed since state transition RPCs should use Mark<Resource><TargetState> naming (AIP-216),

> @getina made 3 comments.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3818#-:-OuD9pQj0dm9BF9oTvPQ:b-8v42pa)* status: all files reviewed, 4 unresolved discussions (waiting on Marco-Premier).

___
*[`src/main/proto/wfa/measurement/edpaggregator/v1alpha/raw_impression_upload_service.proto` line 42 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3818#-OuBaSG82nl6mqZ_QZFh:-OuD9P2q88c_QA8EFHqW:bb7f2mf) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f84d80b30a7336d16ff727ad81e6ac53c4ff1419/src/main/proto/wfa/measurement/edpaggregator/v1alpha/raw_impression_upload_service.proto#L42)):*

one things missing:

Since we added the state we need now grpc methods to alter the state.
When an upload is created it will automatically set to `CREATED` we need methods to update the state to:

- `ACTIVE`
- `COMPLETED`
- `FAILED`

An example can be found here: https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/main/src/main/proto/wfa/measurement/edpaggregator/v1alpha/raw_impression_metadata_batch_service.proto#L47-L54

**same applies to the internal api**
</blockquote></details>

done
___
*[`src/main/proto/wfa/measurement/internal/edpaggregator/raw_impression_upload_service.proto` line 78 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3818#-OuBXnMGC0dAuiuMjsUn:-OuDAOEOClyWR0oFicEQ:bb7f2mf) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f84d80b30a7336d16ff727ad81e6ac53c4ff1419/src/main/proto/wfa/measurement/internal/edpaggregator/raw_impression_upload_service.proto#L78)):*

+1
</blockquote></details>

done

Marco-Premier (approved):
> @Marco-Premier reviewed 3 files and all commit messages, made 1 comment, and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3818)* status: all files reviewed, 2 unresolved discussions (waiting on stevenwarejones).

___
*[`src/main/proto/wfa/measurement/edpaggregator/v1alpha/raw_impression_upload_service.proto` line 96 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3818#-OtFhc8k7Wm4-Ko9xNDB:-OuDEHoJ1hUSPOU_s3jk:bb6ywmx) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ed7d520fa11ab199be76b7935ffa2c7bf5def26a/src/main/proto/wfa/measurement/edpaggregator/v1alpha/raw_impression_upload_service.proto#L96)):*
<details><summary><i>Previously, getina (Tina Ge) wrote…</i></summary><blockquote>

done
</blockquote></details>

LGTM

stevenwarejones (changes requested):
> @stevenwarejones reviewed 2 files and all commit messages, and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3818)* status: all files reviewed, 4 unresolved discussions (waiting on getina and SanjayVas).

___
*[`src/main/proto/wfa/measurement/edpaggregator/v1alpha/raw_impression_upload_service.proto` line 169 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3818#-OuDRvsK1_uUKLpNlV2v:-OuDRvsK1_uUKLpNlV2w:bhmut6a) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/bc1c82e5893c28ddd2217d8e1e5c41be831445df/src/main/proto/wfa/measurement/edpaggregator/v1alpha/raw_impression_upload_service.proto#L169)):*
> ```proto
> 
> // Request message for the `MarkRawImpressionUploadFailed` method.
> message MarkRawImpressionUploadFailedRequest {
> ```

  When marking an upload as FAILED, there's no way to record why it failed. This matters for debugging and operational visibility — update_time tells you when but not why.
  
  Suggest adding a failure detail field to both the internal and public MarkRawImpressionUploadFailedRequest:
```
  // Human-readable detail about the failure.
  string error_message = 3;
```
  And a corresponding field on the RawImpressionUpload resource:
```
  // Detail about the failure, set when state is FAILED.
  string error_message = 7;
```
___
*[`src/main/proto/wfa/measurement/internal/edpaggregator/raw_impression_upload_service.proto` line 56 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3818#-OuDRUy-EqF5C9zl2_Im:-OuDRUy-EqF5C9zl2_In:b-tgl6ul) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/bc1c82e5893c28ddd2217d8e1e5c41be831445df/src/main/proto/wfa/measurement/internal/edpaggregator/raw_impression_upload_service.proto#L56)):*
> ```proto
> 
> // Request message for CreateRawImpressionUpload.
> message CreateRawImpressionUploadRequest {
> ```

  The existing internal pattern (CreateRawImpressionMetadataBatchRequest) has data_provider_resource_id as a top-level request field. This request embeds it inside the RawImpressionUpload message instead, which is inconsistent and means the service
  implementation has to dig into the resource to find the parent.

  Suggest adding data_provider_resource_id as a top-level field to match the existing convention:
```  
  message CreateRawImpressionUploadRequest {
    string data_provider_resource_id = 1;
    RawImpressionUpload raw_impression_upload = 2;
    string request_id = 3;
  }
```

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3818)* status: all files reviewed, 5 unresolved discussions (waiting on getina and SanjayVas).

___
*[`src/main/proto/wfa/measurement/internal/edpaggregator/raw_impression_upload.proto` line 28 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3818#-OuDcTwE6PD7BWchc_LN:-OuDcTwE6PD7BWchc_LO:bfdiul3) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/bc1c82e5893c28ddd2217d8e1e5c41be831445df/src/main/proto/wfa/measurement/internal/edpaggregator/raw_impression_upload.proto#L28)):*
> ```proto
> // Internal representation of a RawImpressionUpload.
> //
> // One row per EDP impression-upload event. All downstream pipeline state
> ```

nit: `per DataProvider impression...`

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3818)* status: all files reviewed, 6 unresolved discussions (waiting on getina and SanjayVas).

___
*[`src/main/proto/wfa/measurement/edpaggregator/v1alpha/raw_impression_upload.proto` line 56 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3818#-OuDdyi25aOpKcXtNF75:-OuDdyi25aOpKcXtNF76:bta888s) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/bc1c82e5893c28ddd2217d8e1e5c41be831445df/src/main/proto/wfa/measurement/edpaggregator/v1alpha/raw_impression_upload.proto#L56)):*
> ```proto
>   string name = 1 [(google.api.field_behavior) = IDENTIFIER];
> 
>   // The current pipeline state of this upload.
> ```

// Denormalized pipeline state, rolled up from child model line states.

SanjayVas (changes requested):
> @SanjayVas reviewed 2 files and all commit messages, and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3818)* status: all files reviewed, 8 unresolved discussions (waiting on getina and stevenwarejones).

___
*[`src/main/proto/wfa/measurement/edpaggregator/v1alpha/raw_impression_upload.proto` line 39 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3818#-OuEl1wE9EXm_p-omu_x:-OuEl1wE9EXm_p-omu_y:bgp9c9q) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/bc1c82e5893c28ddd2217d8e1e5c41be831445df/src/main/proto/wfa/measurement/edpaggregator/v1alpha/raw_impression_upload.proto#L39)):*
> ```proto
>   };
> 
>   // Denormalized pipeline state, rolled up from child model line states.
> ```

This is referencing child model lines, but the child resource is a file. Is there anything about the model line processing that will be exposed to the caller? e.g. if "at least one model line failed", one would assume the caller should have some way to determine which one(s) failed.
___
*[`src/main/proto/wfa/measurement/edpaggregator/v1alpha/raw_impression_upload_service.proto` line 123 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3818#-OuElUqM0usqVMNthUjO:-OuElUqM0usqVMNthUjP:b-mk0gao) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/bc1c82e5893c28ddd2217d8e1e5c41be831445df/src/main/proto/wfa/measurement/edpaggregator/v1alpha/raw_impression_upload_service.proto#L123)):*
> ```proto
>         [(google.api.field_behavior) = OPTIONAL];
> 
>     // Filter by create time range.
> ```

Just to make sure since we've had this come up before: this is filtering for resources that have a `create_time` that falls within this interval, yes? You may want to be more explicit and state that.

Also note whether start and end are required. By default either can be left out to simply get create times before a given value or after a given value. Note that making all of these efficient can be tricky. Consider whether it would be sufficient to just support create time before or create time after.

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3818)* status: all files reviewed, 7 unresolved discussions (waiting on getina).

___
*[`src/main/proto/wfa/measurement/edpaggregator/v1alpha/raw_impression_upload_service.proto` line 42 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3818#-OuBaSG82nl6mqZ_QZFh:-OuHoYo8DZXtVDAmzR0O:bzbhup9) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f84d80b30a7336d16ff727ad81e6ac53c4ff1419/src/main/proto/wfa/measurement/edpaggregator/v1alpha/raw_impression_upload_service.proto#L42)):*
<details><summary><i>Previously, getina (Tina Ge) wrote…</i></summary><blockquote>

done
</blockquote></details>

just following up here to remove the state transitions since they are handled on the children

getina (commented):
> @getina made 6 comments and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3818)* status: all files reviewed, 6 unresolved discussions (waiting on Marco-Premier, SanjayVas, and stevenwarejones).

___
*[`src/main/proto/wfa/measurement/edpaggregator/v1alpha/raw_impression_upload.proto` line 39 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3818#-OuEl1wE9EXm_p-omu_x:-OuJ-Cw11ECaMWwtoeR-:b2e2aai) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/bc1c82e5893c28ddd2217d8e1e5c41be831445df/src/main/proto/wfa/measurement/edpaggregator/v1alpha/raw_impression_upload.proto#L39)):*

This is referencing child model lines, but the child resource is a file. Is there anything about the model line processing that will be exposed to the caller? e.g. if "at least one model line failed", one would assume the caller should have some way to determine which one(s) failed.
</blockquote></details>

The model line is a separate child resource with its own Get/List RPCs (https://github.com/world-federation-of-advertisers/cross-media-measurement/pull/3908). Callers can list model lines for an upload and see which ones failed. The hierarchy is:

RawImpressionUpload (this PR)
  ├── RawImpressionUploadFile (PR #3878)
  └── RawImpressionUploadModelLine (PR #3908)
___
*[`src/main/proto/wfa/measurement/edpaggregator/v1alpha/raw_impression_upload.proto` line 56 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3818#-OuDdyi25aOpKcXtNF75:-OuJ01Hr4ojlRUE7jSGP:btope1r) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/bc1c82e5893c28ddd2217d8e1e5c41be831445df/src/main/proto/wfa/measurement/edpaggregator/v1alpha/raw_impression_upload.proto#L56)):*

// Denormalized pipeline state, rolled up from child model line states.
</blockquote></details>

done, but flagging that this is already a comment about the enum State a few lines above
___
*[`src/main/proto/wfa/measurement/edpaggregator/v1alpha/raw_impression_upload_service.proto` line 42 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3818#-OuBaSG82nl6mqZ_QZFh:-OuINf3_7wGw8RNXJYh3:bb7f2mf) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f84d80b30a7336d16ff727ad81e6ac53c4ff1419/src/main/proto/wfa/measurement/edpaggregator/v1alpha/raw_impression_upload_service.proto#L42)):*

just following up here to remove the state transitions since they are handled on the children
</blockquote></details>

done
___
*[`src/main/proto/wfa/measurement/edpaggregator/v1alpha/raw_impression_upload_service.proto` line 123 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3818#-OuElUqM0usqVMNthUjO:-OuJ0fe7-5Iob_ujvljL:bccgqef) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/bc1c82e5893c28ddd2217d8e1e5c41be831445df/src/main/proto/wfa/measurement/edpaggregator/v1alpha/raw_impression_upload_service.proto#L123)):*

Just to make sure since we've had this come up before: this is filtering for resources that have a `create_time` that falls within this interval, yes? You may want to be more explicit and state that.

Also note whether start and end are required. By default either can be left out to simply get create times before a given value or after a given value. Note that making all of these efficient can be tricky. Consider whether it would be sufficient to just support create time before or create time after.
</blockquote></details>

Added a comment stating that this filters for resources within the create_time interval

@Marco-Premier @stevenwarejones is it enough to just support create time before or after? so changing this interval to a timestamp
___
*[`src/main/proto/wfa/measurement/internal/edpaggregator/raw_impression_upload.proto` line 28 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3818#-OuDcTwE6PD7BWchc_LN:-OuIryffDBHnn7f0z9q6:bb7f2mf) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/bc1c82e5893c28ddd2217d8e1e5c41be831445df/src/main/proto/wfa/measurement/internal/edpaggregator/raw_impression_upload.proto#L28)):*

nit: `per DataProvider impression...`
</blockquote></details>

done
___
*[`src/main/proto/wfa/measurement/internal/edpaggregator/raw_impression_upload_service.proto` line 56 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3818#-OuDRUy-EqF5C9zl2_Im:-OuIsVsy6gpRfS1egdQQ:bb7f2mf) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/bc1c82e5893c28ddd2217d8e1e5c41be831445df/src/main/proto/wfa/measurement/internal/edpaggregator/raw_impression_upload_service.proto#L56)):*

The existing internal pattern (CreateRawImpressionMetadataBatchRequest) has data_provider_resource_id as a top-level request field. This request embeds it inside the RawImpressionUpload message instead, which is inconsistent and means the service
  implementation has to dig into the resource to find the parent.

  Suggest adding data_provider_resource_id as a top-level field to match the existing convention:
```  
  message CreateRawImpressionUploadRequest {
    string data_provider_resource_id = 1;
    RawImpressionUpload raw_impression_upload = 2;
    string request_id = 3;
  }
```

</blockquote></details>

done

stevenwarejones (changes requested):
> @stevenwarejones reviewed 3 files, made 2 comments, and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3818)* status: 6 of 7 files reviewed, 5 unresolved discussions (waiting on getina, Marco-Premier, and SanjayVas).

___
*[`src/main/proto/wfa/measurement/edpaggregator/v1alpha/raw_impression_upload_service.proto` line 123 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3818#-OuElUqM0usqVMNthUjO:-OuJCXO4A4MmCNWu9gYJ:b-x6aw2f) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/bc1c82e5893c28ddd2217d8e1e5c41be831445df/src/main/proto/wfa/measurement/edpaggregator/v1alpha/raw_impression_upload_service.proto#L123)):*
<details><summary><i>Previously, getina (Tina Ge) wrote…</i></summary><blockquote>

Added a comment stating that this filters for resources within the create_time interval

@Marco-Premier @stevenwarejones is it enough to just support create time before or after? so changing this interval to a timestamp
</blockquote></details>

i think if we're going to do that, we would need one field for before and one field for after. I prefer to keep as-is.
___
*[`src/main/proto/wfa/measurement/internal/edpaggregator/raw_impression_upload_service.proto` line 126 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3818#-OuJDXboELmLPeTca63M:-OuJDXboELmLPeTca63N:b-hdmihu) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f69c6576a15a6254f316ee5034fd8cba74f76d13/src/main/proto/wfa/measurement/internal/edpaggregator/raw_impression_upload_service.proto#L126)):*
> ```proto
> }
> 
> // Request message for MarkRawImpressionUploadActive.
> ```

these should be deleted, right?

stevenwarejones (changes requested):
> @stevenwarejones made 4 comments.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3818)* status: 6 of 7 files reviewed, 7 unresolved discussions (waiting on getina, Marco-Premier, and SanjayVas).

___
*[`src/main/proto/wfa/measurement/edpaggregator/v1alpha/raw_impression_upload.proto` line 39 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3818#-OuEl1wE9EXm_p-omu_x:-OuJHc212zkUGKmaxW3X:bjpj50s) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/bc1c82e5893c28ddd2217d8e1e5c41be831445df/src/main/proto/wfa/measurement/edpaggregator/v1alpha/raw_impression_upload.proto#L39)):*
<details><summary><i>Previously, getina (Tina Ge) wrote…</i></summary><blockquote>

The model line is a separate child resource with its own Get/List RPCs (https://github.com/world-federation-of-advertisers/cross-media-measurement/pull/3908). Callers can list model lines for an upload and see which ones failed. The hierarchy is:

RawImpressionUpload (this PR)
  ├── RawImpressionUploadFile (PR #3878)
  └── RawImpressionUploadModelLine (PR #3908)
</blockquote></details>

Agreed that callers need to drill into which model line failed. I think this is fine.
___
*[`src/main/proto/wfa/measurement/internal/edpaggregator/raw_impression_upload_service.proto` line 126 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3818#-OuJDXboELmLPeTca63M:-OuJGYoMA5mkcfm6x8SF:b3zk6iy) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f69c6576a15a6254f316ee5034fd8cba74f76d13/src/main/proto/wfa/measurement/internal/edpaggregator/raw_impression_upload_service.proto#L126)):*

these should be deleted, right?
</blockquote></details>

The parent state is a denormalized rollup from child model line states, so it should be computed at the DB layer when child states change, not set directly via separate RPCs.
___
*[`src/main/proto/wfa/measurement/edpaggregator/v1alpha/raw_impression_upload.proto` line 74 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3818#-OuJIk7a3oQ9o1mG7yyy:-OuJIk7a3oQ9o1mG7yyz:b-5jlqrr) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f69c6576a15a6254f316ee5034fd8cba74f76d13/src/main/proto/wfa/measurement/edpaggregator/v1alpha/raw_impression_upload.proto#L74)):*
> ```proto
> 
>   // Detail about the failure, set when state is FAILED.
>   string error_message = 6 [(google.api.field_behavior) = OUTPUT_ONLY];
> ```

Since we're removing the Mark RPCs from the parent (state is a DB-layer rollup from child model line states), error_message should also be removed from the parent resource. Callers who see FAILED should drill into the RawImpressionUploadModelLine child resources (#3908) to see which model lines failed and why.
___
*[`src/main/proto/wfa/measurement/internal/edpaggregator/raw_impression_upload.proto` line 50 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3818#-OuJIneOA3bTHpN5PqfK:-OuJIneOA3bTHpN5PqfL:b-7ropln) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f69c6576a15a6254f316ee5034fd8cba74f76d13/src/main/proto/wfa/measurement/internal/edpaggregator/raw_impression_upload.proto#L50)):*
> ```proto
> 
>   // Detail about the failure, set when state is FAILED.
>   string error_message = 7;
> ```

ditto

getina (commented):
> @getina made 4 comments.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3818)* status: 6 of 7 files reviewed, 7 unresolved discussions (waiting on Marco-Premier, SanjayVas, and stevenwarejones).

___
*[`src/main/proto/wfa/measurement/edpaggregator/v1alpha/raw_impression_upload.proto` line 74 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3818#-OuJIk7a3oQ9o1mG7yyy:-OuLBuajAYzJ2ARgasRI:bb7f2mf) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f69c6576a15a6254f316ee5034fd8cba74f76d13/src/main/proto/wfa/measurement/edpaggregator/v1alpha/raw_impression_upload.proto#L74)):*

Since we're removing the Mark RPCs from the parent (state is a DB-layer rollup from child model line states), error_message should also be removed from the parent resource. Callers who see FAILED should drill into the RawImpressionUploadModelLine child resources (#3908) to see which model lines failed and why.
</blockquote></details>

done
___
*[`src/main/proto/wfa/measurement/edpaggregator/v1alpha/raw_impression_upload_service.proto` line 123 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3818#-OuElUqM0usqVMNthUjO:-OuLC1ex6Pff5Wg4Ycot:bq0ncok) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/bc1c82e5893c28ddd2217d8e1e5c41be831445df/src/main/proto/wfa/measurement/edpaggregator/v1alpha/raw_impression_upload_service.proto#L123)):*

i think if we're going to do that, we would need one field for before and one field for after. I prefer to keep as-is.
</blockquote></details>

sounds good
___
*[`src/main/proto/wfa/measurement/internal/edpaggregator/raw_impression_upload.proto` line 50 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3818#-OuJIneOA3bTHpN5PqfK:-OuLBsA_2LPMz_-xdeLJ:bb7f2mf) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f69c6576a15a6254f316ee5034fd8cba74f76d13/src/main/proto/wfa/measurement/internal/edpaggregator/raw_impression_upload.proto#L50)):*

ditto
</blockquote></details>

done
___
*[`src/main/proto/wfa/measurement/internal/edpaggregator/raw_impression_upload_service.proto` line 126 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3818#-OuJDXboELmLPeTca63M:-OuLBpNcCtckMYCPFl63:b8qnsld) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f69c6576a15a6254f316ee5034fd8cba74f76d13/src/main/proto/wfa/measurement/internal/edpaggregator/raw_impression_upload_service.proto#L126)):*

The parent state is a denormalized rollup from child model line states, so it should be computed at the DB layer when child states change, not set directly via separate RPCs.
</blockquote></details>

yep - missed this in my commit

stevenwarejones (commented):
> @stevenwarejones reviewed 3 files and all commit messages, and resolved 4 discussions.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3818)* status: all files reviewed, 3 unresolved discussions (waiting on Marco-Premier and SanjayVas).

stevenwarejones (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3818)* status: all files reviewed, 2 unresolved discussions (waiting on Marco-Premier and SanjayVas).

stevenwarejones (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3818)* status: all files reviewed, 2 unresolved discussions (waiting on Marco-Premier and SanjayVas).

SanjayVas (approved):
> state transition RPCs should use Mark naming (AIP-216)

That AIP doesn't state anything about using "Mark" naming. In fact the example it gives uses a specific verb, implying that's the ideal pattern ("Publish" for transitioning to a `PUBLISHED` state). "Mark" is still okay if you can't think of something better or you're really trying to indicate that the method does nothing aside from tracking a state (e.g. where the true state is derived externally to the API).

> @SanjayVas reviewed 6 files and all commit messages, made 2 comments, and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3818#-:-OuOEKfO6ESRNAZEaRg7:b2nw31v)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on getina).

___
*[`src/main/proto/wfa/measurement/edpaggregator/v1alpha/raw_impression_upload_service.proto` line 123 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3818#-OuElUqM0usqVMNthUjO:-OuOFocj43m4eIv7AjKz:b7opuhv) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/bc1c82e5893c28ddd2217d8e1e5c41be831445df/src/main/proto/wfa/measurement/edpaggregator/v1alpha/raw_impression_upload_service.proto#L123)):*
<details><summary><i>Previously, getina (Tina Ge) wrote…</i></summary><blockquote>

sounds good
</blockquote></details>

My question was on whether you truly only needed to support one or the other of "before" or "after". If you need to support both, then this is indeed the right way to do it.

stevenwarejones (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3818)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on getina).

Issue #3832, opened 2026-05-22
Terraform: HTTP cloud function deploy fails when CPU always-allocated with fractional CPU


stevenwarejones:
The `http-cloud-function` terraform module deploys Gen2 cloud functions without specifying `--cpu-throttling`. If the existing Cloud Run service was previously configured with CPU always allocated (`--no-cpu-throttling`) and has fractional CPU (e.g. `0.3333`), GCP rejects the deploy:

```
ERROR: (gcloud.functions.deploy) ResponseError: status=[400], code=[], message=[Could not update Cloud Run service ... spec.template.spec.containers.resources.limits.cpu: Invalid value specified for cpu. Total cpu < 1 is not supported with cpu always allocated (unthrottled).]
```

This is currently blocking all `update-cmms` runs because the `data-availability-cleanup` function in dev and head hits this error.

Fix: Add `--cpu-throttling` to the deploy args in `src/main/terraform/gcloud/modules/http-cloud-function/main.tf`. This explicitly sets CPU throttling (the correct default for on-demand functions), allowing fractional CPU.

stevenwarejones:
Resolved. The `data-availability-cleanup` Cloud Run service in dev had `cpu-throttling=false` (CPU always allocated) with fractional CPU. Fixed via `gcloud run services update data-availability-cleanup --cpu-throttling --region=us-central1 --project=halo-cmm-dev`. Head was not affected.

PR #3831, opened 2026-05-22
fix: Convert EDPA TLS private key to PKCS#8 format

stevenwarejones:
## Summary

- **Root cause**: PR #3826 committed `edp_aggregator_tls.key` in traditional EC format (`-----BEGIN EC PRIVATE KEY-----`). Java TLS libraries expect PKCS#8 format (`-----BEGIN PRIVATE KEY-----`), causing EDPA pods to crash with `java.lang.IllegalStateException`.
- **Fix**: Convert the key to PKCS#8 format using `openssl pkcs8 -topk8 -nocrypt`.
- HEAD environment is currently broken as of nightly/20260522.2.

## Key conversion

```bash
openssl pkcs8 -topk8 -nocrypt \
  -in src/main/k8s/testing/secretfiles/edp_aggregator_tls.key \
  -out edp_aggregator_tls_pkcs8.key
cp edp_aggregator_tls_pkcs8.key src/main/k8s/testing/secretfiles/edp_aggregator_tls.key
```

## Verification

- `openssl pkey -in edp_aggregator_tls.key -noout` -> OK (valid PKCS#8 key)
- Public key extracted from cert matches public key from converted key (no key content change, only format)

## Deployment

After merge, HEAD nightly will pick up the fix. To restore HEAD immediately:

```bash
gh workflow run configure-edp-aggregator.yml -f environment=head -f image-tag=<sha-after-merge> -f apply=true
```

Issue: #3830

PR #3833, opened 2026-05-22
fix: Add --cpu-throttling to cloud function deploy commands

stevenwarejones:
## Summary

- **Root cause**: The `http-cloud-function` and `gcs-bucket-cloud-function` terraform modules deploy Gen2 cloud functions without specifying `--cpu-throttling`. If the existing Cloud Run service was previously configured with CPU always-allocated (`--no-cpu-throttling`) and has fractional CPU (e.g. `0.3333`), GCP rejects the deploy with `Total cpu < 1 is not supported with cpu always allocated`.
- **Impact**: All `update-cmms` workflow runs fail at the terraform step because the `data-availability-cleanup` function in dev and head hits this error.
- **Fix**: Add `--cpu-throttling` to the `gcloud functions deploy` args in both modules. This explicitly enables CPU throttling (the correct default for on-demand functions), which allows fractional CPU values.

## Changes

- `src/main/terraform/gcloud/modules/http-cloud-function/main.tf` — add `--cpu-throttling`
- `src/main/terraform/gcloud/modules/gcs-bucket-cloud-function/main.tf` — add `--cpu-throttling`

## Test plan

- [ ] Merge PR #3831 (base branch) first
- [ ] Run `update-cmms` on dev with `apply=true` — terraform step should pass

Issue: #3832

### DIFF
--- a/src/main/terraform/gcloud/modules/gcs-bucket-cloud-function/main.tf
+++ b/src/main/terraform/gcloud/modules/gcs-bucket-cloud-function/main.tf
@@ -115,6 +115,7 @@ resource "terraform_data" "deploy_gcs_cloud_function" {
         "--trigger-event-filters=bucket=$TRIGGER_BUCKET"
         "--trigger-service-account=$TRIGGER_SERVICE_ACCOUNT"
         "--no-allow-unauthenticated"
+        "--cpu-throttling"
       )
 
       if [[ -n "$EXTRA_ENV_VARS" ]]; then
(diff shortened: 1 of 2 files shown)


stevenwarejones:
Fixed directly via `gcloud run services update data-availability-cleanup --cpu-throttling` on dev. Head was not affected. The `--cpu-throttling` flag is not valid for `gcloud functions deploy`, only for `gcloud run services update`, so the terraform change was incorrect. Closing this PR.

PR #3837, opened 2026-05-23
fix: accumulate event groups in requisition spec

stevenwarejones:
The events field in buildRequisitionInfo was being reassigned inside a for loop over event groups, so only the last event group survived in the requisition spec. Existing tests did not catch this because getMeasurementFilteredVids reads from the same broken spec, making expected match actual.

Issue: #3842


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3837)

SanjayVas (approved):
Wow, surprised we had this bug for this long.

> @SanjayVas reviewed 8 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3837#-:-OtaVGNA0Pah5rANgpoM:b-banmhr)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on stevenwarejones).

PR #3839, opened 2026-05-25
fix: add gRPC service config timeouts to in-process tests

stevenwarejones:
Add 600-second gRPC deadline timeout to all in-process test channels (Kingdom, Duchy, Reporting) via service config to prevent individual RPCs from hanging indefinitely. Change InProcessReportingServer.internalApiChannel from a get() getter to by lazy so it creates one channel instead of a new TLS connection on every access.

Issue: #3843


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3839)

stevenwarejones:
Superseded by #3849, #3850, and #3851 which split this into three focused PRs.

Issue #827, opened 2023-01-25
Postgres tests are flaky


SanjayVas:
The following test targets are marked as flaky:
//src/test/kotlin/org/wfanet/measurement/integration/postgres:PostgresInProcessLifeOfAReportIntegrationTest
//src/test/kotlin/org/wfanet/measurement/reporting/deploy/postgres:PostgresMeasurementsServiceTest
//src/test/kotlin/org/wfanet/measurement/reporting/deploy/postgres:PostgresReportingSetsServiceTest
//src/test/kotlin/org/wfanet/measurement/reporting/deploy/postgres:PostgresReportsServiceTest

SanjayVas:
If these tests tend to pass individually but be flaky when running multiple tests, it may be that these tests need more than one core. This can be indicated on the Bazel target using the `cpu` tag. For example:

```BUILD
tags = ["cpu:2"]
```

PR #3840, opened 2026-05-25
refactor: split report integration tests into protocol-specific targets

stevenwarejones:
Split monolithic report integration tests into focused, protocol-specific targets. Each test now asserts the expected protocol (DIRECT, HMSS, or TRUS_TEE), and InProcessKingdom/InProcessCmmsComponents no longer default hmssEnabled/trusTeeEnabled, requiring explicit values at every call site.

The concurrent report creation test (creating 3 reports at once) is extracted into its own DirectMultipleReport target marked flaky = True, isolating its known BatchCreateMetrics race condition from the stable DirectOnly tests. This means CI flakiness from concurrency issues no longer forces retries of all 26 stable direct-only tests.

Issue: #3844

PR #3841, opened 2026-05-25
fix: force HMSS protocol selection and mark test as flaky

stevenwarejones:
Disable TrusTee for HMSS test to ensure HMSS protocol is selected instead of TrusTee (which has higher priority in MeasurementsService). Mark HMSS test as flaky per issue #827 due to inherent MPC computation participant failures.

Issue: #3845


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3841)

stevenwarejones:
Squashed into PR #3840 — trusTeeEnabled=false, HMSS flaky=True, and DirectOnly timeout=moderate are all part of the refactor commit now.

Issue #3847, opened 2026-05-26
Make hmssEnabled and trusTeeEnabled configurable in InProcessKingdom


stevenwarejones:
InProcessKingdom hardcodes hmssEnabled = true and trusTeeEnabled = true when building the public API server. These should be constructor parameters with default value true so tests can selectively disable protocols.

PR #3850, opened 2026-05-26
feat: Add hmssEnabled and trusTeeEnabled constructor params to InProcessKingdom

stevenwarejones:
Allow tests to selectively disable HMSS and TrusTee protocols. Both default to true, preserving existing behavior.

Issue: #3847


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3850)

PR #3849, opened 2026-05-26
fix: Change InProcessReportingServer.internalApiChannel from get() to by lazy

stevenwarejones:
Change internalApiChannel from a get() getter to by lazy so it creates one channel instead of a new TLS connection on every access.

Issue: #3846

PR #3858, opened 2026-05-27
fix: enable HMSS in ReachMeasurementAccuracyTest

stevenwarejones:
The test has HMSS test methods that require hmssEnabled=true. Setting it to false caused Vid universe size 0 errors on main after the merge of #3840.

Issue: NA


stevenwarejones:
Fix included in #3849 instead.

wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3858)

PR #3861, opened 2026-05-27
fix: Handle malformed and unresolved measurement consumer in EventGroupSync

stevenwarejones:
EventGroupSync crashes when an event group has a malformed measurement_consumer (e.g. "measurementConsumers/" with an empty ID). MeasurementConsumerKey.fromName parses this as a non-null key with an empty ID, passing the existing checkNotNull guard, but then crashing on toName() with IllegalArgumentException: Missing value for measurement_consumer. This crash is fatal to the entire batch.

Fix: In resolveMeasurementConsumers, replace requireNotNull with a safe check that logs and skips malformed keys. Add input validation in validateEventGroup as an additional guard. Add UnresolvedMeasurementConsumerException for the case where MC resolution returns empty, replacing a silent skip that had no failure metric. Only catch UnresolvedMeasurementConsumerException in the sync loop so API errors propagate as expected. Include eventGroupReferenceId and entity key in all log messages.

Issue: #3862

### DIFF
--- a/src/main/kotlin/org/wfanet/measurement/edpaggregator/eventgroups/EventGroupSync.kt
+++ b/src/main/kotlin/org/wfanet/measurement/edpaggregator/eventgroups/EventGroupSync.kt
@@ -61,6 +61,8 @@ import org.wfanet.measurement.edpaggregator.eventgroups.v1alpha.copy
 import org.wfanet.measurement.edpaggregator.eventgroups.v1alpha.mappedEventGroup
 import org.wfanet.measurement.edpaggregator.telemetry.withSpan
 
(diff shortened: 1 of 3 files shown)


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3861)

ugur47 (approved):
> @ugur47 reviewed 3 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3861)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on ugur47).

Issue #3901, opened 2026-06-03
Bytes12IntMap and packed-bytes layout for rank/subpool fingerprint protos


Marco-Premier:
Parent: #3899

## Summary

Add the in-memory data structure and on-disk proto layouts used by the memoized VID labeler's Phase 0 (Pool Assigner) and Phase 1 (Ranker) pipelines.

- **`Bytes12IntMap`** — single-threaded primitive open-addressing hash table mapping 12-byte keys to 32-bit `Int` values. Chunked flat-byte layout (`Array<LongArray>` + 2 × `Array<IntArray>`), power-of-two capacity with bitmask indexing, linear probing with backward-shift deletion, capacity ceiling `2^60` so a single instance can hold >2 B entries on the heap. Reused in two modes: as a true `fp → rank` map (Phase 1 persistent state) and as a membership set by storing a fixed presence sentinel in the value slot and querying via `containsKey` (Phase 0 per-subpool buckets, Phase 1 subpool filter).

- **`RankIndexMap` proto** — Phase 1 persistent per-subpool fp → rank state. Two parallel packed arrays: `bytes fingerprints` (N × 12 B packed back-to-back, no per-element framing) + `repeated fixed32 ranks [packed = true]`. 16 B per entry on the wire. Pairing is positional: the fingerprint at byte offset `i * 12` is paired with `ranks[i]`.

- **`SubpoolFingerprints` proto** — Phase 0 → Phase 1 per-subpool fingerprint blob. Single `bytes fingerprints` field, N × 12 B packed; saves ~17% wire size and ~60 GB heap-object overhead at 1 B entries versus a `repeated bytes` layout. Phase 1 decodes it into a `Bytes12IntMap` used as a set, since the consumer needs `contains(fp)` per impression to filter the input stream.

## Capacity / performance

- Per slot: 16 B (8 + 4 + 4 across the three parallel arrays). ~21 B/entry at default load factor 0.75.
- 360 M entries → ~7 GiB; 1.5 B → ~31 GiB; 2 B → ~42 GiB.
- Hot path: bitmask indexing + linear probing on parallel primitive arrays.
- Capacity ceiling: `MAX_CAPACITY = 2^60` slots; bounded in practice by heap.

PR #3904, opened 2026-06-03
feat: add Bytes12IntMap and packed-bytes layout for rank/subpool fingerprint protos

Marco-Premier:
## Summary

  Introduces the in-memory data structure and on-disk proto layouts used by the
  memoized VID labeler's Phase 0 (Pool Assigner) and Phase 1 (Ranker) pipelines.

  - **`Bytes12IntMap`** — single-threaded, primitive open-addressing hash table
    mapping 12-byte keys to 32-bit `Int` values. Chunked flat-byte layout with
    linear probing and backward-shift deletion. Designed for the India worst-case
    scale (~1.5 B entries / ~31 GiB heap per instance). The same class is reused
    in two modes: as a true `fp → rank` map (Phase 1's persistent rank state) and
    as a membership set by storing a fixed presence sentinel in the value slot
    and querying via `containsKey` (Phase 0 per-subpool buckets, Phase 1 subpool
    membership filter).

  - **`rank_index_map.proto`** — persistent per-subpool fp → rank state read and
    written by the Phase 1 ranker. Decodes directly into a `Bytes12IntMap` used
    as a true map. Uses two parallel packed arrays (`bytes fingerprints` +
    `repeated fixed32 ranks [packed = true]`) so each entry costs exactly 16
    bytes on the wire with no per-element framing overhead.

  - **`subpool_fingerprints.proto`** — per-subpool fingerprint blob produced by
    Phase 0 Pool Assigner VMs and consumed by Phase 1 rankers. Single
    `bytes fingerprints` field packs N × 12 bytes back-to-back; saves ~17% wire
    size and ~60 GB heap-object overhead at 1 B entries versus a `repeated bytes`
    layout. Decodes into a `Bytes12IntMap` used as a **membership set** (value
    slot holds a presence sentinel, lookups go through `containsKey`)

Issue: #3901

### DIFF
--- a/src/main/kotlin/org/wfanet/measurement/edpaggregator/vidlabeler/utils/BUILD.bazel
+++ b/src/main/kotlin/org/wfanet/measurement/edpaggregator/vidlabeler/utils/BUILD.bazel
@@ -1,0 +1,11 @@
+load("@wfa_rules_kotlin_jvm//kotlin:defs.bzl", "kt_jvm_library")
+
+package(default_visibility = [
+    "//src/main/kotlin/org/wfanet/measurement/edpaggregator:__subpackages__",
+    "//src/test/kotlin/org/wfanet/measurement/edpaggregator:__subpackages__",
+])
+
+kt_jvm_library(
+    name = "fingerprint_rank_table",
+    srcs = ["FingerprintRankTable.kt"],
+)
--- a/src/main/kotlin/org/wfanet/measurement/edpaggregator/vidlabeler/utils/BUILD.bazel
+++ b/src/main/kotlin/org/wfanet/measurement/edpaggregator/vidlabeler/utils/BUILD.bazel
@@ -6,6 +6,6 @@ package(default_visibility = [
 ])
 
 kt_jvm_library(
-    name = "fingerprint_rank_table",
-    srcs = ["FingerprintRankTable.kt"],
+    name = "bytes12_int_map",
+    srcs = ["Bytes12IntMap.kt"],
 )
--- a/src/main/kotlin/org/wfanet/measurement/edpaggregator/vidlabeler/utils/Bytes12IntMap.kt
+++ b/src/main/kotlin/org/wfanet/measurement/edpaggregator/vidlabeler/utils/Bytes12IntMap.kt
@@ -17,108 +17,85 @@
 package org.wfanet.measurement.edpaggregator.vidlabeler.utils
 
 /**
- * Single-threaded, primitive open-addressing hash table mapping
- * 12-byte keys to 32-bit `Int` values. Optimized for the memoized
- * VID pipeline, where 12-byte keys are typically the leading bytes
- * of a cryptographic fingerprint (e.g. SHA-256) and values carry
- * domain-specific meaning such as a rank (Phase 1 ranker) or a
- * presence sentinel when the table is used as a membership set
+ * Single-threaded, primitive open-addressing hash table mapping 12-byte keys to 32-bit `Int`
+ * values. Optimized for the memoized VID pipeline, where 12-byte keys are typically the leading
+ * bytes of a cryptographic fingerprint (e.g. SHA-256) and values carry domain-specific meaning such
+ * as a rank (Phase 1 ranker) or a presence sentinel when the table is used as a membership set
  * (Phase 0 subpool buckets, Phase 1 subpool filter).
- *
  *
  * ## Storage layout (chunked)
  *
  * Three parallel arrays-of-primitive-arrays:
- *  * `keysHi: Array<LongArray>` — high 8 bytes of each key.
- *  * `keysLo: Array<IntArray>`  — low 4 bytes of each key.
- *  * `values: Array<IntArray>`  — the associated `Int` value.
+ * * `keysHi: Array<LongArray>` — high 8 bytes of each key.
+ * * `keysLo: Array<IntArray>` — low 4 bytes of each key.
+ * * `values: Array<IntArray>` — the associated `Int` value.
  *
- * Each inner array is one *chunk* of `1 shl chunkShift` slots. The
- * total capacity is `numChunks × chunkSize`. Both `chunkSize` and
- * `numChunks` are powers of two so that the total capacity is itself a
- * power of two and `index = hash and (capacity - 1)` is a single
- * bitwise AND on `Long`s.
+ * Each inner array is one *chunk* of `1 shl chunkShift` slots. The total capacity is `numChunks ×
+ * chunkSize`. Both `chunkSize` and `numChunks` are powers of two so that the total capacity is
+ * itself a power of two and `index = hash and (capacity - 1)` is a single bitwise AND on `Long`s.
  *
- * The two-level layout lets the table grow past the JVM's
- * `Int.MAX_VALUE` single-array limit while preserving the bitmask hot
- * path: a logical `Long` index `i` decomposes to
- * `(chunkIdx, slotIdx) = (i ushr chunkShift, i.toInt() and slotMask)`,
- * which are pure bitwise operations.
+ * The two-level layout lets the table grow past the JVM's `Int.MAX_VALUE` single-array limit while
+ * preserving the bitmask hot path: a logical `Long` index `i` decomposes to `(chunkIdx, slotIdx) =
+ * (i ushr chunkShift, i.toInt() and slotMask)`, which are pure bitwise operations.
  *
- * Collisions are resolved with linear probing on the logical index. On
- * [remove] the freed slot is filled by backward-shifting subsequent
- * occupied slots; the algorithm crosses chunk boundaries transparently
- * because the logical index wraps at `capacity` rather than at any
- * chunk edge.
+ * Collisions are resolved with linear probing on the logical index. On [remove] the freed slot is
+ * filled by backward-shifting subsequent occupied slots; the algorithm crosses chunk boundaries
+ * transparently because the logical index wraps at `capacity` rather than at any chunk edge.
  *
  * ## Memory cost
  *
- * Approximately `16 B` per slot (8 + 4 + 4 across the three parallel
- * arrays). At the default `0.75` load factor that is roughly `21 B`
- * per stored entry. Examples:
- *  * 360 M entries → ~7 GiB
- *  * 1.5 B entries → ~31 GiB
- *  * 2 B entries → ~42 GiB
+ * Approximately `16 B` per slot (8 + 4 + 4 across the three parallel arrays). At the default `0.75`
+ * load factor that is roughly `21 B` per stored entry. Examples:
+ * * 360 M entries → ~7 GiB
+ * * 1.5 B entries → ~31 GiB
+ * * 2 B entries → ~42 GiB
  *
- * Chunked allocation also bounds the *peak* transient memory during
- * [resize]: new chunks are allocated incrementally rather than as a
- * single large primitive array.
+ * Chunked allocation also bounds the *peak* transient memory during [resize]: new chunks are
+ * allocated incrementally rather than as a single large primitive array.
  *
  * ## Capacity ceiling
  *
- * Per chunk: up to `2^30` slots (the largest power of two below the
- * JVM array length limit). Total capacity: up to `2^30 × 2^30 = 2^60`
- * slots, far beyond any realistic use. In practice the table is
- * bounded only by heap memory.
+ * Per chunk: up to `2^30` slots (the largest power of two below the JVM array length limit). Total
+ * capacity: up to `2^30 × 2^30 = 2^60` slots, far beyond any realistic use. In practice the table
+ * is bounded only by heap memory.
  *
  * ## Key assumptions
  *
- * Keys are assumed to be drawn from a uniform distribution (e.g. the
- * leading bytes of a cryptographic hash such as SHA-256). The table
- * applies only a cheap mixing step before reducing to a table index;
- * adversarial or highly-clustered keys can produce pathological probe
- * chains.
+ * Keys are assumed to be drawn from a uniform distribution (e.g. the leading bytes of a
+ * cryptographic hash such as SHA-256). The table applies only a cheap mixing step before reducing
+ * to a table index; adversarial or highly-clustered keys can produce pathological probe chains.
  *
  * ## Absent-key sentinel
  *
- * [get] and [remove] return [NOT_PRESENT] (== `-1`) when the
- * requested key is not in the table. Callers that only ever store
- * non-negative values (e.g. ranks, subpool IDs, or a fixed
- * presence-sentinel such as `0`) can rely on this return value to
- * distinguish "missing" from "present" without a separate
- * [containsKey] probe. Callers that may store `-1` as a real value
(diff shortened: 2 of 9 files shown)


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3904)

stevenwarejones (changes requested):
> @stevenwarejones reviewed 1 file and made 6 comments.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3904)* status: 1 of 7 files reviewed, 6 unresolved discussions (waiting on Marco-Premier).

___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/vidlabeler/utils/Bytes12IntMap.kt` line 353 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3904#-OuDcKxfEMDjbiKFxjeV:-OuDcKxgDn39cBZ2d4pt:bpwru6n) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e3d9ab0da16a76434e5022336a9b2d68c6c6a0f3/src/main/kotlin/org/wfanet/measurement/edpaggregator/vidlabeler/utils/Bytes12IntMap.kt#L353)):*
> ```Kotlin
>    * The cost is one extra `chunkVal[slot]` read on a hit (~1 ns).
>    */
>   fun containsKey(keyHi: Long, keyLo: Int): Boolean = get(keyHi, keyLo) != NOT_PRESENT
> ```

  containsKey delegates to get(), so it returns false when the stored value happens to equal NOT_PRESENT (-1). The KDoc documents this, but the randomized test only generates values in 0 until Int.MAX_VALUE, so the limitation is never exercised.
  Worth either (a) adding a test that documents the expected (incorrect) behavior as a known limitation, or (b) implementing containsKey with its own probe loop that checks the key directly, independent of the stored value.
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/vidlabeler/utils/Bytes12IntMap.kt` line 574 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3904#-OuDdbLP4g8Xl2kgp0xh:-OuDdbLP4g8Xl2kgp0xi:b2uxtfk) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e3d9ab0da16a76434e5022336a9b2d68c6c6a0f3/src/main/kotlin/org/wfanet/measurement/edpaggregator/vidlabeler/utils/Bytes12IntMap.kt#L574)):*
> ```Kotlin
>   /**
>    * Maps a 96-bit key to a logical Long slot index in `[0, capacity)`.
>    * Combines all 96 bits of the key by XOR-ing the 32-bit low half
> ```

Nit: The method-level KDoc says "Combines all 96 bits" — worth clarifying that it combines them via XOR without avalanche mixing. The class-level KDoc already warns about the uniformity assumption, but a reader looking only at this method might
  assume stronger mixing.
___
*[`src/test/kotlin/org/wfanet/measurement/edpaggregator/vidlabeler/utils/Bytes12IntMapTest.kt` line 256 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3904#-OuDeUemA7ZBpCJkccm-:-OuDeUemA7ZBpCJkccm0:b-9a4f3y) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e3d9ab0da16a76434e5022336a9b2d68c6c6a0f3/src/test/kotlin/org/wfanet/measurement/edpaggregator/vidlabeler/utils/Bytes12IntMapTest.kt#L256)):*
> ```Kotlin
> 
>   @Test
>   fun `entries survive resize`() {
> ```

  This test starts with 2 chunks (32 / 16). Consider adding a case that starts single-chunk (initialCapacity = 16, maxChunkShift = 4) and inserts enough to force the transition from 1 chunk to 2+ chunks. That's a distinct transition in installLayout
  where numChunks goes from 1 to >1 and the chunk-index decomposition in the probing hot path activates for re-inserted entries for the first time.
___
*[`src/test/kotlin/org/wfanet/measurement/edpaggregator/vidlabeler/utils/Bytes12IntMapTest.kt` line 544 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3904#-OuDfMNH07hrknJLbuyk:-OuDfMNH07hrknJLbuyl:b-h9tbaz) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e3d9ab0da16a76434e5022336a9b2d68c6c6a0f3/src/test/kotlin/org/wfanet/measurement/edpaggregator/vidlabeler/utils/Bytes12IntMapTest.kt#L544)):*
> ```Kotlin
>   }
> 
>   private fun assertFailsRequire(block: () -> Any?) {
> ```

Use assertThrows<IllegalArgumentException> { ... } instead of the hand-rolled helper. It's already available via the JUnit dep, gives better failure diagnostics, and won't accidentally pass if a check() (IllegalStateException) is thrown where a
  require() (IllegalArgumentException) was expected.
___
*[`src/main/proto/wfa/measurement/edpaggregator/v1alpha/rank_index_map.proto` line 70 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3904#-OuDej12Bq5vajhwkLAd:-OuDej12Bq5vajhwkLAe:bkr6qy6) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e3d9ab0da16a76434e5022336a9b2d68c6c6a0f3/src/main/proto/wfa/measurement/edpaggregator/v1alpha/rank_index_map.proto#L70)):*
> ```proto
>   // The ranked sub-range size for this subpool as configured in the model.
>   // Each value in `ranks` MUST satisfy `0 <= rank < ranked_size`.
>   int64 ranked_size = 2;
> ```

The design doc caps worst-case subpool size at 1.5B (India), well within int32 range. The in-memory map (Bytes12IntMap) already stores ranks as 32-bit Int. Can this just be int32 to match?
___
*[`src/main/proto/wfa/measurement/edpaggregator/v1alpha/rank_index_map.proto` line 82 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3904#-OuDfh12AtTAtXTtDJhm:-OuDfh12AtTAtXTtDJhn:b-2xrhpm) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e3d9ab0da16a76434e5022336a9b2d68c6c6a0f3/src/main/proto/wfa/measurement/edpaggregator/v1alpha/rank_index_map.proto#L82)):*
> ```proto
>   // the rank of the fingerprint at byte offset `i * 12` in
>   // `fingerprints`. Length MUST equal `fingerprints.size() / 12`.
>   repeated fixed32 ranks = 4 [packed = true];
> ```

  Worth noting that the in-memory sentinel NOT_PRESENT (-1 / 0xFFFFFFFF) must never be serialized into this field. Ranks are always in [0, ranked_size) so this should hold by construction, but an explicit note here would help whoever writes the
  serialization code.

Marco-Premier (commented):
> @Marco-Premier made 5 comments and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3904)* status: 1 of 7 files reviewed, 5 unresolved discussions (waiting on stevenwarejones).

___
*[`src/main/proto/wfa/measurement/edpaggregator/v1alpha/rank_index_map.proto` line 70 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3904#-OuDej12Bq5vajhwkLAd:-OuH-6Kh1qye7Jj-Z_hb:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e3d9ab0da16a76434e5022336a9b2d68c6c6a0f3/src/main/proto/wfa/measurement/edpaggregator/v1alpha/rank_index_map.proto#L70)):*

The design doc caps worst-case subpool size at 1.5B (India), well within int32 range. The in-memory map (Bytes12IntMap) already stores ranks as 32-bit Int. Can this just be int32 to match?
</blockquote></details>

Done.
___
*[`src/main/proto/wfa/measurement/edpaggregator/v1alpha/rank_index_map.proto` line 82 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3904#-OuDfh12AtTAtXTtDJhm:-OuH-79k-yJNGZoEdkb7:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e3d9ab0da16a76434e5022336a9b2d68c6c6a0f3/src/main/proto/wfa/measurement/edpaggregator/v1alpha/rank_index_map.proto#L82)):*

Worth noting that the in-memory sentinel NOT_PRESENT (-1 / 0xFFFFFFFF) must never be serialized into this field. Ranks are always in [0, ranked_size) so this should hold by construction, but an explicit note here would help whoever writes the
  serialization code.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/vidlabeler/utils/Bytes12IntMap.kt` line 353 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3904#-OuDcKxfEMDjbiKFxjeV:-OuH-4Tp6qD6sO9tgsTm:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e3d9ab0da16a76434e5022336a9b2d68c6c6a0f3/src/main/kotlin/org/wfanet/measurement/edpaggregator/vidlabeler/utils/Bytes12IntMap.kt#L353)):*

containsKey delegates to get(), so it returns false when the stored value happens to equal NOT_PRESENT (-1). The KDoc documents this, but the randomized test only generates values in 0 until Int.MAX_VALUE, so the limitation is never exercised.
  Worth either (a) adding a test that documents the expected (incorrect) behavior as a known limitation, or (b) implementing containsKey with its own probe loop that checks the key directly, independent of the stored value.
</blockquote></details>

Done.
___
*[`src/test/kotlin/org/wfanet/measurement/edpaggregator/vidlabeler/utils/Bytes12IntMapTest.kt` line 256 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3904#-OuDeUemA7ZBpCJkccm-:-OuH-130-dboCt2TvGbT:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e3d9ab0da16a76434e5022336a9b2d68c6c6a0f3/src/test/kotlin/org/wfanet/measurement/edpaggregator/vidlabeler/utils/Bytes12IntMapTest.kt#L256)):*

This test starts with 2 chunks (32 / 16). Consider adding a case that starts single-chunk (initialCapacity = 16, maxChunkShift = 4) and inserts enough to force the transition from 1 chunk to 2+ chunks. That's a distinct transition in installLayout
  where numChunks goes from 1 to >1 and the chunk-index decomposition in the probing hot path activates for re-inserted entries for the first time.
</blockquote></details>

Done.
___
*[`src/test/kotlin/org/wfanet/measurement/edpaggregator/vidlabeler/utils/Bytes12IntMapTest.kt` line 544 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3904#-OuDfMNH07hrknJLbuyk:-OuH-AM_19gdIG1OhV_I:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e3d9ab0da16a76434e5022336a9b2d68c6c6a0f3/src/test/kotlin/org/wfanet/measurement/edpaggregator/vidlabeler/utils/Bytes12IntMapTest.kt#L544)):*

Use assertThrows<IllegalArgumentException> { ... } instead of the hand-rolled helper. It's already available via the JUnit dep, gives better failure diagnostics, and won't accidentally pass if a check() (IllegalStateException) is thrown where a
  require() (IllegalArgumentException) was expected.
</blockquote></details>

Done.

stevenwarejones (changes requested):
> @stevenwarejones partially reviewed 6 files and all commit messages, made 1 comment, and resolved 3 discussions.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3904)* status: all files reviewed, 2 unresolved discussions (waiting on Marco-Premier).

___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/vidlabeler/utils/Bytes12IntMap.kt` line 353 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3904#-OuDcKxfEMDjbiKFxjeV:-OuHlsxh4xkmxaOZ_2-5:b-tyhy37) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e3d9ab0da16a76434e5022336a9b2d68c6c6a0f3/src/main/kotlin/org/wfanet/measurement/edpaggregator/vidlabeler/utils/Bytes12IntMap.kt#L353)):*

Done.
</blockquote></details>

still missing a test, I believe.

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3904)* status: all files reviewed, 1 unresolved discussion (waiting on Marco-Premier).

Marco-Premier (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3904)* status: 6 of 7 files reviewed, 1 unresolved discussion (waiting on stevenwarejones).

___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/vidlabeler/utils/Bytes12IntMap.kt` line 353 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3904#-OuDcKxfEMDjbiKFxjeV:-OuHyogH15FXNQIgE2YG:bdvg2ff) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e3d9ab0da16a76434e5022336a9b2d68c6c6a0f3/src/main/kotlin/org/wfanet/measurement/edpaggregator/vidlabeler/utils/Bytes12IntMap.kt#L353)):*

still missing a test, I believe.
</blockquote></details>

my bad, added it now

stevenwarejones (approved):
> @stevenwarejones partially reviewed 1 file and all commit messages, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3904)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on Marco-Premier).

kungfucraig (approved):
> @kungfucraig partially reviewed 7 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3904)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on Marco-Premier).

Marco-Premier (commented):
> @Marco-Premier partially reviewed 3 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3904)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on Marco-Premier).

PR #3909, opened 2026-06-03
feat: add ListEventGroupActivities to internal kingdom service

getina:
Add ListEventGroupActivities to internal kingdom service with Spanner query, keyset pagination via typed page token, Filter message with date_interval and event_groups, composite page token tiebreaker, and DataProvider existence validation.

Issue: #3952

### DIFF
--- a/src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/SpannerEventGroupActivitiesService.kt
+++ b/src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/SpannerEventGroupActivitiesService.kt
@@ -228,18 +228,6 @@ class SpannerEventGroupActivitiesService(
         .asStatusRuntimeException(Status.Code.INVALID_ARGUMENT)
     }
 
-    // Validate DataProvider exists.
-    val dataProviderResult =
-      DataProviderReader()
-        .readByExternalDataProviderId(
-          client.singleUse(),
-          ExternalId(request.externalDataProviderId),
-        )
-    if (dataProviderResult == null) {
-      throw DataProviderNotFoundException(ExternalId(request.externalDataProviderId))
-        .asStatusRuntimeException(Status.Code.NOT_FOUND, "DataProvider not found.")
-    }
-
     val pageSize =
       if (request.pageSize == 0) {
         DEFAULT_PAGE_SIZE
@@ -264,16 +252,27 @@ class SpannerEventGroupActivitiesService(
       }
 
     val resultList =
-      EventGroupActivityReader()
-        .readEventGroupActivities(
-          client.singleUse(),
-          request.externalDataProviderId,
-          externalEventGroupIds,
-          pageSize + 1,
-          after,
-          dateInterval,
-        )
-        .map { it.eventGroupActivity }
+      client.readOnlyTransaction().use { txn ->
+        // Validate DataProvider exists.
+        val dataProviderResult =
+          DataProviderReader()
+            .readByExternalDataProviderId(txn, ExternalId(request.externalDataProviderId))
+        if (dataProviderResult == null) {
+          throw DataProviderNotFoundException(ExternalId(request.externalDataProviderId))
+            .asStatusRuntimeException(Status.Code.NOT_FOUND)
+        }
+
+        EventGroupActivityReader()
+          .readEventGroupActivities(
+            txn,
+            request.externalDataProviderId,
+            externalEventGroupIds,
+            pageSize + 1,
+            after,
+            dateInterval,
+          )
+          .map { it.eventGroupActivity }
+      }
 
     if (resultList.isEmpty()) {
       return ListEventGroupActivitiesResponse.getDefaultInstance()
--- a/src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/SpannerEventGroupActivitiesService.kt
+++ b/src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/SpannerEventGroupActivitiesService.kt
@@ -20,6 +20,7 @@ import io.grpc.Status
 import kotlin.coroutines.CoroutineContext
 import kotlin.coroutines.EmptyCoroutineContext
 import org.wfanet.measurement.common.grpc.grpcRequire
+import org.wfanet.measurement.common.identity.ExternalId
 import org.wfanet.measurement.common.identity.IdGenerator
 import org.wfanet.measurement.gcloud.spanner.AsyncDatabaseClient
 import org.wfanet.measurement.internal.kingdom.BatchDeleteEventGroupActivitiesRequest
@@ -39,6 +40,7 @@ import org.wfanet.measurement.kingdom.deploy.gcloud.spanner.common.EventGroupNot
 import org.wfanet.measurement.kingdom.deploy.gcloud.spanner.common.InvalidFieldValueException
 import org.wfanet.measurement.kingdom.deploy.gcloud.spanner.common.KingdomInternalException
 import org.wfanet.measurement.kingdom.deploy.gcloud.spanner.common.RequiredFieldNotSetException
+import org.wfanet.measurement.kingdom.deploy.gcloud.spanner.readers.DataProviderReader
 import org.wfanet.measurement.kingdom.deploy.gcloud.spanner.readers.EventGroupActivityReader
 import org.wfanet.measurement.kingdom.deploy.gcloud.spanner.writers.BatchDeleteEventGroupActivities
 import org.wfanet.measurement.kingdom.deploy.gcloud.spanner.writers.BatchUpdateEventGroupActivities
@@ -226,9 +228,16 @@ class SpannerEventGroupActivitiesService(
         .asStatusRuntimeException(Status.Code.INVALID_ARGUMENT)
     }
 
-    if (request.externalEventGroupId == 0L) {
-      throw RequiredFieldNotSetException("external_event_group_id")
-        .asStatusRuntimeException(Status.Code.INVALID_ARGUMENT)
+    // Validate DataProvider exists.
+    val dataProviderResult =
+      DataProviderReader()
+        .readByExternalDataProviderId(
+          client.singleUse(),
+          ExternalId(request.externalDataProviderId),
+        )
+    if (dataProviderResult == null) {
+      throw DataProviderNotFoundException(ExternalId(request.externalDataProviderId))
+        .asStatusRuntimeException(Status.Code.NOT_FOUND, "DataProvider not found.")
     }
 
     val pageSize =
@@ -240,12 +249,19 @@ class SpannerEventGroupActivitiesService(
 
     val after = if (request.hasPageToken()) request.pageToken.after else null
 
+    val externalEventGroupIds =
+      if (request.hasFilter()) {
+        request.filter.externalEventGroupIdsList
+      } else {
+        emptyList()
+      }
+
     val resultList =
       EventGroupActivityReader()
         .readEventGroupActivities(
           client.singleUse(),
           request.externalDataProviderId,
-          request.externalEventGroupId,
+          externalEventGroupIds,
           pageSize + 1,
           after,
         )
@@ -258,10 +274,12 @@ class SpannerEventGroupActivitiesService(
     return listEventGroupActivitiesResponse {
       for ((index, activity) in resultList.withIndex()) {
         if (index == pageSize) {
+          val lastActivity = this@listEventGroupActivitiesResponse.eventGroupActivities.last()
           nextPageToken = listEventGroupActivitiesPageToken {
             this.after =
               ListEventGroupActivitiesPageTokenKt.after {
-                date = this@listEventGroupActivitiesResponse.eventGroupActivities.last().date
+                date = lastActivity.date
+                externalEventGroupId = lastActivity.externalEventGroupId
               }
           }
         } else {
--- a/src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/SpannerEventGroupActivitiesService.kt
+++ b/src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/SpannerEventGroupActivitiesService.kt
@@ -42,6 +42,7 @@ import org.wfanet.measurement.kingdom.deploy.gcloud.spanner.common.KingdomIntern
 import org.wfanet.measurement.kingdom.deploy.gcloud.spanner.common.RequiredFieldNotSetException
 import org.wfanet.measurement.kingdom.deploy.gcloud.spanner.readers.DataProviderReader
 import org.wfanet.measurement.kingdom.deploy.gcloud.spanner.readers.EventGroupActivityReader
+import org.wfanet.measurement.kingdom.deploy.gcloud.spanner.readers.EventGroupReader
 import org.wfanet.measurement.kingdom.deploy.gcloud.spanner.writers.BatchDeleteEventGroupActivities
 import org.wfanet.measurement.kingdom.deploy.gcloud.spanner.writers.BatchUpdateEventGroupActivities
 
@@ -223,6 +224,11 @@ class SpannerEventGroupActivitiesService(
   ): ListEventGroupActivitiesResponse {
     grpcRequire(request.pageSize >= 0) { "Page size cannot be less than 0" }
 
+    if (request.externalDataProviderId == 0L) {
+      throw RequiredFieldNotSetException("external_data_provider_id")
+        .asStatusRuntimeException(Status.Code.INVALID_ARGUMENT)
+    }
+
     val pageSize =
       if (request.pageSize == 0) {
         DEFAULT_PAGE_SIZE
@@ -239,11 +245,6 @@ class SpannerEventGroupActivitiesService(
         null
       }
 
-    if (request.externalDataProviderId == 0L) {
-      throw RequiredFieldNotSetException("external_data_provider_id")
-        .asStatusRuntimeException(Status.Code.INVALID_ARGUMENT)
-    }
-
     val resultList =
       client.readOnlyTransaction().use { txn ->
         // Validate DataProvider exists.
@@ -255,11 +256,29 @@ class SpannerEventGroupActivitiesService(
             .asStatusRuntimeException(Status.Code.NOT_FOUND)
         }
 
+        // Validate EventGroup exists if specified.
+        if (request.externalEventGroupId != 0L) {
+          val eventGroupResult =
+            EventGroupReader()
+              .readByDataProvider(
+                txn,
+                ExternalId(request.externalDataProviderId),
+                ExternalId(request.externalEventGroupId),
+              )
+          if (eventGroupResult == null) {
+            throw EventGroupNotFoundException(
+                ExternalId(request.externalDataProviderId),
+                ExternalId(request.externalEventGroupId),
+              )
+              .asStatusRuntimeException(Status.Code.NOT_FOUND)
+          }
+        }
+
         EventGroupActivityReader()
           .readEventGroupActivities(
             txn,
             request.externalDataProviderId,
-            if (request.hasExternalEventGroupId()) request.externalEventGroupId else null,
+            request.externalEventGroupId,
             pageSize + 1,
             after,
             dateInterval,
--- a/src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/SpannerEventGroupActivitiesService.kt
+++ b/src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/SpannerEventGroupActivitiesService.kt
@@ -256,6 +256,13 @@ class SpannerEventGroupActivitiesService(
         emptyList()
       }
 
+    val dateInterval =
+      if (request.hasFilter() && request.filter.hasDateInterval()) {
+        request.filter.dateInterval
+      } else {
+        null
+      }
+
     val resultList =
       EventGroupActivityReader()
         .readEventGroupActivities(
@@ -264,6 +271,7 @@ class SpannerEventGroupActivitiesService(
           externalEventGroupIds,
           pageSize + 1,
           after,
+          dateInterval,
         )
         .map { it.eventGroupActivity }
 
--- a/src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/SpannerEventGroupActivitiesService.kt
+++ b/src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/SpannerEventGroupActivitiesService.kt
@@ -239,24 +239,27 @@ class SpannerEventGroupActivitiesService(
         null
       }
 
+    if (request.externalDataProviderId == 0L) {
+      throw RequiredFieldNotSetException("external_data_provider_id")
+        .asStatusRuntimeException(Status.Code.INVALID_ARGUMENT)
+    }
+
     val resultList =
       client.readOnlyTransaction().use { txn ->
-        // Validate DataProvider exists if specified.
-        if (request.externalDataProviderId != 0L) {
-          val dataProviderResult =
-            DataProviderReader()
-              .readByExternalDataProviderId(txn, ExternalId(request.externalDataProviderId))
-          if (dataProviderResult == null) {
-            throw DataProviderNotFoundException(ExternalId(request.externalDataProviderId))
-              .asStatusRuntimeException(Status.Code.NOT_FOUND)
-          }
+        // Validate DataProvider exists.
+        val dataProviderResult =
+          DataProviderReader()
+            .readByExternalDataProviderId(txn, ExternalId(request.externalDataProviderId))
+        if (dataProviderResult == null) {
+          throw DataProviderNotFoundException(ExternalId(request.externalDataProviderId))
+            .asStatusRuntimeException(Status.Code.NOT_FOUND)
         }
 
         EventGroupActivityReader()
           .readEventGroupActivities(
             txn,
             request.externalDataProviderId,
-            request.externalEventGroupId,
+            if (request.hasExternalEventGroupId()) request.externalEventGroupId else null,
             pageSize + 1,
             after,
             dateInterval,
--- a/src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/SpannerEventGroupActivitiesService.kt
+++ b/src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/SpannerEventGroupActivitiesService.kt
@@ -19,6 +19,7 @@ import com.google.type.Date
 import io.grpc.Status
 import kotlin.coroutines.CoroutineContext
 import kotlin.coroutines.EmptyCoroutineContext
+import org.wfanet.measurement.common.grpc.grpcRequire
 import org.wfanet.measurement.common.identity.IdGenerator
 import org.wfanet.measurement.gcloud.spanner.AsyncDatabaseClient
 import org.wfanet.measurement.internal.kingdom.BatchDeleteEventGroupActivitiesRequest
@@ -26,13 +27,19 @@ import org.wfanet.measurement.internal.kingdom.BatchUpdateEventGroupActivitiesRe
 import org.wfanet.measurement.internal.kingdom.BatchUpdateEventGroupActivitiesResponse
 import org.wfanet.measurement.internal.kingdom.DeleteEventGroupActivityRequest
 import org.wfanet.measurement.internal.kingdom.EventGroupActivitiesGrpcKt.EventGroupActivitiesCoroutineImplBase
+import org.wfanet.measurement.internal.kingdom.ListEventGroupActivitiesPageTokenKt
+import org.wfanet.measurement.internal.kingdom.ListEventGroupActivitiesRequest
+import org.wfanet.measurement.internal.kingdom.ListEventGroupActivitiesResponse
 import org.wfanet.measurement.internal.kingdom.batchDeleteEventGroupActivitiesRequest
+import org.wfanet.measurement.internal.kingdom.listEventGroupActivitiesPageToken
+import org.wfanet.measurement.internal.kingdom.listEventGroupActivitiesResponse
 import org.wfanet.measurement.kingdom.deploy.gcloud.spanner.common.DataProviderNotFoundException
 import org.wfanet.measurement.kingdom.deploy.gcloud.spanner.common.EventGroupActivityNotFoundException
 import org.wfanet.measurement.kingdom.deploy.gcloud.spanner.common.EventGroupNotFoundException
 import org.wfanet.measurement.kingdom.deploy.gcloud.spanner.common.InvalidFieldValueException
 import org.wfanet.measurement.kingdom.deploy.gcloud.spanner.common.KingdomInternalException
 import org.wfanet.measurement.kingdom.deploy.gcloud.spanner.common.RequiredFieldNotSetException
+import org.wfanet.measurement.kingdom.deploy.gcloud.spanner.readers.EventGroupActivityReader
 import org.wfanet.measurement.kingdom.deploy.gcloud.spanner.writers.BatchDeleteEventGroupActivities
 import org.wfanet.measurement.kingdom.deploy.gcloud.spanner.writers.BatchUpdateEventGroupActivities
 
@@ -207,5 +214,65 @@ class SpannerEventGroupActivitiesService(
     }
 
     return Empty.getDefaultInstance()
+  }
+
+  override suspend fun listEventGroupActivities(
+    request: ListEventGroupActivitiesRequest
+  ): ListEventGroupActivitiesResponse {
+    grpcRequire(request.pageSize >= 0) { "Page size cannot be less than 0" }
+
+    if (request.externalDataProviderId == 0L) {
+      throw RequiredFieldNotSetException("external_data_provider_id")
+        .asStatusRuntimeException(Status.Code.INVALID_ARGUMENT)
+    }
+
+    if (request.externalEventGroupId == 0L) {
+      throw RequiredFieldNotSetException("external_event_group_id")
+        .asStatusRuntimeException(Status.Code.INVALID_ARGUMENT)
+    }
+
+    val pageSize =
+      if (request.pageSize == 0) {
+        DEFAULT_PAGE_SIZE
+      } else {
+        request.pageSize.coerceAtMost(MAX_PAGE_SIZE)
+      }
+
+    val after = if (request.hasPageToken()) request.pageToken.after else null
+
+    val resultList =
+      EventGroupActivityReader()
+        .readEventGroupActivities(
+          client.singleUse(),
+          request.externalDataProviderId,
+          request.externalEventGroupId,
+          pageSize + 1,
+          after,
+        )
+        .map { it.eventGroupActivity }
+
+    if (resultList.isEmpty()) {
+      return ListEventGroupActivitiesResponse.getDefaultInstance()
+    }
+
+    return listEventGroupActivitiesResponse {
+      for ((index, activity) in resultList.withIndex()) {
+        if (index == pageSize) {
+          nextPageToken = listEventGroupActivitiesPageToken {
+            this.after =
+              ListEventGroupActivitiesPageTokenKt.after {
+                date = this@listEventGroupActivitiesResponse.eventGroupActivities.last().date
+              }
+          }
+        } else {
+          this.eventGroupActivities += activity
+        }
+      }
+    }
+  }
+
+  companion object {
+    private const val MAX_PAGE_SIZE = 1000
+    private const val DEFAULT_PAGE_SIZE = 50
   }
 }
--- a/src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/SpannerEventGroupActivitiesService.kt
+++ b/src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/SpannerEventGroupActivitiesService.kt
@@ -223,11 +223,6 @@ class SpannerEventGroupActivitiesService(
   ): ListEventGroupActivitiesResponse {
     grpcRequire(request.pageSize >= 0) { "Page size cannot be less than 0" }
 
-    if (request.externalDataProviderId == 0L) {
-      throw RequiredFieldNotSetException("external_data_provider_id")
-        .asStatusRuntimeException(Status.Code.INVALID_ARGUMENT)
-    }
-
     val pageSize =
       if (request.pageSize == 0) {
         DEFAULT_PAGE_SIZE
@@ -236,13 +231,6 @@ class SpannerEventGroupActivitiesService(
       }
 
     val after = if (request.hasPageToken()) request.pageToken.after else null
-
-    val externalEventGroupIds =
-      if (request.hasFilter()) {
-        request.filter.externalEventGroupIdsList
-      } else {
-        emptyList()
-      }
 
     val dateInterval =
       if (request.hasFilter() && request.filter.hasDateInterval()) {
@@ -253,20 +241,22 @@ class SpannerEventGroupActivitiesService(
 
     val resultList =
       client.readOnlyTransaction().use { txn ->
-        // Validate DataProvider exists.
-        val dataProviderResult =
-          DataProviderReader()
-            .readByExternalDataProviderId(txn, ExternalId(request.externalDataProviderId))
-        if (dataProviderResult == null) {
-          throw DataProviderNotFoundException(ExternalId(request.externalDataProviderId))
-            .asStatusRuntimeException(Status.Code.NOT_FOUND)
+        // Validate DataProvider exists if specified.
+        if (request.externalDataProviderId != 0L) {
+          val dataProviderResult =
+            DataProviderReader()
+              .readByExternalDataProviderId(txn, ExternalId(request.externalDataProviderId))
+          if (dataProviderResult == null) {
+            throw DataProviderNotFoundException(ExternalId(request.externalDataProviderId))
+              .asStatusRuntimeException(Status.Code.NOT_FOUND)
+          }
         }
 
         EventGroupActivityReader()
           .readEventGroupActivities(
             txn,
             request.externalDataProviderId,
-            externalEventGroupIds,
+            request.externalEventGroupId,
             pageSize + 1,
             after,
             dateInterval,
--- a/src/main/kotlin/org/wfanet/measurement/kingdom/service/internal/testing/EventGroupActivitiesServiceTest.kt
+++ b/src/main/kotlin/org/wfanet/measurement/kingdom/service/internal/testing/EventGroupActivitiesServiceTest.kt
@@ -1046,7 +1046,7 @@ abstract class EventGroupActivitiesServiceTest<T : EventGroupActivitiesCoroutine
       eventGroupActivitiesService.listEventGroupActivities(
         listEventGroupActivitiesRequest {
           externalDataProviderId = dataProvider.externalDataProviderId
-          filter = filter { externalEventGroupIds += eventGroup.externalEventGroupId }
+          externalEventGroupId = eventGroup.externalEventGroupId
         }
       )
 
@@ -1121,7 +1121,7 @@ abstract class EventGroupActivitiesServiceTest<T : EventGroupActivitiesCoroutine
       eventGroupActivitiesService.listEventGroupActivities(
         listEventGroupActivitiesRequest {
           externalDataProviderId = dataProvider.externalDataProviderId
-          filter = filter { externalEventGroupIds += eventGroup.externalEventGroupId }
+          externalEventGroupId = eventGroup.externalEventGroupId
           pageSize = 1
         }
       )
@@ -1133,7 +1133,7 @@ abstract class EventGroupActivitiesServiceTest<T : EventGroupActivitiesCoroutine
       eventGroupActivitiesService.listEventGroupActivities(
         listEventGroupActivitiesRequest {
           externalDataProviderId = dataProvider.externalDataProviderId
-          filter = filter { externalEventGroupIds += eventGroup.externalEventGroupId }
+          externalEventGroupId = eventGroup.externalEventGroupId
           pageToken = firstResponse.nextPageToken
         }
       )
@@ -1171,32 +1171,12 @@ abstract class EventGroupActivitiesServiceTest<T : EventGroupActivitiesCoroutine
       eventGroupActivitiesService.listEventGroupActivities(
         listEventGroupActivitiesRequest {
           externalDataProviderId = dataProvider.externalDataProviderId
-          filter = filter { externalEventGroupIds += eventGroup.externalEventGroupId }
+          externalEventGroupId = eventGroup.externalEventGroupId
         }
       )
 
     assertThat(response.eventGroupActivitiesList).isEmpty()
     assertThat(response.hasNextPageToken()).isFalse()
-  }
-
-  @Test
-  fun `listEventGroupActivities throws INVALID_ARGUMENT when external_data_provider_id is missing`() {
-    runBlocking {
-      val exception =
-        assertFailsWith<StatusRuntimeException> {
-          eventGroupActivitiesService.listEventGroupActivities(listEventGroupActivitiesRequest {})
-        }
-
-      assertThat(exception.status.code).isEqualTo(Status.Code.INVALID_ARGUMENT)
-      assertThat(exception.errorInfo)
-        .isEqualTo(
-          errorInfo {
-            domain = KingdomInternalException.DOMAIN
-            reason = ErrorCode.REQUIRED_FIELD_NOT_SET.name
-            metadata["field_name"] = "external_data_provider_id"
-          }
-        )
-    }
   }
 
   @Test
@@ -1245,7 +1225,7 @@ abstract class EventGroupActivitiesServiceTest<T : EventGroupActivitiesCoroutine
       eventGroupActivitiesService.listEventGroupActivities(
         listEventGroupActivitiesRequest {
           externalDataProviderId = dataProvider.externalDataProviderId
-          filter = filter { externalEventGroupIds += eventGroup.externalEventGroupId }
+          externalEventGroupId = eventGroup.externalEventGroupId
           pageSize = 1
         }
       )
@@ -1312,7 +1292,7 @@ abstract class EventGroupActivitiesServiceTest<T : EventGroupActivitiesCoroutine
       eventGroupActivitiesService.listEventGroupActivities(
         listEventGroupActivitiesRequest {
           externalDataProviderId = dataProvider.externalDataProviderId
-          filter = filter { externalEventGroupIds += eventGroup.externalEventGroupId }
+          externalEventGroupId = eventGroup.externalEventGroupId
           pageSize = 1
         }
       )
@@ -1324,7 +1304,7 @@ abstract class EventGroupActivitiesServiceTest<T : EventGroupActivitiesCoroutine
       eventGroupActivitiesService.listEventGroupActivities(
         listEventGroupActivitiesRequest {
           externalDataProviderId = dataProvider.externalDataProviderId
-          filter = filter { externalEventGroupIds += eventGroup.externalEventGroupId }
+          externalEventGroupId = eventGroup.externalEventGroupId
           pageSize = 1
           pageToken = firstResponse.nextPageToken
         }
@@ -1372,7 +1352,7 @@ abstract class EventGroupActivitiesServiceTest<T : EventGroupActivitiesCoroutine
       eventGroupActivitiesService.listEventGroupActivities(
         listEventGroupActivitiesRequest {
           externalDataProviderId = dataProvider.externalDataProviderId
-          filter = filter { externalEventGroupIds += eventGroup.externalEventGroupId }
+          externalEventGroupId = eventGroup.externalEventGroupId
           pageSize = 5000
         }
       )
@@ -1522,7 +1502,7 @@ abstract class EventGroupActivitiesServiceTest<T : EventGroupActivitiesCoroutine
         eventGroupActivitiesService.listEventGroupActivities(
           listEventGroupActivitiesRequest {
             externalDataProviderId = dataProvider.externalDataProviderId
-            filter = filter { externalEventGroupIds += eventGroup.externalEventGroupId }
+            externalEventGroupId = eventGroup.externalEventGroupId
           }
         )
 
@@ -1588,8 +1568,8 @@ abstract class EventGroupActivitiesServiceTest<T : EventGroupActivitiesCoroutine
       eventGroupActivitiesService.listEventGroupActivities(
         listEventGroupActivitiesRequest {
           externalDataProviderId = dataProvider.externalDataProviderId
+          externalEventGroupId = eventGroup.externalEventGroupId
           filter = filter {
-            externalEventGroupIds += eventGroup.externalEventGroupId
             dateInterval = dateInterval {
               startDate = date {
                 year = 2025
--- a/src/main/kotlin/org/wfanet/measurement/kingdom/service/internal/testing/EventGroupActivitiesServiceTest.kt
+++ b/src/main/kotlin/org/wfanet/measurement/kingdom/service/internal/testing/EventGroupActivitiesServiceTest.kt
@@ -1,4 +1,4 @@
-// Copyright 2025 The Cross-Media Measurement Authors
+// Copyright 2026 The Cross-Media Measurement Authors
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -47,6 +47,7 @@ import org.wfanet.measurement.internal.kingdom.EventGroupDetailsKt.EventGroupMet
 import org.wfanet.measurement.internal.kingdom.EventGroupDetailsKt.EventGroupMetadataKt.adMetadata
 import org.wfanet.measurement.internal.kingdom.EventGroupDetailsKt.eventGroupMetadata
 import org.wfanet.measurement.internal.kingdom.EventGroupsGrpcKt.EventGroupsCoroutineImplBase
+import org.wfanet.measurement.internal.kingdom.ListEventGroupActivitiesRequest
 import org.wfanet.measurement.internal.kingdom.MeasurementConsumersGrpcKt.MeasurementConsumersCoroutineImplBase
 import org.wfanet.measurement.internal.kingdom.MediaType
 import org.wfanet.measurement.internal.kingdom.batchDeleteEventGroupActivitiesRequest
@@ -1040,7 +1041,10 @@ abstract class EventGroupActivitiesServiceTest<T : EventGroupActivitiesCoroutine
       eventGroupActivitiesService.listEventGroupActivities(
         listEventGroupActivitiesRequest {
           externalDataProviderId = dataProvider.externalDataProviderId
-          externalEventGroupId = eventGroup.externalEventGroupId
+          filter =
+            ListEventGroupActivitiesRequest.Filter.newBuilder()
+              .addExternalEventGroupIds(eventGroup.externalEventGroupId)
+              .build()
         }
       )
 
@@ -1109,7 +1113,10 @@ abstract class EventGroupActivitiesServiceTest<T : EventGroupActivitiesCoroutine
       eventGroupActivitiesService.listEventGroupActivities(
         listEventGroupActivitiesRequest {
           externalDataProviderId = dataProvider.externalDataProviderId
-          externalEventGroupId = eventGroup.externalEventGroupId
+          filter =
+            ListEventGroupActivitiesRequest.Filter.newBuilder()
+              .addExternalEventGroupIds(eventGroup.externalEventGroupId)
+              .build()
           pageSize = 1
         }
       )
@@ -1121,7 +1128,10 @@ abstract class EventGroupActivitiesServiceTest<T : EventGroupActivitiesCoroutine
       eventGroupActivitiesService.listEventGroupActivities(
         listEventGroupActivitiesRequest {
           externalDataProviderId = dataProvider.externalDataProviderId
-          externalEventGroupId = eventGroup.externalEventGroupId
+          filter =
+            ListEventGroupActivitiesRequest.Filter.newBuilder()
+              .addExternalEventGroupIds(eventGroup.externalEventGroupId)
+              .build()
           pageToken = firstResponse.nextPageToken
         }
       )
@@ -1143,6 +1153,8 @@ abstract class EventGroupActivitiesServiceTest<T : EventGroupActivitiesCoroutine
           day = 10
         }
       )
+    assertThat(secondResponse.eventGroupActivitiesList.map { it.date })
+      .containsNoneIn(firstResponse.eventGroupActivitiesList.map { it.date })
   }
 
   @Test
@@ -1151,7 +1163,10 @@ abstract class EventGroupActivitiesServiceTest<T : EventGroupActivitiesCoroutine
       eventGroupActivitiesService.listEventGroupActivities(
         listEventGroupActivitiesRequest {
           externalDataProviderId = dataProvider.externalDataProviderId
-          externalEventGroupId = eventGroup.externalEventGroupId
+          filter =
+            ListEventGroupActivitiesRequest.Filter.newBuilder()
+              .addExternalEventGroupIds(eventGroup.externalEventGroupId)
+              .build()
         }
       )
 
@@ -1164,31 +1179,11 @@ abstract class EventGroupActivitiesServiceTest<T : EventGroupActivitiesCoroutine
     runBlocking {
       val exception =
         assertFailsWith<StatusRuntimeException> {
-          eventGroupActivitiesService.listEventGroupActivities(
-            listEventGroupActivitiesRequest {
-              externalEventGroupId = eventGroup.externalEventGroupId
-            }
-          )
+          eventGroupActivitiesService.listEventGroupActivities(listEventGroupActivitiesRequest {})
         }
 
       assertThat(exception.status.code).isEqualTo(Status.Code.INVALID_ARGUMENT)
       assertThat(exception).hasMessageThat().contains("external_data_provider_id")
-    }
-
-  @Test
-  fun `listEventGroupActivities throws INVALID_ARGUMENT when external_event_group_id is missing`() =
-    runBlocking {
-      val exception =
-        assertFailsWith<StatusRuntimeException> {
-          eventGroupActivitiesService.listEventGroupActivities(
-            listEventGroupActivitiesRequest {
-              externalDataProviderId = dataProvider.externalDataProviderId
-            }
-          )
-        }
-
-      assertThat(exception.status.code).isEqualTo(Status.Code.INVALID_ARGUMENT)
-      assertThat(exception).hasMessageThat().contains("external_event_group_id")
     }
 
   @Test
@@ -1237,7 +1232,10 @@ abstract class EventGroupActivitiesServiceTest<T : EventGroupActivitiesCoroutine
       eventGroupActivitiesService.listEventGroupActivities(
         listEventGroupActivitiesRequest {
           externalDataProviderId = dataProvider.externalDataProviderId
-          externalEventGroupId = eventGroup.externalEventGroupId
+          filter =
+            ListEventGroupActivitiesRequest.Filter.newBuilder()
+              .addExternalEventGroupIds(eventGroup.externalEventGroupId)
+              .build()
           pageSize = 1
         }
       )
@@ -1300,7 +1298,10 @@ abstract class EventGroupActivitiesServiceTest<T : EventGroupActivitiesCoroutine
       eventGroupActivitiesService.listEventGroupActivities(
         listEventGroupActivitiesRequest {
           externalDataProviderId = dataProvider.externalDataProviderId
-          externalEventGroupId = eventGroup.externalEventGroupId
+          filter =
+            ListEventGroupActivitiesRequest.Filter.newBuilder()
+              .addExternalEventGroupIds(eventGroup.externalEventGroupId)
+              .build()
           pageSize = 1
         }
       )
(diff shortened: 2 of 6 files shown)


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3909)

stevenwarejones (commented):
The Spanner emulator GLIBCXX issue is a known environment problem. Run tests with:
  ```
  bazel test //src/test/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner:SpannerEventGroupActivitiesServiceTest --test_env=LD_LIBRARY_PATH
```
  Tests should pass with this flag.

> @stevenwarejones made 4 comments.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3909#-:-OuEXUYRAYeLBZklrx2T:b-icybpw)* status: 0 of 4 files reviewed, 3 unresolved discussions (waiting on getina).

___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/internal/testing/EventGroupActivitiesServiceTest.kt` line 1008 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3909#-OuEXEM4-z_DPJnbhNIq:-OuEXEM4-z_DPJnbhNIr:baymk1b) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/25d636150165efcc929f5a67a4696cfbee37aca1/src/main/kotlin/org/wfanet/measurement/kingdom/service/internal/testing/EventGroupActivitiesServiceTest.kt#L1008)):*
> ```Kotlin
>     }
> 
>   @Test
> ```

  There's no test for the MAX_PAGE_SIZE coercion. The implementation caps at 1000 via coerceAtMost, and the proto comment documents "values above this will be coerced to the maximum value." Add a test that creates >1 activity and requests pageSize 
  = 5000 to verify the cap is applied.
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/internal/testing/EventGroupActivitiesServiceTest.kt` line 1067 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3909#-OuEX5yv0pWTC62XEJLp:-OuEX5yv0pWTC62XEJLq:b-4moofr) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/25d636150165efcc929f5a67a4696cfbee37aca1/src/main/kotlin/org/wfanet/measurement/kingdom/service/internal/testing/EventGroupActivitiesServiceTest.kt#L1067)):*
> ```Kotlin
> 
>   @Test
>   fun `listEventGroupActivities paginates with page_token`() = runBlocking {
> ```

  The pagination tests don't explicitly verify that items from the previous page are not repeated on the next page. The paginates with page_token test fetches page 1 (Dec 1) then page 2 (Dec 5, 10) — but doesn't assert Dec 1 is absent from page 2.
  The paginates with page_token and page_size test is better (chains through pages of size 1) but also doesn't assert non-overlap.
  
  Consider adding an assertion like:

  assertThat(secondResponse.eventGroupActivitiesList.map { it.date })
      .containsNoneIn(firstResponse.eventGroupActivitiesList.map { it.date })

  This guards against the cursor accidentally using >= instead of >.
___
*[`src/main/proto/wfa/measurement/internal/kingdom/event_group_activities_service.proto` line 93 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3909#-OuEYAKb5dvg3GnbBVdP:-OuEYAKb5dvg3GnbBVdQ:b-bzs9d4) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/25d636150165efcc929f5a67a4696cfbee37aca1/src/main/proto/wfa/measurement/internal/kingdom/event_group_activities_service.proto#L93)):*
> ```proto
>   // 1000; values above this will be coerced to the maximum value.
>   int32 page_size = 3;
>   ListEventGroupActivitiesPageToken page_token = 4;
> ```

  ListEventGroupActivitiesRequest requires a single external_event_group_id and has no date filter. At ~200 activities per (DataProvider, EventGroup) pair, volume per query is fine, but a data provider with 50 event groups requires 50 RPCs to get
  the full picture.

  Add a Filter message and make external_event_group_id optional:
```
  message ListEventGroupActivitiesRequest {
    // Required. Scoped to a single DataProvider.
    fixed64 external_data_provider_id = 1;
    int32 page_size = 3;
    ListEventGroupActivitiesPageToken page_token = 4;
  
    message Filter {
      // Filter by activity date range.
      google.type.Interval date_interval = 1;
      // Filter to specific event groups. If empty, returns activities
      // across all event groups for the data provider.
      repeated fixed64 external_event_group_ids = 2;
    }
    Filter filter = 5;
  }
```
  The Spanner unique index on (DataProviderId, EventGroupId, ActivityDate) supports this efficiently — add AND EventGroups.ExternalEventGroupId IN UNNEST(@eventGroupIds) when the list is non-empty.
  
  This removes the need for external_event_group_id as a top-level required field (field 2 can be deprecated or removed). The reader and service code would need corresponding updates.

**Review by stevenwarejones — COMMENTED** [review]
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3909)* status: 0 of 4 files reviewed, 3 unresolved discussions (waiting on getina).

___
*[`src/main/proto/wfa/measurement/internal/kingdom/event_group_activities_service.proto` line 93 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3909#-OuEYAKb5dvg3GnbBVdP:-OuEYtv97KViNC2q1Mcl:bw005jn) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/25d636150165efcc929f5a67a4696cfbee37aca1/src/main/proto/wfa/measurement/internal/kingdom/event_group_activities_service.proto#L93)):*

ListEventGroupActivitiesRequest requires a single external_event_group_id and has no date filter. At ~200 activities per (DataProvider, EventGroup) pair, volume per query is fine, but a data provider with 50 event groups requires 50 RPCs to get
  the full picture.

  Add a Filter message and make external_event_group_id optional:
```
  message ListEventGroupActivitiesRequest {
    // Required. Scoped to a single DataProvider.
    fixed64 external_data_provider_id = 1;
    int32 page_size = 3;
    ListEventGroupActivitiesPageToken page_token = 4;
  
    message Filter {
      // Filter by activity date range.
      google.type.Interval date_interval = 1;
      // Filter to specific event groups. If empty, returns activities
      // across all event groups for the data provider.
      repeated fixed64 external_event_group_ids = 2;
    }
    Filter filter = 5;
  }
```
  The Spanner unique index on (DataProviderId, EventGroupId, ActivityDate) supports this efficiently — add AND EventGroups.ExternalEventGroupId IN UNNEST(@eventGroupIds) when the list is non-empty.
  
  This removes the need for external_event_group_id as a top-level required field (field 2 can be deprecated or removed). The reader and service code would need corresponding updates.
</blockquote></details>

just following up - don't deprecate the field of course, just delete and rename all the fields since this PR introduces them.

**Review by stevenwarejones — COMMENTED** [review]
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3909)* status: 0 of 4 files reviewed, 4 unresolved discussions (waiting on afterdate and getina).

___
*[`src/main/proto/wfa/measurement/internal/kingdom/event_group_activities_service.proto` line 105 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3909#-OuEZPRkBfmAQJtXp0X_:-OuEZPRkBfmAQJtXp0Xa:b-sz6fc3) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/25d636150165efcc929f5a67a4696cfbee37aca1/src/main/proto/wfa/measurement/internal/kingdom/event_group_activities_service.proto#L105)):*
> ```proto
> }
> 
> message ListEventGroupActivitiesPageToken {
> ```

if you adopt my other comment - then the page token needs a tiebreaker. ActivityDate is unique per (DataProvider, EventGroup) but not across event groups — two event groups can have activities on the same date.
  
  Update the page token and query ordering:
```
  message ListEventGroupActivitiesPageToken {
    message After {
      google.type.Date date = 1;
      fixed64 external_event_group_id = 2;
    }
    After after = 1;
  }
```
  And update the query to ORDER BY ActivityDate ASC, ExternalEventGroupId ASC with a cursor condition of (ActivityDate, ExternalEventGroupId) > (@afterDate, @afterEventGroupId).

**Review by stevenwarejones — COMMENTED** [review]
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3909)* status: 0 of 4 files reviewed, 5 unresolved discussions (waiting on afterdate and getina).

___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/SpannerEventGroupActivitiesService.kt` line 219 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3909#-OuEZzubA5xMTKWfAecg:-OuEZzubA5xMTKWfAech:b9d7q9b) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/25d636150165efcc929f5a67a4696cfbee37aca1/src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/SpannerEventGroupActivitiesService.kt#L219)):*
> ```Kotlin
>   }
> 
>   override suspend fun listEventGroupActivities(
> ```

  Every other method on this service validates that the DataProvider and EventGroup exist and throws an error if not (FAILED_PRECONDITION on batchUpdate, NOT_FOUND on delete). listEventGroupActivities silently returns an empty list when the parent
  doesn't exist — a typo'd external_data_provider_id would return empty with no indication anything is wrong. 
  
  AIP-132 says: "If a caller specifies a parent, and the parent resource does not exist, the server SHOULD return NOT_FOUND."

  There's also no test for this case — the existing tests only cover missing (0-valued) IDs, not valid-but-non-existent IDs.

**Review by getina — COMMENTED** [review]
> @getina made 4 comments.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3909)* status: 0 of 5 files reviewed, 5 unresolved discussions (waiting on afterdate and stevenwarejones).

___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/SpannerEventGroupActivitiesService.kt` line 219 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3909#-OuEZzubA5xMTKWfAecg:-OucqowQ14-Cwuj7c2nf:bb7f2mf) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/25d636150165efcc929f5a67a4696cfbee37aca1/src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/SpannerEventGroupActivitiesService.kt#L219)):*

Every other method on this service validates that the DataProvider and EventGroup exist and throws an error if not (FAILED_PRECONDITION on batchUpdate, NOT_FOUND on delete). listEventGroupActivities silently returns an empty list when the parent
  doesn't exist — a typo'd external_data_provider_id would return empty with no indication anything is wrong. 
  
  AIP-132 says: "If a caller specifies a parent, and the parent resource does not exist, the server SHOULD return NOT_FOUND."

  There's also no test for this case — the existing tests only cover missing (0-valued) IDs, not valid-but-non-existent IDs.
</blockquote></details>

done
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/internal/testing/EventGroupActivitiesServiceTest.kt` line 1008 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3909#-OuEXEM4-z_DPJnbhNIq:-OuctA8VBnIDj6COOgqU:bb7f2mf) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/25d636150165efcc929f5a67a4696cfbee37aca1/src/main/kotlin/org/wfanet/measurement/kingdom/service/internal/testing/EventGroupActivitiesServiceTest.kt#L1008)):*

There's no test for the MAX_PAGE_SIZE coercion. The implementation caps at 1000 via coerceAtMost, and the proto comment documents "values above this will be coerced to the maximum value." Add a test that creates >1 activity and requests pageSize 
  = 5000 to verify the cap is applied.
</blockquote></details>

done
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/internal/testing/EventGroupActivitiesServiceTest.kt` line 1067 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3909#-OuEX5yv0pWTC62XEJLp:-OuctztJ8ayo69zChtos:bb7f2mf) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/25d636150165efcc929f5a67a4696cfbee37aca1/src/main/kotlin/org/wfanet/measurement/kingdom/service/internal/testing/EventGroupActivitiesServiceTest.kt#L1067)):*

The pagination tests don't explicitly verify that items from the previous page are not repeated on the next page. The paginates with page_token test fetches page 1 (Dec 1) then page 2 (Dec 5, 10) — but doesn't assert Dec 1 is absent from page 2.
  The paginates with page_token and page_size test is better (chains through pages of size 1) but also doesn't assert non-overlap.
  
  Consider adding an assertion like:

  assertThat(secondResponse.eventGroupActivitiesList.map { it.date })
      .containsNoneIn(firstResponse.eventGroupActivitiesList.map { it.date })

  This guards against the cursor accidentally using >= instead of >.
</blockquote></details>

done
___
*[`src/main/proto/wfa/measurement/internal/kingdom/event_group_activities_service.proto` line 105 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3909#-OuEZPRkBfmAQJtXp0X_:-Oucxa4FBaUlhIMTz7NE:bb7f2mf) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/25d636150165efcc929f5a67a4696cfbee37aca1/src/main/proto/wfa/measurement/internal/kingdom/event_group_activities_service.proto#L105)):*

if you adopt my other comment - then the page token needs a tiebreaker. ActivityDate is unique per (DataProvider, EventGroup) but not across event groups — two event groups can have activities on the same date.
  
  Update the page token and query ordering:
```
  message ListEventGroupActivitiesPageToken {
    message After {
      google.type.Date date = 1;
      fixed64 external_event_group_id = 2;
    }
    After after = 1;
  }
```
  And update the query to ORDER BY ActivityDate ASC, ExternalEventGroupId ASC with a cursor condition of (ActivityDate, ExternalEventGroupId) > (@afterDate, @afterEventGroupId).
</blockquote></details>

done

**Review by stevenwarejones — COMMENTED** [review]
> @stevenwarejones reviewed 5 files and all commit messages, made 2 comments, and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3909)* status: all files reviewed, 5 unresolved discussions (waiting on afterdate and getina).

___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/internal/testing/EventGroupActivitiesServiceTest.kt` line 1007 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3909#-Oud0yBz87pjADL9G-FC:-Oud0yBz87pjADL9G-FD:bqrr70o) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/100e95d664cf4f88a688abf5d8df739e14b15830/src/main/kotlin/org/wfanet/measurement/kingdom/service/internal/testing/EventGroupActivitiesServiceTest.kt#L1007)):*
> ```Kotlin
> 
>       assertThat(exception.status.code).isEqualTo(Status.Code.INVALID_ARGUMENT)
>     }
> ```

  The filter redesign was specifically to support listing across multiple event groups in one RPC, but every test filters to a single event group. Add at least: (1) a test with an empty `external_event_group_ids` filter that
  creates activities across 2+ event groups and verifies all are returned, and (2) a test that filters to a subset of event groups. The empty-filter path also exercises the page token tiebreaker on `external_event_group_id`
  when two event groups have activities on the same date.
___
*[`src/main/proto/wfa/measurement/internal/kingdom/event_group_activities_service.proto` line 96 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3909#-Oud0eei01ZBq8yNDR9I:-Oud0eei01ZBq8yNDR9J:b-yfnvvu) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/100e95d664cf4f88a688abf5d8df739e14b15830/src/main/proto/wfa/measurement/internal/kingdom/event_group_activities_service.proto#L96)):*
> ```proto
>   message Filter {
>     // Filter by activity date range.
>     google.type.Interval date_interval = 1;
> ```

  The `Filter` message structure looks good, but `date_interval` is defined in the proto and has its BUILD dep added yet isn't implemented in the reader or service. A caller setting `date_interval` would get unfiltered results
  silently. Either implement it now (add `ActivityDate >= @startDate AND ActivityDate < @endDate` conjuncts in `readEventGroupActivities`) or remove the field and the `interval_proto` dep until it's ready — a silently ignored
  filter is worse than a missing one.

**Review by stevenwarejones — COMMENTED** [review]
> @stevenwarejones resolved 3 discussions.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3909)* status: all files reviewed, 2 unresolved discussions (waiting on getina).

**Review by getina — COMMENTED** [review]
> @getina made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3909)* status: 2 of 5 files reviewed, 2 unresolved discussions (waiting on stevenwarejones).

___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/internal/testing/EventGroupActivitiesServiceTest.kt` line 1007 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3909#-Oud0yBz87pjADL9G-FC:-Oui1mU07dN78KGK3Z-b:bb7f2mf) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/100e95d664cf4f88a688abf5d8df739e14b15830/src/main/kotlin/org/wfanet/measurement/kingdom/service/internal/testing/EventGroupActivitiesServiceTest.kt#L1007)):*

The filter redesign was specifically to support listing across multiple event groups in one RPC, but every test filters to a single event group. Add at least: (1) a test with an empty `external_event_group_ids` filter that
  creates activities across 2+ event groups and verifies all are returned, and (2) a test that filters to a subset of event groups. The empty-filter path also exercises the page token tiebreaker on `external_event_group_id`
  when two event groups have activities on the same date.
</blockquote></details>

done
___
*[`src/main/proto/wfa/measurement/internal/kingdom/event_group_activities_service.proto` line 96 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3909#-Oud0eei01ZBq8yNDR9I:-Oui1xWlCKPcjyi3KpWH:bb7f2mf) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/100e95d664cf4f88a688abf5d8df739e14b15830/src/main/proto/wfa/measurement/internal/kingdom/event_group_activities_service.proto#L96)):*

The `Filter` message structure looks good, but `date_interval` is defined in the proto and has its BUILD dep added yet isn't implemented in the reader or service. A caller setting `date_interval` would get unfiltered results
  silently. Either implement it now (add `ActivityDate >= @startDate AND ActivityDate < @endDate` conjuncts in `readEventGroupActivities`) or remove the field and the `interval_proto` dep until it's ready — a silently ignored
  filter is worse than a missing one.
</blockquote></details>

done

File: src/main/proto/wfa/measurement/internal/kingdom/event_group_activities_service.proto [hunk 1]
```diff
@@ -74,3 +79,41 @@ message BatchDeleteEventGroupActivitiesRequest {
...
+  // The maximum number of results to return.
+  //
+  // If unspecified, at most 50 results will be returned. The maximum value is
+  // 1000; values above this will be coerced to the maximum value.
+  int32 page_size = 2;
+  ListEventGroupActivitiesPageToken page_token = 3;
+
+  message Filter {
+    // Filter by activity date range.
+    google.type.Interval date_interval = 1;
```
stevenwarejones: Use DateInterval (from the API repo date_interval.proto) instead of google.type.Interval. ActivityDate is a Spanner DATE and google.type.Date in the proto — using Timestamps forces unnecessary conversion. DateInterval already exists for this purpose. The public API PR cross-media-measurement-api#281 should be updated to match.

stevenwarejones (changes requested):
See inline comment.

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3909)* status: 2 of 5 files reviewed, 3 unresolved discussions (waiting on getina and stevenwarejones).

stevenwarejones (commented):
still missing an issue trailer

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3909#-:-OuiDB1UEwVP56peCDhr:b-yl00pc)* status: 3 of 5 files reviewed, 3 unresolved discussions (waiting on getina).

stevenwarejones (commented):
> @stevenwarejones resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3909)* status: 3 of 5 files reviewed, 1 unresolved discussion (waiting on getina).

src/main/proto/wfa/measurement/internal/kingdom/event_group_activities_service.proto
getina: done

getina (commented):
can one issue be shared across the cross-media-measurement-api and cross-media-measurement repos?
i want to use https://github.com/world-federation-of-advertisers/cross-media-measurement-api/issues/282 

> @getina made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3909#-:-OuiLBUd1hyx-mjlzozd:b-j0wqdi)* status: 3 of 5 files reviewed, 1 unresolved discussion (waiting on stevenwarejones and stevenwarejones+AGNT).

stevenwarejones (commented):
yes - use same issue in both places - thats the best practice.

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3909#-:-OuiMZ5A5XBi_GV9DPEo:b4427ae)* status: 1 of 5 files reviewed, 1 unresolved discussion (waiting on stevenwarejones+AGNT).

getina (commented):
how do i reference an issue that's in a different repo in this PR?

"It only accepts Issue: #<number> (local repo issue) or Issue: NA. The full cross-repo format is rejected."

> @getina made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3909#-:-OuiNCiI6nPTBwILFdC2:b9jb20d)* status: 1 of 5 files reviewed, 1 unresolved discussion (waiting on stevenwarejones and stevenwarejones+AGNT).

stevenwarejones (approved):
All review comments verified and addressed. Still needs an Issue trailer before merge.

src/main/proto/wfa/measurement/internal/kingdom/event_group_activities_service.proto
stevenwarejones: Verified. Now uses internal kingdom DateInterval with startDate/endDate instead of google.type.Interval.

stevenwarejones (commented):
> @stevenwarejones+AGNT made 8 comments.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3909)* status: 1 of 5 files reviewed, 1 unresolved discussion (waiting on stevenwarejones).

___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/SpannerEventGroupActivitiesService.kt` line 219 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3909#-OuEZzubA5xMTKWfAecg:-OuiNvAnEedj2l_xigS3:b-8fgfb5) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/25d636150165efcc929f5a67a4696cfbee37aca1/src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/SpannerEventGroupActivitiesService.kt#L219)):*
<details><summary><i>Previously, getina (Tina Ge) wrote…</i></summary><blockquote>

done
</blockquote></details>

Verified. DataProviderReader validates existence and throws NOT_FOUND. Test covers non-existent DataProvider.
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/internal/testing/EventGroupActivitiesServiceTest.kt` line 1008 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3909#-OuEXEM4-z_DPJnbhNIq:-OuiNu-541m7tGfw3W2w:bedojj9) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/25d636150165efcc929f5a67a4696cfbee37aca1/src/main/kotlin/org/wfanet/measurement/kingdom/service/internal/testing/EventGroupActivitiesServiceTest.kt#L1008)):*
<details><summary><i>Previously, getina (Tina Ge) wrote…</i></summary><blockquote>

done
</blockquote></details>

Verified. Test with pageSize=5000 confirms coercion to MAX_PAGE_SIZE.
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/internal/testing/EventGroupActivitiesServiceTest.kt` line 1067 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3909#-OuEX5yv0pWTC62XEJLp:-OuiNtZR1dk_L0SCiqz1:b-ibejrf) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/25d636150165efcc929f5a67a4696cfbee37aca1/src/main/kotlin/org/wfanet/measurement/kingdom/service/internal/testing/EventGroupActivitiesServiceTest.kt#L1067)):*
<details><summary><i>Previously, getina (Tina Ge) wrote…</i></summary><blockquote>

done
</blockquote></details>

Verified. Pagination test now asserts containsNoneIn to confirm no overlap between pages.
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/internal/testing/EventGroupActivitiesServiceTest.kt` line 1007 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3909#-Oud0yBz87pjADL9G-FC:-OuiNvpADVR-UfcIl2sY:b-j0ke93) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/100e95d664cf4f88a688abf5d8df739e14b15830/src/main/kotlin/org/wfanet/measurement/kingdom/service/internal/testing/EventGroupActivitiesServiceTest.kt#L1007)):*
<details><summary><i>Previously, getina (Tina Ge) wrote…</i></summary><blockquote>

done
</blockquote></details>

Verified. Two new tests: empty filter across multiple event groups, and filter to subset. The empty-filter test also exercises the tiebreaker on same-date activities.
___
*[`src/main/proto/wfa/measurement/internal/kingdom/event_group_activities_service.proto` line 93 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3909#-OuEYAKb5dvg3GnbBVdP:-OuiNuTQAZoEazd_vPtP:b-whh28x) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/25d636150165efcc929f5a67a4696cfbee37aca1/src/main/proto/wfa/measurement/internal/kingdom/event_group_activities_service.proto#L93)):*

just following up - don't deprecate the field of course, just delete and rename all the fields since this PR introduces them.
</blockquote></details>

Verified. Filter message with external_event_group_ids and date_interval replaces the single event group field.
___
*[`src/main/proto/wfa/measurement/internal/kingdom/event_group_activities_service.proto` line 105 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3909#-OuEZPRkBfmAQJtXp0X_:-OuiNunR08MNFFYhc1xF:b-epk7zq) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/25d636150165efcc929f5a67a4696cfbee37aca1/src/main/proto/wfa/measurement/internal/kingdom/event_group_activities_service.proto#L105)):*
<details><summary><i>Previously, getina (Tina Ge) wrote…</i></summary><blockquote>

done
</blockquote></details>

Verified. Page token After message includes both date and external_event_group_id. ORDER BY and cursor condition are correct.
___
*[`src/main/proto/wfa/measurement/internal/kingdom/event_group_activities_service.proto` line 96 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3909#-Oud0eei01ZBq8yNDR9I:-OuiNvUw5qDivU4uiCxk:b8eapcl) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/100e95d664cf4f88a688abf5d8df739e14b15830/src/main/proto/wfa/measurement/internal/kingdom/event_group_activities_service.proto#L96)):*
<details><summary><i>Previously, getina (Tina Ge) wrote…</i></summary><blockquote>

done
</blockquote></details>

Verified. date_interval is now implemented in the reader with startDate >= and endDate <. Test covers filtering by date range.

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3909)* status: 3 of 5 files reviewed, all discussions resolved (waiting on getina).

stevenwarejones (commented):
> @stevenwarejones reviewed 1 file.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3909)* status: 4 of 5 files reviewed, all discussions resolved (waiting on getina).

stevenwarejones (approved):
> @stevenwarejones reviewed 1 file.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3909)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on getina).

SanjayVas (changes requested):
> @SanjayVas reviewed 1 file and all commit messages, and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3909)* status: all files reviewed, 2 unresolved discussions (waiting on getina).

___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/SpannerEventGroupActivitiesService.kt` line 235 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3909#-OuiXO3b2o9M9CtykfA8:-OuiXO3b2o9M9CtykfA9:biufmdl) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1f8d834c1f2a5cbd5b155c6d12af12e40ade3e5d/src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/SpannerEventGroupActivitiesService.kt#L235)):*
> ```Kotlin
>       DataProviderReader()
>         .readByExternalDataProviderId(
>           client.singleUse(),
> ```

We want to use a single transaction for the entirety of an internal method implementation. `singleUse()` should therefore only be used if your method has only one read. In this case, there are multiple reads, so you want to share a readonly transaction between them. e.g.

```kotlin
client.readOnlyTransaction().use { txn ->
  // Pass txn for all DB reads
}
```




_Code quote:_
```Kotlin
client.singleUse()
```
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/SpannerEventGroupActivitiesService.kt` line 240 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3909#-OuiYTou1TF8MNyIX6tT:-OuiYTou1TF8MNyIX6tU:b-1cnmww) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1f8d834c1f2a5cbd5b155c6d12af12e40ade3e5d/src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/SpannerEventGroupActivitiesService.kt#L240)):*
> ```Kotlin
>     if (dataProviderResult == null) {
>       throw DataProviderNotFoundException(ExternalId(request.externalDataProviderId))
>         .asStatusRuntimeException(Status.Code.NOT_FOUND, "DataProvider not found.")
> ```

nit: don't override the default message if you're just going to specify the same thing



_Code quote:_
```Kotlin
"DataProvider not found."
```

getina (commented):
> @getina made 2 comments and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3909)* status: all files reviewed, 1 unresolved discussion (waiting on SanjayVas).

___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/SpannerEventGroupActivitiesService.kt` line 235 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3909#-OuiXO3b2o9M9CtykfA8:-Oum8Wis5EhE9xkLus-l:b68sb5c) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1f8d834c1f2a5cbd5b155c6d12af12e40ade3e5d/src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/SpannerEventGroupActivitiesService.kt#L235)):*

We want to use a single transaction for the entirety of an internal method implementation. `singleUse()` should therefore only be used if your method has only one read. In this case, there are multiple reads, so you want to share a readonly transaction between them. e.g.

```kotlin
client.readOnlyTransaction().use { txn ->
  // Pass txn for all DB reads
}
```

</blockquote></details>

done - changed to readOnlyTransaction()
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/SpannerEventGroupActivitiesService.kt` line 240 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3909#-OuiYTou1TF8MNyIX6tT:-Oum8TaQ6_ZnXUJBHKEB:bb7f2mf) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1f8d834c1f2a5cbd5b155c6d12af12e40ade3e5d/src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/SpannerEventGroupActivitiesService.kt#L240)):*

nit: don't override the default message if you're just going to specify the same thing
</blockquote></details>

done

src/main/kotlin/org/wfanet/measurement/kingdom/service/internal/testing/EventGroupActivitiesServiceTest.kt
```diff
@@ -1004,6 +1007,588 @@ abstract class EventGroupActivitiesServiceTest<T : EventGroupActivitiesCoroutine
...
+      val exception =
+        assertFailsWith<StatusRuntimeException> {
+          eventGroupActivitiesService.listEventGroupActivities(listEventGroupActivitiesRequest {})
+        }
+
+      assertThat(exception.status.code).isEqualTo(Status.Code.INVALID_ARGUMENT)
+      assertThat(exception).hasMessageThat().contains("external_data_provider_id")
+    }
+
+  @Test
```
stevenwarejones, line 1202: Nit: Don't assert on exception message contents. Just assert the status code. The existing tests in this file do the same thing, but it couples the test to the message string rather than the error semantics.

src/main/kotlin/org/wfanet/measurement/kingdom/service/internal/testing/EventGroupActivitiesServiceTest.kt
```diff
@@ -1004,6 +1007,588 @@ abstract class EventGroupActivitiesServiceTest<T : EventGroupActivitiesCoroutine
...
+      assertFailsWith<StatusRuntimeException> {
+        eventGroupActivitiesService.listEventGroupActivities(
+          listEventGroupActivitiesRequest { externalDataProviderId = 999999L }
+        )
+      }
+
+    assertThat(exception.status.code).isEqualTo(Status.Code.NOT_FOUND)
+    assertThat(exception).hasMessageThat().contains("DataProvider not found")
+  }
+
```
stevenwarejones, line 1404: Nit: Same here — drop the hasMessageThat assertion, just assert the status code.

stevenwarejones (commented):
See inline nits.

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3909)* status: 4 of 5 files reviewed, 3 unresolved discussions (waiting on getina, SanjayVas, and stevenwarejones).

___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/SpannerEventGroupActivitiesService.kt` line 235 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3909#-OuiXO3b2o9M9CtykfA8:-OumOlw91ZoXOE4c5UrL:bnbv43h) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1f8d834c1f2a5cbd5b155c6d12af12e40ade3e5d/src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/SpannerEventGroupActivitiesService.kt#L235)):*
<details><summary><i>Previously, getina (Tina Ge) wrote…</i></summary><blockquote>

done - changed to readOnlyTransaction()
</blockquote></details>

Verified. DataProvider validation and activity read now share a single readOnlyTransaction.

stevenwarejones (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3909)* status: all files reviewed, 3 unresolved discussions (waiting on getina and SanjayVas).

src/main/kotlin/org/wfanet/measurement/kingdom/service/internal/testing/EventGroupActivitiesServiceTest.kt
SanjayVas: More specifically, assert on both the status code and on `exception.errorInfo` to verify that the correct error info is set. Newer tests can also use StatusExceptionSubject to get more detailed test failure messages.

See https://github.com/search?q=repo%3Aworld-federation-of-advertisers%2Fcross-media-measurement+path%3Akingdom%2Fservice%2Finternal+assertThat%28exception.errorInfo%29&type=code for examples of newer Kingdom internal service tests that assert on the error info and https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/main/src/test/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/ReportingSetsServiceTest.kt for example usage of StatusExceptionSubject.

SanjayVas (changes requested):
> @SanjayVas reviewed 4 files and all commit messages, made 3 comments, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3909)* status: all files reviewed, 4 unresolved discussions (waiting on getina, stevenwarejones, and stevenwarejones+AGNT).

___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/internal/testing/EventGroupActivitiesServiceTest.kt` line 1046 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3909#-OumavI-C2f32epV8ccQ:-OumavI-C2f32epV8ccR:b-l4m7xi) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/2ce172a303e6807734b35b3bb9abbb0a9cff0fb8/src/main/kotlin/org/wfanet/measurement/kingdom/service/internal/testing/EventGroupActivitiesServiceTest.kt#L1046)):*
> ```Kotlin
>           externalDataProviderId = dataProvider.externalDataProviderId
>           filter =
>             ListEventGroupActivitiesRequest.Filter.newBuilder()
> ```

Use Kotlin DSL for protobufs rather than Java builders.
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/internal/testing/EventGroupActivitiesServiceTest.kt` line 1052 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3909#-OumaawUA9m1IhSrFejn:-OumaawUA9m1IhSrFejo:b-a35f7d) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/2ce172a303e6807734b35b3bb9abbb0a9cff0fb8/src/main/kotlin/org/wfanet/measurement/kingdom/service/internal/testing/EventGroupActivitiesServiceTest.kt#L1052)):*
> ```Kotlin
>       )
> 
>     assertThat(response.eventGroupActivitiesList).hasSize(2)
> ```

Prefer asserting on the whole protobuf response. If there are fields that vary (e.g. generated IDs or real timestamps)  you can  explicitly ignore them and assert on them separately.

src/main/kotlin/org/wfanet/measurement/kingdom/service/internal/testing/EventGroupActivitiesServiceTest.kt
getina: done

src/main/kotlin/org/wfanet/measurement/kingdom/service/internal/testing/EventGroupActivitiesServiceTest.kt
getina: done

getina (commented):
> @getina made 4 comments and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3909)* status: all files reviewed, 2 unresolved discussions (waiting on SanjayVas and stevenwarejones).

___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/internal/testing/EventGroupActivitiesServiceTest.kt` line 1046 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3909#-OumavI-C2f32epV8ccQ:-OunN1-iAyyhd3iolLlD:bb7f2mf) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/2ce172a303e6807734b35b3bb9abbb0a9cff0fb8/src/main/kotlin/org/wfanet/measurement/kingdom/service/internal/testing/EventGroupActivitiesServiceTest.kt#L1046)):*

Use Kotlin DSL for protobufs rather than Java builders.
</blockquote></details>

done
___
*[`src/main/kotlin/org/wfanet/measurement/kingdom/service/internal/testing/EventGroupActivitiesServiceTest.kt` line 1052 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3909#-OumaawUA9m1IhSrFejn:-OunMv7A4oYnLya4MV_b:b-mrm8ba) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/2ce172a303e6807734b35b3bb9abbb0a9cff0fb8/src/main/kotlin/org/wfanet/measurement/kingdom/service/internal/testing/EventGroupActivitiesServiceTest.kt#L1052)):*

Prefer asserting on the whole protobuf response. If there are fields that vary (e.g. generated IDs or real timestamps)  you can  explicitly ignore them and assert on them separately.
</blockquote></details>

done - updated 7 tests that check response contents to assert the entire protobuf response

src/main/kotlin/org/wfanet/measurement/kingdom/service/internal/testing/EventGroupActivitiesServiceTest.kt
SanjayVas: This is asserting just on the metadataMap and then only that it sets a specific field. I was suggesting to assert on the whole errorInfo so that you can verify the reason/domain, which additionally serves the purpose you were trying to capture with asserting on the message content: disambiguate between different errors with the same status code.

SanjayVas (approved):
> @SanjayVas reviewed 1 file and all commit messages, made 1 comment, and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3909)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on stevenwarejones).

stevenwarejones (commented):
> @stevenwarejones reviewed all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3909)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on getina).

src/main/kotlin/org/wfanet/measurement/kingdom/service/internal/testing/EventGroupActivitiesServiceTest.kt
getina: done - changed to asserting on the entire errorInfo with domain, reason, and metadata

getina (commented):
> @getina made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3909)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on getina).

src/main/proto/wfa/measurement/internal/kingdom/event_group_activities_service.proto
```diff
@@ -74,3 +79,41 @@ message BatchDeleteEventGroupActivitiesRequest {
   // delete. A maximum of 1000 EventGroupActivity can be deleted in a batch.
   repeated google.type.Date external_event_group_activity_ids = 3;
 }
+
+message ListEventGroupActivitiesRequest {
+  // External ID of the parent DataProvider. Required.
+  fixed64 external_data_provider_id = 1;
```
stevenwarejones, line 85: Make `external_data_provider_id` optional. The public API (per cross-media-measurement-api#281) supports `dataProviders/-/eventGroups/-` to list across all DataProviders, which this internal RPC needs to serve. When unset, the service should list across all DataProviders; when set, scope to that DataProvider.

src/main/proto/wfa/measurement/internal/kingdom/event_group_activities_service.proto
```diff
@@ -74,3 +79,41 @@ message BatchDeleteEventGroupActivitiesRequest {
...
+  // 1000; values above this will be coerced to the maximum value.
+  int32 page_size = 2;
+  ListEventGroupActivitiesPageToken page_token = 3;
+
+  message Filter {
+    // Filter by activity date range.
+    DateInterval date_interval = 1;
+    // Filter to specific event groups. If empty, returns activities
+    // across all event groups for the data provider.
+    repeated fixed64 external_event_group_ids = 2;
```
stevenwarejones: Replace `repeated fixed64 external_event_group_ids` with `fixed64 external_event_group_id` (singular, optional). The public API no longer supports filtering to an arbitrary subset of EventGroups — the parent is `halo.wfanet.org/EventGroup` with AIP-159 wildcard ID support, so callers either pin to one EG or wildcard to all. Set → scope to that EG; unset → all EGs for the DataProvider scope.

stevenwarejones (commented):
Following up on the public API design in cross-media-measurement-api#281: the parent there should be `halo.wfanet.org/EventGroup` (matching the resource pattern), using AIP-159 wildcard IDs (`-`) for the cross-EG and cross-DP list cases. That changes what this internal RPC needs to support. Two blocking changes below.

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3909)* status: 4 of 6 files reviewed, 2 unresolved discussions (waiting on getina, SanjayVas, and stevenwarejones).

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3909)* status: all files reviewed, 2 unresolved discussions (waiting on getina).

src/main/proto/wfa/measurement/internal/kingdom/event_group_activities_service.proto
getina: done

src/main/proto/wfa/measurement/internal/kingdom/event_group_activities_service.proto
getina: done

getina (commented):
> @getina partially reviewed 1 file and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3909)* status: all files reviewed, 2 unresolved discussions (waiting on stevenwarejones and stevenwarejones+AGNT).

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3909)* status: 2 of 6 files reviewed, 3 unresolved discussions (waiting on getina, SanjayVas, and stevenwarejones+AGNT).

___
*[`src/main/proto/wfa/measurement/internal/kingdom/event_group_activities_service.proto` line 84 at r9](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3909#-OvBTHp8BsJQ5m50klAG:-OvBTHp8BsJQ5m50klAH:b-tnkwnq) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/42a5e4b19bb5baee70ff4128f8563dfd261a41a3/src/main/proto/wfa/measurement/internal/kingdom/event_group_activities_service.proto#L84)):*
> ```proto
> 
> message ListEventGroupActivitiesRequest {
>   // External ID of the parent DataProvider. If unset, lists across all
> ```

no, it should always filter to a single data provider

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3909)* status: 2 of 6 files reviewed, 3 unresolved discussions (waiting on getina, SanjayVas, and stevenwarejones+AGNT).

___
*[`src/main/proto/wfa/measurement/internal/kingdom/event_group_activities_service.proto` line 84 at r9](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3909#-OvBTHp8BsJQ5m50klAG:-OvBTfk_C2jT-U5g1UMu:b-vfgt52) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/42a5e4b19bb5baee70ff4128f8563dfd261a41a3/src/main/proto/wfa/measurement/internal/kingdom/event_group_activities_service.proto#L84)):*

no, it should always filter to a single data provider
</blockquote></details>

meaning one data provider can't query another data provider's event groups

src/main/proto/wfa/measurement/internal/kingdom/event_group_activities_service.proto
stevenwarejones: The doc on line 87 says "If unset, lists across all EventGroups…" but the field is still plain `fixed64`, so the implementation has to use sentinel `!= 0L` checks at SpannerEventGroupActivitiesService.kt:259 (passed through) and EventGroupActivityReader.kt:107. Add the `optional` keyword:

```proto
  // External ID of the parent EventGroup. If unset, lists across all
  // EventGroups for the scoped DataProvider (or all DataProviders).
  optional fixed64 external_event_group_id = 2;
```

Then switch the reader check to `if (externalEventGroupId != null)` (taking `Long?` instead of `Long = 0L`) and the call site to `if (request.hasExternalEventGroupId()) request.externalEventGroupId else null`. This makes "unset = wildcard" explicit at the type level and matches the doc's wording — the `optional` keyword exists for exactly this case (cf. `access/v1alpha/policies_service.proto:165`). The 0-sentinel pattern is fine for the required `external_event_group_id` fields elsewhere in this proto, but here 0 has a real semantic meaning ("wildcard"), so the sentinel collides.

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3909)* status: 2 of 6 files reviewed, 3 unresolved discussions (waiting on getina, SanjayVas, and stevenwarejones).

src/main/proto/wfa/measurement/internal/kingdom/event_group_activities_service.proto
stevenwarejones: see my other comment - this field should be have the `optional` keyword

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3909)* status: 2 of 6 files reviewed, 3 unresolved discussions (waiting on getina, SanjayVas, and stevenwarejones+AGNT).

src/main/proto/wfa/measurement/internal/kingdom/event_group_activities_service.proto
stevenwarejones: i think my comment went to the wrong spot. sorry. the `external_event_group_id` is the only option

stevenwarejones (changes requested):
> @stevenwarejones made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3909)* status: 2 of 6 files reviewed, 4 unresolved discussions (waiting on getina, SanjayVas, and stevenwarejones+AGNT).

___
*[`src/main/proto/wfa/measurement/internal/kingdom/event_group_activities_service.proto` line 89 at r9](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3909#-OvBUiLx7vODQY28o3i-:-OvBUiLx7vODQY28o3i0:b-4ppop7) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/42a5e4b19bb5baee70ff4128f8563dfd261a41a3/src/main/proto/wfa/measurement/internal/kingdom/event_group_activities_service.proto#L89)):*
> ```proto
>   // External ID of the parent EventGroup. If unset, lists across all
>   // EventGroups for the scoped DataProvider (or all DataProviders).
>   fixed64 external_event_group_id = 2;
> ```

make this `optional`

src/main/proto/wfa/measurement/internal/kingdom/event_group_activities_service.proto
getina: done

getina (commented):
> @getina made 3 comments.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3909)* status: 2 of 6 files reviewed, 4 unresolved discussions (waiting on SanjayVas, stevenwarejones, and stevenwarejones+AGNT).

___
*[`src/main/proto/wfa/measurement/internal/kingdom/event_group_activities_service.proto` line 84 at r9](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3909#-OvBTHp8BsJQ5m50klAG:-OvC2Buv5a5YiDflq_L9:bb7f2mf) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/42a5e4b19bb5baee70ff4128f8563dfd261a41a3/src/main/proto/wfa/measurement/internal/kingdom/event_group_activities_service.proto#L84)):*

meaning one data provider can't query another data provider's event groups
</blockquote></details>

done
___
*[`src/main/proto/wfa/measurement/internal/kingdom/event_group_activities_service.proto` line 89 at r9](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3909#-OvBUiLx7vODQY28o3i-:-OvBzUuMDn0a_6RZw_BR:bb7f2mf) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/42a5e4b19bb5baee70ff4128f8563dfd261a41a3/src/main/proto/wfa/measurement/internal/kingdom/event_group_activities_service.proto#L89)):*

make this `optional`
</blockquote></details>

done

src/main/kotlin/org/wfanet/measurement/kingdom/service/internal/testing/EventGroupActivitiesServiceTest.kt
stevenwarejones: Verified — test at line 1190+ now asserts the whole errorInfo (domain = KingdomInternalException.DOMAIN, reason, and metadata).

src/main/kotlin/org/wfanet/measurement/kingdom/service/internal/testing/EventGroupActivitiesServiceTest.kt
stevenwarejones: Verified — hasMessageThat assertion is gone; test now asserts on status code + errorInfo.

src/main/proto/wfa/measurement/internal/kingdom/event_group_activities_service.proto
stevenwarejones: Verified — `external_data_provider_id` is now required (proto line 84 doc'd "Required"; SpannerEventGroupActivitiesService.kt:242-246 throws INVALID_ARGUMENT when 0; DataProvider existence check at :250-256 is now unconditional). New test at EventGroupActivitiesServiceTest.kt:1182-1199 asserts the missing-field path with full errorInfo.

src/main/proto/wfa/measurement/internal/kingdom/event_group_activities_service.proto
stevenwarejones: Verified — `optional fixed64 external_event_group_id` at proto line 89. Reader signature now `externalEventGroupId: Long? = null` (EventGroupActivityReader.kt:96), called as `if (request.hasExternalEventGroupId()) request.externalEventGroupId else null` (SpannerEventGroupActivitiesService.kt:262). No more sentinel `!= 0L` check.

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3909)* status: 2 of 6 files reviewed, 4 unresolved discussions (waiting on getina, SanjayVas, and stevenwarejones).

___
*[`src/main/proto/wfa/measurement/internal/kingdom/event_group_activities_service.proto` line 84 at r9](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3909#-OvBTHp8BsJQ5m50klAG:-OvC4v68-79r_03eIetl:bkrbmdr) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/42a5e4b19bb5baee70ff4128f8563dfd261a41a3/src/main/proto/wfa/measurement/internal/kingdom/event_group_activities_service.proto#L84)):*
<details><summary><i>Previously, getina (Tina Ge) wrote…</i></summary><blockquote>

done
</blockquote></details>

Verified — `external_data_provider_id` is back to required (doc'd "Required" at proto line 84; INVALID_ARGUMENT when 0 at SpannerEventGroupActivitiesService.kt:242-246).

src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/SpannerEventGroupActivitiesService.kt
```diff
@@ -208,4 +217,80 @@ class SpannerEventGroupActivitiesService(
...
+    val resultList =
+      client.readOnlyTransaction().use { txn ->
+        // Validate DataProvider exists.
+        val dataProviderResult =
+          DataProviderReader()
+            .readByExternalDataProviderId(txn, ExternalId(request.externalDataProviderId))
+        if (dataProviderResult == null) {
+          throw DataProviderNotFoundException(ExternalId(request.externalDataProviderId))
+            .asStatusRuntimeException(Status.Code.NOT_FOUND)
+        }
```
stevenwarejones, line 257: When `external_event_group_id` is set but the EventGroup does not exist, this silently returns an empty list — same failure mode as the DataProvider case you already fixed in this PR (and same AIP-132 reasoning: NOT_FOUND for a missing parent at every level). Add an EventGroup-existence check inside the same readOnlyTransaction, after the DataProvider lookup:

```kotlin
if (request.hasExternalEventGroupId()) {
  val eventGroupResult =
    EventGroupReader()
      .readByDataProvider(
        txn,
        ExternalId(request.externalDataProviderId),
        ExternalId(request.externalEventGroupId),
      )
  if (eventGroupResult == null) {
    throw EventGroupNotFoundException(
        ExternalId(request.externalDataProviderId),
        ExternalId(request.externalEventGroupId),
      )
      .asStatusRuntimeException(Status.Code.NOT_FOUND)
  }
}
```

Add a test mirroring `listEventGroupActivities throws NOT_FOUND for non existent DataProvider` that sets a valid `externalDataProviderId` + an `externalEventGroupId = 999999L` and asserts NOT_FOUND with the full errorInfo (reason = `EVENT_GROUP_NOT_FOUND`, metadata `external_data_provider_id` + `external_event_group_id`).

stevenwarejones (commented):
Re-review of dc0e0be — one issue at a time.

src/main/kotlin/org/wfanet/measurement/kingdom/service/internal/testing/EventGroupActivitiesServiceTest.kt
```diff
@@ -1004,6 +1011,626 @@ abstract class EventGroupActivitiesServiceTest<T : EventGroupActivitiesCoroutine
...
+                month = 12
+                day = 1
+              }
+            }
+          }
+
+      assertThat(response.eventGroupActivitiesList)
+        .ignoringFields(EventGroupActivity.CREATE_TIME_FIELD_NUMBER)
+        .containsExactlyElementsIn(expectedActivities)
+        .inOrder()
```
stevenwarejones, line 1474: The composite page-token tiebreaker (adding `external_event_group_id` to `After`, from the r1 review discussion) was added specifically for the case where multiple EventGroups have activities on the same date — without the tiebreaker, the cursor `(ActivityDate, ExternalEventGroupId)` would either skip or repeat rows at the page boundary. None of the current pagination tests exercise that path:

- `paginates with page_token` and `paginates with page_token and page_size` — single EG, tiebreaker condition never triggers
- `returns activities across multiple event groups with empty filter` — multi-EG, but single page, no pagination boundary
- The `containsNoneIn` non-overlap assertion only runs in single-EG mode

Add a test:

1. Create 2 EventGroups under the same DataProvider
2. Create activities on both EGs on the same date (e.g., 2025-12-01) and on a later date (e.g., 2025-12-05)
3. Call `listEventGroupActivities` with `externalDataProviderId` only (wildcard EG) and `pageSize = 2`
4. Page through and assert:
   - Page 1 contains 2 activities, both on 2025-12-01, ordered by `externalEventGroupId` ASC
   - Next page token is set with `after.date = 2025-12-01` and `after.externalEventGroupId = (the larger of the two EG ids)`
   - Page 2 contains 2 activities, both on 2025-12-05
   - `containsNoneIn` across the two pages — no row repeated

If the cursor silently regressed to `>=` or dropped the OR-clause for the same-date case, the current tests would still pass; this one would fail.

stevenwarejones (commented):
Issue 2: missing test for composite page-token tiebreaker.

src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/SpannerEventGroupActivitiesService.kt
```diff
@@ -208,4 +217,80 @@ class SpannerEventGroupActivitiesService(
...
+
+    val dateInterval =
+      if (request.hasFilter() && request.filter.hasDateInterval()) {
+        request.filter.dateInterval
+      } else {
+        null
+      }
+
+    if (request.externalDataProviderId == 0L) {
+      throw RequiredFieldNotSetException("external_data_provider_id")
```
stevenwarejones: nit: Move the `request.externalDataProviderId == 0L` check up to the top of the method, before the pageSize/after/dateInterval computations. The other methods in this file (batchUpdate, delete, batchDelete) all do required-field checks first — this method is the odd one out, and there's no reason to compute the derived values just to throw INVALID_ARGUMENT immediately after.

stevenwarejones (commented):
Issue 3: validation order nit.

src/main/proto/wfa/measurement/internal/kingdom/event_group_activities_service.proto
```diff
@@ -74,3 +79,41 @@ message BatchDeleteEventGroupActivitiesRequest {
...
+  // The maximum number of results to return.
+  //
+  // If unspecified, at most 50 results will be returned. The maximum value is
+  // 1000; values above this will be coerced to the maximum value.
+  int32 page_size = 3;
+  ListEventGroupActivitiesPageToken page_token = 4;
+
+  message Filter {
+    // Filter by activity date range.
+    DateInterval date_interval = 1;
```
stevenwarejones, line 100: nit: Document the date_interval semantics. The reader implements half-open (`ActivityDate >= startDate AND ActivityDate < endDate`) at EventGroupActivityReader.kt:113-119, which matches the AIP-132 convention, but DateInterval is an internal type that doesn't itself enforce a convention — callers shouldn't have to read the reader to know whether endDate is inclusive.

```proto
  // Filter by activity date range. Half-open: includes activities on or after
  // `start_date` and strictly before `end_date`. Either bound may be unset
  // (open-ended).
  DateInterval date_interval = 1;
```

stevenwarejones (commented):
Issue 4: document date_interval semantics.

stevenwarejones (commented):
> @stevenwarejones reviewed 4 files and all commit messages, and resolved 4 discussions.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3909)* status: all files reviewed, 4 unresolved discussions (waiting on getina).

src/main/proto/wfa/measurement/internal/kingdom/event_group_activities_service.proto
SanjayVas: > this field should be have the `optional` keyword

I don't think this is correct. `optional` is for when we need to differentiate between the default value and unspecified. We still don't need to do that in this case as the default value of 0 is never a valid value for external EventGroup ID, so 0 can indeed be the same as unspecified.

SanjayVas (changes requested):
> @SanjayVas made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3909)* status: all files reviewed, 5 unresolved discussions (waiting on getina).

src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/SpannerEventGroupActivitiesService.kt
getina: done

src/main/kotlin/org/wfanet/measurement/kingdom/deploy/gcloud/spanner/SpannerEventGroupActivitiesService.kt
getina: done

src/main/kotlin/org/wfanet/measurement/kingdom/service/internal/testing/EventGroupActivitiesServiceTest.kt
getina: done

src/main/proto/wfa/measurement/internal/kingdom/event_group_activities_service.proto
getina: reverted back to required with wildcard being default 0

src/main/proto/wfa/measurement/internal/kingdom/event_group_activities_service.proto
getina: confirming if DateInterval can have either bound unset?

getina (commented):
> @getina made 5 comments and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3909)* status: all files reviewed, 4 unresolved discussions (waiting on SanjayVas and stevenwarejones).

src/main/proto/wfa/measurement/internal/kingdom/event_group_activities_service.proto
stevenwarejones: I prefer not to infer that 0 is invalid. but i'm okay with this.

src/main/proto/wfa/measurement/internal/kingdom/event_group_activities_service.proto
stevenwarejones: lets just go with a more generic comment
```
    // Filter by activity date range. Only activities with dates within
    // this interval are returned.
```

stevenwarejones (approved):
> @stevenwarejones reviewed 4 files and all commit messages, made 2 comments, and resolved 3 discussions.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3909)* status: all files reviewed, 1 unresolved discussion (waiting on SanjayVas).

stevenwarejones (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3909)* status: all files reviewed, 2 unresolved discussions (waiting on getina and SanjayVas).

stevenwarejones (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3909#-:-OvFhNXA3sbVSju9Zx4j:bnfp4nl)* status: all files reviewed, 2 unresolved discussions (waiting on getina and SanjayVas).

src/main/proto/wfa/measurement/internal/kingdom/event_group_activities_service.proto
getina: done

getina (commented):
> @getina made 1 comment and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3909)* status: all files reviewed, 1 unresolved discussion (waiting on SanjayVas).

src/main/proto/wfa/measurement/internal/kingdom/event_group_activities_service.proto
SanjayVas: > I prefer not to infer that 0 is invalid

It's explicitly invalid for external IDs in this API for exactly this reason. org.wfanet.measurement.common.IdGenerator generates non-zero IDs.

SanjayVas (approved):
> @SanjayVas reviewed 5 files and all commit messages, made 1 comment, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3909)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on getina).

stevenwarejones (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3909)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on getina).

Issue #3919, opened 2026-06-05
Remove deprecated ReportingUnitComponentSummary.event_group_summaries and its dual-write flag


georgi:
Follow-up cleanup for the EventGroup-to-ReportingSet migration in `ReportingUnitComponentSummary`.

Once consumers have migrated to the `reporting_set` reference (public) / `external_reporting_set_id` (internal), remove the deprecated inline EventGroup enumeration and the operator flag that gates it:

- Remove `ResultGroup.MetricMetadata.ReportingUnitComponentSummary.event_group_summaries` from the public `result_group.proto`.
- Remove `event_group_summaries` from the internal `result_group.proto` and stop writing it in `BasicReportProcessedResultsTransformation`.
- Remove the `--populate-deprecated-reporting-unit-event-group-summaries` flag from the Reporting public API server and the corresponding `populateDeprecatedReportingUnitEventGroupSummaries` plumbing in `BasicReportsService` / `BasicReportProtoConversions`.

This should only happen after the flag has been flipped to new-only (deprecated field empty) in all environments and no consumer depends on `event_group_summaries`.

Issue #3918, opened 2026-06-05
Replace inline EventGroup enumeration in ReportingUnitComponentSummary with a ReportingSet reference


georgi:
### Suggest an idea

basicReports responses can exceed the API gateway's 4 MiB limit, causing valid ListBasicReports and GetBasicReport requests to fail. The root cause is duplicated metadata: ResultGroup.MetricMetadata.ReportingUnitComponentSummary.event_group_summaries inline-lists every EventGroup a component entails, and this list is repeated in every result row.

PR #3920, opened 2026-06-05
feat(reporting): Add ReportingSet reference to ReportingUnitComponentSummary

georgi:
ResultGroup.MetricMetadata.ReportingUnitComponentSummary currently inlines an event_group_summaries block that enumerates every EventGroup in the component. Because this block is repeated in every result row of every BasicReport, the response grows O(number of EventGroups) per row and bloats large reports.

This PR replaces that inline enumeration with a reference to the ReportingSet whose membership is exactly the component's EventGroups, so callers resolve the roster once via GetReportingSet instead of receiving it repeated inline.

API changes
  - Add reporting_set to the public ReportingUnitComponentSummary and external_reporting_set_id to the internal message.
  - Deprecate event_group_summaries on both messages (deprecated = true).
  - The referenced ReportingSet already exists as the per-component primitive ReportingSet of the campaign group, so none is minted.

Behavior
  - reporting_set is always populated.
  - Emission of the deprecated event_group_summaries is controlled by a new opt-out request field, exclude_deprecated_event_group_summaries, on GetBasicReportRequest and ListBasicReportsRequest. It defaults to false, so the deprecated
  field is still populated unless a caller explicitly opts out — existing clients are unaffected and only opt in to the new, leaner response when they're ready. The field is annotated future_disposition = DEPRECATED and is itself
  temporary: it will be removed in a future release once the deprecated field is dropped.
  - This covers all BasicReport read paths via the shared conversion (BasicReportProtoConversions). GetReport is unaffected, since the public Report message does not carry this structure.

  A TODO (#3919) marks each site that still reads/writes the deprecated field so the deprecation warnings stay visible until it can be removed.

Issue: #3918


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3920)

georgi:
@claude review

src/test/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/BasicReportsServiceTest.kt
```diff
@@ -9449,6 +9453,12 @@ class BasicReportsServiceTest {
                           ResultGroupKt.MetricMetadataKt.reportingUnitComponentSummary {
```
stevenwarejones, line 9455: The service-layer tests only exercise the default (flag=true) path. There should be at least one test that constructs `BasicReportsService` with `populateDeprecatedReportingUnitEventGroupSummaries = false` and verifies that the returned `BasicReport` omits `event_group_summaries` while still populating `reporting_set`. The new `BasicReportProtoConversionsTest` covers the conversion layer, but the wiring from server flag → service → conversion is untested end-to-end.

src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/BasicReportProtoConversions.kt
```diff
@@ -402,8 +402,17 @@ fun ResultGroupMetricSpec.toInternal(): InternalResultGroupMetricSpec {
...
 
-/** Converts the internal [InternalBasicReport] to the public [BasicReport]. */
-fun InternalBasicReport.toBasicReport(): BasicReport {
+/**
+ * Converts the internal [InternalBasicReport] to the public [BasicReport].
+ *
+ * @param populateDeprecatedReportingUnitEventGroupSummaries whether to also emit the deprecated
+ *   [ResultGroup.MetricMetadata.ReportingUnitComponentSummary.event_group_summaries] field during
+ *   the EventGroup-to-ReportingSet migration. The replacement `reporting_set` field is always
+ *   populated regardless of this value.
```
stevenwarejones, line 411: `toBasicReport` has `populateDeprecatedReportingUnitEventGroupSummaries: Boolean = true`, but `toResultGroup` and `toMetricMetadata` correctly require the parameter with no default. A default on a migration-only parameter makes it easy to miss call sites that should be explicit. Remove the default so every call site makes a conscious choice, consistent with the other two functions.

src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/gcloud/spanner/BasicReportProcessedResultsTransformation.kt
```diff
@@ -151,6 +151,10 @@ object BasicReportProcessedResultsTransformation {
     }
   }
 
+  // Intentionally writes deprecated event_group_summaries alongside external_reporting_set_id
+  // during
```
stevenwarejones: Nit: the comment splits awkwardly across three lines with `// during` on its own. Reflow to two balanced lines or one line.

stevenwarejones (commented):
Reviewed against the "BasicReport API: Oversized Response" design doc. This correctly implements Option 3 (deprecate + dual-populate, drop later via operator flag). Three issues below.

src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/BasicReportProtoConversions.kt
```diff
@@ -764,17 +803,26 @@ fun InternalMetricMetadata.toMetricMetadata(): MetricMetadata {
...
-                eventGroupSummaries +=
-                  ResultGroupKt.MetricMetadataKt.ReportingUnitComponentSummaryKt.eventGroupSummary {
-                    eventGroup =
-                      MeasurementConsumerEventGroupKey(
-                          internalEventGroupSummary.cmmsMeasurementConsumerId,
-                          internalEventGroupSummary.cmmsEventGroupId,
-                        )
-                        .toName()
-                  }
+              reportingSet =
```
stevenwarejones, line 805: Bug: `reporting_set` is populated unconditionally, but for BasicReports created before this change, `externalReportingSetId` will be empty (proto3 default). This produces a malformed resource name like `measurementConsumers/1234/reportingSets/` when reading old reports via `GetBasicReport` or `ListBasicReports`.

Guard on `externalReportingSetId` being non-empty before setting `reportingSet`:

```kotlin
if (internalReportingUnitComponentSummary.externalReportingSetId.isNotEmpty()) {
  reportingSet =
    ReportingSetKey(
        cmmsMeasurementConsumerId,
        internalReportingUnitComponentSummary.externalReportingSetId,
      )
      .toName()
}
```

stevenwarejones (dismissed):
Requesting changes — see inline comments.

stevenwarejones (commented):
Withdrawn — posting inline instead.

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3920#-:-OuMhKCZEUyHq6nCKTtw:bdv03e0)* status: 0 of 12 files reviewed, 4 unresolved discussions (waiting on georgi and stevenwarejones).

stevenwarejones (commented):
> @stevenwarejones reviewed 12 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3920)* status: all files reviewed, 4 unresolved discussions (waiting on georgi).

src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/gcloud/spanner/BasicReportProcessedResultsTransformation.kt
stevenwarejones: .

src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/BasicReportProtoConversions.kt
stevenwarejones: .

src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/BasicReportProtoConversions.kt
stevenwarejones: .

src/test/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/BasicReportsServiceTest.kt
stevenwarejones: .

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3920)* status: all files reviewed, 4 unresolved discussions (waiting on stevenwarejones).

src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/BasicReportProtoConversions.kt
stevenwarejones: Withdrawn. I traced the read path and  calls  on every read for SUCCEEDED reports, which rebuilds  from the live campaign group ReportingSets via  → . So  is always populated from , not from the stored blob. Old reports are fine.

src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/BasicReportsService.kt
```diff
@@ -132,6 +132,7 @@ class BasicReportsService(
   private val measurementConsumerConfigs: MeasurementConsumerConfigs,
   private val defaultReportStartHour: ZonedHour? = null,
   private val baseExternalImpressionQualificationFilterIds: Iterable<String>,
+  private val populateDeprecatedReportingUnitEventGroupSummaries: Boolean = true,
```
stevenwarejones: The operator flag controls `event_group_summaries` server-wide, but callers who have already migrated to `reporting_set` have no way to opt out of the deprecated field on their own. Since `event_group_summaries` is the root cause of the oversized response problem (97% of payload per the design doc), there should be a request-level field on `GetBasicReportRequest` and `ListBasicReportsRequest` that lets individual callers suppress it — e.g. a bool or a view enum. That way a caller who has adopted `reporting_set` can immediately stop paying the response size cost without waiting for the operator to flip the server-wide flag for everyone.

stevenwarejones (commented):
See inline comment.

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3920)* status: all files reviewed, 6 unresolved discussions (waiting on georgi, kungfucraig, and stevenwarejones+AGNT).

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/common/ReportingApiServerFlags.kt` line 123 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3920#-OuN4MaEAr_xuWZNTJgo:-OuN4MaEAr_xuWZNTJgp:b-61da1o) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8b6f24f7fa9eaa4abb24a6ab93498e77ccbb0c43/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/common/ReportingApiServerFlags.kt#L123)):*
> ```Kotlin
>   // is dropped.
>   @set:CommandLine.Option(
>     names = ["--populate-deprecated-reporting-unit-event-group-summaries"],
> ```

@kungfucraig - this is not my preferred path. This is messy enough that we should just add a new field to the request that if set, drops the old bloated field from the response. In that scenario, there is no need for this flag.

stevenwarejones (changes requested):
> @stevenwarejones+AGNT resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3920)* status: all files reviewed, 5 unresolved discussions (waiting on georgi, kungfucraig, and stevenwarejones).

src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/BasicReportsService.kt
stevenwarejones: To clarify: the operator-level flag approach is fine — the more I thought about it, the dual-write with a server-wide toggle is the right migration strategy. The key missing piece is a request-level flag so that existing callers who have already adopted `reporting_set` can opt out of `event_group_summaries` as fast as possible, without waiting for the operator to flip the server-wide switch for everyone.

kungfucraig (commented):
> @kungfucraig reviewed 1 file and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3920)* status: all files reviewed, 5 unresolved discussions (waiting on georgi, stevenwarejones, and stevenwarejones+AGNT).

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/common/ReportingApiServerFlags.kt` line 123 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3920#-OuN4MaEAr_xuWZNTJgo:-OuNOi9L-56OBsga3Jd-:bceh4ta) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8b6f24f7fa9eaa4abb24a6ab93498e77ccbb0c43/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/common/ReportingApiServerFlags.kt#L123)):*

@kungfucraig - this is not my preferred path. This is messy enough that we should just add a new field to the request that if set, drops the old bloated field from the response. In that scenario, there is no need for this flag.
</blockquote></details>

I agree. I'd like to see a field on both the Get and List requests to control this behavior.

georgi:
@claude review

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3920)* status: 10 of 13 files reviewed, 5 unresolved discussions (waiting on georgi, kungfucraig, and stevenwarejones+AGNT).

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/common/ReportingApiServerFlags.kt` line 123 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3920#-OuN4MaEAr_xuWZNTJgo:-OuNlO1g-WvbjV3FFBWw:b-osqtte) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8b6f24f7fa9eaa4abb24a6ab93498e77ccbb0c43/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/common/ReportingApiServerFlags.kt#L123)):*

I agree. I'd like to see a field on both the Get and List requests to control this behavior.
</blockquote></details>

as agreed out-of-band, lets keep the flag but also allow a request-side field

stevenwarejones (commented):
> @stevenwarejones reviewed 1 file.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3920)* status: 11 of 13 files reviewed, 5 unresolved discussions (waiting on georgi, kungfucraig, and stevenwarejones+AGNT).

src/main/proto/wfa/measurement/reporting/v2alpha/result_group.proto
```diff
@@ -277,8 +277,27 @@ message ResultGroup {
...
       // A summary of each EventGroup entailed by the component.
-      repeated EventGroupSummary event_group_summaries = 3
-          [(google.api.field_behavior) = IMMUTABLE];
+      //
+      // Deprecated in favor of `reporting_set`, which references a ReportingSet
+      // whose membership enumerates the same EventGroups. This field is only
+      // emitted during the EventGroup-to-ReportingSet migration, when enabled by
+      // the server operator, and will be removed in a later CMMS release.
+      // TODO(world-federation-of-advertisers/cross-media-measurement#3919):
+      // Remove this field once consumers have migrated to `reporting_set`.
```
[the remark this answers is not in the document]
georgi: Done in 224db69da — the deprecation comments now name the explicit release: "will be removed in cross-media-measurement release 0.5.42" (in result_group.proto and the new request fields).

src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/BasicReportProtoConversions.kt
georgi: Done in 224db69da — removed the default from `toBasicReport`; all call sites now pass the parameter explicitly.

src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/gcloud/spanner/BasicReportProcessedResultsTransformation.kt
georgi: Done in 224db69da — reflowed to two balanced lines.

src/test/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/BasicReportsServiceTest.kt
georgi: Done in 224db69da — added a service-level test that constructs `BasicReportsService` with `populateDeprecatedReportingUnitEventGroupSummaries = false` and verifies the returned BasicReport omits `event_group_summaries` while still populating `reporting_set` (end-to-end server flag -> service -> conversion).

src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/BasicReportsService.kt
georgi: Added in 2e25fcb51 — `include_deprecated_event_group_summaries` (optional bool, default true) on both GetBasicReportRequest and ListBasicReportsRequest, ANDed with the server flag, with tests for opt-out/opt-in. The server-wide flag is retained per the out-of-band agreement.

georgi:
Addressed the review feedback in 2e25fcb51 and 224db69da:

- Request-level opt-out: added `include_deprecated_event_group_summaries` (optional bool, default true) to `GetBasicReportRequest` and `ListBasicReportsRequest`. The deprecated field is emitted only when the server flag allows it AND the request hasn't opted out, so callers already on `reporting_set` can stop paying the payload cost immediately. The server-wide flag is retained per the out-of-band agreement.
- Deprecation timeline: comments now name the explicit release ("cross-media-measurement release 0.5.42") instead of a vague "later release".
- `toBasicReport`: removed the default value so every call site is explicit, consistent with `toResultGroup`/`toMetricMetadata`.
- Reflowed the awkward comment in `BasicReportProcessedResultsTransformation`.
- Tests: added end-to-end coverage for both the server flag disabled (`populateDeprecatedReportingUnitEventGroupSummaries = false`) and the per-request opt-out/opt-in paths, asserting `event_group_summaries` is omitted while `reporting_set` stays populated.
- The `reporting_set`-empty concern was withdrawn after tracing the read path (it is always rebuilt from the live campaign group ReportingSets, never the stored blob).

src/test/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/BasicReportsServiceTest.kt
```diff
@@ -10000,6 +10028,87 @@ class BasicReportsServiceTest {
...
+        .first()
+    assertThat(componentSummaryWithDeprecated.eventGroupSummariesList).isNotEmpty()
+
+    // When the server is configured to omit the deprecated field, it is omitted regardless of the
+    // request, while reporting_set remains populated.
+    val serviceWithDeprecatedDisabled =
+      BasicReportsService(
+        internalBasicReportsService,
+        internalImpressionQualificationFiltersService,
+        internalReportingSetsService,
```
stevenwarejones: Missing test case: server flag=false with request=true. This is the case that verifies the server flag takes precedence over the request — if the AND logic were accidentally changed to OR, only this test would catch it. Add a test using `serviceWithDeprecatedDisabled` with `includeDeprecatedEventGroupSummaries = true` and assert that `eventGroupSummariesList` is still empty.

stevenwarejones (commented):
See inline comment.

stevenwarejones (commented):
> @stevenwarejones reviewed 7 files and all commit messages, made 3 comments, resolved 5 discussions, and dismissed @stevenwarejones+AGNT from 4 discussions.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3920)* status: all files reviewed, 4 unresolved discussions (waiting on georgi).

___
*[`src/main/proto/wfa/measurement/reporting/v2alpha/basic_reports_service.proto` line 81 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3920#-OubD0fh4Bo8zUUdoNyV:-OubD0fh4Bo8zUUdoNyW:b-h40o5y) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d20858a7cf6d271872db124603dccec9fb44d439/src/main/proto/wfa/measurement/reporting/v2alpha/basic_reports_service.proto#L81)):*
> ```proto
>   // to do so during the EventGroup-to-ReportingSet migration; use the
>   // `reporting_set` reference instead. This field will be removed in
>   // cross-media-measurement release 0.5.42.
> ```

nit: just say future version
___
*[`src/main/proto/wfa/measurement/reporting/v2alpha/basic_reports_service.proto` line 133 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3920#-OubD2svEoh_8Z0hzMr1:-OubD2svEoh_8Z0hzMr2:b-h40o5y) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d20858a7cf6d271872db124603dccec9fb44d439/src/main/proto/wfa/measurement/reporting/v2alpha/basic_reports_service.proto#L133)):*
> ```proto
>   // to do so during the EventGroup-to-ReportingSet migration; use the
>   // `reporting_set` reference instead. This field will be removed in
>   // cross-media-measurement release 0.5.42.
> ```

nit: just say future version
___
*[`src/main/proto/wfa/measurement/reporting/v2alpha/result_group.proto` line 285 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3920#-OubD5gM6s-P2B_e-rJK:-OubD5gM6s-P2B_e-rJL:b-h40o5y) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d20858a7cf6d271872db124603dccec9fb44d439/src/main/proto/wfa/measurement/reporting/v2alpha/result_group.proto#L285)):*
> ```proto
>       // emitted during the EventGroup-to-ReportingSet migration, when enabled
>       // by the server operator, and will be removed in cross-media-measurement
>       // release 0.5.42.
> ```

nit: just say future version

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3920)* status: all files reviewed, 4 unresolved discussions (waiting on georgi).

kungfucraig (approved):
> @kungfucraig reviewed 13 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3920)* status: all files reviewed, 4 unresolved discussions (waiting on georgi).

SanjayVas (changes requested):
> @SanjayVas reviewed 2 files and made 3 comments.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3920)* status: all files reviewed, 7 unresolved discussions (waiting on georgi).

___
*[`src/main/proto/wfa/measurement/reporting/v2alpha/basic_reports_service.proto` line 79 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3920#-OucrZDLD779DyKe8YFj:-OucrZDLD779DyKe8YFk:btukfyp) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d20858a7cf6d271872db124603dccec9fb44d439/src/main/proto/wfa/measurement/reporting/v2alpha/basic_reports_service.proto#L79)):*
> ```proto
>   //
>   // Even when `true`, the field is only emitted while the server is configured
>   // to do so during the EventGroup-to-ReportingSet migration; use the
> ```

Don't do it this way.

1. Invert the condition such that the natural default value of false is used.
2. Either have this field *or* have a server-side flag, not both. This is a temporary request field to delay the new behavior. We'll just drop the request field when the behavior changes.

    Yes, this means that clients that properly handle the change early will need to eventually change their request code or risk compilation failure. That's the natural consequence of having this be a temporary request field.



_Code quote:_
```proto
  // Whether to include the deprecated
  // `ResultGroup.MetricMetadata.ReportingUnitComponentSummary.event_group_summaries`
  // field in the response. Defaults to `true` when unset.
  //
  // Even when `true`, the field is only emitted while the server is configured
  // to do so during the EventGroup-to-ReportingSet migration; use the
```
___
*[`src/main/proto/wfa/measurement/reporting/v2alpha/basic_reports_service.proto` line 82 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3920#-OucqNcbAtvCgtQDjkON:-OucqNcbAtvCgtQDjkOO:bx8a5nh) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d20858a7cf6d271872db124603dccec9fb44d439/src/main/proto/wfa/measurement/reporting/v2alpha/basic_reports_service.proto#L82)):*
> ```proto
>   // `reporting_set` reference instead. This field will be removed in
>   // cross-media-measurement release 0.5.42.
>   optional bool include_deprecated_event_group_summaries = 2
> ```

Don't use the `optional` keyword as that's used to indicate when we need to differentiate between not specified and explicitly set to the default value. See above comment.
___
*[`src/main/proto/wfa/measurement/reporting/v2alpha/basic_reports_service.proto` line 83 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3920#-OucqV1UDEeE4PBlCpIT:-OucqV1UDEeE4PBlCpIU:b-r3rfuj) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d20858a7cf6d271872db124603dccec9fb44d439/src/main/proto/wfa/measurement/reporting/v2alpha/basic_reports_service.proto#L83)):*
> ```proto
>   // cross-media-measurement release 0.5.42.
>   optional bool include_deprecated_event_group_summaries = 2
>       [(google.api.field_behavior) = OPTIONAL];
> ```

Annotate this with a future disposition of DEPRECATED. See https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/main/src/main/proto/wfa/measurement/type/future_disposition.proto

src/test/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/BasicReportsServiceTest.kt
georgi: Done

src/test/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/BasicReportsServiceTest.kt
georgi: Done.

georgi (commented):
> @georgi reviewed 14 files and all commit messages, made 7 comments, resolved 2 discussions, and dismissed @kungfucraig, @stevenwarejones, and @stevenwarejones+AGNT from 5 discussions.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3920)* status: all files reviewed, 5 unresolved discussions (waiting on SanjayVas, stevenwarejones, and stevenwarejones+AGNT).

___
*[`src/main/proto/wfa/measurement/reporting/v2alpha/basic_reports_service.proto` line 79 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3920#-OucrZDLD779DyKe8YFj:-Oumx6eC9AHvq3nCSFGV:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d20858a7cf6d271872db124603dccec9fb44d439/src/main/proto/wfa/measurement/reporting/v2alpha/basic_reports_service.proto#L79)):*

Don't do it this way.

1. Invert the condition such that the natural default value of false is used.
2. Either have this field *or* have a server-side flag, not both. This is a temporary request field to delay the new behavior. We'll just drop the request field when the behavior changes.

    Yes, this means that clients that properly handle the change early will need to eventually change their request code or risk compilation failure. That's the natural consequence of having this be a temporary request field.
</blockquote></details>

Done.
___
*[`src/main/proto/wfa/measurement/reporting/v2alpha/basic_reports_service.proto` line 82 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3920#-OucqNcbAtvCgtQDjkON:-OumwjBE1EwK3w3-DtxJ:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d20858a7cf6d271872db124603dccec9fb44d439/src/main/proto/wfa/measurement/reporting/v2alpha/basic_reports_service.proto#L82)):*

Don't use the `optional` keyword as that's used to indicate when we need to differentiate between not specified and explicitly set to the default value. See above comment.
</blockquote></details>

Done.
___
*[`src/main/proto/wfa/measurement/reporting/v2alpha/basic_reports_service.proto` line 83 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3920#-OucqV1UDEeE4PBlCpIT:-Oumwmp29k3BtB-bUSpb:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d20858a7cf6d271872db124603dccec9fb44d439/src/main/proto/wfa/measurement/reporting/v2alpha/basic_reports_service.proto#L83)):*

Annotate this with a future disposition of DEPRECATED. See https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/main/src/main/proto/wfa/measurement/type/future_disposition.proto
</blockquote></details>

Done.
___
*[`src/main/proto/wfa/measurement/reporting/v2alpha/basic_reports_service.proto` line 133 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3920#-OubD2svEoh_8Z0hzMr1:-Oumwooq9h9B_qNNsbsK:b-tpdb0z) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d20858a7cf6d271872db124603dccec9fb44d439/src/main/proto/wfa/measurement/reporting/v2alpha/basic_reports_service.proto#L133)):*

nit: just say future version
</blockquote></details>

I think dev standards require to put a version there.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/common/ReportingApiServerFlags.kt` line 123 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3920#-OuN4MaEAr_xuWZNTJgo:-OuOh5Hg-9xrFKyuAJkL:bs7wumb) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8b6f24f7fa9eaa4abb24a6ab93498e77ccbb0c43/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/common/ReportingApiServerFlags.kt#L123)):*

as agreed out-of-band, lets keep the flag but also allow a request-side field
</blockquote></details>

Done.

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3920)* status: 13 of 15 files reviewed, 6 unresolved discussions (waiting on georgi, SanjayVas, and stevenwarejones+AGNT).

___
*[`.env` line 11 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3920#-Oun3ne3-kY5tIWf-fIh:-Oun3ne44Fa0mnAvxzcl:b-e4uub4) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ed017212d0c0730251dcf96efc0ee5c0c9cb653c/.env#L11)):*
> ```Dotenv
> AUTH0_AUDIENCE=https://mc-api-phase1-prd
> CLIENT_ID=0Q5sRZlbpI2VNALmbs3X9Upek1UpRgsG
> CLIENT_SECRET=DfgnZ-HtK2zEzx1cxK2qyhoKz5O6AW4AsLbpk3Vh2nWokuh79zeGEacGEPWLojAu
> ```

we'll need to revoke this key. :(

SanjayVas (changes requested):
> @SanjayVas partially reviewed 2 files, made 3 comments, and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3920)* status: 13 of 15 files reviewed, 5 unresolved discussions (waiting on georgi, stevenwarejones, and stevenwarejones+AGNT).

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/gcloud/spanner/BasicReportProcessedResultsTransformation.kt` line 156 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3920#-Oun-vCbBzrbQ-xAhXzb:-Oun-vCbBzrbQ-xAhXzc:b-2b34nz) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8b63ed90fd24ef5b34a2338a8df89ffec7e45391/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/gcloud/spanner/BasicReportProcessedResultsTransformation.kt#L156)):*
Add a TODO instead on the specific line to stop setting that field. From https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/main/docs/code-style.md:

> In cases where the warning cannot yet be addressed (e.g. a deprecated symbol that we can't stop using yet), add a TODO. **Do not** suppress the warning in this case, as it should remain visible until it can be addressed properly.
___
*[`src/main/proto/wfa/measurement/reporting/v2alpha/basic_reports_service.proto` line 79 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3920#-OucrZDLD779DyKe8YFj:-Oun3o113tTVBssZdVxf:b-1dl8mi) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d20858a7cf6d271872db124603dccec9fb44d439/src/main/proto/wfa/measurement/reporting/v2alpha/basic_reports_service.proto#L79)):*

Done.
</blockquote></details>

This didn't actually invert the condition. The field should be `exclude_deprecated_event_group_summaries` such that you get the existing behavior (populating the deprecated field) if the caller doesn't change any code.
___
*[`src/main/proto/wfa/measurement/reporting/v2alpha/basic_reports_service.proto` line 133 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3920#-OubD2svEoh_8Z0hzMr1:-Oun0wSfDp3aWvoNfyTK:bie2nla) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d20858a7cf6d271872db124603dccec9fb44d439/src/main/proto/wfa/measurement/reporting/v2alpha/basic_reports_service.proto#L133)):*

I think dev standards require to put a version there.
</blockquote></details>

Indeed the [API standards](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/main/docs/api-standards.md#deprecation-timelines) say this, but I don't think that standard actually follows from the linked PRs (meaning I think that entry in API standards is incorrect).

We have mentioned versions for deprecation in the past, but it's been more for trying to follow the [update policy](https://docs.google.com/document/d/1AYAJ1fIfFwbfwbAIl7xYrUsJZQKu_I38yqkki2s8zaE/edit?usp=sharing) of avoiding breaking EDP-side changes outside of minor version updates (e.g. documenting that something would be removed *no sooner* than the next minor version, and noting what the minor version would be). This is described in [Planned Breaking API Changes](https://docs.google.com/document/d/1ZTthFQIGKTTv7loknKBmVmKF3tPa9Q-z5HedgiRcPq0/edit?usp=sharing), which we unfortunately never got the cycles to prioritize. As a result, we're still on v0.5.x.

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3920)* status: 13 of 15 files reviewed, 5 unresolved discussions (waiting on georgi, kungfucraig, and stevenwarejones+AGNT).

___
*[`src/main/proto/wfa/measurement/reporting/v2alpha/basic_reports_service.proto` line 79 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3920#-OucrZDLD779DyKe8YFj:-Oun4xUI32TEw318g3E-:b-3i83ir) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d20858a7cf6d271872db124603dccec9fb44d439/src/main/proto/wfa/measurement/reporting/v2alpha/basic_reports_service.proto#L79)):*

This didn't actually invert the condition. The field should be `exclude_deprecated_event_group_summaries` such that you get the existing behavior (populating the deprecated field) if the caller doesn't change any code.
</blockquote></details>

+1 - this is broken now. We need the old callers to get the deprecated value by default and new callers to only not get the field if they set this field

Also - @SanjayVas - @kungfucraig and I already agreed to have a flag _and_ this which I agree is less than desirable but the market wants a flag and the consumers of this API need this field.

stevenwarejones (commented):
> @stevenwarejones reviewed 1 file and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3920)* status: 14 of 15 files reviewed, 5 unresolved discussions (waiting on georgi, kungfucraig, and stevenwarejones+AGNT).

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/common/ReportingApiServerFlags.kt` line 123 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3920#-OuN4MaEAr_xuWZNTJgo:-Oun644CBTaIrjw_lPEJ:bxzmkxc) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8b6f24f7fa9eaa4abb24a6ab93498e77ccbb0c43/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/common/ReportingApiServerFlags.kt#L123)):*

Done.
</blockquote></details>

no. we need this flag.

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3920)* status: 14 of 15 files reviewed, 6 unresolved discussions (waiting on georgi, kungfucraig, and stevenwarejones+AGNT).

src/test/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/BasicReportsServiceTest.kt
stevenwarejones: still need this done since test was reverted - see other comments

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3920)* status: 14 of 15 files reviewed, 6 unresolved discussions (waiting on georgi, kungfucraig, and stevenwarejones+AGNT).

src/test/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/BasicReportsServiceTest.kt
stevenwarejones: Regressed. The operator flag was removed from the BasicReportsService constructor in the latest commits, so the test that was added for this is gone. Once the operator flag is re-added (see discussion on basic_reports_service.proto), this test needs to come back: construct BasicReportsService with populateDeprecatedReportingUnitEventGroupSummaries = false and verify the response omits event_group_summaries while still populating reporting_set.

src/test/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/BasicReportsServiceTest.kt
stevenwarejones: Still needed. Once the operator flag is re-added, add this test: construct service with operator flag=false, send request with exclude=false (i.e. caller wants the deprecated field), and assert it is still omitted because the server overrides.

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3920)* status: 14 of 15 files reviewed, 7 unresolved discussions (waiting on georgi, kungfucraig, and stevenwarejones).

src/test/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/BasicReportsServiceTest.kt
georgi: As we remove the flag, not relevant. Done.

src/test/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/BasicReportsServiceTest.kt
georgi: Done.

georgi (commented):
> @georgi reviewed 7 files and all commit messages, made 6 comments, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3920)* status: all files reviewed, 6 unresolved discussions (waiting on kungfucraig, SanjayVas, stevenwarejones, and stevenwarejones+AGNT).

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/gcloud/spanner/BasicReportProcessedResultsTransformation.kt` line 156 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3920#-Oun-vCbBzrbQ-xAhXzb:-OuqSKbPDUtSld-qUL_L:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8b63ed90fd24ef5b34a2338a8df89ffec7e45391/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/gcloud/spanner/BasicReportProcessedResultsTransformation.kt#L156)):*

Add a TODO instead on the specific line to stop setting that field. From https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/main/docs/code-style.md:

> In cases where the warning cannot yet be addressed (e.g. a deprecated symbol that we can't stop using yet), add a TODO. **Do not** suppress the warning in this case, as it should remain visible until it can be addressed properly.
</blockquote></details>

Done.
___
*[`src/main/proto/wfa/measurement/reporting/v2alpha/basic_reports_service.proto` line 79 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3920#-OucrZDLD779DyKe8YFj:-OuqTlTV5y3m2Cycx7s3:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/d20858a7cf6d271872db124603dccec9fb44d439/src/main/proto/wfa/measurement/reporting/v2alpha/basic_reports_service.proto#L79)):*

+1 - this is broken now. We need the old callers to get the deprecated value by default and new callers to only not get the field if they set this field

Also - @SanjayVas - @kungfucraig and I already agreed to have a flag _and_ this which I agree is less than desirable but the market wants a flag and the consumers of this API need this field.
</blockquote></details>

Done.
___
*[`.env` line 11 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3920#-Oun3ne3-kY5tIWf-fIh:-OuqUWj65iF0rk1-uxQH:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ed017212d0c0730251dcf96efc0ee5c0c9cb653c/.env#L11)):*

we'll need to revoke this key. :(
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/common/ReportingApiServerFlags.kt` line 123 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3920#-OuN4MaEAr_xuWZNTJgo:-OuqU_Fa3syuLMpXunZm:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8b6f24f7fa9eaa4abb24a6ab93498e77ccbb0c43/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/common/ReportingApiServerFlags.kt#L123)):*

no. we need this flag.
</blockquote></details>

Done.

stevenwarejones (approved):
> @stevenwarejones reviewed 8 files and all commit messages, made 1 comment, and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3920)* status: all files reviewed, 4 unresolved discussions (waiting on kungfucraig, SanjayVas, and stevenwarejones+AGNT).

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/common/ReportingApiServerFlags.kt` line 123 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3920#-OuN4MaEAr_xuWZNTJgo:-OurSdc94HLsf-jQ0RAw:b-bocrn8) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/8b6f24f7fa9eaa4abb24a6ab93498e77ccbb0c43/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/common/ReportingApiServerFlags.kt#L123)):*

Done.
</blockquote></details>

as agree offline. agreed on dropping the key.

SanjayVas (approved):
> @SanjayVas reviewed 15 files and all commit messages, made 1 comment, and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3920)* status: all files reviewed, 3 unresolved discussions (waiting on georgi and stevenwarejones+AGNT).

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/BasicReportProtoConversions.kt` line 790 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3920#-OunDNtc-cyINpdWU-aJ:-OunDNtd11O49hr2fJsg:b-taedq8) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f1b479dbc85e8f474ade75c2eb197226764d2045/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/BasicReportProtoConversions.kt#L790)):*
> ```Kotlin
>  */
> // Intentionally reads/writes the deprecated event_group_summaries field during the migration.
> @Suppress("DEPRECATION")
> ```

Same here re: deprecation

src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/BasicReportsService.kt
```diff
@@ -409,7 +409,9 @@ class BasicReportsService(
       }
     )
 
-    return createdInternalBasicReport.toBasicReport()
+    return createdInternalBasicReport.toBasicReport(
+      populateDeprecatedReportingUnitEventGroupSummaries = false
```
stevenwarejones, line 413: Bug on idempotent retry, breaks consistency with Get/List.

SpannerBasicReportsService.createBasicReport checks for an existing report by request_id. If the report has already reached SUCCEEDED, it returns the full report with results rebuilt via withResults(). That response flows through this toBasicReport(false) call, which silently drops event_group_summaries with no way for the caller to opt in.

For first-time creates this is harmless (state is CREATED, no result groups), but on an idempotent retry of a succeeded report the caller gets a different response shape than they would from GetBasicReport with exclude_deprecated_event_group_summaries unset. Get and List both default to populating the deprecated field — Create should be consistent.

Add exclude_deprecated_event_group_summaries to CreateBasicReportRequest to match Get/List:

In basic_reports_service.proto:
```proto
message CreateBasicReportRequest {
  // ... existing fields ...

  // Whether to exclude the deprecated
  // ResultGroup.MetricMetadata.ReportingUnitComponentSummary.event_group_summaries
  // field from the response. When unset, the field is still populated; use the
  // reporting_set reference instead.
  //
  // This is a temporary opt-out to drop the deprecated field early during the
  // EventGroup-to-ReportingSet migration. It will be removed in a future
  // release.
  bool exclude_deprecated_event_group_summaries = 5 [
    (google.api.field_behavior) = OPTIONAL,
    (type.future_disposition) = DEPRECATED
  ];
}
```

In BasicReportsService.kt:
```kotlin
return createdInternalBasicReport.toBasicReport(
  populateDeprecatedReportingUnitEventGroupSummaries =
    !request.excludeDeprecatedEventGroupSummaries
)
```

stevenwarejones: we can also do this in a followup or since this is throwaway code, maybe not at all

stevenwarejones (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3920)* status: all files reviewed, 3 unresolved discussions (waiting on georgi and stevenwarejones+AGNT).

stevenwarejones (approved):
> @stevenwarejones resolved 2 discussions and dismissed @stevenwarejones+AGNT from 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3920)* status: all files reviewed, 1 unresolved discussion (waiting on georgi).

georgi (commented):
> @georgi reviewed 1 file and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3920)* status: all files reviewed, 1 unresolved discussion (waiting on SanjayVas).

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/BasicReportProtoConversions.kt` line 790 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3920#-OunDNtc-cyINpdWU-aJ:-OuvHCwuA8DkdKYfg3pe:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/f1b479dbc85e8f474ade75c2eb197226764d2045/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/BasicReportProtoConversions.kt#L790)):*

Same here re: deprecation
</blockquote></details>

Done.

georgi (commented):
> @georgi resolved 1 discussion and dismissed @SanjayVas from a discussion.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3920)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on georgi).

Issue #3824, opened 2026-05-22
Noise Corrector: Allow EDPs to deviate from the standard Pre-and Post-corrected Report Quality checks


khisba:
### Suggest an idea

We have an issue on the UK platform that one of our data providers sometimes reports the MRC direct (single EDP) reach to be slightly higher (1000 to 3000 absolute difference) than the AMI direct reach for otherwise equivalent reach metrics. This is accepted as an industry standard approach and should not be corrected (i.e. the data is accepted as correct). This EDP does not add noise to direct measurements and so the standard deviation is set to zero. This issue happens often. It impacts certain campaigns usually resulting in a number of report failures.

The current behaviour of the noise corrector is that it attempts to correct the reach values such that AMI reach is higher than MRC reach. The resulting correction always results in a LARGE_CORRECTION failure which causes the Noise Corrector to report an overall failure status and the report is set to failed status (in the case of a report created by the Halo API). In the current market-specific implementation we run the noise corrector in our stack and have been able to work around this, but we are about to move to creating all reports via the Halo API and this will no longer be possible. So an urgent fix is needed.

**Example (numbers changed):**
The AMI direct reach was reported as 6092000 and the equivalent MRC reach as 6094000.
The noise corrector attempts to correct this. It applies a + / - 2 (absolute) correction. This is not sufficient to resolve the issue. I assume because the standard deviation is 0 that constrains the available solutions.

```
"largeCorrections": [{
        "metricTitle": "measurementConsumers/XXX/metrics/AAA,
        "originalValue": "6094000",
        "correctedValue": "6093998"
      }, {
        "metricTitle": "measurementConsumers/XXX/metrics/BBB",
        "originalValue": "6092000",
        "correctedValue": "6092002"
      }
``` 

This results in an overall failure status reported from the noise corrector:
```
"issues": ["ZERO_VARIANCE_MEASUREMENTS_INCONSISTENT_PRE_CORRECTION", "ZERO_VARIANCE_MEASUREMENTS_INCONSISTENT_POST_CORRECTION", "HAS_LARGE_CORRECTIONS"] 
``` 

This is the only type of issue that we have come across from this EDP. And it has been confirmed that the deviation (MRC > AMI reach) is expected and should be tolerated.

I therefore suggest that we either:
1. Enable direct report validation checks to be completely switched off for certain EDPs, on a configurable basis; OR
2. Enable a configurable tolerance on the validation checks for certain EDPs.

[...]

PR #3942, opened 2026-06-08
feat(report-processor): Add ami mrc exempted edps to basic report

ple13:
Issue: #3824


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3942)

kungfucraig (changes requested):
> @kungfucraig reviewed 20 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3942)* status: all files reviewed, 1 unresolved discussion (waiting on ple13 and SanjayVas).

___
*[`src/main/proto/wfa/measurement/reporting/v2alpha/basic_report.proto` line 136 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3942#-OuctYtC-aCwSCgbR8vq:-OuctYtDD2Pn_ap53nTm:beyt8ey) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5c2747b20a996d85c8bcf3dc4d21f4791c8a20e1/src/main/proto/wfa/measurement/reporting/v2alpha/basic_report.proto#L136)):*
> ```proto
>   // The list of data provider resource names for which AMI vs MRC consistency
>   // check is disabled.
>   repeated string ami_mrc_exempted_edps = 14;
> ```

Per offline discussion remove this and introduce a flag in the public reporting server.

ple13 (commented):
> @ple13 made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3942)* status: 16 of 24 files reviewed, 1 unresolved discussion (waiting on kungfucraig and SanjayVas).

___
*[`src/main/proto/wfa/measurement/reporting/v2alpha/basic_report.proto` line 136 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3942#-OuctYtC-aCwSCgbR8vq:-Oud2Bfp54EA-Egudx2C:bj54262) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5c2747b20a996d85c8bcf3dc4d21f4791c8a20e1/src/main/proto/wfa/measurement/reporting/v2alpha/basic_report.proto#L136)):*

Per offline discussion remove this and introduce a flag in the public reporting server.
</blockquote></details>

I've removed this field here and introduced a flag in the public API.

SanjayVas (changes requested):
> @SanjayVas reviewed 13 files and all commit messages, and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3942)* status: 20 of 24 files reviewed, 3 unresolved discussions (waiting on kungfucraig and ple13).

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/common/ReportingApiServerFlags.kt` line 102 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3942#-OudR8Fi2vPV1O0z4qo1:-OudR8Fi2vPV1O0z4qo2:b-axlwct) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1cb0e417c224572bcf5e8156e540b39488323b66/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/common/ReportingApiServerFlags.kt#L102)):*
> ```Kotlin
> 
>   @CommandLine.Option(
>     names = ["--ami-mrc-exempted-edps"],
> ```

Use a singular flag name and document that this can be specified multiple times.



_Code quote:_
```Kotlin
--ami-mrc-exempted-edps
```
___
*[`src/main/proto/wfa/measurement/internal/reporting/v2/basic_report.proto` line 103 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3942#-OudS6qd3PaIASVgzVnz:-OudS6qd3PaIASVgzVo-:b-jqu7tq) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1cb0e417c224572bcf5e8156e540b39488323b66/src/main/proto/wfa/measurement/internal/reporting/v2/basic_report.proto#L103)):*
> ```proto
>   // The list of data provider resource names for which AMI vs MRC consistency
>   // check is disabled.
>   repeated string ami_mrc_exempted_edps = 15;
> ```

We use CMMS DataProvider IDs in the internal Reporting API.



_Suggestion:_
```proto
ami_mrc_exempted_cmms_data_provider_ids
```

ple13 (commented):
> @ple13 made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3942)* status: 13 of 24 files reviewed, 3 unresolved discussions (waiting on kungfucraig and SanjayVas).

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/common/ReportingApiServerFlags.kt` line 102 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3942#-OudR8Fi2vPV1O0z4qo1:-OuhRRTS5AHDWfXp_uGe:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1cb0e417c224572bcf5e8156e540b39488323b66/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/common/ReportingApiServerFlags.kt#L102)):*

Use a singular flag name and document that this can be specified multiple times.
</blockquote></details>

Done.
___
*[`src/main/proto/wfa/measurement/internal/reporting/v2/basic_report.proto` line 103 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3942#-OudS6qd3PaIASVgzVnz:-OuhRUkR0GSgU06lWedh:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1cb0e417c224572bcf5e8156e540b39488323b66/src/main/proto/wfa/measurement/internal/reporting/v2/basic_report.proto#L103)):*

We use CMMS DataProvider IDs in the internal Reporting API.
</blockquote></details>

Done.

kungfucraig (approved):
> @kungfucraig reviewed 13 files and all commit messages, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3942)* status: all files reviewed, 2 unresolved discussions (waiting on SanjayVas).

SanjayVas (changes requested):
> @SanjayVas reviewed 6 files and all commit messages, made 3 comments, and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3942)* status: all files reviewed, 3 unresolved discussions (waiting on ple13).

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/gcloud/spanner/db/BasicReports.kt` line 339 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3942#-OuhTrUzAMt4IkV07ZIq:-OuhTrUzAMt4IkV07ZIr:b-cdvlhs) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ca2882ed5c5273cddbbc5dd293858f633163fb89/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/gcloud/spanner/db/BasicReports.kt#L339)):*
> ```Kotlin
>       externalReportResultId = row.getLong("ExternalReportResultId")
>     }
>     amiMrcExemptedCmmsDataProviderIds += row.getStringList("AmiMrcExemptedEdps")
> ```

You need to handle the conversion between resource names and resource IDs somewhere. The flag takes resource names but the internal API uses and persists just the IDs from those names.
___
*[`src/main/python/wfa/measurement/reporting/job/post_process_report_result_job.py` line 81 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3942#-OuhSrVy4N7Dt1B6ztPo:-OuhSrVy4N7Dt1B6ztPp:b1lhqsf) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ca2882ed5c5273cddbbc5dd293858f633163fb89/src/main/python/wfa/measurement/reporting/job/post_process_report_result_job.py#L81)):*
> ```Python
>                 basic_report.cmms_measurement_consumer_id,
>                 basic_report.external_report_result_id,
>                 list(basic_report.ami_mrc_exempted_cmms_data_provider_ids),
> ```

You're now passing in IDs, but I think the function still expects resource names. You either need to change the function or convert between IDs and resource names somewhere.
___
*[`src/main/resources/reporting/spanner/add-ami-mrc-exempted-edps-to-basic-report.sql` line 20 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3942#-OuhSSq1BTdXL3GaDEn3:-OuhSSq1BTdXL3GaDEn4:bije5x7) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ca2882ed5c5273cddbbc5dd293858f633163fb89/src/main/resources/reporting/spanner/add-ami-mrc-exempted-edps-to-basic-report.sql#L20)):*
> ```TSQL
> -- comment: Add amiMrcExemptedEdps to BasicReport
> 
> ALTER TABLE BasicReports ADD COLUMN AmiMrcExemptedEdps ARRAY<STRING(MAX)>;
> ```

Rename the column to match the field in the internal API so it's clear what these are (IDs from the CMMS API).



_Code quote:_
```TSQL
AmiMrcExemptedEdps
```

src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/gcloud/spanner/db/BasicReports.kt
```diff
@@ -335,6 +336,7 @@ private fun buildBasicReport(row: Struct): BasicReport {
     if (!row.isNull("ExternalReportResultId")) {
       externalReportResultId = row.getLong("ExternalReportResultId")
     }
+    amiMrcExemptedCmmsDataProviderIds += row.getStringList("AmiMrcExemptedEdps")
```
stevenwarejones: `Struct.getStringList` throws NPE on a NULL Spanner column. The new `AmiMrcExemptedEdps` column is nullable (Cloud Spanner cannot add a NOT NULL column to an existing table without a default expression, so nullable is the right call), but that means any existing BasicReport row will read NULL here and crash. The convention in this file for nullable columns is an `isNull` guard — see lines 288, 321, 324, 327, 336. Suggest:\n\n```kotlin\nif (!row.isNull("AmiMrcExemptedEdps")) {\n  amiMrcExemptedCmmsDataProviderIds += row.getStringList("AmiMrcExemptedEdps")\n}\n```

stevenwarejones (changes requested):
> @stevenwarejones reviewed 4 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3942)* status: all files reviewed, 5 unresolved discussions (waiting on ple13).

___
*[`src/main/proto/wfa/measurement/internal/reporting/v2/basic_report.proto` line 103 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3942#-Ous29Ny18TSIcTUosSC:-Ous29Ny18TSIcTUosSD:b-su4qym) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ca2882ed5c5273cddbbc5dd293858f633163fb89/src/main/proto/wfa/measurement/internal/reporting/v2/basic_report.proto#L103)):*
> ```proto
>   // The list of data provider resource names for which AMI vs MRC consistency
>   // check is disabled.
>   repeated string ami_mrc_exempted_cmms_data_provider_ids = 15;
> ```

is this just meant to be an audit record of who was ami mrc exempted at report creation time? its a little confusing to me why we need this field.

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3942)* status: all files reviewed, 5 unresolved discussions (waiting on ple13 and SanjayVas).

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/gcloud/spanner/db/BasicReports.kt` line 339 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3942#-OuhTrUzAMt4IkV07ZIq:-Ourzb6y51fvvqdiDVbL:b8sfes3) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ca2882ed5c5273cddbbc5dd293858f633163fb89/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/gcloud/spanner/db/BasicReports.kt#L339)):*

You need to handle the conversion between resource names and resource IDs somewhere. The flag takes resource names but the internal API uses and persists just the IDs from those names.
</blockquote></details>

+1. Prefer parsing the flag once at startup (in `V2AlphaPublicApiServer.kt`) into bare IDs so malformed config fails fast rather than on every request — the service and `toInternal` then deal in IDs only.

stevenwarejones (commented):
> @stevenwarejones reviewed 1 file and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3942)* status: all files reviewed, 5 unresolved discussions (waiting on ple13 and SanjayVas).

___
*[`src/main/resources/reporting/spanner/add-ami-mrc-exempted-edps-to-basic-report.sql` line 20 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3942#-OuhSSq1BTdXL3GaDEn3:-Ous2pGQ7VulLujvUX6x:bfs2zsr) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ca2882ed5c5273cddbbc5dd293858f633163fb89/src/main/resources/reporting/spanner/add-ami-mrc-exempted-edps-to-basic-report.sql#L20)):*

Rename the column to match the field in the internal API so it's clear what these are (IDs from the CMMS API).
</blockquote></details>

ditto on my other question - why do we need to store this in the db?

src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/gcloud/spanner/db/BasicReports.kt
ple13: Thanks. I've fixed it.

ple13 (commented):
> @ple13 made 5 comments.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3942)* status: 8 of 28 files reviewed, 5 unresolved discussions (waiting on kungfucraig, SanjayVas, stevenwarejones, and stevenwarejones+AGNT).

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/gcloud/spanner/db/BasicReports.kt` line 339 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3942#-OuhTrUzAMt4IkV07ZIq:-Ouv4DslACRdaWY7ukc_:b-i21bbo) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ca2882ed5c5273cddbbc5dd293858f633163fb89/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/gcloud/spanner/db/BasicReports.kt#L339)):*
<details><summary><i>Previously, stevenwarejones+AGNT (Agent AGNT for Steven Ware Jones) wrote…</i></summary><blockquote>

+1. Prefer parsing the flag once at startup (in `V2AlphaPublicApiServer.kt`) into bare IDs so malformed config fails fast rather than on every request — the service and `toInternal` then deal in IDs only.
</blockquote></details>

Thanks. Done.
___
*[`src/main/proto/wfa/measurement/internal/reporting/v2/basic_report.proto` line 103 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3942#-Ous29Ny18TSIcTUosSC:-Ouv5MFc00EMQutDNaDH:b-3pehk7) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ca2882ed5c5273cddbbc5dd293858f633163fb89/src/main/proto/wfa/measurement/internal/reporting/v2/basic_report.proto#L103)):*

is this just meant to be an audit record of who was ami mrc exempted at report creation time? its a little confusing to me why we need this field.
</blockquote></details>

We need this field due to issue #3824.
___
*[`src/main/python/wfa/measurement/reporting/job/post_process_report_result_job.py` line 81 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3942#-OuhSrVy4N7Dt1B6ztPo:-OuuLaxd1jqmlOIXedcn:bpa9ywo) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ca2882ed5c5273cddbbc5dd293858f633163fb89/src/main/python/wfa/measurement/reporting/job/post_process_report_result_job.py#L81)):*

You're now passing in IDs, but I think the function still expects resource names. You either need to change the function or convert between IDs and resource names somewhere.
</blockquote></details>

Thanks for catching this. In API v2, we use the mapping (external_reporting_set_id -> metrics). I've updated the code to map (edp_name -> metrics). The API for report processor remains unchanged.
___
*[`src/main/resources/reporting/spanner/add-ami-mrc-exempted-edps-to-basic-report.sql` line 20 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3942#-OuhSSq1BTdXL3GaDEn3:-Ouv5c_pCN7dfILB7uH_:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ca2882ed5c5273cddbbc5dd293858f633163fb89/src/main/resources/reporting/spanner/add-ami-mrc-exempted-edps-to-basic-report.sql#L20)):*

ditto on my other question - why do we need to store this in the db?
</blockquote></details>

Done.

src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/gcloud/spanner/db/BasicReports.kt
SanjayVas: Thanks for catching that @stevenwarejones!

SanjayVas (changes requested):
> @SanjayVas partially reviewed 12 files and all commit messages, made 3 comments, and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3942)* status: 19 of 28 files reviewed, 4 unresolved discussions (waiting on kungfucraig, ple13, stevenwarejones, and stevenwarejones+AGNT).

___
*[`src/main/proto/wfa/measurement/internal/reporting/v2/basic_report.proto` line 103 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3942#-Ous29Ny18TSIcTUosSC:-OuxADTV0bRlxbM-vi9X:b-i0hts7) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ca2882ed5c5273cddbbc5dd293858f633163fb89/src/main/proto/wfa/measurement/internal/reporting/v2/basic_report.proto#L103)):*

We need this field due to issue #3824.
</blockquote></details>

The BasicReports row is written on create. The report post-processor later calls the internal API to read the BasicReport. If we want to have the exempted EDPs be decided at creation time then it needs to be persisted as part of the BasicReport.

The "audit record" is a secondary benefit of doing it this way.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/BasicReportProtoConversions.kt` line 151 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3942#-OuxAqvc-zPcaRmFDqV_:-OuxAqvc-zPcaRmFDqVa:b-kskb6l) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/dece6b6e279dad74da8ba25ba36de8c0d2957696/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/BasicReportProtoConversions.kt#L151)):*
> ```Kotlin
>     this.amiMrcExemptedCmmsDataProviderIds +=
>       amiMrcExemptedEdps.map {
>         requireNotNull(DataProviderKey.fromName(it)?.dataProviderId) { "Invalid EDP name $it" }
> ```

I'd prefer this happening earlier so we can fail fast. @stevenwarejones also made this recommendation.

e.g. handle this in ReportingApiServerFlags using custom validation. See https://picocli.info/#_single_value_validation

```kotlin
@CommandLine.Spec
private lateinit var spec: CommandLine.CommandSpec

lateinit var amiMrcExemptedCmmsDataProviderIds: List<String>
  private set

@CommandLine.Option(
  names = ["--ami-mrc-exempted-edp"],
  // ... 
)
private fun setAmiMrcExemptedCmmsDataProviderIds(values: List<String>) {
  amiMrcExemptedCmmsDataProviderIds = values.map {
    DataProviderKey.fromName(it)?.dataProviderId ?: throw CommandLine.ParameterException(spec.commandLine(), "$it is not a valid DataProvider resource name")
  }
}

```

stevenwarejones (commented):
> @stevenwarejones partially reviewed 26 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3942)* status: all files reviewed, 4 unresolved discussions (waiting on kungfucraig, ple13, stevenwarejones, and stevenwarejones+AGNT).

src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/gcloud/spanner/db/BasicReports.kt
stevenwarejones: Verified — `isNull` guard added at `BasicReports.kt:340-342`. Thanks.

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3942)* status: all files reviewed, 4 unresolved discussions (waiting on kungfucraig, ple13, SanjayVas, and stevenwarejones).

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/gcloud/spanner/db/BasicReports.kt` line 339 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3942#-OuhTrUzAMt4IkV07ZIq:-OuxGFjP424nIHLGUFDI:b-dlheyz) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ca2882ed5c5273cddbbc5dd293858f633163fb89/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/gcloud/spanner/db/BasicReports.kt#L339)):*

Thanks. Done.
</blockquote></details>

I don't see this in the diff — conversion is happening in `BasicReportProtoConversions.toInternal:148-151` per-request, not at startup. `ReportingApiServerFlags.amiMrcExemptedEdps` is still a `List<String>` of resource names and `V2AlphaPublicApiServer.kt:423` passes it through unchanged. Sanjay raised the same point on R6 with a concrete picocli pattern.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/BasicReportProtoConversions.kt` line 151 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3942#-OuxAqvc-zPcaRmFDqV_:-OuxGTb29haol5TF2O5r:b32v57) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/dece6b6e279dad74da8ba25ba36de8c0d2957696/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/BasicReportProtoConversions.kt#L151)):*

I'd prefer this happening earlier so we can fail fast. @stevenwarejones also made this recommendation.

e.g. handle this in ReportingApiServerFlags using custom validation. See https://picocli.info/#_single_value_validation

```kotlin
@CommandLine.Spec
private lateinit var spec: CommandLine.CommandSpec

lateinit var amiMrcExemptedCmmsDataProviderIds: List<String>
  private set

@CommandLine.Option(
  names = ["--ami-mrc-exempted-edp"],
  // ... 
)
private fun setAmiMrcExemptedCmmsDataProviderIds(values: List<String>) {
  amiMrcExemptedCmmsDataProviderIds = values.map {
    DataProviderKey.fromName(it)?.dataProviderId ?: throw CommandLine.ParameterException(spec.commandLine(), "$it is not a valid DataProvider resource name")
  }
}

```
</blockquote></details>

+1

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3942)* status: all files reviewed, 3 unresolved discussions (waiting on kungfucraig, ple13, and SanjayVas).

___
*[`src/main/proto/wfa/measurement/internal/reporting/v2/basic_report.proto` line 103 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3942#-Ous29Ny18TSIcTUosSC:-OuxG_3I-9edMyNKbgkP:b-95k4r4) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ca2882ed5c5273cddbbc5dd293858f633163fb89/src/main/proto/wfa/measurement/internal/reporting/v2/basic_report.proto#L103)):*

The BasicReports row is written on create. The report post-processor later calls the internal API to read the BasicReport. If we want to have the exempted EDPs be decided at creation time then it needs to be persisted as part of the BasicReport.

The "audit record" is a secondary benefit of doing it this way.
</blockquote></details>

I'd prefer to just simplify and read the list at post processing time.

SanjayVas (commented):
> @SanjayVas made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3942)* status: all files reviewed, 3 unresolved discussions (waiting on kungfucraig, ple13, and stevenwarejones).

___
*[`src/main/proto/wfa/measurement/internal/reporting/v2/basic_report.proto` line 103 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3942#-Ous29Ny18TSIcTUosSC:-OuxvJ9w78rVtF2GTgtu:b-atkpyk) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ca2882ed5c5273cddbbc5dd293858f633163fb89/src/main/proto/wfa/measurement/internal/reporting/v2/basic_report.proto#L103)):*

I'd prefer to just simplify and read the list at post processing time.
</blockquote></details>

There's two concerns with that approach:

1. More difficult to determine what the exemptions were after-the-fact. The operator would have to figure out what values were specified at at the time.
2. More work if we need to switch from specifying the exemptions as a global list for the whole server vs. per-report.

From my reading of the issue description I don't think either is a particularly important use case, as it sounds like there's only going to be one EDP in this list for the time being. Therefore I'd be supportive of @stevenwarejones's recommendation to have the flag defined at the post-processor rather than the API server.

ple13 (commented):
> @ple13 made 3 comments.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3942)* status: 17 of 29 files reviewed, 3 unresolved discussions (waiting on kungfucraig, SanjayVas, stevenwarejones, and stevenwarejones+AGNT).

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/gcloud/spanner/db/BasicReports.kt` line 339 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3942#-OuhTrUzAMt4IkV07ZIq:-OvBfbir5phNjXw3wOrP:b-d59h27) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ca2882ed5c5273cddbbc5dd293858f633163fb89/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/gcloud/spanner/db/BasicReports.kt#L339)):*
<details><summary><i>Previously, stevenwarejones+AGNT (Agent AGNT for Steven Ware Jones) wrote…</i></summary><blockquote>

I don't see this in the diff — conversion is happening in `BasicReportProtoConversions.toInternal:148-151` per-request, not at startup. `ReportingApiServerFlags.amiMrcExemptedEdps` is still a `List<String>` of resource names and `V2AlphaPublicApiServer.kt:423` passes it through unchanged. Sanjay raised the same point on R6 with a concrete picocli pattern.
</blockquote></details>

I'm sorry I missed the part about the parsing once before using. I've made the change regarding this. Now the flag is parsed once when the V2AlphaPublicApiServer starts. Then the parsed result will be used everywhere after that.
___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/BasicReportProtoConversions.kt` line 151 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3942#-OuxAqvc-zPcaRmFDqV_:-OvBgB5_06Tm5BpbPExA:b-bw8e14) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/dece6b6e279dad74da8ba25ba36de8c0d2957696/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/BasicReportProtoConversions.kt#L151)):*
<details><summary><i>Previously, stevenwarejones+AGNT (Agent AGNT for Steven Ware Jones) wrote…</i></summary><blockquote>

+1
</blockquote></details>

Updated.
___
*[`src/main/proto/wfa/measurement/internal/reporting/v2/basic_report.proto` line 103 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3942#-Ous29Ny18TSIcTUosSC:-OvBiBZ61LXg1JDVL6e-:b5olrr3) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ca2882ed5c5273cddbbc5dd293858f633163fb89/src/main/proto/wfa/measurement/internal/reporting/v2/basic_report.proto#L103)):*

There's two concerns with that approach:

1. More difficult to determine what the exemptions were after-the-fact. The operator would have to figure out what values were specified at at the time.
2. More work if we need to switch from specifying the exemptions as a global list for the whole server vs. per-report.

From my reading of the issue description I don't think either is a particularly important use case, as it sounds like there's only going to be one EDP in this list for the time being. Therefore I'd be supportive of @stevenwarejones's recommendation to have the flag defined at the post-processor rather than the API server.
</blockquote></details>

In API Phase 2, report processor (python server) polls for new reports from Reporting server periodically. If we have the flag defined on the post_processor side, I'm not sure how to pair a flag with each report. For now, there is only 1 edp that needs exemption, however, there are reports where no exemption is needed as well.

src/main/python/wfa/measurement/reporting/deploy/v2/common/job/post_process_report_result_job_executor.py
```diff
@@ -58,6 +58,13 @@
     "The path to the certificate collection file for validating the server's cert.",
     required=True)
 flags.DEFINE_alias("cert-collection-file", "cert_collection_file")
+_AMI_MRC_EXEMPTED_EDPS = flags.DEFINE_list(
+    "ami_mrc_exempted_edps",
```
stevenwarejones, line 62: What do you think of adding flag validation?

```python
flags.register_validator(
    "ami_mrc_exempted_edps",
    lambda values: all(
        v.startswith("dataProviders/") and v.count("/") == 1 and v.split("/")[1]
        for v in values
    ),
    message="--ami_mrc_exempted_edps entries must be of the form dataProviders/{id}",
)
```

stevenwarejones (commented):
> @stevenwarejones partially reviewed 22 files and all commit messages, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3942)* status: all files reviewed, 3 unresolved discussions (waiting on kungfucraig, ple13, SanjayVas, and stevenwarejones+AGNT).

src/main/python/wfa/measurement/reporting/deploy/v2/common/job/post_process_report_result_job_executor.py
ple13: Thanks for the suggestion! I've added the validation code.

ple13 (commented):
> @ple13 made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3942)* status: 29 of 31 files reviewed, 3 unresolved discussions (waiting on kungfucraig, SanjayVas, stevenwarejones, and stevenwarejones+AGNT).

stevenwarejones (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3942)* status: all files reviewed, 2 unresolved discussions (waiting on kungfucraig, SanjayVas, and stevenwarejones+AGNT).

SanjayVas (approved):
> @SanjayVas partially reviewed 27 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3942)* status: all files reviewed, 2 unresolved discussions (waiting on kungfucraig and stevenwarejones+AGNT).

___
*[`src/main/proto/wfa/measurement/internal/reporting/v2/basic_report.proto` line 103 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3942#-Ous29Ny18TSIcTUosSC:-OvH5Apt3rZ2EEVOPnxd:b6zxp1z) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ca2882ed5c5273cddbbc5dd293858f633163fb89/src/main/proto/wfa/measurement/internal/reporting/v2/basic_report.proto#L103)):*
> I'm not sure how to pair a flag with each report.

> there are reports where no exemption is needed as well.

From offline discussion: The PR was already using a flag to define the exempted EDPs across the whole instance rather than per-report. The suggestion was just to move this flag from the API server to the report processor so it didn't need to be passed via the internal API. This suggestion is now implemented.

ple13 (commented):
> @ple13 partially reviewed 31 files and all commit messages, resolved 2 discussions, and dismissed @stevenwarejones+AGNT from 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3942)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on kungfucraig and stevenwarejones).

ple13 (commented):
> @ple13 dismissed @stevenwarejones from a discussion.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3942)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on kungfucraig and stevenwarejones).

ple13 (commented):
Resolving comments.

> @ple13 made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3942#-:-OvHQ_kxBGmSpY2_nxYy:b-yaqsea)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on kungfucraig and stevenwarejones).

Issue #3913, opened 2026-06-04
Implement SubpoolAssigner Phase-0 TEE application


Marco-Premier:
Implement the SubpoolAssigner Phase-0 TEE application under `src/main/kotlin/org/wfanet/measurement/edpaggregator/subpoolassigner/`.

The dispatcher publishes one `WorkItem` per fingerprint shard for each (`RawImpressionUpload`, `ModelLine`) pair with memoization enabled. Each VM:
- creates / upserts the `RawImpressionUploadModelLine` row via the EDP Aggregator internal gRPC and loads the compiled VID model from `ModelLineConfig.model_blob_path`,
- reads and decrypts the upload's raw impressions inside the TEE,
- drops impressions outside `ModelLine.[active_start_time, active_end_time)`,
- keeps only impressions whose SHA-256 fingerprint matches its shard (`fp.hash() % N == shardIndex`),
- runs the labeler in pool-emit mode and writes one DEK-encrypted `SubpoolFingerprints` blob per (shard, subpool) to the SubpoolMap storage,
- flips its `PoolAssignmentJob` row to `POOL_ASSIGNMENT_SUCCEEDED`.

Last-shard-out merges the per-shard per-subpool blobs into one cumulative blob per subpool, bin-packs subpools into `RankerJob` rows, flips `RawImpressionUploadModelLineState` from `POOL_ASSIGNING` to `RANKING`, and triggers Phase-1 with one `WorkItem` per `RankerJob`.

Scaffold and `vid_labeler_params.proto` extensions (per-`ModelLineConfig` `model_blob_path` / `active_start_time` / `active_end_time`) landed by PR #<TBD>.

PR #3950, opened 2026-06-09
feat: Add RankIndexBlob proto definitions for internal and public API

getina:
## Summary
- Add proto message and service definitions for RankIndexBlob, a pointer to rank index blobs in Cloud Storage with encrypted DEKs.
- Includes create, batch create, get, list, and delete (soft-delete) methods for both internal and public (v1alpha) APIs.
- BlobType enum: DAY_ONLY (per-dispatch fingerprints), SNAPSHOT (cumulative set).
- Soft-delete via delete_time with show_deleted on List.
- cmms_model_line, subpool_id, and blob_type filters for cross-upload queries with AIP-159 wildcard support.
- max_event_date (google.type.Date) for retention policy queries.
- encrypted_dek (bytes) for blob encryption key material.

## Test plan
- bazel build passes for all proto targets (internal blob type, internal message, internal service, public message, public service)

Closes: #3885

Issue: #3885

PR #3941, opened 2026-06-08
feat: Add RankerJob proto definitions for internal and public API

getina:
## Summary
- Add proto message and service definitions for RankerJob, a Phase-1 ranker job resource for the VID labeling pipeline.
- Includes create, batch create, get, list, MarkSucceeded (custom response with is_last_job), and MarkFailed methods for both internal and public (v1alpha) APIs.
- State enum: CREATED, SUCCEEDED, FAILED (not terminal — operator can retry).
- cmms_model_line as REQUIRED+IMMUTABLE field with resource_reference to halo.wfanet.org/ModelLine.
- subpool_ids (repeated int32) for the subpools this ranker job processes.
- etag for optimistic concurrency control on Mark requests.
- AIP-159 wildcard support on List parent with cmms_model_line filter for cross-upload queries.

## Test plan
- bazel build passes for all proto targets (internal state, internal message, internal service, public message, public service)

Closes: #3896
Issue: #3896

PR #3910, opened 2026-06-03
feat: refactor VidLabelingDispatcher to static shards and VID Repo API + shared dispatch sequencer

jojijac0b:
## Summary

Refactors `VidLabelingDispatcher` to register VID-labeling uploads against the EDP Aggregator metadata store using static shard configuration and the VID Repository API, and extracts the dispatch logic into a shared `VidLabelingDispatchSequencer` that both the dispatcher's "fast path" and the `VidLabelingMonitor` (#4020) call.

**What the dispatcher does now:**
- On a "done" blob event, crawls the directory for raw-impression files and registers a `RawImpressionUpload` + a `RawImpressionUploadFile` per file.
- Resolves active PROD model lines via the VID Repository API (`ListModelLines`), filtered by active window + config, and registers a `RawImpressionUploadModelLine` per resolved model line.
- After registration, runs a best-effort fast-path dispatch via the shared sequencer instead of waiting for the next monitor tick. A fast-path failure never fails the (already successful) registration — the monitor remains the backstop.

**Shared `VidLabelingDispatchSequencer`:**
- The single component both the dispatcher (fast path) and the monitor (#4020 backstop) call, so the per-`(DataProvider, ModelLine)` serialization rule, model-shard resolution, and work-creation steps are defined exactly once.
- Serializes per `(DataProvider, ModelLine)` — different model lines run in parallel, but a model line already in flight is not started again (protects the cumulative rank index). Each model line is claimed via an etag compare-and-swap, so the fast path and the monitor are safe to run concurrently.
- **Non-memoized only.** Memoized Phase-0 dispatch is deferred to #4062.

**Idempotency:** request IDs are derived through a shared `RequestIds` helper (AIP-155); `ALREADY_EXISTS` on creates is acked and skipped for Pub/Sub-redelivery safety. Adds a small `BlobUris` helper for building blob URIs.

**Config / protos:** `VidLabelingConfig` gains the static-shard count, VID-Repository connection fields, and `control_plane_connection`; `VidLabelerParams` uses `raw_impression_upload`.

**Tests:** registration (files + model lines), model-line resolution/filtering, override model lines, `ALREADY_EXISTS` ack on each create, per-`(DataProvider, ModelLine)` sequencing, error propagation, done-marker exclusion, and metrics emission.

## Test plan

[...]

PR #3954, opened 2026-06-10
feat: add parallel RawImpressionStream reader for VID labeling

Marco-Premier:
## Summary
  
  Adds `RawImpressionSource` — the parallel, per-shard raw-impression reader for the VID-labeler memoized-assignment pipeline — together with its supporting utilities. The reader downloads this VM's
  shard of one upload's raw-impression Parquet files, decrypts/decodes them, computes a per-event digest, shard-filters, and hands batches to a caller-supplied per-file sink. It is **agnostic** to what
  downstream does with the events (Phase 0 pool assignment, Phase 2 labeling, …).
  
  ## What's added

  - **`RawImpressionSource`** — the parallel reader + its `BlobSink` contract.
  - **`EventIdDigest` + `EventIdDigestExtractor`** — a 12-byte SHA-256 truncation of the event id (`high: Long`, `low: Int`), used as the shard key and as the downstream map key. Deliberately distinct
  from the C++ labeler's `acting_fingerprint` (FarmHash64); pinned with golden vectors.
  - **`DigestedEvent<R>`** — a shard-surviving row plus its digest, generic over the row type (`ParquetDigestedEvent = DigestedEvent<Map<String, ParquetValue>>` for the Parquet reader).
  - **`ActiveWindow`** utility (+ tests) — a half-open event-time window used **downstream** for per-model-line filtering.

  ## Reader design

  - **Discovery:** gRPC `RawImpressionUploadFileService.ListRawImpressionUploadFiles` (via the common `listResources()` helper); each file's `blob_uri` is read through the injected `ParquetStorageClient`
  (Parquet decode + Parquet Modular Encryption decrypt).
  - **Concurrency:** one coroutine per file (bounded by `maxOpenFiles`); each file's batches are launched onto a **shared CPU dispatcher capped at `workers`** (`limitedParallelism`), so all cores stay
  saturated and even a single large file is parallelized. The per-file `coroutineScope` join provides per-file completion.
  - **Backpressure:** a global `inFlight` semaphore (`maxInFlightBatches`) bounds launched-but-unfinished batches, throttling readers to downstream consumption speed.
  - **Per-file sink lifecycle:** `openSink(blobUri)` mints a `BlobSink` per file. `processBatch` is invoked **concurrently** for the same blob (implementations must be thread-safe). On success `commit()`
  finalizes/publishes; `close()` **always** runs (under `NonCancellable`) to release resources, so a cancelled blob never leaks.
  - **Filtering done here:** only the shard filter, `floorMod(digest.high, totalShards) == shardIndex`. Time-window filtering, labeling, and output writing are the sink's responsibility.
  - **Delivery / failure semantics:** delivers every shard-surviving row (including duplicate digests — dedup is the sink's job). The guarantee is **per-blob, not per-upload**: a failed blob is never
  committed (no partial output), but already-committed blobs are published — so retry / output naming must be idempotent per blob.

  ## Observability
  
  - **Metrics** (OpenTelemetry via `EdpaTelemetry`): `edpa.raw_impression_source.rows_read` / `rows_dropped_other_shard` / `rows_emitted` counters and a `file_processing_duration` histogram.
  - **Logging:** start line (file count + `shardIndex/totalShards`), per-file `FINE` start/complete, an `INFO` progress line every ~10% of files, and a completion summary.

  ## Tuning knobs (constructor)

  `maxOpenFiles`, `workers`, `batchSize`, `maxInFlightBatches`, plus the read/worker dispatchers.

  ## Testing
  
  - `RawImpressionSourceTest` — in-process gRPC fake for discovery + local-filesystem Parquet; covers shard filtering, concurrent multi-file processing, per-file open/commit/close, no-commit-but-close on
  failure, and metrics.
  - `EventIdDigestExtractorTest` — golden vectors.
  - `ActiveWindowTest` — bounds validation + half-open `contains` + open-ended window.

Issue: #3913


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3954)

src/main/proto/wfa/measurement/edpaggregator/v1alpha/rank_index_map.proto
```diff
@@ -0,0 +1,94 @@
...
+// `org.wfanet.measurement.edpaggregator.vidlabeler.utils.Bytes12IntMap`.
+message RankIndexMap {
+  // Identifier of the subpool this map is scoped to.
+  int64 subpool_id = 1;
+
+  // The ranked sub-range size for this subpool as configured in the model.
+  // Each value in `ranks` MUST satisfy `0 <= rank < ranked_size`.
+  // `int32` matches the in-memory Int rank representation and the
+  // `fixed32` element type of the `ranks` field; subpool sizes are
+  // bounded well below `Int.MAX_VALUE` (~1.5 B in the worst case).
```
stevenwarejones, line 75: subpool_id is int64 here and in SubpoolFingerprints, but int32 on RankIndexBlob (PR #3950), RankerJob.subpool_ids (PR #3941), and PoolAssignmentJob.shard_index. The Spanner schema has INT64 but all realistic subpool counts fit int32. Pick one and be consistent — mixing int32 and int64 for the same concept creates unnecessary casting at every boundary.

src/main/kotlin/org/wfanet/measurement/edpaggregator/rawimpressions/EventIdDigestExtractor.kt
```diff
@@ -0,0 +1,101 @@
...
+        ((b[off + 6].toLong() and 0xFF) shl 8) or
+        (b[off + 7].toLong() and 0xFF)
+
+    private fun readIntBE(b: ByteArray, off: Int): Int =
+      ((b[off].toInt() and 0xFF) shl 24) or
+        ((b[off + 1].toInt() and 0xFF) shl 16) or
+        ((b[off + 2].toInt() and 0xFF) shl 8) or
+        (b[off + 3].toInt() and 0xFF)
+  }
+}
```
stevenwarejones, line 97: Add a unit test for FingerprintExtractor with pinned golden vectors — a known input producing an expected Fingerprint(high, low). The C++ Phase 2 VidLabeler also computes fingerprints from event IDs and they must match for shard assignment to be consistent across languages. Without a golden vector test, a divergence in SHA-256 truncation or byte ordering would only surface as a mysterious shard mismatch in production.

stevenwarejones (commented):
Two issues:

src/main/proto/wfa/measurement/edpaggregator/v1alpha/rank_index_map.proto
```diff
@@ -0,0 +1,94 @@
...
+  //
+  // The in-memory map representation
+  // (`org.wfanet.measurement.edpaggregator.vidlabeler.utils.Bytes12IntMap`)
+  // reserves `-1` (`0xFFFFFFFF` when reinterpreted as unsigned) as
+  // its [NOT_PRESENT] sentinel; that value MUST never appear in this
+  // field. Holds by construction because valid ranks are in
+  // `[0, ranked_size)` and `ranked_size` is bounded well below
+  // `Int.MAX_VALUE`.
+  repeated fixed32 ranks = 4 [packed = true];
+}
```
stevenwarejones, line 97: The comment about -1 / 0xFFFFFFFF as NOT_PRESENT sentinel is correct but could be clearer. fixed32 is unsigned in proto — it does not have negative values. The sentinel is a Kotlin-side concern where the fixed32 value is reinterpreted as a signed Int, and -1 (0xFFFFFFFF) is reserved by Bytes12IntMap. Consider rephrasing to: "Bytes12IntMap reserves 0xFFFFFFFF (Int -1 in Kotlin) as its NOT_PRESENT sentinel; that value MUST never appear in this field."

src/main/kotlin/org/wfanet/measurement/edpaggregator/rawimpressions/RawImpressionStream.kt
```diff
@@ -0,0 +1,357 @@
...
+    val emitted: AtomicLong = AtomicLong(),
+  )
+
+  val stats: Stats = Stats()
+
+  private val eventTimeColumn: String =
+    labelerInputFieldMapping[EVENT_TIME_FIELD_PATH]
+      ?: throw IllegalArgumentException(
+        "labelerInputFieldMapping is missing required entry for " +
+          "'$EVENT_TIME_FIELD_PATH'; got keys ${labelerInputFieldMapping.keys}"
```
stevenwarejones: FingerprintExtractor uses ThreadLocal for per-thread MessageDigest reuse. This works correctly with Dispatchers.IO (each IO thread gets its own State, digest is reset at the start of each extract call). Just confirming this is intentional and that extract() is never called across a suspension point within a single coroutine — looking at processFile, the fingerprint + onEvent are sequential with no suspension between them, so this is safe.

src/main/proto/wfa/measurement/edpaggregator/v1alpha/subpool_fingerprints.proto
```diff
@@ -0,0 +1,63 @@
...
+  // each contiguous 12-byte run is one fingerprint (the leading 12
+  // bytes of a uniformly-distributed hash such as SHA-256 of the
+  // upstream account identifier).
+  //
+  // Entry count is derivable as `fingerprints.size() / 12`; no
+  // separate count field is stored because it would introduce a
+  // consistency hazard (count vs. actual length drift) without any
+  // size benefit at scale.
+  bytes fingerprints = 2;
+}
```
stevenwarejones, line 66: subpool_id is int64 here but int32 on RankIndexBlob (PR #3950) and RankerJob.subpool_ids (PR #3941). Same consistency issue as noted on rank_index_map.proto.

stevenwarejones (commented):
Three more issues:

src/main/kotlin/org/wfanet/measurement/edpaggregator/rawimpressions/RawImpressionStream.kt
```diff
@@ -0,0 +1,357 @@
...
+   *   a fresh consumer (its per-file state is thread-confined). State shared
+   *   across files (captured in the factory) must be concurrency-safe.
+   */
+  suspend fun stream(consumerFactories: List<ModelLineConsumerFactory>) {
+    require(consumerFactories.isNotEmpty()) { "consumerFactories must be non-empty" }
+    val semaphore = Semaphore(maxConcurrentReaders)
+    coroutineScope {
+      for (blobKey in discoverBlobKeys()) {
+        launch(readDispatcher) {
+          semaphore.withPermit {
```
stevenwarejones: No operational visibility during execution. StorageEventSource in the ResultsFulfiller has a ProgressTracker that logs at every 10% completion with total events and batches, plus per-file FINE-level logging ("Reading events from ...", "Read N events in M batches"). This component processes thousands of files over minutes — operators need to see progress. Add:
1. Log at stream start with total file count and shard info
2. Per-file FINE-level log on start and completion (with row count)
3. Periodic progress log (e.g. every 10% of files completed)
4. Log at stream completion with stats summary (total read, dropped, emitted)

src/main/kotlin/org/wfanet/measurement/edpaggregator/rawimpressions/FingerprintedEvent.kt
```diff
@@ -0,0 +1,37 @@
...
+ * OPTIONAL columns absent in this row). Includes the event-time
+ * column (epoch micros, as a `Long`) — callers who need an `Instant`
+ * pull it from [row] and convert.
+ *
+ * Downstream consumers project [row] into whatever shape they need
+ * (typically a `LabelerInput` proto) using the
+ * `labeler_input_field_mapping` to translate semantic LabelerInput
+ * field paths to the parquet column names that key this map.
+ */
+data class FingerprintedEvent(val row: Map<String, Any?>, val fingerprint: Fingerprint)
```
stevenwarejones: row: Map<String, Any?> is type-unsafe — every consumer casts values out of the map. common-jvm PR world-federation-of-advertisers/common-jvm#385 adds parquet_row.proto with ParquetRow (map<string, ParquetValue>) and ParquetValue (oneof kind with all Parquet types). Use ParquetRow or Map<String, ParquetValue> instead — readEventIdBytes and readEventMicros would pattern-match on ParquetValue.kind instead of is String / is Long / is Instant casts.

src/main/kotlin/org/wfanet/measurement/edpaggregator/rawimpressions/RawImpressionStream.kt
```diff
@@ -0,0 +1,357 @@
...
+      }
+      stats.emitted.incrementAndGet()
+      onEvent(FingerprintedEvent(row, fingerprint))
+    }
+  }
+
+  /**
+   * Lists the folder containing [doneBlobKey] and drops the done marker,
+   * returning the blob keys of the remaining raw-impression files.
+   */
```
stevenwarejones: discoverBlobKeys() lists the GCS folder to find raw impression files. The design doc flow has the dispatcher creating RawImpressionUploadFile entries in the database — the database should be the source of truth, not GCS listing (which can pick up stale or partially uploaded files). Consider extracting blob discovery into an interface so the production path queries RawImpressionUploadFile rows and the GCS listing approach can be a test fallback.

src/main/kotlin/org/wfanet/measurement/edpaggregator/rawimpressions/RawImpressionStream.kt
```diff
@@ -0,0 +1,357 @@
...
+   * to every consumer whose window contains it, then close every consumer.
+   * Decode + decrypt + fingerprint are paid once per row; suspends until every
+   * file has been fully processed.
+   *
+   * ## How many consumers (Phase 0 vs Phase 2)
+   *
+   *  - **Phase 0 (pool assignment):** at most ONE model line per WorkItem, so
+   *    `consumerFactories` has size 1. The consumer labels in pool-emit mode and
+   *    routes `(subpoolId, fingerprint)` into shared per-subpool accumulators;
+   *    `close()` is a no-op and the accumulators are uploaded once after this
```
stevenwarejones: Stats uses raw AtomicLong counters — only accessible in-process. VidLabelingDispatcherMetrics uses OTel LongCounters that integrate with EdpaTelemetry and surface in operator dashboards. This component runs in the same TEE apps — use OTel counters so operators can monitor read/drop/emit rates. Also add a histogram for per-file processing time.

src/main/kotlin/org/wfanet/measurement/edpaggregator/rawimpressions/RawImpressionStream.kt
stevenwarejones: Document in the stream() KDoc that it delivers every impression row that passes the shard filter, including duplicate fingerprints (the same person can appear in multiple impressions across files). Callers that need unique fingerprints (e.g. Phase 0's SubpoolFingerprints output) are responsible for deduplicating in their consumer.

src/main/proto/wfa/measurement/edpaggregator/v1alpha/subpool_fingerprints.proto
```diff
@@ -0,0 +1,63 @@
...
+// Read pattern: obtain the `fingerprints` byte buffer and slice it in
+// 12-byte chunks (no per-element allocation needed).
+message SubpoolFingerprints {
+  // Identifier of the subpool this blob is scoped to.
+  int64 subpool_id = 1;
+
+  // Fingerprints packed back-to-back. Length MUST be a multiple of 12;
+  // each contiguous 12-byte run is one fingerprint (the leading 12
+  // bytes of a uniformly-distributed hash such as SHA-256 of the
+  // upstream account identifier).
```
stevenwarejones, line 59: Document that subpool_id is the pool's population_offset from the RankedPopulationNode model config — the same value emitted as PoolAssignment.pool_offset in pass-1 mode and matched as RankAssignment.pool_offset in pass-2. This is the identity key that ties Phase 0 output → Phase 1 rank table → Phase 2 labeler lookup. Same documentation needed on RankIndexMap.subpool_id.

src/main/proto/wfa/measurement/edpaggregator/v1alpha/rank_index_map.proto
```diff
@@ -0,0 +1,94 @@
...
+// proto merge concatenates `repeated` fields but replaces singular
+// `bytes`, which would break the pairing).
+//
+// In-memory representation: see
+// `org.wfanet.measurement.edpaggregator.vidlabeler.utils.Bytes12IntMap`.
+message RankIndexMap {
+  // Identifier of the subpool this map is scoped to.
+  int64 subpool_id = 1;
+
+  // The ranked sub-range size for this subpool as configured in the model.
```
stevenwarejones, line 71: Same as SubpoolFingerprints: document that subpool_id is the pool's population_offset from the RankedPopulationNode model config, matching PoolAssignment.pool_offset (pass-1) and RankAssignment.pool_offset (pass-2).

src/main/kotlin/org/wfanet/measurement/edpaggregator/rawimpressions/Fingerprint.kt
```diff
@@ -0,0 +1,35 @@
...
+ *
+ * Use [high] directly for shard partitioning (`fp.high % totalShards`) —
+ * SHA-256's avalanche makes the upper Long a uniform shard hash on its own.
+ * The [low] Int exists only to extend bucket-key resolution past 64 bits
+ * inside `Bytes12IntMap`; it is not part of the shard math.
+ *
+ * @property high big-endian bytes 0..7 of the SHA-256 output.
+ * @property low big-endian bytes 8..11 of the SHA-256 output.
+ */
+data class Fingerprint(val high: Long, val low: Int)
```
stevenwarejones: The VID labeling framework already has a separate concept called fingerprint — acting_fingerprint in LabelerEvent (a FarmHash used for model tree routing and unranked VID assignment). This SHA-256 digest serves a different purpose (shard partitioning and rank table key). Having both called "fingerprint" is confusing — a developer could easily assume they are the same thing. Rename to EventIdDigest (and FingerprintExtractor → EventIdDigestExtractor, FingerprintedEvent → DigestedEvent or similar) to distinguish from the labeler's acting_fingerprint.

src/main/proto/wfa/measurement/edpaggregator/v1alpha/rank_index_map.proto
```diff
@@ -0,0 +1,94 @@
...
+package wfa.measurement.edpaggregator;
+
+option java_package = "org.wfanet.measurement.edpaggregator.v1alpha";
+option java_multiple_files = true;
+option java_outer_classname = "RankIndexMapProto";
+
+// Serialized on-disk form of a memoized VID rank-index map for a single
+// subpool.
+//
+// The map records `(fingerprint, rank)` assignments produced by the
```
stevenwarejones, line 30: The serialization/deserialization code between this proto and Bytes12IntMap (and between Fingerprint and the packed 12-byte slots) doesn't exist yet. When implemented, the byte order must exactly match FingerprintExtractor.readLongBE/readIntBE — a mismatch will silently break rank lookups. Co-locate the serialize/deserialize methods with Bytes12IntMap and add a round-trip golden vector test that verifies a Fingerprint survives pack → unpack with the same (high, low) values.

src/main/kotlin/org/wfanet/measurement/edpaggregator/vidlabeler/utils/ActiveWindow.kt
```diff
@@ -0,0 +1,75 @@
...
+/** Converts an epoch-microseconds `long` to an [Instant]. */
+fun epochMicrosToInstant(epochMicros: Long): Instant {
+  val seconds = Math.floorDiv(epochMicros, 1_000_000L)
+  val nanos = Math.floorMod(epochMicros, 1_000_000L) * 1_000L
+  return Instant.ofEpochSecond(seconds, nanos)
+}
+
+/** Converts an [Instant] to epoch microseconds, truncating sub-microsecond nanos. */
+fun instantToEpochMicros(instant: Instant): Long =
+  Math.addExact(Math.multiplyExact(instant.epochSecond, 1_000_000L), (instant.nano / 1_000).toLong())
```
stevenwarejones: Add an ActiveWindowTest with at least:
1. Throws when endMicros <= startMicros
2. Open-ended window (end = null via of()) contains any event after start
3. Boundary: event exactly at startMicros is included
4. Boundary: event exactly at endMicros is excluded
5. Event before startMicros is excluded

src/main/kotlin/org/wfanet/measurement/edpaggregator/rawimpressions/RawImpressionStream.kt
```diff
@@ -0,0 +1,357 @@
...
+   * inline for every surviving event. Runs entirely on the calling file
+   * coroutine.
+   */
+  private suspend fun processFile(
+    blobKey: String,
+    onEvent: suspend (FingerprintedEvent) -> Unit,
+  ) {
+    val parquetBlob =
+      parquetStorageClient.getBlob(blobKey) ?: error("Raw-impression blob not found: $blobKey")
+    parquetBlob.readRows().collect { row ->
```
stevenwarejones: processFile runs decode and consumer work inline on the same coroutine — while the labeler processes row N, the parquet reader is idle, and vice versa. For large files (EDPs batch by campaign-day), this leaves CPU and I/O alternating idle within each file. StorageEventSource (src/main/kotlin/.../resultsfulfiller/StorageEventSource.kt) uses channelFlow to pipeline reading ahead of consumption. Add a bounded Channel within processFile so the reader can stay ahead by a configurable number of rows, hiding GCS/decode latency behind labeler latency within each file.

src/main/kotlin/org/wfanet/measurement/edpaggregator/rawimpressions/RawImpressionStream.kt
```diff
@@ -0,0 +1,357 @@
...
+
+    /**
+     * Handles one in-window event for the current file. Typically projects the
+     * row into a `LabelerInput`, runs the labeler for this model line, and
+     * appends/routes the result.
+     */
+    suspend fun process(event: FingerprintedEvent)
+
+    /**
+     * The current file has been fully read. Flush + upload this model line's
```
stevenwarejones: RawImpressionStream uses one semaphore for both I/O and CPU concurrency. GCS reads are latency-bound — you want many in-flight requests to saturate the network. CPU work (decode, fingerprint, label) is core-bound. StorageEventSource in the ResultsFulfiller gets this right by launching all readers concurrently while channelFlow backpressure bounds how far ahead they get. Use a bounded Channel within processFile so the reader stays ahead on GCS while channel capacity bounds memory. This naturally separates I/O concurrency (high) from CPU concurrency (bounded by channel + consumer speed).

src/main/kotlin/org/wfanet/measurement/edpaggregator/rawimpressions/RawImpressionStream.kt
```diff
@@ -0,0 +1,357 @@
...
+/**
+ * Streams the local VM's shard of raw impressions for one upload and runs the
+ * full read → decrypt → fingerprint → shard-filter → caller-supplied work chain
+ * **inline, one coroutine per file, in parallel across files** — labeling each
+ * surviving event against one or more model lines in a single pass.
+ *
+ * ## Execution model (why it is shaped this way)
+ *
+ * The read path is CPU-heavy (parquet decode + PME AES-GCM decrypt +
+ * decompression) on top of blocking GCS I/O, and there are always far more
```
stevenwarejones: Nit: consider renaming to RawImpressionSource to parallel StorageEventSource in the ResultsFulfiller. Both are push-based sources that drive consumers, not pull-based streams. Consistent naming convention across the codebase.

stevenwarejones (commented):
> @stevenwarejones reviewed 12 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3954)* status: 12 of 21 files reviewed, 18 unresolved discussions (waiting on Marco-Premier).

src/main/kotlin/org/wfanet/measurement/edpaggregator/rawimpressions/DigestedEvent.kt
```diff
@@ -0,0 +1,40 @@
+/*
```
stevenwarejones, line 1: Nit (non-blocking): consider making this generic — data class DigestedEvent<R>(val row: R, val digest: EventIdDigest). Different readers could exist in the future and the type parameter keeps DigestedEvent reusable without coupling it to ParquetValue.

src/main/kotlin/org/wfanet/measurement/edpaggregator/rawimpressions/RawImpressionSource.kt
```diff
@@ -0,0 +1,480 @@
...
+          openFiles.withPermit { processBlob(blobUri, openSink, cpuDispatcher, inFlight, progress) }
+        }
+      }
+    }
+    progress.logSummary()
+  }
+
+  /**
+   * Reads one input file on its own coroutine: mints its [BlobSink], launches each
+   * batch onto the shared `cpuDispatcher` inside a per-file [coroutineScope] (which
```
stevenwarejones: discoverBlobUris still uses manual do..while(nextPageToken.isNotEmpty()) pagination. Use the listResources() helper from ListResources.kt (org.wfanet.measurement.common.api.grpc) — same feedback as PR #3910.

src/main/kotlin/org/wfanet/measurement/edpaggregator/rawimpressions/EventIdDigestExtractor.kt
stevenwarejones: Verified — renamed to EventIdDigestExtractor. Cross-language parity does not apply here. Indirect test coverage via shard filtering is sufficient.

src/main/kotlin/org/wfanet/measurement/edpaggregator/rawimpressions/RawImpressionStream.kt
stevenwarejones: Verified — renamed to RawImpressionSource.

src/main/kotlin/org/wfanet/measurement/edpaggregator/rawimpressions/RawImpressionStream.kt
stevenwarejones: Verified — Design B separates I/O (readDispatcher = Dispatchers.IO, elastic) from CPU (cpuDispatcher = limitedParallelism(workers)). The inFlight semaphore bounds read-ahead.

src/main/kotlin/org/wfanet/measurement/edpaggregator/rawimpressions/RawImpressionStream.kt
stevenwarejones: Verified — Design B introduces suspension points between extract() calls (inFlight.acquire in onBatch), but each extract() is self-contained (reset → update → digest → return) with no suspension within it. The coroutine may resume on a different IO thread after the suspension, but the next extract() gets that thread's ThreadLocal State, which is fine.

src/main/kotlin/org/wfanet/measurement/edpaggregator/rawimpressions/RawImpressionStream.kt
stevenwarejones: Verified — Design B has RawImpressionSourceMetrics with OTel LongCounters (rows read/dropped/emitted) and a DoubleHistogram for per-file processing duration.

src/main/kotlin/org/wfanet/measurement/edpaggregator/rawimpressions/RawImpressionStream.kt
stevenwarejones: Verified — Design B has ProgressTracker (10% progress + summary), per-file FINE logging with row counts, and OTel counters via RawImpressionSourceMetrics (rows read/dropped/emitted + per-file duration histogram).

stevenwarejones: Verified — streamBlobs KDoc documents that duplicates are delivered and callers are responsible for dedup in their BlobSink.

src/main/kotlin/org/wfanet/measurement/edpaggregator/rawimpressions/RawImpressionStream.kt
stevenwarejones: Verified — Design B pipelines via batch + launch(cpuDispatcher) with inFlight semaphore for backpressure. Gives intra-file parallelism and per-file completion via coroutineScope join.

src/main/kotlin/org/wfanet/measurement/edpaggregator/rawimpressions/RawImpressionStream.kt
stevenwarejones: Verified — Design B discovers files via ListRawImpressionUploadFiles API (paginated), not GCS listing. The database is the source of truth.

src/main/kotlin/org/wfanet/measurement/edpaggregator/rawimpressions/Fingerprint.kt
stevenwarejones: Verified — renamed to EventIdDigest, EventIdDigestExtractor, DigestedEvent.

src/main/kotlin/org/wfanet/measurement/edpaggregator/rawimpressions/FingerprintedEvent.kt
stevenwarejones: Verified — Design B uses Map<String, ParquetValue> from common-jvm. readEventIdBytes pattern-matches on ParquetValue.KindCase.

stevenwarejones (commented):
> @stevenwarejones+AGNT made 11 comments and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3954)* status: 7 of 26 files reviewed, 19 unresolved discussions (waiting on Marco-Premier and stevenwarejones).

src/main/kotlin/org/wfanet/measurement/edpaggregator/rawimpressions/RawImpressionSource.kt
```diff
@@ -0,0 +1,480 @@
...
+  private val workerDispatcher: CoroutineDispatcher = Dispatchers.Default,
+  // Max input files open at once: the memory/I/O governor for reading. Defaults to
+  // ~2× cores — enough open readers to keep the CPU pool fed while some block on GCS.
+  private val maxOpenFiles: Int = DEFAULT_MAX_OPEN_FILES,
+  // CPU parallelism for batch processing (the bottleneck). Defaults to ~#cores.
+  private val workers: Int = DEFAULT_WORKERS,
+  // Number of shard-surviving events per batch (amortizes per-batch handoff).
+  private val batchSize: Int = DEFAULT_BATCH_SIZE,
+  // Read-ahead / memory bound: max batches launched-but-unfinished at once. Keep
+  // it > [workers] so finishing workers always have a decoded batch queued.
```
stevenwarejones, line 168: Nit (non-blocking): workers, maxOpenFiles, batchSize, and maxInFlightBatches are constructor params with sensible defaults, but the runners (SubpoolAssignerAppRunner, VidRankBuilderAppRunner) don't expose them as CLI flags. ResultsFulfillerAppRunner exposes --pipeline-batch-size and --pipeline-workers for operator tuning. Consider adding equivalent flags so operators can tune concurrency per deployment without code changes.

stevenwarejones (commented):
> @stevenwarejones partially reviewed 11 files and all commit messages, and resolved 3 discussions.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3954)* status: 18 of 26 files reviewed, 17 unresolved discussions (waiting on Marco-Premier and stevenwarejones+AGNT).

src/main/kotlin/org/wfanet/measurement/edpaggregator/vidlabeler/utils/ActiveWindow.kt
stevenwarejones: Verified — ActiveWindowTest covers all five scenarios: rejects equal/inverted bounds, half-open contains (startMicros included, endMicros excluded, before excluded), open-ended window via of(start, null), plus instant-micros round-trip conversion.

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3954)* status: 18 of 26 files reviewed, 17 unresolved discussions (waiting on Marco-Premier and stevenwarejones).

stevenwarejones (commented):
> @stevenwarejones reviewed 7 files and resolved 7 discussions.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3954)* status: 21 of 26 files reviewed, 10 unresolved discussions (waiting on Marco-Premier and stevenwarejones+AGNT).

src/main/proto/wfa/measurement/edpaggregator/v1alpha/rank_index_map.proto
stevenwarejones: Verified — serialization code belongs in the ranker implementation PR, not here. The byte-order constraint is documented in the proto comments. The round-trip test should be added when the serialize/deserialize methods are written.

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3954)* status: 21 of 26 files reviewed, 10 unresolved discussions (waiting on Marco-Premier and stevenwarejones).

src/main/proto/wfa/measurement/edpaggregator/v1alpha/rank_index_map.proto
stevenwarejones: No longer actionable on this PR — these protos already merged to main. Filing as a follow-up if still needed.

src/main/proto/wfa/measurement/edpaggregator/v1alpha/rank_index_map.proto
stevenwarejones: No longer actionable on this PR — these protos already merged to main. Filing as a follow-up if still needed.

src/main/proto/wfa/measurement/edpaggregator/v1alpha/rank_index_map.proto
stevenwarejones: No longer actionable on this PR — these protos already merged to main. Filing as a follow-up if still needed.

src/main/proto/wfa/measurement/edpaggregator/v1alpha/subpool_fingerprints.proto
stevenwarejones: No longer actionable on this PR — these protos already merged to main. Filing as a follow-up if still needed.

src/main/proto/wfa/measurement/edpaggregator/v1alpha/subpool_fingerprints.proto
stevenwarejones: No longer actionable on this PR — these protos already merged to main. Filing as a follow-up if still needed.

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3954)* status: 21 of 26 files reviewed, 10 unresolved discussions (waiting on Marco-Premier and stevenwarejones).

stevenwarejones:
Per #3968: rename subpool_id/subpool_ids to pool_offset/pool_offsets and change int32 to int64 where applicable. The value is population_offset (uint64) from the VID model, not a small sequential index. See PR #3969 for the pattern.

stevenwarejones (changes requested):
> @stevenwarejones reviewed 5 files and resolved 7 discussions.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3954)* status: 25 of 26 files reviewed, 3 unresolved discussions (waiting on Marco-Premier).

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3954)* status: all files reviewed, 3 unresolved discussions (waiting on Marco-Premier).

src/main/kotlin/org/wfanet/measurement/edpaggregator/rawimpressions/RawImpressionSource.kt
Marco-Premier: Agreed in principle — RawImpressionSource already takes workers/maxOpenFiles/batchSize/maxInFlightBatches as constructor params, but the runners don't construct it yet (they're scaffolds). I'll add the --pipeline-\*-style flags (matching ResultsFulfillerAppRunner) in the PR that actually wires RawImpressionSource into the Subpool-Assigner / Vid-Rank-Builder apps, where there's something to bind them to.

src/main/kotlin/org/wfanet/measurement/edpaggregator/rawimpressions/RawImpressionSource.kt
Marco-Premier: Done.

src/main/kotlin/org/wfanet/measurement/edpaggregator/vidlabeler/utils/ActiveWindow.kt
Marco-Premier: Done.

src/main/kotlin/org/wfanet/measurement/edpaggregator/rawimpressions/RawImpressionStream.kt
Marco-Premier: Confirmed — intentional and safe. One correction to the reasoning: the safety doesn't actually depend on there being no suspension between calls. extract() is a plain synchronous function that does ThreadLocal.get() and digest.reset() at the start of every call, so the MessageDigest is never held across a suspension — each call is self-contained on whatever IO thread it runs on.

Worth noting after the refactor: the per-event path is now readEventsFromBlob (not processFile), and onEvent does suspend at batch boundaries (workerChannel.send), possibly resuming on a different Dispatchers.IO thread. That's still fine, for exactly the reset-per-call reason above. So no change needed.

src/main/kotlin/org/wfanet/measurement/edpaggregator/rawimpressions/RawImpressionStream.kt
Marco-Premier: Done.

src/main/kotlin/org/wfanet/measurement/edpaggregator/rawimpressions/RawImpressionStream.kt
Marco-Premier: Done.

Marco-Premier: Done.

src/main/kotlin/org/wfanet/measurement/edpaggregator/rawimpressions/RawImpressionStream.kt
Marco-Premier: After the refactor, this is outdated comment now

src/main/kotlin/org/wfanet/measurement/edpaggregator/rawimpressions/RawImpressionStream.kt
Marco-Premier: I added grpc integration to query the files from DB

src/main/kotlin/org/wfanet/measurement/edpaggregator/rawimpressions/DigestedEvent.kt
Marco-Premier: Done.

src/main/kotlin/org/wfanet/measurement/edpaggregator/rawimpressions/Fingerprint.kt
Marco-Premier: Done.

src/main/kotlin/org/wfanet/measurement/edpaggregator/rawimpressions/EventIdDigestExtractor.kt
Marco-Premier: Test added.

EventIdDigest is a SHA-256 12-byte value used purely as the Kotlin-side key for our subpool/rank-index maps and shard split — it is intentionally **not** the C++ labeler's acting_fingerprint (a 64-bit FarmHash used for model-tree routing), the two are never compared, and we chose 96 bits for birthday-collision safety at ~1.5 B fingerprints that a 64-bit hash can't provide; the new golden-vector test pins our values so the persisted key stays stable across Phase 1/Phase 2 and over time.

src/main/kotlin/org/wfanet/measurement/edpaggregator/rawimpressions/FingerprintedEvent.kt
Marco-Premier: Done. To do this I had changed the signature of `override fun readRows(): Flow<Map<String, ParquetValue>>` in the `ParquetStorageClient`

Marco-Premier (commented):
> @Marco-Premier made 13 comments.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3954)* status: 22 of 26 files reviewed, 3 unresolved discussions (waiting on stevenwarejones).

src/main/kotlin/org/wfanet/measurement/edpaggregator/rawimpressions/RawImpressionSource.kt
Marco-Premier: Agreed in principle — RawImpressionSource already takes workers/maxOpenFiles/batchSize/maxInFlightBatches as constructor params, but the runners don't construct it yet (they're scaffolds). I'll add the --pipeline-\*-style flags (matching ResultsFulfillerAppRunner) in the PR that actually wires RawImpressionSource into the Subpool-Assigner / Vid-Rank-Builder apps, where there's something to bind them to.

src/main/kotlin/org/wfanet/measurement/edpaggregator/rawimpressions/RawImpressionSource.kt
Marco-Premier: Done.

src/main/kotlin/org/wfanet/measurement/edpaggregator/vidlabeler/utils/ActiveWindow.kt
Marco-Premier: Done.

src/main/kotlin/org/wfanet/measurement/edpaggregator/rawimpressions/RawImpressionStream.kt
Marco-Premier: Confirmed — intentional and safe. One correction to the reasoning: the safety doesn't actually depend on there being no suspension between calls. extract() is a plain synchronous function that does ThreadLocal.get() and digest.reset() at the start of every call, so the MessageDigest is never held across a suspension — each call is self-contained on whatever IO thread it runs on.

Worth noting after the refactor: the per-event path is now readEventsFromBlob (not processFile), and onEvent does suspend at batch boundaries (workerChannel.send), possibly resuming on a different Dispatchers.IO thread. That's still fine, for exactly the reset-per-call reason above. So no change needed.

src/main/kotlin/org/wfanet/measurement/edpaggregator/rawimpressions/RawImpressionStream.kt
Marco-Premier: Done.

src/main/kotlin/org/wfanet/measurement/edpaggregator/rawimpressions/RawImpressionStream.kt
Marco-Premier: Done.

Marco-Premier: Done.

src/main/kotlin/org/wfanet/measurement/edpaggregator/rawimpressions/RawImpressionStream.kt
Marco-Premier: After the refactor, this is outdated comment now

src/main/kotlin/org/wfanet/measurement/edpaggregator/rawimpressions/RawImpressionStream.kt
Marco-Premier: I added grpc integration to query the files from DB

src/main/kotlin/org/wfanet/measurement/edpaggregator/rawimpressions/DigestedEvent.kt
Marco-Premier: Done.

src/main/kotlin/org/wfanet/measurement/edpaggregator/rawimpressions/Fingerprint.kt
Marco-Premier: Done.

src/main/kotlin/org/wfanet/measurement/edpaggregator/rawimpressions/EventIdDigestExtractor.kt
Marco-Premier: Test added.

EventIdDigest is a SHA-256 12-byte value used purely as the Kotlin-side key for our subpool/rank-index maps and shard split — it is intentionally **not** the C++ labeler's acting_fingerprint (a 64-bit FarmHash used for model-tree routing), the two are never compared, and we chose 96 bits for birthday-collision safety at ~1.5 B fingerprints that a 64-bit hash can't provide; the new golden-vector test pins our values so the persisted key stays stable across Phase 1/Phase 2 and over time.

src/main/kotlin/org/wfanet/measurement/edpaggregator/rawimpressions/FingerprintedEvent.kt
Marco-Premier: Done. To do this I had changed the signature of `override fun readRows(): Flow<Map<String, ParquetValue>>` in the `ParquetStorageClient`

Marco-Premier (commented):
> @Marco-Premier made 13 comments.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3954)* status: 22 of 26 files reviewed, 3 unresolved discussions (waiting on stevenwarejones).

src/main/kotlin/org/wfanet/measurement/edpaggregator/rawimpressions/RawImpressionSource.kt
stevenwarejones: Verified — discoverBlobUris now uses listResources() from ListResources.kt with ResourceList return.

src/main/kotlin/org/wfanet/measurement/edpaggregator/rawimpressions/DigestedEvent.kt
stevenwarejones: Verified — DigestedEvent<R> is generic, with ParquetDigestedEvent typealias for the parquet case. BlobSink.processBatch uses the typealias.

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3954)* status: 22 of 26 files reviewed, 3 unresolved discussions (waiting on stevenwarejones).

src/main/kotlin/org/wfanet/measurement/edpaggregator/rawimpressions/RawImpressionStream.kt
stevenwarejones: Verified — Design B separates I/O from CPU via cpuDispatcher.limitedParallelism + inFlight semaphore. Better than channelFlow for this use case.

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3954)* status: 22 of 26 files reviewed, 3 unresolved discussions (waiting on stevenwarejones).

stevenwarejones (approved):
> @stevenwarejones partially reviewed 4 files and all commit messages, made 1 comment, and resolved 3 discussions.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3954)* status: all files reviewed, 1 unresolved discussion (waiting on Marco-Premier).

___
*[`src/test/kotlin/org/wfanet/measurement/edpaggregator/rawimpressions/RawImpressionSourceTest.kt` line 105 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3954#-OurWGsnEuQCcfGEqxVz:-OurWGsnEuQCcfGEqxW-:byq3e96) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7f595c5e18a61be3e479eeb8b1718932ab60a2f4/src/test/kotlin/org/wfanet/measurement/edpaggregator/rawimpressions/RawImpressionSourceTest.kt#L105)):*
> ```Kotlin
> 
>   private fun row(eventId: String): ParquetRow =
>     ParquetRow.newBuilder()
> ```

nit: use kotlin dsls everywhere

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3954)* status: all files reviewed, 2 unresolved discussions (waiting on Marco-Premier).

___
*[`src/test/kotlin/org/wfanet/measurement/edpaggregator/rawimpressions/RawImpressionSourceTest.kt` line 249 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3954#-OurWKpbDP-hil_KOC4O:-OurWKpbDP-hil_KOC4P:b-7ropln) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7f595c5e18a61be3e479eeb8b1718932ab60a2f4/src/test/kotlin/org/wfanet/measurement/edpaggregator/rawimpressions/RawImpressionSourceTest.kt#L249)):*
> ```Kotlin
>       listOf(
>         ParquetRow.newBuilder()
>           .putColumns("event_id", ParquetValue.newBuilder().setInt64Value(7L).build())
> ```

ditto

stevenwarejones (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3954)* status: all files reviewed, 1 unresolved discussion (waiting on Marco-Premier).

kungfucraig (approved):
> @kungfucraig partially reviewed 26 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3954)* status: all files reviewed, 1 unresolved discussion (waiting on Marco-Premier).

stevenwarejones (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3954)* status: all files reviewed, 1 unresolved discussion (waiting on Marco-Premier).

georgi:
@claude review

stevenwarejones (changes requested):
> @stevenwarejones partially reviewed 7 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3954)* status: all files reviewed, 1 unresolved discussion (waiting on Marco-Premier).

___
*[`src/test/kotlin/org/wfanet/measurement/edpaggregator/rawimpressions/RawImpressionSourceTest.kt` line 105 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3954#-OurWGsnEuQCcfGEqxVz:-OvAZ6gQCJxg7j8rLr2A:b-bvde48) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7f595c5e18a61be3e479eeb8b1718932ab60a2f4/src/test/kotlin/org/wfanet/measurement/edpaggregator/rawimpressions/RawImpressionSourceTest.kt#L105)):*

nit: use kotlin dsls everywhere
</blockquote></details>

still see this not taken care of

Marco-Premier (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3954)* status: 25 of 26 files reviewed, 1 unresolved discussion (waiting on stevenwarejones).

___
*[`src/test/kotlin/org/wfanet/measurement/edpaggregator/rawimpressions/RawImpressionSourceTest.kt` line 105 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3954#-OurWGsnEuQCcfGEqxVz:-OvF3y8j1lTwsbW0YD6n:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7f595c5e18a61be3e479eeb8b1718932ab60a2f4/src/test/kotlin/org/wfanet/measurement/edpaggregator/rawimpressions/RawImpressionSourceTest.kt#L105)):*

still see this not taken care of
</blockquote></details>

Done.

stevenwarejones (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3954)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on Marco-Premier).

stevenwarejones (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3954)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on Marco-Premier).

PR #3959, opened 2026-06-10
feat: add DataProvider validation config proto

jojijac0b:
## Summary

Static process configuration protos for per-`DataProvider` impression validation, in the `config/edpaggregator` package (separate from the over-the-wire API protos in cross-media-measurement-api).

- `DataProviderValidationConfigs` — a `repeated DataProviderValidationConfig` (one entry per opted-in `DataProvider`; absence of an entry means "not validated").
- `DataProviderValidationConfig` — `data_provider` (resource reference), `endpoint`, `tolerance`, and an optional `request_timeout_override`.
- `ToleranceConfig` — a required `warning` band and an optional `failure` band (unset = the report is never failed), plus an optional `minimum_impression_count` (the loader applies 1000 when unset).
- `ToleranceBand` — a `threshold_fraction` and a `minimum_absolute_deviation`; a band triggers only when the deviation strictly exceeds **both**, so small absolute differences don't trip the fractional threshold on medium-sized campaigns.
- `HttpEndpointSink` — the cloud function `endpoint_uri`.

Base of the validation stack (#3960 → #3961 → #3962).

## Test plan

- [x] Proto compiles via Bazel (`data_provider_validation_config_kt_jvm_proto`)

Issue: #3963


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3959)

georgi:
@claude review

src/main/proto/wfa/measurement/config/edpaggregator/edp_validation_config.proto
```diff
@@ -0,0 +1,93 @@
...
+
+option java_package = "org.wfanet.measurement.config.edpaggregator";
+option java_multiple_files = true;
+option java_outer_classname = "EdpValidationConfigProto";
+
+// Static configuration for all DataProvider impression validation.
+message EdpValidationConfigs {
+  // Per-DataProvider validation configurations.
+  // Key: DataProvider resource name (e.g. "dataProviders/{data_provider}").
+  map<string, EdpValidationConfig> configs = 1;
```
stevenwarejones: Make this match the shape of the other configs in this directory. Each sibling file (event_data_provider_configs.proto, data_availability_sync_config.proto, requisition_fetcher_config.proto) uses:

```proto
message FooConfigs {
  repeated FooConfig configs = 1;
}
message FooConfig {
  string data_provider = 1 [
    (google.api.field_behavior) = REQUIRED,
    (google.api.resource_reference).type = "halo.wfanet.org/DataProvider"
  ];
  // ...
}
```

The `map<string, ...>` form here:
- can't annotate the key as a DataProvider `resource_reference` (proto3 map keys take no field options), so the schema doesn't document what the key is or that it's required;
- doesn't get picked up by tooling that walks `resource_reference` fields across configs;
- has unspecified iteration order, which is awkward for textproto golden tests.

Suggested:

```proto
message DataProviderValidationConfigs {
  repeated DataProviderValidationConfig configs = 1;
}
message DataProviderValidationConfig {
  string data_provider = 1 [
    (google.api.field_behavior) = REQUIRED,
    (google.api.resource_reference).type = "halo.wfanet.org/DataProvider"
  ];
  HttpEndpointSink endpoint = 2 [(google.api.field_behavior) = REQUIRED];
  // ... existing fields renumbered ...
}
```

The loader is responsible for enforcing one entry per `data_provider`, same as `DataAvailabilitySyncConfigs` does.

src/main/proto/wfa/measurement/config/edpaggregator/edp_validation_config.proto
```diff
@@ -0,0 +1,93 @@
...
+  FailurePolicy failure_policy = 5;
+}
+
+// HTTP endpoint configuration for a DataProvider's cloud function.
+message HttpEndpointSink {
+  // HTTPS endpoint URI of the cloud function.
+  string endpoint_uri = 1 [(google.api.field_behavior) = REQUIRED];
+
+  // Application-specific configuration passed as JSON.
+  google.protobuf.Struct app_params = 2;
```
stevenwarejones: Drop `app_params`. The cloud function owns its own config; the EDPA isn't going to pre-configure per-DataProvider params here. That leaves just `endpoint_uri`, which is fine as its own message — no need to share with the `HttpEndpointSink` in data_watcher_config.proto since the field sets don't overlap meaningfully.

```proto
message HttpEndpointSink {
  // HTTPS endpoint URI of the cloud function.
  string endpoint_uri = 1 [(google.api.field_behavior) = REQUIRED];
}
```

If a typed param shows up later (e.g. an audience override, a project id), add it as a real field then.

src/main/proto/wfa/measurement/config/edpaggregator/data_provider_validation_config.proto
```diff
@@ -0,0 +1,93 @@
...
+
+  // Timeout for the cloud function call.
+  google.protobuf.Duration request_timeout = 3
+      [(google.api.field_behavior) = REQUIRED];
+
+  // Whether validation is enabled for this DataProvider.
+  bool enabled = 4;
+
+  // Action to take when validation exceeds the failure threshold.
+  FailurePolicy failure_policy = 5;
```
stevenwarejones: Document what happens when the cloud function returns `skipped` (`FILTER_NOT_SUPPORTED`, `ENTITY_NOT_FOUND`, `API_ERROR` per the v2alpha `DataProviderImpressionQueryResponse`). `FailurePolicy` only covers the case where the comparison runs and breaches `failure_threshold_fraction`; it doesn't say what happens when no comparable answer comes back at all.

Plan is to log all skips and treat them as no-op for report disposition (no transition to INVALID) regardless of `failure_policy`. Add that to the doc on `failure_policy`:

```proto
// Action to take when the publisher returned a comparable count AND it
// exceeded the failure threshold. Skipped responses (see
// DataProviderImpressionQueryResponse.SkipReason) are always logged and
// never fail the report, regardless of this setting.
FailurePolicy failure_policy = 5;
```

src/main/proto/wfa/measurement/config/edpaggregator/edp_validation_config.proto
```diff
@@ -0,0 +1,93 @@
...
+
+  // Tolerance thresholds for deviation evaluation.
+  ToleranceConfig tolerance = 2 [(google.api.field_behavior) = REQUIRED];
+
+  // Timeout for the cloud function call.
+  google.protobuf.Duration request_timeout = 3
+      [(google.api.field_behavior) = REQUIRED];
+
+  // Whether validation is enabled for this DataProvider.
+  bool enabled = 4;
```
stevenwarejones: Drop `enabled`. If validation isn't wanted for a DataProvider, just don't include an entry for that DataProvider in `configs` — same signal, one less field, no two-ways-to-say-off footgun. `failure_policy = LOG_ONLY` already covers "validate but don't fail the report".

src/main/proto/wfa/measurement/config/edpaggregator/edp_validation_config.proto
```diff
@@ -0,0 +1,93 @@
...
+  double failure_threshold_fraction = 2;
+
+  // Minimum impression count below which validation is skipped to avoid
+  // false failures on small campaigns. Default: 1000.
+  int64 minimum_impression_count = 3;
+
+  // Statistical confidence multiplier for noise-aware tolerance.
+  // Effective tolerance = max(threshold, k * sqrt(p * (1 - p) / N)).
+  // Default: 4.0.
+  double noise_confidence_multiplier = 4;
```
stevenwarejones: Drop `noise_confidence_multiplier`. Validation compares the publisher's count to the EDP's pre-noise reported value, so there's no DP noise in the comparison to widen the tolerance for. `warning_threshold_fraction` and `failure_threshold_fraction` are enough; `minimum_impression_count` already handles the "small-N is noisy regardless" case.

src/main/proto/wfa/measurement/config/edpaggregator/data_provider_validation_config.proto
```diff
@@ -0,0 +1,93 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+syntax = "proto3";
+
+package wfa.measurement.config.edpaggregator;
+
+import "google/api/field_behavior.proto";
+import "google/api/resource.proto";
```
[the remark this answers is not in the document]
stevenwarejones: +1, this is blocking from my side. Removing both the import and the matching BUILD entry.

src/main/proto/wfa/measurement/config/edpaggregator/edp_validation_config.proto
```diff
@@ -0,0 +1,93 @@
...
+
+// Static configuration for all DataProvider impression validation.
+message EdpValidationConfigs {
+  // Per-DataProvider validation configurations.
+  // Key: DataProvider resource name (e.g. "dataProviders/{data_provider}").
+  map<string, EdpValidationConfig> configs = 1;
+}
+
+// Per-DataProvider validation configuration.
+message EdpValidationConfig {
```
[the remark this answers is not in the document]
stevenwarejones: +1, this is blocking from my side. The PR title was renamed to "DataProvider validation config proto" — file, messages, java_outer_classname, and BUILD targets should follow.

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3959)* status: 0 of 2 files reviewed, 7 unresolved discussions (waiting on jojijac0b and Marco-Premier).

src/main/proto/wfa/measurement/config/edpaggregator/data_provider_validation_config.proto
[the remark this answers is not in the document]
jojijac0b: Kept \`google/api/resource.proto\` (and its BUILD entry) after all — it's now actually used: per the restructure below, \`data_provider\` carries \`(google.api.resource_reference).type = "halo.wfanet.org/DataProvider"\`. What was genuinely unused was \`google/protobuf/struct.proto\` (only \`app_params\` referenced it), so I removed that import and its \`struct_proto\` BUILD dep instead.

src/main/proto/wfa/measurement/config/edpaggregator/edp_validation_config.proto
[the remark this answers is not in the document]
jojijac0b: Done. Renamed the file to \`data_provider_validation_config.proto\`, messages to \`DataProviderValidationConfigs\`/\`DataProviderValidationConfig\`, \`java_outer_classname\` to \`DataProviderValidationConfigProto\`, and BUILD targets to \`data_provider_validation_config_proto\`/\`data_provider_validation_config_kt_jvm_proto\`.

src/main/proto/wfa/measurement/config/edpaggregator/edp_validation_config.proto
jojijac0b: Done. Switched to \`repeated DataProviderValidationConfig configs = 1\` with a required \`string data_provider = 1\` annotated \`(google.api.resource_reference).type = "halo.wfanet.org/DataProvider"\`, matching the sibling configs. Remaining fields renumbered (endpoint=2, tolerance=3, request_timeout=4, failure_policy=5). The loader will enforce one entry per \`data_provider\`, same as \`DataAvailabilitySyncConfigs\`.

src/main/proto/wfa/measurement/config/edpaggregator/edp_validation_config.proto
jojijac0b: Done. \`HttpEndpointSink\` is now just \`endpoint_uri\`. Dropped \`app_params\` and the \`google/protobuf/struct.proto\` import + BUILD dep. Kept it as its own message as you suggested; will add a typed field if a real param shows up later.

src/main/proto/wfa/measurement/config/edpaggregator/data_provider_validation_config.proto
jojijac0b: Done. Added to the \`failure_policy\` doc: it only applies when the publisher returned a comparable count that exceeded the failure threshold; skipped responses (see \`DataProviderImpressionQueryResponse.SkipReason\`) are always logged and never fail the report, regardless of this setting.

src/main/proto/wfa/measurement/config/edpaggregator/edp_validation_config.proto
jojijac0b: Done. Removed \`enabled\` — absence of an entry in \`configs\` now means "not validated" for that DataProvider, and \`failure_policy = LOG_ONLY\` covers "validate but don't fail".

src/main/proto/wfa/measurement/config/edpaggregator/edp_validation_config.proto
jojijac0b: Done. Removed \`noise_confidence_multiplier\` from \`ToleranceConfig\`. \`warning_threshold_fraction\`/\`failure_threshold_fraction\` plus \`minimum_impression_count\` are what's left. Note this cascades to #3960's \`ToleranceEvaluator\`, which used this term in the noise-band formula — I'm removing that term there in the stacked update so the comparison is purely threshold-based against the pre-noise reported value.

src/main/proto/wfa/measurement/config/edpaggregator/BUILD.bazel
```diff
@@ -126,6 +126,26 @@ kt_jvm_proto_library(
...
+    deps = [
+        "@com_google_googleapis//google/api:field_behavior_proto",
+        "@com_google_googleapis//google/api:resource_proto",
+        "@com_google_protobuf//:duration_proto",
+    ],
+)
+
+kt_jvm_proto_library(
+    name = "data_provider_validation_config_kt_jvm_proto",
+    visibility = [
```
stevenwarejones: The visibility here is wrong. It allows only `//src/main/kotlin/org/wfanet/measurement/reporting:__subpackages__`, but reporting doesn't use this config — only the EDPA does. The EDPA-side caller in `org/wfanet/measurement/edpaggregator/...` won't be able to depend on this target.

Drop the `visibility` override and inherit the package default (lines 7-10), which already covers edpaggregator + integration/common — same as every other proto target in this file.

src/main/proto/wfa/measurement/config/edpaggregator/data_provider_validation_config.proto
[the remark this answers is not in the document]
stevenwarejones: Verified — `google.api.resource.proto` is now used by the `data_provider` field's `resource_reference` annotation on line 39, and the matching BUILD dep is legitimate.

src/main/proto/wfa/measurement/config/edpaggregator/data_provider_validation_config.proto
stevenwarejones: Verified — `failure_policy` comment now documents that skipped responses are always logged and never fail the report, regardless of this setting.

src/main/proto/wfa/measurement/config/edpaggregator/edp_validation_config.proto
stevenwarejones: Verified — `repeated DataProviderValidationConfig configs` with a `data_provider` field annotated REQUIRED + `resource_reference.type = "halo.wfanet.org/DataProvider"`. Comment notes the loader enforces uniqueness.

src/main/proto/wfa/measurement/config/edpaggregator/edp_validation_config.proto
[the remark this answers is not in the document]
stevenwarejones: Verified — file renamed to `data_provider_validation_config.proto`, messages to `DataProviderValidationConfigs`/`DataProviderValidationConfig`, `java_outer_classname` to `DataProviderValidationConfigProto`, BUILD targets to `data_provider_validation_config_proto`/`_kt_jvm_proto`.

src/main/proto/wfa/measurement/config/edpaggregator/edp_validation_config.proto
stevenwarejones: Verified — `enabled` removed.

src/main/proto/wfa/measurement/config/edpaggregator/edp_validation_config.proto
stevenwarejones: Verified — `HttpEndpointSink` is now just `endpoint_uri`.

src/main/proto/wfa/measurement/config/edpaggregator/edp_validation_config.proto
stevenwarejones: Verified — `noise_confidence_multiplier` removed.

stevenwarejones (commented):
> @stevenwarejones+AGNT made 7 comments and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3959)* status: 0 of 2 files reviewed, 6 unresolved discussions (waiting on jojijac0b, Marco-Premier, and stevenwarejones).

src/main/proto/wfa/measurement/config/edpaggregator/data_provider_validation_config.proto
```diff
@@ -0,0 +1,90 @@
...
+  ];
+
+  // HTTP endpoint for the DataProvider's validation cloud function.
+  HttpEndpointSink endpoint = 2 [(google.api.field_behavior) = REQUIRED];
+
+  // Tolerance thresholds for deviation evaluation.
+  ToleranceConfig tolerance = 3 [(google.api.field_behavior) = REQUIRED];
+
+  // Timeout for the cloud function call.
+  google.protobuf.Duration request_timeout = 4
```
stevenwarejones: Restructure this as `request_timeout_override`, optional, with the default applied in code:

```proto
// Optional override for the cloud-function call timeout. If unset, the
// EDPA uses the built-in default.
google.protobuf.Duration request_timeout_override = 4;
```

The cloud-function call timeout is usually the same for everyone; per-DataProvider tuning is the exception. Making this REQUIRED forces operators to pick a value with no guidance, and "REQUIRED with a recommended default in the comment" still drifts because operators can pick anything. Putting the default in code keeps the proto signal honest: "this entry is here because we want to deviate from the default."

src/main/proto/wfa/measurement/config/edpaggregator/data_provider_validation_config.proto
```diff
@@ -0,0 +1,90 @@
...
+      [(google.api.field_behavior) = REQUIRED];
+
+  // Action to take when the publisher returned a comparable count AND it
+  // exceeded the failure threshold. Skipped responses (see
+  // `DataProviderImpressionQueryResponse.SkipReason`) are always logged and
+  // never fail the report, regardless of this setting.
+  FailurePolicy failure_policy = 5;
```
stevenwarejones: Drop `FailurePolicy` (and the enum). Make `failure_threshold_fraction` `optional double` — unset means "never fail the report." Keep `warning_threshold_fraction` as-is; warning telemetry is the always-on signal.

```proto
// Deviation fraction that triggers report failure. ...
// If unset, the report is never failed.
// Must be >= `warning_threshold_fraction` when set; the loader enforces this.
optional double failure_threshold_fraction = 2;
```

Use `optional` (proto3 explicit-presence) so the loader can distinguish "unset" from `0.0`.

Move the skip-response paragraph from `failure_policy` onto a top-level comment on `DataProviderValidationConfig`:

```proto
// Skipped responses (see `DataProviderImpressionQueryResponse.SkipReason`)
// are always logged and never fail the report.
```

Reframe the existing "Default: 0.10" comment on `failure_threshold_fraction` as "loader applies if unset", consistent with how `request_timeout_override` works.

stevenwarejones (commented):
> @stevenwarejones partially reviewed 2 files and all commit messages, and resolved 5 discussions.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3959)* status: all files reviewed, 3 unresolved discussions (waiting on jojijac0b and Marco-Premier).

src/main/proto/wfa/measurement/config/edpaggregator/data_provider_validation_config.proto
jojijac0b: Done. Renamed `request_timeout` → `request_timeout_override` (now `optional`, `[(google.api.field_behavior) = OPTIONAL]`), default applied in code: `EdpValidationPostProcessor` uses `config.requestTimeoutOverride` when `hasRequestTimeoutOverride()`, else falls back to the built-in `DEFAULT_TIMEOUT`.

src/main/proto/wfa/measurement/config/edpaggregator/data_provider_validation_config.proto
jojijac0b: Done. Dropped `FailurePolicy` (enum + field, `reserved 5`). `failure_threshold_fraction` is now `optional double` — unset means the report is never failed. Moved the skip-response note to a top-level comment on `DataProviderValidationConfig`. Cascade: `ToleranceEvaluator` now only returns `FAIL` when `hasFailureThresholdFraction()` (added a test for the unset→WARNING case, passing), and `EdpValidationPostProcessor` drops the policy check — it fails the report whenever a DataProvider result is `FAILED`.

src/main/proto/wfa/measurement/config/edpaggregator/BUILD.bazel
jojijac0b: Held off on this one — I think the premise is inverted. The only consumers of `data_provider_validation_config_kt_jvm_proto` are in `reporting/validation` (`ToleranceEvaluator`, `EdpValidationPostProcessor`, consumed by `BasicReportsReportsJob`); there's no `edpaggregator`-side consumer. The package default visibility (lines 7-10) covers `edpaggregator` but not `reporting`, so dropping the override breaks the reporting/validation targets' dependency on this proto. Kept the override as-is. Happy to relocate the validation code under `edpaggregator` instead if that's the intent — lmk.

stevenwarejones: The design changed: the EDPA is the only consumer. The validation cloud-function call is part of the report-disposition path that the EDPA owns; reporting shouldn't be reaching across to invoke a per-EDP webhook. Please move `ToleranceEvaluator`, `EdpValidationPostProcessor`, and the wiring out of `reporting/validation` into `org/wfanet/measurement/edpaggregator/...` (a `validation` package next to the other EDPA components is fine), and update `BasicReportsReportsJob` to call into the EDPA's validation entry point rather than running the validator itself.

Once the consumers move, drop the `visibility` override here — package-default visibility (lines 7-10) is the right scope.

If splitting that move into a follow-up PR keeps this one mergeable, that's fine — flag the follow-up in the issue and keep this proto behind the EDPA-only visibility from the start.

src/main/proto/wfa/measurement/config/edpaggregator/data_provider_validation_config.proto
stevenwarejones: Verified — `request_timeout_override` is `optional` + `[(google.api.field_behavior) = OPTIONAL]`, and `EdpValidationPostProcessor` falls back to `DEFAULT_TIMEOUT` when `hasRequestTimeoutOverride()` is false.

src/main/proto/wfa/measurement/config/edpaggregator/data_provider_validation_config.proto
stevenwarejones: Verified — `FailurePolicy` enum + field removed (`reserved 5; reserved "failure_policy";`), `failure_threshold_fraction` is `optional double` with the unset→never-fail semantic, and the skip-response paragraph moved to the `DataProviderValidationConfig` docstring. Test for unset→WARNING covers the new branch.

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3959)* status: 1 of 2 files reviewed, 3 unresolved discussions (waiting on Marco-Premier and stevenwarejones).

src/main/proto/wfa/measurement/config/edpaggregator/data_provider_validation_config.proto
```diff
@@ -0,0 +1,80 @@
...
+message HttpEndpointSink {
+  // HTTPS endpoint URI of the cloud function.
+  string endpoint_uri = 1 [(google.api.field_behavior) = REQUIRED];
+}
+
+// Tolerance thresholds for deviation between reported and publisher values.
+message ToleranceConfig {
+  // Deviation fraction that triggers a warning. Severe telemetry is emitted
+  // but the report continues. Default: 0.02 (2%).
+  double warning_threshold_fraction = 1;
```
stevenwarejones: Make `warning_threshold_fraction` REQUIRED with a non-zero value, and drop the documented default. Warning telemetry is the always-on monitoring signal — operators must set it explicitly so it's clear what threshold their warnings are derived from. The "loader applies a default" pattern is fine for `failure_threshold_fraction` (where unset = never fail) and `request_timeout_override`, but not for the warning floor.

```proto
// Deviation fraction that triggers a warning. Telemetry is emitted but the
// report continues. Must be > 0.
double warning_threshold_fraction = 1 [(google.api.field_behavior) = REQUIRED];
```

The loader enforces the non-zero invariant.

(`failure_threshold_fraction` as `optional double` is right — leaving it as-is.)

src/main/proto/wfa/measurement/config/edpaggregator/data_provider_validation_config.proto
```diff
@@ -0,0 +1,80 @@
...
+
+  // Deviation fraction that triggers report failure. If unset, the report is
+  // never failed. Must be >= `warning_threshold_fraction` when set; the loader
+  // enforces this.
+  optional double failure_threshold_fraction = 2;
+
+  // Minimum impression count below which validation is skipped to avoid
+  // false failures on small campaigns. Default: 1000.
+  int64 minimum_impression_count = 3;
```
stevenwarejones: Make `minimum_impression_count` `optional int64`, loader applies the default of 1000. Same pattern as `failure_threshold_fraction` and `request_timeout_override`.

```proto
// Minimum impression count below which the cloud-function API call is not
// made and validation is skipped. If unset, the loader applies 1000.
optional int64 minimum_impression_count = 3;
```

Reframe the comment around the API-call behavior — "API calls below this threshold will not be made" — rather than the "false failures on small campaigns" framing, which describes the motivation rather than the observable behavior.

stevenwarejones (commented):
> @stevenwarejones partially reviewed 1 file and all commit messages, and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3959)* status: all files reviewed, 3 unresolved discussions (waiting on jojijac0b and Marco-Premier).

src/main/proto/wfa/measurement/config/edpaggregator/data_provider_validation_config.proto
```diff
@@ -0,0 +1,80 @@
...
+// Tolerance thresholds for deviation between reported and publisher values.
+message ToleranceConfig {
+  // Deviation fraction that triggers a warning. Severe telemetry is emitted
+  // but the report continues. Default: 0.02 (2%).
+  double warning_threshold_fraction = 1;
+
+  // Deviation fraction that triggers report failure. If unset, the report is
+  // never failed. Must be >= `warning_threshold_fraction` when set; the loader
+  // enforces this.
+  optional double failure_threshold_fraction = 2;
```
stevenwarejones: Document the boundary semantic explicitly on both threshold fields. The evaluator (#3960) uses strict `>`, so equality is PASS / WARNING respectively. Without it in the proto doc, the contract is implicit.

```proto
// Deviation fraction that triggers a warning. ...
// A deviation strictly greater than this fraction triggers WARNING; equality is PASS.
double warning_threshold_fraction = 1;

// Deviation fraction that triggers report failure. ...
// A deviation strictly greater than this fraction triggers FAIL; equality is WARNING.
optional double failure_threshold_fraction = 2;
```

src/main/proto/wfa/measurement/config/edpaggregator/BUILD.bazel
stevenwarejones: Verified — the `visibility` override on `data_provider_validation_config_kt_jvm_proto` is gone (now falls back to package default, which covers `edpaggregator`), and the consumers (`ToleranceEvaluator`, `ValidationCloudFunctionClient`, `EdpValidationPostProcessor`, `EdpValidationMetrics`) all moved to `edpaggregator/validation/` across the stack.

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3959)* status: 1 of 2 files reviewed, 4 unresolved discussions (waiting on jojijac0b, Marco-Premier, and stevenwarejones).

src/main/proto/wfa/measurement/config/edpaggregator/data_provider_validation_config.proto
stevenwarejones: Still open from last round — `warning_threshold_fraction` is still un-REQUIRED with the documented default. Can you address before the stack merges?

src/main/proto/wfa/measurement/config/edpaggregator/data_provider_validation_config.proto
stevenwarejones: Still open from last round — no boundary semantic in the comments on `warning_threshold_fraction` / `failure_threshold_fraction`. The evaluator in #3960 uses strict `>`, so equality is PASS / WARNING; this should be documented here so the contract isn't implicit. Can you address before merge?

src/main/proto/wfa/measurement/config/edpaggregator/data_provider_validation_config.proto
stevenwarejones: Still open from last round — `minimum_impression_count` is still a plain `int64` (not `optional int64`), so the loader can't distinguish "operator set to 0 to disable the floor" from "operator omitted the field." Can you address before merge?

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3959)* status: 1 of 2 files reviewed, 4 unresolved discussions (waiting on Marco-Premier and stevenwarejones).

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3959)* status: all files reviewed, 3 unresolved discussions (waiting on Marco-Premier).

src/main/proto/wfa/measurement/config/edpaggregator/data_provider_validation_config.proto
```diff
@@ -0,0 +1,80 @@
...
+}
+
+// HTTP endpoint configuration for a DataProvider's cloud function.
+message HttpEndpointSink {
+  // HTTPS endpoint URI of the cloud function.
+  string endpoint_uri = 1 [(google.api.field_behavior) = REQUIRED];
+}
+
+// Tolerance thresholds for deviation between reported and publisher values.
+message ToleranceConfig {
```
stevenwarejones, line 64: Restructure `ToleranceConfig`. Factor warning and failure into a shared `ToleranceBand` and add an absolute-deviation floor on each band — small absolute differences shouldn't trip the fractional threshold (e.g. publisher=1001 vs reported=900 today flags FAIL on 10.1% even though 101 impressions is noise).

Use this:

```proto
message ToleranceConfig {
  ToleranceBand warning = 1 [(google.api.field_behavior) = REQUIRED];

  // If unset, the report is never failed. Use `hasFailure()` to check.
  ToleranceBand failure = 2;

  // Minimum publisher impression count below which validation is skipped.
  optional int64 minimum_impression_count = 3;
}

// Threshold for a tolerance band. The band is exceeded when BOTH the fractional
// deviation AND the absolute deviation are strictly greater than these values.
message ToleranceBand {
  // Fractional deviation that triggers the band, e.g. 0.10 for 10%.
  double threshold_fraction = 1 [(google.api.field_behavior) = REQUIRED];

  // Absolute deviation, in impressions, that must also be exceeded for the
  // band to trigger.
  int64 minimum_absolute_deviation = 2 [(google.api.field_behavior) = REQUIRED];
}
```

This is a breaking change vs the in-flight stack — `ToleranceEvaluator` (#3960) and `EdpValidationPostProcessor` (#3962) need to read `config.warning.thresholdFraction` / `config.warning.minimumAbsoluteDeviation` and the same for `failure`. Do it now while the stack is unmerged.

Supersedes the open `warning_threshold_fraction` / `failure_threshold_fraction` threads.

jojijac0b: Done in `429f53569`. Restructured `ToleranceConfig` to your shape: `ToleranceBand warning = 1 [REQUIRED]`, `ToleranceBand failure = 2` (optional, `hasFailure()`), `optional int64 minimum_impression_count = 3`. New `ToleranceBand { double threshold_fraction [REQUIRED]; int64 minimum_absolute_deviation [REQUIRED] }` — a band trips only when the deviation is strictly `>` BOTH the fraction and the absolute floor (closes the 1001-vs-900 medium-N gap). Cascaded into #3960 (`ToleranceEvaluator` reads the bands with the AND semantic, regression tests added) and #3962. As you noted, this supersedes the `warning_threshold_fraction` / `failure_threshold_fraction` threads.

src/main/proto/wfa/measurement/config/edpaggregator/data_provider_validation_config.proto
jojijac0b: Addressed via the `ToleranceBand` restructure in `429f53569` (see the `ToleranceConfig` restructure thread). `warning` is now a `[REQUIRED]` `ToleranceBand` whose `threshold_fraction` is itself `[REQUIRED]`, with no documented default.

src/main/proto/wfa/measurement/config/edpaggregator/data_provider_validation_config.proto
jojijac0b: Addressed in `429f53569`. `ToleranceBand.threshold_fraction` and `minimum_absolute_deviation` both document the strict-`>` boundary (equality does not trigger), and `ToleranceEvaluator.evaluate()` (#3960) documents the same (deviation exactly at the warning fraction is PASS, at the failure fraction is WARNING).

src/main/proto/wfa/measurement/config/edpaggregator/data_provider_validation_config.proto
jojijac0b: Done in `429f53569` — `optional int64 minimum_impression_count`, comment reframed around the API-call-skip behavior plus the loader-applies-1000 default.

src/main/proto/wfa/measurement/config/edpaggregator/data_provider_validation_config.proto
stevenwarejones: Verified — `warning` is now a `[REQUIRED]` `ToleranceBand` whose `threshold_fraction` is itself `[REQUIRED]`, addressed by the `ToleranceConfig` restructure.

src/main/proto/wfa/measurement/config/edpaggregator/data_provider_validation_config.proto
stevenwarejones: Verified — `ToleranceBand.threshold_fraction` and `minimum_absolute_deviation` both document the strict-`>` boundary (equality does not trigger).

src/main/proto/wfa/measurement/config/edpaggregator/data_provider_validation_config.proto
stevenwarejones: Verified — `optional int64 minimum_impression_count` with the loader-applies-1000 default and the reframed comment.

src/main/proto/wfa/measurement/config/edpaggregator/data_provider_validation_config.proto
stevenwarejones: Verified — `ToleranceConfig` matches the proposed shape: `warning: ToleranceBand` REQUIRED (line 70), `failure: ToleranceBand` non-optional message with `hasFailure()` for presence (line 75), `minimum_impression_count: optional int64` (line 81), `ToleranceBand.threshold_fraction` and `minimum_absolute_deviation` both REQUIRED (lines 95, 100). Class docstring covers the AND semantic and strict-`>` boundary.

Minor nit on line 73: comment says `has_failure()` but Kotlin readers see `hasFailure()` (camelCase). Either align to `hasFailure()` or drop the parenthetical.

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3959)* status: 1 of 2 files reviewed, 4 unresolved discussions (waiting on Marco-Premier and stevenwarejones).

stevenwarejones (approved):
> @stevenwarejones partially reviewed 1 file and all commit messages, and resolved 4 discussions.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3959)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on Marco-Premier).

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3959)* status: all files reviewed, 1 unresolved discussion (waiting on jojijac0b and Marco-Premier).

___
*[`src/main/proto/wfa/measurement/config/edpaggregator/data_provider_validation_config.proto` line 56 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3959#-OvztIcG0Fi9M_GYtfXl:-OvztIcG0Fi9M_GYtfXm:bmeffj3) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/429f53569c37d89da9809c4312a4bf685977b5c0/src/main/proto/wfa/measurement/config/edpaggregator/data_provider_validation_config.proto#L56)):*
> ```proto
>       [(google.api.field_behavior) = OPTIONAL];
> 
>   reserved 5;
> ```

git rid of these reserved - we can still have breaking changes.

src/main/proto/wfa/measurement/config/edpaggregator/data_provider_validation_config.proto
jojijac0b: Both done in `1d4209670`:
- **Line 73:** `has_failure()` → `hasFailure()` (the Kotlin accessor).
- **Also dropped `reserved 5; reserved "failure_policy";`** per your other note — the proto is unreleased, so we don't need to reserve the field number (breaking changes are fine until first release).

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3959)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on Marco-Premier and stevenwarejones).

___
*[`src/main/proto/wfa/measurement/config/edpaggregator/data_provider_validation_config.proto` line 56 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3959#-OvztIcG0Fi9M_GYtfXl:-Ow4-T_aBoCXOegh4J3X:b6ioqje) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/429f53569c37d89da9809c4312a4bf685977b5c0/src/main/proto/wfa/measurement/config/edpaggregator/data_provider_validation_config.proto#L56)):*

git rid of these reserved - we can still have breaking changes.
</blockquote></details>

Verified — `reserved 5` / `reserved "failure_policy"` are gone in the latest force-push. Field numbers freely renumbered.

src/main/proto/wfa/measurement/config/edpaggregator/data_provider_validation_config.proto
```diff
@@ -0,0 +1,98 @@
...
+// Skipped responses (see `DataProviderImpressionQueryResponse.SkipReason`) are
+// always logged and never fail the report.
+message DataProviderValidationConfig {
+  // Resource name of the `DataProvider` this configuration applies to.
+  string data_provider = 1 [
+    (google.api.field_behavior) = REQUIRED,
+    (google.api.resource_reference).type = "halo.wfanet.org/DataProvider"
+  ];
+
+  // HTTP endpoint for the DataProvider's validation cloud function.
```
stevenwarejones: nit: say "HTTPS endpoint" here to match `HttpEndpointSink.endpoint_uri` on line 59 — the CF client (#3961) presents a Google ID token bound to the endpoint as audience, which only works against HTTPS.

stevenwarejones (commented):
> @stevenwarejones reviewed all commit messages, made 1 comment, and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3959)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on jojijac0b and Marco-Premier).

___
*[`src/main/proto/wfa/measurement/config/edpaggregator/data_provider_validation_config.proto` line 57 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3959#-Ow40_SK945SRK-VssJ6:-Ow40_SK945SRK-VssJ7:b-igbchb) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1d4209670c38e7a4573841a0f45b02dfce0d6201/src/main/proto/wfa/measurement/config/edpaggregator/data_provider_validation_config.proto#L57)):*
> ```proto
> }
> 
> // HTTP endpoint configuration for a DataProvider's cloud function.
> ```

nit: https

src/main/proto/wfa/measurement/config/edpaggregator/data_provider_validation_config.proto
stevenwarejones: Verified — `hasFailure()` (correct Kotlin accessor) on line 70 of the head SHA, and `reserved 5` / `reserved "failure_policy"` are gone. Agreed: proto is unreleased, no need to reserve.

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3959)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on jojijac0b, Marco-Premier, and stevenwarejones).

stevenwarejones (approved):
> @stevenwarejones partially reviewed 1 file.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3959)* status: all files reviewed, 1 unresolved discussion (waiting on jojijac0b and Marco-Premier).

jojijac0b:
Done — line 45 now reads "HTTPS endpoint for the DataProvider's validation cloud function." to match `HttpEndpointSink.endpoint_uri`. Cascaded the rebase through #3960 → #3961 → #3962 so the stack stays linear.

src/main/proto/wfa/measurement/config/edpaggregator/data_provider_validation_config.proto
stevenwarejones: Verified — line 45 now says `// HTTPS endpoint for the DataProvider's validation cloud function.` Satisfied.

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3959)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on Marco-Premier, stevenwarejones, and ugur47).

___
*[`src/main/proto/wfa/measurement/config/edpaggregator/data_provider_validation_config.proto` line 57 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3959#-Ow40_SK945SRK-VssJ6:-OwNukzH9F53r8c7s9O1:b-c3tecd) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1d4209670c38e7a4573841a0f45b02dfce0d6201/src/main/proto/wfa/measurement/config/edpaggregator/data_provider_validation_config.proto#L57)):*

nit: https
</blockquote></details>

Still outstanding — line 45 (the `endpoint` field doc) got the HTTPS fix in the latest force-push, but the message-level KDoc on line 57 still says `// HTTP endpoint configuration for a DataProvider's cloud function.` Same one-character fix here: `// HTTPS endpoint configuration ...`

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3959)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on Marco-Premier, stevenwarejones, and ugur47).

___
*[`src/main/proto/wfa/measurement/config/edpaggregator/data_provider_validation_config.proto` line 57 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3959#-Ow40_SK945SRK-VssJ6:-OwPiLXoE0a-GE697dS0:b-s3et1z) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1d4209670c38e7a4573841a0f45b02dfce0d6201/src/main/proto/wfa/measurement/config/edpaggregator/data_provider_validation_config.proto#L57)):*
<details><summary><i>Previously, stevenwarejones+AGNT (Agent AGNT for Steven Ware Jones) wrote…</i></summary><blockquote>

Still outstanding — line 45 (the `endpoint` field doc) got the HTTPS fix in the latest force-push, but the message-level KDoc on line 57 still says `// HTTP endpoint configuration for a DataProvider's cloud function.` Same one-character fix here: `// HTTPS endpoint configuration ...`
</blockquote></details>

Reminder — still outstanding. The other three PRs in the stack have all moved on past their open asks; this one-character doc fix (`HTTP` → `HTTPS` on line 57) is the only thing left on #3959. Please push a fix-up so the stack can move.

src/main/proto/wfa/measurement/config/edpaggregator/data_provider_validation_config.proto
```diff
@@ -0,0 +1,98 @@
...
+  // Tolerance thresholds for deviation evaluation.
+  ToleranceConfig tolerance = 3 [(google.api.field_behavior) = REQUIRED];
+
+  // Optional override for the cloud-function call timeout. If unset, the EDPA
+  // uses the built-in default.
+  google.protobuf.Duration request_timeout_override = 4
+      [(google.api.field_behavior) = OPTIONAL];
+}
+
+// HTTP endpoint configuration for a DataProvider's cloud function.
```
stevenwarejones: Reminder — still says `HTTP`. The `endpoint` field doc on line 45 got the HTTPS fix in the last push, but this message-level KDoc was missed. Same one-character fix: `// HTTPS endpoint configuration ...` The other three PRs in the stack have all moved past their open asks; this is the only thing left on #3959.

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3959)* status: 1 of 2 files reviewed, 2 unresolved discussions (waiting on jojijac0b, Marco-Premier, stevenwarejones, and ugur47).

stevenwarejones (commented):
> @stevenwarejones partially reviewed 1 file and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3959)* status: all files reviewed, 2 unresolved discussions (waiting on jojijac0b, Marco-Premier, and ugur47).

src/main/proto/wfa/measurement/config/edpaggregator/data_provider_validation_config.proto
jojijac0b: Fixed in `7b9b57263` — the message-level KDoc on line 57 now says `// HTTPS endpoint configuration for a DataProvider's cloud function.` Cascaded through #3960 → #3961 → #3962.

stevenwarejones (approved):
> @stevenwarejones partially reviewed 1 file and all commit messages, resolved 1 discussion, and dismissed @stevenwarejones+AGNT from a discussion.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3959)* status: all files reviewed, 1 unresolved discussion (waiting on Marco-Premier, stevenwarejones+AGNT, and ugur47).

ugur47 (approved):
> @ugur47 partially reviewed 2 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3959)* status: all files reviewed, 1 unresolved discussion (waiting on Marco-Premier and stevenwarejones+AGNT).

src/main/proto/wfa/measurement/config/edpaggregator/data_provider_validation_config.proto
stevenwarejones: Verified — `HttpEndpointSink` message-level KDoc on line 57 now says `// HTTPS endpoint configuration for a DataProvider's cloud function.` Cascaded through #3960 → #3961 → #3962. Satisfied.

stevenwarejones (commented):
> @stevenwarejones+AGNT made 1 comment and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3959)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on Marco-Premier).

jojijac0b (commented):
> @jojijac0b partially reviewed 2 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3959)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on Marco-Premier).

PR #3960, opened 2026-06-10
feat: add ToleranceEvaluator for impression validation

jojijac0b:
## Summary

Adds `ToleranceEvaluator` — a pure two-count comparison of a reported impression count against an expected count (the publisher's source-of-truth count, scaled to the EDPA's measured population by the caller).

- `evaluate(reportedCount, expectedCount, config)` returns PASS / WARNING / FAIL, or SKIPPED when there is no expected count to compare against.
- Each `ToleranceBand` (warning, failure) triggers only when the deviation strictly exceeds **both** its `threshold_fraction` and its `minimum_absolute_deviation`. Boundaries are exclusive (`>`): a deviation exactly at the warning fraction is PASS, exactly at the failure fraction is WARNING.
- An unset failure band means the report is never failed.
- Fails fast (`require`) if a set failure band's `threshold_fraction` is below the warning band's.
- Scaling and minimum-count gating are the caller's responsibility (the post-processor, #3962) — the evaluator carries no statistical assumptions.

Stacked on #3959 (config proto).

## Test plan

`ToleranceEvaluatorTest` (passes on the dev server):
- [x] PASS on exact match / within warning; WARNING / FAIL above the respective bands
- [x] strict-`>` boundary (deviation exactly at warning → PASS, at failure → WARNING)
- [x] absolute-deviation floor downgrades FAIL→WARNING and WARNING→PASS when the abs gap is within the floor
- [x] WARNING instead of FAIL when the failure band is unset
- [x] SKIPPED when the expected count is zero
- [x] throws when the failure fraction is below the warning fraction

Issue: #3963


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3960)

georgi:
@claude review

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3960)* status: 4 of 6 files reviewed, all discussions resolved (waiting on Marco-Premier).

stevenwarejones:
The PR summary is out of date — it describes a noise-aware formula (`max(threshold, k * sqrt(p * (1-p) / N))`, "confidence multiplier (default 4.0)") that the code doesn't implement. The current `evaluate()` does a flat-threshold comparison after rescaling the publisher count by `vidSamplingWidth`; no noise band, no `k`, no `sqrt`. Please update the summary and test plan to match.

src/main/kotlin/org/wfanet/measurement/edpaggregator/validation/ToleranceEvaluator.kt
```diff
@@ -0,0 +1,121 @@
...
+        effectiveTolerance = null,
+      )
+    }
+
+    val deviationFraction = abs(expectedCount - reportedCount).toDouble() / expectedCount.toDouble()
+
+    val warningTolerance = config.warningThresholdFraction
+
+    // An unset failure threshold means the report is never failed; only the warning band applies.
+    val verdict =
```
stevenwarejones, line 108: The boundary semantics are strict (`>`), so `deviation == warningThreshold` returns PASS and `deviation == failureThreshold` returns WARNING. That's a reasonable product choice — thresholds are exclusive upper bounds, "deviations above X warn / fail" — but it's not documented and not pinned by a test, so a refactor could silently flip to inclusive (`>=`).

Add boundary tests:

```kotlin
@Test
fun `evaluate returns PASS when deviation equals warning threshold`() {
  // deviation = exactly 0.02 → PASS
}

@Test
fun `evaluate returns WARNING when deviation equals failure threshold`() {
  // deviation = exactly 0.10 → WARNING
}
```

Filed a parallel comment on the proto (#3959) to add a sentence to the `warning_threshold_fraction` / `failure_threshold_fraction` doc making the strict-greater-than semantic explicit.

src/main/kotlin/org/wfanet/measurement/edpaggregator/validation/ToleranceEvaluator.kt
```diff
@@ -0,0 +1,121 @@
...
+   * @param config tolerance configuration with warning and failure thresholds.
+   * @param vidSamplingWidth VID sampling fraction (0.0 to 1.0) applied by the EDPA.
+   * @return the evaluation result with verdict and deviation details.
+   */
+  fun evaluate(
+    reportedCount: Long,
+    publisherCount: Long,
+    config: ToleranceConfig,
+    vidSamplingWidth: Double = 1.0,
+  ): EvaluationResult {
```
stevenwarejones, line 78: Fail fast on the warn/fail invariant. The proto promises the loader enforces `failure_threshold_fraction >= warning_threshold_fraction`, but no loader exists in this stack yet — until it lands, this evaluator is the only thing standing between a malformed config and a misleading verdict.

```kotlin
fun evaluate(...): EvaluationResult {
  require(
    !config.hasFailureThresholdFraction() ||
      config.failureThresholdFraction >= config.warningThresholdFraction
  ) {
    "failureThresholdFraction must be >= warningThresholdFraction"
  }
  // ... existing body
}
```

Add a test for the violation. Once a real loader exists, this `require` becomes belt-and-suspenders rather than the sole check — fine to keep.

Separately: where will the config loader live, and which PR adds it? The "the loader enforces ..." promises in #3959 currently point at nothing.

src/main/kotlin/org/wfanet/measurement/edpaggregator/validation/ToleranceEvaluator.kt
```diff
@@ -0,0 +1,121 @@
...
+  /**
+   * Evaluates the deviation between reported and publisher impression counts.
+   *
+   * @param reportedCount impression count from the EDPA report.
+   * @param publisherCount impression count from the publisher's source of truth.
+   * @param config tolerance configuration with warning and failure thresholds.
+   * @param vidSamplingWidth VID sampling fraction (0.0 to 1.0) applied by the EDPA.
+   * @return the evaluation result with verdict and deviation details.
+   */
+  fun evaluate(
```
stevenwarejones, line 74: Drop `vidSamplingWidth` entirely. Scale the publisher count to the EDPA's measured population in the caller, then pass the already-scaled value in. The evaluator becomes a pure two-count comparison with no statistical assumptions baked in.

```kotlin
fun evaluate(
  reportedCount: Long,
  expectedCount: Long,
  config: ToleranceConfig,
): EvaluationResult
```

Rename the second parameter to `expectedCount` so it's clear the comparison is reported-vs-expected, not reported-vs-raw-publisher. The caller (likely `EdpValidationPostProcessor`) holds the VID-sampling context; encoding "expected = publisher * vidSamplingWidth" in a separate, named step there is clearer than burying it inside the evaluator.

This also removes:
- the `(publisherCount * vidSamplingWidth).toLong()` truncation issue
- the silent acceptance of `vidSamplingWidth > 1.0` or `< 0.0`
- the `vidSamplingWidth = 1.0` defaulted parameter
- the "applies VID sampling width" test, which moves to the caller

Update the docstring (lines 22-29) and the `EvaluationResult.publisherCount` field — should become `expectedCount` or `comparisonBaseline`.

src/main/kotlin/org/wfanet/measurement/edpaggregator/validation/ToleranceEvaluator.kt
```diff
@@ -0,0 +1,121 @@
...
+    val deviationFraction = abs(expectedCount - reportedCount).toDouble() / expectedCount.toDouble()
+
+    val warningTolerance = config.warningThresholdFraction
+
+    // An unset failure threshold means the report is never failed; only the warning band applies.
+    val verdict =
+      when {
+        config.hasFailureThresholdFraction() &&
+          deviationFraction > config.failureThresholdFraction -> Verdict.FAIL
+        deviationFraction > warningTolerance -> Verdict.WARNING
```
stevenwarejones: Still open from last round — code uses strict `>` (lines 98, 99) with no documenting comment, no boundary tests. Can you address before merge?

Tiny add-on while you're here: the class docstring (lines 22-29) still has leftover wording from earlier rounds — "publisher's count scaled by the VID sampling fraction" and "too noisy to validate." Worth a once-over.

src/main/kotlin/org/wfanet/measurement/edpaggregator/validation/ToleranceEvaluator.kt
```diff
@@ -0,0 +1,121 @@
...
+  /**
+   * Evaluates the deviation between reported and publisher impression counts.
+   *
+   * @param reportedCount impression count from the EDPA report.
+   * @param publisherCount impression count from the publisher's source of truth.
+   * @param config tolerance configuration with warning and failure thresholds.
+   * @param vidSamplingWidth VID sampling fraction (0.0 to 1.0) applied by the EDPA.
+   * @return the evaluation result with verdict and deviation details.
+   */
+  fun evaluate(
```
stevenwarejones, line 74: Still open from last round — no `require` added, and the bigger question is unanswered: where will the config loader live, and which PR adds it? The proto in #3959 has multiple "the loader enforces this" promises (uniqueness, warn-≤-fail ordering, optional-field defaults) that today point at code that doesn't exist. Without a loader, this evaluator is the only thing standing between a malformed config and a misleading verdict.

Either add the `require` here (and a test) so the evaluator is honest about what it trusts, or point at the PR that adds the loader so I can verify the invariants are checked once.

src/main/kotlin/org/wfanet/measurement/edpaggregator/validation/ToleranceEvaluator.kt
```diff
@@ -0,0 +1,121 @@
...
+    reportedCount: Long,
+    publisherCount: Long,
+    config: ToleranceConfig,
+    vidSamplingWidth: Double = 1.0,
```
stevenwarejones: Disagree on the simplification reasoning. Most callers will have `vidSamplingWidth = 1.0`, so a defaulted parameter buys nothing — they pass it or they don't, same evaluator math. The point of dropping it is that scaling is the caller's responsibility, not the evaluator's. The evaluator should be a pure two-count comparison.

Drop the parameter, take `expectedCount: Long`, let the post-processor compute `expectedCount = (publisherCount * vidSamplingWidth).toLong()` (or `.roundToLong()` — caller's call) and pass it in.

stevenwarejones (commented):
> @stevenwarejones partially reviewed 2 files and all commit messages, and resolved 3 discussions.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3960)* status: 3 of 6 files reviewed, 3 unresolved discussions (waiting on jojijac0b and Marco-Premier).

src/main/kotlin/org/wfanet/measurement/edpaggregator/validation/ToleranceEvaluator.kt
jojijac0b: Done in `9855add4a` — dropped `vidSamplingWidth`. `evaluate(reportedCount, expectedCount, config)` is now a pure two-count comparison; the post-processor (#3962) scales the publisher count to the EDPA measured population before calling. Since the evaluator no longer sees the publisher count, I also moved the `minimum_impression_count` gate up into the post-processor (with tests for the below-minimum skip and the scaling).

src/main/kotlin/org/wfanet/measurement/edpaggregator/validation/ToleranceEvaluator.kt
jojijac0b: Done in `9855add4a` — added `require(!config.hasFailure() || config.failure.thresholdFraction >= config.warning.thresholdFraction)` at the top of `evaluate()`, with a regression test. On the loader: there is still no config loader in this stack, so the proto's "loader enforces …" promises (uniqueness, warn≤fail, optional defaults) are not yet backed by code — this `require` is the interim guard for the warn≤fail invariant. I think the loader belongs in a `ConfigLoader` under `edpaggregator/validation` next to the evaluator; I will track it as a follow-up unless you want it in this stack.

src/main/kotlin/org/wfanet/measurement/edpaggregator/validation/ToleranceEvaluator.kt
jojijac0b: Done in `9855add4a` — documented the strict-`>` boundary on `evaluate()` and added boundary tests using exact-binary fractions (0.5/0.25) so they do not rely on float equality, plus absolute-floor tests for the 1001-vs-900 case. Also cleaned the stale "VID sampling fraction" / "too noisy to validate" wording from the class docstring — the evaluator no longer scales or gates on the minimum (both moved to the caller).

src/main/kotlin/org/wfanet/measurement/edpaggregator/validation/ToleranceEvaluator.kt
stevenwarejones: Verified — `require(!config.hasFailure() || config.failure.thresholdFraction >= config.warning.thresholdFraction)` at the top of `evaluate()` (lines 79-84) with the `evaluate throws when failure fraction is below warning fraction` regression test. Loader question is fine to defer to the wiring PR.

src/main/kotlin/org/wfanet/measurement/edpaggregator/validation/ToleranceEvaluator.kt
stevenwarejones: Verified — `evaluate(reportedCount, expectedCount, config)` is now pure two-count; `EvaluationResult.expectedCount` (line 55), scaling moved to the post-processor in #3962, no defaulted parameter remaining.

src/main/kotlin/org/wfanet/measurement/edpaggregator/validation/ToleranceEvaluator.kt
stevenwarejones: Verified — docstring on `evaluate()` (lines 60-72) documents the strict-`>` boundary; `bandExceeded()` (line 140) uses strict `>` on both fraction and absolute. Boundary tests landed: `evaluate returns PASS when deviation exactly equals warning fraction`, `evaluate returns WARNING when deviation exactly equals failure fraction`, plus the absolute-floor cases (`...within the failure floor`, `...absolute is within floor`). Docstring rot cleaned up — "too noisy to validate" / VID-sampling wording gone.

stevenwarejones (changes requested):
> @stevenwarejones+AGNT made 6 comments.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3960)* status: 3 of 6 files reviewed, 6 unresolved discussions (waiting on jojijac0b, Marco-Premier, and stevenwarejones).

stevenwarejones (commented):
> @stevenwarejones partially reviewed 3 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3960)* status: all files reviewed, 3 unresolved discussions (waiting on Marco-Premier).

stevenwarejones (approved):
> @stevenwarejones resolved 3 discussions.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3960)* status: all files reviewed, 3 unresolved discussions (waiting on jojijac0b and Marco-Premier).

src/main/kotlin/org/wfanet/measurement/edpaggregator/validation/ToleranceEvaluator.kt
jojijac0b: Addressed in `9855add4a` (see the follow-up thread on this field): the strict-`>` boundary is documented on `evaluate()` and pinned by boundary tests using exact-binary fractions (deviation exactly at the warning fraction → PASS, at the failure fraction → WARNING).

src/main/kotlin/org/wfanet/measurement/edpaggregator/validation/ToleranceEvaluator.kt
jojijac0b: Addressed in `9855add4a`: added `require(!config.hasFailure() || config.failure.thresholdFraction >= config.warning.thresholdFraction)` with a regression test. No loader exists in the stack yet — this is the interim guard; proposed a `ConfigLoader` in `edpaggregator/validation` as a follow-up.

src/main/kotlin/org/wfanet/measurement/edpaggregator/validation/ToleranceEvaluator.kt
jojijac0b: Addressed in `9855add4a`: `vidSamplingWidth` dropped — `evaluate(reportedCount, expectedCount, config)` is now a pure two-count comparison and the post-processor (#3962) scales the publisher count before calling.

jojijac0b:
Summary updated — it now describes the pure two-count `evaluate(reportedCount, expectedCount, config)`, the `ToleranceBand` warning/failure bands with the absolute-deviation floor, the warn≤fail `require`, and a real test plan. No more noise-aware-formula wording.

src/main/kotlin/org/wfanet/measurement/edpaggregator/validation/ToleranceEvaluator.kt
stevenwarejones: :lgtm:

stevenwarejones (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3960)* status: all files reviewed, 3 unresolved discussions (waiting on Marco-Premier and stevenwarejones+AGNT).

src/main/kotlin/org/wfanet/measurement/edpaggregator/validation/ToleranceEvaluator.kt
stevenwarejones: Verified — strict-`>` boundary documented on `evaluate()` KDoc (lines 63-65), and pinned by `evaluate returns PASS when deviation exactly equals warning fraction` and `evaluate returns WARNING when deviation exactly equals failure fraction` tests in `ToleranceEvaluatorTest`.

src/main/kotlin/org/wfanet/measurement/edpaggregator/validation/ToleranceEvaluator.kt
stevenwarejones: Verified — `require(!config.hasFailure() || config.failure.thresholdFraction >= config.warning.thresholdFraction)` at `ToleranceEvaluator.kt:79-84`, with regression test `evaluate throws when failure fraction is below warning fraction`. Noted: no loader yet; this is the interim guard.

On the loader follow-up: please file a tracking issue (or add a sub-task under #3963) so the "the loader enforces this" promises in #3959 (uniqueness, warn-≤-fail ordering, optional-field defaults) have a home. The `require` here closes the warn/fail-ordering case; uniqueness and `minimum_impression_count` defaulting still need a real owner.

src/main/kotlin/org/wfanet/measurement/edpaggregator/validation/ToleranceEvaluator.kt
stevenwarejones: Verified — `evaluate(reportedCount: Long, expectedCount: Long, config: ToleranceConfig)` is now a pure two-count comparison; no `vidSamplingWidth` left in the evaluator. The post-processor (#3962) scales the publisher count before calling: `EdpValidationPostProcessor.kt:225` does `(publisherCount * row.vidSamplingWidth).toLong()` and passes `expectedCount` in.

stevenwarejones (commented):
> @stevenwarejones+AGNT made 3 comments and resolved 3 discussions.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3960)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on Marco-Premier and ugur47).

jojijac0b:
Filed the loader follow-up as #4122 — it tracks the three "the loader enforces this" promises that currently have interim in-code homes: config uniqueness, `warning ≤ failure` ordering (the `require` here), and optional-field defaults (`minimum_impression_count` → 1000 in #3962). Linked back to #3963 and these threads.

stevenwarejones (approved):
LGTM — all prior blocking threads resolved: strict-`>` boundary documented and tested, `require(failure >= warning)` invariant added with regression test, `vidSamplingWidth` dropped from the evaluator (caller scales). Pure two-count comparison, well-tested.

ugur47 (changes requested):
> @ugur47 made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3960)* status: all files reviewed, 1 unresolved discussion (waiting on jojijac0b and Marco-Premier).

___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/validation/ToleranceEvaluator.kt` line 81 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3960#-OwbSdyl1eOS6dLL02Gp:-OwbSdyl1eOS6dLL02Gq:b-97oum6) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a76c646ce6d082e85ba861675eaa72d45d1e07ec/src/main/kotlin/org/wfanet/measurement/edpaggregator/validation/ToleranceEvaluator.kt#L81)):*
> ```Kotlin
>     require(
>       !config.hasFailure() ||
>         config.failure.thresholdFraction >= config.warning.thresholdFraction
> ```

Nothing forces failure.minimumAbsoluteDeviation >= warning.minimumAbsoluteDeviation.


Here is what can happen here : 

- warning { fraction 0.02, floor 500 }, failure { fraction 0.10, floor 100 } → passes require.
- expected = 1000, reported = 800 → absDev 200, fraction 0.20.
  - failure band: 0.20 > 0.10 ✓ and 200 > 100 ✓ → FAIL
  - warning band would be: 0.20 > 0.02 ✓ but 200 > 500 ✗ → not even a warning



I'd add this check and add a test case for it. But let me know if this is expected

jojijac0b:
@ugur47 good catch — fixed in `f601ffe83`. Added the floor-ordering guard next to the existing fraction check:

```kotlin
require(
  !config.hasFailure() ||
    config.failure.minimumAbsoluteDeviation >= config.warning.minimumAbsoluteDeviation
) {
  "failure band minimum_absolute_deviation must be >= warning band minimum_absolute_deviation"
}
```

So a config where a deviation could FAIL without ever having been a WARNING (your warn-floor 500 / fail-floor 100 case) is now rejected up front. Added a regression test `evaluate throws when failure floor is below warning floor` using exactly that example (reported 800 vs expected 1000).

stevenwarejones (approved):
LGTM on the new head — self-added `require(failure.minimumAbsoluteDeviation >= warning.minimumAbsoluteDeviation)` extends the warn-≤-fail invariant to the absolute-deviation floor, matching the existing `threshold_fraction` invariant, with a regression test. Nice defensive addition.

ugur47 (commented):
> @ugur47 partially reviewed 6 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3960)* status: all files reviewed, 1 unresolved discussion (waiting on jojijac0b and Marco-Premier).

ugur47 (approved):
:lgtm:

> @ugur47 made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3960#-:-Ox0aBIlAjt_H1DVHjbR:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on jojijac0b and Marco-Premier).

jojijac0b:
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/validation/ToleranceEvaluator.kt` line 81 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3960#-OwbSdyl1eOS6dLL02Gp:-Oyd5IoU7hghkxaTKRuS:bggrle) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/a76c646ce6d082e85ba861675eaa72d45d1e07ec/src/main/kotlin/org/wfanet/measurement/edpaggregator/validation/ToleranceEvaluator.kt#L81)):*

Nothing forces failure.minimumAbsoluteDeviation >= warning.minimumAbsoluteDeviation.


Here is what can happen here : 

- warning { fraction 0.02, floor 500 }, failure { fraction 0.10, floor 100 } → passes require.
- expected = 1000, reported = 800 → absDev 200, fraction 0.20.
  - failure band: 0.20 > 0.10 ✓ and 200 > 100 ✓ → FAIL
  - warning band would be: 0.20 > 0.02 ✓ but 200 > 500 ✗ → not even a warning



I'd add this check and add a test case for it. But let me know if this is expected
</blockquote></details>

Done. Added the symmetric guard (failure.minimumAbsoluteDeviation >= warning.minimumAbsoluteDeviation) next to the fraction one, plus a test with your exact scenario (evaluate throws when failure floor is below warning floor).

jojijac0b (commented):
> @jojijac0b resolved 1 discussion and dismissed @ugur47 from a discussion.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3960)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on Marco-Premier).

jojijac0b (commented):
> @jojijac0b partially reviewed 6 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3960)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on Marco-Premier).

PR #3966, opened 2026-06-11
build: Add virtual-people-core-serving module dependency

jojijac0b:
## Summary

Adds the `virtual-people-core-serving` Bazel module — the canonical VirtualPeople `Labeler` used by the VID Labeler TEE app to assign VIDs from a compiled model.

- `bazel_dep` on `virtual-people-core-serving` 0.2.2 + `git_override` (pinned commit).
- Pins `googleapis` to `0.0.0-20251127-8cd3749f` to avoid the yanked `googleapis@0.0.0-bzlmod.1` that `virtual-people-core-serving` otherwise resolves.
- Regenerated `MODULE.bazel.lock` (no Maven changes).

Per dev standards, the dependency introduction is isolated in this PR.

## Test plan

- [x] `bazel mod deps --lockfile_mode=update` resolves cleanly

Issue: #3899

PR #3951, opened 2026-06-09
feat: Add LabelerInputEventGroupSpec for VID labeler-based data generation

stevenwarejones:
Add a new spec format that defines labeler input IDs with demographics and runs them through a VID labeler model (CompiledNode) to produce VIDs. This exercises the full labeling pipeline: fingerprint extraction, demographic correction matrix, hash-based pool routing.

Issue: #3949

Issue #3899, opened 2026-06-03
VID Labeling Dispatcher Cloud Function


Marco-Premier:
Cloud Function that fans out `DataProvider` raw impression uploads into the downstream VID labeling pipeline.
  
On each upload event the dispatcher resolves the applicable `ModelLine`s (type `PROD`, within the active time window), interacts with the VID Repo API, persists a new `RawImpressionUpload` together
  with its child resources (`RawImpressionUploadFile`, `RawImpressionUploadModelLine`, and `PoolAssignmentJob` for memoization-enabled model lines), and enqueues one work item per shard to trigger
  downstream processing.
  
Scope: dispatcher Kotlin implementation, `VidLabelingConfig` proto extensions, VID Repo gRPC integration, shard-based work-item creation, and deployment configuration.

PR #3967, opened 2026-06-11
feat: Add VidAssigner over VirtualPeople Labeler

jojijac0b:
## Summary

Adds the VID Labeler "labeler core" — a thin, testable seam over the canonical VirtualPeople `Labeler`, so the TEE app assigns VIDs without reimplementing the model-tree traversal/hashing.

- `VidAssigner` — interface `assign(LabelerInput): LabelerOutput`.
- `VirtualPeopleVidAssigner` — wraps a `Labeler`; `fromCompiledNodeBlob(...)` builds it from a serialized `CompiledNode` (the `model_blob_path` payload resolved by the dispatcher).
- `VidModelLoader` — loads and builds each compiled model at most once per process, keyed by model blob URI.

Stacked on #3966 (adds `virtual-people-core-serving`).

Follow-ups (separate PRs): the impression → `LabelerInput` mapping (depends on the raw-impression Parquet schema) and the full Path-A VidLabeler pipeline (stacked on the `RawImpressionStream` reader).

## Test plan

- [x] `bazel test //src/test/.../vidlabeler:VidModelLoaderTest` passes
- [x] `bazel build //src/main/.../vidlabeler:vid_assigner` (against virtual-people-core-serving)
- [x] `VirtualPeopleVidAssignerTest` added as an `@Ignore`'d stub for the `fromCompiledNodeBlob` round-trip; activates once #3951 lands `reference_test_model.textproto` on main.

Issue: #3899

### DIFF
--- a/src/main/kotlin/org/wfanet/measurement/edpaggregator/vidlabeler/BUILD.bazel
+++ b/src/main/kotlin/org/wfanet/measurement/edpaggregator/vidlabeler/BUILD.bazel
@@ -6,6 +6,18 @@ package(default_visibility = [
 ])
 
 kt_jvm_library(
+    name = "vid_assigner",
+    srcs = ["VidAssigner.kt"],
+    deps = [
+        "@wfa_common_jvm//imports/java/com/google/protobuf",
+        "@wfa_common_jvm//imports/kotlin/kotlinx/coroutines:core",
+        "@wfa_virtual_people_common//src/main/proto/wfa/virtual_people/common:event_kt_jvm_proto",
+        "@wfa_virtual_people_common//src/main/proto/wfa/virtual_people/common:model_kt_jvm_proto",
+        "@wfa_virtual_people_core_serving//src/main/kotlin/org/wfanet/virtualpeople/core/labeler",
+    ],
+)
+
+kt_jvm_library(
     name = "vid_labeler_app",
     srcs = ["VidLabelerApp.kt"],
     deps = [
--- a/src/main/kotlin/org/wfanet/measurement/edpaggregator/vidlabeler/VidAssigner.kt
+++ b/src/main/kotlin/org/wfanet/measurement/edpaggregator/vidlabeler/VidAssigner.kt
@@ -1,0 +1,85 @@
+/*
+ * Copyright 2026 The Cross-Media Measurement Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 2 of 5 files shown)


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3967)

georgi:
@claude review

stevenwarejones (commented):
This code does the exact same thing the upstream does. I'm not sure why its needed.

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3967#-:-OuxLkNwAjSEn5r_VqoE:bi1eqfm)* status: 1 of 5 files reviewed, all discussions resolved (waiting on Marco-Premier).

jojijac0b:
> This code does the exact same thing the upstream does. I'm not sure why its needed.
> 
> > @stevenwarejones reviewed 1 file and all commit messages, and made 1 comment.
> > _[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3967#-:-OuxLkNwAjSEn5r_VqoE:bi1eqfm)_ status: 1 of 5 files reviewed, all discussions resolved (waiting on Marco-Premier).

You're right that VirtualPeopleVidAssigner is a pure pass-through to Labeler — that's intentional. It's a DI seam, not new logic, and it's there for two reasons:

1. Testability. VidLabeler/VidLabelingSink (#3990) depend on the VidAssigner interface, which lets their unit tests inject a fake assigner that returns known VIDs — so we can test the real sink logic (active-window filtering, building/encrypting/writing labeled output) without standing up a real model. This isn't just convenience: Labeler has a private constructor and can only be created via Labeler.build(compiledNode) with a valid CompiledNode, so there's no way to fake it directly. The interface is what makes substitution possible — VidLabelingSinkTest relies on a 4-line FixedVidAssigner for exactly this.

2. Caching. VidModelLoader builds each model once per model_blob_path and reuses it; Labeler has no caching, so without it we'd rebuild the model per file/shard.

The real VirtualPeopleVidAssigner still gets coverage against an actual model — that's VirtualPeopleVidAssignerTest, currently @Ignored pending the reference model fixture from #3951.

If you'd rather drop the interface and have VidModelLoader cache Labeler directly (with consumers using Labeler), I'm open to it — the tradeoff is losing the fake-injection seam in the sink tests, which would force every sink test to
  construct a real CompiledNode. Happy to go that way if you think it's cleaner; otherwise I'd lean toward keeping the seam. Let me know what you prefer.

stevenwarejones:
Closing since the code fits well into 3990

PR #3974, opened 2026-06-11
fix: gap detection no longer flags data-without-done folders as gaps

dawn-wang22:
## Summary

Fixes a false positive in the EDP-aggregator data availability monitor's **gap detection**.

`findGaps` ran only over `datesWithDoneBlob`. A date folder that has data but no `done` blob (e.g. a recent day still being finalized, or a partially-written day) was therefore reported both as a **gap** and under `datesWithoutDoneBlob`.

**Fix:** compute gaps over all present date folders (`datesWithDoneBlob ∪ datesWithoutDoneBlob`) via a new `presentDates()` helper, so a folder that exists but lacks a `done` blob is reported only under `datesWithoutDoneBlob`, not as a spurious gap.

## Test plan

- Updated the two "without done blob" tests to assert such a folder is not reported as a gap.
- `bazel test //src/test/kotlin/org/wfanet/measurement/edpaggregator/dataavailability:DataAvailabilityMonitorTest` passes.

## Note / scope

This PR intentionally covers **only** the gap-detection fix.

The separate "late-arriving data after done blob" false positive — caused by the post-`done` `DataAvailabilitySync` step touching every blob's `updateTime` (Custom-Time + resource-id metadata) so `updateTime` lands after the `done` blob — is more involved and will be addressed in a **separate diff**.

Issue: #3983


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3974)

stevenwarejones (commented):
drop the late arriving as we do need a way to flag that. keep Gap detection counts data-without-done folders as gaps

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3974#-:-OutCr9VD9Y-qhGva9VW:bl6zedc)* status: 0 of 3 files reviewed, all discussions resolved.

georgi:
@claude review

src/main/kotlin/org/wfanet/measurement/edpaggregator/dataavailability/DataAvailabilityMonitor.kt
```diff
@@ -497,6 +497,14 @@ class DataAvailabilityMonitor(
     )
   }
 
+  /**
+   * All date folders that physically exist for the model line, whether or not they have a "done"
+   * blob. Used for gap detection so that a folder with data but no "done" blob (e.g. a date still
+   * being finalized) is not mistaken for a missing date.
+   */
+  private fun presentDates(dateInfo: DateInfo): List<LocalDate> =
+    (dateInfo.datesWithDoneBlob + dateInfo.datesWithoutDoneBlob).sorted()
```
stevenwarejones, line 523: This fix removes the protection that prevented `DataAvailabilitySync` from calling `replaceDataAvailabilityIntervals` while a middle date is still mid-finalization.

Pre-fix, a data-without-done date landed in `gapDates`, which `Sync.sync` checks at line 229 and uses to skip the Kingdom update under `errorIfGapsExist`. Post-fix, that same date lands only in `datesWithoutDoneBlob`, so `gapDates` is empty and Sync proceeds — telling the Kingdom "I have 3/10 and 3/12" while 3/11 is still being written. Partial availability updates are visible to downstream and have to be rewound when the unfinalized date completes.

The alert-clarity win is real and worth keeping. But the Sync-skip behavior is load-bearing and silently changes here.

Recommend widening the Sync skip condition to cover both conditions:

```kotlin
val modelLinesBlocked = gapResult.statuses.filter {
  !it.gapDates.isNullOrEmpty() || !it.datesWithoutDoneBlob.isNullOrEmpty()
}
```

with an updated log message and (probably) a renamed `errorIfGapsExist` config field — though that rename could land in a follow-up since it is a config-schema change.

Needs a regression test: "sync skips replaceDataAvailabilityIntervals when an in-range date has data but no done blob" — fixture is the same 3-day setup with the middle date missing `done`; assert `replaceDataAvailabilityIntervals` was not called.

stevenwarejones (commented):
Spot-check on gap detection semantics — see inline.

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3974)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on dawn-wang22 and stevenwarejones).

src/main/kotlin/org/wfanet/measurement/edpaggregator/dataavailability/DataAvailabilityMonitor.kt
stevenwarejones: Correction on the suggested fix, after reading the cmms-api docs ([data_provider.proto#L130](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/21cf2f45e7b115a31e38cd8b321c74597a25a204/src/main/proto/wfa/measurement/api/v2alpha/data_provider.proto#L130)): the contract is "one Interval per model line, continuous availability." So the right bounds for the in-range check are the same bounds being sent to the Kingdom — already computed in `Sync.sync` as `availabilityEntries` (from `computeModelLineBounds`). Anything inside those bounds (a gap, or a date present-but-unfinalized) breaks the continuity guarantee and should block. Anything outside (today still being written; a backfill earlier than the established range) is normal and shouldn't.

```kotlin
val modelLinesBlocked = gapResult.statuses.filter { status ->
  val bound = availabilityEntries.firstOrNull { it.key == status.modelLineKey.toName() }
  val startDate = bound?.value?.startTime?.toLocalDate(timeZone)
  val endDate = bound?.value?.endTime?.toLocalDate(timeZone)
  val inRangeUnfinalized =
    startDate != null && endDate != null &&
      status.datesWithoutDoneBlob.orEmpty().any { it in startDate..endDate }
  !status.gapDates.isNullOrEmpty() || inRangeUnfinalized
}
```

This ties the in-range check to the *actual* interval being sent, not an inference about which `done` dates exist.

The "any datesWithoutDoneBlob blocks" version I posted earlier is too broad — replace with the above.

Test coverage to add:
- **in-range unfinalized → blocked**: 3-day fixture, all 3 days have `done` and metadata in Spanner so bounds are `[day1, day3]`, then a 4th day in the middle (`day2.5`) has data but no `done` → assert `replaceDataAvailabilityIntervals` NOT called.
- **trailing-margin unfinalized → not blocked**: established range `[day1, day3]`, today (day4) has data but no `done` → assert `replaceDataAvailabilityIntervals` IS called.
- **pre-leading unfinalized → not blocked**: established range `[day3, day5]`, a backfill date (day1) has data but no `done` → assert `replaceDataAvailabilityIntervals` IS called.
- **gap inside bounds → blocked**: established range `[day1, day3]`, day2 missing entirely → existing behavior, assert still blocked.
- **healthy → updates**: all dates in range have `done` + data → assert `replaceDataAvailabilityIntervals` called once with the expected interval.

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3974)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on stevenwarejones).

src/main/kotlin/org/wfanet/measurement/edpaggregator/dataavailability/DataAvailabilityMonitor.kt
stevenwarejones: Folding in a related observation: the same in-range predicate should also drive `DataAvailabilityMonitorFunction.logStatusIssues` and `hasIssues`. Today's SEVERE alert:

```kotlin
if (!status.datesWithoutDoneBlob.isNullOrEmpty()) {
  logger.log(Level.SEVERE, "ALERT: ... has dates without done blob: ${status.datesWithoutDoneBlob}")
}
```

fires for *any* unfinalized date — including today's in-progress folder and old deleted-files folders. Both are normal and shouldn't page.

If `ModelLineStatus` exposes the in-range subset (computed from `datesWithoutDoneBlob` ∩ `[startDate, endDate]`), the same field drives:
- Sync's skip decision (above)
- Function's SEVERE alert + `hasIssues`

The raw `datesWithoutDoneBlob` stays on the status for INFO logging / metrics, so out-of-range cases are still observable, just not paging.

Margin-side coverage (today still being written) is handled separately by staleness — if today's folder doesn't get a `done` blob within `max_stale_days`, staleness fires.

Test coverage to add (in addition to the Sync-skip tests above):
- **in-range unfinalized → SEVERE log + hasIssues=true** in `DataAvailabilityMonitorFunctionTest` (or unit-level equivalent)
- **trailing-margin unfinalized → no SEVERE log + hasIssues=false**
- **pre-leading unfinalized → no SEVERE log + hasIssues=false**

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3974)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on stevenwarejones).

src/main/kotlin/org/wfanet/measurement/edpaggregator/dataavailability/DataAvailabilityMonitor.kt
stevenwarejones: Refinement to the previous comment — my SEVERE-only-when-in-range advice was wrong. The clearer rule is:

**A folder enumerable as `datesWithoutDoneBlob` always contains at least one non-`done` file** — in GCS a date prefix is only enumerable if some blob lives under it, and the `done` blob isn't there, so by elimination something else is. That's always a workflow bug (data uploaded, finalization failed) and should always alert, regardless of margin or middle.

Truly empty folders are invisible to GCS enumeration, so the "empty folder on the margin vs. middle" case doesn't physically exist in production.

So the actual rules are:

| Condition | Alert (SEVERE) | Block availability update |
|---|---|---|
| `datesWithoutDoneBlob` (anywhere) | yes | only if in `[startDate, endDate]` |
| `gapDates` (always interior by def) | yes | yes |

Which means:

- `DataAvailabilityMonitorFunction.logStatusIssues` / `hasIssues` on `datesWithoutDoneBlob` — **keep as-is**, fires on any unfinalized date.
- `Sync.sync` skip — needs the in-range check (gated by `[startDate, endDate]` from `computeModelLineBounds`).

Only the Sync-side change is needed. The Function's alert logic stays.

Revised test coverage:
- **in-range unfinalized → Sync skipped, SEVERE alert fires**
- **trailing-margin unfinalized → Sync proceeds, SEVERE alert fires**
- **pre-leading unfinalized → Sync proceeds, SEVERE alert fires**
- **gap inside bounds → Sync skipped, SEVERE alert fires**
- **healthy → Sync proceeds, no SEVERE**

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3974)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on stevenwarejones).

src/main/kotlin/org/wfanet/measurement/edpaggregator/dataavailability/DataAvailabilityMonitor.kt
stevenwarejones, line 523: nit: the disjointness of `datesWithDoneBlob` and `datesWithoutDoneBlob` is load-bearing for the new `presentDates` (and for the existing metric-count math), but only implicitly enforced by the producer construction in `getDateInfo`. Worth a `check` before constructing `DateInfo` to fail loudly if a future refactor ever breaks it:

```kotlin
check(datesWithDone.intersect(datesWithoutDoneBlobList.toSet()).isEmpty()) {
  "datesWithDoneBlob and datesWithoutDoneBlob must be disjoint"
}
```

stevenwarejones (commented):
One additional nit.

src/main/kotlin/org/wfanet/measurement/edpaggregator/dataavailability/DataAvailabilityMonitor.kt
```diff
@@ -497,6 +497,14 @@ class DataAvailabilityMonitor(
     )
   }
 
+  /**
+   * All date folders that physically exist for the model line, whether or not they have a "done"
+   * blob. Used for gap detection so that a folder with data but no "done" blob (e.g. a date still
+   * being finalized) is not mistaken for a missing date.
```
stevenwarejones: nit: "physically exist" reads as if the function inspects storage. It just unions two precomputed sets. Suggest:

```
All date folders enumerated for the model line, whether or not they have a "done" blob. Used for gap detection so that a folder with data but no "done" blob (e.g. a date still being finalized) is not mistaken for a missing date.
```

stevenwarejones (commented):
One more nit.

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3974)* status: all files reviewed, 2 unresolved discussions (waiting on dawn-wang22).

dawn-wang22:
Thanks for the review, Steven. Addressed in the latest push:

**Block in-range unfinalized dates (not just gaps).** `DataAvailabilitySync.sync` now skips `replaceDataAvailabilityIntervals` when a model line has either a true gap **or** an unfinalized date (a folder with data but no `done` blob) that falls *inside* the finalized range. The range is `[earliestDate, latestDate]` computed entirely from the finalized (`done`-bearing) dates, so an unfinalized date that only *trails* `latestDate` or *leads* `earliestDate` is treated as window extension and does **not** block publishing. This is done with `LocalDate` comparisons only — no timezone/instant math.

**`earliestDate` on `ModelLineStatus`.** Added `earliestDate: LocalDate?` (populated in both `buildFullStatus` and `buildGapStatus`, `null` when there are no uploaded dates) so the Sync gate has the lower bound it needs. KDoc updated.

**Disjointness invariant.** `getDateInfo` now `check(...)`s that `datesWithDoneBlob` and `datesWithoutDoneBlob` never overlap — the in-range gate relies on a date being either finalized or unfinalized, never both.

**`presentDates` KDoc** reworded to make explicit that gap detection runs over the union of finalized + unfinalized folders.

**Regression tests** (`DataAvailabilitySyncTest`), all green alongside `DataAvailabilityMonitorTest`:
- in-range unfinalized date → **skips** publish
- unfinalized date trailing the finalized range → **publishes**
- unfinalized date leading the finalized range → **publishes**
- true gap between finalized dates → **skips** publish
- all dates finalized and contiguous → **publishes**

src/main/kotlin/org/wfanet/measurement/edpaggregator/dataavailability/DataAvailabilityMonitor.kt
stevenwarejones: Verified — the in-range gating using `earliestDate`/`latestDate` from `ModelLineStatus` is actually better than my suggested `availabilityEntries`-based check. It's more conservative under the realistic timing window where a `done` blob exists but Sync hasn't yet persisted that date's metadata to Spanner: the local check blocks publishing in that window, while the `computeModelLineBounds`-based check would let it through. Erring on the side of not publishing partial availability is the right direction here. Plus simpler code — no proto-to-LocalDate conversion at the call site.

All 5 test cases covered cleanly (in-range, trailing, leading, gap, healthy). LGTM.

stevenwarejones: Verified.

stevenwarejones (commented):
> @stevenwarejones+AGNT made 2 comments and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3974)* status: 2 of 5 files reviewed, 2 unresolved discussions (waiting on dawn-wang22 and stevenwarejones).

stevenwarejones (approved):
> @stevenwarejones reviewed 3 files and all commit messages, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3974)* status: all files reviewed, 1 unresolved discussion (waiting on dawn-wang22).

src/test/kotlin/org/wfanet/measurement/edpaggregator/dataavailability/DataAvailabilitySyncTest.kt
```diff
@@ -1091,6 +1091,137 @@ class DataAvailabilitySyncTest {
...
+  fun `sync calls replaceDataAvailabilityIntervals when an unfinalized date trails the range`():
+    Unit = runBlocking {
+    val storageClient =
+      FakeBlobMetadataStorageClient(FileSystemStorageClient(File(tempFolder.root.toString())))
+    val edpPath = "edp/edpa_edp"
+
+    // Finalized range is [03-13, 03-15] (03-15 from seedBlobDetails). 03-16 has data but no "done"
+    // blob, trailing *after* the finalized range, so it must not block publishing.
+    writeModelLineDate(storageClient, edpPath, "2026-03-13", finalized = true)
+    writeModelLineDate(storageClient, edpPath, "2026-03-14", finalized = true)
```
stevenwarejones, line 1167: nit: the leading-edge (and trailing-edge) tests assert the publish proceeds, but do not assert that the leading-edge unfinalized date is still surfaced to operators. The contract we landed on is "do not block, but still alert," and both halves should be visible in one test so the intent is unmistakable.

The `STATUS_WITHOUT_DONE_BLOB` metric path is unconditional in `recordMetrics`, so it is covered in principle by the existing combined-fixture test, but a leading-edge-specific assertion would prove the case directly:

```kotlin
val metrics = collectMetrics()  // wire InMemoryMetricExporter as in DataAvailabilityMonitorTest
assertThat(getDateStatusCount(metrics, DataAvailabilityMonitorMetrics.STATUS_WITHOUT_DONE_BLOB))
  .isEqualTo(1)
```

Same addition would also be valuable on the trailing-edge test for symmetry.

stevenwarejones (commented):
One more on the new tests.

stevenwarejones (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3974)* status: all files reviewed, 2 unresolved discussions (waiting on dawn-wang22).

dawn-wang22:
Thanks Steven — both nits addressed in `548e1fc17`:

1. **`presentDates` KDoc** — reworded to drop "physically exist" (it just unions two precomputed sets, doesn't inspect storage); used your suggested phrasing.

2. **Leading/trailing-edge tests** — both now assert the "don't block, but still surface it" contract explicitly: after confirming `replaceDataAvailabilityIntervals` IS called, they wire `createMetricsEnvironment()` and assert `STATUS_WITHOUT_DONE_BLOB == 1`, proving the out-of-range unfinalized date still alerts even though publishing proceeds.

(The earlier in-range gating + disjointness `check` you already marked Verified are unchanged.) Both test targets green.

src/main/kotlin/org/wfanet/measurement/edpaggregator/dataavailability/DataAvailabilityMonitor.kt
stevenwarejones: Verified — KDoc now reads cleanly and accurately describes what the function does (sorted union of two precomputed sets) without implying any storage inspection. LGTM.

src/test/kotlin/org/wfanet/measurement/edpaggregator/dataavailability/DataAvailabilitySyncTest.kt
stevenwarejones: Verified — both trailing-edge and leading-edge tests now assert `STATUS_WITHOUT_DONE_BLOB == 1` alongside the existing `replaceDataAvailabilityIntervals` check, making the "don't block, but still alert" contract explicit in one test each. Reused the existing `createMetricsEnvironment` / `getDateStatusCount` helpers — clean. LGTM.

stevenwarejones (commented):
> @stevenwarejones+AGNT made 2 comments and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3974)* status: 3 of 5 files reviewed, all discussions resolved (waiting on stevenwarejones and ugur47).

stevenwarejones (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3974)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on ugur47).

ugur47 (approved):
> @ugur47 reviewed 5 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3974)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dawn-wang22).

PR #3929, opened 2026-06-05
feat: Add PoolAssignmentJob proto definitions for internal and public API

getina:
## Summary
- Add proto message and service definitions for PoolAssignmentJob, a Phase-0 pool assignment job resource for the VID labeling pipeline.
- Includes create, batch create, get, list, MarkSucceeded (custom response with is_last_shard and subpool_ids), and MarkFailed methods for both internal and public (v1alpha) APIs.
- State enum: CREATED, SUCCEEDED, FAILED (not terminal — operator can retry).
- cmms_model_line as REQUIRED+IMMUTABLE field with resource_reference to halo.wfanet.org/ModelLine.
- etag for optimistic concurrency control on Mark requests.
- AIP-159 wildcard support on List parent with cmms_model_line filter for cross-upload queries.

## Test plan
- bazel build passes for all proto targets (internal state, internal message, internal service, public message, public service)

Issue: #3883

Issue #3958, opened 2026-06-10
Add VidLabelingMonitorFunction for pipeline health monitoring and dispatch sequencing


stevenwarejones:
## Summary

Add a VidLabelingMonitorFunction (periodic Cloud Scheduler-triggered function) that monitors the health of the VID labeling pipeline and drives dispatch sequencing. This parallels DataAvailabilityMonitorFunction which monitors the data availability pipeline.

## Motivation

The VidLabelingDispatcher (PR #3910) currently creates RawImpressionUpload + RawImpressionUploadFile entries AND dispatches WorkItems in a single invocation. This creates several problems:

1. **Cross-dispatch concurrency:** Two uploads for the same (DataProvider, ModelLine) can run Phase 1 concurrently, corrupting the cumulative rank index blob (both rankers read the same prior blob, allocate overlapping ranks, last writer wins silently).
2. **Re-upload to same path:** If an EDP uploads additional files to an existing date folder and writes a new done blob, the dispatcher crashes with ALREADY_EXISTS on WorkItem creation because the requestId is deterministic from the done blob path (same path → same requestId → same uploadId → same WorkItem IDs).
3. **Stuck pipeline detection:** If a phase transition fails (all PoolAssignmentJobs SUCCEEDED but RawImpressionUploadModelLine still in POOL_ASSIGNING), nothing detects or recovers the stall.

## Proposed Solution

Split the dispatcher into two responsibilities:
- **VidLabelingDispatcher** (event-driven, triggered by done blob): Only registers the upload — creates RawImpressionUpload + RawImpressionUploadFile entries. Does NOT create WorkItems.
- **VidLabelingMonitorFunction** (periodic, triggered by Cloud Scheduler): Scans pipeline state and drives work forward when conditions are met.

## VidLabelingMonitorFunction Checks

Modeled after DataAvailabilityMonitor, the monitor should check for:

### Upload health
- **Uploads without WorkItems:** RawImpressionUpload exists with files registered but no RawImpressionUploadModelLine or PoolAssignmentJob children. If no other upload for the same (DataProvider, ModelLine) is in a non-terminal state → resolve model lines, create WorkItems, dispatch.
- **Queued uploads:** Uploads waiting behind an in-progress upload for the same (DataProvider, ModelLine). When the prior upload completes → dispatch the queued one. Ensures sequential processing.

### Phase transition health
[...]

PR #3985, opened 2026-06-12
feat: Phase-0 SubpoolAssigner — pool-emit labeling to encrypted per-subpool fingerprint maps

Marco-Premier:
## Summary

  Implements the **Phase-0 SubpoolAssigner** TEE application of the memoized VID-assignment
  pipeline: per fingerprint shard, it reads an EDP's raw impressions, runs the VID model in
  **pool-emit** mode to discover each fingerprint's subpool, and writes encrypted, RecordIO
  `SubpoolFingerprints` maps to storage. The last shard out merges every shard's per-subpool
  slices into one map per subpool.

  Layering mirrors `ResultsFulfiller`:
  - **`SubpoolAssignerAppRunner`** — process bootstrap / CLI / shared singletons (KMS clients,
    gRPC channels + stubs, Pub/Sub).
  - **`SubpoolAssignerApp`** — thin `BaseTeeApplication` adapter: unpacks the `WorkItem`, resolves
    WorkItem-scoped dependencies, delegates.
  - **`SubpoolAssigner`** — the unit-testable per-WorkItem logic (the `ResultsFulfiller` analog).

  > Stacked on #3954 (`pool-assigner-utility-classes`) — **review that first**. Builds on the
  > `PoolAssignmentJob` / `RankerJob` / `RawImpressionUploadModelLine` APIs from #3929.

  ## Per-shard flow

  1. `RawImpressionSource.streamBlobs` → one shared `SubpoolAssignmentSink`: project each parquet row
     to a `LabelerInput` (`LabelerInputMapper`), drop rows outside the model line's active window,
     pool-emit label, and add the 12-byte `EventIdDigest` to the per-subpool set.
  2. Generate a per-shard DEK; stream each subpool's fingerprints to its own encrypted RecordIO blob,
     freeing each in-memory map as soon as its blob is durable.
  3. `MarkPoolAssignmentJobSucceeded(name, etag, encrypted_dek)` — persists the DEK on the job and
     atomically learns whether this is the last shard out (returns the discovered pool offsets).
  4. **Last-shard-out**: recover every shard's DEK via `ListPoolAssignmentJobs`, stream-concat each
     subpool's per-shard slices into one merged blob (disjoint by shard ⇒ no dedup), delete the temp
     per-shard blobs.
  5. On failure: best-effort `MarkPoolAssignmentJobFailed`, then rethrow so the TEE framework nacks.

  ## Key components

  | Component | Role |
  |---|---|
  | `LabelerInputMapper` (`rawimpressions`) | Reflective parquet-row → `LabelerInput` projection via `labeler_input_field_mapping`. Shared with Phase-2. |
  | `PoolEmitLabeler` | Interface for the pool-emit model call. Real C++/JNI impl is a follow-up. |
  | `SubpoolAssignmentSink` | `RawImpressionSource.BlobSink`: project → window-filter → label → accumulate. |
  | `SubpoolFingerprintsAccumulator` | Thread-safe `subpool → Bytes12IntMap`-as-set; streams packed chunks and frees maps incrementally (no whole-subpool materialization). |
  | `SubpoolFingerprintsStore` (`rawimpressions`) | RecordIO read / write / streaming merge / delete, envelope-encrypted. Shared with Phase-2. |
  | `SubpoolAssignerMetrics` | OpenTelemetry counters for labeling outcomes. |

  ## Why RecordIO

  A single protobuf message / `ByteString` is `Int`-bounded (~2 GB ⇒ ~178M fingerprints). A merged,
  un-sharded subpool can exceed that, so blobs are **streams of `SubpoolFingerprints` records**;
  reads, writes, and the merge are all streaming, so memory stays bounded regardless of subpool size.

  ## Proto changes (contract only)

  Adds `encrypted_dek` to `PoolAssignmentJob` (public + internal):
  - on the **resource** as `OUTPUT_ONLY` (server-populated, read back via `Get`/`List` — consistent
    with how `state` is set by the `Mark*` custom methods),
  - on **`MarkPoolAssignmentJobSucceededRequest`** as the input the SubpoolAssigner supplies, so the
    DEK is persisted atomically with the `SUCCEEDED` transition.

  The service implementation that persists/returns this field is a **separate PR**; the app code is
  correct against the contract and goes fully live once that lands.

  ## Out of scope / follow-ups (TODOs in code)

  - **Control-plane fan-out** (`runLastShardOut`): RankerJob bin-packing/creation, the
    `RawImpressionUploadModelLine` → `RANKING` flip, and publishing one VidRankBuilder WorkItem per
    RankerJob — needs bin-packing policy + the dispatcher's WorkItem-publish mechanism.
  - **`PoolEmitLabeler`** production impl (load the compiled model from `model_blob_path`, C++/JNI).
  - **`SubpoolAssignerAppRunner`** production providers (storage clients, KEK URI, labeler loader),
    and the `RawImpressionUploadFileService` stub.
  - Deeper redelivery idempotency beyond the per-shard `SUCCEEDED` short-circuit.

  ## Testing

  Unit tests (all green): `LabelerInputMapperTest`, `SubpoolFingerprintsAccumulatorTest`
  (dedup / fan-out / big-endian packing / chunking / 8-thread concurrency), `SubpoolAssignmentSinkTest`
  (routing / window-drop / unrouted / fan-out / OTel counters), `SubpoolFingerprintsStoreTest`
  (RecordIO round-trip / streaming merge / delete).

Issue: #3913


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3985)

georgi:
@claude review

src/main/kotlin/org/wfanet/measurement/edpaggregator/subpoolassigner/SubpoolAssigner.kt
```diff
@@ -0,0 +1,257 @@
...
+    logger.info(
+        "Shard $shardIndex/$totalShards for $modelLine: wrote $subpoolCount subpools " +
+            "(labeled=${sink.labeled}, droppedOutsideWindow=${sink.droppedOutsideWindow}, " +
+            "unrouted=${sink.unrouted})")
+
+    // 3. Mark this shard SUCCEEDED (idempotent on redelivery).
+    val job =
+        poolAssignmentJobsStub.getPoolAssignmentJob(
+            getPoolAssignmentJobRequest { name = poolAssignmentJob })
+    if (job.state == PoolAssignmentJob.State.SUCCEEDED) {
```
stevenwarejones: This early-return silently drops the last-shard-out merge on Pub/Sub redelivery. If the last VM marks SUCCEEDED, starts the merge, then crashes (ack-deadline expiry, OOM, transient network failure) and Pub/Sub redelivers the WorkItem, this branch returns lastShardOut=false, acks the message, and the merged subpool blobs never exist. Phase 1 reads /merged/subpool-N and finds nothing — silent stall, no error signal.

The TODO acknowledges the root cause: LastShardResult.pool_offsets is returned only by MarkPoolAssignmentJobSucceeded, not by Get. The design (§Stuck phase transition) routes recovery through VidLabelingMonitorFunction detecting "all PoolAssignmentJobs SUCCEEDED but parent still in POOL_ASSIGNING," but that path requires a redelivered WorkItem to actually re-run the merge — which today it can't, because pool_offsets is gone.

Tracked by #3991 — persist pool_offsets on RawImpressionUploadModelLine so this branch can recover the merge input list via Get instead of skipping. Fix shape:

```kotlin
if (job.state == PoolAssignmentJob.State.SUCCEEDED) {
  val parent = rawImpressionUploadModelLinesStub.getRawImpressionUploadModelLine(...)
  val poolOffsets = parent.poolOffsetsList
  if (poolOffsets.isNotEmpty()) {
    runLastShardOut(poolOffsets, mergedDek = dek)
    return summary(subpoolCount, sink, lastShardOut = true)
  }
  return summary(subpoolCount, sink, lastShardOut = false)
}
```

This is blocking — without it, any crash during the merge silently stalls the pipeline.

stevenwarejones (commented):
Last-shard-out recovery: persist pool_offsets so redelivery can re-run the merge.

stevenwarejones (commented):
nit: New Kotlin files use 4-space indentation; repo standard is 2-space (ktfmt --google-style). Run `ktfmt --google-style` on the new files before next push.

src/main/kotlin/org/wfanet/measurement/edpaggregator/subpoolassigner/SubpoolAssigner.kt
```diff
@@ -0,0 +1,257 @@
...
+      if (job.state == PoolAssignmentJob.State.CREATED) {
+        poolAssignmentJobsStub.markPoolAssignmentJobFailed(
+            markPoolAssignmentJobFailedRequest {
+              name = poolAssignmentJob
+              etag = job.etag
+              errorMessage = (cause.message ?: cause::class.java.simpleName).take(MAX_ERROR_MESSAGE)
+            })
+      }
+    } catch (e: Exception) {
+      logger.warning("Failed to mark PoolAssignmentJob $poolAssignmentJob FAILED: $e")
```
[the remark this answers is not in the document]
stevenwarejones: +1 on the bot — this is a real bug. The stack-trace-dropping log line is the only diagnostic surviving a best-effort markFailed failure, and we'll need it. Apply the fix as suggested:

```kotlin
logger.log(Level.WARNING, "Failed to mark PoolAssignmentJob $poolAssignmentJob FAILED", e)
```

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3985)* status: 0 of 23 files reviewed, 2 unresolved discussions (waiting on Marco-Premier and stevenwarejones).

src/main/kotlin/org/wfanet/measurement/edpaggregator/rawimpressions/LabelerInputMapper.kt
```diff
@@ -0,0 +1,198 @@
...
+ *
+ * Stateless and therefore safe to share across threads. Field paths are resolved to descriptors
+ * once at construction, so per-row mapping does no string-keyed descriptor lookups.
+ *
+ * Note: `SubpoolAssignerParams.event_template_field_mapping` is intentionally NOT handled here.
+ * Those fields target the model's event-template message (the EDP's event proto), whose descriptor
+ * is only known once the compiled model is loaded by the labeler; event-template projection is
+ * therefore the labeler-loading layer's responsibility.
+ */
+class LabelerInputMapper(labelerInputFieldMapping: Map<String, String>) {
```
stevenwarejones, line 56: @Marco-Premier two follow-up issues on the mapper. Please add TODOs in this file linking both:

#3992 — Add enum lookup (gender) and age-range mapping. This one we should do — gender values that don't match the proto enum names and age as a single int / string bucket are routine in real EDP data, and the alternative (every EDP normalizes upstream) produces inconsistent shapes across EDPs. Sketch of the proto shape and complexity estimate (~280 LOC including tests) in the issue.

#3993 — Identity composition, required-field validation, schema-drift detection. Lower priority but worth a TODO so they don't get forgotten. The required-field check in particular is one missing-event_id away from a catastrophic-collision bug; worth fixing eventually even if not in this PR.

Add `TODO(world-federation-of-advertisers/cross-media-measurement#3992)` and `TODO(world-federation-of-advertisers/cross-media-measurement#3993)` above the class declaration so the next person editing this file sees both.

stevenwarejones (commented):
Two follow-up mapping issues — please add TODOs linking both.

src/main/kotlin/org/wfanet/measurement/edpaggregator/rawimpressions/LabelerInputMapper.kt
Marco-Premier: Done.

src/main/kotlin/org/wfanet/measurement/edpaggregator/subpoolassigner/SubpoolAssigner.kt
Marco-Premier: Done.

src/main/kotlin/org/wfanet/measurement/edpaggregator/subpoolassigner/SubpoolAssigner.kt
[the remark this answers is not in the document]
Marco-Premier: Done.

Marco-Premier (commented):
> @Marco-Premier made 3 comments.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3985)* status: 0 of 26 files reviewed, 3 unresolved discussions (waiting on stevenwarejones and stevenwarejones+AGNT).

src/main/kotlin/org/wfanet/measurement/edpaggregator/rawimpressions/LabelerInputMapper.kt
stevenwarejones: Verified — both TODOs landed at LabelerInputMapper.kt:50-55 with the right shape (#3992 framed as "do this", #3993 noted as a follow-up that depends on #3992's reshaped config).

src/main/kotlin/org/wfanet/measurement/edpaggregator/subpoolassigner/SubpoolAssigner.kt
stevenwarejones: Verified — and the implementation is meaningfully better than the simple persist-on-parent fix I proposed in #3991. Three recovery states (not-last-shard-out / already-complete / interrupted) cleanly handled; uses the persisted `encrypted_merged_dek` to keep the re-merge consistent with the already-published `VidRankBuilderParams.encrypted_subpool_maps_dek` instead of diverging on a freshly generated DEK. The proto additions on `RawImpressionUploadModelLine` (`pool_offsets`, `max_event_date`, `encrypted_merged_dek`) match exactly what the recovery path needs. TODO referencing #3999 at the GCS-write site is a nice touch.

src/main/kotlin/org/wfanet/measurement/edpaggregator/subpoolassigner/SubpoolAssigner.kt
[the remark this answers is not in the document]
stevenwarejones: Not applied — `SubpoolAssigner.kt:429` is still `logger.warning("Failed to mark PoolAssignmentJob $poolAssignmentJob FAILED: $e")`. The `$e` interpolation drops the stack trace via `Throwable.toString()`. Per the bot's original suggestion:

```kotlin
logger.log(Level.WARNING, "Failed to mark PoolAssignmentJob $poolAssignmentJob FAILED", e)
```

(Needs `import java.util.logging.Level`.)

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3985)* status: 0 of 26 files reviewed, 3 unresolved discussions (waiting on stevenwarejones).

src/main/kotlin/org/wfanet/measurement/edpaggregator/subpoolassigner/SubpoolAssigner.kt
```diff
@@ -0,0 +1,556 @@
...
+        (0 until totalShards).map { shard ->
+          SubpoolFingerprintsStore.SubpoolBlob(
+            shardSubpoolKey(shard, poolOffset),
+            requireNotNull(shardDeks[shard]) { "Missing DEK for shard $shard; cannot merge" },
+          )
+        }
+      // TODO(world-federation-of-advertisers/cross-media-measurement#3999): once mergeSubpool
+      //   enforces ifGenerationMatch=0, treat a 412 here as "already merged" and continue (this
+      //   recovery path re-runs the merge idempotently), rather than as a lost race.
+      store.mergeSubpool(inputs, mergedSubpoolKey(poolOffset), mergedDek)
```
stevenwarejones, line 284: Verified — recovery is correct *contingent on #3999 landing*. Without `ifGenerationMatch=0` on the merged-blob write, a late-redelivered non-last-out shard's `recoverIfLastShardOut` can race with the real last-out's `runLastShardOut` here (both write to `merged/subpool-N`, last writer wins on bytes). Same DEK on both sides so the merged plaintext is identical, but there's a secondary race: the first finisher deletes per-shard blobs while the other is still mid-merge → truncated output. The TODO is the right pointer; the fix lands when #3999 does.

stevenwarejones (commented):
Verification notes on three retry/redelivery paths I traced through carefully.

src/main/kotlin/org/wfanet/measurement/edpaggregator/subpoolassigner/SubpoolAssigner.kt
```diff
@@ -0,0 +1,556 @@
...
+      }
+      logger.info(
+        "markRawImpressionUploadModelLineRanking(${parent.name}) already advanced " +
+          "(${e.status.code}); treating as done"
+      )
+    }
+  }
+
+  /** Best-effort transition of this job to FAILED so operators see the failure; never throws. */
+  private suspend fun markFailedBestEffort(cause: Throwable) {
```
stevenwarejones, line 427: Verified — the retry-vs-DLQ split works as intended. Transient failures (GCS 503 during merge, UNAVAILABLE on fan-out RPCs) throw past the state==SUCCEEDED check (which intentionally does nothing in that case), nack the message, redeliver → `recoverIfLastShardOut` succeeds on retry. Persistent failures eventually hit Pub/Sub's retry exhaustion → DLQ → the Monitor (#3958) picks up "all jobs SUCCEEDED but parent stuck in POOL_ASSIGNING." The SUCCEEDED branch deliberately doesn't mark the job FAILED because the per-shard work *did* complete — fan-out failure is a parent-level concern, not a job-level one.

src/main/kotlin/org/wfanet/measurement/edpaggregator/subpoolassigner/SubpoolAssigner.kt
```diff
@@ -0,0 +1,556 @@
...
+  }
+
+  /**
+   * Returns the parent `RawImpressionUploadModelLine` for this (`RawImpressionUpload`,
+   * `ModelLine`), located via `ListRawImpressionUploadModelLines` filtered by model line, or `null`
+   * if absent. Carries the resource name (for the state flip), the `pool_offsets` union, the
+   * reduced `max_event_date`, and the current pipeline state — all populated by the
+   * last-shard-out's `MarkPoolAssignmentJobSucceeded`.
+   */
+  private suspend fun getParent(): RawImpressionUploadModelLine? {
```
stevenwarejones, line 477: Verified — the List-with-filter approach is correct given the dispatcher doesn't pass a `raw_impression_upload_model_line_id` on `CreateRawImpressionUploadModelLine` (server assigns the resource ID). At realistic scale (handful of rows per upload, server-side `cmms_model_line` filter, ~2 calls per WorkItem) the cost is negligible — single round trip, single indexed row return. AIP-133 client-assigned IDs would let this become a direct `Get` by name, but not worth the proto churn here.

stevenwarejones (commented):
> @stevenwarejones reviewed 26 files and all commit messages, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3985)* status: all files reviewed, 5 unresolved discussions (waiting on Marco-Premier).

stevenwarejones (changes requested):
> @stevenwarejones resolved 4 discussions.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3985)* status: all files reviewed, 1 unresolved discussion (waiting on stevenwarejones).

src/main/kotlin/org/wfanet/measurement/edpaggregator/subpoolassigner/SubpoolAssigner.kt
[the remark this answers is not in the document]
Marco-Premier: Done.

Marco-Premier (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3985)* status: 25 of 26 files reviewed, 1 unresolved discussion (waiting on stevenwarejones and stevenwarejones+AGNT).

src/main/kotlin/org/wfanet/measurement/edpaggregator/subpoolassigner/SubpoolAssigner.kt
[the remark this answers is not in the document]
stevenwarejones: Verified — `SubpoolAssigner.kt:430` is now `logger.log(Level.WARNING, "Failed to mark PoolAssignmentJob $poolAssignmentJob FAILED", e)`, applied in commit ab83a0d67. Stack trace is preserved.

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3985)* status: 25 of 26 files reviewed, 1 unresolved discussion (waiting on stevenwarejones).

stevenwarejones (approved):
> @stevenwarejones reviewed 9 files and all commit messages, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3985)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on Marco-Premier).

src/main/kotlin/org/wfanet/measurement/edpaggregator/subpoolassigner/SubpoolAssigner.kt
```diff
@@ -0,0 +1,557 @@
...
+      "Shard $shardIndex/$totalShards for $modelLine: wrote $subpoolCount subpools " +
+        "(labeled=${sink.labeled}, droppedOutsideWindow=${sink.droppedOutsideWindow}, " +
+        "unrouted=${sink.unrouted})"
+    )
+
+    // 3. Mark this shard SUCCEEDED, reporting its discovered subpool offsets and latest event date
+    // so the service can union the offsets and reduce the dates to a max across shards, persisting
+    // both on the parent RawImpressionUploadModelLine.
+    val markResponse =
+      poolAssignmentJobsStub.markPoolAssignmentJobSucceeded(
```
stevenwarejones, line 183: Populate `request_id` here. The field is on `MarkPoolAssignmentJobSucceededRequest` as of #3997 (already merged) and is the AIP-155 retry-idempotency key — without it, a Pub/Sub redelivery hits the etag-mismatch path on the second attempt and the server cannot return the cached `LastShardResult` payload. Use the same deterministic-UUID pattern `VidRankBuilder.markSucceededRequestId()` uses, e.g. `requestId = deterministicUuid("$poolAssignmentJob|succeeded")`.

src/main/kotlin/org/wfanet/measurement/edpaggregator/subpoolassigner/SubpoolAssigner.kt
```diff
@@ -0,0 +1,557 @@
...
+
+  /**
+   * Flips the parent `RawImpressionUploadModelLine` `POOL_ASSIGNING` -> `RANKING`. Only attempted
+   * when the parent is still `POOL_ASSIGNING`; any failure is swallowed so a redelivered
+   * last-shard-out (where another runner may have already advanced the state) is a no-op.
+   */
+  private suspend fun markParentRanking(parent: RawImpressionUploadModelLine) {
+    if (parent.state != RawImpressionUploadModelLine.State.POOL_ASSIGNING) return
+    try {
+      rawImpressionUploadModelLinesStub.markRawImpressionUploadModelLineRanking(
```
stevenwarejones, line 399: Add a TODO to pass `etag = parent.etag` once #4005 lands. PR #4005 adds the `etag` field on `MarkRawImpressionUploadModelLine{PoolAssigning,Ranking,Labeling,Completed,Failed}Request` for AIP-154 optimistic locking. The state-machine guard (POOL_ASSIGNING → RANKING fires once) is sufficient for correctness today, but the etag-CAS will tighten the contract and catch a stale-parent-read bug class the state guard cannot.

src/main/kotlin/org/wfanet/measurement/edpaggregator/subpoolassigner/SubpoolAssigner.kt
```diff
@@ -0,0 +1,557 @@
...
+      return recoverIfLastShardOut()
+    }
+
+    // 1. Read -> label -> accumulate.
+    val sink = SubpoolAssignmentSink(mapper, labeler, accumulator, activeWindow, metrics)
+    rawImpressionSource.streamBlobs(openSink = { sink })
+
+    // 2. Generate this shard's DEK and stream each subpool to its own RecordIO blob, freeing the
+    // subpool's in-memory map as soon as its blob is durable.
+    // TODO(world-federation-of-advertisers/cross-media-measurement#3999): once
```
stevenwarejones, line 154: FYI — wrote up the missing primitive in common-jvm: world-federation-of-advertisers/common-jvm#389 adds `writeBlobIfAbsent` to `ConditionalOperationStorageClient`, forwarded through `AeadStorageClient` / `StreamingAeadStorageClient` / `MesosRecordIoStorageClient` so it's available end-to-end through your encryption + RecordIO stack.

Once that PR lands, the workaround here (per-attempt UUID in blob keys + AIP-155 row insert as the linearization point + lifecycle-policy-dependent orphan cleanup) can collapse to deterministic blob keys + a single `writeBlobIfAbsent` call. The lost-race path becomes a clean `BlobChangedException` catch where you ack the Pub/Sub message instead of writing.

No need to adopt in this PR — your current design is correct, just dependent on bucket lifecycle. Track as a follow-up so the cleanup happens once common-jvm#389 merges.

As a side bonus, common-jvm#389 also fixes a pre-existing race in `GcsStorageClient.writeBlob` (the prior `storage.create()` + `Blob.write(generationMatch)` split could fail under concurrent writers despite being documented as unconditional).

src/main/kotlin/org/wfanet/measurement/edpaggregator/rawimpressions/SubpoolFingerprintsStore.kt
```diff
@@ -0,0 +1,210 @@
...
+   *   `ifGenerationMatch=0` (write-if-absent) precondition so a `WorkItem` delivered to two VMs
+   *   (Pub/Sub at-least-once) can't last-writer-wins clobber each other's ciphertext — the etag CAS
+   *   on `MarkPoolAssignmentJobSucceeded` only guards the row, after the bytes are already on GCS.
+   *   This needs a write-if-absent primitive on the underlying stack (`GcsStorageClient`
+   *   `BlobWriteOption.doesNotExist()`, forwarded through the AEAD/KMS and
+   *   `MesosRecordIoStorageClient` wrappers) in common-jvm; once it exists, surface the `412` so
+   *   [SubpoolAssigner] can take the lost-race path. Not yet available — written unconditionally
+   *   for now.
+   */
+  suspend fun writeBlob(
```
stevenwarejones, line 97: Same adoption path applies here as on #4009: world-federation-of-advertisers/common-jvm#389 adds `writeBlobIfGeneration(blobKey, expectedGeneration, content)` end-to-end through your `MesosRecordIo(Aead(Gcs))` stack. Once it lands, the per-attempt-UUID model can collapse to deterministic blob keys with a read-gen-at-start + write-with-precondition pattern that handles fresh runs, crash-after-write-before-row recovery, and concurrent redelivery in one shape.

Full walkthrough of the pattern (with the critical "don't retry on 412" discipline and the blob-first/row-second ordering note) is on the corresponding comment in #4009 — see https://github.com/world-federation-of-advertisers/cross-media-measurement/pull/4009#discussion_r3423927318.

Not blocking this PR; track as a follow-up after common-jvm#389 lands.

Marco-Premier: Done.

src/main/kotlin/org/wfanet/measurement/edpaggregator/subpoolassigner/SubpoolAssigner.kt
Marco-Premier: Done.

src/main/kotlin/org/wfanet/measurement/edpaggregator/subpoolassigner/SubpoolAssigner.kt
Marco-Premier: #4005 was merged already, therefore I rebased the code and added the parent etag already

Marco-Premier (commented):
> @Marco-Premier made 3 comments.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3985)* status: 24 of 26 files reviewed, 3 unresolved discussions (waiting on kungfucraig and stevenwarejones).

src/main/kotlin/org/wfanet/measurement/edpaggregator/rawimpressions/SubpoolFingerprintsStore.kt
stevenwarejones: Verified — TODO at SubpoolFingerprintsStore.kt:88-94 in commit 9f7de6e12 now cross-links world-federation-of-advertisers/common-jvm#389 and spells out the read-gen-at-start + write-with-precondition + don't-retry-on-412 adoption pattern. Same cross-link in SubpoolAssigner.kt:155. (The other two #3999 sites — `mergeSubpool` and the consumer in last-shard-out — still reference only #3999, which is fine: the issue body links to #389.)

src/main/kotlin/org/wfanet/measurement/edpaggregator/subpoolassigner/SubpoolAssigner.kt
stevenwarejones: Verified — `requestId = deterministicUuid("$poolAssignmentJob|succeeded")` at SubpoolAssigner.kt:189 in commit 9f7de6e12. Helper extracted from `rankerJobRequestId` and shared, doc captures the AIP-155 reasoning.

src/main/kotlin/org/wfanet/measurement/edpaggregator/subpoolassigner/SubpoolAssigner.kt
stevenwarejones: Verified — `markRawImpressionUploadModelLineRanking` now passes `etag = parent.etag` at SubpoolAssigner.kt:399 in commit 9f7de6e12, and the `ABORTED` swallow's doc captures the etag-mismatch path. Thanks for picking it up immediately after #4005 merged.

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3985)* status: 24 of 26 files reviewed, 3 unresolved discussions (waiting on kungfucraig and stevenwarejones).

stevenwarejones (approved):
LGTM again. Verified the three follow-ups from yesterday's round are addressed in commit 9f7de6e12:
- `request_id` populated on `MarkPoolAssignmentJobSucceeded` via the new shared `deterministicUuid` helper
- `etag` passed on `MarkRawImpressionUploadModelLineRanking` (now that #4005 is merged)
- TODOs cross-link `common-jvm#389` for the eventual write-if-absent adoption

Re-approving.

stevenwarejones (approved):
> @stevenwarejones reviewed 8 files and all commit messages, and resolved 3 discussions.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3985)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on kungfucraig).

kungfucraig (approved):
> @kungfucraig reviewed 26 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3985)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on Marco-Premier).

Marco-Premier (commented):
:lgtm:

> @Marco-Premier reviewed 2 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/3985#-:-OvTnrFY7hqzjvhvIvu0:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on Marco-Premier).

Issue #4000, opened 2026-06-15
EDPA v1alpha: make Filter.state repeated across ListXxx services to save RPCs in Monitor scans


stevenwarejones:
## Problem

`ListXxx` filters in the EDPA `v1alpha` surface use a **singular** `state` field. The `VidLabelingMonitorFunction` (#3958) needs to ask "for this `(DataProvider, ModelLine)`, is any `RawImpressionUploadModelLine` currently in `POOL_ASSIGNING`, `RANKING`, or `LABELING`?" — i.e., **any of N non-terminal states**. With singular `state`, the answer requires either:

- **3 List RPCs** per `(DataProvider, ModelLine)` per Monitor cycle (one per state), then OR the results client-side.
- **1 List RPC** with no state filter, then full client-side filtering — wastes bandwidth as upload history grows.

Both work but are unnecessarily chatty. At ~100 EDPs × ~5 ModelLines × Monitor running every minute, the concurrent-dispatch-prevention gate alone would issue ~1500 List RPCs/minute under option 1, where 1 round trip would suffice.

## Affected services (family-wide)

All `ListXxx.Filter` messages with a `state` field across the EDPA v1alpha + internal mirrors:

| Resource | PR | Status |
|---|---|---|
| `RawImpressionUploadModelLine` | #3908 | merged |
| `RawImpressionUpload` | #3818 | merged |
| `RawImpressionMetadataBatch` | (in-tree) | merged |
| `PoolAssignmentJob` | #3929 | merged |
| `RankerJob` | #3941 | merged |
| `RequisitionMetadata` | (in-tree) | merged |

`VidLabelingJob` (#3984, open) should follow the same pattern when it lands.

## Proposed fix

Replace the singular `state` field with `repeated states` on each `Filter` message. The empty list matches all states (back-compat semantic for callers that previously omitted the field).

```proto
// Before
message Filter {
  RawImpressionUploadModelLine.State state = 1 [(google.api.field_behavior) = OPTIONAL];
  // ...
}

// After
message Filter {
  // Filter to resources whose state is in this set. Empty matches all states.
  repeated RawImpressionUploadModelLine.State states = 1 [(google.api.field_behavior) = OPTIONAL];
  // ...
}
```

### Wire-compatibility note

[...]

PR #3984, opened 2026-06-12
feat: Add VidLabelingJob proto definitions for internal and public API

getina:
## Summary
- Add public and internal LabelingJob proto definitions for Phase 2 (VID labeling) of the raw impression upload pipeline
- Add LabelingJobService with Create, BatchCreate, Get, List, MarkSucceeded, and MarkFailed RPCs
- Add internal LabelingState enum with CREATED, SUCCEEDED, FAILED states
- Follows the same patterns as PoolAssignmentJob and RankerJob

## Test plan
- [x] bazel build passes for all new proto targets
- [x] clang-format passes

Issue: #3980

PR #4001, opened 2026-06-15
feat: Make Filter.state repeated across EDPA v1alpha List services

stevenwarejones:
Replaces singular `state` filter field with `repeated state_in` on every ListXxx.Filter in the edpaggregator v1alpha surface (plus internal mirrors). Lets callers — most notably the VidLabelingMonitorFunction (#3958) — ask "any of these states" in one List RPC instead of fanning out per state.

The `_in` suffix follows the established repo convention for set-membership filter fields (`blob_uri_in`, `entity_type_in`, `type_in`), reads naturally as the SQL `IN` operator, and disambiguates that the field is a filter set rather than the resource's state vector. Each public proto carries an `api-linter: core::0140::prepositions=disabled` block with an aip.dev/not-precedent justification.

Source-incompatible (callers using the old `filter.state` symbol need a rename to `filter.stateInList`, and DSL `state = X` becomes `stateIn += X`), but **wire-compatible**: field number 1 is preserved and proto3's wire format treats singular and repeated as the same encoding for scalar fields. An older client writing a single `state` value to field 1 is interpreted by the new server as a one-element list, and a new client writing a single `state_in` element produces the same wire bytes as the old singular write.

12 protos: 6 resources × public + internal. Doc on every Filter is uniform: "Filter to resources whose state is in this set. Empty matches all states."

VidLabelingJob (#3984, open) should adopt this shape directly.

Issue: #4000

### DIFF
--- a/src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/spanner/db/RawImpressionMetadataBatch.kt
+++ b/src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/spanner/db/RawImpressionMetadataBatch.kt
@@ -207,7 +207,7 @@ fun AsyncDatabaseClient.ReadContext.readRawImpressionMetadataBatches(
     }
 
     if (filter.statesList.isNotEmpty()) {
-      conjuncts.add("State IN UNNEST(@states)")
+      conjuncts.add("CAST(State AS INT64) IN UNNEST(@states)")
     }
 
     if (after != null) {
--- a/src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/spanner/db/RawImpressionMetadataBatch.kt
+++ b/src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/spanner/db/RawImpressionMetadataBatch.kt
@@ -206,8 +206,8 @@ fun AsyncDatabaseClient.ReadContext.readRawImpressionMetadataBatches(
       conjuncts.add("DeleteTime IS NULL")
     }
 
-    if (filter.statesList.isNotEmpty()) {
-      conjuncts.add("CAST(State AS INT64) IN UNNEST(@states)")
+    if (filter.stateInList.isNotEmpty()) {
+      conjuncts.add("CAST(State AS INT64) IN UNNEST(@state_in)")
     }
 
     if (after != null) {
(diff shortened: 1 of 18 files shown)


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4001)

Marco-Premier (approved):
> @Marco-Premier reviewed 12 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4001)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on stevenwarejones).

SanjayVas (approved):
> Backwards-incompatible field rename + cardinality change. Safe today because none of these protos have shipped externally yet.

FYI this is actually safe in terms of binary compatibility. It may cause *build* breakage for clients when they update, but wouldn't change behavior for any existing clients even if this had already shipped.

> @SanjayVas reviewed 18 files and all commit messages, and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4001#-:-OvCDNxpAcQgRA_nklFN:bbj2iv1)* status: all files reviewed, 1 unresolved discussion (waiting on stevenwarejones).

___
*[`src/main/proto/wfa/measurement/edpaggregator/v1alpha/pool_assignment_job_service.proto` line 154 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4001#-OvCDrZb7OT_CEYe5qEZ:-OvCDrZcAN_BbOdfL3Le:b-ehocab) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0fe04d69004f80e57f1e48bbef6eff9549d89d4e/src/main/proto/wfa/measurement/edpaggregator/v1alpha/pool_assignment_job_service.proto#L154)):*
> ```proto
>     //     aip.dev/not-precedent: This is a filter input field, not
>     //     a resource state field. --)
>     repeated PoolAssignmentJob.State states = 1
> ```

nit: elsewhere we use prepositions to make this clearer, though there's the caveat of needing to disable another lint rule



_Suggestion:_
```proto
state_in
```

stevenwarejones (commented):
Right — thanks for the correction. Updated the PR description to say *source-incompatible but wire-compatible*, with the field-1 preservation note (singular and repeated proto3 scalars share the same wire encoding, so older clients keep working at the byte level).

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4001#-:-OvCGGfg-GqH04TtXk5L:bi01yi0)* status: all files reviewed, 1 unresolved discussion (waiting on stevenwarejones).

stevenwarejones (commented):
> @stevenwarejones reviewed 18 files and all commit messages, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4001)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on stevenwarejones).

PR #4003, opened 2026-06-15
feat: Add create_time_interval filter to EDPA v1alpha List services

stevenwarejones:
Adds google.type.Interval create_time_interval to the Filter messages of RawImpressionMetadataBatch, RawImpressionMetadataBatchFile, RawImpressionUploadFile, and RequisitionMetadata (public v1alpha + internal mirrors). Brings these services in line with the phase-job services (PoolAssignmentJob, RankerJob, RawImpressionUploadModelLine, RawImpressionUpload) which already have it.

Required for the VidLabelingMonitorFunction (#3958) staleness sweeps. Wire-compatible additive change; no existing callers break.

Issue: #4002

PR #3997, opened 2026-06-15
feat: Add request_id and etag-idempotency docs to MarkRankerJob/MarkPoolAssignmentJob*Succeeded RPCs

stevenwarejones:
Adds AIP-155 `request_id` field to `MarkRankerJobSucceeded` and `MarkPoolAssignmentJobSucceeded` request messages (public + internal) so retries can recover the original response payload after a lost network reply. Updates the RPC docstrings to describe the two-layer contract:

- **AIP-154 etag** on each request: optimistic locking. Server returns `ABORTED` on etag mismatch.
- **AIP-155 request_id** on each request: retry idempotency. Server returns the original response on retry with the same `request_id` without re-transitioning the resource.

The third RPC in the family (`MarkVidLabelingJobSucceeded`) is handled in #3984 directly with the same shape.

(History note: an earlier version of this PR proposed reusing `etag` as the idempotency key. That was rejected as inconsistent with AIP-154's strict ABORTED-on-mismatch rule; this version uses AIP-155 `request_id` instead, matching the rest of the family.)

Issue: #3996

PR #4013, opened 2026-06-17
feat: Add RawImpressionUpload internal and public API

Marco-Premier:
## Summary
  
  Implements the internal (Spanner) and public `v1alpha` gRPC service for the
  `RawImpressionUpload` resource — the top-level parent of the VID labeling
  pipeline. The proto definitions already landed in earlier PRs (#3984, #3950,
  #3941, #3929, #3908, #3878, #3818, #4003, #4001, #3997); this PR provides the
  service implementations on top of the schema from #3989.

  Follows the existing `RawImpressionMetadataBatch` / `ImpressionMetadata`
  service patterns.
  ## Changes
  
  ### Internal API (Spanner)
  - **`SpannerRawImpressionUploadService`** — `Create` / `Get` / `List`, backed by
    the `RawImpressionUpload` table. Request-id idempotency and server-generated
    resource IDs.
  - **`db/RawImpressionUpload.kt`** — read / insert / exists / list helpers, with
    `state` and `create_time` interval filtering plus keyset pagination.
  - **`internal/Errors.kt`** — new `RAW_IMPRESSION_UPLOAD_NOT_FOUND` reason,
    `RAW_IMPRESSION_UPLOAD_RESOURCE_ID` metadata, and
    `RawImpressionUploadNotFoundException`.
    
  ### Public API (v1alpha)
  - **`RawImpressionUploadService`** — public ⇄ internal translation, `done_blob_uri`
    validation, state mapping, and error mapping (`NOT_FOUND`).
  - **`RawImpressionUploadKey`** — resource-name parsing/assembly for
    `dataProviders/{data_provider}/rawImpressionUploads/{raw_impression_upload}`.
  - **`Errors.kt`** — public `RAW_IMPRESSION_UPLOAD_NOT_FOUND` reason,
    `RAW_IMPRESSION_UPLOAD` metadata, and `RawImpressionUploadNotFoundException`.
    
  ### Wiring
  - Registered the service in internal `Services` / `InternalApiServices` and
    public v1alpha `Services`.
  Adding `RAW_IMPRESSION_UPLOAD_NOT_FOUND` to the shared `InternalErrors.Reason`
  enum makes every exhaustive `when` over it non-exhaustive, so the existing
  `ImpressionMetadataService`, `RequisitionMetadataService`,
  `RawImpressionMetadataBatchService`, and `RawImpressionMetadataBatchFileService`
  each get the new reason added to their grouped `→ Status.INTERNAL` arm. No
  behavioral change to those services.

  ### Tests
  - Abstract internal service test (`service/internal/testing`).
  - Spanner-emulator-backed internal test.
  - Spanner-emulator-backed public service test.

  ## Testing
  - Spanner-emulator tests compile but cannot execute on the current devbox
    (`GLIBCXX_3.4.32 not found` for the emulator binary — pre-existing
    environment issue, reproduces on existing `SpannerRawImpressionMetadataBatch`
    tests). Expected to pass in CI.

  ## Dependencies / Follow-ups
  - Depends on the schema in **#3989**. This branch targets that schema **as it
    currently stands** (string `RawImpressionUploadId` as PK). Once the #3989
    review feedback lands (INT64 surrogate PK + separate
    `RawImpressionUploadResourceId STRING(36)`, plus the
    `RawImpressionUploadByCreateTime` index), `db/RawImpressionUpload.kt` and
    `SpannerRawImpressionUploadService` will be updated to the Batch-style
    surrogate-id pattern (re-introducing the `IdGenerator` and keying reads on the
    resource id).

Issue: #3887


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4013)

stevenwarejones:
CI's Lint check is failing — `ktfmt produced diffs` ([log](https://github.com/world-federation-of-advertisers/cross-media-measurement/actions/runs/27685872469/job/81884315668)). Run `ktfmt --google-style` over the new/modified Kotlin files before the next push.

src/main/resources/edpaggregator/spanner/create-vid-labeling-schema.sql
```diff
@@ -0,0 +1,242 @@
...
+
+CREATE UNIQUE INDEX RankIndexBlobByResourceId
+  ON RankIndexBlob(DataProviderResourceId, RankIndexBlobResourceId);
+
+CREATE UNIQUE NULL_FILTERED INDEX RankIndexBlobByCreateRequestId
+  ON RankIndexBlob(DataProviderResourceId, CreateRequestId);
+
+CREATE INDEX RankIndexBlobByMaxEventDate
+  ON RankIndexBlob(DataProviderResourceId, CmmsModelLine, BlobType, MaxEventDate)
+  WHERE DeleteTime IS NULL;
```
stevenwarejones: CI blocker — Spanner GoogleSQL dialect doesn't support `WHERE` clauses on `CREATE INDEX`. The two `WHERE DeleteTime IS NULL` clauses (here and at line 209) fail at schema migration time:

```
Error parsing Spanner DDL statement: CREATE INDEX RankIndexBlobByMaxEventDate
  WHERE DeleteTime IS NULL : Syntax error on line 3, column 23: Expecting 'NOT' but found 'NULL'
```

This breaks every downstream test that needs the schema — 30+ EDPA service + integration tests are failing for this single reason (all show the same Liquibase migration error).

The existing pattern in this repo (e.g. `RawImpressionMetadataBatchByState` in `refactor-raw-impression-schema.sql`) is to use plain `CREATE INDEX` without a soft-delete filter and apply `WHERE DeleteTime IS NULL` at query time. Suggested fix:

```sql
CREATE INDEX RankIndexBlobByMaxEventDate
  ON RankIndexBlob(DataProviderResourceId, CmmsModelLine, BlobType, MaxEventDate);

CREATE INDEX RankIndexBlobByUploadAndType
  ON RankIndexBlob(DataProviderResourceId, CmmsModelLine, BlobType, RawImpressionUploadId);
```

src/main/resources/edpaggregator/spanner/create-vid-labeling-schema.sql
```diff
@@ -0,0 +1,242 @@
...
+CREATE UNIQUE INDEX RawImpressionUploadModelLineByResourceId
+  ON RawImpressionUploadModelLine(DataProviderResourceId, RawImpressionUploadModelLineResourceId);
+
+CREATE UNIQUE NULL_FILTERED INDEX RawImpressionUploadModelLineByCreateRequestId
+  ON RawImpressionUploadModelLine(DataProviderResourceId, RawImpressionUploadId, CreateRequestId);
+
+-- =============================================================================
+-- PoolAssignmentJob — Phase-0 gate.
+-- One row per (upload, model line, shard); pre-created by the dispatcher
+-- in POOL_ASSIGNMENT_CREATED state. The "last shard out" of a (upload,
```
stevenwarejones: nit: the comment says *"pre-created by the dispatcher in `POOL_ASSIGNMENT_CREATED` state"* but the actual enum value (`wfa.measurement.internal.edpaggregator.PoolAssignmentState`) is `POOL_ASSIGNMENT_STATE_CREATED`. Small doc typo.

stevenwarejones (commented):
> @stevenwarejones reviewed 27 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4013)* status: all files reviewed, 2 unresolved discussions (waiting on getina and Marco-Premier).

src/main/kotlin/org/wfanet/measurement/edpaggregator/service/internal/testing/RawImpressionUploadServiceTest.kt
```diff
@@ -0,0 +1,480 @@
...
+            ListRawImpressionUploadsRequestKt.filter {
+              createTimeIn = interval { startTime = Instant.now().minusSeconds(3600).toProtoTime() }
+            }
+        }
+      )
+    assertThat(pastResponse.rawImpressionUploadsList).hasSize(2)
+  }
+
+  @Test
+  fun `listRawImpressionUploads returns remaining uploads using page token`() = runBlocking {
```
stevenwarejones: CI blocker — this test fails JUnit validation with `Method ... should be void`. The last expression in the `runBlocking { ... }` block is `assertThat(...).containsExactlyElementsIn(...)`, which returns Truth's `Ordered` subject. Kotlin then infers the method's return type as `Ordered`, not `Unit`, and JUnit rejects the method.

Fix: annotate explicitly.

```kotlin
fun `listRawImpressionUploads returns remaining uploads using page token`(): Unit = runBlocking {
  ...
}
```

Worth scanning every `() = runBlocking {` in this file and the parallel `RawImpressionUploadServiceTest` in `service/v1alpha/testing/` — any method whose `runBlocking` block doesn't end in a `Unit`-returning expression will hit the same failure once the schema bug at `create-vid-labeling-schema.sql:205` is fixed and the suite actually runs to validation. Easiest blanket fix: add `: Unit =` to all of them.

stevenwarejones:
CI / merge-state issue — **stale base regression**.

This PR is reverting #3974 (the gap-detection fix that landed on main June 16). Three files in this PR's diff are restoring the pre-#3974 state:
- `src/main/kotlin/.../dataavailability/DataAvailabilityMonitor.kt` (29 line diff — removes `earliestDate` field from `DateInfo`, reverts `findGaps(presentDates(...))` back to `findGaps(sortedDates)`)
- `src/main/kotlin/.../dataavailability/DataAvailabilitySync.kt` (31 lines)
- `src/test/kotlin/.../dataavailability/DataAvailabilitySyncTest.kt` (162 deletions)

Marco's branch was forked before #3974 landed and hasn't been rebased; merging as-is silently undoes that bug fix.

**Fix:** rebase onto current main and keep main's version of these three files. RawImpressionUpload service files don't touch `dataavailability/`, so the resolution is simply "take main's side" on all three.

src/main/kotlin/org/wfanet/measurement/edpaggregator/service/internal/testing/RawImpressionUploadServiceTest.kt
Marco-Premier: Done.

src/main/resources/edpaggregator/spanner/create-vid-labeling-schema.sql
Marco-Premier: I copied this comment over pr #3989

src/main/resources/edpaggregator/spanner/create-vid-labeling-schema.sql
Marco-Premier: Sorry I had a wrong base. This PR depends on #3989 and I have left already a comment there to fix this

Marco-Premier (commented):
1. Will run the rebase after @getina merges #3989 since this PR is stacked on top of that branch
2. Linting done

> @Marco-Premier made 4 comments.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4013#-:-OvLJTVC1LrC00JwT7u_:bl0zpk3)* status: 24 of 27 files reviewed, 3 unresolved discussions (waiting on getina and stevenwarejones).

src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/spanner/db/RawImpressionUpload.kt
```diff
@@ -0,0 +1,247 @@
...
+        conjuncts.add("CreateTime >= @createTimeStart")
+      }
+      if (filter.createTimeIn.hasEndTime()) {
+        conjuncts.add("CreateTime < @createTimeEnd")
+      }
+    }
+
+    if (after != null) {
+      conjuncts.add(
+        "(CreateTime > @afterCreateTime) OR (CreateTime = @afterCreateTime AND RawImpressionUploadResourceId > @afterRawImpressionUploadResourceId)"
```
stevenwarejones: Blocking — the pagination conjunct leaks rows across data providers when combined with any other conjunct.

The added clause is internally `A OR B`:

```kotlin
conjuncts.add(
  "(CreateTime > @afterCreateTime) OR (CreateTime = @afterCreateTime AND RawImpressionUploadResourceId > @afterRawImpressionUploadResourceId)"
)
```

After `joinToString(" AND ")` and without outer parens, SQL precedence (`AND` binds tighter than `OR`) parses the assembled `WHERE` as:

```sql
(DataProviderResourceId = @dpri AND <state-filter> AND CreateTime > @afterCreateTime)
OR
(CreateTime = @afterCreateTime AND RawImpressionUploadResourceId > @afterResId)
```

The second OR-branch is unscoped by `DataProviderResourceId` (and by any state/createTime filter), so the second page can return rows from other data providers whose `CreateTime` exactly equals `@afterCreateTime` and whose resource ID sorts after `@afterResId`.

Fix — wrap the OR in parens:

```kotlin
conjuncts.add(
  "((CreateTime > @afterCreateTime) OR (CreateTime = @afterCreateTime AND RawImpressionUploadResourceId > @afterRawImpressionUploadResourceId))"
)
```

The existing pagination test doesn't catch this because it uses one DP with no state filter. Worth adding a test that paginates with a state filter active, and one with two DPs whose rows share the same `CreateTime`.

Note: `db/RawImpressionMetadataBatch.kt` has the same shape — pre-existing, out of scope for this PR, but worth a follow-up issue.

src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/spanner/InternalApiServices.kt
```diff
@@ -32,6 +32,7 @@ object InternalApiServices {
       SpannerImpressionMetadataService(databaseClient, coroutineContext),
       SpannerRawImpressionMetadataBatchService(databaseClient, coroutineContext, idGenerator),
       SpannerRawImpressionMetadataBatchFileService(databaseClient, coroutineContext, idGenerator),
+      SpannerRawImpressionUploadService(databaseClient, coroutineContext),
```
stevenwarejones: Blocking — `idGenerator` is not threaded into `SpannerRawImpressionUploadService` here.

```kotlin
return Services(
  SpannerRequisitionMetadataService(databaseClient, coroutineContext, idGenerator),
  SpannerImpressionMetadataService(databaseClient, coroutineContext),
  SpannerRawImpressionMetadataBatchService(databaseClient, coroutineContext, idGenerator),
  SpannerRawImpressionMetadataBatchFileService(databaseClient, coroutineContext, idGenerator),
  SpannerRawImpressionUploadService(databaseClient, coroutineContext),  // missing idGenerator
)
```

The service accepts `idGenerator` and uses it in `idGenerator.generateNewId { id -> txn.rawImpressionUploadExists(...) }`, but the wiring drops it on the floor, so production always gets `IdGenerator.Default` while every sibling service gets the injected one. Add `, idGenerator` to the constructor call to stay consistent.

stevenwarejones (commented):
Fresh review pass — found one blocking pagination bug, one wiring miss, and two test-coverage gaps. The previously raised schema/lint/Unit-annotation issues are acknowledged separately.

src/test/kotlin/org/wfanet/measurement/edpaggregator/service/v1alpha/RawImpressionUploadServiceTest.kt
```diff
@@ -0,0 +1,491 @@
...
+          errorInfo {
+            domain = Errors.DOMAIN
+            reason = Errors.Reason.INVALID_FIELD_VALUE.name
+            metadata[Errors.Metadata.FIELD_NAME.key] = "page_token"
+          }
+        )
+    }
+
+  @Test
+  fun `listRawImpressionUploads throws INVALID_ARGUMENT for unspecified state filter`(): Unit =
```
stevenwarejones, line 388: The public service rejects both `STATE_UNSPECIFIED` and `UNRECOGNIZED`:

```kotlin
if (state == RawImpressionUpload.State.STATE_UNSPECIFIED ||
    state == RawImpressionUpload.State.UNRECOGNIZED) {
  throw InvalidFieldValueException("filter.state_in")...
}
```

but only the `STATE_UNSPECIFIED` case is tested. Add a parallel test that asserts the same `INVALID_ARGUMENT` / `filter.state_in` response for `UNRECOGNIZED`.

src/main/kotlin/org/wfanet/measurement/edpaggregator/service/internal/testing/RawImpressionUploadServiceTest.kt
```diff
@@ -0,0 +1,483 @@
...
+            ListRawImpressionUploadsRequestKt.filter {
+              createTimeIn = interval { startTime = Instant.now().minusSeconds(3600).toProtoTime() }
+            }
+        }
+      )
+    assertThat(pastResponse.rawImpressionUploadsList).hasSize(2)
+  }
+
+  @Test
+  fun `listRawImpressionUploads returns remaining uploads using page token`(): Unit = runBlocking {
```
stevenwarejones, line 347: The pagination test creates 3 uploads with no filters and pages through them. Combined with the pagination-predicate fix on `db/RawImpressionUpload.kt`, please add a test that paginates while a state filter (or createTime filter) is active — that's the test shape that would have caught the precedence bug, and it locks in the corrected predicate going forward.

src/main/kotlin/org/wfanet/measurement/edpaggregator/service/internal/testing/RawImpressionUploadServiceTest.kt
stevenwarejones: Verified — every `runBlocking { }` test method in both `service/internal/testing/RawImpressionUploadServiceTest.kt` and `service/v1alpha/testing/RawImpressionUploadServiceTest.kt` now has the `: Unit =` annotation.

src/main/resources/edpaggregator/spanner/create-vid-labeling-schema.sql
stevenwarejones: Acknowledged — tracked on #3989 (-OvLHz5x3YlcA9Xpn0oE). Resolving here.

src/main/resources/edpaggregator/spanner/create-vid-labeling-schema.sql
stevenwarejones: Acknowledged — tracked on #3989 via the parallel comment there (gh-3430418550). Resolving here as moved-to-#3989.

stevenwarejones (commented):
Acknowledged — lint is clean now. Flagging that the dataavailability/* regression (DataAvailabilityMonitor, DataAvailabilitySync, DataAvailabilitySyncTest) will need to be re-checked after the rebase onto post-#3989 main, since #3989's base is also pre-#3974.

> @stevenwarejones+AGNT made 4 comments and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4013#-:-OvLpVuz0RW-DvXbd1dd:b-6bo98l)* status: 24 of 27 files reviewed, 6 unresolved discussions (waiting on getina, Marco-Premier, and stevenwarejones).

stevenwarejones (commented):
> @stevenwarejones reviewed 3 files and all commit messages, and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4013)* status: all files reviewed, 4 unresolved discussions (waiting on getina and Marco-Premier).

src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/spanner/InternalApiServices.kt
Marco-Premier: Done.

src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/spanner/db/RawImpressionUpload.kt
Marco-Premier: Done.

src/main/kotlin/org/wfanet/measurement/edpaggregator/service/internal/testing/RawImpressionUploadServiceTest.kt
Marco-Premier: Done.

src/test/kotlin/org/wfanet/measurement/edpaggregator/service/v1alpha/RawImpressionUploadServiceTest.kt
Marco-Premier: Done.

Marco-Premier (commented):
> @Marco-Premier made 4 comments.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4013)* status: 22 of 28 files reviewed, 4 unresolved discussions (waiting on getina and stevenwarejones).

src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/spanner/InternalApiServices.kt
stevenwarejones: Verified in commit 0478af98b. `InternalApiServices.build` now passes `idGenerator` through to `SpannerRawImpressionUploadService` (line 33). All Spanner services in this module now share the same injectable generator (consistent with `SpannerRequisitionMetadataService`, `SpannerRawImpressionMetadataBatchService`, etc.). Closing.

src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/spanner/db/RawImpressionUpload.kt
stevenwarejones: Verified in commit 0478af98b. `readRawImpressionUploads` now scopes the `after` conjunct correctly: `DataProviderResourceId = @dataProviderResourceId` is the first WHERE clause (line 177), and the `(CreateTime > X) OR (CreateTime = X AND ResourceId > Y)` predicate operates within that already-narrowed scope. The cross-DP leak path is closed.

The new dedicated `RawImpressionUploadPaginationTest.kt` is exactly the regression guard I asked for — single-transaction multi-row insert to force identical `CreateTime`, then asserts provider B's row does not surface on provider A's next page. Better than expanding the existing test (which can't construct identical commit timestamps because the service commits per-row). Closing.

src/main/kotlin/org/wfanet/measurement/edpaggregator/service/internal/testing/RawImpressionUploadServiceTest.kt
stevenwarejones: Verified — addressed differently and better than I asked. Marco added a new dedicated test file `RawImpressionUploadPaginationTest.kt` at the DB layer (where the leak condition is actually constructible: same-CreateTime requires a single-transaction multi-row insert that the v1alpha service can't produce). Test inserts one row per DP in one transaction → forces identical `CreateTime` → asserts the `after` conjunct correctly scopes to the queried DP and does not leak provider B's row onto provider A's next page. Regression guard locked in. Closing.

src/test/kotlin/org/wfanet/measurement/edpaggregator/service/v1alpha/RawImpressionUploadServiceTest.kt
stevenwarejones: Verified in commit 0478af98b. `listRawImpressionUploads` at lines 195-205 explicitly rejects both `STATE_UNSPECIFIED` and `UNRECOGNIZED` in the state filter. Test coverage exists for both:
- `listRawImpressionUploads throws INVALID_ARGUMENT for unspecified state filter` (line 388)
- `listRawImpressionUploads throws INVALID_ARGUMENT for unrecognized state filter` (line 414)

Closing.

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4013)* status: 20 of 28 files reviewed, 4 unresolved discussions (waiting on getina and stevenwarejones).

stevenwarejones (approved):
:lgtm:

> @stevenwarejones reviewed 10 files and all commit messages, made 1 comment, and resolved 4 discussions.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4013#-:-OvjizUR5jdo1e7t6402:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on getina).

SanjayVas (approved):
> @SanjayVas reviewed 28 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4013)* status: all files reviewed, 1 unresolved discussion (waiting on getina and Marco-Premier).

___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/spanner/db/RawImpressionUpload.kt` line 81 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4013#-OvkhC19BmhMOlk_KB_b:-OvkhC19BmhMOlk_KB_c:bd0htxs) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/0478af98bbfd89007c58d59970b40e83a7f301bc/src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/spanner/db/RawImpressionUpload.kt#L81)):*
> ```Kotlin
> 
> /** Finds an existing [RawImpressionUpload] by request ID for idempotency. */
> suspend fun AsyncDatabaseClient.ReadContext.findExistingUploadByRequestId(
> ```

nit: Make sure it is clear that this is the request ID for the Create method. The "existing" part is redundant.

The request ID is bound to the method. which is why the DB column is `CreateRequestId`.



_Suggestion:_
```Kotlin
findUploadByCreateRequestId
```

Marco-Premier (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4013)* status: 26 of 28 files reviewed, all discussions resolved (waiting on getina, SanjayVas, and stevenwarejones).

stevenwarejones (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4013)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on getina and SanjayVas).

stevenwarejones (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4013)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on getina and SanjayVas).

Issue #3889, opened 2026-06-03
RawImpressionUploadFileService implementation


Marco-Premier:
Implement the public and internal Kotlin services for `RawImpressionUploadFileService`, backed by the Spanner DAO.

PR #3878, opened 2026-06-02
feat: Add RawImpressionUploadFile proto definitions for internal and public API

getina:
## Summary
- Add proto message and service definitions for RawImpressionUploadFile, a child resource of RawImpressionUpload for tracking individual raw impression files.
- Includes create, batch create, get, list, delete (soft), and batch delete methods for both internal and public (v1alpha) APIs.
- Both internal and public proto definitions in a single PR per task guidance.
- Soft-delete via delete_time with show_deleted on List, matching existing RawImpressionMetadataBatchFile pattern.
- AIP-159 wildcard support documented on public List parent field.

## Test plan
- bazel build passes for all 4 new proto targets (internal message, internal service, public message, public service)

Closes: #3881
Issue: #3881

PR #4009, opened 2026-06-16
feat: implement VidRankBuilder (Phase-1 ranker) for memoized VID assignment

Marco-Premier:
# SUMMARY

The orchestration and TEE-app layer of the Phase-1 ranker — wires the subpool ranker into a WorkItem-driven TEE application and the Phase-0 → Phase-1 → Phase-2 state machine.

  ## What's included
  
  **`VidRankBuilder`** — per-WorkItem orchestrator:
  - Ranks each assigned subpool via `SubpoolRanker`, then marks its `RankerJob` `SUCCEEDED` using **etag optimistic locking** (AIP-154) so a redelivered/duplicate ranker can't corrupt state — a stale
  write is detected (`FAILED_PRECONDITION`/`ABORTED`) and handled idempotently.
  - **Last-job-out** fans out Phase-2: flips the parent `RawImpressionUploadModelLine` `RANKING → LABELING` (also etag-guarded), making `LABELING` the durable completion marker.
  - **Recovery path** for redelivery of an already-`SUCCEEDED` last job: re-derives the last-out condition (all `RankerJob`s succeeded, parent still `RANKING`) and re-runs the idempotent fan-out; a
  missing parent is treated as a hard invariant violation.
  - Best-effort `RankerJob` → `FAILED` transition on unexpected errors so operators see failures.

  **`VidRankBuilderApp` / `VidRankBuilderAppRunner`** — TEE app wiring (stores, service stubs, queue subscriber) and the entrypoint.

  **`vid_rank_builder_params.proto`** — adds `subpool_ranked_sizes` so each WorkItem carries the `ranked_size` per subpool.

  **Phase-0 producer wiring** — `PoolEmitLabeler.rankedSize(poolOffset)` plus `SubpoolAssigner` stamping `subpool_ranked_sizes` at last-shard-out, so the Phase-1 ranker can size its rank `BitSet` and cap
  allocation **without loading the model itself**.

  ## Design notes
  Params carry everything the TEE needs (model-line windows, blob URIs, ranked sizes), so the TEE apps never call the VID Repository. State transitions are etag-CAS guarded end-to-end; the last-out 
  fan-out is idempotent under at-least-once delivery.

  ## Testing
  `VidRankBuilderTest` (non-last vs last job, redelivery recovery, missing-parent failure, per-subpool failure handling) and updated `SubpoolAssigner` tests covering the new `rankedSize` plumbing.

Issue: #3921 
Issue: #4015

PR #4014, opened 2026-06-17
feat: Implement RawImpressionUploadFile public and internal API

Marco-Premier:
## Summary

  Implements the internal (Spanner-backed) and public `v1alpha`
  `RawImpressionUploadFileService`. The gRPC/proto definitions already exist
  (#3878, plus the `create_time_in` filter from #4003) and the DB schema landed
  with the VID labeling pipeline tables (#3989); this PR provides the service
  implementations, resource keys, error types, registration, and tests.

  The design mirrors the existing `RawImpressionMetadataBatchFile` services so the
  two stay consistent.

  ## Changes

  ### Internal API
  - **`db/RawImpressionUploadFile.kt`** — Spanner access layer: get / batch-get /
    list / insert / soft-delete, parent-`RawImpressionUpload` existence check,
    `request_id` idempotency lookup, `blob_uri_in` + `create_time_in` (half-open
    interval) filters, and keyset pagination ordered by
    `(CreateTime, RawImpressionUploadId, FileResourceId)`.
  - **`SpannerRawImpressionUploadFileService.kt`** — `Create`,
    `BatchCreate`, `Get`, `List`, `Delete`, `BatchDelete`. File resource IDs are
    server-generated; idempotent on `request_id`; duplicate `blob_uri` is allowed
    (no uniqueness constraint in the schema).
  - New internal error reasons/exceptions: `RAW_IMPRESSION_UPLOAD_NOT_FOUND`,
    `RAW_IMPRESSION_UPLOAD_FILE_NOT_FOUND`.

  ### Public API (v1alpha)
  - **`RawImpressionUploadFileService.kt`** — translates resource names ↔ internal
    IDs, maps internal errors to public statuses, supports the AIP-159 `-`
    wildcard on the upload segment for cross-upload `List`.
  - **`RawImpressionUploadKey` / `RawImpressionUploadFileKey`** resource keys.
  - New public error reasons/exceptions for upload / upload-file not-found.

  ### Wiring
  - Registered the new service in the internal and v1alpha `Services` aggregators
    and `InternalApiServices.build`.
  - Extended the exhaustive `when (InternalErrors.getReason(...))` blocks in the
    sibling v1alpha services (Impression, RequisitionMetadata,
    RawImpressionMetadataBatch, RawImpressionMetadataBatchFile) to handle the two
    new enum values.

  ## Testing
  - Abstract internal test (`service/internal/testing`) covering create
    (incl. idempotency, validation, parent-not-found), batch create, get, list
    (pagination, `show_deleted`, `blob_uri_in`, `create_time_in`, cross-upload
    wildcard), delete, and batch delete.
  - `SpannerRawImpressionUploadFileServiceTest` (Spanner emulator) running the
    abstract suite.
  - `RawImpressionUploadFileServiceTest` (v1alpha) for the public surface.

  All new targets build; both emulator-backed suites pass locally. Formatted with
  `ktfmt 0.54 --google-style`.

  ## Related
  - Protos: #3878, #4003
  - Schema: #3989

Issue: #3889


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4014)

src/main/resources/edpaggregator/spanner/create-vid-labeling-schema.sql
```diff
@@ -0,0 +1,242 @@
...
+
+CREATE UNIQUE INDEX RankIndexBlobByResourceId
+  ON RankIndexBlob(DataProviderResourceId, RankIndexBlobResourceId);
+
+CREATE UNIQUE NULL_FILTERED INDEX RankIndexBlobByCreateRequestId
+  ON RankIndexBlob(DataProviderResourceId, CreateRequestId);
+
+CREATE INDEX RankIndexBlobByMaxEventDate
+  ON RankIndexBlob(DataProviderResourceId, CmmsModelLine, BlobType, MaxEventDate)
+  WHERE DeleteTime IS NULL;
```
stevenwarejones: CI failing — 381 emulator-backed test failures, all cascading from a single broken DDL statement in the inherited base. Root cause: `create-vid-labeling-schema.sql:205,209` uses `CREATE INDEX ... WHERE DeleteTime IS NULL`, which Spanner GoogleSQL doesn't support. The Liquibase migration fails at changeset #11, so every test in this PR's emulator suites fails at setup.

The schema file is unchanged between this PR and its base (`getina/vid-labeling-spanner-schema`), so the fix belongs in #3989, not here. Once that PR rebases to use `CREATE NULL_FILTERED INDEX`, this PR inherits the fix and the 381 failures clear automatically. Flagging here so it's visible — no action for you on this PR's diff. Same blocker hit #4013 too.

stevenwarejones (commented):
FYI — schema DDL blocker inherited from base; fix lands in #3989.

stevenwarejones (commented):
> @stevenwarejones partially reviewed 27 files.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4014)* status: 25 of 27 files reviewed, 1 unresolved discussion (waiting on getina and Marco-Premier).

src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/spanner/SpannerRawImpressionUploadFileService.kt
```diff
@@ -0,0 +1,516 @@
...
+          }
+          val fileId: Long =
+            idGenerator.generateNewId { id ->
+              txn.rawImpressionUploadFileExists(
+                file.dataProviderResourceId,
+                rawImpressionUploadId,
+                id,
+              )
+            }
+          val fileResourceId: String = UUID.randomUUID().toString()
```
stevenwarejones: Blocker — `fileResourceId = UUID.randomUUID().toString()` (also line 232 in `batchCreate`) creates a retry-orphan hazard. Sequence:

1. Client calls `createRawImpressionUploadFile` WITHOUT `requestId` (allowed — AIP-155 idempotency is opt-in).
2. Server commits the row, generates `fileResourceId = U1`, returns the response.
3. The response is lost on the wire (TCP reset, deadline exceeded, anything).
4. Client retries the same Create. No `requestId`, so the idempotency cache at `findExistingUploadFilesByRequestIds` misses. Server commits a SECOND row with `fileResourceId = U2`.

Result: two `RawImpressionUploadFile` rows for what the client believed was one logical file, and the client never sees `U1`. `U1` is orphaned and only the soft-delete sweep or List-then-Delete can find it. The data in `U1` (its `blob_uri`) silently persists.

Two fixable options:

1. Make `request_id` a required field (drop it from "opt-in for idempotency" to "required for at-least-once safety"). Validate non-empty up front; the server-derived idempotency cache then always catches retries. Matches AIP-155's "should" for resource creation in unreliable networks.
2. Document explicitly in `RawImpressionUploadFile.create` proto (and the public service KDoc) that callers MUST supply `request_id` for at-least-once safety, and orphan rows from missing-request_id retries are the caller's responsibility to GC. Weaker, but consistent with sibling services if they have the same behavior.

Either way: the current code reads as "idempotent by design" because `findExistingUploadFilesByRequestIds` exists, but the design is incomplete without one of the above. Flagging this because the same shape exists in the sibling `SpannerRawImpressionMetadataBatchFileService` — worth deciding the policy once and applying it across.

stevenwarejones (commented):
Holistic review: retry-orphan hazard, idgen consistency, test coverage gap, error message.

stevenwarejones (commented):
IdGenerator injection consistency.

src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/spanner/SpannerRawImpressionUploadFileService.kt
```diff
@@ -0,0 +1,516 @@
...
+            val blobUri: String = subRequest.rawImpressionUploadFile.blobUri
+            val fileId: Long =
+              idGenerator.generateNewId { id ->
+                txn.rawImpressionUploadFileExists(
+                  request.dataProviderResourceId,
+                  rawImpressionUploadId,
+                  id,
+                )
+              }
+            val fileResourceId: String = UUID.randomUUID().toString()
```
stevenwarejones: nit: inconsistent ID generation. `fileId` (line 109-115) goes through `idGenerator.generateNewId { ... }` (injectable, deterministic in tests), but `fileResourceId` on the next line uses `UUID.randomUUID().toString()` (not injectable). The public name flows through `fileResourceId`, so tests can't assert resource names without parsing them post-hoc, and the test seam designed for one ID is bypassed for the other.

Two options:
1. Run `fileResourceId` through `idGenerator` too (`IdGenerator.Default` would still produce a UUID string; tests can inject a deterministic generator).
2. Keep `UUID.randomUUID()` if a stable contract on `fileResourceId` opacity is wanted, but document why this ID escapes the injection seam.

Same pattern shows up in `batchCreateRawImpressionUploadFiles` (line 232).

src/main/kotlin/org/wfanet/measurement/edpaggregator/service/internal/testing/RawImpressionUploadFileServiceTest.kt
```diff
@@ -0,0 +1,850 @@
...
+
+    val file1: RawImpressionUploadFile = createFile(uploadId, requestId = requestId)
+    val file2: RawImpressionUploadFile =
+      createFile(uploadId, blobUri = "gs://other-uri", requestId = requestId)
+
+    assertThat(file2).isEqualTo(file1)
+  }
+
+  @Test
+  fun `createRawImpressionUploadFile allows duplicate blob_uri`() = runBlocking {
```
stevenwarejones: nit: this test doesn't exercise what its name claims. Without `requestId`s, every `createFile` call creates a new row regardless of `blob_uri`, so two calls producing two files is trivially true — the test passes even if the schema later adds a `UNIQUE(BlobUri)` constraint that breaks the intent.

To actually validate "duplicate blob_uri is allowed when requestIds differ":

```kotlin
val rid1 = UUID.randomUUID().toString()
val rid2 = UUID.randomUUID().toString()
val file1 = createFile(uploadId, requestId = rid1)
val file2 = createFile(uploadId, requestId = rid2)
assertThat(file1.fileResourceId).isNotEqualTo(file2.fileResourceId)
assertThat(file1.blobUri).isEqualTo(file2.blobUri)
```

That way, the test enforces "different request_ids with the same blob_uri produce two rows" — which is the actual contract.

stevenwarejones (commented):
Test coverage gap.

src/main/kotlin/org/wfanet/measurement/edpaggregator/service/v1alpha/RawImpressionUploadFileService.kt
```diff
@@ -0,0 +1,579 @@
...
+          fileKey.dataProviderId != uploadKey.dataProviderId ||
+            fileKey.rawImpressionUploadId != uploadKey.rawImpressionUploadId
+        ) {
+          throw InvalidFieldValueException("requests.$index.name") {
+              "File $name does not belong to parent ${request.parent}"
+            }
+            .asStatusRuntimeException(Status.Code.INVALID_ARGUMENT)
+        }
+
+        if (!nameSet.add(name)) {
```
stevenwarejones, line 543: nit: this throws `InvalidFieldValueException("requests.$index.name")` with no message body, so the client sees only "Invalid value for requests.2.name" with no hint about why. The internal-layer counterpart at `SpannerRawImpressionUploadFileService.kt:452` includes a detailed message (`"file_resource_id is duplicate in the batch of requests"`). Match it here:

```kotlin
throw InvalidFieldValueException("requests.$index.name") {
    "$it is duplicate in the batch of requests"
  }
```

stevenwarejones (commented):
Error message consistency.

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4014)* status: all files reviewed, 5 unresolved discussions (waiting on getina and Marco-Premier).

src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/spanner/SpannerRawImpressionUploadFileService.kt
Marco-Premier: Done.

src/main/resources/edpaggregator/spanner/create-vid-labeling-schema.sql
Marco-Premier: this is not part of this PR

Marco-Premier (commented):
> @Marco-Premier made 2 comments and resolved 3 discussions.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4014)* status: 23 of 29 files reviewed, 2 unresolved discussions (waiting on getina and stevenwarejones).

src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/spanner/SpannerRawImpressionUploadFileService.kt
stevenwarejones: Verified — Marco took the strongest possible fix: `request_id` is now REQUIRED on both `createRawImpressionUploadFile` (line 96-98) and `batchCreateRawImpressionUploadFiles` (line 200-203). Without a request_id, callers can't even reach the create path, so the retry-orphan hazard I described (lost response → retry without request_id → second row with new fileResourceId orphaning the first) is structurally impossible. Tests added: `createRawImpressionUploadFile throws INVALID_ARGUMENT if request_id not set` (internal:245), `batchCreateRawImpressionUploadFiles throws INVALID_ARGUMENT if request_id not set` (internal:264), and v1alpha equivalents (`for missing request_id`).

The public service also propagates the requirement: v1alpha at line 97 rejects empty requestId and passes through to internal at line 112.

Closing.

src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/spanner/SpannerRawImpressionUploadFileService.kt
stevenwarejones: Not addressed — `fileResourceId` is still generated via `UUID.randomUUID()` at lines 128 and 245, not through the injectable `idGenerator` (which is only used for the numeric `fileId`).

That said: this matches the established pattern from `SpannerRawImpressionMetadataBatchFileService` on `main` (`"file-${UUID.randomUUID()}"`). Per the principle of matching pre-existing convention, downgrading from blocker to acknowledged. Filing as a follow-up to track both files together rather than blocking this PR. Closing.

src/main/kotlin/org/wfanet/measurement/edpaggregator/service/internal/testing/RawImpressionUploadFileServiceTest.kt
stevenwarejones: Verified — test renamed to `createRawImpressionUploadFile allows duplicate blob_uri with different request_ids` (internal:110) and now uses `rid1` and `rid2` (lines 113-114) to actually exercise the "two different request_ids on same blob_uri produces two files" contract. Now that request_id is required (per the UUID blocker fix), this is the right shape — exercises what its name claims. Closing.

src/main/kotlin/org/wfanet/measurement/edpaggregator/service/v1alpha/RawImpressionUploadFileService.kt
stevenwarejones: Verified — `batchDelete` duplicate-name error at v1alpha:518-523 now includes a meaningful message body: `"$it is duplicate in the batch of requests"`. Closing.

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4014)* status: 18 of 30 files reviewed, 2 unresolved discussions (waiting on getina and stevenwarejones).

stevenwarejones (approved):
LGTM. All 4 prior session blockers resolved:

- `request_id` REQUIRED on Create + BatchCreate — the strongest possible fix to the retry-orphan hazard (without request_id callers can't reach the create path, making the lost-response-then-retry-with-fresh-fileResourceId scenario structurally impossible). Tests cover both rejection paths.
- Pagination predicate DP-scoped — no cross-DP leak.
- "Allows duplicate blob_uri" test renamed and now exercises the actual contract with two distinct request_ids.
- batchDelete duplicate-name error has a meaningful message body.

`fileResourceId` still uses `UUID.randomUUID()` rather than the injected `idGenerator`, but matches the established sibling pattern in `SpannerRawImpressionMetadataBatchFileService` on main. Worth a follow-up to address both files together; not blocking here.

Code is clean: 36 internal tests + 29 v1alpha tests, AIP-159 wildcard correctly handled, idempotency via the UNIQUE NULL_FILTERED INDEX, sibling-service exhaustive when blocks consistently extended.

Note: `Build and test` CI not running on this PR because the base is a stacked PR (`marcopremier/raw-impression-upload-api-impl`) rather than `main`. Will fire once the parent (#4013) lands. The schema DDL bug from #3989 is fixed in the inherited base.

Approving — land after #3989 and #4013.

stevenwarejones (approved):
> @stevenwarejones reviewed 28 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4014)* status: all files reviewed, 2 unresolved discussions (waiting on getina, Marco-Premier, and stevenwarejones+AGNT).

src/main/resources/edpaggregator/spanner/create-vid-labeling-schema.sql
stevenwarejones: Agreed — the schema file isn't in this PR's diff, and the fix lives in #3989 (which has now been rebased into this PR's base). Confirmed `WHERE DeleteTime IS NULL` no longer appears in the inherited schema. Closing.

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4014)* status: all files reviewed, 2 unresolved discussions (waiting on getina, Marco-Premier, and stevenwarejones).

stevenwarejones (approved):
> @stevenwarejones resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4014)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on getina).

SanjayVas (approved):
> @SanjayVas reviewed 30 files and all commit messages, and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4014)* status: all files reviewed, 2 unresolved discussions (waiting on getina and Marco-Premier).

___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/service/Errors.kt` line 458 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4014#-OvkkCUlCApGp3FXbxFr:-OvkkCUlCApGp3FXbxFs:b-etf0gw) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1b5a77f98c1371aa42ff84bf6a7caeeaa988a117/src/main/kotlin/org/wfanet/measurement/edpaggregator/service/Errors.kt#L458)):*
> ```Kotlin
>       get() = Errors.Reason.RAW_IMPRESSION_UPLOAD_NOT_FOUND
> 
>     override fun fromInternal(
> ```

nit: I don't think this is used
___
*[`src/main/proto/wfa/measurement/edpaggregator/v1alpha/raw_impression_upload_file_service.proto` line 77 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4014#-OvklOliCRrX7-JnChvV:-OvklOliCRrX7-JnChvW:b-n07jjr) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/1b5a77f98c1371aa42ff84bf6a7caeeaa988a117/src/main/proto/wfa/measurement/edpaggregator/v1alpha/raw_impression_upload_file_service.proto#L77)):*
> ```proto
>   // A request ID provided by the client to ensure idempotency.
>   //
>   // AIP-155 recommends that request IDs `should` be optional. This field
> ```

nit: This still shouldn't be required. It's the caller's choice whether to even implement retries.

This is a system API, so we can enforce this in our own calling code.

Marco-Premier (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4014)* status: 29 of 30 files reviewed, 1 unresolved discussion (waiting on getina, SanjayVas, and stevenwarejones).

stevenwarejones (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4014)* status: all files reviewed, 1 unresolved discussion (waiting on getina and SanjayVas).

___
*[`src/main/proto/wfa/measurement/edpaggregator/v1alpha/raw_impression_upload_file_service.proto` line 77 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4014#-OvklOliCRrX7-JnChvV:-OvqDaW185CGDi7mVfpl:budl5k) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/095880eceef5e5baa0b004685bc46084bc42ddc3/src/main/proto/wfa/measurement/edpaggregator/v1alpha/raw_impression_upload_file_service.proto#L77)):*

nit: This still shouldn't be required. It's the caller's choice whether to even implement retries.

This is a system API, so we can enforce this in our own calling code.
</blockquote></details>

Doesn't AIP 155 allow required request IDs?

src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/spanner/SpannerRawImpressionUploadFileService.kt
```diff
@@ -0,0 +1,528 @@
...
+    return result.copy {
+      createTime = commitTimestamp
+      updateTime = commitTimestamp
+    }
+  }
+
+  override suspend fun batchCreateRawImpressionUploadFiles(
+    request: BatchCreateRawImpressionUploadFilesRequest
+  ): BatchCreateRawImpressionUploadFilesResponse {
+    if (request.requestsList.isEmpty()) {
```
stevenwarejones, line 171: nit: This early-return bypasses the parent validations below. A caller passing `BatchCreate(empty list, parent="")` silently succeeds at the internal layer, even though the public layer rejects it (RawImpressionUploadFileService.kt:159-162). Either move the empty check after the two `isEmpty()` parent checks, or drop it entirely — an empty `forEachIndexed` plus an empty transaction is already a correct no-op.

stevenwarejones (commented):
Re-review nit.

src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/spanner/db/RawImpressionUploadFile.kt
```diff
@@ -0,0 +1,367 @@
...
+  fileResourceId: String,
+  blobUri: String,
+  createRequestId: String,
+) {
+  bufferInsertMutation("RawImpressionUploadFile") {
+    set("DataProviderResourceId").to(dataProviderResourceId)
+    set("RawImpressionUploadId").to(rawImpressionUploadId)
+    set("FileId").to(fileId)
+    set("FileResourceId").to(fileResourceId)
+    if (createRequestId.isNotEmpty()) {
```
stevenwarejones: nit: `createRequestId` is now REQUIRED at both service layers (SpannerRawImpressionUploadFileService.kt:95-98, RawImpressionUploadFileService.kt:97-100), so this `isNotEmpty()` guard is unreachable. Inline as `set("CreateRequestId").to(createRequestId)`. Same applies to the `nonEmptyRequestIds` filter at line 136 in this file.

stevenwarejones (commented):
Re-review nit.

SanjayVas (commented):
> @SanjayVas made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4014)* status: all files reviewed, 3 unresolved discussions (waiting on getina, Marco-Premier, and stevenwarejones).

___
*[`src/main/proto/wfa/measurement/edpaggregator/v1alpha/raw_impression_upload_file_service.proto` line 77 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4014#-OvklOliCRrX7-JnChvV:-Ovquwed9JfeLR-eox6Q:b-3vdq9m) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/095880eceef5e5baa0b004685bc46084bc42ddc3/src/main/proto/wfa/measurement/edpaggregator/v1alpha/raw_impression_upload_file_service.proto#L77)):*

Doesn't AIP 155 allow required request IDs?
</blockquote></details>

Strictly speaking it states that the field **should** be optional (rather than **must**). I'm making the same statement here - it should still be optional even in this case IMO.

This is just an optional nit though. I don't feel terribly strongly.

src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/spanner/db/RawImpressionUploadFile.kt
```diff
@@ -0,0 +1,367 @@
...
+/** Buffers an insert mutation for a [RawImpressionUploadFile] row. */
+fun AsyncDatabaseClient.TransactionContext.insertRawImpressionUploadFile(
+  fileId: Long,
+  dataProviderResourceId: String,
+  rawImpressionUploadId: Long,
+  fileResourceId: String,
+  blobUri: String,
+  createRequestId: String,
+) {
+  bufferInsertMutation("RawImpressionUploadFile") {
```
stevenwarejones, line 220: Companion to gh-3462699369 on #4009: the public proto now has `size_bytes` REQUIRED, but it never round-trips through Spanner because:

1. The internal proto (`internal/edpaggregator/raw_impression_upload_file.proto`) has no `size_bytes` field.
2. The schema has no `SizeBytes` column on `RawImpressionUploadFile` (will be added in #3989 — see companion ask there).
3. This Spanner service (Db.kt `insertRawImpressionUploadFile`, `buildRawImpressionUploadFileResult`) doesn't write or read the column.
4. The v1alpha service doesn't validate `request.rawImpressionUploadFile.sizeBytes` is set before forwarding (silently fits `0` through).

Fix:
- **Internal proto**: add `int64 size_bytes = 8;` to `RawImpressionUploadFile` (next field number after `delete_time = 7`).
- **Db.kt insertRawImpressionUploadFile**: take `sizeBytes: Long` parameter, `set("SizeBytes").to(sizeBytes)`.
- **Db.kt buildRawImpressionUploadFileResult**: populate `sizeBytes = struct.getLong("SizeBytes")`.
- **SpannerRawImpressionUploadFileService**: forward `file.sizeBytes` from request to the insert.
- **v1alpha RawImpressionUploadFileService**: validate `request.rawImpressionUploadFile.sizeBytes != 0` (or `> 0` — caller decides whether 0 is genuinely empty or unset; current proto KDoc says 0 means genuinely empty, but the validation should reject unset proto-default if so).

Otherwise `VidRankBuilder.binPackFiles` (#4009) reads `0` for every file and bin-packs everything into a single VidLabelingJob — the exact failure mode that motivated making the field REQUIRED.

Standing rule: as long as #3989 is open, all SQL changes go there. The schema column add is asked on #3989 (gh-... companion).

stevenwarejones (commented):
size_bytes plumbing missing.

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4014)* status: all files reviewed, 4 unresolved discussions (waiting on getina, Marco-Premier, and stevenwarejones).

src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/spanner/db/RawImpressionUploadFile.kt
Marco-Premier: Done.

Marco-Premier (commented):
> @Marco-Premier made 1 comment and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4014)* status: 27 of 32 files reviewed, 2 unresolved discussions (waiting on getina, SanjayVas, stevenwarejones, and stevenwarejones+AGNT).

stevenwarejones (commented):
> @stevenwarejones reviewed 3 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4014)* status: 30 of 32 files reviewed, 2 unresolved discussions (waiting on getina, SanjayVas, and stevenwarejones+AGNT).

___
*[`src/main/proto/wfa/measurement/edpaggregator/v1alpha/raw_impression_upload_file_service.proto` line 77 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4014#-OvklOliCRrX7-JnChvV:-OvuT_CV88wz30XVD9h4:b3das8q) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/095880eceef5e5baa0b004685bc46084bc42ddc3/src/main/proto/wfa/measurement/edpaggregator/v1alpha/raw_impression_upload_file_service.proto#L77)):*

Strictly speaking it states that the field **should** be optional (rather than **must**). I'm making the same statement here - it should still be optional even in this case IMO.

This is just an optional nit though. I don't feel terribly strongly.
</blockquote></details>

the vid labeler in the edp aggregator is a state machine with each call potentially flipping important states along the way. I think for this use case, we should make them required.

src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/spanner/SpannerRawImpressionUploadFileService.kt
stevenwarejones: Verified in commit 83e79036. `request.requestsList.isEmpty()` early-return now sits AFTER the two parent validations at SpannerRawImpressionUploadFileService.kt:163-170 — a caller passing `BatchCreate(empty list, parent="")` now correctly fails fast with `INVALID_ARGUMENT` instead of silently succeeding. Closing.

src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/spanner/db/RawImpressionUploadFile.kt
stevenwarejones: Verified in commit 83e79036 — full size_bytes plumbing landed:

- Internal proto: `int64 size_bytes = 8;` added to `RawImpressionUploadFile`.
- Public proto: same field as REQUIRED + IMMUTABLE.
- v1alpha service: validates `sizeBytes == 0L` → INVALID_ARGUMENT on both `createRawImpressionUploadFile` (line 97-100) and `batchCreateRawImpressionUploadFiles` per-element (line 200-205), forwards to internal.
- Spanner service: forwards `file.sizeBytes` on both Create and BatchCreate paths.
- Db.kt: `insertRawImpressionUploadFile` takes `sizeBytes: Long` and `set("SizeBytes").to(sizeBytes)`; `buildRawImpressionUploadFileResult` reads `struct.getLong("SizeBytes")`; BASE_SQL selects the column.

Value now round-trips through Spanner. Pending the `SizeBytes INT64 NOT NULL` column add on #3989 (gh-3468267360) and the dispatcher's blob-listing fix on #3910 (gh-3468405104). Closing here.

src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/spanner/db/RawImpressionUploadFile.kt
stevenwarejones: Verified in commit 83e79036. Db.kt:217-218 inlines `set("CreateRequestId").to(createRequestId)` (no more isNotEmpty guard), and `findExistingUploadFilesByRequestIds` at line 137 dropped the `nonEmptyRequestIds` filter. Closing.

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4014)* status: 27 of 32 files reviewed, 2 unresolved discussions (waiting on getina, SanjayVas, and stevenwarejones).

stevenwarejones (approved):
> @stevenwarejones reviewed 2 files and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4014)* status: all files reviewed, 1 unresolved discussion (waiting on getina and SanjayVas).

Marco-Premier (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4014)* status: 29 of 32 files reviewed, all discussions resolved (waiting on getina, SanjayVas, and stevenwarejones).

Marco-Premier (commented):
> @Marco-Premier reviewed 31 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4014)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on getina, SanjayVas, and stevenwarejones).

Marco-Premier (commented):
> @Marco-Premier reviewed 34 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4014)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on getina, SanjayVas, and stevenwarejones).

PR #4016, opened 2026-06-17
feat: rank-index storage and in-heap rank allocator

Marco-Premier:
# SUMMARY

Introduces the storage and in-memory primitives the Phase-1 ranker is built on. No orchestration or TEE-app logic — pure data structures, codecs, encrypted I/O, and the on-disk format.

  ## What's included

  **On-disk format — `rank_index_map.proto`**
  `RankIndexMap` is the serialized per-subpool rank map, used for both `SNAPSHOT` and `DAY_ONLY` blobs. Layout is two positionally-paired packed arrays plus a recency array: `fingerprints` (12-byte
  stride), `ranks` (`fixed32`), and `last_seen_days` (`uint16` big-endian, 2-byte stride). The compact `uint16` encoding for `last_seen` halves wire + heap cost vs `fixed32` (≈720 MB vs 1.4 GB of recency
  state at US 360 M scale) and covers epoch days through 2149.

  **Encrypted RecordIO storage**
  - `EncryptedRecordIoStore` — shared base providing DEK generation and the envelope-encrypted RecordIO client plumbing.
  - `RankIndexStore` — reads/writes a blob as a **stream of `RankIndexMap` records** (a 1.5 B-entry cumulative snapshot far exceeds protobuf's ~2 GB message cap, so it cannot be a single message).
  Returns the SHA-256 of the plaintext payload for the `RankIndexBlob.blob_checksum` corruption guard, and re-validates it on read. `SubpoolFingerprintsStore` is refactored onto the shared base
  (API-preserving).

  **In-heap primitives**
  - `RankAllocator` — single-subpool allocation state: the cumulative `(fingerprint → rank)` map (`Bytes12IntMap`), a `BitSet` of taken ranks rebuilt from the map at load, and a `ShortArray` `last_seen`
  retention index. Supports load-from-snapshot, aged-rank freeing, allocate/renew, and streaming the cumulative / day-only records back out in bounded-memory chunks.
  - `EventIdDigestBytes` / `LastSeenDayBytes` — big-endian read/write codecs for the 12-byte fingerprint and `uint16` epoch-day packed formats.

  ## Testing
  `RankAllocatorTest`, `EventIdDigestBytesTest`, `RankIndexStoreTest`; the existing `SubpoolFingerprintsStoreTest` continues to pass after the base refactor.

Issue: #3921

Issue #3921, opened 2026-06-05
Implement VidRankBuilder Phase-1 TEE application


Marco-Premier:
Implement the VidRankBuilder Phase-1 TEE application under `src/main/kotlin/org/wfanet/measurement/edpaggregator/vidrankbuilder/`.

The SubpoolAssigner's last-shard-out publishes one `WorkItem` per `RankerJob` row for each (`RawImpressionUpload`, `ModelLine`) pair with memoization enabled. A `RankerJob` may cover one or more subpools (Phase-0 bin-packs small subpools into a single VM while a large subpool gets its own). Each VM:
- unwraps `VidRankBuilderParams.encrypted_subpool_maps_dek` via the DataProvider's KMS to recover the plaintext DEK Phase-0 used to encrypt the `SubpoolFingerprints` blobs,
- for each `(subpoolId, blobUri)` in `subpool_map_blob_uris`:
  - loads the prior cumulative rank-index blob from `vid_rank_map_storage_params` and rebuilds the in-heap `Bytes12IntMap` (12-byte fingerprint → rank) + rank `BitSet`,
  - prunes aged-out rank-index blobs per the retention policy (`MaxEventDate < today − RETENTION_DAYS`) and frees the corresponding bits,
  - reads and decrypts the SubpoolAssigner's `SubpoolFingerprints` blob at `blobUri`,
  - allocates ranks to new fingerprints via `bitSet.nextClearBit(cursor)`, capping at `ranked_size` (overflow falls back to the unranked path at Phase-2),
  - writes the new day-only + cumulative rank-index blobs (DEK-encrypted) to `vid_rank_map_storage_params` and inserts the matching `RankIndexBlob` rows via the EDP Aggregator internal gRPC,
- flips its `RankerJob` row from `RANKING` to `RANKER_SUCCEEDED`.

Last-`RankerJob`-out for `(RawImpressionUpload, ModelLine)` flips `RawImpressionUploadModelLineState` from `RANKING` to `LABELING` and publishes one VidLabeler WorkItem per shard (`VidRankBuilderParams.total_shards` controls the fan-out).

Scaffold and `vid_rank_builder_params.proto` landed by PR #<TBD>.

Issue #4011, opened 2026-06-16
Memoized VID pipeline: support out-of-window backfills and historical event-date reports via anchored cumulative SNAPSHOTs


stevenwarejones:
## Problem

The current memoized VID pipeline (Phase 0 #3985, Phase 1 #4009) maintains a **single global cumulative `RankIndexMap` SNAPSHOT** per `(EDP, model line, subpool)`. This is the right shape for "rolling-window reports run from today" but breaks two distinct use cases:

### Failure mode A — backfills with `max_event_date < today - RETENTION_DAYS`

A late-arriving upload (e.g. impressions for an event date 6 months old, retention = 90 days):

1. `SubpoolRanker.rank()` loads the current cumulative SNAPSHOT.
2. Backfilled fingerprints get assigned/renewed ranks. The `last_seen` field on those entries is stamped to `eventDay = backfill's max event date`, which is below `today - RETENTION_DAYS`.
3. The very next forward dispatch runs `freeAgedRanks(cutoffEpochDay = today - 90)`. Every rank touched by the backfill is now `last_seen < cutoff` → **freed immediately**.
4. A subsequent dispatch reassigns those freed slots to different fingerprints. The DAY_ONLY blob written by the backfill in step 2 references `R_F`, but `R_F` now belongs to a different fingerprint. **Labels for the backfilled events become incorrect (point at the wrong fingerprint's VID slot, or at a freed-and-unreassigned slot).**

### Failure mode B — historical reports

A report targeting an event-date window earlier than the current SNAPSHOT (e.g., "Mar–Jun 2025 reach, run in Jan 2026"):

1. Labeler loads the current SNAPSHOT (built from all data through today).
2. Cookie X observed in Mar got rank `R_F`. Cookie value churned (different user); reassigned in Oct as rank `R_F'` after `R_F` was freed by retention.
3. SNAPSHOT today contains X → `R_F'`. Mar events of X label as `VID(R_F')` — wrong VID, wrong reach.

This is fundamental to single-cumulative storage: there's no record of "what the rank for X was as of Mar."

## Scope of breakage

Failure mode A is a **correctness bug** that surfaces any time backfill latency exceeds `RETENTION_DAYS`. Probably bounded operationally for in-pipeline EDPs but is silent corruption when it happens.

[...]

PR #4017, opened 2026-06-17
feat: subpool ranker, retention, and metrics

Marco-Premier:
# SUMMARY

Adds the per-subpool ranking domain logic that drives one subpool through a dispatch, built on the storage + allocator primitives from #4016.

  ## What's included
  
  **`SubpoolRanker`** — ranks a single subpool end-to-end:
  1. **Idempotency gate** — if a `SNAPSHOT` row already exists for this `(upload, subpool)`, the subpool is already done; skip.
  2. **Load prior state** — read the prior cumulative `SNAPSHOT` (checksum-validated) into a `RankAllocator`, rebuilding the rank `BitSet` and `last_seen` index.
  3. **Materialize today's fingerprints** from the Phase-0 merged blob.
  4. **Retention** — GC aged-out `DAY_ONLY` blobs, then free aged ranks in-heap (`last_seen < cutoff`), **excluding** fingerprints seen this dispatch so their VIDs stay stable.
  5. **Allocate / renew** ranks for today's fingerprints (renewal preserves the existing rank → stable VID; overflow past `ranked_size` falls back to the unranked path).
  6. **Persist** the new `DAY_ONLY` + `SNAPSHOT` blobs and rows under per-attempt-unique keys so a concurrent or redelivered attempt never overwrites another's bytes.

  **`SubpoolRetention`** — deletes `DAY_ONLY` blobs older than the retention window (storage GC, independent of in-heap rank freeing).

  **`VidRankBuilderMetrics`** — OpenTelemetry counters for allocated / renewed / overflow / freed / cumulative size.

  ## Design notes
  Retention is an in-heap pass over the cumulative map's `last_seen` index rather than a re-scan of every surviving `DAY_ONLY` blob. Renewal keeps already-ranked fingerprints on their existing rank so
  VIDs are stable across dispatches.

  ## Testing
  `SubpoolRankerTest` (cold subpool, warm renew + allocate, retention free-and-reuse, overflow, idempotency-skip) and `SubpoolRetentionTest`.

Issue: #3921

### DIFF
--- a/src/main/kotlin/org/wfanet/measurement/edpaggregator/vidrankbuilder/SubpoolRanker.kt
+++ b/src/main/kotlin/org/wfanet/measurement/edpaggregator/vidrankbuilder/SubpoolRanker.kt
@@ -17,6 +17,7 @@
 package org.wfanet.measurement.edpaggregator.vidrankbuilder
 
 import com.google.type.Date
+import com.google.type.date
 import java.nio.ByteBuffer
 import java.security.MessageDigest
 import java.time.LocalDate
@@ -44,19 +45,78 @@ import org.wfanet.measurement.edpaggregator.vidlabeler.utils.Bytes12IntMap
  * `RankerJob` covers.
  *
  * For one subpool it:
- * 1. **idempotency gate** — if a `SNAPSHOT` `RankIndexBlob` row already exists for this (upload,
- *    subpool), a prior attempt already committed its blobs and rows atomically, so the subpool is
- *    done and is skipped (cheap re-delivery no-op),
- * 2. loads the prior cumulative `SNAPSHOT` (if any) into a [RankAllocator] (rebuilding its rank
- *    `BitSet` and per-rank `last_seen`), validating the blob checksum,
- * 3. materializes this dispatch's fingerprint set from the Phase-0 merged `SubpoolFingerprints`
+ * 1. loads the prior cumulative `SNAPSHOT` (the most recent by `create_time` across all uploads —
+ *    the current cumulative state) into a [RankAllocator] (rebuilding its rank `BitSet` and
+ *    per-rank `last_seen`), validating the blob checksum,
+ * 2. decides whether this dispatch is a **backfill** (see "Backfill" below),
+ * 3. **idempotency gate** — if the row this run would commit already exists for this (upload,
+ *    subpool), a prior attempt already finished, so the subpool is skipped (cheap re-delivery
+ *    no-op). The gate row is the `SNAPSHOT` on a normal dispatch and the `DAY_ONLY` on a backfill
+ *    (a backfill writes no `SNAPSHOT` — see below),
+ * 4. materializes this dispatch's fingerprint set from the Phase-0 merged `SubpoolFingerprints`
  *    blob,
- * 4. garbage-collects aged-out `DAY_ONLY` blobs ([SubpoolRetention]) and frees aged-out ranks in
+ * 5. garbage-collects aged-out `DAY_ONLY` blobs ([SubpoolRetention]) and frees aged-out ranks in
  *    heap via the `last_seen` index, excluding fingerprints seen this dispatch,
- * 5. allocates ranks to today's fingerprints (renewing already-ranked ones, capping at
- *    `ranked_size`),
- * 6. writes the new `DAY_ONLY` + `SNAPSHOT` blobs (DEK-encrypted) under **per-attempt-unique keys**
- *    and records both as `RankIndexBlob` rows.
+ * 6. allocates ranks to today's fingerprints (renewing already-ranked ones, capping at
+ *    `ranked_size`); on a backfill it instead replays each fingerprint's historical rank (see
+ *    below),
+ * 7. writes the new `DAY_ONLY` blob (DEK-encrypted) and — on a normal dispatch only — the new
+ *    cumulative `SNAPSHOT`, recording each as a `RankIndexBlob` row.
+ *
+ * ## Backfill (Problem 1 — give a person back their original rank)
+ *
+ * The pipeline keeps a single live cumulative snapshot per (EDP, model line, subpool). That is
+ * correct for forward data appended in chronological order, but a **backfill** — an upload whose
+ * newest event date ([maxEventDate]) is older than the most recent data already ranked — can land a
+ * fingerprint whose original rank was freed and re-handed to someone else by the time the latest
+ * snapshot was written. Giving it a brand-new rank then splits one person into two VIDs (reach
+ * overcount).
+ *
+ * A dispatch is treated as a backfill when **all** hold:
+ * * the latest cumulative `SNAPSHOT` carries a `max_event_date` (the snapshot-selection key) and
+ *   that date is strictly newer than this dispatch's [maxEventDate] — i.e. fresher data already
+ *   exists, so this is not forward append, and
+ * * the gap (`latest.max_event_date - this.max_event_date`) is within [retentionDays] — a dispatch
+ *   older than the retention window has no live ranks left to reclaim, so reassigning is pointless
+ *   (the residual corruption there is Problem 3, explicitly out of scope), and
+ * * an **old snapshot** exists: the `SNAPSHOT` whose `max_event_date` is the greatest value
+ *   strictly less than this dispatch's [maxEventDate] — the cumulative state as of just before the
+ *   backfilled day. If none exists (the data provider uploaded data older than anything on record),
+ *   it is not a backfill and the normal path runs.
+ *
+ * On a backfill, both the latest cumulative (already loaded into the allocator) and that old
+ * snapshot's `(fp -> rank)` map are held in heap, and each of today's fingerprints is assigned:
+ * 1. if present in the **old** snapshot -> its rank there (give the person back their original
+ *    rank),
+ * 2. else if present in the **latest** snapshot -> the rank it holds there,
+ * 3. else -> a freshly allocated rank.
+ *
+ * ### Backfill writes only a `DAY_ONLY` blob — never a new `SNAPSHOT`
+ *
+ * A backfill deliberately does **not** write a new cumulative `SNAPSHOT`. The prior-snapshot lookup
+ * ([findPriorSnapshot]) selects the latest snapshot by `create_time`, and a backfill run has the
+ * newest `create_time` but carries old data; persisting its cumulative would make the next forward
+ * dispatch load this stale, older-day cumulative as its baseline and corrupt the forward chain.
+ * Phase-2 labels each upload from that upload's own `DAY_ONLY` blob (the corrected `(fp -> rank)`
+ * for exactly this dispatch's fingerprints), so the backfilled day is labeled correctly without a
+ * snapshot. The idempotency gate therefore keys off the `DAY_ONLY` row on a backfill.
+ *
+ * ### Monitoring (the accepted undercount)
+ *
+ * When a fingerprint's old rank was already re-handed to a *different* fingerprint (the old slot is
+ * taken by "Bob" in the latest snapshot), reusing the old rank makes the two share a VID for the
+ * backfilled day (reach undercount). We still reuse the old rank (one-person-one-VID is the goal)
+ * but **count** these: [Result.backfillRankCollisions] is the number of backfilled fingerprints
+ * that reused an old rank while holding a *different* rank in the latest snapshot, so the size of
+ * the undercount is observable in metrics and logs.
+ *
+ * ## Memory
+ *
+ * A backfill holds **two** in-heap maps for the subpool — the latest cumulative and the old
+ * snapshot. At the India worst case (~1.5 B fingerprints) a single cumulative already dominates
+ * heap, so loading a second can OOM an `n2d-highmem-16`; backfill-capable rankers should run on a
+ * larger machine type (e.g. `n2d-highmem-32`+). Normal (non-backfill) dispatches hold only one map,
+ * as before.
  *
  * ## Idempotency & concurrency
  *
@@ -68,9 +128,9 @@ import org.wfanet.measurement.edpaggregator.vidlabeler.utils.Bytes12IntMap
  * write-before-CAS clobber hazard without needing a write-if-absent storage primitive.
  *
  * @param subpoolFingerprintsStore reads the Phase-0 merged `SubpoolFingerprints` blob.
- * @param rankIndexStore reads the prior snapshot and writes the new rank-index blobs.
+ * @param rankIndexStore reads the prior/old snapshots and writes the new rank-index blobs.
  * @param rankIndexBlobsStub metadata-storage service for the idempotency gate, locating the prior
- *   snapshot, and inserting the new blob rows.
+ *   and old snapshots, and inserting the new blob rows.
  * @param retention garbage-collects aged-out `DAY_ONLY` blobs.
  * @param dataProvider EDP resource name (`dataProviders/{dp}`), parent of the upload wildcard for
  *   the prior-snapshot lookup.
@@ -79,8 +139,9 @@ import org.wfanet.measurement.edpaggregator.vidlabeler.utils.Bytes12IntMap
  * @param vidRankMapBlobPrefix static blob prefix for the vid-rank-map bucket.
  * @param kekUri KEK URI used to wrap each new blob's DEK.
  * @param encryptedSubpoolMapsDek DEK that decrypts the Phase-0 merged blobs.
- * @param maxEventDate newest event date this dispatch covers; stamped as `last_seen` and on the
- *   `DAY_ONLY` row.
+ * @param maxEventDate newest event date this dispatch covers; stamped as `last_seen`, on the
+ *   `DAY_ONLY` row, and on the `SNAPSHOT` row (the snapshot-selection key), and used for backfill
+ *   detection.
  * @param retentionDays retention window in days (must exceed the max measurement-report window).
  * @param today the UTC date treated as "now" for the rank-age cutoff.
  * @param metrics OpenTelemetry instruments.
@@ -109,6 +170,10 @@ class SubpoolRanker(
    * @property freed ranks released by the retention pass.
    * @property cumulativeSize total ranked fingerprints after this dispatch.
    * @property skipped whether the subpool was already ranked for this upload (idempotency gate).
+   * @property backfill whether this dispatch was handled as a backfill.
+   * @property backfillReusedOldRank backfilled fingerprints given back their old-snapshot rank.
+   * @property backfillRankCollisions subset of [backfillReusedOldRank] whose old rank differs from
+   *   the rank they hold in the latest snapshot (the accepted reach-undercount sizing).
    */
   data class Result(
     val poolOffset: Long,
@@ -118,25 +183,37 @@ class SubpoolRanker(
     val freed: Long,
     val cumulativeSize: Long,
     val skipped: Boolean = false,
+    val backfill: Boolean = false,
+    val backfillReusedOldRank: Long = 0L,
+    val backfillRankCollisions: Long = 0L,
   )
 
   /** Ranks [poolOffset]; [subpoolBlobUri] is its Phase-0 merged blob; caps at [rankedSize]. */
   suspend fun rank(poolOffset: Long, subpoolBlobUri: String, rankedSize: Int): Result {
-    // Idempotency gate: an existing SNAPSHOT row for this (upload, subpool) means a prior attempt
-    // committed both blobs + rows atomically, so the subpool is already done — skip the re-rank.
-    if (findUploadSnapshot(poolOffset) != null) {
+    val eventDay = epochDayOf(maxEventDate)
+    val cutoffEpochDay = (today.toEpochDay() - retentionDays).toInt()
+
+    // 1. Locate the latest cumulative snapshot (most recent create_time), then decide whether this
+    // dispatch is a backfill and, if so, which older snapshot to replay ranks from.
+    val priorSnapshot = findPriorSnapshot(poolOffset)
+    val oldSnapshot = resolveBackfillOldSnapshot(poolOffset, eventDay, priorSnapshot)
+    val isBackfill = oldSnapshot != null
+
+    // 2. Idempotency gate, keyed on the row this run commits: SNAPSHOT on a normal dispatch,
+    // DAY_ONLY on a backfill (which writes no SNAPSHOT). A non-null result means a prior attempt
+    // already finished this (upload, subpool), so skip the re-rank.
+    val gateType =
+      if (isBackfill) RankIndexBlob.BlobType.DAY_ONLY else RankIndexBlob.BlobType.SNAPSHOT
+    if (findUploadBlob(poolOffset, gateType) != null) {
       logger.info(
         "Subpool $poolOffset for $modelLine already ranked for $rawImpressionUpload; skipping"
       )
       return Result(poolOffset, 0, 0, 0, 0, 0, skipped = true)
     }
 
-    val eventDay = epochDayOf(maxEventDate)
-    val cutoffEpochDay = (today.toEpochDay() - retentionDays).toInt()
     val allocator = RankAllocator(poolOffset, rankedSize, eventDay)
 
-    // 1. Load the prior cumulative snapshot (validating its checksum), if one exists.
-    val priorSnapshot = findPriorSnapshot(poolOffset)
+    // 3. Load the latest cumulative snapshot (validating its checksum), if one exists.
     if (priorSnapshot != null) {
       allocator.loadFrom(
         rankIndexStore.readBlob(
@@ -147,7 +224,7 @@ class SubpoolRanker(
       )
     }
 
-    // 2. Materialize this dispatch's fingerprint set for the subpool from the Phase-0 merged blob.
+    // 4. Materialize this dispatch's fingerprint set for the subpool from the Phase-0 merged blob.
     val todayFps = Bytes12IntMap()
     subpoolFingerprintsStore.readBlob(subpoolBlobUri, encryptedSubpoolMapsDek).collect { record ->
       val fps = record.fingerprints
@@ -163,30 +240,44 @@ class SubpoolRanker(
       }
     }
 
-    // 3a. Storage GC of aged-out DAY_ONLY blobs (independent of rank freeing).
+    // 5a. Storage GC of aged-out DAY_ONLY blobs (independent of rank freeing).
     retention.deleteAgedBlobs(poolOffset)
-    // 3b. Free aged-out ranks in heap from the `last_seen` index, before allocating so freed slots
+    // 5b. Free aged-out ranks in heap from the `last_seen` index, before allocating so freed slots
     // are reusable today. Fingerprints seen this dispatch are excluded so their ranks (VIDs)
     // survive.
     allocator.freeAgedRanks(cutoffEpochDay, todayFps)
 
-    // 4. Allocate / renew ranks for today's fingerprints.
-    todayFps.forEach { keyHi, keyLo, _ -> allocator.assign(keyHi, keyLo) }
+    // 6. Assign ranks to today's fingerprints.
+    var reusedOldRank = 0L
+    var rankCollisions = 0L
+    if (oldSnapshot != null) {
+      // Backfill: replay each fingerprint's historical rank (old -> latest -> new).
+      val oldRanks = loadFingerprintRanks(oldSnapshot)
+      todayFps.forEach { keyHi, keyLo, _ ->
+        val oldRank = oldRanks.get(keyHi, keyLo)
+        if (oldRank != Bytes12IntMap.NOT_PRESENT && oldRank in 0 until rankedSize) {
+          // The fingerprint's rank in the latest snapshot, captured before we overwrite it; a
+          // different value means the old slot was re-handed to someone else (accepted undercount).
+          val latestRank = allocator.get(keyHi, keyLo)
+          allocator.assignAt(keyHi, keyLo, oldRank)
+          reusedOldRank++
+          if (latestRank != Bytes12IntMap.NOT_PRESENT && latestRank != oldRank) {
+            rankCollisions++
+          }
+        } else {
+          allocator.assign(keyHi, keyLo)
+        }
+      }
+    } else {
+      todayFps.forEach { keyHi, keyLo, _ -> allocator.assign(keyHi, keyLo) }
+    }
 
-    // 5. Write the new DAY_ONLY + SNAPSHOT blobs and record both as RankIndexBlob rows. Keys carry
-    // a
-    // per-attempt UUID so a concurrent/re-delivered attempt never overwrites another's bytes; the
-    // winning (deterministic-request_id) row's blob_uri + DEK are always self-consistent.
+    // 7. Write the new blob(s) and record the RankIndexBlob row(s). Keys carry a per-attempt UUID
+    // so
+    // a concurrent/re-delivered attempt never overwrites another's bytes; the winning
+    // (deterministic-request_id) row's blob_uri + DEK are always self-consistent.
     val attemptId = UUID.randomUUID().toString()
     val dek = rankIndexStore.generateDek(kekUri)
-    val snapshotKey =
-      RankIndexStore.snapshotKey(
-        vidRankMapBlobPrefix,
-        rawImpressionUpload,
-        modelLine,
-        poolOffset,
-        attemptId,
-      )
     val dayOnlyKey =
       RankIndexStore.dayOnlyKey(
         vidRankMapBlobPrefix,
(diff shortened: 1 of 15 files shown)


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4017)

src/main/kotlin/org/wfanet/measurement/edpaggregator/vidrankbuilder/SubpoolRanker.kt
```diff
@@ -0,0 +1,363 @@
...
+        ResourceList(response.rankIndexBlobsList, response.nextPageToken)
+      }
+      .collect { page -> page.firstOrNull()?.let { existing = it } }
+    return existing
+  }
+
+  /**
+   * The newest non-deleted `SNAPSHOT` blob for [poolOffset] of this (data provider, model line)
+   * across all uploads, or `null` on a cold (Day-1) subpool. The most recent `create_time` is the
+   * current cumulative state (the design's N−1 recovery baseline).
```
stevenwarejones: Backfill safety — out-of-window backfills are in scope, and the **single-cumulative-SNAPSHOT model can't label backfilled events correctly even within retention**. A backfill for event date E < today needs the rank state in force at E; today's live cumulative is post-churn (ranks reassigned, fingerprints renewed) and silently mis-labels every fingerprint whose rank changed since E.

Concrete fix — keep **two cumulatives**, no DAY_ONLY on the labeling read path:

1. **Live cumulative** — state at end-of-today (what this PR already produces; cold Phase-1 unchanged).
2. **Anchor cumulative** — state at end of `today - RETENTION_DAYS - 1` (one day before the oldest retained DAY_ONLY). Rolls forward only at retention-boundary refresh.

To label backfill at E in (anchor, today]: read both, look up F in each.
- In both at the same rank → label that rank (stable through retention, exact).
- In both at different ranks → label anchor's rank (pre-churn).
- Anchor only → label anchor's rank.
- Live only → label live's rank.
- Neither → allocate fresh.

Storage ~2x. Cold Phase-1 load stays single-blob. Backfill is a two-blob read, no daily-delta scan. Out-of-window backfills (event-date older than the anchor) still need #4011's anchored-N model.

Posting on `findPriorSnapshot` because this method is the entry point into the cumulative model; the fix is in the data-model shape, not in this function. Should be settled before #4017 lands; happy to amend #4011 to record the two-cumulative shape.

stevenwarejones (commented):
Posting one design-scope item — backfill safety with the single-cumulative-SNAPSHOT model.

stevenwarejones (commented):
findUploadSnapshot idempotency gate.

src/main/kotlin/org/wfanet/measurement/edpaggregator/vidrankbuilder/SubpoolRanker.kt
```diff
@@ -0,0 +1,363 @@
...
+                  blobType = RankIndexBlob.BlobType.SNAPSHOT
+                  cmmsModelLine = modelLine
+                  this.poolOffset = poolOffset
+                }
+              this.pageToken = pageToken
+            }
+          )
+        ResourceList(response.rankIndexBlobsList, response.nextPageToken)
+      }
+      .collect { page -> page.firstOrNull()?.let { existing = it } }
```
stevenwarejones: The idempotency-gate logic is structurally wrong. `.collect { page -> page.firstOrNull()?.let { existing = it } }` captures "the first element of the LAST page that had any results," not "the first match." Today it only works because the filter is *expected* to return ≤1 row — but if duplicates ever exist (race conditions, partial idempotency upstream, operator backfill), this silently picks an arbitrary winner.

Also: every dispatch walks ALL pages even though the answer is binary (exists/not). For an idempotency gate called on every retry, this scans the whole filter result wastefully.

Fix both in one pass — short-circuit + bound the page size:

```kotlin
private suspend fun findUploadSnapshot(poolOffset: Long): RankIndexBlob? {
  rankIndexBlobsStub
    .listResources { pageToken: String ->
      val response =
        listRankIndexBlobs(
          listRankIndexBlobsRequest {
            parent = rawImpressionUpload
            filter =
              ListRankIndexBlobsRequestKt.filter {
                blobType = RankIndexBlob.BlobType.SNAPSHOT
                cmmsModelLine = modelLine
                this.poolOffset = poolOffset
              }
            pageSize = 1
            this.pageToken = pageToken
          }
        )
      ResourceList(response.rankIndexBlobsList, response.nextPageToken)
    }
    .collect { page ->
      page.firstOrNull()?.let { return@findUploadSnapshot it }
    }
  return null
}
```

Aside: this is a smell about the API shape — needing `List+filter+pageSize=1` to answer "does this single (upload, model line, subpool, SNAPSHOT) row exist" is messy. A dedicated `GetRankIndexBlob` by `(upload, model line, subpool, blob type)` or an `existsRankIndexBlob` RPC would be the right primitive. Worth tracking as a follow-up on the API side.

stevenwarejones (commented):
Silent default for required maxEventDate.

src/main/kotlin/org/wfanet/measurement/edpaggregator/vidrankbuilder/SubpoolRanker.kt
```diff
@@ -0,0 +1,363 @@
...
+    metrics.ranksAllocatedCounter.add(allocator.allocated)
+    metrics.ranksRenewedCounter.add(allocator.renewed)
+    metrics.ranksFreedCounter.add(allocator.freed)
+    metrics.overflowFingerprintsCounter.add(allocator.overflow)
+    metrics.subpoolsRankedCounter.add(1)
+  }
+
+  /** Epoch-day of a UTC [Date]; falls back to [today] when the date is unset. */
+  private fun epochDayOf(date: Date): Int {
+    if (date.year == 0) return today.toEpochDay().toInt()
```
stevenwarejones: Silent default for an unset required field. `maxEventDate` is declared REQUIRED in `vid_rank_builder_params.proto:161` and is the constructor parameter at line 97. But `epochDayOf` silently substitutes `today.toEpochDay()` when `date.year == 0`, which:

1. Violates the "no defaults for critical items" rule. `maxEventDate` drives both `last_seen` stamping (every fingerprint touched this dispatch) and the `MaxEventDate` on each `DAY_ONLY` row (which retention uses to age out blobs). Wrong value → wrong rank-freeing → wrong VIDs.
2. Defeats Phase-0's contract. If Phase-0 ever ships an unset `maxEventDate` (e.g., a shard with `unrouted == labeled`, see `SubpoolAssigner.kt:182` where `maxTimestampUsec?.let { ... }` is conditional), Phase-1 should fail loudly so operators notice — not silently corrupt rank state with `today`.

Fix — require non-unset:
```kotlin
private fun epochDayOf(date: Date): Int {
  require(date.year != 0) { "maxEventDate must be set on VidRankBuilderParams" }
  return LocalDate.of(date.year, date.month, date.day).toEpochDay().toInt()
}
```

Better: validate at `VidRankBuilderApp.runWork` alongside the other `require(...)` checks at lines 119-123 so we fail before doing any work. The default belongs in neither place.

src/main/kotlin/org/wfanet/measurement/edpaggregator/vidrankbuilder/SubpoolRanker.kt
```diff
@@ -0,0 +1,363 @@
...
+   *   latest-snapshot lookup and bound SNAPSHOT retention.
+   */
+  private suspend fun findPriorSnapshot(poolOffset: Long): RankIndexBlob? {
+    var latest: RankIndexBlob? = null
+    rankIndexBlobsStub
+      .listResources { pageToken: String ->
+        val response =
+          listRankIndexBlobs(
+            listRankIndexBlobsRequest {
+              parent = "$dataProvider/rawImpressionUploads/-"
```
stevenwarejones: Hardcoded AIP-159 wildcard string. Use the named constant:

```kotlin
parent = RawImpressionUploadKey(dataProvider, ResourceKey.WILDCARD_ID).toName()
```

Or, since the parser is loose, just `"$dataProvider/rawImpressionUploads/${ResourceKey.WILDCARD_ID}"`. The named constant documents intent (a reader otherwise has to recognize `-` as a wildcard rather than a literal segment) and won't silently break if the wildcard convention changes. Sibling code uses it — see `RawImpressionUploadFileService.kt:332` in #4014.

stevenwarejones (commented):
Hardcoded wildcard string.

stevenwarejones (commented):
Doc the orphan-bytes case.

src/main/kotlin/org/wfanet/measurement/edpaggregator/vidrankbuilder/SubpoolRetention.kt
```diff
@@ -0,0 +1,116 @@
...
+    val cutoff = today.minusDays(retentionDays.toLong() + 1L) // strictly older than RETENTION_DAYS
+    val candidates = listDeletionCandidates(poolOffset, cutoff)
+    if (candidates.isEmpty()) return
+
+    logger.info(
+      "Subpool $poolOffset for $modelLine: garbage-collecting ${candidates.size} aged DAY_ONLY blob(s)"
+    )
+    for (candidate in candidates) {
+      // Soft-delete the row first so a concurrent retry skips it, then delete the bytes.
+      rankIndexBlobsStub.deleteRankIndexBlob(deleteRankIndexBlobRequest { name = candidate.name })
```
stevenwarejones, line 76: nit: document the orphan-bytes case. If `deleteRankIndexBlob` succeeds and `rankIndexStore.delete` fails (transient GCS error, concurrent retry that already deleted the bytes), the row is gone but the bytes leak on Cloud Storage. The ordering is correct — but a reader has no signal that this case is handled.

Sibling code calls this out explicitly. From `SubpoolAssigner.kt:289-295`:
> "A crash in the narrow post-flip/pre-delete window leaks the temp blobs (no corruption) — they are reclaimed by the monitor (#3958) or the bucket lifecycle policy."

Add the equivalent here so the leak-vs-corruption distinction and the downstream sweeper are visible to the next reader:

```kotlin
for (candidate in candidates) {
  // Soft-delete the row first so a concurrent retry skips it, then delete the bytes. A
  // failure between the two leaks the bytes (no corruption — the row is gone, so nothing
  // references them) and they're reclaimed by the monitor (#3958) or the bucket lifecycle.
  rankIndexBlobsStub.deleteRankIndexBlob(deleteRankIndexBlobRequest { name = candidate.name })
  rankIndexStore.delete(candidate.blobUri)
}
```

stevenwarejones (commented):
> @stevenwarejones reviewed 7 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4017)* status: all files reviewed, 5 unresolved discussions (waiting on Marco-Premier).

src/test/kotlin/org/wfanet/measurement/edpaggregator/vidrankbuilder/SubpoolRankerTest.kt
```diff
@@ -0,0 +1,329 @@
...
+  private suspend fun readSnapshot(): Map<Pair<Long, Int>, Pair<Int, Int>> {
+    val row =
+      fake.rows.single {
+        it.blobType == RankIndexBlob.BlobType.SNAPSHOT && it.name.startsWith("$UPLOAD/")
+      }
+    return decode(rankStore.readBlob(row.blobUri, row.encryptedDek).toList())
+  }
+
+  @Test
+  fun `cold subpool allocates sequential ranks and stamps last_seen`() = runBlocking {
```
stevenwarejones, line 190: Test coverage gaps in this file — these are real failure modes the current 5 tests don't exercise.

**A. Prior-snapshot checksum mismatch.** `rankIndexStore.readBlob(..., priorSnapshot.blobChecksum)` throws on mismatch (the integrity guard against silent rank-state corruption). No test exercises this path — if a corrupt blob lands and the checksum check regresses, every dispatch silently rebuilds the allocator from garbage. Add a test that seeds a prior snapshot whose `blobChecksum` doesn't match the bytes, calls `ranker.rank(...)`, and asserts it throws.

**B. `findPriorSnapshot` picking the newest of N.** `seedPriorSnapshot` is called exactly once in every test, so the "newest by create_time" logic at lines 274-283 is untested. If it regressed to first-of-page or oldest-of-page (same shape as the bug I flagged in `findUploadSnapshot`), tests pass. Add: seed two prior SNAPSHOTs under different uploads with different `createTime`s and assert the allocator loads the entries from the newer one.

**C. Empty Phase-0 merged blob.** When a subpool got zero impressions this dispatch, the merged blob is empty → `todayFps` stays empty → no `assign()` calls. Production still writes an empty DAY_ONLY and a SNAPSHOT carrying just the prior cumulative + inserts both rows. No test exercises this. Add: `writePhase0(emptyList())`, call `rank(...)`, assert no `assign`/`renewal`/`overflow`, and that a SNAPSHOT row is still inserted carrying the prior cumulative unchanged.

These shore up the actual correctness guarantees the production code makes (checksum protects state; "newest" actually means newest; zero-impression subpools don't break the row-insert contract).

stevenwarejones (commented):
Test coverage gaps for SubpoolRanker.

src/test/kotlin/org/wfanet/measurement/edpaggregator/vidrankbuilder/SubpoolRetentionTest.kt
```diff
@@ -0,0 +1,133 @@
...
+      maxEventDate = date {
+        year = 2000
+        month = 1
+        day = 1
+      } // old
+    }
+  }
+
+  @Test
+  fun `deleteAgedBlobs soft-deletes the row and removes the blob bytes`() = runBlocking {
```
stevenwarejones, line 141: Two test coverage gaps in this file:

**E. Row delete succeeds, bytes delete fails.** The KDoc at SubpoolRetention.kt:71 takes the "row first, then bytes" ordering deliberately. The real failure mode is "row gone, bytes leaked." Today if `rankIndexStore.delete` throws (transient GCS error), the loop aborts and the rest of the candidates are untouched — half-deleted batch. Add a test where the stub `deleteRankIndexBlob` succeeds but `rankStore.delete` throws on the first candidate; assert the row IS deleted (the KDoc claim holds) and the leaked bytes are surfaced as a metric / logged (or, if we accept "leak then lifecycle-sweep," document that's the design and assert the loop aborts rather than silently swallowing).

**F. Cutoff boundary off-by-one.** `cutoff = today.minusDays(retentionDays.toLong() + 1L)` and the predicate is "strictly older than `RETENTION_DAYS`." With `retentionDays=30, today=epoch-day 100`:
- a blob with `MaxEventDate = epoch-day 69` (i.e., `today - 31`) **should** be deleted (cutoff)
- a blob with `MaxEventDate = epoch-day 70` (i.e., `today - 30`) **should not** be deleted (exactly retention)

Neither boundary is tested. The current `writeBlob` helper hardcodes `date { year = 2000 }` for the "old" case, which is way past the cutoff. An off-by-one in retention is exactly the kind of bug that silently frees still-active ranks. Add a parameterized test (or two): one at `today - retentionDays - 1` (must delete) and one at `today - retentionDays` (must NOT delete).

stevenwarejones (commented):
Test coverage gaps for SubpoolRetention.

stevenwarejones (commented):
Test for unset maxEventDate (companion to require).

src/test/kotlin/org/wfanet/measurement/edpaggregator/vidrankbuilder/SubpoolRankerTest.kt
stevenwarejones, line 190: **D. Companion test for the `epochDayOf` fix.** Pairs with the blocker on `SubpoolRanker.kt:338` (silent `today` fallback for `date.year == 0`). Once that fix lands as `require(date.year != 0) { ... }`, add a test that exercises the require:

- Construct a `SubpoolRanker` with `maxEventDate = Date.getDefaultInstance()` (unset).
- Call `rank(...)`.
- Assert `IllegalArgumentException` with the expected message.

This guards two things at once:
1. That the require actually fires (regression guard if someone later "fixes" the require to a silent default again).
2. That `VidRankBuilderApp.runWork`'s upfront validation (the better location for the same check) actually catches an unset `params.maxEventDate` before construction.

If you push the validation up to `VidRankBuilderApp.runWork` (recommended), add the same test in `VidRankBuilderAppTest` (or whatever file currently exercises `runWork`'s `require(...)` chain at lines 119-123).

src/main/kotlin/org/wfanet/measurement/edpaggregator/vidrankbuilder/SubpoolRanker.kt
Marco-Premier: Done.

src/main/kotlin/org/wfanet/measurement/edpaggregator/vidrankbuilder/SubpoolRanker.kt
Marco-Premier: Still WIP

src/main/kotlin/org/wfanet/measurement/edpaggregator/vidrankbuilder/SubpoolRanker.kt
Marco-Premier: Done.

src/main/kotlin/org/wfanet/measurement/edpaggregator/vidrankbuilder/SubpoolRanker.kt
Marco-Premier: Done.

src/test/kotlin/org/wfanet/measurement/edpaggregator/vidrankbuilder/SubpoolRankerTest.kt
Marco-Premier: Done.

Marco-Premier: Done.

src/test/kotlin/org/wfanet/measurement/edpaggregator/vidrankbuilder/SubpoolRetentionTest.kt
Marco-Premier: Done.

Marco-Premier (commented):
> @Marco-Premier made 7 comments and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4017)* status: 2 of 7 files reviewed, 7 unresolved discussions (waiting on stevenwarejones).

stevenwarejones (commented):
> @stevenwarejones reviewed 9 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4017)* status: all files reviewed, 7 unresolved discussions (waiting on Marco-Premier).

src/main/kotlin/org/wfanet/measurement/edpaggregator/vidrankbuilder/SubpoolRanker.kt
```diff
@@ -0,0 +1,588 @@
...
+    poolOffset: Long,
+    eventDay: Int,
+    priorSnapshot: RankIndexBlob?,
+  ): RankIndexBlob? {
+    if (priorSnapshot == null || !priorSnapshot.hasMaxEventDate()) return null
+    val latestEventDay = priorSnapshot.maxEventDate.epochDay()
+    // Forward append (newest or same day) — not a backfill.
+    if (eventDay >= latestEventDay) return null
+    // Older than the retention window — no live ranks remain to reclaim (Problem 3, out of scope).
+    if (latestEventDay - eventDay > retentionDays) return null
```
stevenwarejones: Emit a metric and log a warning when this branch fires. Today an out-of-retention backfill silently falls through to the forward-append path, which fresh-ranks every fingerprint — including those still tracked by the data provider as the same person. The KDoc calls this 'Problem 3, out of scope', but blind corruption isn't acceptable; ops needs to know when this lands so they can run reconciliation, extend retention, or whatever the runbook says.

```
if (latestEventDay - eventDay > retentionDays) {
  logger.warning(
    "Subpool $poolOffset for $modelLine: out-of-retention backfill " +
      "(gap=${latestEventDay - eventDay} days, retention=$retentionDays); " +
      "treating as forward append — Problem 3, ranks may diverge."
  )
  metrics.outOfRetentionBackfillCounter.add(1)
  return null
}
```

Add `outOfRetentionBackfillCounter` to VidRankBuilderMetrics.

src/main/kotlin/org/wfanet/measurement/edpaggregator/vidrankbuilder/SubpoolRanker.kt
```diff
@@ -0,0 +1,588 @@
...
+    // 3b. Free aged-out ranks in heap from the `last_seen` index, before allocating so freed slots
+    // are reusable today. Fingerprints seen this dispatch are excluded so their ranks (VIDs)
+    // survive.
+    allocator.freeAgedRanks(cutoffEpochDay, todayFps)
+
+    // 4. Assign ranks to today's fingerprints.
+    if (oldSnapshot != null) {
+      // Backfill: replay each fingerprint's historical rank into the day-only delta while extending
+      // the cumulative forward (see class KDoc and RankAllocator.assignBackfill).
+      val oldRanks = loadFingerprintRanks(oldSnapshot)
```
stevenwarejones, line 246: Emit a metric for the loaded old-snapshot's size so ops can alert before OOM. The class KDoc explicitly warns this load can OOM an n2d-highmem-16 at India scale (~1.5B fingerprints, two in-heap maps), but there's no runtime signal — the process just dies. Bytes12IntMap.size is an O(1) maintained counter, so the metric call is free.

```
val oldRanks = loadFingerprintRanks(oldSnapshot)
metrics.backfillOldSnapshotEntries.record(oldRanks.size)
```

Add `backfillOldSnapshotEntries` as a histogram in VidRankBuilderMetrics.

src/main/kotlin/org/wfanet/measurement/edpaggregator/vidrankbuilder/SubpoolRanker.kt
stevenwarejones: Verified — `findUploadBlob` (SubpoolRanker.kt:355) now uses `.firstOrNull { page -> page.firstOrNull() != null }?.firstOrNull()`, which short-circuits at the first matching page and cancels pagination.

src/main/kotlin/org/wfanet/measurement/edpaggregator/vidrankbuilder/SubpoolRanker.kt
stevenwarejones: Verified — backfill safety addressed comprehensively. A backfill now writes both DAY_ONLY and SNAPSHOT (SubpoolRanker.kt:277-292), the SNAPSHOT's `max_event_date` is inherited from the prior snapshot (line 283), `resolveBackfillOldSnapshot` (line 418-460) gates on the retention window, and `assignBackfill` (RankAllocator.kt:249-298) handles all four cases (new fingerprint, old rank free, old rank taken, already-ranked) consistently — DAY_ONLY labels with the historical rank, SNAPSHOT carries forward without lowering existing ranks. The 13 new backfill tests in SubpoolRankerTest.kt:338-637 cover the boundary cases, collision counting, retention-window edges, and the 'closest older snapshot when several exist' case.

src/main/kotlin/org/wfanet/measurement/edpaggregator/vidrankbuilder/SubpoolRanker.kt
stevenwarejones: Verified — all three sites now use `ResourceKey.WILDCARD_ID`: `findPriorSnapshot` (SubpoolRanker.kt:377), `resolveBackfillOldSnapshot` (line 437), and `SubpoolRetention.listDeletionCandidates` (line 94).

src/main/kotlin/org/wfanet/measurement/edpaggregator/vidrankbuilder/SubpoolRanker.kt
stevenwarejones: Verified — `epochDayOf` (SubpoolRanker.kt:550-553) now uses `require(date.year != 0) { "maxEventDate must be set on VidRankBuilderParams" }`, and the companion test at SubpoolRankerTest.kt:327-335 exercises it.

src/main/kotlin/org/wfanet/measurement/edpaggregator/vidrankbuilder/SubpoolRetention.kt
stevenwarejones: Verified — the orphan-bytes case is documented at SubpoolRetention.kt:72-75: "A failure between the two leaks the bytes (no corruption — the row is gone, so nothing references them) and they're reclaimed by the monitor (#3958) or the bucket lifecycle."

src/test/kotlin/org/wfanet/measurement/edpaggregator/vidrankbuilder/SubpoolRankerTest.kt
stevenwarejones: Verified — all four gaps covered: A (prior-snapshot checksum mismatch) at SubpoolRankerTest.kt:275-286, B (loads newest across uploads) at 288-308, C (empty subpool still writes snapshot) at 311-324, D (year=0 fails loudly) at 327-335. Plus 13 new backfill scenarios at lines 338-637 covering boundary, collision, ordering, and out-of-window cases.

stevenwarejones: Verified — companion test at SubpoolRankerTest.kt:327-335 asserts `IllegalArgumentException` with "maxEventDate must be set" when `Date.getDefaultInstance()` is passed. Pairs with the `require` fix at SubpoolRanker.kt:551.

src/test/kotlin/org/wfanet/measurement/edpaggregator/vidrankbuilder/SubpoolRetentionTest.kt
stevenwarejones: Verified — both gaps covered. E (row delete succeeds, bytes delete fails) at SubpoolRetentionTest.kt:172-209 asserts the IOException propagates and the row was deleted first. F (boundary cases) at lines 212-242 covers exactly-at-cutoff (kept) and one-day-past (deleted).

stevenwarejones (changes requested):
> @stevenwarejones+AGNT made 8 comments.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4017)* status: all files reviewed, 9 unresolved discussions (waiting on Marco-Premier and stevenwarejones).

src/main/kotlin/org/wfanet/measurement/edpaggregator/vidrankbuilder/SubpoolRanker.kt
stevenwarejones: Correction on my own comment — this should **throw**, not log-and-continue.

The retention gate exists precisely *because* an out-of-retention backfill corrupts the cumulative. Proceeding past it makes the gate cosmetic: the corruption happens, we just write to a counter on the way down. That's worse than not having the check.

Throw a structured exception (mirroring the `require(date.year != 0)` fail-loud pattern at SubpoolRanker.kt:551):

```
if (latestEventDay - eventDay > retentionDays) {
  metrics.outOfRetentionBackfillCounter.add(1)
  throw OutOfRetentionBackfillException(
    "Subpool $poolOffset for $modelLine: out-of-retention backfill " +
      "(gap=${latestEventDay - eventDay} days, retention=$retentionDays). " +
      "Forward-append would silently re-rank still-tracked fingerprints. " +
      "Extend retention or mark the dispatch as skip-ranked before retrying."
  )
}
```

The exception surfaces as `FAILED_PRECONDITION` through Phase-2 and marks the `RankerJob` as FAILED. Pipeline halts on that subpool — operator decides: extend retention and retry, skip the backfill, or backfill manually. No silent reach undercount in produced labels.

The metric still increments so ops can dashboard on frequency.

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4017)* status: all files reviewed, 9 unresolved discussions (waiting on Marco-Premier and stevenwarejones).

src/main/kotlin/org/wfanet/measurement/edpaggregator/vidrankbuilder/SubpoolRanker.kt
Marco-Premier: Done.

src/main/kotlin/org/wfanet/measurement/edpaggregator/vidrankbuilder/SubpoolRanker.kt
Marco-Premier: Done.

src/main/kotlin/org/wfanet/measurement/edpaggregator/vidrankbuilder/SubpoolRanker.kt
Marco-Premier: Done.

src/main/kotlin/org/wfanet/measurement/edpaggregator/vidrankbuilder/SubpoolRanker.kt
Marco-Premier: Done.

src/main/kotlin/org/wfanet/measurement/edpaggregator/vidrankbuilder/SubpoolRanker.kt
Marco-Premier: Done.

src/main/kotlin/org/wfanet/measurement/edpaggregator/vidrankbuilder/SubpoolRanker.kt
Marco-Premier: Done.

src/test/kotlin/org/wfanet/measurement/edpaggregator/vidrankbuilder/SubpoolRankerTest.kt
Marco-Premier: Done.

Marco-Premier: Done.

src/test/kotlin/org/wfanet/measurement/edpaggregator/vidrankbuilder/SubpoolRetentionTest.kt
Marco-Premier: Done.

Marco-Premier (commented):
> @Marco-Premier made 9 comments.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4017)* status: 6 of 11 files reviewed, 9 unresolved discussions (waiting on stevenwarejones and stevenwarejones+AGNT).

src/main/kotlin/org/wfanet/measurement/edpaggregator/vidrankbuilder/SubpoolRanker.kt
stevenwarejones: Confirming closure — verification verified, Marco's 'Done' acknowledges.

src/main/kotlin/org/wfanet/measurement/edpaggregator/vidrankbuilder/SubpoolRanker.kt
stevenwarejones: Confirming closure — my prior verification stands and Marco's 'Done' acknowledges. Thread can be resolved on his side.

src/main/kotlin/org/wfanet/measurement/edpaggregator/vidrankbuilder/SubpoolRanker.kt
stevenwarejones: Confirming closure — verification verified, Marco's 'Done' acknowledges.

src/main/kotlin/org/wfanet/measurement/edpaggregator/vidrankbuilder/SubpoolRanker.kt
stevenwarejones: Confirming closure — verification verified, Marco's 'Done' acknowledges.

src/main/kotlin/org/wfanet/measurement/edpaggregator/vidrankbuilder/SubpoolRanker.kt
stevenwarejones: Verified in commit 2936e8982. `metrics.backfillOldSnapshotEntriesHistogram.record(oldRanks.size)` is now called immediately after `loadFingerprintRanks` at SubpoolRanker.kt:247. The histogram (`edpa.vid_rank_builder.backfill_old_snapshot_entries`) lets ops alert before OOM on the second in-heap map.

src/main/kotlin/org/wfanet/measurement/edpaggregator/vidrankbuilder/SubpoolRanker.kt
stevenwarejones: Verified in commit 2936e8982. `resolveBackfillOldSnapshot` at SubpoolRanker.kt:435-443 now throws `OutOfRetentionBackfillException` (structured, with poolOffset/modelLine/gapDays/retentionDays fields) after incrementing `metrics.outOfRetentionBackfillCounter`. Out-of-retention backfill fails loud instead of silently forward-appending — the orchestrator marks the RankerJob FAILED so an operator can act. Resolves the silent-corruption hazard.

src/test/kotlin/org/wfanet/measurement/edpaggregator/vidrankbuilder/SubpoolRankerTest.kt
stevenwarejones: Confirming closure — test coverage verified, Marco's 'Done' acknowledges.

stevenwarejones: Confirming closure — companion test verified, Marco's 'Done' acknowledges.

src/test/kotlin/org/wfanet/measurement/edpaggregator/vidrankbuilder/SubpoolRetentionTest.kt
stevenwarejones: Confirming closure — test coverage verified, Marco's 'Done' acknowledges.

stevenwarejones (commented):
> @stevenwarejones+AGNT made 9 comments.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4017)* status: 6 of 11 files reviewed, 9 unresolved discussions (waiting on stevenwarejones).

stevenwarejones (approved):
> @stevenwarejones reviewed 4 files and all commit messages, and resolved 9 discussions.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4017)* status: 10 of 11 files reviewed, all discussions resolved (waiting on stevenwarejones+AGNT).

kungfucraig (approved):
> @kungfucraig reviewed 11 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4017)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on Marco-Premier).

Marco-Premier (commented):
> @Marco-Premier reviewed 4 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4017)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on Marco-Premier).

Marco-Premier (commented):
> @Marco-Premier reviewed 11 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4017)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on Marco-Premier).

PR #4025, opened 2026-06-18
fix: avoid failing BasicReports on AddProcessedResultValues preconditions

### LINKED ISSUE #4024: Reporting post-processor incorrectly fails BasicReports when AddProcessedResultValues returns FAILED_PRECONDITION or transient errors
## Summary

The post-processor cron job's `_process_basic_report` method in `src/main/python/wfa/measurement/reporting/job/post_process_report_result_job.py` catches any exception from either the local post-processor or the gRPC `AddProcessedResultValues` call and unconditionally marks the BasicReport as FAILED. That over-broad catch can corrupt BasicReports that already have valid processed results, and incorrectly fails BasicReports for transient gRPC errors that would otherwise resolve on the next tick.

Discovered while investigating a recurring `BasicReportStateInvalidException` pattern in staging that produced an unexpected `state: FAILED` on BasicReports whose underlying Reports were all `SUCCEEDED`.

## Background

The reporting v2 BasicReport state machine is:
```
CREATED → REPORT_CREATED → UNPROCESSED_RESULTS_READY → SUCCEEDED
                                                    ↘ FAILED
```

State-transitioning RPCs:
- `BatchCreateReportingSetResults` calls `associateBasicReport` (`SpannerReportResultsService.kt:378`), which requires state `REPORT_CREATED` and transitions to `UNPROCESSED_RESULTS_READY`.
- `AddProcessedResultValues` calls `markBasicReportCompleted` (`SpannerReportResultsService.kt:612`), which requires state `UNPROCESSED_RESULTS_READY` and transitions to `SUCCEEDED`.
- `FailBasicReport` transitions to `FAILED` from any state.

Both `associateBasicReport` and `markBasicReportCompleted` throw `BasicReportStateInvalidException` (returned to clients as `FAILED_PRECONDITION`) when the current state does not match what the operation requires.

The OSS deployment runs the post-processing pipeline as one Pod per tick: `basic-reports-reports` initContainer (Kotlin, calls `BatchCreateReportingSetResults`), then `report-result-post-processor` main container (Python, calls `AddProcessedResultValues`). They share a Pod via the report-result-post-processor CronJob defined in `src/main/k8s/reporting_v2.cue`.

## Bug

In `post_process_report_result_job.py`:

```python
try:
    add_processed_result_values_request = self._post_processor.process(
        basic_report.cmms_measurement_consumer_id,
        basic_report.external_report_result_id,
        self._ami_mrc_exempted_edps,
    )
    if add_processed_result_values_request:
        self._report_results_stub.AddProcessedResultValues(
            add_processed_result_values_request)
except Exception:
    logging.warning(...)
    self._basic_reports_stub.FailBasicReport(...)
    succeeded = False
```

The `except Exception:` catches every failure mode from both calls and treats them all as "this BasicReport is unprocessable." That conflates three distinct cases:

1. **Solver / data-parsing failure** in `_post_processor.process(...)` → BasicReport really cannot be processed → FAIL is correct.
2. **`FAILED_PRECONDITION` from `AddProcessedResultValues`** → state is no longer `UNPROCESSED_RESULTS_READY` because another writer already advanced it (typically to `SUCCEEDED`). The work is done. Failing here corrupts a BasicReport that already has valid `result_details`.
3. **Transient gRPC errors** (`UNAVAILABLE`, `DEADLINE_EXCEEDED`, etc.) from `AddProcessedResultValues` → should retry on the next tick, not fail.

## Evidence from a real incident

Staging cluster `origin-fe-stg-k8s-v2-cluster`, 2026-06-17 16:50 UTC. The cron tick at `*/5` produced `BasicReportStateInvalidException` errors for 6 BasicReports. Investigation found a second writer running the same Kotlin pipeline concurrently against the same Spanner; that writer raced the cronjob, sometimes won the `associateBasicReport` step, and proceeded to call `AddProcessedResultValues` itself within the same tick.

Server log at 16:50:41 / 16:50:42 (`postgres-internal-reporting-server-container`):

```
gRPC ... error: FAILED_PRECONDITION
  BasicReport with external key (ONYdS_z_3xo, ec6460c81464f04bb947c818fdb83dpoabdelv2)
  is in state SUCCEEDED which is invalid for the operation

gRPC ... error: FAILED_PRECONDITION
  BasicReport with external key (ONYdS_z_3xo, abdel-single-edp-test-1)
  is in state SUCCEEDED which is invalid for the operation
```

Both RPCs were called by the post-processor pod (`user-agent: grpc-python/1.74.0`). For both BasicReports the cron Pod's `_process_basic_report` then called `FailBasicReport`, flipping `SUCCEEDED → FAILED`.

Final state check confirmed: both BasicReports are now `state: FAILED` even though their underlying Reports are `SUCCEEDED` and their `result_details` were successfully written by the racing writer.

For comparison, the other 4 BasicReports in the same incident window (`af2b9f4e…`, `aa8a522e…`, `a67f6a14…`, `a3318317…`) ended up `SUCCEEDED` — they hit the same race at the associate step but, by the time the OSS post-processor reached them, their state machine had completed cleanly. So this bug only manifests for the subset where the racing writer's full pipeline (including `AddProcessedResultValues`) completed in the brief window between the OSS post-processor's `ListBasicReports` call and its per-BasicReport `AddProcessedResultValues` call.

## Reproduction

Tests in `src/test/python/wfa/measurement/reporting/job/post_process_report_result_job_test.py`:

- `test_execute_skips_basic_report_already_processed` — mocks `AddProcessedResultValues` to raise `grpc.RpcError` with code `FAILED_PRECONDITION` and details mimicking the real server message. Asserts `FailBasicReport` is **not** called.
- `test_execute_does_not_fail_on_transient_grpc_error` — mocks `AddProcessedResultValues` to raise `grpc.RpcError` with code `UNAVAILABLE`. Asserts `FailBasicReport` is **not** called.

Both currently fail on `main`; both pass with the fix proposed in the linked PR.

## Proposed fix

Split the exception handling so that:

- `_post_processor.process(...)` exceptions still call `FailBasicReport` (unchanged behavior for genuine processing failures).
- `AddProcessedResultValues` raising `grpc.RpcError` with code `FAILED_PRECONDITION` logs an INFO message and returns success (the work is done by another writer; skip silently).
- `AddProcessedResultValues` raising any other `grpc.RpcError` logs a WARNING and returns failure but does **not** call `FailBasicReport`, allowing the next tick to retry.

The fix is local to one method and includes regression tests for both failure modes.

## Related work / not in scope

The root cause of having two concurrent writers in the same cluster is a separate operational concern (someone is running the Kotlin executor outside the cron Pod). Fixing the post-processor's exception handling is the durable code-level defense that makes the system robust regardless of operational topology.

A separate change to make `associateBasicReport` idempotent (so the cron's "loser" call doesn't log `BasicReportStateInvalidException` as a WARNING either) would further reduce log noise but is not required to prevent BasicReport corruption.

stevenwarejones:
Fixes #4024.

Issue: #4024


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4025)

SanjayVas (changes requested):
> @SanjayVas made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4025)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on stevenwarejones).

___
*[`src/main/python/wfa/measurement/reporting/job/post_process_report_result_job.py` line 181 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4025#-OvQiPtOBNGotmCXSlJ8:-OvQiPtOBNGotmCXSlJ9:b-9ezgzl) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/347111a8aafd6c0e2c26c8d145fd29b00d21280d/src/main/python/wfa/measurement/reporting/job/post_process_report_result_job.py#L181)):*
> ```Python
>         returns False so the caller falls through to its normal error path.
>         """
>         try:
> ```

This can just read the ErrorInfo to disambiguate. The reason will be `BASIC_REPORT_STATE_INVALID` and the `basicReportState` metadata field will tell you the state.

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4025)* status: 0 of 5 files reviewed, 1 unresolved discussion (waiting on SanjayVas).

___
*[`src/main/python/wfa/measurement/reporting/job/post_process_report_result_job.py` line 181 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4025#-OvQiPtOBNGotmCXSlJ8:-OvRVasi3CyR7u0xmDGW:b3xkyrq) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/347111a8aafd6c0e2c26c8d145fd29b00d21280d/src/main/python/wfa/measurement/reporting/job/post_process_report_result_job.py#L181)):*

This can just read the ErrorInfo to disambiguate. The reason will be `BASIC_REPORT_STATE_INVALID` and the `basicReportState` metadata field will tell you the state.
</blockquote></details>

Good call — thanks. Updated in the latest commit to parse the ErrorInfo on the response trailers and check `BASIC_REPORT_STATE_INVALID` + `basicReportState` rather than doing a separate `GetBasicReport`. Pulled in `googleapis-common-protos` as an explicit dep so the Python `google.rpc.status` / `google.rpc.error_details` modules are reachable. Tests adapted to mock the trailer-attached ErrorInfo instead of a separate GetBasicReport response.

stevenwarejones (commented):
> @stevenwarejones reviewed 5 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4025)* status: all files reviewed, 1 unresolved discussion (waiting on SanjayVas).

SanjayVas (changes requested):
> @SanjayVas reviewed 5 files and all commit messages, made 4 comments, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4025)* status: all files reviewed, 4 unresolved discussions (waiting on stevenwarejones).

___
*[`src/main/python/wfa/measurement/reporting/job/post_process_report_result_job.py` line 46 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4025#-Ovw57PhC5yAcoa6dKlg:-Ovw57PhC5yAcoa6dKlh:b-h1hcls) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7d667de2b2780456516984151f6bf91fb6eaff56/src/main/python/wfa/measurement/reporting/job/post_process_report_result_job.py#L46)):*
> ```Python
> 
> 
> def _basic_report_state_past_unprocessed(
> ```

nit: move this below the `_MAX_PAGE_SIZE` constant
___
*[`src/main/python/wfa/measurement/reporting/job/post_process_report_result_job.py` line 56 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4025#-Ovw3T8g8mOodGblqfNT:-Ovw3T8h87oQMtdtnqtN:b-n66dzm) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7d667de2b2780456516984151f6bf91fb6eaff56/src/main/python/wfa/measurement/reporting/job/post_process_report_result_job.py#L56)):*
> ```Python
>     The server attaches a google.rpc.Status with an ErrorInfo to the
>     grpc-status-details-bin trailing metadata, per
>     google.rpc.error_details.proto. We parse it without taking a dependency
> ```

I don't think we should need to parse this ourselves. What's the concern with depending on grpcio-status for this? It seems pretty lightweight on top of grpcio. The functions are documented as experimental, but they have been in that state for 8 years.
___
*[`src/main/python/wfa/measurement/reporting/job/post_process_report_result_job.py` line 80 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4025#-Ovw4QNVC4mZPVUaZS7C:-Ovw4QNVC4mZPVUaZS7D:b985pck) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7d667de2b2780456516984151f6bf91fb6eaff56/src/main/python/wfa/measurement/reporting/job/post_process_report_result_job.py#L80)):*
> ```Python
>             )
>             return None
>         for detail in status.details:
> ```

nit: extract a function for extracting ErrorInfo from a `grpc.RpcError`.
___
*[`src/main/python/wfa/measurement/reporting/job/post_process_report_result_job.py` line 85 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4025#-Ovw4K1WA2QqpO2XXMK2:-Ovw4K1WA2QqpO2XXMK3:b-4j8zzg) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7d667de2b2780456516984151f6bf91fb6eaff56/src/main/python/wfa/measurement/reporting/job/post_process_report_result_job.py#L85)):*
> ```Python
>             error_info = error_details_pb2.ErrorInfo()
>             detail.Unpack(error_info)
>             if error_info.reason != _BASIC_REPORT_STATE_INVALID_REASON:
> ```

nit: This should also check the domain just for completeness.

stevenwarejones (commented):
> @stevenwarejones reviewed 3 files and all commit messages, made 2 comments, and resolved 3 discussions.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4025)* status: 4 of 6 files reviewed, 1 unresolved discussion (waiting on SanjayVas).

___
*[`src/main/python/wfa/measurement/reporting/job/post_process_report_result_job.py` line 56 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4025#-Ovw3T8g8mOodGblqfNT:-OvylerBBSpQ9FUxkzBF:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7d667de2b2780456516984151f6bf91fb6eaff56/src/main/python/wfa/measurement/reporting/job/post_process_report_result_job.py#L56)):*

I don't think we should need to parse this ourselves. What's the concern with depending on grpcio-status for this? It seems pretty lightweight on top of grpcio. The functions are documented as experimental, but they have been in that state for 8 years.
</blockquote></details>

Done.
___
*[`src/main/python/wfa/measurement/reporting/job/post_process_report_result_job.py` line 80 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4025#-Ovw4QNVC4mZPVUaZS7C:-OvymAaS1EEE73XITbEY:b-k5ixvb) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/7d667de2b2780456516984151f6bf91fb6eaff56/src/main/python/wfa/measurement/reporting/job/post_process_report_result_job.py#L80)):*

nit: extract a function for extracting ErrorInfo from a `grpc.RpcError`.
</blockquote></details>

done. extracted _extract_error_info

stevenwarejones (commented):
> @stevenwarejones reviewed 2 files.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4025)* status: all files reviewed, 1 unresolved discussion (waiting on SanjayVas).

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4025)* status: all files reviewed, 1 unresolved discussion (waiting on SanjayVas).

SanjayVas (approved):
> @SanjayVas reviewed 5 files and all commit messages, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4025)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on stevenwarejones).

PR #4039, opened 2026-06-18
fix: paginate through all REPORT_CREATED BasicReports per cron tick

### LINKED ISSUE #4038: BasicReportsReportsJob processes at most 10 BasicReports per cron tick due to pagination cap
## Summary

`BasicReportsReportsJob` processes at most 10 BasicReports per cron tick, total, regardless of how many are in `REPORT_CREATED` state. When more than `BATCH_SIZE` (= 10) BasicReports are awaiting processing for a single MeasurementConsumer, the 11th and beyond are never picked up until items at the head of the list move out of `REPORT_CREATED`.

In a staging environment where >10 BasicReports were stuck at the head of the list, a freshly-created BasicReport sat in `REPORT_CREATED` for hours with no cron tick ever touching it — its Report could not progress and its child Metrics could not advance (because the cron's `getReport()` chain is the de facto trigger for Metric sync via `MetricsService.syncAndConvertInternalMetricsToPublicMetrics`).

## Root cause

`BasicReportsReportsJob.execute()` calls the paginating helper with `BATCH_SIZE` as the **total** limit, not the per-page page size:

```kotlin
val resourceLists =
  internalBasicReportsStub.listResources(BATCH_SIZE, null) {
    pageToken: ListBasicReportsPageToken?,
    remaining ->
    val listBasicReportsResponse =
      internalBasicReportsStub.listBasicReports(
        listBasicReportsRequest {
          pageSize = remaining
          ...
```

The helper in `common/api/grpc/ListResources.kt` defines:

```kotlin
inline fun ... S.listResources(
  limit: Int,                          // total cap on emitted resources
  ...,
  crossinline list: suspend S.(pageToken: T, remaining: Int) -> ResourceList<R, T>,
): Flow<ResourceList<R, T>> {
  ...
  while (true) {
    val resourceList = list(nextPageToken, remaining)
    emit(resourceList)
    remaining -= resourceList.size
    nextPageToken = resourceList.nextPageToken
    if (nextPageToken == emptyPageToken || remaining == 0) break  // ← breaks at remaining=0
  }
}
```

After the first page returns `BATCH_SIZE` items, `remaining` becomes 0 and the helper breaks out — even when the response carries a `next_page_token`.

## Observed impact

Staging cluster `origin-fe-stg-k8s-v2-cluster`, MeasurementConsumer `ONYdS_z_3xo`. Two cron ticks 5 hours apart (13:40 UTC and 18:45 UTC) returned the **identical** 10 BasicReports — meaning the head 10 hadn't drained AND nothing past position 10 was ever fetched. A BasicReport created at 13:35 UTC sat untouched by any cron tick for the entire 5+ hour observation window.

Plausibly the same mechanism that caused the recently-reported ~17,888 metrics stuck in RUNNING state for >2 days: BasicReports past position 10 never trigger their Reports' metric sync.

## Fix

Switch to the no-limit `listResources` overload (it internally uses `Int.MAX_VALUE`) and set `pageSize = BATCH_SIZE` explicitly on each request. This makes `BATCH_SIZE` the page size (its intended meaning) and lets the helper paginate until the response stops returning `next_page_token`.

## Tests

Two regression tests added in `BasicReportsReportsJobTest`:

- `execute paginates through all REPORT_CREATED basic reports across multiple pages` — sets up a 2-page response (11 items: page 1 has 10 + `next_page_token`, page 2 has 1) and asserts that `listBasicReports` is called twice and all 11 BasicReports are processed.
- `execute does not fetch second page when first response has no next page token` — sanity test that ensures the fix doesn't introduce unnecessary calls when pagination isn't needed.

Both fail on `main`; both pass with the fix.

## Related / follow-up

This bug compounds with anything else that delays the head of the `REPORT_CREATED` list (Reports that take a long time to complete, or BasicReports whose Reports are stuck for reasons external to this code). Even with this fix, throughput is one cron-tick-worth of work every 5 minutes; if BasicReport creation rate exceeds that, backlog still grows. Tuning `BATCH_SIZE` or processing concurrency may be worth a follow-up. For this PR, the pagination correctness fix is the minimal change.

stevenwarejones:
Fixes #4038.

Issue: #4038

### DIFF
--- a/src/main/kotlin/org/wfanet/measurement/reporting/job/BasicReportsReportsJob.kt
+++ b/src/main/kotlin/org/wfanet/measurement/reporting/job/BasicReportsReportsJob.kt
@@ -106,13 +106,12 @@ class BasicReportsReportsJob(
           .measurementConsumerId
 
       val resourceLists =
-        internalBasicReportsStub.listResources(BATCH_SIZE, null) {
-          pageToken: ListBasicReportsPageToken?,
-          remaining ->
+        internalBasicReportsStub.listResources(pageToken = null as ListBasicReportsPageToken?) {
+          pageToken: ListBasicReportsPageToken? ->
           val listBasicReportsResponse =
             internalBasicReportsStub.listBasicReports(
               listBasicReportsRequest {
-                pageSize = remaining
+                pageSize = BATCH_SIZE
                 if (pageToken != null) {
                   this.pageToken = pageToken
                 }
--- a/src/main/kotlin/org/wfanet/measurement/reporting/job/BasicReportsReportsJob.kt
+++ b/src/main/kotlin/org/wfanet/measurement/reporting/job/BasicReportsReportsJob.kt
@@ -121,8 +121,7 @@ class BasicReportsReportsJob(
                   }
               }
             )
-          val nextPageToken =
-            if (response.hasNextPageToken()) response.nextPageToken else null
+          val nextPageToken = if (response.hasNextPageToken()) response.nextPageToken else null
           ResourceList(response.basicReportsList, nextPageToken)
         }
 
--- a/src/main/kotlin/org/wfanet/measurement/reporting/job/BasicReportsReportsJob.kt
+++ b/src/main/kotlin/org/wfanet/measurement/reporting/job/BasicReportsReportsJob.kt
@@ -106,9 +106,8 @@ class BasicReportsReportsJob(
           .measurementConsumerId
 
       val resourceLists =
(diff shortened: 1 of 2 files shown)


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4039)

SanjayVas (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4039)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on stevenwarejones).

PR #4045, opened 2026-06-19
feat: implement public ListEventGroupActivities RPC in Kingdom

getina:
## Summary
- Implements the public v2alpha `ListEventGroupActivities` RPC handler in the Kingdom server (`EventGroupActivitiesService.kt`).
- Translates the public request to the internal `ListEventGroupActivities` service and converts results back: parses the EventGroup parent (AIP-159 wildcard `-` supported for the EventGroup ID; DataProvider always required), authorizes the DataProvider principal, coerces page size, round-trips the opaque page token to/from the internal `ListEventGroupActivitiesPageToken`, and converts the `DateInterval` filter.
- Bumps `cross-media-measurement-api` to 0.95.0 (brings in the public `ListEventGroupActivities` proto from world-federation-of-advertisers/cross-media-measurement-api#281 and world-federation-of-advertisers/cross-media-measurement-api#283) and updates the lockfile.

Builds on internal service #3909 and public proto world-federation-of-advertisers/cross-media-measurement-api#281.

Issue: #4046


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4045)

src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/EventGroupActivitiesService.kt
```diff
@@ -227,6 +245,81 @@ class EventGroupActivitiesService(
...
+      eventGroupActivities +=
+        internalResponse.eventGroupActivitiesList.map { it.toEventGroupActivity() }
+      if (internalResponse.hasNextPageToken()) {
+        nextPageToken = internalResponse.nextPageToken.toByteString().base64UrlEncode()
+      }
+    }
+  }
+
+  companion object {
+    private const val DEFAULT_PAGE_SIZE = 1000
```
stevenwarejones: Default page size of 1000 disagrees with every sibling kingdom v2alpha list service (Certificates, ClientAccounts, EventGroupMetadataDescriptors all use 50; EventGroups uses 10) and with the internal SpannerEventGroupActivitiesService it wraps (50). A client that omits page_size gets the maximum page on every call, which makes pagination effectively a no-op for the default request and triggers 1000-row Spanner scans when the internal service was designed around 50.

The public proto comment "at most 1000 resources will be returned" is the ceiling, not the default — sibling services document the same ceiling and still pick a smaller default. Change DEFAULT_PAGE_SIZE to 50 (matching the internal service and the rest of the kingdom). Update the proto comment in a follow-up cmma PR to clarify default vs max if needed.

```kotlin
  private const val DEFAULT_PAGE_SIZE = 50
  private const val MAX_PAGE_SIZE = 1000
```

stevenwarejones (commented):
Issue 1: default page size.

src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/EventGroupActivitiesService.kt
```diff
@@ -227,6 +245,81 @@ class EventGroupActivitiesService(
...
+        }
+      }
+
+    val internalResponse: InternalListEventGroupActivitiesResponse =
+      try {
+        internalEventGroupActivitiesStub.listEventGroupActivities(internalRequest)
+      } catch (e: StatusException) {
+        throw when (e.status.code) {
+          Status.Code.NOT_FOUND -> Status.NOT_FOUND
+          else -> Status.INTERNAL
```
stevenwarejones: This remap is narrower than the other three methods in the same file (batchUpdate 121-128, delete 161-168, batchDelete 217-224), all of which propagate INVALID_ARGUMENT and FAILED_PRECONDITION through. With the current code, an INVALID_ARGUMENT from the internal service (e.g. `external_data_provider_id == 0`, negative page_size before the client-side gate, or any future internal validation) surfaces externally as INTERNAL — hiding a client error as a server failure.

Match the sibling pattern:

```kotlin
  throw when (e.status.code) {
    Status.Code.INVALID_ARGUMENT -> Status.INVALID_ARGUMENT
    Status.Code.FAILED_PRECONDITION -> Status.FAILED_PRECONDITION
    Status.Code.NOT_FOUND -> Status.NOT_FOUND
    else -> Status.UNKNOWN
  }.toExternalStatusRuntimeException(e)
```

The test at EventGroupActivitiesServiceTest.kt:940 (`listEventGroupActivities throws INTERNAL when internal service throws INVALID_ARGUMENT`) codifies the current incorrect behavior — flip it to assert INVALID_ARGUMENT.

stevenwarejones (commented):
Issue 2: error remap is narrower than sibling methods.

src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/EventGroupActivitiesService.kt
```diff
@@ -227,6 +245,81 @@ class EventGroupActivitiesService(
...
   }
+
+  override suspend fun listEventGroupActivities(
+    request: ListEventGroupActivitiesRequest
+  ): ListEventGroupActivitiesResponse {
+    val parentKey: EventGroupKey =
+      grpcRequireNotNull(EventGroupKey.fromName(request.parent)) {
+        "Parent is either unspecified or invalid"
+      }
+
```
stevenwarejones, line 256: The public EventGroup resource (cmma) declares an alternative pattern: `measurementConsumers/{mc}/eventGroups/{eg}`. The current code only parses the DataProvider-rooted form and rejects the MC-rooted form. Everything needed to support both is derivable from `parent` — no public API change required. This PR should:

1. Parse parent as either DP-rooted or MC-rooted EventGroup.
2. Authorize: DP principal must match the DP root; MC principal must match the MC root.
3. Convert the internal `ListEventGroupActivitiesRequest` (cmm-internal proto, same repo) to use a oneof scope:

```proto
   message ListEventGroupActivitiesRequest {
     oneof scope {
       fixed64 external_data_provider_id = 1;
       fixed64 external_measurement_consumer_id = 6;
     }
     // If unset, lists across all EventGroups within the chosen scope.
     fixed64 external_event_group_id = 2;
     // ... rest unchanged
   }
```

4. Internal service: validate `scope` is set (else INVALID_ARGUMENT), existence-check the scope-side actor (DP or MC), reader scopes by `MeasurementConsumers.ExternalMeasurementConsumerId` when MC scope is set.
5. Reader: add `externalMeasurementConsumerId: Long` param; conjunct on `MeasurementConsumers.ExternalMeasurementConsumerId = @...` when set.
6. Tests cover all four combinations: DP+specific, DP+wildcard, MC+specific, MC+wildcard, plus the auth matrix.

EventGroups already has `MeasurementConsumerId` so no schema change is needed for MC scoping, but a `(MeasurementConsumerId, ExternalEventGroupId)` index may be wanted for the MC+specific path (currently only `(DataProviderId, ExternalEventGroupId)` exists).

This is a meaningful expansion of this PR. If you'd rather land DP-only here and do the MC support as a stacked follow-up, that's fine — let me know which direction and I'll re-review accordingly.

stevenwarejones (commented):
Issue 3: support MC-rooted parent via oneof scope.

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4045)* status: all files reviewed, 3 unresolved discussions (waiting on getina).

src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/EventGroupActivitiesService.kt
getina: deferring to a subsequent pr. tracking issue subsequent pr. tracking issue https://github.com/world-federation-of-advertisers/cross-media-measurement/issues/4080

src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/EventGroupActivitiesService.kt
getina: done

src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/EventGroupActivitiesService.kt
getina: done
proto comment will be in a subsequent pr. tracking issue https://github.com/world-federation-of-advertisers/cross-media-measurement/issues/4080

getina (commented):
> @getina made 3 comments.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4045)* status: 2 of 4 files reviewed, 3 unresolved discussions (waiting on stevenwarejones).

src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/EventGroupActivitiesService.kt
stevenwarejones: can you put a TODO in the code to implement referencing the issue number?

stevenwarejones (commented):
> @stevenwarejones reviewed 3 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4045)* status: 3 of 4 files reviewed, 3 unresolved discussions (waiting on getina).

src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/EventGroupActivitiesService.kt
stevenwarejones: Verified — INVALID_ARGUMENT, FAILED_PRECONDITION, and NOT_FOUND now propagate; everything else maps to UNKNOWN. Test name and assertion are flipped to expect INVALID_ARGUMENT.

src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/EventGroupActivitiesService.kt
stevenwarejones: Verified — DEFAULT_PAGE_SIZE is now 50 in the companion object and the test constant matches. Thanks for filing #4080 for the proto comment.

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4045)* status: 3 of 4 files reviewed, 3 unresolved discussions (waiting on stevenwarejones).

stevenwarejones (commented):
> @stevenwarejones reviewed 1 file and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4045)* status: all files reviewed, 1 unresolved discussion (waiting on stevenwarejones+AGNT).

src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/EventGroupActivitiesService.kt
```diff
@@ -227,6 +245,83 @@ class EventGroupActivitiesService(
...
+    val resolvedPageSize: Int =
+      if (request.pageSize == 0) DEFAULT_PAGE_SIZE else request.pageSize.coerceAtMost(MAX_PAGE_SIZE)
+
+    val pageToken: ListEventGroupActivitiesPageToken? =
+      if (request.pageToken.isEmpty()) {
+        null
+      } else {
+        try {
+          ListEventGroupActivitiesPageToken.parseFrom(request.pageToken.base64UrlDecode())
+        } catch (e: IOException) {
```
stevenwarejones, line 281: The catch covers only IOException, but `String.base64UrlDecode()` calls `Base64.getUrlDecoder().decode(...)`, which throws `IllegalArgumentException` for non-base64 input (e.g. `"$$$"`). The existing "malformed page_token" test uses `"not-a-valid-token"` — all hyphen + lowercase letters, which are valid base64url characters, so the failure path it exercises is the protobuf parse one. A non-base64 token escapes the catch and surfaces externally as UNKNOWN instead of INVALID_ARGUMENT.

Catch both:

```kotlin
} catch (e: IOException) {
  throw Status.INVALID_ARGUMENT.withCause(e)
    .withDescription("page_token is malformed")
    .asRuntimeException()
} catch (e: IllegalArgumentException) {
  throw Status.INVALID_ARGUMENT.withCause(e)
    .withDescription("page_token is malformed")
    .asRuntimeException()
}
```

Add a separate test that sends a non-base64 token (e.g. `"$$$"`) and asserts INVALID_ARGUMENT, in addition to the existing protobuf-parse-failure test.

src/test/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/EventGroupActivitiesServiceTest.kt
```diff
@@ -63,6 +75,8 @@ private val EVENT_GROUP_NAME = "$DATA_PROVIDER_NAME/eventGroups/AAAAAAAAAHs"
 private val EVENT_GROUP_EXTERNAL_ID =
   apiIdToExternalId(EventGroupKey.fromName(EVENT_GROUP_NAME)!!.eventGroupId)
 private const val MODEL_PROVIDER_NAME = "modelProviders/AAAAAAAAAHs"
+private const val DEFAULT_PAGE_SIZE = 50
+private const val MAX_PAGE_SIZE = 1000
 
 @RunWith(JUnit4::class)
 class EventGroupActivitiesServiceTest {
```
stevenwarejones, line 85: nit: A few test gaps worth filling for `listEventGroupActivities`:

1. **Wildcard returns activities from multiple EventGroups, asserting names.** The only test that exercises name construction (`listEventGroupActivities returns EventGroupActivities`) uses a specific-EventGroup parent where the parent's `eventGroupId` happens to equal each row's `externalEventGroupId`, so a regression where `toEventGroupActivity` reads `parentKey.eventGroupId` instead of `source.externalEventGroupId` would pass it. Add a wildcard test that returns activities from two different EventGroups and asserts the public names use each row's own id.
2. **Wildcard authorization rejection.** `listEventGroupActivities throws PERMISSION_DENIED when principal is wrong` uses the specific-EventGroup parent. Add a parallel case with `dataProviders/<other>/eventGroups/-` to pin the auth check on the wildcard branch.
3. **Filter present but `date_interval` unset.** The contract is "empty Filter behaves like no filter". Add a test that sends `filter = {}` and verifies the internal request has no filter.
4. **`startDate` only / `endDate` only.** `DateInterval.toInternal()` conditionally copies each field; both branches are exercised together but never in isolation.

src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/EventGroupActivitiesService.kt
getina: done

src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/EventGroupActivitiesService.kt
getina: done

src/test/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/EventGroupActivitiesServiceTest.kt
getina: done

getina (commented):
> @getina made 3 comments and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4045)* status: all files reviewed, 2 unresolved discussions (waiting on stevenwarejones and stevenwarejones+AGNT).

src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/EventGroupActivitiesService.kt
stevenwarejones: Verified — TODO at EventGroupActivitiesService.kt:259 references #4080.

src/main/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/EventGroupActivitiesService.kt
stevenwarejones: Verified — second catch (IllegalArgumentException) added; new test sends `"$$$"` and asserts INVALID_ARGUMENT.

src/test/kotlin/org/wfanet/measurement/kingdom/service/api/v2alpha/EventGroupActivitiesServiceTest.kt
stevenwarejones: Verified — all four gaps covered, plus separate start_date-only and end_date-only tests. Wildcard test asserts the two activities resolve to distinct EventGroup-rooted names.

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4045)* status: 2 of 4 files reviewed, 2 unresolved discussions (waiting on stevenwarejones).

stevenwarejones (approved):
:lgtm:

> @stevenwarejones reviewed 2 files and all commit messages, made 1 comment, and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4045#-:-Ow3rAad5O0CUlrXTPcn:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on getina).

SanjayVas (approved):
> @SanjayVas reviewed 4 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4045)* status: all files reviewed, 1 unresolved discussion (waiting on getina).

___
*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4045#-Ow4Me920kHnR-2FU1E5:-Ow4Me920kHnR-2FU1E6:bxkhmox) (no related file):*
I think your references to `cross-media-measurement-api` in the PR description are incorrect. You need to qualify your reference when it's for another repo. e.g. `world-federation-of-advertisers/cross-media-measurement-api#281`

getina (commented):
> @getina made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4045)* status: all files reviewed, 1 unresolved discussion (waiting on SanjayVas).

___
*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4045#-Ow4Me920kHnR-2FU1E5:-Ow4ce880d_5xAtJCOZg:b-7ql1td) (no related file):*

I think your references to `cross-media-measurement-api` in the PR description are incorrect. You need to qualify your reference when it's for another repo. e.g. `world-federation-of-advertisers/cross-media-measurement-api#281`
</blockquote></details>

fixed

SanjayVas (approved):
> @SanjayVas resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4045)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on getina).

Issue #3980, opened 2026-06-12
VidLabelingJobService implementation


Marco-Premier:
Implement the public and internal Kotlin services for VidLabelingJobService, backed by the Spanner DAO.

PR #4044, opened 2026-06-18
feat: Implement PoolAssignmentJob internal and public API

getina:
Implement the internal Spanner service and public v1alpha PoolAssignmentJob service, covering Create, BatchCreate, Get, List, MarkSucceeded, and MarkFailed, with AIP-159 wildcard cross-upload List, per-element BatchCreate validation and batch dedup on (cmms_model_line, shard_index), request_id idempotency scoped to the parent upload, and timestamp-based etag optimistic concurrency.

Issue: #3893
Issue: #4173 
Issue: #4159

PR #4051, opened 2026-06-22
feat: Add VidLabelingJob public and internal API

Marco-Premier:
## Summary
  Implements the `VidLabelingJob` resource service (Phase-2 VID labeling) for the EDP
  Aggregator, with a public `v1alpha` API and an internal Cloud Spanner implementation.

  A `VidLabelingJob` is created (one per file batch) when a `RawImpressionUploadModelLine`
  transitions `RANKING → LABELING`. When the last job for a `(upload, model line)` reaches
  `SUCCEEDED`, the service signals which model lines just completed so the caller can flip
  the parent to `COMPLETED` and trigger `DataAvailabilitySync`.

  ## What's included
  - **Public service** (`service/v1alpha/VidLabelingJobService.kt`): `CreateVidLabelingJob`,
    `BatchCreateVidLabelingJobs`, `GetVidLabelingJob`, `ListVidLabelingJobs`,
    `MarkVidLabelingJobSucceeded`, `MarkVidLabelingJobFailed`.
  - **Internal Spanner service** (`deploy/gcloud/spanner/SpannerVidLabelingJobService.kt`)
    plus the DB layer (`deploy/gcloud/spanner/db/VidLabelingJob.kt`) and the
    `VidLabelingJobKey` / `RawImpressionUploadKey` resource keys.
  - Behavior: `request_id` (UUID4) idempotency on Create/BatchCreate and both Mark RPCs,
    `etag` optimistic concurrency on the Mark RPCs, structured `List` filter
    (state / create-time interval / model line), keyset pagination, and the `-` upload
    wildcard for cross-upload listing (AIP-159).
  - Per-model-line "last job out" completion is computed in a shared
    `computeCompletedModelLines(...)` and returned (the caller performs the state
    transition).

  ## Error mapping
  Internal errors are typed (carry `errorInfo`) and mapped at the public layer:
  - missing parent upload → `RawImpressionUploadNotFoundException` → `NOT_FOUND`
  - `CreateRequestId` collision → `VidLabelingJobAlreadyExistsException` → `ALREADY_EXISTS`
  - unknown job → `VID_LABELING_JOB_NOT_FOUND` → `NOT_FOUND`
  - bad state transition → `VID_LABELING_JOB_STATE_INVALID` → `FAILED_PRECONDITION`
  - etag mismatch → `ETAG_MISMATCH` → `ABORTED`

  Adding the two new `Errors.Reason` values required extending the exhaustive `when`
  blocks in the sibling EDPA v1alpha services (mapped to `INTERNAL`).

  ## Tests
  Abstract internal suite (run against the Spanner emulator) plus a public-API suite:
  Create/Get/List/BatchCreate/Mark happy paths and validation; `NOT_FOUND` for a missing
  parent upload (Create + BatchCreate, internal and public); BatchCreate idempotency,
  batch-size limit, and malformed sub-request `request_id`; full-response equality on
  idempotent `MarkVidLabelingJobSucceeded` replay (preserves `last_vid_labeling_job_result`);
  and `NOT_FOUND` for `MarkVidLabelingJobFailed`.

  ## ⚠️  Depends on base schema PR #3989
  This PR intentionally contains **no** schema change. It requires
  `world-federation-of-advertisers/cross-media-measurement#3989` to:
  1. **add** the `VidLabelingJob.MarkRequestId` column (used for Mark idempotency), and
  2. **remove** the unused `VidLabelingJob.Etag` column (the service derives the `etag`
     from `UpdateTime` and no longer writes the column).

  Until #3989 is merged/rebased in, the Spanner-emulator tests for this service will fail
  (`MarkRequestId` not yet present; `Etag` still `NOT NULL`). `TODO(...#3989)` markers note
  the dependency in the code.

  ## Notes for reviewers
  - Stacked on `getina/vid-labeling-spanner-schema`; `RawImpressionUploadKey` and the
    upload-id resolver are defined locally here (the base branch lacks them) and will be
    de-duplicated on rebase.
  - Please add an `Issue:` trailer for the tracking issue before merge (per dev-standards).

Issue: #3980


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4051)

src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/spanner/db/VidLabelingJob.kt
```diff
@@ -0,0 +1,380 @@
...
+    set("VidLabelingJobId").to(vidLabelingJobId)
+    set("State").to(state)
+    set("UpdateTime").to(Value.COMMIT_TIMESTAMP)
+    if (block != null) {
+      block()
+    }
+  }
+}
+
+// TODO(world-federation-of-advertisers/cross-media-measurement#3989): `MarkRequestId` is added and
```
stevenwarejones: Schema mismatch is structurally worse than #4052. This PR doesn't touch the schema at all, but the code is pinned to a hypothetical future schema in #3989. Two distinct problems make every test in this PR fail today, not just Mark tests:

1. `insertVidLabelingJob` (db/VidLabelingJob.kt:288-310) does NOT set `Etag`, but base schema declares `Etag STRING(36) NOT NULL`. **Every Create insert throws NULL constraint violation.** That's failure on test #1, not just on MarkSucceeded.
2. `MarkRequestId` is read at db/VidLabelingJob.kt:348, 377 and written at SpannerVidLabelingJobService.kt:480, 589 — column doesn't exist in base schema.

Two options:

**Option A — Include the schema changes in this PR.** You've already touched the schema mid-stack on #4052 (you added `MarkRequestId` to `PoolAssignmentJob` there). Same precedent applies here:

```sql
ALTER TABLE VidLabelingJob ADD COLUMN MarkRequestId STRING(36);
ALTER TABLE VidLabelingJob DROP COLUMN Etag;
```

Or edit the existing `CREATE TABLE` in the schema file directly. Plus the Liquibase changeset bump. This makes the PR self-contained and unblocks tests immediately.

**Option B — Revert the code paths that depend on the schema delta.** Drop the MarkRequestId reads/writes (effectively shipping Mark RPCs without idempotency — a regression vs. the design intent). Set `Etag` to empty string on insert to satisfy the NOT NULL constraint. MarkSucceeded/MarkFailed idempotency lands in a follow-up PR stacked on #3989-with-changes.

Option A is cleaner. What's not okay is the current state: code reads/writes a column that doesn't exist AND fails to set a column that the schema requires. CI hasn't run yet (no checks reported on the head SHA), so the failures aren't visible — but `bazel test` against current base is fully red.

stevenwarejones (commented):
Schema mismatch — Etag NOT NULL violated on insert, MarkRequestId column missing. Option A (add schema here) or Option B (revert idempotency).

src/main/kotlin/org/wfanet/measurement/edpaggregator/service/v1alpha/VidLabelingJobService.kt
```diff
@@ -0,0 +1,483 @@
...
+    val jobKey =
+      VidLabelingJobKey.fromName(request.name)
+        ?: throw InvalidFieldValueException("name")
+          .asStatusRuntimeException(Status.Code.INVALID_ARGUMENT)
+
+    if (request.etag.isEmpty()) {
+      throw RequiredFieldNotSetException("etag")
+        .asStatusRuntimeException(Status.Code.INVALID_ARGUMENT)
+    }
+    if (request.requestId.isNotEmpty()) {
```
stevenwarejones: Same shape as my comment on #4052's `MarkRankerJobSucceeded` — `request_id` is optional on BOTH `MarkVidLabelingJobSucceeded` (v1alpha:320-327, internal:445) and `MarkVidLabelingJobFailed` (v1alpha:371-378, internal:568). The idempotency check only fires when both sides have a request_id; without one, Pub/Sub redelivery falls through to `checkMarkPrecondition` and hits the poison-pill path:

- `MarkSucceeded` after already-SUCCEEDED: not in `validPreviousStates = {CREATED, FAILED}` → `FAILED_PRECONDITION`. Infinite poison-pill on every redelivery.
- `MarkFailed` after already-FAILED: passes the state check (FAILED IS in the valid set), but re-runs the update — bumps `UpdateTime`, overwrites `ErrorMessage`, writes `MarkRequestId`. Idempotent on the outcome but not on the operation.

Worse than #4052 — this PR adds idempotency to BOTH Mark RPCs (a deliberate design improvement over #4052), so a missing `request_id` defeats the safety net you explicitly built.

Make `request_id` REQUIRED on both Mark RPCs:

1. **Proto**: add `[(google.api.field_behavior) = REQUIRED]` to `MarkVidLabelingJobSucceededRequest.request_id` AND `MarkVidLabelingJobFailedRequest.request_id`.
2. **Public service** (`VidLabelingJobService.kt:320` and `:371`): reject empty before the UUID-format check:
   ```kotlin
   if (request.requestId.isEmpty()) {
     throw RequiredFieldNotSetException("request_id")
   }
   ```
3. **Internal service** (`SpannerVidLabelingJobService.kt:validateMarkRequest`): pull the empty check into the validator.
4. **Tests**: add tests that empty request_id → INVALID_ARGUMENT on both Mark RPCs.

Same rationale as #4052 and the precedent you set on `RawImpressionUploadFile.batchCreate` in #4014: state-transition operations the design depends on being idempotent need the idempotency key to actually be required.

stevenwarejones (commented):
request_id should be REQUIRED on both Mark RPCs — see inline.

stevenwarejones (commented):
nit: vlj prefix is opaque — use vidLabelingJob.

src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/spanner/SpannerVidLabelingJobService.kt
```diff
@@ -0,0 +1,732 @@
...
+    if (request.vidLabelingJob.cmmsModelLinesList.isEmpty()) {
+      throw RequiredFieldNotSetException("vid_labeling_job.cmms_model_lines")
+    }
+    if (request.vidLabelingJob.rawImpressionUploadFilesList.isEmpty()) {
+      throw RequiredFieldNotSetException("vid_labeling_job.raw_impression_upload_files")
+    }
+  }
+
+  companion object {
+    private const val VID_LABELING_JOB_RESOURCE_ID_PREFIX = "vlj"
```
stevenwarejones: nit: `"vlj"` is opaque. Same as the `"rj"` nit I posted on #4052 — your own recently-landed PRs (`SpannerRawImpressionUploadService` → `"rawImpressionUpload"`, `SpannerRawImpressionUploadFileService` → `"rawImpressionUploadFile"`) use the full resource type name. Match that here:

```kotlin
private const val VID_LABELING_JOB_RESOURCE_ID_PREFIX = "vidLabelingJob"
```

A resource ID like `vidLabelingJob-3c8f...` reads clearly in logs and Cloud Console; `vlj-3c8f...` requires the reader to know the abbreviation.

src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/spanner/db/VidLabelingJob.kt
```diff
@@ -0,0 +1,380 @@
...
+fun AsyncDatabaseClient.TransactionContext.insertVidLabelingJob(
+  rawImpressionUploadId: Long,
+  vidLabelingJobId: Long,
+  vidLabelingJobResourceId: String,
+  dataProviderResourceId: String,
+  cmmsModelLines: List<String>,
+  rawImpressionUploadFiles: List<String>,
+  createRequestId: String,
+) {
+  bufferInsertMutation("VidLabelingJob") {
```
stevenwarejones, line 311: Cross-link to my comment on #3989 about inconsistent `Etag` handling across the four job tables. VidLabelingJob is the worst case — `insertVidLabelingJob` (line 297-310) does NOT set `Etag` AT ALL, but the schema declares `Etag STRING(36) NOT NULL`. **Every Create insert throws NULL constraint violation today.**

Pick the explicit-write pattern (the one `RawImpressionUploadModelLine` and `PoolAssignmentJob` already use): compute the etag inside the transaction (e.g. `ETags.computeETag(...)` of a stable input) and write it as the actual value. Match what the schema's `NOT NULL` claims to enforce.

Rationale (full version on #3989 thread):
- Schema honesty: `NOT NULL` is meaningless if the column is never populated.
- Forward compatibility: the column can later carry non-UpdateTime semantics (content hash, schema version).
- Silent regression: any future bumping of `UpdateTime` for non-semantic reasons silently invalidates client etag caches. Explicit writes don't.

This obviates my Option B from the schema-mismatch comment above (drop Etag from VidLabelingJob). Option A (add MarkRequestId here, write Etag explicitly) is the right call. Don't drop Etag — populate it.

Same fix needed on the Mark path: `updateVidLabelingJobState` and its callers in `markVidLabelingJobSucceeded` / `markVidLabelingJobFailed`.

stevenwarejones (commented):
Etag column NOT NULL but not set — write explicitly. See #3989 thread.

src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/spanner/db/VidLabelingJob.kt
stevenwarejones: **Correcting myself**: ignore the "drop `Etag` column" part of Option A above. My newer Etag comment (cross-link 3453854522, on db/VidLabelingJob.kt:297) supersedes it: keep `Etag STRING(36) NOT NULL` and write it explicitly on insert/update — matching `RawImpressionUploadModelLine` and `PoolAssignmentJob` and what the schema's `NOT NULL` constraint claims.

So the corrected Option A is:

1. **Schema (#3989)**: add `MarkRequestId STRING(36)` to VidLabelingJob. Keep `Etag STRING(36) NOT NULL` as-is.
2. **This PR**: write `Etag` explicitly in `insertVidLabelingJob` (line 297-310) and `updateVidLabelingJobState`. Drop the derive-from-UpdateTime read path.

Option B (revert idempotency code paths) is still a valid alternative if you don't want to add the column dependency on #3989 in this PR's stack.

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4051)* status: 0 of 24 files reviewed, 4 unresolved discussions (waiting on getina, Marco-Premier, and stevenwarejones).

stevenwarejones:
Make `request_id` REQUIRED on every state-mutating RPC in this service: `CreateVidLabelingJob`, `BatchCreateVidLabelingJobs` (per-element), and every `Mark*` RPC. Same for `etag` on Marks (currently REQUIRED, keep it that way).

Rationale:
- etag is the optimistic-concurrency contract — clients must declare what state they think they're acting on. Optional etag is a blind write, which has no place in a pipeline where every transition is gated by precondition.
- request_id is the AIP-155 idempotency contract — clients must provide an at-most-once token so retries are safe. Optional request_id means a network-retried call silently re-runs or silently fails depending on server state.

Proto changes (`vid_labeling_job_service.proto`):
- `CreateVidLabelingJobRequest.request_id`: OPTIONAL → REQUIRED + UUID4
- `BatchCreateVidLabelingJobsRequest`: each element's `request_id` REQUIRED
- Every `Mark*Request`: add `string request_id = N [REQUIRED, UUID4]` (currently absent on Marks)

Public + internal layers: reject empty `request_id` and `etag` with INVALID_ARGUMENT before forwarding.

This is v1alpha with no production clients, so the breaking change is free. Cross-posting on #4018, #4044, #4052 — keeping all four services consistent. Depends on Mark RPCs adopting MarkRequestId-based idempotency per the #3989 schema feedback.

src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/spanner/db/VidLabelingJob.kt
Marco-Premier: Done.

src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/spanner/db/VidLabelingJob.kt
Marco-Premier: Done.

src/main/kotlin/org/wfanet/measurement/edpaggregator/service/v1alpha/VidLabelingJobService.kt
Marco-Premier: Done.

Marco-Premier (commented):
> @Marco-Premier made 3 comments and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4051)* status: 0 of 25 files reviewed, 3 unresolved discussions (waiting on getina, stevenwarejones, and stevenwarejones+AGNT).

src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/spanner/SpannerVidLabelingJobService.kt
```diff
@@ -0,0 +1,729 @@
...
+
+        txn.updateVidLabelingJobState(
+          dataProviderResourceId = request.dataProviderResourceId,
+          rawImpressionUploadId = result.rawImpressionUploadId,
+          vidLabelingJobId = result.vidLabelingJobId,
+          state = State.VID_LABELING_STATE_SUCCEEDED,
+          etag = newEtag,
+        ) {
+          set("MarkRequestId").to(request.requestId)
+        }
```
stevenwarejones, line 462: Clear `ErrorMessage` on the FAILED → SUCCEEDED transition. `VALID_MARK_PREVIOUS_STATES` (Spanner.kt:658) allows a FAILED row to retry into SUCCEEDED, but neither the buffered update here nor the response copy at Spanner.kt:501-506 clears `ErrorMessage`. Proto says `error_message` is set when state is FAILED — a succeeded row carrying stale failure text contradicts that.

```
txn.updateVidLabelingJobState(
  ...,
  state = State.VID_LABELING_STATE_SUCCEEDED,
  etag = newEtag,
) {
  set("MarkRequestId").to(request.requestId)
  set("ErrorMessage").to(null as String?)
}

// response copy:
currentJob.copy {
  state = State.VID_LABELING_STATE_SUCCEEDED
  etag = newEtag
  clearErrorMessage()
  clearUpdateTime()
}
```

Same fix landed on #4044 M7 (-OvV8E-V-6P86CPzIOvF) and #4052 (-OvVTu7z6OAih9sSMpx0 area). Add a test: CREATED → markFailed → markSucceeded, assert returned `error_message` is empty.

src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/spanner/SpannerVidLabelingJobService.kt
```diff
@@ -0,0 +1,729 @@
...
+
+        // Read all jobs for this upload to detect which model lines have all their jobs done.
+        // The buffered update above is not visible in this read, so the current job still
+        // shows its previous state; we treat it as SUCCEEDED via its ID below.
+        val allJobsForUpload: List<VidLabelingJobResult> = buildList {
+          txn
+            .readVidLabelingJobs(
```
stevenwarejones: Replace the `limit = Int.MAX_VALUE` scan with N targeted COUNT queries (one per model line on this job). At scale (hundreds of VidLabelingJobs per upload), reading every job's full row just to ask 'which model lines are now complete' inflates the transaction's read set and drives unnecessary Spanner CPU. Same fix applies to the replay path at Spanner.kt:448.

`computeCompletedModelLines` (Spanner.kt:682-695) iterates only over `currentJob.cmmsModelLinesList`, so a single COUNT per model line suffices:

```
suspend fun AsyncDatabaseClient.ReadContext.countOtherNonSucceededVidLabelingJobsForModelLine(
  dataProviderResourceId: String,
  rawImpressionUploadId: Long,
  cmmsModelLine: String,
  excludeVidLabelingJobId: Long,
): Long {
  // SELECT COUNT(*) FROM VidLabelingJob
  // WHERE DataProviderResourceId=@dp AND RawImpressionUploadId=@up
  //   AND @cmmsModelLine IN UNNEST(CmmsModelLines)
  //   AND VidLabelingJobId != @excludeId
  //   AND CAST(State AS INT64) != @succeededState
}

// Spanner.kt — drop computeCompletedModelLines, just:
val completedModelLines = currentJob.cmmsModelLinesList.filter { modelLine ->
  txn.countOtherNonSucceededVidLabelingJobsForModelLine(
    request.dataProviderResourceId, result.rawImpressionUploadId,
    modelLine, result.vidLabelingJobId,
  ) == 0L
}
```

Same pattern as #4052 (-OvV7_HP1PFz88jFeWGE) and #4044 M6, adapted for the multi-modelline VidLabelingJob shape. The buffered state update for the current row isn't visible in the read, so excluding by ID is the correct semantics.

stevenwarejones (commented):
> @stevenwarejones reviewed 26 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4051)* status: all files reviewed, 5 unresolved discussions (waiting on getina, Marco-Premier, and stevenwarejones+AGNT).

src/main/kotlin/org/wfanet/measurement/edpaggregator/service/v1alpha/VidLabelingJobService.kt
```diff
@@ -0,0 +1,491 @@
...
+              if (uploadKey.rawImpressionUploadId == WILDCARD_ID) ""
+              else uploadKey.rawImpressionUploadId
+            this.pageSize = pageSize
+            if (internalPageToken != null) {
+              pageToken = internalPageToken
+            }
+            if (request.hasFilter()) {
+              filter =
+                InternalListRequestKt.filter {
+                  if (request.filter.state != VidLabelingJob.State.STATE_UNSPECIFIED) {
```
stevenwarejones, line 288: Reject `UNRECOGNIZED` here too. The current guard only handles `STATE_UNSPECIFIED`; an `UNRECOGNIZED` value (future enum from a newer client, or a corrupt request) falls through to `toInternal()` (line 488), throws `IllegalStateException`, and surfaces to the client as `UNKNOWN` instead of `INVALID_ARGUMENT`.

```
if (request.filter.state == VidLabelingJob.State.UNRECOGNIZED) {
  throw InvalidFieldValueException("filter.state")
}
if (request.filter.state != VidLabelingJob.State.STATE_UNSPECIFIED) {
  state = request.filter.state.toInternal()
}
```

Same fix shape as #4044 M11 (-OvV99Cp62-_LbCbCVIA), adapted for the single-state filter this PR uses (vs the `state_in` list on the others).

src/main/kotlin/org/wfanet/measurement/edpaggregator/service/v1alpha/VidLabelingJobService.kt
```diff
@@ -0,0 +1,491 @@
...
+      RawImpressionUploadKey.fromName(request.parent)
+        ?: throw InvalidFieldValueException("parent")
+          .asStatusRuntimeException(Status.Code.INVALID_ARGUMENT)
+
+    if (request.requestsList.isEmpty()) {
+      throw RequiredFieldNotSetException("requests")
+        .asStatusRuntimeException(Status.Code.INVALID_ARGUMENT)
+    }
+
+    if (request.requestsList.size > MAX_BATCH_SIZE) {
```
stevenwarejones, line 145: nit: include the field-message body for consistency with #4018, #4044, #4052. The current exception drops the helpful "must contain at most 50" detail, so clients see only "requests has invalid value":

```
if (request.requestsList.size > MAX_BATCH_SIZE) {
  throw InvalidFieldValueException("requests") {
      "$it must contain at most $MAX_BATCH_SIZE elements"
    }
}
```

src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/spanner/db/VidLabelingJob.kt
stevenwarejones: Verified — explicit-write etag throughout. Insert at Db.kt:307 (`set(\"Etag\").to(etag)`), update at Db.kt:327, buildResult reads from the column at Db.kt:373. Matches the explicit-write pattern across #3989 / #4018 / #4044 / #4052.

src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/spanner/db/VidLabelingJob.kt
stevenwarejones: Verified — `insertVidLabelingJob` (Db.kt:307) now sets `Etag` from the constructor parameter, the schema's NOT NULL constraint is satisfied. Plus the schema commit (3e6e90090) added the `MarkRequestId` column at create-vid-labeling-schema.sql, unblocking the Mark idempotency tests.

src/main/kotlin/org/wfanet/measurement/edpaggregator/service/v1alpha/VidLabelingJobService.kt
stevenwarejones: Verified — `MarkVidLabelingJobSucceededRequest.request_id` (public.proto:212-215) and `MarkVidLabelingJobFailedRequest.request_id` are both REQUIRED + UUID4. Validated at the public layer via `validateMarkRequest` (Spanner.kt:608+) which rejects empty/non-UUID with INVALID_ARGUMENT. Internal replay check simplified to `result.markRequestId == request.requestId` (no `isNotEmpty` guard needed).

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4051)* status: all files reviewed, 7 unresolved discussions (waiting on getina, Marco-Premier, and stevenwarejones).

stevenwarejones (commented):
all these top-level comments look good. posted some more nits.

> @stevenwarejones made 1 comment and resolved 3 discussions.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4051#-:-Ovo_iK7-oPRyug-lbM0:bcnfm6l)* status: all files reviewed, 4 unresolved discussions (waiting on getina and Marco-Premier).

src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/spanner/SpannerVidLabelingJobService.kt
Marco-Premier: Done.

src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/spanner/SpannerVidLabelingJobService.kt
Marco-Premier: Done.

src/main/kotlin/org/wfanet/measurement/edpaggregator/service/v1alpha/VidLabelingJobService.kt
Marco-Premier: Done.

Marco-Premier (commented):
> @Marco-Premier made 3 comments and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4051)* status: 22 of 26 files reviewed, 3 unresolved discussions (waiting on getina, stevenwarejones, and stevenwarejones+AGNT).

src/main/kotlin/org/wfanet/measurement/edpaggregator/service/v1alpha/VidLabelingJobService.kt
```diff
@@ -0,0 +1,497 @@
...
+      InternalErrors.Reason.RAW_IMPRESSION_METADATA_BATCH_FILE_ALREADY_EXISTS,
+      InternalErrors.Reason.REQUISITION_METADATA_NOT_FOUND,
+      InternalErrors.Reason.REQUISITION_METADATA_NOT_FOUND_BY_CMMS_REQUISITION,
+      InternalErrors.Reason.REQUISITION_METADATA_ALREADY_EXISTS,
+      InternalErrors.Reason.REQUISITION_METADATA_ALREADY_EXISTS_BY_BLOB_URI,
+      InternalErrors.Reason.REQUISITION_METADATA_ALREADY_EXISTS_BY_CMMS_REQUISITION,
+      InternalErrors.Reason.REQUISITION_METADATA_STATE_INVALID,
+      InternalErrors.Reason.REQUIRED_FIELD_NOT_SET,
+      InternalErrors.Reason.INVALID_FIELD_VALUE,
+      null -> Status.INTERNAL.withCause(e).asRuntimeException()
```
stevenwarejones, line 432: Map `REQUIRED_FIELD_NOT_SET` and `INVALID_FIELD_VALUE` to `INVALID_ARGUMENT`, not `INTERNAL`. Currently lines 430-432 route both into the `null -> INTERNAL` arm — if the internal layer ever throws one of these (e.g. via a future change that adds internal validation past what the public layer covers), clients see `INTERNAL` instead of `INVALID_ARGUMENT`.

```
InternalErrors.Reason.REQUIRED_FIELD_NOT_SET,
InternalErrors.Reason.INVALID_FIELD_VALUE ->
  Status.INVALID_ARGUMENT.withCause(e).asRuntimeException()
```

Same bug fixed in #4058 commit a485e21c5; apply the same shape here.

src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/spanner/SpannerVidLabelingJobService.kt
stevenwarejones: Verified in commit 91d044979. Spanner.kt:479-481 buffers `set("ErrorMessage").to(null as String?)` on the SUCCEEDED transition, and the response copy at line 497-503 calls `clearErrorMessage()`. FAILED → SUCCEEDED retry no longer carries stale failure text.

src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/spanner/SpannerVidLabelingJobService.kt
stevenwarejones: Verified in commit 91d044979. `countOtherNonSucceededVidLabelingJobsForModelLine` added to Db.kt and wired at Spanner.kt:451 (replay path) and Spanner.kt:488 (forward path). Iterates only over `currentJob.cmmsModelLinesList` — one COUNT per model line on the current job, instead of the full-row scan.

src/main/kotlin/org/wfanet/measurement/edpaggregator/service/v1alpha/VidLabelingJobService.kt
stevenwarejones: Verified in commit 91d044979. Public.kt:146-148 now includes the helpful field-message body (`\"$it must contain at most $MAX_BATCH_SIZE elements\"`). Matches the consistent shape across #4018, #4044, #4052.

src/main/kotlin/org/wfanet/measurement/edpaggregator/service/v1alpha/VidLabelingJobService.kt
stevenwarejones: Verified in commit 91d044979. Public.kt:284-287 now rejects `UNRECOGNIZED` with `INVALID_ARGUMENT` before falling into `toInternal()`. Future enum values from newer clients now surface correctly.

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4051)* status: 22 of 26 files reviewed, 4 unresolved discussions (waiting on getina, Marco-Premier, and stevenwarejones).

stevenwarejones (commented):
> @stevenwarejones reviewed 4 files and all commit messages, and resolved 3 discussions.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4051)* status: all files reviewed, 1 unresolved discussion (waiting on getina and Marco-Premier).

src/main/kotlin/org/wfanet/measurement/edpaggregator/service/v1alpha/VidLabelingJobService.kt
Marco-Premier: Done.

Marco-Premier (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4051)* status: 25 of 26 files reviewed, 1 unresolved discussion (waiting on getina, stevenwarejones, and stevenwarejones+AGNT).

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4051)* status: all files reviewed, 1 unresolved discussion (waiting on getina and stevenwarejones+AGNT).

src/main/resources/edpaggregator/spanner/create-vid-labeling-schema.sql
```diff
@@ -0,0 +1,279 @@
...
+  RawImpressionUploadModelLineId INT64 NOT NULL,
+  RawImpressionUploadModelLineResourceId STRING(63) NOT NULL,
+  CmmsModelLine STRING(MAX) NOT NULL,
+  CreateRequestId STRING(36),
+  State `wfa.measurement.internal.edpaggregator.RawImpressionUploadModelLineState` NOT NULL,
+  ErrorMessage STRING(MAX),
+  PoolOffsets ARRAY<INT64>,
+  MaxEventDate DATE,
+  EncryptedMergedDek BYTES(MAX),
+  Etag STRING(36) NOT NULL,
```
stevenwarejones: Rebase needed — this branch is on a stale snapshot of `getina/vid-labeling-spanner-schema`. The base branch has added `MarkRequestId` columns + matching unique indexes on RawImpressionUploadModelLine and RankerJob, the four `*ByMarkRequestId` UNIQUE indexes (PoolAssignmentJob, RankerJob, VidLabelingJob, RawImpressionUploadModelLine), and the `RankIndexBlobByNaturalKey` unique index — all missing from this PR's snapshot. A `git fetch origin getina/vid-labeling-spanner-schema && git rebase origin/getina/vid-labeling-spanner-schema` should bring them back.

This PR's own service code compiles against the schema as-is (VidLabelingJob.MarkRequestId is present at line 260), but downstream PRs (#4052 RankerJob, future #4018 RawImpressionUploadModelLine wiring) need the columns this PR is silently dropping if it lands first.

stevenwarejones (commented):
Rebase needed.

src/main/kotlin/org/wfanet/measurement/edpaggregator/service/v1alpha/VidLabelingJobService.kt
stevenwarejones: Verified in commit 3d5f4433b. `REQUIRED_FIELD_NOT_SET` and `INVALID_FIELD_VALUE` now map to `INVALID_ARGUMENT` (VidLabelingJobService.kt:439-441), separated from the `null -> INTERNAL` arm at line 430. Closing.

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4051)* status: all files reviewed, 2 unresolved discussions (waiting on getina, Marco-Premier, and stevenwarejones).

stevenwarejones (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4051#-:-Ovu7tSY9OngKtbRwrbU:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on getina and Marco-Premier).

Marco-Premier (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4051)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on getina, SanjayVas, and stevenwarejones).

src/main/resources/edpaggregator/spanner/create-vid-labeling-schema.sql
stevenwarejones: Verified — rebased to commit c87e69d6. Schema is now identical to the base branch (`getina/vid-labeling-spanner-schema`), so all MarkRequestId columns + indexes are back. This PR's only contribution to the schema file is now empty (no diff). Closing.

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4051)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on getina, SanjayVas, and stevenwarejones).

stevenwarejones (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4051)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on getina and SanjayVas).

SanjayVas (changes requested):
> @SanjayVas reviewed 6 files and made 4 comments.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4051)* status: all files reviewed, 4 unresolved discussions (waiting on getina and Marco-Premier).

___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/spanner/db/VidLabelingJob.kt` line 43 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4051#-OvvxE2kA9KNEJxw8Er2:-OvvxE2kA9KNEJxw8Er3:b-284pqe) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c87e69d6c2e1fdbf77112ac66b7137e8e9d73414/src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/spanner/db/VidLabelingJob.kt#L43)):*
> ```Kotlin
>   val rawImpressionUploadId: Long,
>   val vidLabelingJobId: Long,
>   val markRequestId: String,
> ```

Is this for the mark succeeded method or the mark failed method? Request IDs are scoped to the corresponding method, so these two need to be differentiated.
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/spanner/db/VidLabelingJob.kt` line 64 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4051#-OvvxlLg5Imb0g9SSfIC:-OvvxlLhAn0qlpJJj2bx:b-7s4dj3) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c87e69d6c2e1fdbf77112ac66b7137e8e9d73414/src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/spanner/db/VidLabelingJob.kt#L64)):*
> ```Kotlin
>  */
> suspend fun AsyncDatabaseClient.ReadContext.getRawImpressionUploadIdForVidLabeling(
> ```

This function seems like it should be defined in RawImpressionUpload.kt since that's the entity/table it's concerned with. I'm not sure what the "ForVidLabeling" is in the name and how that's salient when this is just trying to get an internal ID from resource IDs.
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/spanner/db/VidLabelingJob.kt` line 126 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4051#-OvvzKsl4OEUvigena2C:-OvvzKsl4OEUvigena2D:b-xiz2zd) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c87e69d6c2e1fdbf77112ac66b7137e8e9d73414/src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/spanner/db/VidLabelingJob.kt#L126)):*
> ```Kotlin
> 
> /** Finds an existing [VidLabelingJob] by request ID for idempotency. */
> suspend fun AsyncDatabaseClient.ReadContext.findVidLabelingJobByRequestId(
> ```

Need to disambiguate between the methods as each has its own request ID.



_Suggestion:_
```Kotlin
CreateRequestId
```
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/spanner/db/VidLabelingJob.kt` line 160 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4051#-OvvznNH7B0bFPAu_Tnm:-OvvznNH7B0bFPAu_Tnn:b45h) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c87e69d6c2e1fdbf77112ac66b7137e8e9d73414/src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/spanner/db/VidLabelingJob.kt#L160)):*
> ```Kotlin
> 
> /** Finds existing [VidLabelingJob] entries by request IDs for batch idempotency. */
> suspend fun AsyncDatabaseClient.ReadContext.findVidLabelingJobsByRequestIds(
> ```





_Suggestion:_
```Kotlin
CreateRequestIds
```

Marco-Premier (commented):
> @Marco-Premier made 4 comments.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4051)* status: 17 of 29 files reviewed, 4 unresolved discussions (waiting on getina, SanjayVas, and stevenwarejones).

___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/spanner/db/VidLabelingJob.kt` line 43 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4051#-OvvxE2kA9KNEJxw8Er2:-OvyaFYt6Iq8cMe6Xj7Z:b-gpqb6l) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c87e69d6c2e1fdbf77112ac66b7137e8e9d73414/src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/spanner/db/VidLabelingJob.kt#L43)):*

Is this for the mark succeeded method or the mark failed method? Request IDs are scoped to the corresponding method, so these two need to be differentiated.
</blockquote></details>

I think the same comment is on #3989 already. Will address this once the comment supports this columns from #3989 
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/spanner/db/VidLabelingJob.kt` line 64 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4051#-OvvxlLg5Imb0g9SSfIC:-OvyaO89A7g7-h7_5vyt:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c87e69d6c2e1fdbf77112ac66b7137e8e9d73414/src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/spanner/db/VidLabelingJob.kt#L64)):*

This function seems like it should be defined in RawImpressionUpload.kt since that's the entity/table it's concerned with. I'm not sure what the "ForVidLabeling" is in the name and how that's salient when this is just trying to get an internal ID from resource IDs.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/spanner/db/VidLabelingJob.kt` line 126 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4051#-OvvzKsl4OEUvigena2C:-OvyaOVb2ObmQEDbQUoU:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c87e69d6c2e1fdbf77112ac66b7137e8e9d73414/src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/spanner/db/VidLabelingJob.kt#L126)):*

Need to disambiguate between the methods as each has its own request ID.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/spanner/db/VidLabelingJob.kt` line 160 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4051#-OvvznNH7B0bFPAu_Tnm:-OvyaP1KDLKpwJ9KtMJ0:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c87e69d6c2e1fdbf77112ac66b7137e8e9d73414/src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/spanner/db/VidLabelingJob.kt#L160)):*
> ```Kotlin
> 
> /** Finds existing [VidLabelingJob] entries by request IDs for batch idempotency. */
> suspend fun AsyncDatabaseClient.ReadContext.findVidLabelingJobsByRequestIds(
> ```

Done.

SanjayVas (commented):
> @SanjayVas reviewed 25 files and all commit messages, and resolved 3 discussions.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4051)* status: 28 of 29 files reviewed, 1 unresolved discussion (waiting on getina and stevenwarejones).

Marco-Premier (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4051)* status: 25 of 29 files reviewed, 1 unresolved discussion (waiting on getina, SanjayVas, and stevenwarejones).

___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/spanner/db/VidLabelingJob.kt` line 43 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4051#-OvvxE2kA9KNEJxw8Er2:-Ow2XJt32vI2Coo3ngvZ:bb6un1j) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/c87e69d6c2e1fdbf77112ac66b7137e8e9d73414/src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/spanner/db/VidLabelingJob.kt#L43)):*

I think the same comment is on #3989 already. Will address this once the comment supports this columns from #3989
</blockquote></details>

Done

SanjayVas (approved):
> @SanjayVas reviewed 6 files and all commit messages, made 1 comment, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4051)* status: all files reviewed, 1 unresolved discussion (waiting on getina, Marco-Premier, and stevenwarejones).

___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/spanner/db/VidLabelingJob.kt` line 388 at r9](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4051#-Ow4IqaIE7lBN-0vyBc4:-Ow4IqaJ9t_popkqQfL-:byzw2hj) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/5073733ecf3a0c6fc60ec295a5946f9dc53276eb/src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/spanner/db/VidLabelingJob.kt#L388)):*
> ```Kotlin
>       struct.getLong("RawImpressionUploadId"),
>       struct.getLong("VidLabelingJobId"),
>       if (struct.isNull("MarkSucceededRequestId")) ""
> ```

nit: Braces are required if the `if` expression doesn't fit on one line.

https://developer.android.com/kotlin/style-guide#braces

Marco-Premier (commented):
> @Marco-Premier reviewed 23 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4051)* status: all files reviewed, 1 unresolved discussion (waiting on getina and stevenwarejones).

Marco-Premier (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4051)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on getina and stevenwarejones).

Issue #3927, opened 2026-06-05
RankJobService implementation


Marco-Premier:
Implement the public and internal Kotlin services for RankJobService, backed by the Spanner DAO.

PR #4052, opened 2026-06-22
feat(edpaggregator): Add RankerJob public and internal API

Marco-Premier:
## Summary

  Implements the **Phase-1 `RankerJob`** resource API for the VID Labeler
  memoized-assignment pipeline: a public `v1alpha` service plus an internal Cloud
  Spanner service, resource keys, internal error reasons, and Spanner-emulator
  tests.

  The last `RankerJob` to succeed for a `(upload, model line)` reports
  `is_last_job`, so the caller can transition the parent
  `RawImpressionUploadModelLine` from `RANKING` → `LABELING` and trigger Phase 2.

  > Stacked on `getina/vid-labeling-spanner-schema` (schema PR #3989). Please set
  > that branch as the base when opening the PR, and merge it first.

  ## What's included

  - **Public API** (`service/v1alpha/RankerJobService.kt`): `CreateRankerJob`,
    `BatchCreateRankerJobs`, `GetRankerJob`, `ListRankerJobs`,
    `MarkRankerJobSucceeded`, `MarkRankerJobFailed`.
  - **Internal API** (`deploy/gcloud/spanner/SpannerRankerJobService.kt` +
    `db/RankerJob.kt`): Spanner-backed implementation with `request_id`
    idempotency and `etag` optimistic concurrency.
  - **Resource keys** (`service/RankerJobKey.kt`, incl. `RawImpressionUploadKey`),
    new internal error reasons in `service/internal/Errors.kt`, and service
    wiring (`Services.kt`, `InternalApiServices.kt`, `IdVariable.kt`, BUILD files).
  - **Tests**: abstract internal contract test + concrete Spanner-emulator test,
    and a public-API `v1alpha` test.

  ## Behavior notes
  
  - `MarkRankerJobSucceeded` is idempotent on `request_id`; `is_last_job` is
    computed from the current committed state of the sibling jobs, so the Phase-2
    (`RANKING` → `LABELING`) transition the caller drives MUST be idempotent.
  - Validation lives in the public layer (batch size ≤ 50, `request_id` UUID
    format + in-batch uniqueness, sub-request `parent`/`ranker_job` presence,
    `state_in` filter values), returning `INVALID_ARGUMENT` for client errors.
  - `ListRankerJobs` supports the AIP-159 wildcard `-` parent to list across all
    uploads for a data provider.

  ## ⚠️  Dependency on base schema PR #3989

  The `MarkRequestId STRING(36)` column that `MarkRankerJobSucceeded` idempotency
  relies on is intentionally **not** added here — it belongs to the base schema PR
  (`getina/vid-labeling-spanner-schema`, #3989) and is left as a `TODO` in
  `create-vid-labeling-schema.sql`. The code compiles, but the two
  Spanner-emulator test targets will pass only once that column is added in the
  base branch. (Verified green locally with the column present.)

  ## Testing
  
  - `ktfmt 0.54 --google-style`: clean.
  - `bazel build` of all main + test targets: passing.
  - `RankerJobServiceTest` (v1alpha) and `SpannerRankerJobServiceTest`: passing
    with the `MarkRequestId` column present (see dependency note above).

Issue: #3927

### DIFF
--- a/src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/spanner/SpannerRankerJobService.kt
+++ b/src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/spanner/SpannerRankerJobService.kt
@@ -29,6 +29,7 @@ import kotlinx.coroutines.flow.collectIndexed
 import org.wfanet.measurement.common.IdGenerator
 import org.wfanet.measurement.common.generateNewId
 import org.wfanet.measurement.edpaggregator.deploy.gcloud.spanner.db.RankerJobResult
+import org.wfanet.measurement.edpaggregator.deploy.gcloud.spanner.db.countOtherNonSucceededRankerJobs
 import org.wfanet.measurement.edpaggregator.deploy.gcloud.spanner.db.findRankerJobByRequestId
 import org.wfanet.measurement.edpaggregator.deploy.gcloud.spanner.db.findRankerJobsByRequestIds
 import org.wfanet.measurement.edpaggregator.deploy.gcloud.spanner.db.getRankerJobByResourceId
@@ -36,7 +37,6 @@ import org.wfanet.measurement.edpaggregator.deploy.gcloud.spanner.db.getRawImpre
 import org.wfanet.measurement.edpaggregator.deploy.gcloud.spanner.db.insertRankerJob
 import org.wfanet.measurement.edpaggregator.deploy.gcloud.spanner.db.rankerJobExists
 import org.wfanet.measurement.edpaggregator.deploy.gcloud.spanner.db.readRankerJobs
-import org.wfanet.measurement.edpaggregator.deploy.gcloud.spanner.db.readRankerJobsForModelLine
 import org.wfanet.measurement.edpaggregator.deploy.gcloud.spanner.db.updateRankerJobState
 import org.wfanet.measurement.edpaggregator.service.internal.EtagMismatchException
 import org.wfanet.measurement.edpaggregator.service.internal.InvalidFieldValueException
@@ -446,14 +446,12 @@ class SpannerRankerJobService(
           return@run TransactionResult(
             currentJob,
             isLastJob =
-              computeIsLastJob(
-                txn.readRankerJobsForModelLine(
-                  request.dataProviderResourceId,
-                  result.rawImpressionUploadId,
-                  currentJob.cmmsModelLine,
-                ),
+              txn.countOtherNonSucceededRankerJobs(
+                request.dataProviderResourceId,
+                result.rawImpressionUploadId,
+                currentJob.cmmsModelLine,
                 result.rankerJobId,
-              ),
+              ) == 0L,
             isReplay = true,
           )
         }
@@ -475,26 +473,27 @@ class SpannerRankerJobService(
           etag = newEtag,
         ) {
           set("MarkRequestId").to(request.requestId)
+          // Clear any error message from a prior FAILED attempt; it is only set while FAILED.
+          set("ErrorMessage").to(null as String?)
         }
 
-        // Read all jobs for this (upload, model line). The buffered update above is not visible in
-        // this read, so the current job still shows its previous state; it is treated as SUCCEEDED
-        // via its ID below.
+        // Count sibling jobs in this (upload, model line) that are not yet SUCCEEDED, excluding the
+        // current job. The buffered update above is not visible in this read, so excluding the
+        // current job by ID yields the correct "last job out" semantics.
         val isLastJob =
-          computeIsLastJob(
-            txn.readRankerJobsForModelLine(
-              request.dataProviderResourceId,
-              result.rawImpressionUploadId,
-              currentJob.cmmsModelLine,
-            ),
+          txn.countOtherNonSucceededRankerJobs(
+            request.dataProviderResourceId,
+            result.rawImpressionUploadId,
+            currentJob.cmmsModelLine,
             result.rankerJobId,
-          )
+          ) == 0L
 
         TransactionResult(
           updatedJob =
             currentJob.copy {
               state = State.RANKER_STATE_SUCCEEDED
               etag = newEtag
+              clearErrorMessage()
               clearUpdateTime()
             },
           isLastJob = isLastJob,
@@ -690,19 +689,5 @@ class SpannerRankerJobService(
 
     private val VALID_MARK_PREVIOUS_STATES =
       setOf(State.RANKER_STATE_CREATED, State.RANKER_STATE_FAILED)
-
-    /**
-     * Returns whether marking [currentJobId] as SUCCEEDED makes it the last ranker job to complete
-     * for its (upload, model line). [jobsForModelLine] are the sibling jobs as read before the
-     * buffered transition is visible, so [currentJobId] is treated as already SUCCEEDED.
-     */
-    private fun computeIsLastJob(
-      jobsForModelLine: List<RankerJobResult>,
-      currentJobId: Long,
-    ): Boolean {
-      return jobsForModelLine.all {
-        it.rankerJob.state == State.RANKER_STATE_SUCCEEDED || it.rankerJobId == currentJobId
-      }
-    }
   }
 }
--- a/src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/spanner/SpannerRankerJobService.kt
+++ b/src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/spanner/SpannerRankerJobService.kt
@@ -27,9 +27,7 @@ import kotlin.coroutines.EmptyCoroutineContext
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.collectIndexed
 import org.wfanet.measurement.common.IdGenerator
-import org.wfanet.measurement.common.api.ETags
 import org.wfanet.measurement.common.generateNewId
-import org.wfanet.measurement.common.toInstant
 import org.wfanet.measurement.edpaggregator.deploy.gcloud.spanner.db.RankerJobResult
 import org.wfanet.measurement.edpaggregator.deploy.gcloud.spanner.db.findRankerJobByRequestId
 import org.wfanet.measurement.edpaggregator.deploy.gcloud.spanner.db.findRankerJobsByRequestIds
@@ -128,6 +126,7 @@ class SpannerRankerJobService(
             }
 
           val resourceId = "$RANKER_JOB_RESOURCE_ID_PREFIX-${UUID.randomUUID()}"
+          val newEtag = UUID.randomUUID().toString()
 
           txn.insertRankerJob(
             rawImpressionUploadId = rawImpressionUploadId,
@@ -137,7 +136,7 @@ class SpannerRankerJobService(
             cmmsModelLine = job.cmmsModelLine,
             poolOffsets = job.poolOffsetsList,
             createRequestId = request.requestId,
-            etag = "", // Placeholder; real etag computed from commit timestamp.
+            etag = newEtag,
           )
 
           job.copy {
@@ -145,9 +144,9 @@ class SpannerRankerJobService(
             this.rawImpressionUploadResourceId = rawImpressionUploadResourceId
             rankerJobResourceId = resourceId
             state = State.RANKER_STATE_CREATED
+            etag = newEtag
             clearCreateTime()
             clearUpdateTime()
-            clearEtag()
           }
         }
       } catch (e: SpannerException) {
@@ -169,7 +168,6 @@ class SpannerRankerJobService(
       createdJob.copy {
         createTime = commitTimestamp
         updateTime = commitTimestamp
-        etag = ETags.computeETag(commitTimestamp.toInstant())
       }
     }
   }
@@ -267,6 +265,7 @@ class SpannerRankerJobService(
                 }
 
               val resourceId = "$RANKER_JOB_RESOURCE_ID_PREFIX-${UUID.randomUUID()}"
+              val newEtag = UUID.randomUUID().toString()
 
               txn.insertRankerJob(
                 rawImpressionUploadId = rawImpressionUploadId,
@@ -276,7 +275,7 @@ class SpannerRankerJobService(
                 cmmsModelLine = subRequest.rankerJob.cmmsModelLine,
                 poolOffsets = subRequest.rankerJob.poolOffsetsList,
                 createRequestId = subRequest.requestId,
-                etag = "", // Placeholder; real etag computed from commit timestamp.
+                etag = newEtag,
               )
 
               subRequest.rankerJob.copy {
@@ -284,9 +283,9 @@ class SpannerRankerJobService(
                 this.rawImpressionUploadResourceId = rawImpressionUploadResourceId
                 rankerJobResourceId = resourceId
                 state = State.RANKER_STATE_CREATED
+                etag = newEtag
                 clearCreateTime()
                 clearUpdateTime()
-                clearEtag()
               }
             }
           }
@@ -304,7 +303,6 @@ class SpannerRankerJobService(
       }
 
     val commitTimestamp: Timestamp = transactionRunner.getCommitTimestamp().toProto()
-    val computedEtag = ETags.computeETag(commitTimestamp.toInstant())
     return batchCreateRankerJobsResponse {
       rankerJobs +=
         results.map { result ->
@@ -314,7 +312,6 @@ class SpannerRankerJobService(
             result.copy {
               createTime = commitTimestamp
               updateTime = commitTimestamp
-              etag = computedEtag
             }
           }
         }
@@ -407,6 +404,7 @@ class SpannerRankerJobService(
       request.rankerJobResourceId,
       request.etag,
       request.requestId,
+      requireRequestId = true,
     )
 
     val transactionRunner =
@@ -469,12 +467,14 @@ class SpannerRankerJobService(
           currentJob = currentJob,
         )
 
+        val newEtag = UUID.randomUUID().toString()
+
         txn.updateRankerJobState(
           dataProviderResourceId = request.dataProviderResourceId,
           rawImpressionUploadId = result.rawImpressionUploadId,
           rankerJobId = result.rankerJobId,
           state = State.RANKER_STATE_SUCCEEDED,
-          etag = "", // Placeholder; real etag computed from commit timestamp.
+          etag = newEtag,
         ) {
           set("MarkRequestId").to(request.requestId)
         }
@@ -496,8 +496,8 @@ class SpannerRankerJobService(
           updatedJob =
             currentJob.copy {
               state = State.RANKER_STATE_SUCCEEDED
+              etag = newEtag
               clearUpdateTime()
-              clearEtag()
             },
           isLastJob = isLastJob,
         )
@@ -509,10 +509,7 @@ class SpannerRankerJobService(
           txnResult.updatedJob
         } else {
           val commitTimestamp: Timestamp = transactionRunner.getCommitTimestamp().toProto()
-          txnResult.updatedJob.copy {
-            updateTime = commitTimestamp
-            etag = ETags.computeETag(commitTimestamp.toInstant())
-          }
+          txnResult.updatedJob.copy { updateTime = commitTimestamp }
         }
       isLastJob = txnResult.isLastJob
     }
@@ -525,6 +522,7 @@ class SpannerRankerJobService(
       request.rankerJobResourceId,
       request.etag,
       requestId = "",
+      requireRequestId = false,
     )
 
     val transactionRunner =
@@ -554,12 +552,14 @@ class SpannerRankerJobService(
           currentJob = currentJob,
         )
 
+        val newEtag = UUID.randomUUID().toString()
+
         txn.updateRankerJobState(
           dataProviderResourceId = request.dataProviderResourceId,
           rawImpressionUploadId = result.rawImpressionUploadId,
           rankerJobId = result.rankerJobId,
           state = State.RANKER_STATE_FAILED,
-          etag = "", // Placeholder; real etag computed from commit timestamp.
+          etag = newEtag,
         ) {
           set("ErrorMessage").to(request.errorMessage)
         }
@@ -567,21 +567,19 @@ class SpannerRankerJobService(
         currentJob.copy {
           state = State.RANKER_STATE_FAILED
           errorMessage = request.errorMessage
+          etag = newEtag
           clearUpdateTime()
-          clearEtag()
         }
       }
 
     val commitTimestamp: Timestamp = transactionRunner.getCommitTimestamp().toProto()
-    return updatedJob.copy {
-      updateTime = commitTimestamp
-      etag = ETags.computeETag(commitTimestamp.toInstant())
-    }
+    return updatedJob.copy { updateTime = commitTimestamp }
   }
 
   /**
    * Validates parent identifiers, the job resource ID, the etag, and the request ID common to the
-   * Mark requests. [requestId] may be empty when the RPC does not support idempotency.
+   * Mark requests. When [requireRequestId] is true, [requestId] must be set; it may be empty only
+   * for Mark RPCs that do not support idempotency.
    */
   private fun validateMarkRequest(
     dataProviderResourceId: String,
@@ -589,6 +587,7 @@ class SpannerRankerJobService(
     rankerJobResourceId: String,
     etag: String,
     requestId: String,
+    requireRequestId: Boolean,
   ) {
     if (dataProviderResourceId.isEmpty()) {
       throw RequiredFieldNotSetException("data_provider_resource_id")
@@ -604,6 +603,10 @@ class SpannerRankerJobService(
     }
     if (etag.isEmpty()) {
       throw RequiredFieldNotSetException("etag")
+        .asStatusRuntimeException(Status.Code.INVALID_ARGUMENT)
+    }
+    if (requireRequestId && requestId.isEmpty()) {
+      throw RequiredFieldNotSetException("request_id")
         .asStatusRuntimeException(Status.Code.INVALID_ARGUMENT)
     }
     if (requestId.isNotEmpty()) {
@@ -671,7 +674,7 @@ class SpannerRankerJobService(
   }
 
   companion object {
-    private const val RANKER_JOB_RESOURCE_ID_PREFIX = "rj"
+    private const val RANKER_JOB_RESOURCE_ID_PREFIX = "rankerJob"
     private const val MAX_PAGE_SIZE = 100
     private const val MAX_BATCH_SIZE = 50
     private const val DEFAULT_PAGE_SIZE = 50
--- a/src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/spanner/SpannerRankerJobService.kt
+++ b/src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/spanner/SpannerRankerJobService.kt
@@ -1,0 +1,696 @@
+/*
+ * Copyright 2026 The Cross-Media Measurement Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.wfanet.measurement.edpaggregator.deploy.gcloud.spanner
+
+import com.google.cloud.spanner.ErrorCode
+import com.google.cloud.spanner.Options
+import com.google.cloud.spanner.SpannerException
+import com.google.protobuf.Timestamp
+import io.grpc.Status
+import java.util.UUID
+import kotlin.coroutines.CoroutineContext
+import kotlin.coroutines.EmptyCoroutineContext
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.collectIndexed
+import org.wfanet.measurement.common.IdGenerator
+import org.wfanet.measurement.common.api.ETags
+import org.wfanet.measurement.common.generateNewId
+import org.wfanet.measurement.common.toInstant
+import org.wfanet.measurement.edpaggregator.deploy.gcloud.spanner.db.RankerJobResult
+import org.wfanet.measurement.edpaggregator.deploy.gcloud.spanner.db.findRankerJobByRequestId
+import org.wfanet.measurement.edpaggregator.deploy.gcloud.spanner.db.findRankerJobsByRequestIds
+import org.wfanet.measurement.edpaggregator.deploy.gcloud.spanner.db.getRankerJobByResourceId
+import org.wfanet.measurement.edpaggregator.deploy.gcloud.spanner.db.getRawImpressionUploadIdForRanker
+import org.wfanet.measurement.edpaggregator.deploy.gcloud.spanner.db.insertRankerJob
+import org.wfanet.measurement.edpaggregator.deploy.gcloud.spanner.db.rankerJobExists
+import org.wfanet.measurement.edpaggregator.deploy.gcloud.spanner.db.readRankerJobs
+import org.wfanet.measurement.edpaggregator.deploy.gcloud.spanner.db.readRankerJobsForModelLine
+import org.wfanet.measurement.edpaggregator.deploy.gcloud.spanner.db.updateRankerJobState
+import org.wfanet.measurement.edpaggregator.service.internal.EtagMismatchException
+import org.wfanet.measurement.edpaggregator.service.internal.InvalidFieldValueException
+import org.wfanet.measurement.edpaggregator.service.internal.RankerJobAlreadyExistsException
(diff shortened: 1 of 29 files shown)


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4052)

src/main/resources/edpaggregator/spanner/create-vid-labeling-schema.sql
```diff
@@ -178,6 +178,11 @@ CREATE TABLE RankerJob (
   Etag STRING(36) NOT NULL,
   ErrorMessage STRING(MAX),
   CreateRequestId STRING(36),
+  -- TODO(getina/vid-labeling-spanner-schema, PR #3989): Add `MarkRequestId STRING(36)` column.
```
stevenwarejones: The schema file is in this PR's diff (5 lines added vs base) — you added the TODO comment but the column itself is unresolved. And you've ALREADY added `MarkRequestId STRING(36)` to PoolAssignmentJob in this same file at line 144 — RankerJob just needs the same one-line addition. Pick one of two:

**Option A — Add the column here:** drop the TODO and add `MarkRequestId STRING(36)` to the `RankerJob` table. One-line schema change matching what you already did for PoolAssignmentJob, makes this PR self-contained, unblocks the Spanner-emulator tests immediately, and #3989 inherits the column when this lands.

```sql
CreateRequestId STRING(36),
MarkRequestId STRING(36),  -- AIP-155 idempotency key for MarkRankerJobSucceeded.
CreateTime TIMESTAMP NOT NULL OPTIONS (allow_commit_timestamp = true),
```

**Option B — Don't touch the schema in this PR at all:** revert your 5-line addition (drop the TODO too), and split the MarkRequestId code paths out of this PR into a follow-up stacked on #3989-with-column. This PR ships Create/Get/List/MarkFailed only; MarkSucceeded with idempotency lands in the follow-up.

What's not okay is the current state: schema is touched, column is needed, column isn't added. The code at `db/RankerJob.kt:391` and `SpannerRankerJobService.kt:479` reads/writes a column that doesn't exist, so every Spanner-emulator test fails at the first SELECT.

stevenwarejones (commented):
Schema TODO — add the column here or drop the schema edit entirely.

src/main/kotlin/org/wfanet/measurement/edpaggregator/service/v1alpha/RankerJobService.kt
```diff
@@ -0,0 +1,475 @@
...
+      RankerJobKey.fromName(request.name)
+        ?: throw InvalidFieldValueException("name")
+          .asStatusRuntimeException(Status.Code.INVALID_ARGUMENT)
+
+    if (request.etag.isEmpty()) {
+      throw RequiredFieldNotSetException("etag")
+        .asStatusRuntimeException(Status.Code.INVALID_ARGUMENT)
+    }
+
+    if (request.requestId.isNotEmpty()) {
```
stevenwarejones: `request_id` is optional on `MarkRankerJobSucceeded` (validated for UUID format only if non-empty at v1alpha:330-337, internal:609-616). The idempotency check at `SpannerRankerJobService.kt:445-449` only fires when both sides have a request_id — without one, a Pub/Sub redelivery of a successful Mark falls through to `checkMarkPrecondition`, which rejects with `RANKER_JOB_STATE_INVALID` because SUCCEEDED isn't in `VALID_MARK_PREVIOUS_STATES`. The redelivery is poison-pilled.

The `VidRankBuilder` TEE caller (#4009) does pass a deterministic request_id, so the happy path works today — but the service contract is wrong. Any future caller that forgets request_id is one redelivery away from an infinite poison-pill loop.

Make `request_id` REQUIRED:

1. **Proto**: add `[(google.api.field_behavior) = REQUIRED]` to `MarkRankerJobSucceededRequest.request_id`.
2. **Public service** (`RankerJobService.kt:330`): reject empty before the UUID-format check:
   ```kotlin
   if (request.requestId.isEmpty()) {
     throw RequiredFieldNotSetException("request_id")
   }
   ```
3. **Internal service** (`SpannerRankerJobService.kt:validateMarkRequest`): same — pull the empty check into the validator instead of leaving it conditional.
4. **Test**: add a test that empty request_id → INVALID_ARGUMENT. The existing `Marked SUCCEEDED returns isLastJob on replay with same request_id` test already exercises the happy path.

Same pattern just landed on `RawImpressionUploadFile.batchCreate` in #4014 — required there, should be required here. State-transition operations the design depends on being idempotent need the idempotency key to actually be required.

stevenwarejones (commented):
request_id should be REQUIRED on MarkRankerJobSucceeded — see inline.

src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/spanner/SpannerRankerJobService.kt
```diff
@@ -0,0 +1,696 @@
...
+    if (request.rankerJob.cmmsModelLine.isEmpty()) {
+      throw RequiredFieldNotSetException("ranker_job.cmms_model_line")
+    }
+    if (request.rankerJob.poolOffsetsList.isEmpty()) {
+      throw RequiredFieldNotSetException("ranker_job.pool_offsets")
+    }
+  }
+
+  companion object {
+    private const val RANKER_JOB_RESOURCE_ID_PREFIX = "rj"
```
stevenwarejones: nit: `"rj"` is opaque. Mixed convention in the codebase, but your own recently-landed PRs (`SpannerRawImpressionUploadService` → `"rawImpressionUpload"`, `SpannerRawImpressionUploadFileService` → `"rawImpressionUploadFile"`) use the full resource type name. Match that here:

```kotlin
private const val RANKER_JOB_RESOURCE_ID_PREFIX = "rankerJob"
```

A resource ID like `rankerJob-3c8f...` reads clearly in logs and Cloud Console; `rj-3c8f...` requires the reader to know the abbreviation. The older `"req"` / `"file"` / `"batch"` prefixes in `RequisitionMetadata` / `RawImpressionMetadataBatchFile` / `RawImpressionMetadataBatch` predate the convention you established on the recent PRs — no reason to revert to abbreviation here.

stevenwarejones (commented):
nit: rj prefix is opaque — use rankerJob.

stevenwarejones (commented):
> @stevenwarejones reviewed 25 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4052)* status: all files reviewed, 3 unresolved discussions (waiting on getina and Marco-Premier).

src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/spanner/SpannerRankerJobService.kt
```diff
@@ -0,0 +1,696 @@
...
+          txn.insertRankerJob(
+            rawImpressionUploadId = rawImpressionUploadId,
+            rankerJobId = rankerJobId,
+            rankerJobResourceId = resourceId,
+            dataProviderResourceId = dataProviderResourceId,
+            cmmsModelLine = job.cmmsModelLine,
+            poolOffsets = job.poolOffsetsList,
+            createRequestId = request.requestId,
+            etag = "", // Placeholder; real etag computed from commit timestamp.
+          )
```
stevenwarejones, line 130: Cross-link to my comment on #3989 about inconsistent `Etag` handling across the four job tables. RankerJob is one of the inconsistent cases — `set("Etag").to("")` placeholder here (line 351 in db file) plus `etag = ""` argument here in the service.

Switch to the explicit-write pattern (the one `RawImpressionUploadModelLine` and `PoolAssignmentJob` already use): compute the etag inside the transaction (e.g. `ETags.computeETag(...)` of a stable input) and write it as the actual value. Match what the schema's `NOT NULL` claims to enforce.

Rationale (full version on #3989 thread):
- Schema honesty: `NOT NULL` is meaningless if the column is always `""`.
- Forward compatibility: the column can later carry non-UpdateTime semantics (content hash, schema version).
- Silent regression: any future bumping of `UpdateTime` for non-semantic reasons silently invalidates client etag caches. Explicit writes don't.

Same fix needed on the Mark path: `updateRankerJobState` (db/RankerJob.kt:358) also takes `etag` as a parameter and the service passes `""`. Compute the new etag and write it.

stevenwarejones (commented):
Etag should be written explicitly, not placeholder — see #3989 thread.

stevenwarejones:
Make `request_id` REQUIRED on every state-mutating RPC in this service: `CreateRankerJob`, `BatchCreateRankerJobs` (per-element), and every `Mark*` RPC. Same for `etag` on Marks (currently REQUIRED, keep it that way).

Rationale:
- etag is the optimistic-concurrency contract — clients must declare what state they think they're acting on. Optional etag is a blind write, which has no place in a pipeline where every transition is gated by precondition.
- request_id is the AIP-155 idempotency contract — clients must provide an at-most-once token so retries are safe. Optional request_id means a network-retried call silently re-runs or silently fails depending on server state.

Proto changes (`ranker_job_service.proto`):
- `CreateRankerJobRequest.request_id`: OPTIONAL → REQUIRED + UUID4
- `BatchCreateRankerJobsRequest`: each element's `request_id` REQUIRED
- Every `Mark*Request`: add `string request_id = N [REQUIRED, UUID4]` (currently absent on Marks)

Public + internal layers: reject empty `request_id` and `etag` with INVALID_ARGUMENT before forwarding.

This is v1alpha with no production clients, so the breaking change is free. Cross-posting on #4018, #4044, #4051 — keeping all four services consistent. Depends on Mark RPCs adopting MarkRequestId-based idempotency per the #3989 schema feedback.

src/main/kotlin/org/wfanet/measurement/edpaggregator/service/v1alpha/RankerJobService.kt
Marco-Premier: now the request id is required on all grpc as per comment in #4044

src/main/resources/edpaggregator/spanner/create-vid-labeling-schema.sql
Marco-Premier: option A

Marco-Premier (commented):
> @Marco-Premier made 2 comments and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4052)* status: 19 of 27 files reviewed, 3 unresolved discussions (waiting on getina and stevenwarejones).

src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/spanner/SpannerRankerJobService.kt
Marco-Premier: Done.

Marco-Premier (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4052)* status: 19 of 27 files reviewed, 3 unresolved discussions (waiting on getina and stevenwarejones).

src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/spanner/db/RankerJob.kt
```diff
@@ -0,0 +1,423 @@
...
+        val result = RankerJobEntity.buildResult(row)
+        if (!row.isNull("CreateRequestId")) {
+          put(row.getString("CreateRequestId"), result)
+        }
+      }
+  }
+}
+
+/** Reads [RankerJob] entries for the same (upload, model line) group. */
+suspend fun AsyncDatabaseClient.ReadContext.readRankerJobsForModelLine(
```
stevenwarejones: Replace this O(N) per-row read with a single COUNT. At scale (hundreds/thousands of RankerJobs per (upload, model_line)), `readRankerJobsForModelLine` reads every row's (id, state) just so `computeIsLastJob` can check 'all SUCCEEDED except this one' — a single COUNT(*) achieves the same result in O(1).

```
suspend fun AsyncDatabaseClient.ReadContext.countOtherNonSucceededRankerJobs(
  dataProviderResourceId: String,
  rawImpressionUploadId: Long,
  cmmsModelLine: String,
  excludeRankerJobId: Long,
): Long {
  val sql =
    """
    SELECT COUNT(*) AS cnt
    FROM RankerJob
    WHERE DataProviderResourceId = @dataProviderResourceId
      AND RawImpressionUploadId = @rawImpressionUploadId
      AND CmmsModelLine = @cmmsModelLine
      AND RankerJobId != @excludeRankerJobId
      AND CAST(State AS INT64) != @succeededState
    """
  ...
}

// Spanner.kt — drop computeIsLastJob, just:
val isLastJob =
  txn.countOtherNonSucceededRankerJobs(
    request.dataProviderResourceId, result.rawImpressionUploadId,
    currentJob.cmmsModelLine, result.rankerJobId,
  ) == 0L
```

Same fix landed on #4044 M6 (-OvV7_HP1PFz88jFeWGE). The buffered state update for the current row isn't visible in the read, so excluding by ID is the correct semantics. Apply to both the replay path (Spanner.kt:449) and the forward path (Spanner.kt:484).

src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/spanner/SpannerRankerJobService.kt
```diff
@@ -0,0 +1,708 @@
...
+
+        txn.updateRankerJobState(
+          dataProviderResourceId = request.dataProviderResourceId,
+          rawImpressionUploadId = result.rawImpressionUploadId,
+          rankerJobId = result.rankerJobId,
+          state = State.RANKER_STATE_SUCCEEDED,
+          etag = newEtag,
+        ) {
+          set("MarkRequestId").to(request.requestId)
+        }
```
stevenwarejones, line 470: Clear `ErrorMessage` on the FAILED → SUCCEEDED transition. `VALID_MARK_PREVIOUS_STATES` (Spanner.kt:691) allows a FAILED row to retry into SUCCEEDED, but neither the buffered update nor the response copy clears `ErrorMessage`. Proto says `error_message` is "set when state is FAILED" — a succeeded row carrying stale failure text contradicts that.

In the buffered update (Spanner.kt:470-478):

```
txn.updateRankerJobState(
  ...,
  state = State.RANKER_STATE_SUCCEEDED,
  etag = newEtag,
) {
  set("MarkRequestId").to(request.requestId)
  set("ErrorMessage").to(null as String?)
}
```

And in the response copy (Spanner.kt:494-499):

```
updatedJob =
  currentJob.copy {
    state = State.RANKER_STATE_SUCCEEDED
    etag = newEtag
    clearErrorMessage()
    clearUpdateTime()
  },
```

Same fix landed on #4044 M7 (-OvV8E-V-6P86CPzIOvF). Add a test that goes CREATED → markFailed → markSucceeded, asserting the returned `error_message` is empty.

src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/spanner/SpannerRankerJobService.kt
stevenwarejones: Verified — explicit-write etag throughout. Insert sets `Etag` from the constructor parameter (Db.kt:336, 349), updates set it via parameter (Db.kt:361, 369), buildResult reads it from the column (Db.kt:417: `etag = struct.getString(\"Etag\")`). No more derive-from-UpdateTime. Matches the explicit-write pattern landed across #3989 / #4018 / #4044 / #4051.

src/main/kotlin/org/wfanet/measurement/edpaggregator/service/v1alpha/RankerJobService.kt
stevenwarejones: Verified — `request_id` is now REQUIRED on every state-mutating RPC. Public proto: CreateRankerJobRequest.request_id (REQUIRED + UUID4), MarkRankerJobSucceededRequest.request_id (REQUIRED + UUID4), MarkRankerJobFailedRequest.request_id (REQUIRED + UUID4). Public layer enforces non-empty + UUID format at RankerJobService.kt:96-103 (Create) and 329-345 (Mark*). Internal Spanner replay check at SpannerRankerJobService.kt:442-445 simplified to `result.markRequestId == request.requestId` (no `isNotEmpty` guard needed).

src/main/resources/edpaggregator/spanner/create-vid-labeling-schema.sql
stevenwarejones: Verified — `MarkRequestId STRING(36)` added to RankerJob at create-vid-labeling-schema.sql:181, matching PoolAssignmentJob's column at line 144.

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4052)* status: 19 of 27 files reviewed, 5 unresolved discussions (waiting on getina, Marco-Premier, and stevenwarejones).

stevenwarejones (commented):
> @stevenwarejones reviewed 8 files and all commit messages, and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4052)* status: all files reviewed, 3 unresolved discussions (waiting on getina and Marco-Premier).

stevenwarejones (commented):
These all look good now. Thx. Just a few more nits.

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4052#-:-OvoY7g538qa-cxLdHCY:btbnek3)* status: all files reviewed, 3 unresolved discussions (waiting on getina and Marco-Premier).

src/main/proto/wfa/measurement/edpaggregator/v1alpha/ranker_job_service.proto
```diff
@@ -206,7 +206,7 @@ message MarkRankerJobSucceededRequest {
   //
   // This field must be a UUID4 (RFC 4122) string.
   string request_id = 3 [
```
stevenwarejones, line 226: Document the deterministic-request_id contract for clients. The KDoc currently says 'a retry with the same request_id returns the original response' — accurate, but doesn't tell clients how to construct request_id so retries actually share one. A client that follows the proto literally and generates fresh UUIDs per call will hit FAILED_PRECONDITION on natural retries (the second call's request_id won't match `result.markRequestId`, fall through to validPreviousStates, fail because the row is already SUCCEEDED).

Recommend:

```
// A request ID provided by the client to ensure idempotency on retry. To survive
// retries (network blips, redeliveries, process restarts), derive this deterministically
// from the resource being marked — e.g. UUIDv5/MD5-hash of (name, operation) — so
// every attempt for the same logical mark uses the same request_id. Sending fresh
// UUIDs per attempt defeats AIP-155 and will surface as FAILED_PRECONDITION on retry
// against an already-transitioned row.
```

Apply the same KDoc to MarkRankerJobFailedRequest.request_id (line 247). VidRankBuilder #4009 already follows this pattern (`deterministicUuid("$rankerJob|succeeded")`); this just makes the contract explicit for future clients.

src/main/proto/wfa/measurement/edpaggregator/v1alpha/ranker_job_service.proto
```diff
@@ -238,4 +238,14 @@ message MarkRankerJobFailedRequest {
 
   // Human-readable detail about the failure.
   string error_message = 3 [(google.api.field_behavior) = OPTIONAL];
+
+  // A request ID provided by the client to ensure idempotency
+  // on retry. A retry with the same `request_id` returns the
+  // original response without re-transitioning the resource.
+  //
+  // This field must be a UUID4 (RFC 4122) string.
+  string request_id = 4 [
```
stevenwarejones, line 273: Apply the same deterministic-request_id KDoc here as on MarkRankerJobSucceededRequest.request_id (line 208). Same hazard: a client that doesn't use a deterministic request_id will get FAILED_PRECONDITION on retried markFailed calls against an already-FAILED row instead of a clean idempotent return.

src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/spanner/SpannerRankerJobService.kt
Marco-Premier: Done.

src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/spanner/db/RankerJob.kt
Marco-Premier: Done.

src/main/kotlin/org/wfanet/measurement/edpaggregator/service/v1alpha/RankerJobService.kt
Marco-Premier: Done.

src/main/proto/wfa/measurement/edpaggregator/v1alpha/ranker_job_service.proto
Marco-Premier: Done.

src/main/proto/wfa/measurement/edpaggregator/v1alpha/ranker_job_service.proto
Marco-Premier: Done.

Marco-Premier (commented):
> @Marco-Premier made 5 comments.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4052)* status: 23 of 27 files reviewed, 5 unresolved discussions (waiting on getina, stevenwarejones, and stevenwarejones+AGNT).

src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/spanner/SpannerRankerJobService.kt
stevenwarejones: Verified in commit 12f9cb89e. Spanner.kt:476-478 buffers `set("ErrorMessage").to(null as String?)` on the SUCCEEDED transition, and the response copy at line 491-499 calls `clearErrorMessage()`. A FAILED → SUCCEEDED retry no longer carries stale failure text in either the DB row or the response.

src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/spanner/db/RankerJob.kt
stevenwarejones: Verified in commit 12f9cb89e. `countOtherNonSucceededRankerJobs` added at Db.kt:210 and wired at Spanner.kt:449 (replay path) and Spanner.kt:484 (forward path). The full-row scan plus in-memory `computeIsLastJob` is gone. Single COUNT per call now.

src/main/kotlin/org/wfanet/measurement/edpaggregator/service/v1alpha/RankerJobService.kt
stevenwarejones: Verified — `request_id` REQUIRED on every state-mutating RPC throughout. No further follow-up needed on this thread.

src/main/proto/wfa/measurement/edpaggregator/v1alpha/ranker_job_service.proto
stevenwarejones: Verified in commit e8464001f. `MarkRankerJobSucceededRequest.request_id` (proto:212-228) now documents the deterministic-derivation contract and the FAILED_PRECONDITION hazard for clients that send fresh UUIDs. The rpc-level KDoc at lines 48-54 also notes the requirement.

src/main/proto/wfa/measurement/edpaggregator/v1alpha/ranker_job_service.proto
stevenwarejones: Verified in commit e8464001f. Same deterministic-request_id contract documented on `MarkRankerJobFailedRequest.request_id` and the rpc-level KDoc at lines 60-65.

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4052)* status: 23 of 27 files reviewed, 5 unresolved discussions (waiting on getina and stevenwarejones).

stevenwarejones (approved):
> @stevenwarejones reviewed 4 files and all commit messages, and resolved 5 discussions.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4052)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on getina).

stevenwarejones (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4052)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on getina and SanjayVas).

src/main/resources/edpaggregator/spanner/create-vid-labeling-schema.sql
```diff
@@ -0,0 +1,279 @@
...
+  RawImpressionUploadModelLineId INT64 NOT NULL,
+  RawImpressionUploadModelLineResourceId STRING(63) NOT NULL,
+  CmmsModelLine STRING(MAX) NOT NULL,
+  CreateRequestId STRING(36),
+  State `wfa.measurement.internal.edpaggregator.RawImpressionUploadModelLineState` NOT NULL,
+  ErrorMessage STRING(MAX),
+  PoolOffsets ARRAY<INT64>,
+  MaxEventDate DATE,
+  EncryptedMergedDek BYTES(MAX),
+  Etag STRING(36) NOT NULL,
```
stevenwarejones: Rebase needed — this branch is on a stale snapshot of `getina/vid-labeling-spanner-schema`. The base branch has added `MarkRequestId` columns + matching unique indexes on RawImpressionUploadModelLine and VidLabelingJob, the four `*ByMarkRequestId` UNIQUE indexes (PoolAssignmentJob, RankerJob, VidLabelingJob, RawImpressionUploadModelLine), and the `RankIndexBlobByNaturalKey` unique index — all missing from this PR's snapshot. A `git fetch origin getina/vid-labeling-spanner-schema && git rebase origin/getina/vid-labeling-spanner-schema` brings them back.

This PR's own service code compiles against the schema as-is (RankerJob.MarkRequestId is present at line 181), but the missing items would regress Tina's consolidation work when this lands.

stevenwarejones (commented):
Rebase needed.

src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/spanner/SpannerRankerJobService.kt
stevenwarejones: Verified in commit e8464001f. `RANKER_JOB_RESOURCE_ID_PREFIX = "rankerJob"` at SpannerRankerJobService.kt:685 — resource IDs now read as `rankerJob-{uuid}`, matching the established convention from SpannerRawImpressionUploadService. Closing.

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4052)* status: all files reviewed, 1 unresolved discussion (waiting on getina, Marco-Premier, and SanjayVas).

stevenwarejones (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4052)* status: all files reviewed, 1 unresolved discussion (waiting on getina, Marco-Premier, and SanjayVas).

src/main/resources/edpaggregator/spanner/create-vid-labeling-schema.sql
SanjayVas: +1. This PR diff is misleading to reviewers as it makes it look like this file is added by this PR rather than the base.

SanjayVas (changes requested):
> @SanjayVas reviewed 3 files and all commit messages, and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4052)* status: all files reviewed, 2 unresolved discussions (waiting on getina, Marco-Premier, stevenwarejones, and stevenwarejones+AGNT).

___
*[`src/main/proto/wfa/measurement/edpaggregator/v1alpha/ranker_job_service.proto` line 51 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4052#-Ovw6KYI3VZJ5a1SnWTn:-Ovw6KYI3VZJ5a1SnWTo:byy8vdu) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/e8464001fc4f30d6c9b46ece6b5e018337ab9b7e/src/main/proto/wfa/measurement/edpaggregator/v1alpha/ranker_job_service.proto#L51)):*
> ```proto
>   // version (AIP-154). A retry with the same `request_id`
>   // returns the original response without modifying state
>   // (AIP-155); `request_id` must be derived deterministically so
> ```

nit: all this extra documentation about request_id and retries seems pretty overkill to me, as it's just the natural consequence of the existing documentation for the `request_id` field. Meaning the fact that you need to ensure the same request ID for equivalent requests is obvious.

Documenting preconditions is still useful, e.g. what states the resource is expected to be in.

stevenwarejones (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4052)* status: all files reviewed, 2 unresolved discussions (waiting on getina, Marco-Premier, and stevenwarejones+AGNT).

src/main/resources/edpaggregator/spanner/create-vid-labeling-schema.sql
Marco-Premier: Done.

Marco-Premier (commented):
> @Marco-Premier made 1 comment and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4052)* status: 19 of 29 files reviewed, 1 unresolved discussion (waiting on getina, SanjayVas, and stevenwarejones).

stevenwarejones (approved):
> @stevenwarejones reviewed 23 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4052)* status: all files reviewed, 1 unresolved discussion (waiting on getina and SanjayVas).

SanjayVas (changes requested):
> @SanjayVas reviewed 27 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4052)* status: all files reviewed, 2 unresolved discussions (waiting on getina and Marco-Premier).

___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/spanner/db/RankerJob.kt` line 68 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4052#-Ow-lApN8a9lQ2G_uYiQ:-Ow-lApN8a9lQ2G_uYiR:be10pck) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/fd0f5b1a0e96abc19c9b41dbd0432eb32a7b0e16/src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/spanner/db/RankerJob.kt#L68)):*
> ```Kotlin
>  */
> suspend fun AsyncDatabaseClient.ReadContext.getRawImpressionUploadIdForRanker(
> ```

https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4051#-OvvxlLg5Imb0g9SSfIC:-OvvxlLhAn0qlpJJj2bx:b-7s4dj3

Call the function in RawImpressionUpload.kt defined in #4058

Marco-Premier (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4052)* status: 20 of 30 files reviewed, 2 unresolved discussions (waiting on getina, SanjayVas, and stevenwarejones).

___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/spanner/db/RankerJob.kt` line 68 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4052#-Ow-lApN8a9lQ2G_uYiQ:-Ow2hnJB4Mpq5bnKyEgX:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/fd0f5b1a0e96abc19c9b41dbd0432eb32a7b0e16/src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/spanner/db/RankerJob.kt#L68)):*

https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4051#-OvvxlLg5Imb0g9SSfIC:-OvvxlLhAn0qlpJJj2bx:b-7s4dj3

Call the function in RawImpressionUpload.kt defined in #4058
</blockquote></details>

Done.

SanjayVas (approved):
> @SanjayVas reviewed 24 files and all commit messages, and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4052)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on getina and stevenwarejones).

Marco-Premier (commented):
> @Marco-Premier reviewed 26 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4052)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on getina and stevenwarejones).

PR #4053, opened 2026-06-22
feat(post-processor): snap k_plus_reach to satisfy per-window rounding identities (partial #4049)

stevenwarejones:
Partial fix for #4049. Within each BasicMetricSet, snap k_plus_reach so that k_plus_reach[0] == reach and sum(k_plus_reach) <= impressions. Independent round() calls on the solver's float outputs could drift these two integer identities apart, breaking validation rules consumers rely on.

The reach snap is unconditional: k_plus_reach[0] is overwritten with the canonical reach value. The impressions residue (sum(k_plus_reach) - impressions, after the reach snap) is absorbed bucket-by-bucket from the highest-frequency end, cascading into the next-highest bucket if the top one can't hold it all. The cascade stops before index 0 so the reach identity is never re-broken; if the residue still can't be absorbed (which only happens when reach > impressions in the input -- upstream data inconsistency, not a rounding issue), a warning is logged and the impressions identity is left violated by the leftover residue.

Addresses Rule 4 from the issue, and the only sensible reading of Rule 3 (per-BMS reach == k_plus_reach[0]). Does not touch any cross-BMS identity: Rules 1, 2, and 5 compare values across separate BasicMetricSets in the same report and require separate fixes. PR #4054 addresses Rules 1 and 2; Rule 5 is tracked in #4059.

Issue: #4049

Issue #4049, opened 2026-06-19
Adjust BasicReport output to match Origin's validation criteria w.r.t. K+ reach


khisba:
### Bug description

The BasicReport outputs are not meeting the expected noise corrector validation criteria.

E.g. (Synthetic data from Staging)
This the result of our validations:
“[ami] kPlusReach (slice: total | groups: dataProviders/Jpl6H0aRX2w, dataProviders/WAXeoAvvMNY) – sum of all kPlus buckets (13432561) exceeds impressions (13432556);
[ami] cumulativeWeeklyReach (slice: total | groups: dataProviders/WAXeoAvvMNY) – final week value (2642854) does not equal kPlusReach 1+ (2642848)"

### Steps To Reproduce

Environment: Origin and Aquila Staging
Run a report using Halo API.
Validate the results meet the expected Noise Corrector rules:

1. Cumulative measurements are non-decreasing.
2. Whole campaign reach matches last cumulative reach.
3. Whole campaign reach equals to the sum of k-reaches.
4. Impression is greater than or equal to weighted sum of k-reaches.
5. Check for the consistency between ordered metrics. (ami >= [custom, mrc])

Some of these criteria are not met. We see small deviations.

Rules 4 and 2 are breached here:

“[ami] kPlusReach (slice: total | groups: dataProviders/Jpl6H0aRX2w, dataProviders/WAXeoAvvMNY) – sum of all kPlus buckets (13432561) exceeds impressions (13432556);
[ami] cumulativeWeeklyReach (slice: total | groups: dataProviders/WAXeoAvvMNY) – final week value (2642854) does not equal kPlusReach 1+ (2642848)"

### Component(s) affected

Reporting Server  report-result-post-processor-cronjob

### Version

0.5.42

### Environment

Origin and Aquila Staging

kungfucraig:
What module is doing the validation?

cmmbrandon:
@kungfucraig We have a script that validates this to ensure compliance as part of CMM API. But we've also manually validated this behaviour directly from Halo API outputs.

khisba:
I think this is a rounding issue when the Halo API sums frequency bin values to compute k+ reach.

There are a couple of types of validation checks we see fail frequently:

**1. The sum of the k+ reach <= impression count.**

[...]

PR #4054, opened 2026-06-22
feat: snap whole-campaign reach to last weekly cumulative reach

stevenwarejones:
Partial fix for #4049. The QP solver constrains whole_campaign.reach to equal the last weekly cumulative window's reach (see report.py:_add_cumulative_whole_campaign_relations_to_spec), but each ReportingSetResult is rounded independently after correction. A residual up to the solver tolerance (0.1) can amplify into a 1-unit integer mismatch across the two RSRs, breaking an identity consumers rely on.

For each pair of ReportingSetResults that share dimension modulo metric_frequency_spec selector (i.e. same reporting_set, venn_region, IQF, grouping, event_filters; one with selector=total, one with selector=weekly), snap whole_campaign.reach and the last weekly cumulative window's reach to min(both). k_plus_reach[0] is also snapped on each side, and the rest of k_plus_reach is forward-clamped so the per-BMS reach == k_plus_reach[0] identity and the non-increasing invariant stay satisfied. Taking the min (rather than always pulling whole_campaign to last_weekly) prevents re-breaking the sum(k_plus_reach) <= impressions identity fixed by PR #4053: snapping a side upward would also bump its k_plus_reach[0] upward, which could push the sum above impressions. Snapping downward never has that effect.

The dim-key matching is keyed on the server-computed grouping_dimension_fingerprint and filter_fingerprint already on each ReportingSetResult (the same fingerprints the ReportingSetResultsByDimensions UNIQUE INDEX uses). This avoids relying on Python protobuf SerializeToString(), which is not guaranteed deterministic across map iteration orders.

After the cross-window snap, a backward sweep over the weekly series clamps any earlier window whose cumulative reach now exceeds its successor's. The sweep runs unconditionally rather than only after a snap-down, so a pre-existing Rule 1 (cumulative non-decreasing) violation in the weekly series is fixed even when the cross-window reaches happened to agree. The clamp reuses the snap helper so derived fields and k_plus_reach monotonicity stay consistent on every touched window.

Derived fields (percent_reach, percent_k_plus_reach, average_frequency, grps) are recomputed via a single _recompute_derived_fields helper shared with compute_basic_metric_set, so the construction and mutation paths never drift.

When two ReportingSetResults share the dim key plus the same selector kind, the reconciler raises ValueError rather than logging and skipping. PR #4057 closes the only known input path that produces this collision, so reaching this branch implies upstream data is already corrupted (one RSR shipping values derived from another cadence's solver run, per Issue #4056) and failing loud is the correct response over silently shipping inconsistent data. The same fail-loud treatment applies to a total-selector RSR with more than one reporting window (a data-model violation in _group_results_by_window) and to an unknown impression_qualification_filter oneof variant in the dim-key helper (defense in depth against future proto evolution).

Addresses Rules 1, 2, 3, and 4 from the issue. Rule 5 (cross-IQF consistency, ami >= [custom, mrc]) compares values across separate BasicMetricSets with different impression qualification filters and would require its own snap pass; tracked in #4059. Per-BMS identities (Rules 3 and 4) are handled by PR #4053.

Issue: #4049


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4054)

ple13 (approved):
> @ple13 reviewed 2 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4054)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on kungfucraig).

ple13 (approved):
> @ple13 reviewed 2 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4054)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on stevenwarejones).

Issue #4056, opened 2026-06-23
Post-processor silently collapses multiple metric_frequency_spec cadences per dimension into one


stevenwarejones:
### Bug description

The report post-processor groups RSRs for solver input by `(impression_filter, edp_combination)` only -- it does not include the `metric_frequency_spec` selector value (e.g. `DayOfWeek`). If a single `ReportResult` contains two `ReportingSetResult` rows that share the same `(reporting_unit, dimension_spec, IQF)` but differ only in their `metric_frequency_spec` value (for example, two `weekly` cadences with different `DayOfWeek`s), the second one to be enumerated silently overwrites the first at `_weekly_cumulative_reaches[impression_filter][edp_combination]` in `post_process_report_summary_v2._process_union_results`. The solver then runs on only one cadence's worth of data; the other cadence's RSR receives processed values derived from the surviving cadence's solver output, regardless of its own unprocessed data.

The public BasicReport API permits this input shape today: `BasicReport.result_group_specs` is `repeated ResultGroupSpec`, each `ResultGroupSpec` carries one `MetricFrequencySpec metric_frequency`, and `CreateBasicReportRequestValidation.validateResultGroupSpecs` does not de-dup across specs. The Spanner `UNIQUE INDEX` on `ReportingSetResults` includes the full `MetricFrequencySpecFingerprint`, so both rows persist; the collapse happens in the post-processor only.

As an immediate guard, PR #4057 rejects two `ResultGroupSpec`s that share `(reporting_unit, dimension_spec, metric_frequency selector kind)` but differ in their full `metric_frequency` value. The reconciler added in PR #4054 also logs-and-skips if it sees the same dim-key collision at its layer. Both are defensive -- neither fixes the underlying inability of the post-processor to handle multiple cadences for one slice.

### Proper fix (deferred)

1. **Solver input grouping.** Extend `post_process_report_summary_v2._process_union_results` and the `ReportSummaryV2` shape it consumes so that the keying includes the full `metric_frequency_spec`, not just the selector kind. Same for `_weekly_non_cumulative_measurements` and `_whole_campaign_measurements`.

[...]

PR #4057, opened 2026-06-23
fix: reject result_group_specs that collide on dim + selector kind

stevenwarejones:
Reject two result_group_specs that share (reporting_unit, dimension_spec, metric_frequency selector kind) but differ in the full metric_frequency value (e.g. weekly: MONDAY vs weekly: SUNDAY for the same slice). Without this rejection, the two specs persist as distinct ReportingSetResult rows, but the post-processor groups solver input by (impression_filter, edp_combination) only -- ignoring metric_frequency entirely -- so the second cadence's measurements silently overwrite the first's, and both RSRs receive processed values derived from a single cadence's solver run.

Specs sharing the full metric_frequency value remain allowed -- the job-level dedup merges them into a single RSR before the post-processor sees them.

The collision key canonicalizes reportingUnit and dimensionSpec before comparing, using two new helpers added to Normalization.kt (normalizeReportingUnit, normalizeDimensionSpec). The downstream pipeline normalizes filter order (via Normalization.normalizeEventFilters) before generating RSRs, so two specs with permuted dimensionSpec.filters and different metric_frequency values would otherwise pass the validator yet still collapse to the same dim in the post-processor. Two regression tests pin this: one for the basic MONDAY-vs-TUESDAY case, one for the filter-permuted case (`[age_group, gender]` MONDAY vs `[gender, age_group]` TUESDAY). Both fail without the fix, pass with it.

Issue: #4056


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4057)

SanjayVas (changes requested):
> @SanjayVas reviewed 3 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4057)* status: 3 of 7 files reviewed, 1 unresolved discussion (waiting on kungfucraig and stevenwarejones).

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/CreateBasicReportRequestValidation.kt` line 260 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4057#-Ovqrxb-9IAbF9UACqPE:-Ovqrxb-9IAbF9UACqPF:bcjsygh) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ea0d09e3de1177c96e9e3905c40d368d04709810/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/CreateBasicReportRequestValidation.kt#L260)):*
> ```Kotlin
>    * same dimension in the post-processor.
>    *
>    * Proto messages don't define stable hashCode/equals across runtimes, so key on the serialized
> ```

Protobuf serialization is also not stable/deterministic across builds. equals and hashCode work just fine within a single build.

If you only care about having a unique key within a single build, then relying on protobuf equals/hashCode is fine and is clearer in intent than comparing serialized values.

stevenwarejones (commented):
> @stevenwarejones reviewed 7 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4057)* status: all files reviewed, 1 unresolved discussion (waiting on kungfucraig and SanjayVas).

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/CreateBasicReportRequestValidation.kt` line 260 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4057#-Ovqrxb-9IAbF9UACqPE:-Ovu3z8J9ccxKhsCb8Xg:b-kghtrs) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/ea0d09e3de1177c96e9e3905c40d368d04709810/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/CreateBasicReportRequestValidation.kt#L260)):*

Protobuf serialization is also not stable/deterministic across builds. equals and hashCode work just fine within a single build.

If you only care about having a unique key within a single build, then relying on protobuf equals/hashCode is fine and is clearer in intent than comparing serialized values.
</blockquote></details>

good point. done.

SanjayVas (approved):
> @SanjayVas reviewed 5 files and all commit messages, made 1 comment, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4057)* status: all files reviewed, 1 unresolved discussion (waiting on kungfucraig and stevenwarejones).

___
*[`src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/CreateBasicReportRequestValidation.kt` line 262 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4057#-OvvWbYC5TMHKrrID3Nx:-OvvWbYC5TMHKrrID3Ny:bpx37gf) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/eb3c5e22f094eca8089c360c607f63b4dd36078f/src/main/kotlin/org/wfanet/measurement/reporting/service/api/v2alpha/CreateBasicReportRequestValidation.kt#L262)):*
> ```Kotlin
>    */
>   private data class ResultGroupSpecCollisionKey(
>     val reportingUnit: InternalReportingUnit,
> ```

nit: maybe document that these must be normalized?

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4057)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on kungfucraig).

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4057)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on kungfucraig).

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4057)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on kungfucraig).

Issue #3928, opened 2026-06-05
RankIndexBlobService implementation


Marco-Premier:
Implement the public and internal Kotlin services for RankIndexBlob, backed by the Spanner DAO.

PR #4058, opened 2026-06-23
feat: Add RankIndexBlob internal and public API

Marco-Premier:
## Summary

  Implements the `RankIndexBlob` resource for the EDP Aggregator, providing both the
  internal (Spanner-backed) and `v1alpha` public gRPC services. `RankIndexBlob` is a
  soft-deletable pointer to a rank-index blob in Cloud Storage along with its encrypted
  DEK, used by the VID labeler memoized-assignment pipeline to track, per `(upload,
  model line, subpool)`, the `DAY_ONLY` and `SNAPSHOT` blobs and their lifecycle.

  Stacked on `getina/vid-labeling-spanner-schema` — **set the base branch to
  `getina/vid-labeling-spanner-schema`, not `main`.**

  ## What's included

  - **Public API** (`v1alpha`): `RankIndexBlobService` with `CreateRankIndexBlob`,
    `BatchCreateRankIndexBlobs`, `GetRankIndexBlob`, `ListRankIndexBlobs`,
    `DeleteRankIndexBlob`, `BatchDeleteRankIndexBlobs`. Validates requests, maps resource
    names to internal resource IDs, and translates internal error reasons to gRPC statuses.
  - **Internal API**: `SpannerRankIndexBlobService` + `db/RankIndexBlob.kt`, persisting
    rows interleaved under `RawImpressionUpload`.
  - **Resource key**: `RankIndexBlobKey` / `RawImpressionUploadKey` and `IdVariable`
    entries.
  - **Errors**: `RawImpressionUploadNotFound`, `RankIndexBlobNotFound`,
    `RankIndexBlobAlreadyExists` reasons and exception types.
  - **Schema**: adds the `RankIndexBlobByNaturalKey` unique index enforcing one row per
    `(data provider, upload, model line, blob type, pool offset)`.
  - **Wiring**: registers the service in both internal and public `Services`, threading
    the shared `idGenerator`.

  ## Design notes
  
  - **`request_id` is REQUIRED** on `CreateRankIndexBlob` (and each batch sub-request) for
    at-least-once retry safety, consistent with the decision applied to the sibling job
    APIs (#4014, #4044, #4051, #4052). Idempotency is enforced both in-service
    (`findRankIndexBlobByRequestId`) and at the DB level (`CreateRequestId` unique index).
  - **Soft deletion** via `DeleteTime`; `Delete`/`BatchDelete` return the resource with
    `delete_time` populated. `show_deleted=true` on `List` returns active **and**
    soft-deleted rows per AIP-164.
  - This resource has **no `etag`** (it is an immutable blob pointer with no in-place
    mutation other than soft-delete), so the optimistic-concurrency pattern used by the
    job tables does not apply.
  - Pagination uses keyset ordering on `(CreateTime, RankIndexBlobResourceId)`, scoped to
    a single data provider; the upload segment may be the `-` wildcard to list across
    uploads (AIP-159).

  ## Testing

  - New abstract `RankIndexBlobServiceTest` with the `SpannerRankIndexBlobServiceTest`
    concrete subclass, plus the public `RankIndexBlobServiceTest`.
  - Coverage includes create/batch idempotency, natural-key `ALREADY_EXISTS`, validation
    failures, filtering, `show_deleted` (active + deleted), pagination including rows that
    share a `CreateTime` (resource-ID tie-breaker), wildcard listing, soft-delete, and
    batch parent/name mismatch and negative `page_size`.

  //src/test/.../spanner:SpannerRankIndexBlobServiceTest    PASSED
  //src/test/.../service/v1alpha:RankIndexBlobServiceTest   PASSED

Issue: #3928


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4058)

src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/spanner/SpannerRankIndexBlobService.kt
```diff
@@ -0,0 +1,606 @@
...
+        rankIndexBlobResourceId,
+      )
+        ?: throw RankIndexBlobNotFoundException(
+            dataProviderResourceId,
+            rawImpressionUploadResourceId,
+            rankIndexBlobResourceId,
+          )
+          .asStatusRuntimeException(Status.Code.NOT_FOUND)
+
+    if (result.rankIndexBlob.hasDeleteTime()) {
```
stevenwarejones, line 527: Make Delete idempotent on an already-soft-deleted row. Today a successful Delete followed by a retry returns NOT_FOUND, indistinguishable from 'resource never existed' — a client retrying after a network blip can't tell whether their original call landed.

Per AIP-135: 'calling Delete on an already-soft-deleted resource should succeed (and the resource will appear unchanged).' Remove the throw and return the existing row:

```
if (result.rankIndexBlob.hasDeleteTime()) {
  return@run result.rankIndexBlob  // already deleted; idempotent no-op
}
txn.softDeleteRankIndexBlob(...)
return@run result.rankIndexBlob
```

The caller's `deletedBlob.copy { deleteTime = commitTimestamp }` at line 369 needs to become conditional on whether the row was already deleted — preserve the original `delete_time` on replay.

Apply the same change to `batchDeleteRankIndexBlobs` at Spanner.kt:442-449: don't throw NOT_FOUND for already-deleted rows; include them in the response with their original `delete_time`. Distinguish 'never existed' (NOT_FOUND, correct) from 'already deleted' (idempotent success).

Add tests: Delete → Delete (same name) succeeds twice; BatchDelete with mix of active and already-deleted rows returns all of them.

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4058)* status: 0 of 27 files reviewed, 1 unresolved discussion (waiting on getina and Marco-Premier).

stevenwarejones (commented):
> @stevenwarejones reviewed 27 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4058)* status: all files reviewed, 1 unresolved discussion (waiting on getina and Marco-Premier).

src/main/kotlin/org/wfanet/measurement/edpaggregator/service/v1alpha/RankIndexBlobService.kt
```diff
@@ -0,0 +1,561 @@
...
+
+    val internalResponse: InternalListResponse =
+      try {
+        internalRankIndexBlobStub.listRankIndexBlobs(
+          internalListRequest {
+            dataProviderResourceId = uploadKey.dataProviderId
+            rawImpressionUploadResourceId =
+              if (uploadKey.rawImpressionUploadId == ResourceKey.WILDCARD_ID) ""
+              else uploadKey.rawImpressionUploadId
+            this.pageSize = request.pageSize
```
stevenwarejones: Coerce `pageSize` at the public layer to match repo precedent. `SpannerImpressionMetadataService` in main coerces at the public layer (ImpressionMetadataService.kt:604-609 in main): `0 -> DEFAULT_PAGE_SIZE`, `> MAX_PAGE_SIZE -> MAX_PAGE_SIZE`. #4058 passes `request.pageSize` raw to the internal stub and relies on the internal layer to coerce (Spanner.kt:313-314). Public layer is the validation boundary; coerce there.

```
val pageSize =
  if (request.pageSize == 0) DEFAULT_PAGE_SIZE
  else request.pageSize.coerceAtMost(MAX_PAGE_SIZE)

// then:
this.pageSize = pageSize
```

Add `DEFAULT_PAGE_SIZE = 50` and `MAX_PAGE_SIZE = 100` to the companion (already present in the internal layer; lift to public to match).

src/main/kotlin/org/wfanet/measurement/edpaggregator/service/v1alpha/RankIndexBlobService.kt
```diff
@@ -0,0 +1,561 @@
...
+        InternalErrors.Reason.REQUISITION_METADATA_NOT_FOUND,
+        InternalErrors.Reason.REQUISITION_METADATA_NOT_FOUND_BY_CMMS_REQUISITION,
+        InternalErrors.Reason.REQUISITION_METADATA_ALREADY_EXISTS,
+        InternalErrors.Reason.REQUISITION_METADATA_ALREADY_EXISTS_BY_BLOB_URI,
+        InternalErrors.Reason.REQUISITION_METADATA_ALREADY_EXISTS_BY_CMMS_REQUISITION,
+        InternalErrors.Reason.REQUISITION_METADATA_STATE_INVALID,
+        InternalErrors.Reason.ETAG_MISMATCH,
+        InternalErrors.Reason.REQUIRED_FIELD_NOT_SET,
+        InternalErrors.Reason.INVALID_FIELD_VALUE,
+        null -> Status.INTERNAL.withCause(e).asRuntimeException()
```
stevenwarejones, line 477: Map `REQUIRED_FIELD_NOT_SET` and `INVALID_FIELD_VALUE` to `INVALID_ARGUMENT`, not `INTERNAL`. The current catch arm routes both into the `null -> INTERNAL` branch — if the internal layer ever throws one of these (e.g. via a future change that adds internal validation past what the public layer covers), the client sees `INTERNAL` rather than the documented `INVALID_ARGUMENT`.

```
InternalErrors.Reason.REQUIRED_FIELD_NOT_SET,
InternalErrors.Reason.INVALID_FIELD_VALUE ->
  Status.INVALID_ARGUMENT.withCause(e).asRuntimeException()
```

Same shape as the bug fixed on #4018 / #4044.

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4058)* status: all files reviewed, 3 unresolved discussions (waiting on getina and Marco-Premier).

src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/spanner/SpannerRankIndexBlobService.kt
Marco-Premier: Done.

src/main/kotlin/org/wfanet/measurement/edpaggregator/service/v1alpha/RankIndexBlobService.kt
Marco-Premier: Done.

src/main/kotlin/org/wfanet/measurement/edpaggregator/service/v1alpha/RankIndexBlobService.kt
Marco-Premier: Done.

Marco-Premier (commented):
> @Marco-Premier made 3 comments.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4058)* status: 24 of 27 files reviewed, 3 unresolved discussions (waiting on getina, stevenwarejones, and stevenwarejones+AGNT).

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4058)* status: all files reviewed, 3 unresolved discussions (waiting on getina and stevenwarejones+AGNT).

src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/spanner/SpannerRankIndexBlobService.kt
stevenwarejones: Verified in commit a485e21c5. `deleteOneRankIndexBlob` at Spanner.kt:539-541 returns the already-soft-deleted row unchanged for idempotency. Caller at line 370-375 preserves the original `delete_time` rather than overwriting it. BatchDelete at lines 458-464 + 470-476 applies the same idempotency: already-deleted rows skip re-deletion and keep their original delete_time in the response. AIP-135 compliant. NOT_FOUND is now only returned for rows that never existed.

src/main/kotlin/org/wfanet/measurement/edpaggregator/service/v1alpha/RankIndexBlobService.kt
stevenwarejones: Verified in commit a485e21c5. `pageSize` is now coerced at the public layer (RankIndexBlobService.kt:217-219): `0 -> DEFAULT_PAGE_SIZE`, `coerceAtMost(MAX_PAGE_SIZE)`, then passed to the internal stub at line 246. Matches the established pattern from `ImpressionMetadataService` in main.

src/main/kotlin/org/wfanet/measurement/edpaggregator/service/v1alpha/RankIndexBlobService.kt
stevenwarejones: Verified in commit a485e21c5. `REQUIRED_FIELD_NOT_SET` and `INVALID_FIELD_VALUE` now map to `INVALID_ARGUMENT` (RankIndexBlobService.kt:434-436), separated from the `null -> INTERNAL` arm. Any internal-layer validation failure now surfaces correctly to clients.

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4058)* status: all files reviewed, 3 unresolved discussions (waiting on getina and stevenwarejones).

stevenwarejones (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4058#-:-Ovq-7lDEswgSywvonXU:bnfp4nl)* status: all files reviewed, 3 unresolved discussions (waiting on getina).

stevenwarejones:
**Issue D1 (blocking, design-doc cross-reference):** `blob_checksum` should be REQUIRED.

`rank_index_blob.proto:88-91` currently marks `blob_checksum` as OPTIONAL. The design doc (Lifecycle Management § Corrupted cumulative blob) specifies SHA-256 checksum is persisted at write and validated on every load — that is the corruption-detection mechanism for the cumulative blob. #4017 SubpoolRanker.kt:211-216 always passes the checksum to `rankIndexStore.readBlob`. If a writer omits the checksum, the integrity validation is silently skipped on reload, defeating the heal-rank-index recovery path described in the doc.

Fix: change field_behavior to `REQUIRED` and add a runtime guard in `validateCreateRequest` (Public.kt) + `validateRankIndexBlob` (Spanner.kt) that rejects empty `blob_checksum` with `INVALID_ARGUMENT`.

---

**Issue D2 (blocking, design-doc cross-reference):** `max_event_date` KDoc is wrong and should be REQUIRED on SNAPSHOT.

`rank_index_blob.proto:99-104` says "Only set for DAY_ONLY blobs; unset for SNAPSHOT blobs." This contradicts the design doc (Rank index blob persistence & life cycle § SNAPSHOT rows): "SNAPSHOT rows also carry MaxEventDate (the newest event-date represented in the cumulative), used as the selection key to locate the 'old' snapshot for a given event-date during backfill."

#4017 SubpoolRanker.kt:74-79 + line 443 explicitly select the backfill old-snapshot via the `max_event_date_on_or_before` filter. If SNAPSHOTs don't carry max_event_date, the filter never matches and backfill silently falls through to forward-append, corrupting cumulative rank state.

Fix the KDoc:

```
// The newest event date among the impressions in the blob (for DAY_ONLY) or
// represented in the cumulative (for SNAPSHOT, inherited from the prior snapshot
// on backfill). Required on both blob types — DAY_ONLY uses it for retention,
// SNAPSHOT uses it for backfill 'old snapshot' selection (see VidRankBuilder).
google.type.Date max_event_date = 8 [
  (google.api.field_behavior) = REQUIRED,
  (google.api.field_behavior) = IMMUTABLE
];
```

Add runtime guards in `validateCreateRequest` (Public.kt) + `validateRankIndexBlob` (Spanner.kt): reject missing `max_event_date` with `INVALID_ARGUMENT`.

Posting as a top-level comment because both fields existed in the base PR and weren't modified in this PR's diff (line-anchor inline comments require touched lines).

stevenwarejones (changes requested):
> @stevenwarejones made 1 comment and resolved 3 discussions.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4058)* status: all files reviewed, 1 unresolved discussion (waiting on getina and Marco-Premier).

___
*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4058#-Ovq3e9KEFmE9jVjohIv:-Ovq3e9KEFmE9jVjohIw:bosy10f) (no related file):*
See two new blocking issues

Marco-Premier (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4058)* status: 22 of 28 files reviewed, 1 unresolved discussion (waiting on getina and stevenwarejones).

___
*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4058#-Ovq3e9KEFmE9jVjohIv:-OvsXmiU7OBmXMOYqEp_:b-896fix) (no related file):*

See two new blocking issues
</blockquote></details>

Done.

src/main/resources/edpaggregator/spanner/create-vid-labeling-schema.sql
```diff
@@ -0,0 +1,292 @@
...
+  RawImpressionUploadModelLineId INT64 NOT NULL,
+  RawImpressionUploadModelLineResourceId STRING(63) NOT NULL,
+  CmmsModelLine STRING(MAX) NOT NULL,
+  CreateRequestId STRING(36),
+  State `wfa.measurement.internal.edpaggregator.RawImpressionUploadModelLineState` NOT NULL,
+  ErrorMessage STRING(MAX),
+  PoolOffsets ARRAY<INT64>,
+  MaxEventDate DATE,
+  EncryptedMergedDek BYTES(MAX),
+  Etag STRING(36) NOT NULL,
```
stevenwarejones: Rebase needed — this branch is on a stale snapshot of `getina/vid-labeling-spanner-schema`. The base branch has added `MarkRequestId` columns + matching unique indexes on RawImpressionUploadModelLine, PoolAssignmentJob, RankerJob, and VidLabelingJob since this rebase. A `git fetch origin getina/vid-labeling-spanner-schema && git rebase origin/getina/vid-labeling-spanner-schema` should bring them back.

stevenwarejones (commented):
Rebase needed.

stevenwarejones (commented):
D1 & D2 look good now. just the rebase.

> @stevenwarejones reviewed 4 files and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4058#-:-Ovu30Sk0mXeA52Mgpz_:b-cy7ym5)* status: 25 of 29 files reviewed, 2 unresolved discussions (waiting on getina and Marco-Premier).

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4058)* status: all files reviewed, 2 unresolved discussions (waiting on getina and Marco-Premier).

src/main/resources/edpaggregator/spanner/create-vid-labeling-schema.sql
Marco-Premier: Done.

Marco-Premier (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4058)* status: all files reviewed, 2 unresolved discussions (waiting on getina and stevenwarejones).

stevenwarejones (approved):
:lgtm:

> @stevenwarejones reviewed 2 files and all commit messages, made 1 comment, and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4058#-:-OvuApMx2ZIJu7VJM7DV:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on getina and SanjayVas).

SanjayVas (changes requested):
> @SanjayVas reviewed 4 files and made 3 comments.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4058)* status: all files reviewed, 3 unresolved discussions (waiting on getina and Marco-Premier).

___
*[`src/main/proto/wfa/measurement/edpaggregator/v1alpha/rank_index_blob.proto` line 72 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4058#-Ovvt-sfC4c7SOfgNPr9:-Ovvspxx0ERmUigZlCbg:b-5rthk5) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/65822f09b7e90783f74d3b7d5c88f8ef3af3b94a/src/main/proto/wfa/measurement/edpaggregator/v1alpha/rank_index_blob.proto#L72)):*
> ```proto
>   // The pool offset this blob belongs to.
>   //
>   // Marked REQUIRED, but note that 0 is a valid pool offset and is also the
> ```

This is incorrect. REQUIRED in AIP terms means that the value must not be the default. See https://google.aip.dev/203#required

You an use the `optional` keyword to allow distinguishing between unset and 0, where the `REQUIRED` field behavior can be satisfied using `0` as the value *and* the server can enforce this requirement.
___
*[`src/main/proto/wfa/measurement/edpaggregator/v1alpha/rank_index_blob_service.proto` line 74 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4058#-Ovvu4XH04rC9iU6ojum:-Ovvu4XI7Ae7q35IvIr0:bg105i7) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/65822f09b7e90783f74d3b7d5c88f8ef3af3b94a/src/main/proto/wfa/measurement/edpaggregator/v1alpha/rank_index_blob_service.proto#L74)):*
> ```proto
>   // This field must be a UUID4 (RFC 4122) string.
>   string request_id = 3 [
>     (google.api.field_behavior) = REQUIRED,
> ```

nit: same note as the other PR where this should always be optional.
___
*[`src/main/resources/edpaggregator/spanner/create-vid-labeling-schema.sql` line 268 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4058#-OvvuNi25FQn4CZk_rrR:-OvvuNi25FQn4CZk_rrS:b-x6a383) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/65822f09b7e90783f74d3b7d5c88f8ef3af3b94a/src/main/resources/edpaggregator/spanner/create-vid-labeling-schema.sql#L268)):*
> ```TSQL
>   ON RankIndexBlob(DataProviderResourceId, CmmsModelLine, BlobType, RawImpressionUploadId);
> 
> -- Enforces the design invariant of at most one row per
> ```

You cannot modify an existing changeset if it has already been applied in any environment, including halo-cmm-dev. You must add this as a new entry in the changelog.

Marco-Premier (commented):
> @Marco-Premier made 3 comments.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4058)* status: 22 of 30 files reviewed, 3 unresolved discussions (waiting on getina, SanjayVas, and stevenwarejones).

___
*[`src/main/proto/wfa/measurement/edpaggregator/v1alpha/rank_index_blob.proto` line 72 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4058#-Ovvt-sfC4c7SOfgNPr9:-OvyhYBH58NjWrF7S-Uz:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/65822f09b7e90783f74d3b7d5c88f8ef3af3b94a/src/main/proto/wfa/measurement/edpaggregator/v1alpha/rank_index_blob.proto#L72)):*

This is incorrect. REQUIRED in AIP terms means that the value must not be the default. See https://google.aip.dev/203#required

You an use the `optional` keyword to allow distinguishing between unset and 0, where the `REQUIRED` field behavior can be satisfied using `0` as the value *and* the server can enforce this requirement.
</blockquote></details>

Done.
___
*[`src/main/proto/wfa/measurement/edpaggregator/v1alpha/rank_index_blob_service.proto` line 74 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4058#-Ovvu4XH04rC9iU6ojum:-OvyeHkq8dF2A_s6d-a_:b-a8ie5k) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/65822f09b7e90783f74d3b7d5c88f8ef3af3b94a/src/main/proto/wfa/measurement/edpaggregator/v1alpha/rank_index_blob_service.proto#L74)):*

nit: same note as the other PR where this should always be optional.
</blockquote></details>

This is required as per the other PR:  #4044 / #4078 as previously discussed with @stevenwarejones 
___
*[`src/main/resources/edpaggregator/spanner/create-vid-labeling-schema.sql` line 268 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4058#-OvvuNi25FQn4CZk_rrR:-OvyhZbL9rbOfCh-CicX:bq5ywgd) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/65822f09b7e90783f74d3b7d5c88f8ef3af3b94a/src/main/resources/edpaggregator/spanner/create-vid-labeling-schema.sql#L268)):*

You cannot modify an existing changeset if it has already been applied in any environment, including halo-cmm-dev. You must add this as a new entry in the changelog.
</blockquote></details>

reverted this file as it's not part of this pr but rather #3989

stevenwarejones (commented):
> @stevenwarejones reviewed 19 files and all commit messages, and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4058)* status: 29 of 30 files reviewed, 3 unresolved discussions (waiting on getina, Marco-Premier, and SanjayVas).

___
*[`src/main/proto/wfa/measurement/edpaggregator/v1alpha/rank_index_blob.proto` line 72 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4058#-Ovvt-sfC4c7SOfgNPr9:-OvyjDzj5PMhzYZr7CAb:b-afc7jl) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/65822f09b7e90783f74d3b7d5c88f8ef3af3b94a/src/main/proto/wfa/measurement/edpaggregator/v1alpha/rank_index_blob.proto#L72)):*

Done.
</blockquote></details>

thx for catching this.
___
*[`src/main/proto/wfa/measurement/edpaggregator/v1alpha/rank_index_blob_service.proto` line 74 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4058#-Ovvu4XH04rC9iU6ojum:-OvyjGeg1faY_bpApdGG:b-4giil5) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/65822f09b7e90783f74d3b7d5c88f8ef3af3b94a/src/main/proto/wfa/measurement/edpaggregator/v1alpha/rank_index_blob_service.proto#L74)):*

This is required as per the other PR:  #4044 / #4078 as previously discussed with @stevenwarejones
</blockquote></details>

@SanjayVas - vid labeling needs required request_ids almost everywhere since its a giant state machine that also needs to deal with multiple pub/sub re-delivery, not knowing who lost the race, etc.

SanjayVas (changes requested):
> @SanjayVas reviewed 28 files and all commit messages, made 2 comments, and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4058)* status: all files reviewed, 2 unresolved discussions (waiting on getina, Marco-Premier, and stevenwarejones).

___
*[`src/main/proto/wfa/measurement/edpaggregator/v1alpha/rank_index_blob_service.proto` line 74 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4058#-Ovvu4XH04rC9iU6ojum:-OvzjwDVBrBMgYJpE2Ht:bv3ea11) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/65822f09b7e90783f74d3b7d5c88f8ef3af3b94a/src/main/proto/wfa/measurement/edpaggregator/v1alpha/rank_index_blob_service.proto#L74)):*

@SanjayVas - vid labeling needs required request_ids almost everywhere since its a giant state machine that also needs to deal with multiple pub/sub re-delivery, not knowing who lost the race, etc.
</blockquote></details>

This is a design philosophy issue w.r.t. separation of concerns[^1]. An API should be defined with no knowledge of how its callers are implemented. IMO the request ID feature should be optional on the service level such that the caller chooses whether to use it. You could *theoretically* have a caller that has its own mechanism to ensure it doesn't make the same request more than once.

In this case it happens to be that we control all the callers since it's a system API, thus we can ensure that all of our calling code uses it. That said I can see the argument for making this required: Since this is a system API where we control the callers we can be more prescriptive on how they must behave. Hence why I've left this as an optional non-blocking comment.

Note that the same principle applies to documentation: The API documentation should not reference anything about specific caller implementations. If caller behavior changes but API behavior does not, then there should be no change needed to keep API documentation in sync. The related principle is keeping all of the documentation in the right place, which is why I was concerned about the [extraneous documentation in ranker_job_service.proto](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4052#-Ovw6KYI3VZJ5a1SnWTn:-Ovw6KYI3VZJ5a1SnWTo:byy8vdu) and about duplicated documentation in the PR to update AGENTS.md.

[^1]: https://en.wikipedia.org/wiki/Separation_of_concerns
___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/spanner/db/RankIndexBlob.kt` line 70 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4058#-Ovzve1FD70QY89wMZf-:-Ovzve1GEHg2oAu_wKzV:bf71r9l) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3fbcb1dcc80389892237068f89445520b37b5b38/src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/spanner/db/RankIndexBlob.kt#L70)):*
> ```Kotlin
>  */
> suspend fun AsyncDatabaseClient.ReadContext.getRawImpressionUploadIdForRankIndexBlob(
> ```

Same note as https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4051#-OvvxlLg5Imb0g9SSfIC:-OvvxlLhAn0qlpJJj2bx:b-7s4dj3

This should be defined in RawImpressionUpload.kt. In fact, it looks like the same function as in that comment.



_Code quote:_
```Kotlin
getRawImpressionUploadIdForRankIndexBlob
```

Marco-Premier (commented):
> @Marco-Premier made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4058)* status: 23 of 32 files reviewed, 2 unresolved discussions (waiting on getina, SanjayVas, and stevenwarejones).

___
*[`src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/spanner/db/RankIndexBlob.kt` line 70 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4058#-Ovzve1FD70QY89wMZf-:-Ow2wWI44uidcVoYJJER:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/3fbcb1dcc80389892237068f89445520b37b5b38/src/main/kotlin/org/wfanet/measurement/edpaggregator/deploy/gcloud/spanner/db/RankIndexBlob.kt#L70)):*

Same note as https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4051#-OvvxlLg5Imb0g9SSfIC:-OvvxlLhAn0qlpJJj2bx:b-7s4dj3

This should be defined in RawImpressionUpload.kt. In fact, it looks like the same function as in that comment.
</blockquote></details>

Done.
___
*[`src/main/proto/wfa/measurement/edpaggregator/v1alpha/rank_index_blob_service.proto` line 74 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4058#-Ovvu4XH04rC9iU6ojum:-Ow2wsNuBXnOg-L2pg5C:b68ck8y) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/65822f09b7e90783f74d3b7d5c88f8ef3af3b94a/src/main/proto/wfa/measurement/edpaggregator/v1alpha/rank_index_blob_service.proto#L74)):*

This is a design philosophy issue w.r.t. separation of concerns[^1]. An API should be defined with no knowledge of how its callers are implemented. IMO the request ID feature should be optional on the service level such that the caller chooses whether to use it. You could *theoretically* have a caller that has its own mechanism to ensure it doesn't make the same request more than once.

In this case it happens to be that we control all the callers since it's a system API, thus we can ensure that all of our calling code uses it. That said I can see the argument for making this required: Since this is a system API where we control the callers we can be more prescriptive on how they must behave. Hence why I've left this as an optional non-blocking comment.

Note that the same principle applies to documentation: The API documentation should not reference anything about specific caller implementations. If caller behavior changes but API behavior does not, then there should be no change needed to keep API documentation in sync. The related principle is keeping all of the documentation in the right place, which is why I was concerned about the [extraneous documentation in ranker_job_service.proto](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4052#-Ovw6KYI3VZJ5a1SnWTn:-Ovw6KYI3VZJ5a1SnWTo:byy8vdu) and about duplicated documentation in the PR to update AGENTS.md.

[^1]: https://en.wikipedia.org/wiki/Separation_of_concerns
</blockquote></details>

thanks @SanjayVas I see your point which makes totally sense, I am also incline to keep the `require`. I just went through all the state-change api call in the entire pipeline and I think having it requires helps a lot as it surfaces in many different places.

Regarding the documentation you mentioned, I am not sure I follow, are there still open gaps there for this PR?

thanks

SanjayVas (approved):
> @SanjayVas reviewed 26 files and all commit messages, made 1 comment, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4058)* status: all files reviewed, 1 unresolved discussion (waiting on getina, Marco-Premier, and stevenwarejones).

___
*[`src/main/proto/wfa/measurement/edpaggregator/v1alpha/rank_index_blob_service.proto` line 74 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4058#-Ovvu4XH04rC9iU6ojum:-Ow4JsJQ7nj-yMBBRuDR:bambuan) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/65822f09b7e90783f74d3b7d5c88f8ef3af3b94a/src/main/proto/wfa/measurement/edpaggregator/v1alpha/rank_index_blob_service.proto#L74)):*
> Regarding the documentation you mentioned, I am not sure I follow, are there still open gaps there for this PR?

Nothing is immediately jumping out in this PR, but I had seen some comments before along the lines of "<specific caller> uses this to do X"

Marco-Premier (commented):
> @Marco-Premier reviewed 26 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4058)* status: all files reviewed, 1 unresolved discussion (waiting on getina and stevenwarejones).

Marco-Premier (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement/4058)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on getina and stevenwarejones).

